PR #340, opened 2019-04-23
Extend lub/glb to more types

ggreif:
This is a peek at the final version and request for comment on the further direction.

Added to the type algebra:
* tuples
* objects
* arrays
* functions
* `Shared`
* futures

Laundry list:

- [x] test `glb` for variants
- [x] remove `TBR`?
- [ ] ~consider `var` in objects, arrays (do they add indeterminism/effects?)~
- [x] are there any other meaningful points in the lattice?
- [ ] ~move the clause(s) `t1', t2' when eq t1' t2' -> t1` further up?~ Costly
- [x] alpha-equivalence in polymorphic function types, need to `open`
- [x] go contravariantly with `Func` type param bounds
- [ ] can we relax _sharing_ (resp. _control_) mode checking for functions?
- [x] style issue: make the use of `t1'` etc. consistent in touched functions
- [x] recursion in types
- [x] ~make `is_recursive_con` more precise~ Not needed any more
- [x] `@glb (Int, Nat)/2` appears in types, sometimes we need to reduce!
- [ ] check whether we need to observe polarity for bounded params


rossberg (commented):
Yes, there are quite a few more cases, just follow the cases for subtyping.

ggreif (commented):
Handling `Mut` is the only open question. ~Apart from the suggested `lazy` mechanics, which are admittedly gross, but all other alternatives I have considered are worse.~
_Update_: The only effect base I make use of now is the rewriting of `Con` kinds, as done in other places too.

rossberg (commented):
Hm... I don't understand why we need all this laziness and loop breaker stuff. It should suffice to create a new Def con (which essentially is a mu) when you encounter one (or two). You initialise it with type Pre and update it on the way back. In all cases where neither of the two types is a Def con you do not need to do anything special.

crusso (commented):
Actually, I think we could allow [mut T] <: [T] if we make [-] invariant.... but then [T] wouldn't be immutable but statefull and just readonly. Let's not go there....

ggreif (commented):
Almost there (I hope).

crusso (changes requested):
I wonder it's worth adding some tests that test lub and glub involving (bounded) type parameters (unless you already have some)

rossberg (commented):
This is going in the right direction, but the Con case still seems wrong for abstract constructors. See comments.

crusso (approved):
I'm happy to merge provide you add some TODOs for the remaining issues (possibly linked to a single JIRA or GITHUB tracking item)

rossberg (approved):
Thanks, LGTM

Issue #400, opened 2019-05-13
Syntactic form of actor files


nomeata:
## Assumptions:

To deploy an actor to the platform, we want a WebAssembly module that manifestly exports all public message entry points (i.e. we no longer pursue the upload-code-that-installs-an-actor-idea).

The system initializes an actor using an explicit function call to a `"start"` function (NB: not the `(start)` function in the WebAssembly module).

This initialization should support passing arguments, so that a single module can be configured upon instantiations.

## Requirements:

It should be possible to use the same `.as` file to produce a deployable WebAssembly module, and to import it from ActorScript code, e.g. for testing (or eventually to install it as a dynamic actor). The lack of that ability currently bites @matthewhammer.

It should be possible express modules that are passed values upon initialization in ActorScript.

A nice error message needs to be produced when the files are not of the expected format.

## Status quo:

We currently have a hack in the compiler that expects the actor file to end with a manifest `actor` expression or declaration. If no such expression is there, then `actor {}` is assumed. This fulfills neither of the above requirements.

## Proposal:

When compiling for the DFINITY platform, the ActorScript source is expected to end with a manifest actor class declaration. The parameters of the class are the parameters expected by the WebAssembly module upon initialization. As a short-hand, the file may also end with a manifest actor (implicitly turning it into a class with a parameter of type `()`), or an expression of type `()` (implicitly appending `actor {}`). As before, all declarations that come before the actor class are implicitly moved into the class, as private declarations. The compiler checks these syntactic requirements during type-checking or earlier.

## Rationale:

Because class declarations are static, this allows the same file to serve as a deployable actor and as an importable library. It also provides an intuitive way of specifying configurable canisters.

## Issues

actor class declarations have a name; if that name is used to dynamically instantiate the actor again, we have a problem. We can probably implement that now already (at the cost of embedding an almost identical copy of the WebAssembly module in its own data section), but I would propose that we statically rule it out for now (“Error: top-level actor cannot instantiate itself again.”), and if we do need this, do it with system support (in the current System API we could have a `mod.self : () -> (modref)` call; in the eventual System API this might look different.)

PR #452, opened 2019-05-29
Units design proposal

rossberg:
Okay, this proposal attempts to define a simple and coherent model for source files, import semantics, and compilation, It is intended to address #400 and other related discussion; it's a sort of remix of the approaches already discussed in that thread. Please take a look.


nomeata (commented):
Thanks for starting to write this up.

I am slightly disappointed that you don’t even acknowledge the work that the rest of the team poured into the [current `import` design](https://dfinity.atlassian.net/wiki/spaces/AST/pages/147357892/ActorScript+library+import), including some intense hours in Zurich, and that design was not callenged more timely. It also “ makes each of these use cases possible and convenient while unifying them as much as possible.”, maybe even more uniformly.

matthewhammer (approved):
LGTM.

My only outstanding comments are about the precise compiler flags we choose.  I like @nomeata's suggestion of distinguishing the three conceptual cases, since then the compiler can know the user's intention with the output, and issue more helpful messages.  I don't see the benefit to the alternative, where one or more flags are conflated.

nomeata (changes requested):
(I guess I should make it explicit that there are open issues, lest it gets merged accidentally because of one approving review.)

crusso (changes requested):
I think this still need some iteration.

FTR, I still don't see what is wrong with the existing import mechanism Joachim came up with, which is clean yet expressive. There is no reason to require an import to be a module - all we need it to be is pure and the same mechanism that is used to reconstruct a module value from a wasm module, which will need to be module-type directed, can be used to construct any old value from a wasm module and an arbitrary AS type. With Joachim's proposal, pattern matching on libraries comes for free too and doesn't need to be built into the syntax.

I also don't grok the objection to requiring `module { ... }` around library implementations (when they are, in fact, literal modules). An advantage of this that the implementation could also be just a projection or, if extend the notion of static appropriately, an application of a pure function (perhaps for pure functors).
Provided the syntax for fields and expressions doesn't overlap then we could even treat the fields* as shorthand for `module { fields* }` yet still support arbitrary (static) expressions as libraries).

On the other hand, I agree that the use case for importing non-module libraries is quite limited, but I don't think it's fundamentally flawed. 

I sympathise with @kRITZCREEK  that prefix `import` makes the IDE problem a little simpler, but a simple lex of the file for `IMPORT STRING` token sequences should be sufficient yet much quicker than a full parse.

matthewhammer (changes requested):
Good stuff!  I have two minor suggestions.

luc-blaeser:

PR #459, opened 2019-05-29
Failing test case related to large Word64 literals

nomeata:
and their conversions in the interpreter.


ggreif (approved):
It is okay to check in these test cases to capture the status quo.

However we need to know that conversions are a bit ill-defined at the moment because we do not trap correctly. I.e. `0xFFFFFFFFFFFFFFFF` -> `Nat` -> `Int` (by subtyping) -> `Word64` may return the wrong (unexpected) result.

nomeata (commented):
I can’t approve this, because I opened this. But I do approve it. So gabor, just approve it yourself and press the merge button.

PR #480, opened 2019-06-06
[AST-77][AST-89] Fix correct check_ir.check_pat

crusso:
correct check_ir.check_pat subtyping checks; fix bug in source intepretation of variants; add test

Note the coverage checker complains about some patterns that seem complete to me. Gabor, please take a look at the test output. I expect the coverage checker might not be taking account of subtyping...


ggreif (approved):
Looks good to me. Some comments. The important point is that the front-end and IR type-checkers don't contradict each other. Otherwise the users find themselves [between a rock and a hard place](https://media1.tenor.com/images/cd91daaa4f689b8c27073caaf11271c3/tenor.gif).

PR #517, opened 2019-06-21
Describe separate the compilation schemes

nomeata:
this doesn’t really change the meaning of the document, but describes
the intermediate step a with a bit more detail. This can be seen as a
compromise that I can live with being merged into `master`.


matthewhammer (commented):
Noted minor typos; added a question about release times.

PR #729, opened 2019-10-04
flatten async returns types in IDL and compiler ACT-138

crusso:
Fixes the hack that was PR #731

- Use Promise control flag to track syntactic arity of async returns.
- Propagate to back-end and generated IDL
- Flatten accordingly using modified ICReplyPrim taking list of types
- Add test idl-nary test and revised expected drun-test output.

@nomeata, please verify the test outputs make sense - I think we were systematically boxing the replies in a one tuple; not anymore.


ggreif (commented):
I only looked at the results. One comment below. Also one question: For query methods will we have `: async ()` as default return (instead of `: ()`)?

nomeata (approved):
Approving as quickfix

ggreif (commented):
Oops I missed one.

PR #745, opened 2019-10-08
Motoko-IDL field escaping proofs

nomeata:
these functions are so small it would be embarrassing if they would not
hold their promises. So I did a pen-and-pencil proof of the necessary
properties.


nomeata:
@ggreif should we merge this? Or maybe wait until October to be able to claim that we have merged a 2 year old PR :-)

nomeata:
@ggreif ping?

ggreif (approved):
LGTM!

PR #901, opened 2019-11-21
Definedness: Delayify the body of `actor {…}`

### LINKED ISSUE #900: Should actor be allowed to use its self reference eagerly?
We assume actor creation is synchronous at the moment. So is there a reason to forbid
```
> actor a { public func foo(){}; a.foo() };
stdin:1.1-1.41: definedness error, cannot use a before a has been defined
```
when we allow
```
> actor a { public func foo(){}; foo() };
let a : actor {foo : shared () -> ()} = {foo = func}
```
(This may be not intentional and might just require some relaxation in the definedness checker.)

nomeata:
this fixes #900
```
> actor a { public func foo(){}; a.foo() };
stdin:1.1-1.41: definedness error, cannot use a before a has been
defined
```

As the test case output shows, the interpreter(s) are not doing this
properly: They evaluate the body of `actor {…}` eagerly. I noticed this
previously in
https://github.com/dfinity-lab/motoko/pull/899/commits/a69230ca6e40e4d054a81f794b88d422c041eed3

Interpreting `actor {…}` should synchronosly create something you can
use as an actor, but it should _schedule_ its initialization.

Can someone take over that part? @crusso?


nomeata:
I assume this is not needed any more. Or not until we have recursive actors…

PR #924, opened 2019-11-25
fix trapping behaviour of intToWord64

ggreif:
(it should wrap, of course)

Eliminates the test failure in `test/run/ok/conversions.wasm-run.ok`.


nomeata (approved):
Yay!

PR #1105, opened 2020-01-13
More extra-rounds

nomeata:
we have to tell `drun` to wait for a certain amount of time (rounds)
after each message to allow for subsequent computation in the call tree.
This bumps this number to something that might cover all our test cases (even
`flatten-awaitables` and `large-array.)

This hack should go away with DFN-1269.


ggreif (approved):
LGTM! I had these discrepancies locally, hopefully the are eliminated now. Will we get triggered by DFN-1269?

PR #1682, opened 2020-07-01
Use Nat8 for floatToFormattedText

ggreif:
this is just a small tweak, no functionality change

Both latter args of `floatToFormattedText ` were `Word8`, now `Nat8`. Shorter, slightly better semantics.


crusso (approved):
Still don't like the fmtFloat->Text name. I agree its arbitrary, but it would be easier to grep for things if names followed some pattern...

PR #1865, opened 2020-08-21
[Draft] Unambiguous record syntax, take 1

rossberg:
First attempt at reconciling record object syntax, by requiring labels to be written with a dot. But I am hesitant to land this change. 

After some getting used to it, the clutter in literals turns out to be not such a big deal. And there is some precedent for this syntax in other languages. After adjusting all the test suite, it feels like it notably improves readability. The particularly ugly empty record rarely occurs outside artificial tests.

However: it would be inconsistent not to adjust the type syntax in the same manner. And from my first go at trying that, it starts looking really ugly and cluttered. :(

Uploading the experiment so far, so you can see how it affects code, in form of the test suite. Will experiment with prefix syntax (`.{…}`) next (cf. #1870).

List of changes:

* Change record-style object syntax short-hand to be unambiguous with blocks, by requiring a dot before each label:
  `{.a = 1; .b = 2}`

* Empty records are written `{.}`, analogous to empty variant types.

* Record labels are no longer bindings, so they are non-recursive:
  `let x = 0; let r = {.x = x}  // now works as expected`

* Allow record punning short-hand, analogous to patterns:
  `let x = 1; let r = {.x}  // equivalent to the above`

* Syntax for object patterns changes analogously:
  `func f({.a : Nat = a; .b : Nat}) : Nat = a + b;`

* Syntax for types is not yet adapted, up next.

* Clean up grammar.

* This is breaking change. If you adjust your code but still see errors about type mismatches between types `{.}` and `()`, then you likely have missed changing an empty record literal to `{.}` somewhere.

Pros of this change:

- obsoletes the need for context-dependent disambiguation between records and blocks
- improves clarity of code for human reader
- allows non-recursive record fields and punning
- ...but without legalising dubious equations like `x = x + 1`
- symmetric to variant type syntax

Cons of this change:

- mandatory dot adds notable clutter
- empty record becomes a syntactic special case
- object types may become really cluttered if we change them consistently (currently trying this)

Meta-con:

- breaking change, will affect a lot of code!


nomeata:
Record syntax stuff has been resolved, right? Closing this.

PR #1870, opened 2020-08-24
[Draft] Unambiguous record syntax, take 2

rossberg:
Second attempt at reconciling record object syntax (cf. #1865), by requiring a dot before the left curly. I am not really content with this approach either.

After a day of working with this, the syntax looks weirder and more arbitrary overall than putting a dot before each label. It's especially weird when it appears after another expression, as in `f .{a = 2}` or `shared .{caller} ...`, where it looks like an odd form of selector. Could require parens here, but that could make it more clumsy than before.

Uploading this experiment as well, see the test suite changes for how it affects code. I will experiment with a third option, which is kind of middle grounds between this and #1865, where the dot is placed after the curly ({. }).

List of changes:

* Change record-style object syntax short-hand to be unambiguous with blocks, by requiring a dot before the left curly brace:
  `.{a = 1; b = 2}`

* Syntax for object patterns changes analogously:
  `func f(.{a : Nat = a; b : Nat}) : Nat = a + b`

* Syntax for types is not yet adapted (should it change as well?).

* The old syntax is still supported for now, but generates a deprecation warning. This should help to transition code.

* Cleaned up grammar as far as possible for now, moving all deprecated syntax to a separate section.

* Some more tweaks to grammar and gen-grammar.

* This PR does not change the recursive binding behaviour of record literals, since that would legalise bindings like `x = x + 1`. Maybe this should be addressed differently (e.g., by adding a syntax for accessing identifiers from outer scopes, like `..x` or something). Consequently, punning also isn't supported yet.

Pros of this change:

- obsoletes the need for context-dependent disambiguation between records and blocks (once the deprecated syntax is gone)
- improves clarity of code for human reader

Cons of this change:

- dot adds some clutter
- object types would become cluttered as well if we changed them consistently
- would suggest changing the syntax of variant types as well to make the `#` prefix symmetric

Meta-con:

- breaking change, will affect a lot of code!


nomeata:
Record syntax stuff has been resolved, right? Closing this.

PR #1926, opened 2020-09-14
Add debugging utilities

osa1:
This adds two RTS modules which are built only in debug mode:

- print.rs: Implements `println` and `format` macros that are similar to
  the standard ones, except writes formatted strings to stack-allocated
  arrays.

- debug.rs: Implements various heap dump utilities and an exported
  `printObject` function for printing arbitrary objects. A new function
  `Prim.printObject` added which is no-op in release mode and prints the
  argument's fields in debug mode (with `--sanity-checks`). Some output
  examples:

  Example output when printing a closure:

      0x10064: <Closure size=0x0>

  When printing an object:

      0x16f2c: <Object size=0x1 hash_ptr=0x1006f field=[0x16f3b]>

  When printing a scalar:

      <Scalar 0x2>


osa1:
We've already merged this stuff in another PR, closing.

PR #1993, opened 2020-09-29
Test for PR #1992

crusso:
Adds a test to check gc handles iterators properly, as a PR to osa1/fix_tuple_len_fields (not master!)


ggreif (changes requested):
Checked that the third line in @osa1's patch is triggered by the tests, the first too. But the second (blob-like text) isn't. Let me add a test case too.

ggreif (commented):
Actually the second fix is also exercised, but I had to help it with
``` C
    TUPLE_LEN(c) = 30;
```
so that it actually started crashing.

ggreif (approved):
As such, this can go in. I think it is OK to send both the fix and the tests (at the same time or one after another) now that all three lines of the fix are checked. I'll leave this to both of you.

osa1 (approved):
Great, thanks @crusso !

PR #2070, opened 2020-10-23
Proof of concept for Blob/Text iterator duplication

ggreif:
From Slack:

> @ggreif: I was thinking of backtracking parser combinators (triggered by the Education talk, i.e. IC Fellows). To do this efficiently we need a way to snapshot `Text` (or `Blob` for that matter) iterators. So that we can pick up form a certain iteration state later. We could provide a primitive (or maybe extend the `Iter` interface).
> @crusso: Yeah, we've thought about that before. There current Iterator interface is just too inexpressive (as well as inefficient).
> @ggreif: This is because the `Iter` will mutate itself, and in absence of a clone feature one has to track the char offset and do `Text` slicing, which is O(n).

...
https://github.com/dfinity/motoko-base/pull/112#issuecomment-654776248
...

> @crusso: That would be cool, but go ahead and do it properly, maybe with a short design doc. Let's get it right this time!

> @ggreif: So I think I have something that would work for flat blobs, but needs more work for ropes. It involves looking into closures :open_mouth:. I doubt this can be done generically, but specialisations for `Blob` and `Text` seem possible. Basically one needs to mini-GC  (only) the iterator (deep copy). Maybe the Rust GC already provides this. The caveat is that the original object must not be mutated


luc-blaeser:

PR #2105, opened 2020-11-11
[Draft] Require even more brackets, Rust-style

rossberg:
This PR goes into the direction of a more Rust-style syntax, where argument parens are mandatory and substatements are supposed to be blocks most of the time. Baseline is #2102.

Having gone through the whole exercise, I'm not yet convinced that it is the best way forward:

* The impact on code is substantial. The change touched around 1300 lines in the test suite (on top of the 500 lines for the base PR). I spent many hours applying all these changes semi-manually. It seems problematic to impose such sweeping changes on users without providing a conversion tool, even at this early state.

* If the test suite is a meaningful metric, then the change also adds significant clutter. Most notably, I encountered
  - many inline if's that get notably more noisy with extra braces;
  - plenty of simple case's that suffer similarly;
  - many function calls without parens.

  I assume that the latter is mostly an artefact of our specific FP background, but the other two points will probably affect regular user code in a similar manner. Some of the clutter would be offset later when we allow removing parens around control constructs, but it is unlikely that it can offset all I had to add.

* Mandatory braces introduce the old else-if problem. Avoiding the extra braces around nested if's requires a little grammar hack. Unfortunately, there are more cases like that, for example `label`, which you want to be able to use in front of various control constructs without extra braces. To avoid too many special hacks, I opted for uniformly allowing braces to be omitted around expressions that start with a keyword. That still is pretty ad-hoc, and still leads to a more complicated grammar, when the original plan was simplifying it.

* There are a number of other constructs where the need to make a binary decision between a construct's operand being a block or an expression hurts or is arbitrary. E.g. for `await`, you usually don't want braces, except when you do. Similarly for `debug` and others. For `async`, a block usually makes more sense, but then the asymmetry with `await` is kind of surprising/confusing. As is the general arbitrariness of this choice.

I think I'll need to sleep over this. Opinions?


nomeata:
This doesn’t look like it’s going to be merged any time soon; closing.

PR #2115, opened 2020-11-14
RTS: Allocation-free `?t`

### LINKED ISSUE #2114: The things we do at night
I just had the following dream:

I was at a tech conference, and everyone was kinda glad to have tech conferences again. @rossberg, me and some random guy were chatting, and that guy was explaining something to us that was essentially an iterator like ours, but it looked like it wasn't allocating `?` values like ours, so Andreas hopefully asked how that worked, and the guy said that no, actually it does, because of the `null` vs `?null` confusion, and Andreas said something about how Claudio would have liked that if it didn't. And then i said, well, can't we still make the iterator return either a special `null` sentinel or the plain value, unless it has to return an actual `null` and only box that? Andreas thought about it and said yes, that might work, and that that's kinda why they hired me (nice praise). Then I woke up.

Concretely, the proposal is that the heal representation of `?t` is either
  * A special, fixed value to indicate `null` (must be a pointer value, so that it doesn't clash with small scalars. The GC will not follow it if it's in the static area of the heap.)
  * A pointer to a heap allocated `Some v` box (like now)
  * any other vanilla value `v`, denoting `?v`.

Constructing a `?v` would just return `v` as is, unless it is the null value or a Some box, and only _then_ allocate a new Some box.

That should be safe, avoid any allocations in the usual iterators, right? Additionally, it would even make `[var ?t]` a relatively efficient data structure, right?

(Since now the `Some` box only contains `null` or `Some` it could further be changed to a heap allocated number that just gives the size of the `???null` tower, instead of the unary encoding, maybe as a next step, but that's not essential here.)

Ok, back to sleep, the rest of the dream wasn't bad either.

nomeata:
this implements the ideas in #2114: All `?t` values, with the exception
of `??…?null`, are represented without extra allocation.

This in particular allows iterations to work without allocation; see the
`iter-no-alloc.wasm-run` test case.

I am quite content how localized these changes were.

Fixes #2114.

### DIFF
--- a/rts/motoko-rts/src/gc.rs
+++ b/rts/motoko-rts/src/gc.rs
@@ -324,8 +324,12 @@ unsafe fn scav(
             evac(begin_from_space, begin_to_space, end_to_space, field_addr);
         }
 
-        TAG_BITS64 | TAG_BITS32 | TAG_BLOB | TAG_NULL => {
+        TAG_BITS64 | TAG_BITS32 | TAG_BLOB => {
             // These don't include pointers, skip
+        }
+
+        TAG_NULL => {
+            rts_trap_with("encountered NULL object tag in dynamic object in scav\0".as_ptr());
         }
 
         TAG_FWD_PTR | _ => {


nomeata:
I wrote up a blog post about this scheme at https://www.joachim-breitner.de/blog/787-A_mostly_allocation-free_optional_type

PR #2132, opened 2020-11-19
import actor class from Candid

### LINKED ISSUE #2120: Support actor init args in mo_idl/idl_to_mo.ml
Currently there is an assertion failure in
```
Fatal error: exception "Assert_failure mo_idl/idl_to_mo.ml:105:9"
Raised at file "mo_idl/idl_to_mo.ml", line 105, characters 9-21
```
when compiling .did files with services that take init arguments.

One such example is
https://github.com/dfinity-lab/dfinity/pull/6864/files#diff-47dc8fc25392fec08b3f4d2b12e98c0f6284b751ce0ceed8c450a7055bf23e11R40

For now we have a workaround to exclude those args from the `.did`: https://github.com/dfinity-lab/dfinity/pull/6864/commits/69a813a913299668368e4b35cc3d6073de648795

Please note, that this issue can be split, to
- first be able to send messages to existing services (that came into existence with init args)
- second, instantiating such services.

Clearly, the first part is sufficient to deal with the problem described above, and as such has priority.

chenyan-dfinity:
Fix https://github.com/dfinity/motoko/issues/2120

### DIFF
--- a/src/mo_idl/idl_to_mo.ml
+++ b/src/mo_idl/idl_to_mo.ml
@@ -101,6 +101,11 @@ let check_prog (env: typ I.Env.t) actor : M.typ =
        | M.Con (c, _) -> M.{lab = id; typ = M.Typ c}::fs
        | _ -> assert false) !m_env fs in
      M.Obj (M.Actor, List.sort M.compare_field fs)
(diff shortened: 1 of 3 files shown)


luc-blaeser:

PR #2223, opened 2021-01-05
New GC: in-place compaction

osa1:
This implement the GC algorithm "A fast garbage compaction algorithm" by
Jonkers, also described in The GC Handbook section 3.3.

This algorithm allows fast bump allocation and efficient use of Wasm linear
memory as before, without copying live data to a new space.

Bitmap and mark stack for marking is allocated after the heap pointer. Bitmap
has a constant size (heap_size / word_size_in_bits) so we allocate it first.
Mark stack dynamically grows (initial size is 64 words). Right before marking
the heap looks like

      | static data | heap | bitmap | mark stack |

Because we don't allocate anything during marking it's possible to grow mark
stack by just bumping the heap pointer.

After marking the algorithm makes two passes over the bitmap and moves live
objects to the beginning of the heap. See the references for details.

In this algorithm it's difficult to support interior pointers (even if the
offset is a known constant), so BigInt representation is refactored. Data
pointers now point to Blob headers rather than Blob payloads. When calling a
tommath function we stack allocate a mp_int struct with the data pointer
pointing to a blob payload and pass that to the library function. On return we
update BigInt fields with the values in the stack allocated struct.

Bitmap and mark stack implementations are tested using quickcheck, and can be
used in other GC implementations.

Perf test results:

> In terms of gas, 2 tests regressed and the mean change is +3.1%.
> In terms of size, 2 tests regressed and the mean change is +5.0%.

Some other numbers from the perf tests:

| test    | max heap (words)   | max bitmap (words) | max mark stack (words) | Wasm pages |
| ------- | ------------------ | ------------------ | ---------------------- | ---------- |
| reversi |             5,372  |                168 |           64 (default) |          3 |
| qr      |        13,453,988  |            420,438 |           64 (default) |        850 |

Note that allocation for bookkeeping (mainly bitmap, mark stack never grows
above 64 words in the current tests) is `heap_size / 32` (3.12%). When the
residency is lower than that, the bitmap allocation exceeds the copied data
size in the current collector. In those cases this collector allocates more.

It turns out this case is very common, as residency is extremely low in most
programs (I think mainly because we only do GC between messages). For example,
in "qr", when the heap size is 13,456,988 words, live data size is only 442
words. That's extremely low, 0.003%.

As a result, in the collection described above, current GC allocates 442 words
to copy the live data, but the new collector allocates 420,438 words for the
bitmap. In terms of Wasm pages, the current collector allocates 827 pages while
the new one allocates 850.

In general, whenever residency is smaller than 3.12%, the new collector will
allocate more than the current one.

---

TODOs:

- [ ] Growing the mark stack doesn't work on CI because the implementation requires that the allocations for mark stack will be consecutive heap location, which doesn't work when we use malloc/free-style allocation in the test suite. Not sure how to fix this yet.


crusso (commented):
Looks better to me, but I don't think I should be giving the final approval given my Rust knowledge. I'll defer to @nomeata.

Thanks for making those changes!

osa1:
This will be merged with #2223.

PR #2245, opened 2021-01-11
do async blocks, ir (low-level)

### LINKED ISSUE #1482: Support direct abstraction of code that awaits into functions, without requiring an unnecessary async
### Story

A user should be able to abstract code that awaits into a function, *without* calling that function requiring an extra, expensive message send, i.e. trip through the scheduler and consensus.

### Demo

Consider the current Motoko code:

```
import Mailserver "canister:mailserver";_

actor {
    public func sendAlert() : async Bool {
        let ok1 = 
          if (valid_email("johndoe@mailinator.com")) {
              await Mailserver.send("To: johndoe@mailinator.com\nSubject: A Subject\n\n A Body");
          } else { true };
        let ok2  = if (valid_email("johndoe@mailinator.com")) {
              await Mailserver.send("To: marydoe@mailinator.com\nSubject: B Subject\n\n B Body");
           } else { true };
        return (ok1 and ok2);
    };
```
Clearly, this calls for some refactoring into
```
import Mailserver "canister:mailserver";_

actor {
    public func sendAlert() : async Bool {
        let ok1 : Bool = await sendEmail("johndoe@mailinator.com", "A Subject", "A Body");
        let ok2 : Bool = await sendEmail("marydoe@mailinator.com", "B Subject", "B Body");
        return ok1 and ok2;
    };
    func sendEmail(address : Text, subject: Text, body : Text) : async Bool {
        if (valid_email("johndoe@mailinator.com") { 
          return await Mailserver.send(
                  "To: " # address # "\nSubject: " # subject # "\n\n" # body);
        } else { return true }
    };
};
```
This code abstracts calls to `send` into a function `sendEmail`. Our current compilation scheme causes each call to `sendEmail` to to do one additional message send, increasing latency and the number of context switches.

This is a real example contributed by Dfinity-Alin who rightly asks:
 
> Is this the idiomatic (and most efficient) way to return a promise from a function?
> 
> It looks like the 2 sets of awaits (the one within `sendEmail()` on one hand; and the ones in  `sendAlert()` on the other) result in 2 execution rounds before the canister call is actually made.
> If I instead inline the code of `sendEmail()`, there's only one initial execution round for this canister (followed by an execution round for mailserver; and another execution round for this canister, to handle the response).

A simple demo would show reduced latency of using the old compilation vs new compilation scheme in this example.

The problem is exacerbated by recursion, where each recursive calls to an asynchronous local function induces an additional send.

Consider this recursive evaluator, that calls an asynchronous power function (perhaps provided by a plugin canister in a spreadsheet formula evaluator):

```
import P "mo:prim";
actor a {

  public query func pow(i :Int, n:Int) : async Int {
     power(i, n);
  };

  public type Expression = {
    #const : Int;
    #add : (Expression, Expression);
    #mul : (Expression, Expression);
    #sub : (Expression, Expression);
    #pow : (Expression, Expression);
  };


  func evalAsync(exp : Expression) : async Int {
    switch (exp) {
      case (#const(n)) n;
      case (#add(e1, e2)) (await evalAsync(e1)) + (await evalAsync(e2));
      case (#mul(e1, e2)) (await evalAsync(e1)) * (await evalAsync(e2));
      case (#sub(e1, e2)) (await evalAsync(e1)) - (await evalAsync(e2));
      case (#pow(e1, e2)) await (pow(await (evalAsync e1), await (evalAsync e2)));
    }
  };
};
```
With the current compilation scheme, every call to `evalAsync` in a nested expression causes a context switch, even though only calls to query `pow()` actually require one.

The second demo could use a recursive function such as `evalAsync` to show very dramatic reductions in latency.

### Design

The current proposal is to introduce a new `do async {...}` expression that produces an async value without, on its own, doing a context switch (#2180).

In future, local async functions would desugar to the new `do async {...}` block, instead of context switching `async` blocks.

### Notes

This is just a more realistic example of our inability to abstract await into functions, without doing an unnecessary trip through the scheduler, see #680.

See #1633 for example perf gains in recursive scenarios using a hacked solution.

POC implementation of desing in interpreter, using `do async {}` blocks #2180.

### Progress

As stacked PRs:

- [x] #2180 Implements source typing and AST interpreter. First Draft.
- [x] #2244 Implements high-level ir typing and interpretation. First Draft.
- [x] #2245 Implements lowering to low-level, compilable ir (i.e. await/async transformations). In progress.
- [x] #2277 Implements interpretation of await but not async lowered code, for self-testing

This story is now blocked since Andreas is questioning our semantics of both `await` (longstanding) and `do async` (as proposed here).

crusso:
# Story

Addresses #1482.

# Description

The do-async block, `do async <block>,` returns a future containing the result of `<block>`.
Unlike an `async` expression, `do async <block>` evaluates `<block>` __synchronously__ at first,  switching to asynchronous computation _only_ at the first await of a _pending_ future within `<block>`.

The expression `do async <block>` has type `async T` provided

* `<block>` has type `T`;
* the expression occurs in an asynchronous context (i.e. in the body of a (non-`query`) `shared` function, asynchronous function, `async` expression or another `do async` block).

Do-async blocks can be nested, with the exit of a `return` or `throw` determined by lexical scoping.

It is a compile-time error for `<block>` to contain a `break` to a label declared outside the do-async block.

The expression `do async <block>` evaluates by entering the evaluation of `<block>`, executing synchronously  --  just like a `do <block>` --
until the first await of a pending future, if any.
Unless evaluation traps, it returns a future holding the actual or eventual result of `<block>`.

During the synchronous execution of `<block>`:

* a `return` expression exits the block with a completed future, storing the returned value;
* a `throw` expression resumes execution from the nearest enclosing `try ... catch ...` handler within `<block>` or, if there is none, exits the do-async block
  with a failed future storing the error;
* a trap causes `do async <block>` to trap;
* an `await` expression of
** a successfully completed future continues execution with the contents of that future, or
** a failed, completed future propagates the error contents of the future as if thrown (see above), or
** a pending future exits the do-async block with a new, pending future whose completion with the result of `<block>` is suspended, waiting on the result of the awaited future.

NOTE: Evaluation of a do-async block is synchronous while possible, switching to asynchronous when necessary.
By contrast, evaluation of the similar `async <block-or-exp>` (without `do`) immediately returns a pending future and always computes the result of `<block-or-exp>` _asynchronously_,
never blocking evaluation of the enclosing expression.

NOTE: Do-async blocks are useful for abstracting asynchronous code into re-useable functions, without introducing redundant and expensive `async` expressions just to
satisfy the type checker.

# Guidance

You can review this PR in one go by changing base to master, or by first reviewing the stacked PRs:

#2180 (frontend changes) typing and AST interpreter
#2244 (high-level ir changes) ir and high-level IR interpreter

This PR does the low-level ir changes needed to express async/await lowering and adds tests and examples.

See design/asynccps.md for more details of the lowering.

Once this is in, I suggest we abandon the old sugar for asynchronous local functions and make do-exp the desugaring. I have not done that yet.

- [x] fix premature refund bug (cf. calc.mo)
- [x] cleanup construct ml
- [x] cleanup async.ml/CPSAwait case
- [x] relax effect, ir_effect to be less conservative
- [x] update asynccps.md 
- [x] document
- [x] dogfood in prelude where app to avoid redundant context switches
- [x] more checks in check_ir
- [x] run on replica while smoking a cigarette
- [x] add test for every possible static error in test/fail
- [x]  fix interpretation of CPSDoAsync, CPSAwait (even though we never execute them in our tests)
- [ ] remove silly scheduling of success and reject continuation from both interpreters (these context switches are redundant)
- [ ] typing of `async e` and `do async {}` is identical now - share the code (or better, identify the constructs)


nomeata:
Do you want to keep this PR alive, or is this dead (possibly superseded by https://github.com/dfinity/motoko/issues/2945)?

kritzcreek (commented):
Just cleaning up my review queue

Issue #2251, opened 2021-01-12
Use cargo's build-std instead of xargo


osa1:
Apparently cargo recently gained the ability to build Rust's standard library: https://doc.rust-lang.org/cargo/reference/unstable.html#build-std this command replaces xargo.

I just tried this locally, it's just a matter of updating rustc and cargo:

```diff
diff --git a/nix/default.nix b/nix/default.nix
index 7588024a..4f33d260 100644
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -41,7 +41,7 @@ let
         # Rust nightly
         (self: super: let
           moz_overlay = import self.sources.nixpkgs-mozilla self super;
-          rust-channel = moz_overlay.rustChannelOf { date = "2020-07-22"; channel = "nightly"; };
+          rust-channel = moz_overlay.rustChannelOf { date = "2020-12-29"; channel = "nightly"; };
         in rec {
           rustc-nightly = rust-channel.rust.override {
             targets = [ "wasm32-unknown-unknown" "wasm32-unknown-emscripten" ];
```

and updating the Makefile:

```diff
diff --git a/rts/Makefile b/rts/Makefile
index c23f3fdf..7d980bbd 100644
--- a/rts/Makefile
+++ b/rts/Makefile
@@ -164,7 +164,7 @@ _build/wasm/libmotoko_rts.a: | _build/wasm
        # NB. codegen-units=1 is to make debugging easier, not strictly
        # necessary
        cd motoko-rts && \
-           xargo -v rustc --target=wasm32-unknown-emscripten --release -- \
+           cargo -v rustc -Zbuild-std=core --target=wasm32-unknown-emscripten --release -- \
                -Crelocation-model=pic -Ccodegen-units=1
        cp motoko-rts/target/wasm32-unknown-emscripten/release/libmotoko_rts.a $@

```

However I couldn't make this work in nix-shell and nix-build. There's always some dependency issue, and other errors that I don't understand.

When I try locally outside of nix-shell I can build the release RTS and link it, but the final linker command fails when producing debug RTS:

```
wasm-ld -o mo-rts-debug.wasm \
        --import-memory --shared --no-entry --gc-sections \
        --export=__wasm_call_ctors \
        --export=memcpy \
        --export=memcmp \
        --whole-archive \
        _build/wasm/libmotoko_rts_debug.a _build/wasm/rts_debug.o ...
[...]

PR #2253, opened 2021-01-13
Bump rustc version

osa1:
This PR bumps rustc version which enables cargo's `build-std` feature
(#2251) and also allows us to remove some feature flags that became
stable in the meantime (ptr_offset_from, maybe others too).

(`build-std` feature exists in the current version too, but it's too
buggy or too different than the currently documented version, so I
couldn't use it)


osa1:
We updated rustc in #2761, closing.

PR #2512, opened 2021-04-30
[Bug Repro] mo_idl sometimes export types that are irrelevant and inaccessible.

### LINKED ISSUE #2513: BUG: mo_idl sometimes export types that are irrelevant and inaccessible
Repro and description in PR #2512

crusso:
A minor refactoring of internals.mo to define an actor type definition (for the management canister) reveals that mo_idl is to agressive in output candid types for a program.

The .did output for test/mo_idl/fun_self_arg.mo contains the @ManagementCanister.ic actor type from internal.mo. 

I suspect what is going wrong is mo_idl is using the constructor environment, not type environment, to construct the IDL, including actor types that shouldn't really be accessible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Motoko
+# Motoko 
 
 A simple language for writing Dfinity actors.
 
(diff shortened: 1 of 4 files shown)


luc-blaeser:

Issue #2033, opened 2020-10-08
Collect garbage when the heap is full


osa1:
Currently Motoko programs only do GC in these places:

- In `canister_init`: first function invoked by the system after installing a canister
- In `post_upgrade`: Invoked by the system after upgrading a canister
- After responding to a message (but not if a message is a query? I don't understand this part)

In particular Motoko doesn't do GC if a single message the program fills the heap. Here's an example (run with `run.sh -d`):

```
import Prim "mo:prim";

actor a {

    class range(x : Nat, y : Int) {
        var i = x;
        public func next() : ?Nat { if (i > y) null else {let j = i; i += 1; ?j} };
    };

    public func allocate() : async () {
        // Allocate 4GiB of arrays, 4KiB each. This shouldn't fill up the heap
        // as the arrays are not used and could be collected.
        for (i in range(0, 1024 * 1024 * 4)) {
            let a = Prim.Array_init<Nat>(1024, 0); // 4 KiB
        }
    };

};

a.allocate(); //OR-CALL ingress allocate 0x4449444C0000

//SKIP run
//SKIP run-low
//SKIP run-ir
//SKIP ic-ref-run
```

`allocate` allocates 4GiB of 4KiB-large arrays but doesn't keep any of them alive, so ideally this program should not cause heap overflow.

(Btw, this program fails with "heap out of bounds", which seems weird. I'd expect it to fail with "heap overflow" or something like that.)

To fix this we need to at least do a GC when the heap is full [1], in `alloc_words` [2]. Later we may want to do GC more often, in different places, but we have to do a GC when the heap is full to avoid trapping, so I think it's fine to start with just doing a GC in `alloc_words` when the heap is full for now. Once this is implemented adding more GC points will also be easier.

The main difficulty is to pass roots to the GC at the GC site. As far as I understand we currently don't need this because a property of the current GC sites is that they don't have any roots in locals, stack, or globals, so the only roots are those in the `static_roots` array.

I think we currently don't store pointers in globals (as far as I can see globals are only used for multi-value returns), so we only need to refactor our use of function locals and the Wasm stack. Because Wasm doesn't provide a way to traverse the stack or function locals we can't use "stack maps" or anything like that that tells the GC locations of pointers in Wasm stack, locals, and globals. I think we have no choice other than maintaining our own stack, for roots. The idea is that at a GC point the stack will have all dynamic objects (other than those pointed by the static_roots array) that the GC needs to retain.

[...]

Issue #916, opened 2019-11-22
Optimise for non-pointer variant representation


ggreif:
Currently objects of variant type are always allocated on the heap and referred to by a (skewed) pointer the form `0bPPPPPPPPPPPPPPPPPPPPPPPPPPPPPP11`. When the variant's payload is `()` (unit), which is a common case, then the heap allocated portion (variant key, hash, unit pointer) will convey 0 bits of effective information.

It would be interesting to get rid of heap allocation altogether in this case.

Here is a suggested scheme to do exactly that.

Define a new hash function `H : string -> u32` in terms of the `IDL : string -> u32` one by the following rules:
- if `btst (IDL variant) 1 == 0` then `H variant := IDL variant`
- if `IDL variant == 0xFFFFFFFF` then statically reject the variant by Motoko type checker (just like collisions)
- otherwise `H variant := rotr (IDL variant) n` with `n` being the smallest in `1..31`, such that `btst (rotr (IDL variant) n) 1 == 0`.

Claim: `H variant` can be used by the Motoko runtime to refer to payload-less variants as scalars, avoiding heap allocation altogether. The price is
- excluding `0xFFFFFFFF` from the IDL hash space (justifiable)
- introducing a few more collisions (e.g. `0xFFFF7FFF` and `0xFFFFFF7F` suddenly collide)
- slightly increased complexity of `H` (compile-time only)
- cannot reuse the Motoko hash as IDL hash in external messaging.

That said, we could also simply mask the hash as `H variant := bitand 0xFFFFFFFD (IDL variant)` and be even better :-) But this will cause _ugly_ collisions, like `H "b" == H "c"`.

PR #2522, opened 2021-05-06
New GC: in-place compaction

osa1:
This merges #2223 and master and enables the new collector with a new moc flag
`--compacting-gc`. The goal is to (1) avoid bitrot in #2223 (2) allow easy
testing and benchmarking of different collectors.

We include both collectors in the Wasms. Binary sizes grow between 0.8% (CanCan
backend) to 1.8% (in simple tests in `run-drun`).

Some benchmark results here:
https://github.com/dfinity/motoko/issues/2033#issuecomment-842379468

An improvement to the compacting GC is currently being implemented in branch
`osa1/compacting_gc_3`.


osa1:
> I don't remember checking the code size if we ship both collectors. I will check that and update.

CanCan backend size grows from 372,584 bytes to 375,780 bytes. That's 3,196 bytes, 0.85%. So at least for this particular app it should be fine to ship with two GCs. The overhead is constant so it will have more impact on smaller applications. I checked a few simple tests from the run-drun directory and for example it adds 1.8% to `life-mut`, which is 124 lines (including whitespace).

I tried `wasm-opt --vacuum` (using wasm-opt version 95) on binaries with just the copying GC and both GCs, but it didn't make any difference.

osa1:
No idea what's wrong with the CI but I don't think it's related to this PR.

osa1:
Commit history is completely confused after changing the base branch to master. Should I squash the commits maybe?

osa1:
> Commit history is completely confused after changing the base branch to master. Should I squash the commits maybe?

The problem seems to be that one of the old base branches of this PR is merged to master, but as a single squashed commit. So this branch now shows the non-squashed commits from that branch.

nomeata:
> > Commit history is completely confused after changing the base branch to master. Should I squash the commits maybe?
> 
> The problem seems to be that one of the old base branches of this PR is merged to master, but as a single squashed commit. So this branch now shows the non-squashed commits from that branch.

Which PR was that that was merged into master? #2223 does not seem to be merged.

osa1:
> Which PR was that that was merged into master?

Commit history showing commits from https://github.com/dfinity/motoko/pull/2210 so I guess #2223 was based on #2210 at some point, or somehow included the commits before squash-merging #2210.

nomeata:
Yes, that’s expected, even with `git post-squash`. Don’t worry whether commit history of a PR is clean.

nomeata (changes requested):
I’d like to have a look before it enters master, if you don’t mind, but didn’t get through it yet, so blocking this.

rts/motoko-rts/src/bitmap.rs
```diff
@@ -0,0 +1,116 @@
...
+
+pub unsafe fn free_bitmap() {
+    BITMAP_PTR = core::ptr::null_mut();
+}
+
+pub unsafe fn get_bit(idx: u32) -> bool {
+    let byte_idx = idx / 8;
+    let byte = *BITMAP_PTR.add(byte_idx as usize);
+    let bit_idx = idx % 8;
+    (byte >> bit_idx) & 0b1 == 0b1
```
crusso, line 31: Would it be worth checking this generates wasm bittest instruction? Might be faster, but who knows. I think there a bittest intrinsic in Rust now. Or just make an issue for later.

default.nix
```diff
@@ -370,6 +375,7 @@ rec {
   in fix_names ({
       run        = test_subdir "run"        [ moc ] ;
       run-dbg    = snty_subdir "run"        [ moc ] ;
+      run-compacting-gc = compacting_gc_subdir "run" [ moc ] ;
```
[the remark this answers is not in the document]
osa1: Ah, I didn't realize we have different jobs for run-drun and other tests. I'll update.

rts/motoko-rts/src/bitmap.rs
osa1: Which instruction do you mean? I don't see a bittest instruction here: https://webassembly.github.io/spec/core/appendix/index-instructions.html

default.nix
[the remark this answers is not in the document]
osa1: I updated the code to run drun and ic-ref tests with `--compacting-gc`.

rts/motoko-rts/src/bitmap.rs
crusso: If there even are dedicated bittest instructions - I couldn't find any. Maybe I was thinking of x86

rts/motoko-rts/src/bitmap.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+pub unsafe fn set_bit(idx: u32) {
+    let byte_idx = idx / 8;
+    let byte = *BITMAP_PTR.add(byte_idx as usize);
+    let bit_idx = idx % 8;
+    let new_byte = byte | (0b1 << bit_idx);
```
crusso, line 38: ditto

rts/motoko-rts/src/bitmap.rs
crusso: Just confirmed with Andreas that there aren't any. Don't bother.

rts/motoko-rts/src/bitmap.rs
```diff
@@ -0,0 +1,116 @@
...
+                return None;
+            }
+
+            // Inner loop iterates bits in the current word
+            while self.current_word != 0 {
+                if self.current_word & 0b1 == 0b1 {
+                    let bit_idx = (self.current_word_idx * 64) + (64 - self.bits_left);
+                    self.current_word >>= 1;
+                    self.bits_left -= 1;
+                    return Some(bit_idx);
```
crusso, line 98: Is this Some allocation cheap in Rust or would it be worth avoiding iterators altogether for this low-level code.

osa1: `Option<u32>` values are represented as two `u32`s on Wasm, one as tag (indicating whether the value is `None` or `Some(...)`) and one for the value of `Some`. Here's an example function that returns a `Option<u32>`: https://godbolt.org/z/eYT6oMhhf it takes two arguments, one for the address of the resulting `Option<u32>`, one for the argument in the source code (the `u32`). It writes `1` to the first memory location (so tag = `Some`), and the argument to the second memory location.

In our case, heap is 4GiB and each word (32-bit) gets a bit, so we need `1024 * 1024 * 1024` bits in the worst case (actually less than that, as we can't allocate a bitmap if the heap is full). So we could use one larger than that to indicate the `None` case here.

I will implement this and update the benchmark results.

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -0,0 +1,312 @@
...
+
+    mark_stack(heap_base);
+
+    thread_roots(static_roots, heap_base);
+    thread(closure_table_loc, heap_base);
+    update_fwd_refs(heap_base);
+    update_bwd_refs(heap_base);
+
+    free_bitmap();
+    free_mark_stack();
```
nomeata, line 50: May not matter, but it’s probably nice to free in reverse order of allocation?
```suggestion
    free_mark_stack();
    free_bitmap();
```

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -0,0 +1,312 @@
...
+
+unsafe fn thread_obj_fields(obj: *mut Obj, heap_base: u32) {
+    match obj.tag() {
+        TAG_OBJECT => {
+            let obj = obj as *mut Object;
+            let obj_payload = obj.payload_addr();
+            for i in 0..obj.size() {
+                thread(obj_payload.add(i as usize), heap_base);
+            }
+        }
```
nomeata, line 216: There is some repetition here and in other places.

How about a higher-order function (sorry for the non-Rusty syntax)
```
with_ptrs_in_obj : (*mut Obj) -> (*mut SkewedPtr -> ()) -> ()
```
that you can use is in `thread_obj_fields` and `mark_fields` and in the copying GC? (Assuming Rust will inline it, of course)? I think this would greatly improve readability, and also avoid bugs as we change the object layout or add new object types.

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -0,0 +1,312 @@
...
+    (*pointed).tag = field as u32;
+}
+
+/// Unthread all references, replacing with `new_loc`
+unsafe fn unthread(obj: *mut Obj, new_loc: u32) {
+    // NOTE: For this to work heap addresses need to be greater than the largest value for object
+    // headers. Currently this holds. TODO: Document this better.
+    let mut header = (*obj).tag;
+    while header > TAG_NULL {
+        // TODO: is `header > TAG_NULL` the best way to distinguish a tag from a pointer?
```
nomeata, line 304: This is a bit hairy. Maybe leave a note in the cod where we define header tags that these tags must be smaller than any possible dynamic pointer?

Also, this would limit our wiggle room if we want to put variant tags or array lengths etc into the header… (#916)

nomeata (commented):
Very nice! Only minor comments

rts/motoko-rts/src/bitmap.rs
osa1: > In our case, heap is 4GiB and each word (32-bit) gets a bit, so we need 1024 * 1024 * 1024 bits in the worst case (actually less than that, as we can't allocate a bitmap if the heap is full). So we could use one larger than that to indicate the None case here.

This made 2.7% reduction in instructions in some cases so I will commit my code that implements this now.

nomeata: Nice catch!

osa1: Done.

rts/motoko-rts/src/gc/copying.rs
```diff
@@ -0,0 +1,242 @@
+use crate::alloc;
```
crusso, line 1: I assume the code in here has just moved, and isn't substantially changed right?

osa1: Yes, no need to review this file.

rts/motoko-rts/src/mark_stack.rs
```diff
@@ -0,0 +1,61 @@
...
+pub unsafe fn push_mark_stack(obj: usize) {
+    if STACK_LEN == (*STACK_PTR).len.to_words() {
+        grow_stack();
+    }
+
+    *(STACK_PTR.payload_addr() as *mut usize).add(STACK_LEN.0 as usize) = obj;
+    STACK_LEN += Words(1);
+}
+
+pub unsafe fn pop_mark_stack() -> Option<usize> {
```
osa1, line 53: I should try to get rid of this `Option` as well, I think it could also improve perf. (https://github.com/dfinity/motoko/pull/2522/files#r649170651)

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -0,0 +1,312 @@
...
+    free_mark_stack();
+}
+
+unsafe fn mark_static_roots(static_roots: SkewedPtr, heap_base: u32) {
+    let root_array = static_roots.as_array();
+
+    // Static objects are not in the dynamic heap so don't need marking.
+    for i in 0..root_array.len() {
+        let obj = root_array.get(i).unskew() as *mut Obj;
+        mark_fields(obj, heap_base);
```
crusso, line 59: Why do we mark their fields? Because they may just be ref cells whose contents has been updated to point into the non-static heap?

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -0,0 +1,312 @@
...
+    }
+}
+
+unsafe fn push_mark_stack(obj: SkewedPtr, heap_base: u32) {
+    if obj.is_tagged_scalar() {
```
crusso, line 64: would it make sense to do this at the call-sites to avoid the function call altogether?

crusso: (not a requirement for now)

rts/motoko-rts/src/gc/mark_compact.rs
nomeata: Yes. Else all stuff you put in your precious top-level `var state = …` will be lost :)

osa1: So I think this part is a bit confusing. Currently static heap is just an array with `MutBox` elements. The `MutBox`es also live in the static heap so none of the array of the `MutBox`es need marking (or evacuation). However `MutBox`es point to dynamic heap so fields of `MutBox`es need to be marked. That's what we do in this code.

There's nothing else in the static heap currently, only the array and `MutBox`es.

You will probably ask "why not just store the MutBox fields in the array"? That's because the `MutBox`es can be referenced from other objects. We can't have interior pointers so if we remove the `MutBox`es and store fields directly in the array then we won't be able to refer to those fields directly, we will need to refer to the array, and then store the index of the reference for the field.

nomeata: > There's nothing else in the static heap currently, only the array and MutBoxes.

I don’t think that’s true; shouldn’t there be lots of constant values (texts etc.) in the static heap?

(The “static roots” are all MutBoxes assembled collected in one array, if that’s what you mean.)

rts/motoko-rts/src/bitmap.rs
crusso: Cool!

rts/motoko-rts/src/mark_stack.rs
crusso: Cool again

rts/motoko-rts/src/gc/mark_compact.rs
crusso: I think the fact that you can distinguish headers from pointers is unfortunately fundamental to the threading algorithm, unless there's some other technique that can be used.

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -0,0 +1,312 @@
...
+    closure_table_loc: *mut SkewedPtr,
+) {
+    let heap_size = Bytes(heap_end - heap_base);
+
+    alloc_bitmap(heap_size);
+    alloc_mark_stack();
+
+    mark_static_roots(static_roots, heap_base);
+
+    // TODO: We could skip the is_tagged_scalar, heap_base etc. checks
```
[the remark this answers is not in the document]
osa1: Hm, so the value of `closure_table_loc` (pointer to the static variable that holds pointer to the table) cannot be NULL as it's defined in static memory: https://github.com/dfinity/motoko/blob/f23cedd26cea622d9e650fa043884f6cf21767c6/rts/motoko-rts/src/closure_table.rs#L30

The the value of the variable is `0` initially, until we call `remember_closure` for the first time. So I think this comment is invalid and should be removed.

rts/motoko-rts/src/gc/mark_compact.rs
osa1: I can try it, but this function is called in 11 places so the code will suffer a little bit.

crusso:
I guess one thing we could consider is using stable memory for the mark stack and bits, but maybe that's too slow.

rts/motoko-rts/src/gc/mark_compact.rs
osa1: I wonder if LLVM can do "partial inlining" and only inline the early return path, effectively doing the transformation you suggest.

rts/motoko-rts/src/gc/mark_compact.rs
osa1: > I don’t think that’s true; shouldn’t there be lots of constant values (texts etc.) in the static heap?

Err, sorry, I should've said "as far as garbage collection concerned".

> And this doesn’t affect GC

What doesn't affect GC? The array is used by the GC as the root set, right?

rts/motoko-rts/src/gc/mark_compact.rs
nomeata: Right. It’ll take away roughly ¼ of the possible header values. Still plenty of room for other things.

rts/motoko-rts/src/gc/mark_compact.rs
nomeata: I mean, it doesn’t affect GC that there is other stuff in the static heap. I am just confirming your “as far as GC is concerned” :-)

rts/motoko-rts/src/gc/mark_compact.rs
crusso: To be honest, I have no expectation for what LLVM will do for you. Would a macro not let you keep the code more or less unchanged?

rts/motoko-rts/src/mark_stack.rs
osa1: I don't understand why but if I apply the same technique here the benchmark starts using more instructions. No idea why. The new `pop_mark_stack` is actually smaller in generated Wasm. This is the new version I've tried:

```rust
pub unsafe fn pop_mark_stack() -> usize {
    if STACK_LEN.0 == 0 {
        0
    } else {
        STACK_LEN -= Words(1);
        *(STACK_PTR.payload_addr() as *mut usize).add(STACK_LEN.0 as usize)
    }
}
```

rts/motoko-rts/src/gc/mark_compact.rs
osa1: IIRC I had tried this before but it did not work well because of some difficulty with handing `BigInt`s. Now that we have a simpler `BigInt` representation I think it should work out nicely. I will update.

osa1: I implemented this in 9af5279 . It improved performance (instructions used) slightly in copying GC (< 1%) but made compacting GC 3% worse. The code is much better though..

rts/motoko-rts/src/gc/copying.rs
osa1: This file is now changed with 9af5279.

rts/motoko-rts/src/gc/mark_compact.rs
osa1: Done as a part of 9af5279.

rts/motoko-rts/src/mark_stack.rs
osa1: Here's the patch for this change:

```diff
diff --git a/rts/motoko-rts-tests/src/mark_stack.rs b/rts/motoko-rts-tests/src/mark_stack.rs
index 0499a31bf..4938f252a 100644
--- a/rts/motoko-rts-tests/src/mark_stack.rs
+++ b/rts/motoko-rts-tests/src/mark_stack.rs
@@ -24,7 +24,7 @@ fn test_(objs: Vec<usize>) -> TestResult {
 
         for obj in objs.iter().rev() {
             let popped = pop_mark_stack();
-            if popped != Some(*obj) {
+            if popped != *obj {
                 free_mark_stack(); // TODO: Does not really free
                 return TestResult::error(format!(
                     "Unexpected object popped, expected={:?}, popped={:?}",
diff --git a/rts/motoko-rts/src/gc/mark_compact.rs b/rts/motoko-rts/src/gc/mark_compact.rs
index ef8a814ee..5890d67db 100644
--- a/rts/motoko-rts/src/gc/mark_compact.rs
+++ b/rts/motoko-rts/src/gc/mark_compact.rs
@@ -80,7 +80,11 @@ unsafe fn push_mark_stack(obj: SkewedPtr, heap_base: u32) {
 }
 
 unsafe fn mark_stack(heap_base: u32) {
-    while let Some(obj) = pop_mark_stack() {
+    loop {
+        let obj = pop_mark_stack();
+        if obj == 0 {
+            break;
+        }
         mark_fields(obj as *mut Obj, heap_base);
     }
 }
diff --git a/rts/motoko-rts/src/mark_stack.rs b/rts/motoko-rts/src/mark_stack.rs
index be87acc6f..1bf8145b9 100644
--- a/rts/motoko-rts/src/mark_stack.rs
+++ b/rts/motoko-rts/src/mark_stack.rs
@@ -50,12 +50,11 @@ pub unsafe fn push_mark_stack(obj: usize) {
     STACK_LEN += Words(1);
 }
 
-pub unsafe fn pop_mark_stack() -> Option<usize> {
+pub unsafe fn pop_mark_stack() -> usize {
     if STACK_LEN.0 == 0 {
-        None
+        0
     } else {
         STACK_LEN -= Words(1);
-        let p = *(STACK_PTR.payload_addr() as *mut usize).add(STACK_LEN.0 as usize);
-        Some(p)
+        *(STACK_PTR.payload_addr() as *mut usize).add(STACK_LEN.0 as usize)
     }
 }
```

crusso: Any idea why the cost went up?

osa1: Here's the diff of before vs. after this patch (mo-rts.wasm):

```diff
--- mo-rts.wat  2021-06-14 15:44:34.892004845 +0300
+++ ../../motoko/rts/mo-rts.wat 2021-06-14 15:44:34.180013747 +0300
@@ -19343,7 +19343,7 @@
     local.get 5
     i32.const 1
     i32.add
-    local.set 10
+    local.set 8
     block  ;; label = @1
       local.get 5
       i32.const 5
@@ -19352,7 +19352,7 @@
       local.tee 2
       i32.eqz
       br_if 0 (;@1;)
-      local.get 10
+      local.get 8
       i32.const 8
       i32.add
       local.set 1
@@ -19402,7 +19402,6 @@
         local.tee 2
         i32.const 15796
         i32.add
-        local.tee 8
         local.get 1
         i32.const -1
         i32.add
@@ -19419,21 +19418,27 @@
         i32.const 8
         i32.add
         i32.load
+        local.tee 1
+        i32.eqz
+        br_if 1 (;@1;)
+        local.get 1
         local.get 4
         call $motoko_rts::gc::mark_compact::mark_fields::hc1bca295efbdeb41
-        local.get 8
+        global.get 1
+        i32.const 15796
+        i32.add
         i32.load
         local.tee 1
         br_if 0 (;@2;)
       end
     end
     block  ;; label = @1
-      local.get 10
+      local.get 8
       i32.load offset=4 align=1
       local.tee 2
       i32.eqz
       br_if 0 (;@1;)
-      local.get 10
+      local.get 8
       i32.const 8
       i32.add
       local.set 1
```

So the new version, after a lot of inlining, ends up being larger. No idea why though. This code is for `compacting_gc`, so `mark_stack` is inlined in `mark_compact`, which is inlined in `compacting_gc`. Somehow the version with this change has more stack traffic.

crusso:
LGTM.

rts/motoko-rts/src/bitmap.rs
```diff
@@ -0,0 +1,125 @@
...
+//
+// Heap is 4GiB and each 32-bit word gets a bit, so this is one larger than the bit for the last
+// word in heap.
+//
+// (We actually need less bits than that as when the heap is full we can't allocate bitmap and mark
+// stack and can't do GC)
+pub const BITMAP_ITER_END: u32 = 1024 * 1024 * 1024;
+
+impl BitmapIter {
+    /// Returns the next bit, or `MAX_BIT` if there are no more bits set.
```
crusso, line 91:
```suggestion
    /// Returns the next bit, or `BITMAP_ITER_END` if there are no more bits set.
```

rts/motoko-rts/src/gc/copying.rs
```diff
@@ -0,0 +1,158 @@
...
+
+/// Evacuate (copy) an object in from-space to to-space, update end_to_space. If the object was
+/// already evacuated end_to_space is not changed.
+///
+/// Arguments:
+///
+/// - begin_from_space: Where the dynamic heap starts. Used for two things:
+///
+///   - An object is static if its address is below this value. These objects either don't point to
+///     dynamic heap, or listed in static_roots array. Objects in static_roots are scavenged
```
crusso, line 16:
```suggestion
///     dynamic heap, or are listed in static_roots array. Objects in static_roots are scavenged
```

rts/motoko-rts/src/gc/copying.rs
```diff
@@ -0,0 +1,158 @@
...
+// "static roots", we just scavenge them.
+unsafe fn evac_static_roots(
+    begin_from_space: usize,
+    begin_to_space: usize,
+    end_to_space: &mut usize,
+    roots: *mut Array,
+) {
+    // The array and the objects pointed by the array are all static so we don't evacuate them. We
+    // only evacuate fields of objects in the array.
+    for i in 0..roots.len() {
```
crusso, line 99: Do we actually know that all the values in roots are pointers to heap objects and never unboxed value or even static objects? Otherwise this code might crash, right?

crusso: My reasoning is that env_static_roots eventually calls visit_pointer_fields, but there's not check that the obj is a proper object.

rts/motoko-rts/src/mark_stack.rs
crusso: Maybe it would be shorter if you just stored the sentinel value 0 at the bottom of the stack and avoided doing the conditional testing on pop?

rts/motoko-rts/src/gc/copying.rs
osa1: Yeah, so the static heap (other than static data/constants) have a an array of `MutBox`es (`roots` in this function). These `MutBox`es also live in static heap but they can point to dynamic heap. This code evacuates those `MutBox` fields.

rts/motoko-rts/src/mark_stack.rs
crusso: And maybe just maintain pointer to top of stack rather than computing from length on push and pop

rts/motoko-rts/src/gc/copying.rs
crusso: Right, just remembered that after (re)seeing the explanation to a previous comment. Sorry about the gaslighting

rts/motoko-rts/src/mark_stack.rs
osa1: > And maybe just maintain pointer to top of stack rather than computing from length on push and pop

This made a benchmark use 2.3% less instructions. Pushed as 836992cfc.

osa1: > Maybe it would be shorter if you just stored the sentinel value 0 at the bottom of the stack and avoided doing the conditional testing on pop?

We would need a conditional to avoid popping the sentinel value, right? Otherwise a push after popping the sentinel value would break the stack. So I think we can't remove the conditional in `pop_mark_stack`.

PR #2545, opened 2021-05-31
Start decoding extended name sections

### LINKED ISSUE #2543: mo-ld: Implement extended names section
Our `mo-ld` linker handles the official parts of the `names` section (functions and locals), but not the extensions in       https://github.com/WebAssembly/extended-name-section/blob/master/proposals/extended-name-section/Overview.md, and LLVM starts to produce them. In #2532 we start to ignore these subsections to not fail, but we should probably implement them.

osa1:
I can't test this yet as upgrading to LLVM 12 (#2542) fails with other (I think unrelated) problems.

Fixes #2543

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -8142,13 +8142,10 @@ and conclude_module env start_fi_o =
     let open Wasm_exts.CustomModule in
     { module_;
       dylink = None;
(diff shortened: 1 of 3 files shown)


osa1:
This is merged to #2542 via another PR, closing.

PR #2554, opened 2021-06-03
Change the version label from master to latest

lsgunnlsgunn:
To be more PC and "accurate" changing the label for the docs bucket from "master" to "latest"


nomeata (approved):

lsgunnlsgunn:
Looks like this isn't going to be possible with how the site-entry page is being circumvented.

PR #2570, opened 2021-06-10
CI: Automatically create tags for releases

### LINKED ISSUE #2566: CI: Create Github releases from CI
We should configure our Github Actions that that 
 * when a new tag is created
 * and the tag looks like a release
 * and the release artifacts built (probably in separate steps for linux and darwin)
 * we automatically create a new Github release and include the files with that release, in a format and naming convention that works for `vessel`
 * bonus points for extracting the relevant section from the `Changelog.md` and using is in the release.

This should be documented in <./CI.md>.

@kritzcreek mentioned that he has done this before. Can you point to prior art that we can cargo-cult?

nomeata:
Whenever we push a tag, this will:

 * Fetch these assets from the cache
   ```
   motoko-linux64-$VERSION.tar.gz
   motoko-macos-$VERSION.tar.gz
   moc-interpreter-$VERSION.js
   moc-$VERSION.js
   ```

 * Extracts the changes from the changelog, failing hard if it doesn’t _precisely_
   look as expected. This also means that nothing happens if you use some tag
   like `some-revision-I-wanted-to-remember`

 * Creates a Github Release, with the body of the changelog entry as the body,
   and the above files as assets. This means one can download motoko from
   ```
   https://github.com/dfinity/motoko/releases/download/0.1.2/motoko-linux64-0.1.2.tar.gz
   ```

Note that the `nix-build` step needs stuff from both linux and darwin. This
will fail if the nix cache was not warm. This could happen if you push the tag
while there were still jobs runnig. In that case, just restart the github
action manually.

With this we can stop deploying via DFINITY’s internal infrastructure.

This also removes the hydra-related infrastructure for releases, and updates `CI.md`.

Fixes #2566


.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+      id: get_version
+      run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
+
+    - run: nix-build --max-jobs 1 release-files.nix --argstr releaseVersion '${{ steps.get_version.outputs.VERSION }}'
+
+    - name: Extract changelog
+      id: read_release
+      run: |
+        export VERSION='${{ steps.get_version.outputs.VERSION }}'
+        perl -0777 -ne '/^# Motoko compiler changelog\n\n== (??{quotemeta($ENV{VERSION})}) \(\d\d\d\d-\d\d-\d\d\)\n\n(.*?)^==/sm or die "Changelog does not look right for this version\n" ; print $1' Changelog.md > changelog-extract.md
```
kritzcreek, line 35: You just had to

nomeata:
See https://github.com/nomeata/motoko/releases for an example (I practiced on my fork)

kritzcreek (approved):
Very neat :)

.github/workflows/release.yml
nomeata: Is there an alternative :-D

PR #2573, opened 2021-06-10
GC unit test infra, start adding GC tests

osa1:
This refactors the RTS to allow testing garbage collectors and adds a simple GC
test (more tests to be implemented in a new PR).

Major changes:

- New trait `Memory` is introduced, with a `alloc_words` method for heap
  allocation.

- All allocating functions now take a generic `Memory` argument. Part of this
  is necessary for testing the collectors, but this also enables more testing:
  we now test more code paths in mark stack implementation.

- Copying and mark-compact GC entry points now take these arguments:
  - `mem: &mut M`: memory implementation
  - `get_hp`: callback to get heap pointer
  - `set_hp: SetHp`: callback to set heap pointer
  - `mem_base: u32`: beginning of dynamic heap
  - `mem_end: u32`: end of dynamic heap
  - `static_roots: SkewedPtr`: pointer to the static root array
  - `closure_table_loc: *mut SkewedPtr`: address of the closure table in
    dynamic heap

- Monomorphised versions of copying and mark-compact GC entry points added to
  be called from compiled code

- Copying GC refactored to avoid using `grow_memory`, it now uses `alloc_words`
  to allocate in to-space.

- New crate `motoko-rts-macros` added. This defines a macro `ic_mem_fn` which
  is used to generate monomorphised versions of allocationg functions, to be
  called from compiled code.

- Randomized testing in RTS tests now use proptest instead of quickcheck.
  proptest is much more flexible than quickcheck, allows easier testing in some
  cases. 

Perf changes:

Using CanCan backend benchmark as described here:
https://github.com/dfinity/motoko/issues/2033#issuecomment-842379468

- With copying GC the the benchmark uses 1.5% more instructions. I suspect this
  is caused by the change in copying GC to use `alloc_words` instead of
  `grow_memory`.

- With compacting GC the benchmark uses 0.12% less instructions

TODOs:

- I had to disable the test `big-array-access`. It's a failing test, and the
  backtrace changes depending on RTS build flags (debug or release).


osa1:
The test `big-array-access` now fails with a different call stack in different builds (release and debug). I will either have to update the test somehow so that it won't fail, or update the test runner so that it can ignore the call stacks, or find another way ..

rts/motoko-rts-macros/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "motoko-rts-macros"
+version = "0.1.0"
+authors = ["Ömer Sinan Ağacan <omeragacan@gmail.com>"]
```
crusso, line 4: I wonder if we should be avoiding emails since they asked us to purge them. I don't actually care ;->

rts/motoko-rts-macros/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+    let fn_wrapper_ident = syn::Ident::new(&format!("ic_{}", fn_ident), fn_ident.span());
+    let fn_name = fn_ident.to_string();
+    let wrapper_ret = fun_sig.output.clone();
+    let wrapper_args: Vec<(syn::Ident, syn::Type)> = fun_sig
+        .inputs
+        .iter()
+        .enumerate()
+        .map(|(i, arg)| match arg {
+            syn::FnArg::Receiver(_) => {
+                panic!("IC functions can't have recievers (`&self`, `&mut self`, etc.)")
```
crusso, line 34:
```suggestion
                panic!("IC functions can't have receivers (`&self`, `&mut self`, etc.)")
```

rts/motoko-rts-macros/Cargo.toml
osa1: Good point, this should be `dfinity <sdk@dfinity.org>`. I'll update.

rts/motoko-rts/src/memory.rs
```diff
@@ -0,0 +1,35 @@
+#[cfg(feature = "ic")]
+pub mod ic;
+
+use crate::rts_trap_with;
+use crate::types::*;
+
+pub trait Memory {
+    unsafe fn alloc_words(&mut self, n: Words<u32>) -> SkewedPtr;
+
+    unsafe fn get_hp(&self) -> usize;
```
osa1, line 10: We discussed this trait with Ulan and realized that this method is only used by the collectors and so should not be a part of this trait. I will remove this.

rts/motoko-rts-tests/src/gc.rs
```diff
@@ -0,0 +1,211 @@
...
+            heap_ptr_offset,
+        );
+    }
+}
+
+/// Check the dynamic heap:
+///
+/// - Objects should point to objects with the expected tags (as specified by the `objects` argument)
+/// - Each tag should be seen at most once
+/// - All of `roots` should be seen
```
osa1, line 92: @ulan points out that we should check that we only see the transitive closure of roots here.

osa1:
I just run cancan backend benchmark and it reports 2.5% increase in cycles. No idea where do those extra cycles come from yet.

osa1:
Interestingly, mo-rts.wasm file is only 74 bytes larger, that's 0.04% (159,640 to 159,714).

rts/motoko-rts-tests/src/gc.rs
```diff
@@ -0,0 +1,286 @@
...
+//
+// To convert an offset into an address, add heap array's address to the offset.
+
+mod heap;
+mod utils;
+
+use heap::MotokoHeap;
+use utils::{read_word, ObjectIdx, GC, GC_IMPLS, WORD_SIZE};
+
+// use motoko_rts::debug;
```
ulan, line 14: Nit: commented out code

rts/motoko-rts-tests/src/gc.rs
```diff
@@ -0,0 +1,286 @@
...
+fn test_gcs(heap_descr: &TestHeap) {
+    for gc in &GC_IMPLS {
+        test_gc(*gc, &heap_descr.heap, &heap_descr.roots);
+    }
+}
+
+fn test_gc(gc: GC, refs: &HashMap<u32, Vec<u32>>, roots: &[u32]) {
+    let heap = MotokoHeap::new(refs, roots, gc);
+
+    // println!("{:?}", heap.heap);
```
ulan, line 54: ditto

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -0,0 +1,85 @@
...
+
+        // Grow memory if needed
+        grow_memory(new_hp as usize);
+
+        skew(old_hp as usize)
+    }
+}
+
+unsafe fn grow_memory(ptr: usize) {
+    let total_pages_needed = ((ptr / 65536) + 1) as i32;
```
ulan, line 77: Nit: Is there a named constant that can be used instead of 65536?

rts/motoko-rts/src/memory.rs
```diff
@@ -0,0 +1,57 @@
...
+    (*blob).header.tag = TAG_BLOB;
+    (*blob).len = size;
+    ptr
+}
+
+/// Helper for allocating arrays
+#[ic_mem_fn]
+pub unsafe fn alloc_array<M: Memory>(mem: &mut M, len: u32) -> SkewedPtr {
+    // Array payload should not be larger than half of the memory
+    if len > 1 << (32 - 2 - 1) {
```
ulan, line 45: Nit: Is there a named constant that can be used instead of 1 << 32 ?

rts/motoko-rts/src/gc/mark_compact/mark_stack.rs
```diff
@@ -39,9 +39,9 @@ pub unsafe fn free_mark_stack() {
 }
 
 /// Doubles the stack size
-unsafe fn grow_stack() {
+unsafe fn grow_stack<M: Memory>(mem: &mut M) {
     let stack_cap: Words<u32> = STACK_BLOB_PTR.len().to_words();
-    let p = alloc_words(stack_cap).unskew() as *mut usize;
+    let p = mem.alloc_words(stack_cap).unskew() as *mut usize;
```
ulan, line 44: A question about the existing code:
I see that `STACK_TOP` is expected to be equal to `p` in line 47, but how do we ensure that equality. Shouldn't we use `p` to set `STACK_TOP` or `STACK_PTR`?

ulan (approved):
I did a very light review and skipped macro magic. The change lgtm on the high-level without digging deep.

osa1:
After carefully adding two inline attributes, CanCan benchmark now uses 1.5% more instructions with copying GC, but 0.12% less in compacting GC. I will push the commit with inline attributes now.

rts/motoko-rts/src/gc/mark_compact/mark_stack.rs
osa1: Here's how the Wasm heap looks like before starting compacting GC:

```
| static heap | dynamic heap |
^             ^              ^
0          heap base         hp
```

In compacting GC we first allocate a bitmap (size depends on dynamic heap size, does not grow) and a mark stack (size unknown, grows dynamically). At that point the heap looks like this:

```
| static heap | dynamic heap | bitmap | mark stack |
^             ^              ^        ^            ^
0          heap base         hp    stack base,  stack top
                                   stack ptr
```

As we push stuff to the mark stack stack ptr gets incremented towards stack top:

```
... | mark stack |
    ^    ^       ^
    |    p       |
stack base    stack top 
```

We grow the stack when `p == stack_top`, but we don't support moving the stack and don't have a linked-array representation, so the assumption here is that if we allocate more to grow the mark stack, new memory should be `stack top`. This effectively grows the mark stack without copying (moving) it. The assertion is checking that we didn't allocate anything after allocating the mark stack so `alloc_words` returned `stack top`.

What make sure that new allocation will be `stack top` is the rest of the GC code which does not allocate during marking. If we do any allocation during marking the assertion will fail.

@ulan Does this make sense?

rts/motoko-rts/src/memory/ic.rs
osa1: This is Wasm page size (64KiB), I'll add a constant for this.

I'm a bit confused about what this code is doing though. I think it should divide by page size, round up. But the code is not doing that. It should be

```rust
(ptr + WASM_PAGE_SIZE - 1) / WASM_PAGE_SIZE
```

Current version returns a different result for e.g. `WASM_PAGE_SIZE`, it returns 2 but the correct value would be 1.

So I think it allocates more than necessary when `ptr` is a multiple of `WASM_PAGE_SIZE`.

@ulan do you agree with my analysis?

osa1: To avoid overflows when `ptr` is larger than `u32::MAX - WASM_PAGE_SIZE` we should do this arithmetic on `u64`s.

ulan: I agree with the analysis. The overflow handling is indeed subtle.
+1 to doing the expected arithmetic in `u64`

rts/motoko-rts/src/gc/mark_compact/mark_stack.rs
ulan: Thanks, got it! That's a nice trick. It will break if we switch to page-based layout. We would need non-contiguous stack at that point.

rts/motoko-rts/src/memory.rs
osa1: This is Wasm heap size (4GiB). I will add a `WASM_HEAP_SIZE_BYTES` constant for this.

osa1: Done in #2632.

rts/motoko-rts/src/memory/ic.rs
osa1: I've added a constant for Wasm page size in #2632.

rts/motoko-rts-macros/src/lib.rs
```diff
@@ -0,0 +1,145 @@
+use proc_macro::TokenStream;
+use quote::quote;
+
+#[proc_macro_attribute]
+pub fn ic_fn(_attr: TokenStream, input: TokenStream) -> TokenStream {
```
crusso, line 5: Would it be possible/worth it to check that the argument/result types are restricted to the handful of legal Wasm types?

rts/motoko-rts-macros/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+        #[export_name = #fn_name]
+        unsafe extern "C" fn #fn_wrapper_ident(#(#wrapper_params_syn,)*) #wrapper_ret {
+            #fn_ident(#(#wrapper_args_syn,)*)
+        }
+    )
+    .into()
+}
+
+#[proc_macro_attribute]
+pub fn ic_mem_fn(attr: TokenStream, input: TokenStream) -> TokenStream {
```
crusso, line 65: Ditto here I guess.

rts/motoko-rts-tests/src/gc.rs
```diff
@@ -0,0 +1,274 @@
...
+    }
+}
+
+/// Check the dynamic heap:
+///
+/// - All and only reachable objects should be in the heap. Reachable objects are those in the
+///   transitive closure of roots.
+///
+/// - Objects should point to right objects. E.g. if object with index X points to objects with
+///   indices Y and Z in the `objects` map, it should point to objects with indicex Y and Z on the
```
crusso, line 83:
```suggestion
///   indices Y and Z in the `objects` map, it should point to objects with indices Y and Z on the
```

rts/motoko-rts-tests/src/gc.rs
```diff
@@ -0,0 +1,274 @@
...
+use motoko_rts::gc::mark_compact::compacting_gc_internal;
+use motoko_rts::types::*;
+
+use std::collections::{HashMap, HashSet};
+use std::fmt::Write;
+
+pub fn test() {
+    println!("Testing garbage collection ...");
+
+    // TODO: Add more tests
```
crusso, line 24: A nice one to tru might be some large binary tree with a linear number of nodes due to sharing, eg:
```
let p0 = ();
let p1= (p0, p0);
...
let pn = (pn-1, pn-1);
```
and then maybe a cyclic version of that.

osa1: Noted, I will add this case when I open a PR to add tests.

rts/motoko-rts-macros/src/lib.rs
osa1: In macros we only have the syntactic information, for example we know the names of the argument types but nothing else about those types. However the compiler already checks that types are "FFI safe", for example if I try something like

```rust
pub struct MyStruct {
    x: usize,
    y: String,
}

#[no_mangle]
extern "C" fn my_fn() -> MyStruct {
    MyStruct {
        x: 10,
        y: "test".to_owned(),
    }
}
```

Here it's not clear how to return this struct using C ABI, so the compiler generates this:

```
warning: `extern` fn uses type `MyStruct`, which is not FFI-safe
 --> src/lib.rs:7:26
  |
7 | extern "C" fn my_fn() -> MyStruct {
  |                          ^^^^^^^^ not FFI-safe
  |
  = note: `#[warn(improper_ctypes_definitions)]` on by default
  = help: consider adding a `#[repr(C)]` or `#[repr(transparent)]` attribute to this struct
  = note: this struct has unspecified layout
```

rts/motoko-rts/src/gc/copying.rs
```diff
@@ -63,96 +141,28 @@ unsafe fn evac(
...
     let obj = obj as *mut Obj;
 
     crate::visitor::visit_pointer_fields(obj, begin_from_space, |field_addr| {
-        evac(
-            begin_from_space,
-            begin_to_space,
-            end_to_space,
-            field_addr as usize,
-        );
+        evac(mem, begin_from_space, begin_to_space, field_addr as usize);
```
crusso, line 150: Now `end_to_space` is maintained as heap ptr in mem, right?

rts/motoko-rts/src/memory.rs
```diff
@@ -0,0 +1,57 @@
...
+///
+/// ```
+/// unsafe fn allocating_function<M: Memory>(mem: &mut M) { ... }
+/// ```
+///
+/// we implement (or generate with a macro)
+///
+/// ```
+/// #[no_mangle]
+/// unsafe extern "C" fn export_name() { allocating_function(crate::memory::ic::IcMemory) }
```
crusso, line 23: Is `export_name` just `allocating_function`?

crusso (approved):
I must admit, this PR is kinda large to review now, but hopefully some of the comments are useful....

rts/motoko-rts/src/memory/ic.rs
osa1: `grow_memory` refactored in #2633. Marking as resolved.

rts/motoko-rts-tests/src/gc.rs
osa1: > Indeed, I wonder if adding some monadic syntax to let you build heap graphs using monadic syntax would be useful for writing tests. Does Rust let you do that easily?

Not easily possible with Rust syntax, but we could use macros.

rts/motoko-rts-tests/src/gc/heap.rs
```diff
@@ -0,0 +1,342 @@
...
+            // Store length: tag + refs
+            write_word(
+                dynamic_heap,
+                heap_offset,
+                u32::try_from(refs.len() + 1).unwrap(),
+            );
+            heap_offset += WORD_SIZE;
+
+            // Store object value (tag)
+            write_word(dynamic_heap, heap_offset, obj << 1);
```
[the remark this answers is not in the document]
osa1: I'll add a function for this and use it.

rts/motoko-rts-tests/src/gc/heap.rs
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    // println!("object addresses={:#?}", object_addrs);
+
+    // Second pass adds fields
+    for (obj, refs) in refs {
+        let obj_offset = object_addrs.get(obj).unwrap() - heap_start;
+        for (ref_idx, ref_) in refs.iter().enumerate() {
+            // -1 for skewing
+            let ref_addr = object_addrs.get(ref_).unwrap().wrapping_sub(1);
```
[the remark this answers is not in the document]
osa1: We have `unskew` in the RTS. I'll try to use it, or add a similar one in tests.

rts/motoko-rts-tests/src/gc/heap.rs
```diff
@@ -0,0 +1,342 @@
...
+
+        let mutbox_addr = heap.as_ptr() as usize + mutbox_offset;
+        write_word(
+            heap,
+            root_addr_offset,
+            u32::try_from(mutbox_addr).unwrap().wrapping_sub(1),
+        );
+
+        root_addr_offset += WORD_SIZE;
+        mutbox_offset += 2 * WORD_SIZE;
```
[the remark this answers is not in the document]
osa1: This is actually `size_of::<MutBox>()`. I'll update.

rts/motoko-rts-tests/src/gc/utils.rs
```diff
@@ -0,0 +1,30 @@
+use byteorder::{ReadBytesExt, WriteBytesExt, LE};
+
+/// A unique object index, used in heap descriptions
+pub type ObjectIdx = u32;
```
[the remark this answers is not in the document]
osa1: Good point, yes. I'll document this. We could also make this type opaque (newtype) and provide a constructor that checks for this invariant. Not sure if it's necessary though.

rts/motoko-rts-tests/src/mark_stack.rs
```diff
@@ -1,40 +1,57 @@
...
-            push_mark_stack(*obj);
+        for obj in &objs {
+            push_mark_stack(mem, *obj as usize);
         }
 
         for obj in objs.iter().rev() {
             let popped = pop_mark_stack();
-            if popped != Some(*obj) {
+            if popped != Some(*obj as usize) {
                 free_mark_stack(); // TODO: Does not really free
```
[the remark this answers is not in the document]
osa1: I think this TODO is invalid, there's nothing to be done to free the mark stack, so just setting `STACK_TOP` etc. `NULL` is enough. I will remove the TODO.

rts/motoko-rts/src/float.rs
```diff
@@ -1,9 +1,12 @@
...
 use crate::types::{Bytes, SkewedPtr};
 
+use motoko_rts_macros::ic_mem_fn;
+
 // The meaning of the `mode` parameter is documented in motoko-base, function Float.format()
-#[no_mangle]
-unsafe extern "C" fn float_fmt(a: f64, prec: u32, mode: u32) -> SkewedPtr {
+#[ic_mem_fn]
+unsafe fn float_fmt<M: Memory>(mem: &mut M, a: f64, prec: u32, mode: u32) -> SkewedPtr {
     // prec and mode are tagged (TODO (osa): what tag???)
```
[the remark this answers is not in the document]
osa1: Right, so this code was ported from the C implementation of RTS and I added this TODO to ask about it later and document it. Would be good to know what "tag" means here and document it.

rts/motoko-rts/src/gc/copying.rs
```diff
@@ -1,12 +1,93 @@
...
-/// Evacuate (copy) an object in from-space to to-space, update end_to_space. If the object was
-/// already evacuated end_to_space is not changed.
+    // Note the stats
+    let new_live_size = end_to_space - begin_to_space;
+    note_live_size(Bytes(new_live_size as u32));
+
+    let reclaimed = (end_from_space - begin_from_space) - (end_to_space - begin_to_space);
+    note_reclaimed(Bytes(reclaimed as u32));
+
+    // Copy to-space to the beginning of from-space
```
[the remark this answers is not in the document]
osa1: > Is that to accommodate a memory.grow on the next round of mutation and avoid interpreting pointers as offset relative to the current to_space base?

It is to allow bump allocation until the end of Wasm heap without loss of efficiency. If keep using to-space we will have less room to bump the heap pointer until end of Wasm heap. Before copying GC the heap is like this:

```
| static heap | dynamic heap |........(unallocated)...........|
^             ^              ^                                ^
0           heap base      heap pointer                  end of Wasm heap
                           (allocation pointer)              (4GiB)
```

Allocation is done by bumping the heap pointer until it reaches end of Wasm heap, and calling `memory.grow` on the way to allocate Wasm pages when needed.

After copying live data to to-space it looks like this:

```
| static heap | from-space   |to-space  |....(unallocated)....|
^             ^              ^          ^                     ^
0           heap base      to-space   heap pointer          end of Wasm heap
```

Now if we don't move to-space back to from-space we'll have much less space to allocate (the "unallocated" part).

We could implement a slow path for allocation that, when `hp == end of Wasm heap`, checks if we have any space between heap base and beginning of current space. Or more generally, we could make allocation area a list of ranges on the heap (though I think we will need at most two ranges in this list), so handle cases like:

```
| static heap | to-space | from-space | to-space |
                                                 ^
                                            end of Wasm heap
```

Which becomes this after GC:

```
| static heap | alloc area | live data | alloc area |
                                                    ^
                                               end of Wasm heap
```

rts/motoko-rts/src/gc/copying.rs
osa1: Right.

rts/motoko-rts/src/memory.rs
osa1: `export_name` is whatever we want to call this function in the .wasm file generated for the RTS.  Without `no_mangle` the compiler mangles the name, so for example `allocating_function` above will have a name like `motoko_rts::allocation_function::h19cee68178dfb200`. We add `no_mangle` to the wrapper function and then give it a name that we later use to refer to from the generated code.

Does this make sense? Perhaps I should update this comment to make it more clear.

rts/motoko-rts-tests/src/gc.rs
```diff
@@ -0,0 +1,274 @@
...
+                        static_roots,
+                        closure_table_address,
+                        // note_live_size
+                        |_live_size| {},
+                        // note_reclaimed
+                        |_reclaimed| {},
+                    );
+                }
+            }
+        }
```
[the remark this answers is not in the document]
osa1: So the problem here is `compacting_gc_internal` and `copying_gc_internal` functions have generic parameters, so they're like C++ templates rather than OCaml/Haskell functions with type parameters. We can't return them from functions for two reasons (1) type system not powerful enough for higher rank types, so we can't return a value of type `forall x y z . ...` (2) the runtime does not support values with type parameters (which would require monomorphising in runtime). I also mentioned this briefly in the `GC` type, copying here:

```rust
/// Enum for the GC implementations. GC functions are generic so we can't put them into arrays or
/// other data types, we use this type instead.
#[derive(Debug, Clone, Copy)]
pub enum GC {
    Copying,
    MarkCompact,
}
```

rts/motoko-rts-tests/src/gc/utils.rs
```diff
@@ -28,3 +31,15 @@ pub fn read_word(heap: &[u8], offset: usize) -> u32 {
 pub fn write_word(heap: &mut [u8], offset: usize, word: u32) {
     (&mut heap[offset..]).write_u32::<LE>(word).unwrap()
 }
+
+/// Make a scalar value to be used in heap object payload
+pub fn make_scalar(value: u32) -> u32 {
+    // Scalar values can be at most 31 bits
+    assert_eq!(value >> 31, 0);
```
crusso, line 38: does the assert disappear in production?

osa1: No, that would be `debug_assert_eq`. But even if I used that, this is in tests, and tests are run in debug mode (which does more checking, for example overflows in arithmetic is checked in debug mode).

crusso (approved):
Curious why you got rid of ic_fn, but LGTM

rts/motoko-rts-tests/src/gc/utils.rs
crusso: Nvm.

osa1:
> Curious why you got rid of ic_fn, but LGTM

It wasn't as useful as `ic_mem_fn`. What it did was it converted this

```
#[ic_fn]
fn blah() { ... }
```

into

```
// original function, to be used in tests
fn blah(...) { ... }

// wrapper to be called from generated code
#[cfg(feature = "ic")]
#[no_mangle]
fn ic_blah(...) { blah(...) }
```

I can do the same without using the macro by adding `cfg(...)` and `no_mangle` manually:

```
#[cfg_feature = "ic"]
#[no_mangle]
fn blah() { ... }
```

and use it in both tests and from generated code.

rts/motoko-rts/src/gc/copying.rs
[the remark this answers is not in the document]
crusso: Yeah, all makes sense, thanks for the explanation!

PR #2587, opened 2021-06-12
Release build: embed actual version number

nomeata:
this way we’ll get something like
```
$ moc --version
Motoko compiler 0.1 (source cq98vl2b-xnb74m6b-2k3959b5-7jjskw5s)
```

I still want to include the source id, as a safe guard against two
builds for the same version floating around.

This means that the create-release-on-tag job will have to rebuild. I am
_not_ running the tests there, hoping that they won’t break if one
changes the banner…

Only release builds will have a version numbers. Builds of arbitrary
revisions will print what the do now, and executables built in the nix
shell will continue to report the `git describe` output, as of now.

This also fixes an embarrassing but that was actually deploying linux binaries in `motoko-macos-0.1.tar.gz`, which means the previous release was kinda broken. We should do another release once this is merged.

### DIFF
--- a/release-files.nix
+++ b/release-files.nix
@@ -31,7 +31,7 @@ let
     nixpkgs.runCommandNoCC "motoko-release-${releaseVersion}" {} ''
       mkdir $out
       cp ${as_tarball "x86_64-linux" (with linux; [ mo-ide mo-doc moc ])} $out/motoko-linux64-${releaseVersion}.tar.gz
-      cp ${as_tarball "x86_64-darwin" (with linux; [ mo-ide mo-doc moc ])} $out/motoko-macos-${releaseVersion}.tar.gz
+      cp ${as_tarball "x86_64-darwin" (with darwin; [ mo-ide mo-doc moc ])} $out/motoko-macos-${releaseVersion}.tar.gz
       cp ${as_js "moc" linux.js.moc} $out/moc-${releaseVersion}.js
       cp ${as_js "moc-interpreter" linux.js.moc_interpreter} $out/moc-interpreter-${releaseVersion}.js
     '';
--- a/release-files.nix
+++ b/release-files.nix
@@ -7,8 +7,8 @@
 { releaseVersion ? "latest" }:
 let
   nixpkgs = import ./nix { };
-  linux = import ./default.nix { system = "x86_64-linux"; internal = true; };
-  darwin = import ./default.nix { system = "x86_64-darwin"; internal = true; };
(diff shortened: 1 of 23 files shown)


nomeata:
Changes to this flow are really hard to test. I had to create a forked repo, add the cachix secret key, and make test releases there. Oh well.

src/source_id/gen.sh
```diff
@@ -2,13 +2,18 @@
 
 cd "$(dirname "${BASH_SOURCE[0]}")"
 
-file="source_id.ml"
+file="generated.ml"
+
+if [ -n "$MOTOKO_RELEASE" ]
```
crusso, line 7: I wonder if it would have been simpler to just us the ppx extension to include a version file rather than generate some source code, but no biggie. I guess this predates our embrace of ppx.

release-files.nix
```diff
@@ -31,7 +31,7 @@ let
     nixpkgs.runCommandNoCC "motoko-release-${releaseVersion}" {} ''
       mkdir $out
       cp ${as_tarball "x86_64-linux" (with linux; [ mo-ide mo-doc moc ])} $out/motoko-linux64-${releaseVersion}.tar.gz
-      cp ${as_tarball "x86_64-darwin" (with linux; [ mo-ide mo-doc moc ])} $out/motoko-macos-${releaseVersion}.tar.gz
+      cp ${as_tarball "x86_64-darwin" (with darwin; [ mo-ide mo-doc moc ])} $out/motoko-macos-${releaseVersion}.tar.gz
       cp ${as_js "moc" linux.js.moc} $out/moc-${releaseVersion}.js
       cp ${as_js "moc-interpreter" linux.js.moc_interpreter} $out/moc-interpreter-${releaseVersion}.js
```
crusso, line 36: Why only `linux.js.moc(_interpreter)` and no `darwin....`? Oversight or deliberate?

crusso (approved):
Approving just to unblock you if necessary (not that confident reviewing this)

crusso:
> this way we’ll get something like
> 
> ```
> $ moc --version
> Motoko compiler 0.1 (source cq98vl2b-xnb74m6b-2k3959b5-7jjskw5s)
> ```

Looks sensible, but why no third revision number?

release-files.nix
nomeata: Deliberate, for the JS builds it doesn’t matter which OS built them.

nomeata:
> Looks sensible, but why no third revision number?

That was just a local test build, sorry for the confusion.

PR #2594, opened 2021-06-14
RTS makefile: Track .rs files in subdirectories

osa1:
This fixes tracking of .rs files in subdirectories by the RTS Makefile.
Previously a file in e.g. `motoko-rts/src/my_dir` would not be tracked, only
the ones at the top-level `motoko-rts/src` would be tracked. With this patch it
now recursively tracks all .rs files.

### DIFF
--- a/rts/Makefile
+++ b/rts/Makefile
@@ -157,9 +157,9 @@ $(MUSL_WASM_A): $(MUSL_WASM_O)
 RTS_RUST_WASM_A=_build/wasm/libmotoko_rts.a
 RTS_RUST_DEBUG_WASM_A=_build/wasm/libmotoko_rts_debug.a
 
+RTS_RUST_FILES=$(shell ls **/*.rs)
+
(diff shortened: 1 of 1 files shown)


rts/Makefile
```diff
@@ -157,9 +157,9 @@ $(MUSL_WASM_A): $(MUSL_WASM_O)
 RTS_RUST_WASM_A=_build/wasm/libmotoko_rts.a
 RTS_RUST_DEBUG_WASM_A=_build/wasm/libmotoko_rts_debug.a
 
-TOMMATH_BINDINGS_RS=_build/tommath_bindings.rs
-
+RTS_RUST_FILES=$(shell ls **/*.rs)
```
nomeata, line 160: Don’t you have to set `SHELL` to enable globstar?
```suggestion
# this relies on bash and globstar, see https://stackoverflow.com/questions/2483182/recursive-wildcards-in-gnu-make
RTS_RUST_FILES=$(shell ls **/*.rs)
```

nomeata (approved):

PR #2596, opened 2021-06-14
Lg/typo fix

lsgunnlsgunn:


crusso (approved):
LGTM. Skimming it, I think that section could use another pass by either myself or Andreas, but I'm fine with fixing these typos first.

PR #2605, opened 2021-06-18
Fix issue #2319 (crash on import of candid class)

### LINKED ISSUE #2319: Bad error message when importing `.did` that isn't a service
Users out there ran into an exception: https://forum.dfinity.org/t/build-error-with-imported-actor-class/1931

They built a canister using `actor class…`, which produced this `.did`:
```
type anon_class_1_1 = 
 service {
 };
service : () -> anon_class_1_1
```
which then tripped this code:
```
    match actor with
    | Some {it=ServT ms; _} ->
       let fs = List.map (check_meth env) ms in
       let fs = M.Env.fold (fun id t fs ->
         match t with
         | M.Con (c, _) -> M.{lab = id; typ = M.Typ c}::fs
         | _ -> assert false) !m_env fs in
       M.Obj (M.Actor, List.sort M.compare_field fs)
    | None -> assert false
    | _ -> assert false 
```

These `assert false` should produce more helpful error messages.

(Of course, we also should support importing canister classes, but that's a bigger thing; this issue is _just_ about proper error messages.)

crusso:
* errors, doesn't crash, on import of candid service class
* improves on UnsupportedCandid error reporting, adding source location and message.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,7 @@
 # Motoko compiler changelog
 
+* Fix issue #2319 (crash on import of Candid class)
+
 == 0.6.4 (2021-06-12)
 
 * For release builds, the banner (`moc --version`) now includes the release
--- a/src/exes/candid_tests.ml
+++ b/src/exes/candid_tests.ml
@@ -101,8 +101,8 @@ let mo_of_test tenv test : (string * expected_behaviour, string) result =
     | ParsesEqual (false, i1, i2)
     -> Ok (defs ^ not_equal (deser ts i1) (deser ts i2), ShouldPass)
   with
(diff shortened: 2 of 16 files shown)


chenyan-dfinity (approved):

crusso:
@chenyan-dfinity PTAL
BTW, do you know if we actually type-check (i.e. kind-check) the imported candid file before or during import? There might be a an unlikely, but possible, security hole if we don't.

chenyan-dfinity (approved):
Yes, we type check the did file when importing. https://github.com/dfinity/motoko/blob/master/src/pipeline/pipeline.ml#L337

PR #2607, opened 2021-06-19
typo

cPlayIt:
Not best-practice to use former/latter with lists more than 2


crusso (approved):

crusso:
Incorporated into #2619

PR #2609, opened 2021-06-19
reworded

cPlayIt:
a bit confusing as written, being that a pun is typically play on a word implying a different meaning of a term, does alias make more sense here?

### DIFF
--- a/doc/modules/language-guide/pages/pattern-matching.adoc
+++ b/doc/modules/language-guide/pages/pattern-matching.adoc
@@ -12,7 +12,7 @@ Consider the following function call:
 let name : Text = fullName({ first = "Jane"; mid = "M"; last = "Doe" });
 ....
 
-We see that an object with three fields is being built, then passed to the function `fullName`, and the returned result is being captured and brought into scope by binding it to the `name` identifier. The latter step is referred to as pattern matching, and `name : Text` is a pattern. When we consider the callee, we'll encounter something like the following:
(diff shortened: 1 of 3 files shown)


doc/modules/language-guide/pages/pattern-matching.adoc
```diff
@@ -21,7 +21,7 @@ func fullName({ first : Text; mid : Text; last : Text }) : Text {
 };
 ....
 
-The input is an (anonymous) object, which is deconstructed into its three `Text` fields, whose values are bound to the identifiers `first`, `mid` and `last`. They can be freely used in the block that forms the body of the function. Above we have resorted to _name punning_ for object field patterns, their general form allows the binding identifier to be distinct from the field name, as in `mid = middle_name : Text`.
+The input is an (anonymous) object, which is deconstructed into its three `Text` fields, whose values are bound to the identifiers `first`, `mid` and `last`. They can be freely used in the block that forms the body of the function. Above we have resorted to _name aliasing_ for object field patterns, this general form allows the binding identifier to be distinct from the field name, as in `mid = middle_name : Text`.
```
crusso, line 24:
```suggestion
The input is an (anonymous) object, which is deconstructed into its three `Text` fields, whose values are bound to the identifiers `first`, `mid` and `last`. They can be freely used in the block that forms the body of the function. Above we have resorted to _name punning_ (a form of aliasing) for object field patterns, using the name of a field to also name its contents. A more general form of field pattern allows the content to be named separately from the field, as in `...; mid = m : Text; ...`. Here `mid` determines which field to match, and `m` names the content of that field within the scope of the pattern.
```

crusso (approved):

crusso:
Incorporated into #2619

PR #2610, opened 2021-06-19
word mix-up

cPlayIt:
I believe destructured is the correct word here.

### DIFF
--- a/doc/modules/language-guide/pages/pattern-matching.adoc
+++ b/doc/modules/language-guide/pages/pattern-matching.adoc
@@ -21,7 +21,7 @@ func fullName({ first : Text; mid : Text; last : Text }) : Text {
 };
 ....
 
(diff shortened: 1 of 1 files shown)


crusso (commented):
I actually prefer deconstructed since destructuring implies destruction of the original value (which is not the case).

@rossberg?

cPlayIt:
Destructuring assignment is a specialized term in JavaScript and PHP
documentation, however I am not familiar with how things are represented in
other languages.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment
https://www.php.net/manual/en/migration71.new-features.php#migration71.new-features.symmetric-array-destructuring

rossberg:
"Destructuring" is the established term in a number of communities, so I'm fine with using it.

crusso:
Superseeded by #2620

PR #2613, opened 2021-06-19
possible error

cPlayIt:
I'm just learning so a bit unsure, but I believe scrutineer (with the "r") may have been the word the author was going for, but in this context "alternative" probably makes more sense? ...

### DIFF
--- a/doc/modules/language-guide/pages/pattern-matching.adoc
+++ b/doc/modules/language-guide/pages/pattern-matching.adoc
@@ -108,7 +108,7 @@ The last kind of pattern is the `or`-pattern. As their name suggests, these are 
 Since pattern matching has a rich history and interesting mechanics, a few additional comments are justified.
 
 - nomenclature
-The (usually structured) expression that is being matched is frequently called the _scrutinee_ and the patterns appearing behind the keyword `case` are the _alternatives_. When every possible scutinee is matched by (at least one) alternative, then we say that the scrutinee is _covered_. The patterns are tried in top-down fashion and thus in case of _overlapping_ patterns the one higher-up is selected. A scrutinee is considered _dead_ (or _inactive_), if for every value that it matches there is higher-up alternative that is also matched.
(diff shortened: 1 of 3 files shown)


ggreif (approved):

ggreif:
There are no such things as _dead scrutinees_ in a call-by-value language. You are correct that _alternative_ is the right word. It is also defined a few lines up.

cPlayIt:
Thanks, never worked in a strongly typed language.

doc/modules/language-guide/pages/pattern-matching.adoc
```diff
@@ -108,7 +108,7 @@ The last kind of pattern is the `or`-pattern. As their name suggests, these are
...
 
 - nomenclature
-The (usually structured) expression that is being matched is frequently called the _scrutinee_ and the patterns appearing behind the keyword `case` are the _alternatives_. When every possible scutinee is matched by (at least one) alternative, then we say that the scrutinee is _covered_. The patterns are tried in top-down fashion and thus in case of _overlapping_ patterns the one higher-up is selected. A scrutinee is considered _dead_ (or _inactive_), if for every value that it matches there is higher-up alternative that is also matched.
+The (usually structured) expression that is being matched is frequently called the _scrutinee_ and the patterns appearing behind the keyword `case` are the _alternatives_. When every possible scutinee is matched by (at least one) alternative, then we say that the scrutinee is _covered_. The patterns are tried in top-down fashion and thus in case of _overlapping_ patterns the one higher-up is selected. An alternative is considered _dead_ (or _inactive_), if for every value that it matches there is higher-up alternative that is also matched.
```
crusso, line 111:
```suggestion
The (usually structured) expression that is being matched is frequently called the _scrutinee_ and the patterns appearing behind the keyword `case` are the _alternatives_. When every possible scrutinee is matched by (at least one) alternative, then we say that the scrutinee is _covered_. The patterns are tried in top-down fashion and thus in case of _overlapping_ patterns the one higher-up is selected. An alternative is considered _dead_ (or _inactive_), if for every value that it matches there is higher-up alternative that is also matched.
```

crusso:
Incorporated into #2619

PR #2626, opened 2021-06-24
Stable Memory: upgrade mechanism

crusso:
Design and implementation of direct StableMemory access, coexisting with Motoko stable variables.

See design/StableMemory.md for overview.

This PR implement the basic upgrade mechanism and operations for growing (virtual) Stable Memory, plus load/store for Nat32 only.

Split into several PR's for ease of reviewing.

* #2753 implements numeric load/store operations for fixed-size numerics (including floats), (with tests)
* #2762 implements load/store of blob contents (with tests)
* #2723 extends the (mutable) game of life test to add another upgrade that uses stable memory.
* https://github.com/dfinity/motoko-base/pull/280 is the proposed new base library

TODO:
- [x] StableMemory base library:
     * mostly adding test/run-drun/stable-mem/StableMemory.mo to base, but :
     * should we use Nat32 or Nat for offsets, presaging StableMemory64? Using Nat32 unless told otherwise.
- [ ] Documentation
- [ ] We are missing load/store of Nat/Int/Char/Bool/Principal/Text. Do we want them? Can always add later or hand roll


design/StableMemory.md
```diff
@@ -0,0 +1,88 @@
...
+
+To provide more fine-grained access to stable memory we propose:
+
+A) extending Motoko with access to (monomorphic) Candid serialization primitives.
+   This is most easily achieved by extending the syntax with saturated serialization and deserialization
+   constructs, along the lines of debug_show.
+
+```
+   encode<T>(v : T) : Blob
+   decode<T>(b : Blob) : T
```
kritzcreek, line 21: Should this return ?T. I'd want to be able to handle deserialization failure.

crusso: I'm not sure we have that option with the current serialization code. Even if we did, we have to trust that the contents actually has type `T` I think, but maybe I'm mistaken. @rossberg?

rossberg: Hm, wouldn't these signatures depend on adding type-passing polymorphism first? Short of that, I suspect we can only support monomorphic functions.

And yeah, we probably need to make these return options. I assume that's can be made to work with the deserialisation code, since it already has a mechanism for abort and fall back for mismatches on option types.

design/StableMemory.md
```diff
@@ -0,0 +1,88 @@
...
+Direct reads and writes of word-sized data to/from the stack are not supported but can be emulated at cost.
+
+The Minimal Motoko prims could be:
+
+```
+module StableMemory {
+  stableSize : () -> (page_count : Nat32); // hidden?
+  stableGrow : (new_pages : Nat32) -> (old_page_count : Nat32); // hidden?
+  stableWrite : (offset : Nat32, b : Blob) -> (); // write size + contents of blob to memory, growing stable memory region on demand
+  stableRead : (offset : Nat32) -> Blob // read Blob size & contents from memory at offset into fresh blob, trapping if out-of-bounds
```
rossberg, line 50: Hm, this interface involves a substantial allocation and de/serialisation overhead for every single access to stable memory. That would be pretty expensive for some needs.

But more importantly, the serialisation format produces blobs whose size are variable not just by type but also by value. That makes it practically impossible to update any data in place with just this API.

In order to provide sufficient control over memory layout, I think we need something more low-level, that e.g. enables reading/writing basic types like numbers (and perhaps vectors thereof) directly, with fixed size.

We can still provide the variable-size blob functions, but I suspect their utility is limited to cases where the app is fine with essentially custom-implementing full dynamic memory management inside stable memory (which still requires fixed-size read/writes to boot).

design/StableMemory.md
```diff
@@ -0,0 +1,108 @@
...
+
+The raw API of the previous section might be workable but is totally unsafe, allowing writes to arbitrary offsets, possible corrupting data, and reads from random offset,
+producing unpredictable behaviour.
+
+Maybe a stable `ref T` type is a safer abstraction for stable memory?
+
+* type `ref T`:  represent a fixed address in stable memory. The address stores current offset in stable memory of contents of r (T must be stable).
+* `ref v` : allocates the next address, storing the serialized blob of `v` in stable memory, writing address of that blob into `r`, returns the address.
+* `! r`: deserializes `v` from some blob in stable memory at current offset stored in stable reference `r`.
+* `r := v' frees content blob `v` of r and serializes v' to (re)allocated memory, updating
```
crusso, line 68:
```suggestion
* `r := v' frees content blob `v` of r and serializes `v` to (re)allocated memory, updating
```

design/StableMemory.md
crusso: Sure they should probably just be overloaded constructs applicable to any T  where T shared (just like  equality and similar to debug_show). Just sketching here.

design/StableMemory.md
crusso: In that case, maybe we should provide just raw word level memory access, and let the user write custom picklers or whatever. 
That pretty low-level though.

The internet identity service actually uses the technique of an array fixed size regions, with candid serialization in and out of each entry. That supports a modicum of lazy upgrade, but writes can fail if they exceed the slot budget.

crusso:
@rossberg PTAL at the idea of using a much safer, cross-heap ref type. 

Otherwise, let's do the simplest thing and just give access to raw memory, with some convenience functions for writing common types, safety be damned.

rossberg:
Ah, sorry, I somehow missed the update.

Unfortunately, I think the ref abstraction still is way too limited. I think a basic requirement is that you ought to be able to lay out arbitrary (arrays|records of)* base values in stable memory, and be able to do in-place update of individual fields in them. I don't see how the ref interface can achieve that. Short of introducing a full Candid-style type algebra for describing data layout, a safe abstraction won't work I'm afraid.

That is, I believe there are basically two possible approaches for exposing stable memory:

1. Provide direct low-level access to stable memory, so that everything else can be coded as libraries that have full control over memory layout (including locality).
2. Introduce a fully typed high-level abstraction and type algebra, and let the compiler/runtime take control.

Obviously, (1) is not very safe and user-friendly, but much more flexible, lightweight and easy to provide, that's why I suggested starting with it. Moreover, if somebody wants to build, say, a high-performance database or file system in stable memory, I think they'll need (1).

(2) is essentially what we do now with stable variables, just with another implementation underneath. Such a "live" implementation of stable vars could be done, and would be nice and safe. But I doubt that it would satisfy all production use cases. It's sufficient for simple apps, but not for building custom persistence libraries, which was our goal here.

In an ideal world, a variant of (2) should actually be implementable on top of (1).

crusso:
`Peek` and `Poke`, here we come!

rossberg:
Somehow that gives me a warm fuzzy feeling. 38911 bytes free.

crusso:
Oh, did you have a C-64 too? I cut my teeth on that (well, after the Vic-20)

crusso:
@rossberg 

I've been thinking about this off and on, but the trickiest bit is actually supporting both  raw access and the existing stable vars while providing full access to raw stable memory. Absent any other stability mechanism, we need to reserve some area for meta-data (e.g. higher watermark of accessed raw memory to determine beginning of stable variable data.)


If we don't want to do that, then we need some protocol for reading/writing raw stable memory so users stable memory and and stable variable storage can co-exist.

Something like:
On initialization, if stable memory pages > 0 and first word of stable memory is non-zero then memory starts with encoding of stable vars and contains no raw user data.

Otherwise, if word 0 == 0 then address 1 contains highwater mark of raw access, stable memory starts at address 8 and any stable var storage starts at highwater mark+1.

The raw memory api maintains the higherwater mark during access and prevents writes to location 0-7. On pre-upgrade, we write stable memory starting at highwater mark+1.

What I don't like about this approach is that it makes it difficult for a Motoko program to use the stable memory of an application written in another language, say Rust, since we impose some semantics on the structure of stable memory to accommodate Motoko's stable variables.

Maybe we should make it all or nothing. If your program declares stables vars, then you don't have access to stable memory. But if it doesn't then you have full access. 

We can even provide some functionality for a program to upgrade from using stable vars to do not using them by providing a (type annotated) deserialization construct that would allow the upgraded program to read the stable vars as a record before itself writing to stable memory with raw access.

rossberg:
> On initialization, if stable memory pages > 0 and first word of stable memory is non-zero then memory starts with legacy encoding of stable vars and contains no raw user data.

Do we know that it can't be zero in the legacy encoding?

The meta data doesn't need to be in stable memory itself, except during the upgrade, correct? And Motoko code cannot access stable memory during upgrade. So can't we save the meta data (including the high water mark) to the top end of the stable memory (relative to its current size), after having written the stable var? Then it wouldn't interfere with raw memory layout.

There is another minor problem, though: to save stable vars and meta data, we'll generally need to grow the stable memory larger than what was requested by user code. That observably affects bounds checks, i.e., some raw memory access that ought to trap might not -- unless we implement custom bounds checking for it, which could be expensive.

> Oh, did you have a C-64 too? I cut my teeth on that (well, after the Vic-20)

Actually, no, but of course everybody else in school had one. My own nerdism started with an Atari ST.

crusso:
> > On initialization, if stable memory pages > 0 and first word of stable memory is non-zero then memory starts with legacy encoding of stable vars and contains no raw user data.
> 
> Do we know that it can't be zero in the legacy encoding?

I'd have to double check, but its basically a size plus serialization blob, which I don't think can ever be entirely empty, so the size should be zero. If there are no stable variables at all, no  blob is written and pages should be 0. I think we might even distinguish install from upgrade by looking at the stable memory page size (yuk)

> The meta data doesn't need to be in stable memory itself, except during the upgrade, correct? And Motoko code cannot access stable memory during upgrade. So can't we save the meta data (including the high water mark) to the top end of the stable memory (relative to its current size), after having written the stable var? Then it wouldn't interfere with raw memory layout.

Might work.

> There is another minor problem, though: to save stable vars and meta data, we'll generally need to grow the stable memory larger than what was requested by user code. That observably affects bounds checks, i.e., some raw memory access that ought to trap might not -- unless we implement custom bounds checking for it, which could be expensive.

I was assuming we'd need custom bounds checking.

That was part of the reason I suggested going for either stable vars or raw API. So the raw access could be at efficient as possible, and less fiddly. 


> > Oh, did you have a C-64 too? I cut my teeth on that (well, after the Vic-20)
> 
> Actually, no, but of course everybody else in school had one. My own nerdism started with an Atari ST.

rossberg:
> That was part of the reason I suggested going for either stable vars or raw API. So the raw access could be at efficient as possible, and less fiddly.

I see. That could be a fallback, though as you point out, it would require more interface surface to enable transition. I'm not sure how exactly we would check the mutual exclusion, though. The presence of a stable declaration sets some flag that needs to be checked in every call to a StableMem prim?

If bounds checks are the only issue, then I'd allow both to coexist. From the outset, that seems almost simpler.

crusso:
> > That was part of the reason I suggested going for either stable vars or raw API. So the raw access could be at efficient as possible, and less fiddly.
> 
> I see. That could be a fallback, though as you point out, it would require more interface surface to enable transition. I'm not sure how exactly we would check the mutual exclusion, though. The presence of a stable declaration sets some flag that needs to be checked in every call to a StableMem prim?

Since the flab value would be statically known, I think you would just compile the Prims differently depending on the value of the flag (i.e. trap or do the right thing, sans check).

> If bounds checks are the only issue, then I'd allow both to coexist. From the outset, that seems almost simpler.

crusso:
FTR, @nomeata questioned why we are giving up the on the idea of a more lazy implementation of stable vars.

rossberg:
> FTR, @nomeata questioned why we are giving up the on the idea of a more lazy implementation of stable vars.

@nomeata, we're not removing stable vars, but we also want to enable the choice of implementing more sophisticated abstractions as efficient in-mem libs (like K/V stores, databases) that allow upgrades without having to serialise gigs of data (imagine the cost of doing that for s.th like a BigMap distributed over 100 actors). Another way to say this is that we need to be on par with what folks can build in Rust.

design/StableMemory.md
matthewhammer: > but I suspect their utility is limited to cases where the app is fine with essentially custom-implementing full dynamic memory management inside stable memory (which still requires fixed-size read/writes to boot).

It seems very limited or bad when stated that way (which is a fair way to state it), but still, I think most apps will have a data model that makes that problem very domain-specific, and thus, possibly tractable to solve on a per canister basis, at least for initial cases.

For instance, the II service uses arrays of fixed-sized fields, indexed by array offset, as @crusso mentioned above. No need for general-purpose MM there.

If anything, I would imagine that apps want a lot of control over when data becomes "garbage", and may want a quasi-manual process for marking users' stable data as garbage, not something automagical or otherwise obscured by the language semantics.

crusso:
@rossberg PTAL

design/StableMemory.md
```diff
@@ -35,73 +30,204 @@ ic0.stable_write : (offset : i32, src : i32, size : i32) -> ();             // *
...
+canisters, as follows.
 
+During execution, abstract stable memory (StableMemory) is aligned
+with IC stable memory, at address 0, for reasonable efficiency (apart
+from maintaining `max`).
 
+During upgrade, we compute the size of the stable variable blob, shift
+that portion of StableMemory to the end of user memory (growing if
+necessary) and write the old format into the space just freed,
+followed by the size of StableMem.
```
rossberg, line 141: This may be more complicated and expensive than necessary, I think. Why shift all the memory? As you observe, you can distinguish old layout from new layout by checking whether the byte at address 0 is non-zero. So isn't it sufficient to just back-up this one byte somewhere else in the new scheme and null it out for upgrade?

Along with the idea of keeping grow explicit and dropping `max`, stable memory could then be used like this during upgrade:
```
[0]  0
[1..N-1]  StableMemory bytes
[N..M-7]  serialised stable vars (+ possible page padding)
[M-6..M-3]  value N/64Ki = StableMemory.size()
[M-2]  backed-up StableMemory byte 0
[M-1]  version byte
```
where

M = physical size of stable memory (as seen by system)
N = logical size of stable memory (as seen through StableMemory interface)

both being multiples of 64Ki, i.e., page boundaries.

crusso: I'm not shifting all memory just the portion occupied by stables vars but I'll consider this scheme instead. However, the first word (not byte) must be 0x0.

rossberg: > However, the first word (not byte) must be 0x0.

I was wondering about that. Isn't the first byte always non-zero in a Candid blob?

crusso: I think we write the size of the encoding plus the encoding itself at the moment, so the first 4 bytes aren't candid at all. But I'll check.

design/StableMemory.md
```diff
@@ -35,73 +30,204 @@ ic0.stable_write : (offset : i32, src : i32, size : i32) -> ();             // *
...
+The minimal Motoko prims could be:
 
 ```
 module StableMemory {
-  stableSize : () -> (page_count : Nat32); // hidden?
-  stableGrow : (new_pages : Nat32) -> (old_page_count : Nat32); // hidden?
-  stableWrite : (offset : Nat32, b : Blob) -> (); // write size + contents of blob to memory, growing stable memory region on demand
-  stableRead : (offset : Nat32) -> Blob // read Blob size & contents from memory at offset into fresh blob, trapping if out-of-bounds
+  writeNat8 : (offset : Nat8, n : Nat8) -> ();
+  readNat8 : (offset : Nat8) -> Nat8;
```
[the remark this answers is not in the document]
crusso: Ok. Should also be offset : Nat32 (not Nat8)

design/StableMemory.md
```diff
@@ -35,73 +30,204 @@ ic0.stable_write : (offset : i32, src : i32, size : i32) -> ();             // *
...
+ic0.stable_read_i32 : (offset : i32, size : i32) -> i32; // *
+// similarly for i64, f32, f64
+```
 
-In future, the GC could mark dead references as free, returning them to the allocator -
-there is no need to collect StableMemory store for references as they are not shared and cannot, by typing, occur in stable memory (or be sent to other actors).
+## Bikeshedding:
 
-Drawbacks:
+It might be preferable to arrange the API by type, with one nested module per type:
```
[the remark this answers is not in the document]
crusso: Right, but being uniform, you'd be able to abstract over them using combinators. Maybe if we add a size field, we could write array operations, parameterized by a load/store/size operations given by a single module.

```
module class Array<T>(t : module { load: Nat32 -> t; store : Nat32 -> t; size : ...} {
     func load( n : Nat32 -> [T]) : ()) { ...};
     func store( n : Nat32, a: [T]) : ()) { ... }
     let size = ...
}
```
But then maybe these should be record, not modules to allow more computation...

design/StableMemory.md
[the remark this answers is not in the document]
crusso: Should we do the load/store naming for blobs too?

design/StableMemory.md
crusso: confirmed

crusso:
@rossberg PTAL

design/StableMemory.md
```diff
@@ -35,73 +30,204 @@ ic0.stable_write : (offset : i32, src : i32, size : i32) -> ();             // *
...
-Q:
-* should we read/write size + contents or read/write size and contents separately?
-* should we have a more streaming API, that returns the next free offset?
-* should we use a more malloc style API, returning opaque addresses. If so, should we be able to update their blob content (size permitting) and even free them? Should we use GC to discover unreachable blobs and free them? If opaque addresses aren't sharable, we don't need to worry about gc'ing the blob themselves, but they can't be stored in stable vars either unless we allow Candid(ish) stable data to additionally include opaque address.
-
-# Stable Ref types?
-
-The raw API of the previous section might be workable but is totally unsafe, allowing writes to arbitrary offsets, possible corrupting data, and reads from random offset,
-producing unpredictable behaviour.
+Each operation will maintain an internal global i32 `max` bounding the maximum address read or written (from above), initially 0, always <= 0xFFFFFFF0
```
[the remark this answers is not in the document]
crusso: Done

design/StableMemory.md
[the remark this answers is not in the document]
rossberg: Yes, consistency ftw.

design/StableMemory.md
```diff
@@ -46,38 +48,40 @@ The minimal Motoko prims could be:
...
 ```
 
-Each operation will maintain an internal global i32 `max` bounding the maximum address read or written (from above), initially 0, always <= 0xFFFFFFF0
-(since we need at least two words to store 0x00 and `max` in raw stable memory on upgrade.)
-`max` is really the size of the user-accessed address range.
+(NOTE: Motoko's `Nat32` value are always boxed - it might be more efficient to use `Nat` which is unboxed for 30(?)-bit values)
+
 
 ```
 fun writeNat8(offset, b) =
```
rossberg, line 68: Nit: adjust name

design/StableMemory.md
```diff
@@ -144,90 +153,115 @@ If there are no stable variables, we shift the first two words of
...
+  [N..N+3] StableVariable data len
+  [N+4..(N+4)+len-1] StableVariable data
+  [(N+4)+len-1,..M-1] 0...0 // zero padding
+(case !size > 0)
+[0..3]  0...0
+[4..N-1]  StableMemory bytes
+[N..N+3]  StableVariable data len
+[N+4..(N+4)+len-1] StableVariable data
+[(N+4)+len..M-3] 0...0 // zero padding
+[M-3..M-6] value N/64Ki = !size
```
rossberg, line 184:
```suggestion
[(N+4)+len..M-7] 0...0 // zero padding
[M-6..M-3] value N/64Ki = !size
```

design/StableMemory.md
```diff
@@ -144,90 +153,115 @@ If there are no stable variables, we shift the first two words of
...
+[M-5..M-2] saved StableMemory bytes
+[M-1]  version byte
```
rossberg, line 186: On second thought, why be greedy instead of aligning to a full word.

design/StableMemory.md
```diff
@@ -144,90 +153,115 @@ If there are no stable variables, we shift the first two words of
...
-    let blob = serialise(v) in
-    // stable grow page including address | 4 + |blob| + 4 + max |
-    [size(serialize(v))@@blob@@max@@mem[4 + |blob| + 4, ..., max -1 ]@@mem[0, ..., 4 + |blob| + 4  -1]
+(case !size == 0)
+  (case v == None)
+  <empty> (ie. ic0.stable_size() == 0)
+  (case v_opt = Some v)
+  [N..N+3] StableVariable data len
+  [N+4..(N+4)+len-1] StableVariable data
+  [(N+4)+len-1,..M-1] 0...0 // zero padding
```
rossberg, line 177: Don't we already know N=0 in this case?

design/StableMemory.md
```diff
@@ -144,90 +153,115 @@ If there are no stable variables, we shift the first two words of
...
-    mem.writeWord(4 + size, max) // save max
+
+```ocaml
+func save v_opt : value option =
+  let len, data = match v_opt with
+    | Some v -> serialize(v)
+    | None -> 0, []
+  in
+  if !size == 0 then
+    assert (ic0.stable_size() == 0);
```
rossberg, line 205: Is that correct? It might already have been grown in previous upgrades.

design/StableMemory.md
```diff
@@ -144,90 +153,115 @@ If there are no stable variables, we shift the first two words of
...
+      let len = mem[N,..,N+3] in
+      if len > 0
+        assert (N+4+len-1 <= ic0.stable_size() * pagesize)
+        let v = deserialize(len, N+4) in
+        mem[N+4, ..., N+4+len-1] := [0, ..., 0]; // clear memory
+        Some v
+      else None
+    else
+      size := 0;
+      let len = mem[0,..,3] in
```
rossberg, line 252:
```suggestion
      let len = marker in
```

design/StableMemory.md
```diff
@@ -133,7 +142,7 @@ canisters, as follows.
 
 During execution, abstract stable memory (StableMemory) is aligned
 with IC stable memory, at address 0, for reasonable efficiency (apart
-from maintaining `max`).
+from bound checks against logical `size()`).
 
 During upgrade, we compute the size of the stable variable blob, shift
 that portion of StableMemory to the end of user memory (growing if
```
rossberg, line 148: You might need to update this and following paragraphs as well.

design/StableMemory.md
crusso: You mean for the version? Yeah, I agree.

crusso: I'll leave 4 bytes for the version.

design/StableMemory.md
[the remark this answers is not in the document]
crusso: Done

design/StableMemory.md
crusso: Done

design/StableMemory.md
crusso: Done

design/StableMemory.md
crusso: I hate arithmetic

design/StableMemory.md
crusso: Sure, simplified

design/StableMemory.md
crusso: Quite right, removing the assert.

design/StableMemory.md
crusso: K

crusso:
@rossberg PTAL

design/StableMemory.md
```diff
@@ -55,35 +55,37 @@ module StableMemory {
...
    mem[offset]
+
+fun storeNat8(offset, b) =
+   assert (offset < StableMemory.size() * wasm_page_size);
+   mem[offset] := b
+
 ```
 
-(To avoid overflow on the rhs could do the check as `assert ((offset >> 6) < StableMemory.size())`.)
+(To avoid overflow on the rhs, we could do implement the check as `assert ((offset >> 6) < StableMemory.size())`.)
```
rossberg, line 78:
```suggestion
(To avoid overflow on the rhs, we could implement the check as `assert ((offset >> 6) < StableMemory.size())`.)
```

design/StableMemory.md
```diff
@@ -168,22 +179,22 @@ Stable memory layout (during execution):
...
-  [(N+4)+len-1,..M-1] 0...0 // zero padding
+  [0..3] StableVariable data len
+  [4..4+len-1] StableVariable data
+  [4+len-1,..M-1] 0...0 // zero padding
 (case !size > 0)
 [0..3]  0...0
 [4..N-1]  StableMemory bytes
 [N..N+3]  StableVariable data len
 [N+4..(N+4)+len-1] StableVariable data
 [(N+4)+len..M-3] 0...0 // zero padding
```
rossberg, line 194:
```suggestion
[(N+4)+len..M-13] 0...0 // zero padding
```

design/StableMemory.md
```diff
@@ -249,19 +259,40 @@ fun restore() : value option =
...
+FURTHER CONSIDERATIONS: It would be nice if there was some way to
+allow a Motoko actor (perhaps intended to upgrade from a foreign
+canister with foreign stable memory format) to have unadulterated,
+full-speed access to stable memory, without the protocol required by
+the possibility of stable variables.  Perhaps we could special case
+programs that have *no* stable variables to support this raw
+semantics.  I.e. if the program declares no stable variables, we
+install and upgrade by simply setting `(StableMemory.)size :=
+ic0.stable_size()`, never consulting or altering physical memory and
+(ideally) omitting the additional bounds checking.
```
rossberg, line 298: Yeah, I was thinking about this as well. But it seems like we always have to take into account the possibility that there were stable variables before the upgrade, written by the prior code. Or that at some point in time there was an even older version that grew the physical size of stable memory past (the current) StableMemory.size. Dammit.

Something else I wondered was whether we could somehow make it so that StableMem is always located at the _end_ of stable memory -- then instead of custom bounds check we'd only need to add an offset, which should be cheaper. But that would require shifting all of StableMem each time stable vars are backed up whose value takes more space than what was previously reserved. That would be pretty costly. OTOH, that cost would never come up if you didn't use both kinds of stable data at the same time. But still probably not worth it.

rossberg (approved):

design/StableMemory.md
crusso: > Yeah, I was thinking about this as well. But it seems like we always have to take into account the possibility that there were stable variables before the upgrade, written by the prior code.

But can't we just ignore them, and let the (sv-less) code decide what to do with the populated memory?

> Or that at some point in time there was an even older version that grew the physical size of stable memory past (the current) StableMemory.size. Dammit.

Is that a problem? At that point we just set StableMemory.size to ic0.stable_size() and the actor can see the full memory and do as they see fit? If they want to keep a finger on StableMemory, they can use a dedicated address and upgrade hook to save it. If we wanted to restore it,  we'd need to extend the interface.
I think you could even use this technique to migrate code from pure stable variables to pure StableMemory, provided we give people access to Candid(ish) deserialization.

> Something else I wondered was whether we could somehow make it so that StableMem is always located at the _end_ of stable memory -- then instead of custom bounds check we'd only need to add an offset, which should be cheaper. But that would require shifting all of StableMem each time stable vars are backed up whose value takes more space than what was previously reserved. That would be pretty costly. OTOH, that cost would never come up if you didn't use both kinds of stable data at the same time. But still probably not worth it.

rossberg: It would be fishy to arbitrarily change StableMem.size behind the app's back. It might have certain assumptions attached to the size it knows, i.e., that the page number is a multiple of something or that each page is initialised in a certain way.

So I think you can perhaps simplify some logic if they _are_ already equal. But you cannot _make_ them equal.

crusso:
@rossberg should I still go ahead an implement this or wait until 64 bit stable memory is supported and just keep the two separate?

rossberg:
I'd say go ahead. I don't assume it would be hard to change it to 64 bit later?

crusso:
@rossberg, so I've almost completed the (de)stabilization bits, but was wondering about the API. Do you want to index/ report size of memory with Nat32 or Nat? Nat is actually more compact (unboxed for small values) but will require a trapping conversion.

I'll start with a very small API (maybe just byte and word level writes) so I can write sanity tests.

design/StableMemory.md
```diff
@@ -0,0 +1,290 @@
...
+   assert (offset < StableMemory.size() * wasm_page_size);
+   mem[offset]
+
+fun storeNat8(offset, b) =
+   assert (offset < StableMemory.size() * wasm_page_size);
+   mem[offset] := b
+
+```
+
+(To avoid overflow on the rhs, we could implement the check as `assert ((offset >> 6) < StableMemory.size())`.)
```
rossberg, line 78: That probably would have to be
```suggestion
(To avoid overflow on the rhs, we could implement the check as `assert (((offset + 0xffff) >> 6) < StableMemory.size())`.)
```

design/StableMemory.md
```diff
@@ -0,0 +1,290 @@
...
+
+Actually implementing the sketched assignments using the existing IC
+API involves writing the contents to memory and then copying stable
+memory - even for individual words - this could be optimized by an
+improved system API offering direct load and stores from/to the stack:
+
+```
+ic0.stable_write_i32 : (offset : i32, val: i32) -> ();   // *
+ic0.stable_read_i32 : (offset : i32, size : i32) -> i32; // *
+// similarly for i64, f32, f64
```
rossberg, line 93: You'll likely also want i8 and i16 versions (that extend to i32 values, either signed or unsigned). Essentially, you want the analogous of all of Wasm's load/store instructions.

design/StableMemory.md
```diff
@@ -0,0 +1,290 @@
...
+
+During execution, abstract stable memory (StableMemory) is aligned
+with IC stable memory, at address 0, for reasonable efficiency (apart
+from bound checks against logical `size()`).
+
+During upgrade, if StableMemory has zero pages, we use the existing format, writing
+(non_zero) length and content of any stable variables from address 0 or leaving ic0.stable_mem()
+at zero with no pages allocated (if there are no stable variables).
+Otherwise, we compute the length and data of the stable variable encoding;
+save the first word of StableMemory at a known offset from the end of physical memory;
```
rossberg, line 155:
```suggestion
save the first word of StableMemory at a known offset from the end of stable memory;
```

design/StableMemory.md
```diff
@@ -0,0 +1,290 @@
...
+        mem[N,..,N+3] := 0;
+        assert len > 0
+        assert (N+4+len-1 <= ic0.stable_size() * pagesize)
+        (N+4, len)
+      else
+        (4, marker)
+    in
+    assert (0 < len <= ic0.stable_size() * pagesize)
+    let v = deserialise<Ts>(offset, len) in
+    mem[offset, len] := 0 // clear serialization memory
```
rossberg, line 257:
```suggestion
    mem[offset,..,offset+len-1] := 0 // clear serialization memory
```

design/StableMemory.md
```diff
@@ -0,0 +1,290 @@
...
+
+NOTE: We still need to do some work during updgrade and postupgrade,
+but if stable variables and user-defined pre/post upgrade hooks are
+avoided, then the work is minimal and highly unlikely to exhaust cycle
+budget.
+
+REMARK:
+
+* An actor that has no stable variables and allocates no StableMem
+  should requize no physical stable memory
```
rossberg, line 274:
```suggestion
  should require no physical stable memory
```

src/codegen/compile.ml
```diff
@@ -2926,6 +2941,21 @@ module Blob = struct
...
+      let (set_len, get_len) = new_local env "len" in
+      get_x ^^
+      as_ptr_len env ^^
+      set_len ^^
+      set_ptr ^^
+      get_len ^^
+      from_0_to_n env (fun get_i ->
+        get_ptr ^^ get_i ^^ G.i (Binary (Wasm.Values.I32 I32Op.Add)) ^^
+        compile_unboxed_const 0l ^^
+        G.i (Store {ty = I32Type; align = 0; offset = 0l; sz = Some Wasm.Types.Pack8})))
```
rossberg, line 2956: Would it be hard to optimise this to write i64 zeroes, plus a couple of odd extra bytes in the end if necessary?

Even better, can't we use memory.fill? That's standard now, and I think Wasmtime supports it.

src/codegen/compile.ml
```diff
@@ -3697,6 +3727,148 @@ module IC = struct
...
+  (* stable memory bounds check *)
+  let _guard env =
+    match E.mode env with
+    | Flags.ICMode | Flags.RefMode ->
+      Func.share_code1 env "__stablemem_guard"
+        ("offset", I32Type) []
+        (fun env get_offset ->
+          get_offset ^^
+          compile_unboxed_const (Int32.of_int page_size_bits) ^^
+          G.i (Binary (Wasm.Values.I32 I32Op.ShrU)) ^^
```
rossberg, line 3752: See above, I believe this isn't quite correct. You'll need to add 0xffff before the shift to "round up".

src/codegen/compile.ml
```diff
@@ -3697,6 +3727,148 @@ module IC = struct
...
+    | Flags.ICMode | Flags.RefMode ->
+      Func.share_code2 env "__stablemem_guard"
+        (("offset", I32Type), ("size", I32Type)) []
+        (fun env get_offset get_size ->
+          get_offset ^^ G.i (Convert (Wasm.Values.I64 I64Op.ExtendUI32)) ^^
+          get_size ^^ G.i (Convert (Wasm.Values.I64 I64Op.ExtendUI32)) ^^
+          G.i (Binary (Wasm.Values.I64 I64Op.Add)) ^^
+          get_mem_size env ^^
+          compile_const_64 (Int64.of_int page_size_bits) ^^
+          G.i (Binary (Wasm.Values.I64 I64Op.Shl)) ^^
```
rossberg, line 3770: Is there a reason why you're not using the right shift here?

src/codegen/compile.ml
```diff
@@ -3697,6 +3727,148 @@ module IC = struct
...
+    match E.mode env with
+    | Flags.ICMode | Flags.RefMode ->
+      Func.share_code2 env "__stablemem_ensure"
+        (("offset", I32Type), ("size", I32Type)) []
+        (fun env get_offset get_size ->
+          get_offset ^^ G.i (Convert (Wasm.Values.I64 I64Op.ExtendUI32)) ^^
+          get_size ^^ G.i (Convert (Wasm.Values.I64 I64Op.ExtendUI32)) ^^
+          G.i (Binary (Wasm.Values.I64 I64Op.Add)) ^^
+          compile_const_64 (Int64.of_int page_size_bits) ^^
+          G.i (Binary (Wasm.Values.I64 I64Op.ShrU)) ^^
```
rossberg, line 3864: Same here, I believe.

src/codegen/compile.ml
```diff
@@ -3697,6 +3727,148 @@ module IC = struct
...
+          G.i (Convert (Wasm.Values.I32 I64Op.WrapI64)) ^^ (* TBR *)
+          compile_add_const 1l ^^
```
rossberg, line 3866: Why add 1?

design/StableMemory.md
crusso: I'm not seeing this. Are you sure?

design/StableMemory.md
```diff
@@ -0,0 +1,290 @@
...
+  storeNat8 : (offset : Nat32, n : Nat8) -> ();
+    // traps outside logical address space
+  ...
+  loadBlob : (offset : Nat32, size : i32) -> Blob
+     // read Blob contents from memory at [offset,..,offset+size-1] into fresh blob, trapping if exceeding logical address space
+  storeBlob : (offset : Nat32, b : Blob) -> (); // write contents of blob to memory, trapping if exceeding logical address space
+}
+```
+
+NOTE: Motoko's `Nat32` value are always boxed - it might be more efficient to use `Nat` which is unboxed for 30(?)-bit values.
```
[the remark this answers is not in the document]
crusso: Yeah, but my thinking was we index arrays by Nat and might generalize to 64 bit later.
Nat32 has some clarity to it though.

design/StableMemory.md
crusso: But I do think that should be 16, not 6, which maybe a typo.

crusso: My reasoning is check the pageindex of the address (>>16) is strictly less than the page count.

src/codegen/compile.ml
crusso: grow takes the desired page *count* and really grows if necessary. But maybe I'm wrong. Will double check.

src/codegen/compile.ml
crusso: Yeah, I was actually planning to use i32 but either of those would better.

src/codegen/compile.ml
crusso: I think there is an off by one earlier, since we want the address offset+size-1 to be accessible, not offset+size.

rossberg: The add 1 on this line adds an entire page, right? So this seems to imply that `ensure` always grows by at least one page. Should it really?

src/codegen/compile.ml
crusso: The replica's appear to support MemoryFill, but our wasm lib is to old. We'll need to upgrade that first. I'll leave that to a separate PR.

src/codegen/compile.ml
crusso: It seems more complicated actually (see version using Shr in comments, but perhaps you had something better in mind)?

src/codegen/compile.ml
crusso: See above

src/codegen/compile.ml
crusso: ensure only grows if necessary, so it shouldn't, no. 
Also,the code assumes size > 0, so at least one page will be necessary.

crusso:
@rossberg PTAL before I implement more of the API

src/codegen/compile.ml
```diff
@@ -3865,11 +3880,89 @@ module StableMem = struct
...
+          get_mem_size env ^^
+          set_size ^^
+
+          get_size ^^ G.i (Convert (Wasm.Values.I64 I64Op.ExtendUI32)) ^^
+          get_pages ^^ G.i (Convert (Wasm.Values.I64 I64Op.ExtendUI32)) ^^
+          G.i (Binary (Wasm.Values.I64 I32Op.Add)) ^^
+          compile_const_64 65536L ^^
+          G.i (Compare (Wasm.Values.I64 I64Op.LtU)) ^^
+          G.if_ []
+            (G.nop) (* fallthrough *)
```
rossberg, line 3909: Nit: If you invert the condition (`GeU`) then you can use an else-less if and drop the nop.

test/run-drun/stable-mem-grow.mo
```diff
@@ -0,0 +1,55 @@
...
+      i += 4
+    };
+
+    n += 1;
+
+  };
+
+  public func testGrow() : async () {
+    var i : Nat32 = 0;
+    while (i < 10) {
```
rossberg, line 26: Don't you like for loops? :)

test/run-drun/stable-mem.mo
```diff
@@ -0,0 +1,89 @@
...
+
+    if (n == 0) return;
+    assert (n == StableMemory.size());
+    P.debugPrint (debug_show {testBounds=n});
+    // test bounds check
+    var i : Nat32 = n * 65536 - 3;
+    let max = i + 16;
+    while (i < max) {
+      try {
+        await async {
```
rossberg, line 54: Is this so that you can use exception handling?

test/run-drun/stable-mem.mo
```diff
@@ -0,0 +1,89 @@
...
+          ignore StableMemory.loadNat32(i);
```
rossberg, line 55: Assert false after this, like below?

rossberg (approved):
LGTM, but I'm still not very effective at reading backend code, so might have missed something. Somebody else might want to provide another pair of eyes on that.

src/codegen/compile.ml
crusso: Wasn't aware we have that!

test/run-drun/stable-mem-grow.mo
crusso: No base in tests, so I'd need to implement an iterator first. So, no.

test/run-drun/stable-mem.mo
crusso: It's so I can observe the trap.

test/run-drun/stable-mem.mo
crusso: K

src/codegen/compile.ml
nomeata: I don’t think our `G.` module provides else-less ifs. But the wasm encoder might be smart here.

design/StableMemory.md
```diff
@@ -0,0 +1,290 @@
...
+```
+module StableMemory {
+  size : () -> (logical_page_count : i32); // <= ic0.stable_size()
+  grow : () -> (new_pages : i32) -> (old_logical_page_count : i32);
+  loadNat8 : (offset : Nat32) -> Nat8;
+    // traps outside logical address space
+  storeNat8 : (offset : Nat32, n : Nat8) -> ();
+    // traps outside logical address space
+  ...
+  loadBlob : (offset : Nat32, size : i32) -> Blob
```
crusso, line 58:
```suggestion
  loadBlob : (offset : Nat32, size : Nat) -> Blob
```

ggreif:
> * We are missing load/store of Nat/Int/Char/Bool/Principal/Text. Do we want them? Can always add later or hand roll

As you have `Blob` load/store, going to `Text` and `Principal` should be trivial (check: do we have primitives for those?). `Bool` is just comparison against `0`. So what remains is `Nat/Int/Char`. `Char` can be done with `charToNat32`/`nat32ToChar`. So we only need `intToBlob` and `blobToInt`/`blobToNat`.

crusso:
> > * We are missing load/store of Nat/Int/Char/Bool/Principal/Text. Do we want them? Can always add later or hand roll
> 
> As you have `Blob` load/store, going to `Text` and `Principal` should be trivial (check: do we have primitives for those?). `Bool` is just comparison against `0`. So what remains is `Nat/Int/Char`. `Char` can be done with `charToNat32`/`nat32ToChar`. So we only need `intToBlob` and `blobToInt`/`blobToNat`.

So we have Text.encodeUft8/Text.decodeUtf8, Principal.toBlob/toText/fromText (but not fromBlob), and I guess users could write their own Int and Nat conversions (at cost), or rely on the upcoming Candid serialization prims for those.


Is there an invariant on the structure of Principals that prevents us from safely adding Principal.fromBlob (to avoid the detour through text?).

design/StableMemory.md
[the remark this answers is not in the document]
ggreif: Please note: `Nat32` is only allocated on the heap if it has more than 31 bits.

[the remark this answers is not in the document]
crusso: > Please note: `Nat32` is only allocated on the heap if it has more than 31 bits.

Is that actually true? I thought we only did that optimization for Nat?

test/run-drun/stable-mem.mo
crusso: done (in later PR)

src/codegen/compile.ml
crusso: Rewritten as full if, inverting test, and pushing continuation into else branch.

design/StableMemory.md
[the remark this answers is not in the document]
ggreif: Nope, I  double-checked. `Int/Nat32/64` only go to heap when necessary, see:
``` ocaml
module BoxedSmallWord = struct
  (* We store proper 32bit Word32 in immutable boxed 32bit heap objects.

     Small values are stored unboxed, tagged, see BitTagged.

     The heap layout of a BoxedSmallWord is:

       ┌─────┬─────┐
       │ tag │ i32 │
       └─────┴─────┘

  *)
```
`BoxedSmallWord.(un)box` does all the magic for us.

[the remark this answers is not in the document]
ggreif: But I filed #2774 to fend off a potential pessimization.

PR #2630, opened 2021-06-28
Fix sleb128_decode overflow checking

### LINKED ISSUE #2625: RTS: sleb128_decode overflow check is not right
It rejects valid values, for example, `-1100000` is encoded as

```
[160, 238, 188, 127]
[a0, ee, bc, 7f] // or in hex
[10100000, 11101110, 10111100, 01111111] // or in binary
```

but sleb128_decode rejects this encoding as "overflow". The function: https://github.com/dfinity/motoko/blob/01867bdb76d6ceb7e986297464a41146f3d34839/rts/motoko-rts/src/leb128.rs#L63-L90

First of all, incrementing shift and then checking the bits makes no sense, we should increment shift as last thing in the loop.

But even after fixing that this check still rejects `-2147483648` which is encoded as

```
[128, 128, 128, 128, 120]
[80, 80, 80, 80, 78]
[10000000, 10000000, 10000000, 10000000, 01111000]
```

Too tired to fix this today so just reporting it for now..

osa1:
Fixes #2625

Refactors decode functions to allow checking overflows.

New tests added for checking edge cases + overflows.

### DIFF
--- a/rts/motoko-rts-tests/src/leb128.rs
+++ b/rts/motoko-rts-tests/src/leb128.rs
@@ -1,12 +1,47 @@
 use motoko_rts::buf::Buf;
-use motoko_rts::leb128::{leb128_decode, leb128_encode, sleb128_decode, sleb128_encode};
+use motoko_rts::leb128::{
+    leb128_decode, leb128_decode_checked, leb128_encode, sleb128_decode, sleb128_decode_checked,
+    sleb128_encode,
+};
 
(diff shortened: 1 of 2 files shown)


crusso (commented):
LGTM.

Are the option returning versions used on hot paths? Would it be worth writing ones that either panic on overflow or return boolean out parameter indicating success instead of allocating an option?

crusso (approved):

PR #2632, opened 2021-06-29
RTS: Introduce WASM_PAGES_SIZE and WASM_HEAP_SIZE constants

osa1:


rts/motoko-rts/src/alloc.rs
```diff
@@ -11,14 +11,14 @@ pub use alloc_impl::alloc_words;
...
+use crate::constants::WASM_HEAP_SIZE;
 use crate::rts_trap_with;
 use crate::types::{size_of, Array, Blob, Bytes, SkewedPtr, Words, TAG_ARRAY, TAG_BLOB};
 
 #[no_mangle]
 pub unsafe extern "C" fn alloc_array(len: u32) -> SkewedPtr {
     // Array payload should not be larger than half of the memory
-    if len > 1 << (32 - 2 - 1) {
-        // 2 for word size, 1 to divide by two
+    if Words(len) > WASM_HEAP_SIZE / 2 {
```
ulan, line 21: Nit (feel free to ignore): how about `WASM_HEAP_SIZE_IN_WORDS` to make it clear for the reader?

rts/motoko-rts/src/constants.rs
```diff
@@ -0,0 +1,11 @@
+use crate::types::{Bytes, Words};
+
+/// Wasm word size. RTS only works correctly on platforms with this word size.
+pub const WORD_SIZE: u32 = 4;
+
+/// Wasm page size (64KiB) in bytes
+pub const WASM_PAGE_SIZE: Bytes<u32> = Bytes(64 * 1024);
+
+/// Wasm heap size (4GiB) in words. Note that `to_bytes` on this value will overflow as 4GiB in
+/// bytes is `u32::MAX + 1`.
```
ulan, line 10: The overflow is a bit of a footgun.  Alternatively we could introduce `MAX_HEAP_SIZE` and make it smaller than `WASM_HEAP_SIZE` by a couple of pages, so that we don't have to worry about accidentally overflowing u32.

ulan (approved):
lgtm, thanks!

Feel free to ignore the suggestions below.

rts/motoko-rts/src/alloc.rs
osa1: I used the type `Words`, which is a newtype and requires explicit unwrapping or conversion to `Bytes`. Do you still think `_IN_WORDS` suffix would be helpful? I can add it, just wanted to make sure you know that `Words` is a newtype and won't be implicitly converted to anything else and when used in arithmetic with `Bytes` or scalars will cause compile errors.

rts/motoko-rts/src/alloc.rs
```diff
@@ -11,14 +11,14 @@ pub use alloc_impl::alloc_words;
 #[cfg(feature = "gc")]
 pub(crate) use alloc_impl::grow_memory;
 
+use crate::constants::WASM_HEAP_SIZE;
 use crate::rts_trap_with;
 use crate::types::{size_of, Array, Blob, Bytes, SkewedPtr, Words, TAG_ARRAY, TAG_BLOB};
 
 #[no_mangle]
 pub unsafe extern "C" fn alloc_array(len: u32) -> SkewedPtr {
     // Array payload should not be larger than half of the memory
```
osa1, line 20: I wish we documented why this is the case. I believe this is because the copying GC will want to copy it and overflow the heap if this is 2GiB, but I don't know if there are other reasons.

rts/motoko-rts/src/constants.rs
osa1: I'm not sure if we want to reduce heap size to avoid a bug in the RTS. In Rust overflow in arithmetic is checked in debug mode, and we run the tests in both debug mode and release mode, so hopefully any bugs caused by this will be caught before getting merged.

crusso (approved):

PR #2633, opened 2021-06-30
RTS: fix pages calculation in grow_memory

osa1:
Originally discussed here: https://github.com/dfinity/motoko/pull/2573#discussion_r660366802

Previously grow_memory would allocate one more page than necessary when
the argument is a multiple of Wasm page size. New calculation fixes
that. The code is basically

    (ptr + WASM_PAGE_SIZE - 1) / WASM_PAGE_SIZE

I.e. divide argument by WASM_PAGE_SIZE, round up. However we do this in
u64 to avoid overflows in `ptr` is larger than `u32::MAX -
WASM_PAGE_SIZE + 1`

Also moves `total_pages - current_pages` to the slow path to improve
performance slightly.

CanCan backend benchmark reports 0.002% increase in cycles.

### DIFF
--- a/rts/motoko-rts/src/alloc/gc.rs
+++ b/rts/motoko-rts/src/alloc/gc.rs
@@ -24,7 +24,7 @@ pub unsafe extern "C" fn alloc_words(n: Words<u32>) -> SkewedPtr {
     skew(old_hp as usize)
 }
 
-/// Page allocation. Ensures that the memory up to the given pointer is allocated.
+/// Page allocation. Ensures that the memory up to, but excluding, the given pointer is allocated.
 pub(crate) unsafe fn grow_memory(ptr: usize) {
     let page_size = u64::from(WASM_PAGE_SIZE.0);
     let total_pages_needed = (((ptr as u64) + page_size - 1) / page_size) as usize;
--- a/rts/motoko-rts/src/alloc/gc.rs
+++ b/rts/motoko-rts/src/alloc/gc.rs
@@ -26,7 +26,8 @@ pub unsafe extern "C" fn alloc_words(n: Words<u32>) -> SkewedPtr {
 
 /// Page allocation. Ensures that the memory up to the given pointer is allocated.
 pub(crate) unsafe fn grow_memory(ptr: usize) {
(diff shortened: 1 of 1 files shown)


osa1:
I managed to improve perf of new code a little bit, pushing the commit now.

crusso (commented):
Is that correct? What if current pages = 0 and ptr  = 0, then this allocates nothing  and dereferencing ptr traps. Or is my arithmetic wrong. On mobile.

osa1:
Good point, `ptr = 0` is another case where the new function will return differently than the current one, but that case cannot happen. This function is called to make sure we have enough Wasm pages, so for example when we allocate `N` words we do `grow_memory(HP + N)`. For `HP + N` to be 0 both `HP` and `N` need to be 0. `HP` can't be zero because we will always have some static data in heap and `HP` will start with a larger value. `N` can't be zero because I think we never try to allocate 0 words.

crusso:
Doesn't the problem recur at every page boundary. Suppose HP = PAGESIZE-1 and N = 1. Then we need to allocate another page but the algorithm won't, or will it?

crusso:
(I'm assuming ptr is meant to be useable (in bounds) after the call.)

osa1:
> (I'm assuming ptr is meant to be useable (in bounds) after the call.)

`ptr` is not guaranteed to be allocated after this call, `ptr - 1` is.

crusso:
Ah, maybe adjust the comment to say ptr (exclusive)

rts/motoko-rts/src/alloc/gc.rs
```diff
@@ -26,11 +26,11 @@ pub unsafe extern "C" fn alloc_words(n: Words<u32>) -> SkewedPtr {
 
 /// Page allocation. Ensures that the memory up to the given pointer is allocated.
```
crusso, line 27:
```suggestion
/// Page allocation. Ensures that the memory up to, but excluding, the given pointer is allocated.
```

crusso:
That makes sense now, HP points at the next available byte (not last allocated), right?

crusso (approved):

osa1:
> That makes sense now, HP points at the next available byte (not last allocated), right?

Exactly, yeah.

PR #2641, opened 2021-07-01
Compacting GC: merge threading and marking

osa1:
This PR merges threading of object fields with marking. The result is
-14% in instructions in CanCan backend benchmark.

Changes:

- Mark phase of compacting GC now threads fields of marked objects.
  After marking all fields are now threaded.

- Because all fields are threaded during marking, the pass for threading
  is now gone.

- However we still need two passes for unthreading and moving the
  objects (see below). So the first pass after marking unthreads fields.
  Then another pass moves objects.

Original idea was to merge marking and threading, and doing unthreading +
moving in one pass. It turns out that doesn't work. To see why, suppose the
heap is like this:

    | ..., object A, ..., object B, ... |

where A and B point to each other. If we unthread + move (slide) in one pass,
then after sliding A, it's possible that live objects between A and B will be
slid to A's original location. So when we unthread B which will have a pointer
to A's old location in its header we'll read random data from whatever's moved
to A's original location.

The solution is to do these in different passes: first unthread, then slide.

The result is still 3 passes as before, but for some reason that I don't
understand, this version is 14% faster on CanCan backend benchmark (42,005,245
instructions to 35,775,591).

I think "An Efficient Garbage Compaction Algorithm" is describing a way to
implement in-sliding compaction in two passes instead of three, but I found
that paper difficult to follow. I will revisit it now to see if this can be
improved further.

### DIFF
--- a/rts/motoko-rts-tests/src/gc.rs
+++ b/rts/motoko-rts-tests/src/gc.rs
@@ -21,23 +21,49 @@ use std::fmt::Write;
 pub fn test() {
     println!("Testing garbage collection ...");
 
-    // TODO: Add more tests
-
     for test_heap in test_heaps() {
         test_gcs(&test_heap);
     }
 }
 
 fn test_heaps() -> Vec<TestHeap> {
-    vec![TestHeap {
-        heap: hashmap! {
-            0 => vec![0, 2],
-            2 => vec![0],
-            3 => vec![3],
+    vec![
+        // Just a random test that covers a bunch of cases:
+        // - Self references
+        // - Forward pointers
+        // - Backwards pointers
+        // - More than one fields in an object
+        TestHeap {
+            heap: hashmap! {
+                0 => vec![0, 2],
+                2 => vec![0],
+                3 => vec![3],
+            },
+            roots: vec![0, 2, 3],
+            closure_table: vec![0],
         },
-        roots: vec![0, 2, 3],
-        closure_table: vec![0],
-    }]
+        // Tests pointing to the same object in multiple fields of an object. Also has unreachable
+        // objects.
+        TestHeap {
+            heap: hashmap! {
+                0 => vec![],
+                1 => vec![],
+                2 => vec![],
+            },
+            roots: vec![1],
+            closure_table: vec![0, 0],
+        },
+        // Root points backwards in heap. Caught a bug in mark-compact collector.
+        TestHeap {
+            heap: hashmap! {
+                0 => vec![],
+                1 => vec![2],
+                2 => vec![1],
+            },
+            roots: vec![2],
+            closure_table: vec![],
+        },
+    ]
 }
 
 #[derive(Debug)]
@@ -66,17 +92,6 @@ fn test_gc(
     closure_table: &[ObjectIdx],
 ) {
     let heap = MotokoHeap::new(refs, roots, closure_table, gc);
-
-    // Check `check_dynamic_heap` sanity
-    check_dynamic_heap(
-        refs,
-        roots,
-        closure_table,
-        &**heap.heap(),
-        heap.heap_base_offset(),
-        heap.heap_ptr_offset(),
-        heap.closure_table_ptr_offset(),
-    );
 
     for _ in 0..3 {
         gc.run(heap.clone());
--- a/rts/motoko-rts-tests/src/mark_stack.rs
+++ b/rts/motoko-rts-tests/src/mark_stack.rs
@@ -32,12 +32,12 @@ fn test_<M: Memory>(mem: &mut M, n_objs: u32) -> TestCaseResult {
         alloc_mark_stack(mem);
 
         for obj in &objs {
-            push_mark_stack(mem, *obj as usize);
+            push_mark_stack(mem, *obj as usize, obj.wrapping_sub(1));
         }
 
-        for obj in objs.iter().rev() {
+        for obj in objs.iter().copied().rev() {
             let popped = pop_mark_stack();
-            if popped != Some(*obj as usize) {
+            if popped != Some((obj as usize, obj.wrapping_sub(1))) {
                 free_mark_stack();
                 return Err(TestCaseError::Fail(
                     format!(
--- a/rts/motoko-rts/src/gc/mark_compact.rs
+++ b/rts/motoko-rts/src/gc/mark_compact.rs
@@ -157,7 +157,7 @@ unsafe fn update_refs<SetHp: Fn(u32)>(set_hp: SetHp, heap_base: u32) {
         // Update references to the object's new location and restore object header
         unthread(p, free);
 
-        // All references to the object now point to the new location, move the object
+        // All references to the object now point to the new location (object only moved in next pass)
         let p_size_words = object_size(p as usize);
 
         free += p_size_words.to_bytes().0;
--- a/rts/motoko-rts/src/gc/mark_compact.rs
+++ b/rts/motoko-rts/src/gc/mark_compact.rs
@@ -90,8 +90,6 @@ unsafe fn mark_compact<M: Memory, SetHp: Fn(u32)>(
 
     mark_stack(mem, heap_base);
 
-    thread_roots(static_roots, heap_base);
-
     update_refs(set_hp, heap_base);
 
     free_mark_stack();
@@ -136,15 +134,6 @@ unsafe fn mark_fields<M: Memory>(mem: &mut M, obj: *mut Obj, obj_tag: Tag, heap_
     });
 }
 
-unsafe fn thread_roots(static_roots: SkewedPtr, heap_base: u32) {
-    // Static roots
-    let root_array = static_roots.as_array();
-    for i in 0..root_array.len() {
-        thread_obj_fields(root_array.get(i).unskew() as *mut Obj, heap_base);
-    }
-    // No need to thread closure table here as it's on heap and we already marked it
-}
-
 /// Expects all fields to be threaded. First pass updates fields, second pass moves objects.
 unsafe fn update_refs<SetHp: Fn(u32)>(set_hp: SetHp, heap_base: u32) {
     let mut free = heap_base;
@@ -182,12 +171,6 @@ unsafe fn update_refs<SetHp: Fn(u32)>(set_hp: SetHp, heap_base: u32) {
     }
 
     set_hp(free);
-}
-
-unsafe fn thread_obj_fields(obj: *mut Obj, heap_base: u32) {
-    visit_pointer_fields(obj, obj.tag(), heap_base as usize, |field_addr| {
-        thread(field_addr)
-    });
 }
 
 unsafe fn thread(field: *mut SkewedPtr) {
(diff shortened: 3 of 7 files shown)


rts/motoko-rts-tests/src/gc.rs
```diff
@@ -67,17 +93,6 @@ fn test_gc(
...
-    // Check `check_dynamic_heap` sanity
-    check_dynamic_heap(
-        refs,
-        roots,
-        closure_table,
-        &**heap.heap(),
-        heap.heap_base_offset(),
-        heap.heap_ptr_offset(),
-        heap.closure_table_ptr_offset(),
-    );
```
osa1, line 79: I removed this as it doesn't allow garbage in heap (as we don't want garbage after a GC). I should probably modify this to take a "post GC" parameter and check for garbage depending on that.

rts/motoko-rts-tests/src/mark_stack.rs
```diff
@@ -32,12 +32,12 @@ fn test_<M: Memory>(mem: &mut M, n_objs: u32) -> TestCaseResult {
         alloc_mark_stack(mem);
 
         for obj in &objs {
-            push_mark_stack(mem, *obj as usize);
+            push_mark_stack(mem, *obj as usize, obj.wrapping_sub(1));
```
crusso, line 35: is that an unskew?

rts/motoko-rts-tests/src/mark_stack.rs
```diff
@@ -32,12 +32,12 @@ fn test_<M: Memory>(mem: &mut M, n_objs: u32) -> TestCaseResult {
...
         }
 
-        for obj in objs.iter().rev() {
+        for obj in objs.iter().copied().rev() {
             let popped = pop_mark_stack();
-            if popped != Some(*obj as usize) {
+            if popped != Some((obj as usize, obj.wrapping_sub(1))) {
```
crusso, line 40: ditto

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -108,11 +104,12 @@ unsafe fn mark_static_roots<M: Memory>(mem: &mut M, static_roots: SkewedPtr, hea
     // Static objects are not in the dynamic heap so don't need marking.
     for i in 0..root_array.len() {
         let obj = root_array.get(i).unskew() as *mut Obj;
-        mark_fields(mem, obj, heap_base);
+        mark_fields(mem, obj, obj.tag(), heap_base);
     }
 }
 
 unsafe fn push_mark_stack<M: Memory>(mem: &mut M, obj: SkewedPtr, heap_base: u32) {
```
crusso, line 111: maybe mark_object would be a better name?

crusso (changes requested):

osa1:
Thanks @crusso for the review, I'll address your comments next week.

> The result is still 3 passes as before, but for some reason that I don't understand, this version is 14% faster

I was thinking about this and I think I have an idea. Previously the algorithm was something like this at a high level:

- Mark live objects
- Visit live objects from start to end, unthread and update forward pointers to the object's new location, thread fields
- Visit live objects from start to end, unthread and update backward pointers to the object's new location, move the object

With this PR:

- Mark live objects, thread fields
- Visit live objects from start to end, unthread objects and update both forward and backward pointers to the object's new location
- Visit live objects from start to end, move objects

I think the difference in performance is because with this PR we visit object fields once (to push fields to the mark stack and thread), instead of twice as before (once to push fields to the mark stack, once again to thread). This can be seen in the diff if you search for `thread_obj_fields`: the use in `update_fwd_refs` is now gone.

---

As mentioned in the PR description, "An Efficient Garbage Compaction Algorithm" describes a way to do this in two passes, but I found the paper difficult to follow. @crusso mentioned "High-Performance Garbage Collection for Memory-Constrained Environments" to me (thanks @crusso!) which describes a compaction algorithm that works in two passes. I don't know if it's the same as "An Efficient Garbage Collection Algorithm" or not, but it's quite simple and I think should improve our GC even more. Here's how it works: during marking we only thread backward pointers and skip forward pointers. So in our original example from PR description:

```
| ..., object A, ..., object B, ... |
```

during marking we only thread the field of B that points to A, not the field of A that points to B.

Now the second pass is similar, we scan the heap from start to end. When we see A we unthread it and update the pointers to A's final location, then move A, then thread its forward pointers. Now when we reach B we can safely unthread it as its header will be readable (i.e. won't be overwritten by objects between A and B).

It's quite smart and should improve this PR even more. I will implement this next week.

rts/motoko-rts/src/gc/mark_compact/mark_stack.rs
```diff
@@ -50,20 +50,25 @@ unsafe fn grow_stack<M: Memory>(mem: &mut M) {
     (*STACK_BLOB_PTR).len = new_cap.to_bytes();
 }
 
-pub unsafe fn push_mark_stack<M: Memory>(mem: &mut M, obj: usize) {
+pub unsafe fn push_mark_stack<M: Memory>(mem: &mut M, obj: usize, obj_tag: Tag) {
+    // We add 2 words in a push, and `STACK_PTR` and `STACK_TOP` are both multiples of 2, so we can
```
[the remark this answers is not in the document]
osa1: > Is the size of the mark stack ever an issue?

I had some benchmarks on mark stack sizes when I first implemented this GC, but those benchmarks were not realistic (more like micro benchmarks) so I should run that again on CanCan. In the worst case stack size will be `heap size / 2` (one object points to every object in the heap), but that case never happens.

rts/motoko-rts-tests/src/mark_stack.rs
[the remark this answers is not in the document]
osa1: So this part is a bit hacky. Now that push needs a pointer + tag (u32) I needed to come up with something for the tag here. The mark stack doesn't care about the tag so just randomly used `value - 1`. I could use a constant, but I thought perhaps that hides bugs.

> Is Tag just an abbreviation for usize? If it is, might be nice to make them distinct types.

Yeah I think making them distinct makes sense. I'll do that in a separate PR.

osa1: Not really, I just wanted a value derived from `obj` for the `tag` argument of `push_mark_stack`. Since the actual value does not matter (push/pop does not care about validness of tags) I more or less randomly used `obj - 1`.

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -123,18 +120,19 @@ unsafe fn push_mark_stack<M: Memory>(mem: &mut M, obj: SkewedPtr, heap_base: u32
...
+    while let Some((obj, tag)) = pop_mark_stack() {
+        mark_fields(mem, obj as *mut Obj, tag, heap_base);
     }
 }
 
-unsafe fn mark_fields<M: Memory>(mem: &mut M, obj: *mut Obj, heap_base: u32) {
-    visit_pointer_fields(obj, heap_base as usize, |field_addr| {
+unsafe fn mark_fields<M: Memory>(mem: &mut M, obj: *mut Obj, obj_tag: Tag, heap_base: u32) {
+    visit_pointer_fields(obj, obj_tag, heap_base as usize, |field_addr| {
         push_mark_stack(mem, *field_addr, heap_base);
```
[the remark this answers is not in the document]
osa1: Yeah having two `push_mark_stack`s is a bit confusing. I'll rename.

osa1:
> should improve this PR even more

Actually I'm not sure about this, because we will have to visit object fields twice and the -14% will be lost. I will implement it in a new PR so that we can compare.

rts/motoko-rts-tests/src/mark_stack.rs
crusso: Oh, never mind, I didn't realize this was just dummy data.

crusso: (should have looked at the context)

rts/motoko-rts/src/gc/mark_compact/mark_stack.rs
```diff
@@ -50,20 +50,25 @@ unsafe fn grow_stack<M: Memory>(mem: &mut M) {
...
     }
 
     *STACK_PTR = obj;
-    STACK_PTR = STACK_PTR.add(1);
+    *(STACK_PTR.add(1)) = obj_tag as usize;
+    STACK_PTR = STACK_PTR.add(2);
 }
 
-pub unsafe fn pop_mark_stack() -> Option<usize> {
+pub unsafe fn pop_mark_stack() -> Option<(usize, Tag)> {
```
crusso, line 65: I guess you could return a sentinal, not an option, for perf, but perhaps not worth the effort.

crusso (approved):

rts/motoko-rts/src/gc/mark_compact/mark_stack.rs
nomeata: Maybe Rust even does that automatically? It does for some types, but the list doens't mention tuples :https://doc.rust-lang.org/std/option/#representation

nomeata: It seems it does! https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5007ac15b58e9dca74e4b54789e139f2

osa1:
Last commit removes a redundant pass over roots so it should improve the perf even more. I'll rerun the benchmark.

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -108,11 +102,12 @@ unsafe fn mark_static_roots<M: Memory>(mem: &mut M, static_roots: SkewedPtr, hea
     // Static objects are not in the dynamic heap so don't need marking.
     for i in 0..root_array.len() {
         let obj = root_array.get(i).unskew() as *mut Obj;
-        mark_fields(mem, obj, heap_base);
+        mark_fields(mem, obj, obj.tag(), heap_base);
```
osa1, line 105: An optimization I've done in #2647: we know `obj` here is a `MutBox` so we can check that with a `debug_assert` and inline and specialize `mark_fields` for `MutBox`.

osa1: I implemented this.

osa1:
After all the optimizations current status is -17.15% in instructions compared to master.

rts/motoko-rts/src/gc/mark_compact.rs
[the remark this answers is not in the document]
osa1: Done in #2648.

rts/motoko-rts/src/gc/mark_compact.rs
osa1: Done in #2648.

osa1:
#2647 performs better than this one, closing.

Remaining comments above are addressed in: #2648

PR #2643, opened 2021-07-02
ci-pr.nix: Fix wasm-hash-for

### LINKED ISSUE #2631: dfinity-ci reports in PRs are no longer reliable
I realized today that @dfinity-ci's comments on binary size and perf changes are no longer reliable, I think it always says "This PR does not affect the produced WebAssembly code.". For example: https://github.com/dfinity/motoko/pull/2573#issuecomment-864967187 No idea what's wrong but perhaps it's related to the refactorings in the CI setup for open sourcing.

cc @crusso

nomeata:
in 58219bd06074ac924fe63d2007fa4bcf7106d9a3 I broke this script: The way
I used timeout was broken. This meant that we never produce any `.wasm`
files, which meant that the pr diff never noticed a differnce, and thus
didn’t print any performance regressions.

This fixes the bug and adds a safeguard that fails if no wasm is
generated.

Note that this compiles all files with `-ref-system-api`, meaning that
many tests in `run-drun` are not taken into account.

Fixes #2631


crusso:
@osa1 this might fix your issue with getting perf info

ci-pr.nix
```diff
@@ -28,11 +28,17 @@ let
...
           # ignore all errors
           echo -n $file
-          if timeout 10s -- moc $file -no-check-ir -ref-system-api -o $file.wasm 2>/dev/null
+          if timeout 10s moc $file -no-check-ir -ref-system-api -o $file.wasm 2>/dev/null
           then echo " failed (ignored)"
           else echo " ok"
           fi
         done
+
+        if ! find -name \*.wasm | grep -q .
```
crusso, line 37:
```suggestion
        if ! find -name */*.wasm | grep -q .
```

crusso: (not CI actually fails due to broken pipe around here...

nomeata: That changes the meaning. `\*.wasm` passes a pattern to `find`, `*/*.wasm` uses the shell to expand.

crusso: Actually, I couldn't figure out how to fix it myself. Any suggestions?

crusso: I think #2644 fixes this...

nomeata: Oh, there was breakage on CI? I thought you are just suggesting a different implementation.

Are you sure the broken pipe causes the error?

Anyways, won't be able to look until at the laptop again. (Phone right now)

ggreif:
@crusso any doubts with this left?

ggreif (approved):
Looks good to me!

PR #2644, opened 2021-07-02
Claudio/fix wasm hash for

crusso:
@nomeata Look ok to you?

### DIFF
--- a/ci-pr.nix
+++ b/ci-pr.nix
@@ -34,7 +34,7 @@ let
           fi
         done
 
-        if ! find . -name \*.wasm | cat | grep -q .
+        if ! ls */*.wasm
         then
           echo "No wasm files generated. wasm-hash-for broken?"
           exit 1
--- a/ci-pr.nix
+++ b/ci-pr.nix
@@ -34,7 +34,7 @@ let
           fi
         done
 
-        if ! ls */*.wasm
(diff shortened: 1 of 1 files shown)


osa1:
What is this PR about? Would be good to at least have a short PR description.

crusso:
It's trying to fix #2643 which in turn is trying to fix the perf regression reporting

ci-pr.nix
```diff
@@ -34,7 +34,7 @@ let
           fi
         done
 
-        if ! find -name \*.wasm | grep -q .
+        if ! ls */*.wasm
```
nomeata, line 37: Hmm, I dislike using non- matching globs like this. And using LS should not be needed: the check is already “done” when the bash is done expanding the glob, the `ls` kinda does unnecessary extra stuff.

It's surprising that this is relatively non-trivial to do elegantly…

nomeata: How about this? https://stackoverflow.com/a/34195247

crusso: compgen isn't available in the CI env

crusso: @nomeata this ok?

nomeata (approved):
This produces a large command line and may break once we have too many test files, but it breaks in the happy case (so it’s safe), and we can worry then.

PR #2647, opened 2021-07-05
 Compacting GC: thread backwards pointers in marking

osa1:
This PR implements the mark-compact GC algorithm briefly described by
"High-Performance Garbage Collection for Memory-Constrained Environments"
section 5.1.2 (I think the original paper is "An Efficient Garbage Compaction
Algorithm" but that paper is a bit difficult to follow).

The idea is as follows: when marking an object we thread backwards pointers of
that object. After we linearly scan the heap as before. For a live object, we
unthread it and update backwards pointers to it to the objects new location,
move the object, and then thread its forwards pointers. After the pass all
objects are moved and all references are updated.

This reduces number of passes in the original mark-compact collector to 2.

CanCan backend benchmark reports -19.8% in instructions. Compared to copying
GC, mark-compact GC is now 36% slower, instead of 70% as before.

This PR also optimizes `mark_static_roots` by inlining `mark_fields` in the
body and specializing it for `MutBox`: static root array only points to static
`MutBox`es so no need to call more general `mark_fields`. This optimization
gives us approximately -2% in instructions.

See also #2641 for another, slower variant of the algorithm.

### DIFF
--- a/rts/motoko-rts-tests/src/gc.rs
+++ b/rts/motoko-rts-tests/src/gc.rs
@@ -29,15 +29,43 @@ pub fn test() {
 }
 
 fn test_heaps() -> Vec<TestHeap> {
-    vec![TestHeap {
-        heap: hashmap! {
-            0 => vec![0, 2],
-            2 => vec![0],
-            3 => vec![3],
+    vec![
+        // Just a random test that covers a bunch of cases:
+        // - Self references
+        // - Forward pointers
+        // - Backwards pointers
+        // - More than one fields in an object
+        TestHeap {
+            heap: hashmap! {
+                0 => vec![0, 2],
+                2 => vec![0],
+                3 => vec![3],
+            },
+            roots: vec![0, 2, 3],
+            closure_table: vec![0],
         },
-        roots: vec![0, 2, 3],
-        closure_table: vec![0],
-    }]
+        // Tests pointing to the same object in multiple fields of an object. Also has unreachable
+        // objects.
+        TestHeap {
+            heap: hashmap! {
+                0 => vec![],
+                1 => vec![],
+                2 => vec![],
+            },
+            roots: vec![1],
+            closure_table: vec![0, 0],
+        },
+        // Root points backwards in heap. Caught a bug in mark-compact collector.
+        TestHeap {
+            heap: hashmap! {
+                0 => vec![],
+                1 => vec![2],
+                2 => vec![1],
+            },
+            roots: vec![2],
+            closure_table: vec![],
+        },
+    ]
 }
 
 #[derive(Debug)]
@@ -66,17 +94,6 @@ fn test_gc(
     closure_table: &[ObjectIdx],
 ) {
     let heap = MotokoHeap::new(refs, roots, closure_table, gc);
-
-    // Check `check_dynamic_heap` sanity
-    check_dynamic_heap(
-        refs,
-        roots,
-        closure_table,
-        &**heap.heap(),
-        heap.heap_base_offset(),
-        heap.heap_ptr_offset(),
-        heap.closure_table_ptr_offset(),
-    );
 
     for _ in 0..3 {
         gc.run(heap.clone());
--- a/rts/motoko-rts/src/gc/mark_compact.rs
+++ b/rts/motoko-rts/src/gc/mark_compact.rs
@@ -141,7 +141,7 @@ unsafe fn mark_fields<M: Memory>(mem: &mut M, obj: *mut Obj, obj_tag: Tag, heap_
         mark_object(mem, field_value, heap_base);
 
         // Thread if backwards pointer
-        if field_value.unskew() < field_addr as usize {
+        if field_value.unskew() < obj as usize {
             thread(field_addr);
         }
     });
--- a/rts/motoko-rts/src/gc/mark_compact.rs
+++ b/rts/motoko-rts/src/gc/mark_compact.rs
@@ -175,20 +175,21 @@ unsafe fn update_refs<SetHp: Fn(u32)>(set_hp: SetHp, heap_base: u32) {
     let mut bit = bitmap_iter.next();
     while bit != BITMAP_ITER_END {
         let p = (heap_base + (bit * WORD_SIZE)) as *mut Obj;
+        let p_new = free;
 
         // Update backwards references to the object's new location and restore object header
-        unthread(p, free);
+        unthread(p, p_new);
 
         // Move the object
         let p_size_words = object_size(p as usize);
-        if free as usize != p as usize {
-            memcpy_words(free as usize, p as usize, p_size_words);
+        if p_new as usize != p as usize {
+            memcpy_words(p_new as usize, p as usize, p_size_words);
         }
 
         free += p_size_words.to_bytes().0;
 
         // Thread forward pointers of the object
-        thread_fwd_pointers(p, heap_base);
+        thread_fwd_pointers(p_new as *mut Obj, heap_base);
 
         bit = bitmap_iter.next();
     }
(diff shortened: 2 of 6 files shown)


osa1:
CanCan backend bench reports -19.8% instructions compared to master.

osa1:
After porting optimizations done in this branch to #2641 this PR is now -3.24% compared to #2641.

rts/motoko-rts-tests/src/gc.rs
```diff
@@ -67,17 +95,6 @@ fn test_gc(
 ) {
     let heap = MotokoHeap::new(refs, roots, closure_table, gc);
 
-    // Check `check_dynamic_heap` sanity
```
osa1, line 70: This should be "create_dynamic_heap sanity".

rts/motoko-rts-tests/src/gc.rs
```diff
@@ -67,17 +95,6 @@ fn test_gc(
...
-    check_dynamic_heap(
-        refs,
-        roots,
-        closure_table,
-        &**heap.heap(),
-        heap.heap_base_offset(),
-        heap.heap_ptr_offset(),
-        heap.closure_table_ptr_offset(),
-    );
```
osa1, line 79: Same as previous PR, this is removed as `check_dynamic_heap` rejects garbage, but we want to test heaps with garbage as well. So currently `check_dynamic_heap` is only called after a GC, not before.

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -123,87 +126,89 @@ unsafe fn mark_object<M: Memory>(mem: &mut M, obj: SkewedPtr, heap_base: u32) {
...
 
-unsafe fn thread_roots(static_roots: SkewedPtr, heap_base: u32) {
-    // Static roots
-    let root_array = static_roots.as_array();
-    for i in 0..root_array.len() {
-        thread_obj_fields(root_array.get(i).unskew() as *mut Obj, heap_base);
-    }
-    // No need to thread closure table here as it's on heap and we already marked it
+        // Thread if backwards pointer
+        if field_value.unskew() < field_addr as usize {
```
crusso, line 144:
```suggestion
        if field_value.unskew() < field_addr as usize {
```
I guess that works, but shouldn't this really be tesing againt the address of the header, not some field offset from the hearder? At least morally?

crusso: visit_pointer_fields only visits unmarked fields, right? So there is no danger of threading the same field twice?

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -123,87 +126,89 @@ unsafe fn mark_object<M: Memory>(mem: &mut M, obj: SkewedPtr, heap_base: u32) {
...
-
-        free += size;
-
-        bit = bitmap_iter.next();
+/// Specialized version of `mark_fields` for root `MutBox`es.
+unsafe fn mark_root_mutbox_fields<M: Memory>(mem: &mut M, mutbox: *mut MutBox, heap_base: u32) {
+    let field_addr = &mut (*mutbox).field;
+    // TODO: Not sure if this check is necessary?
+    if pointer_to_dynamic_heap(field_addr, heap_base as usize) {
+        // TODO: We should be able to omit the "already marked" check here as no two root MutBox
```
crusso, line 155: That's probably true, but I'd check with Joachim first. Or just play it safe. Does the set of root boxes ever get large? My intuition is that it corresponds to top-level vars and recursive declarations that need back-patching, but I'm not sure.

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -123,87 +126,89 @@ unsafe fn mark_object<M: Memory>(mem: &mut M, obj: SkewedPtr, heap_base: u32) {
...
-        if free as usize != p as usize {
-            memcpy_words(free as usize, p as usize, p_size_words);
+        if p_new as usize != p as usize {
+            memcpy_words(p_new as usize, p as usize, p_size_words);
         }
 
         free += p_size_words.to_bytes().0;
 
+        // Thread forward pointers of the object
+        thread_fwd_pointers(p_new as *mut Obj, heap_base);
```
crusso, line 194: Is this really the right order in which to do this?
Something tells me you should be unthreading the forward pointers before doing the field move, as the test for whether something is a forward pointer depends on the address of the field (or original object).
Can you convince me this is correct?

crusso (commented):
I think I'll need some handholding to convince me.

rts/motoko-rts/src/gc/mark_compact.rs
osa1: > as the test for whether something is a forward pointer depends on the address of the field

Two things:

- This algorithm never reorders objects, so if A has a forward pointer to B, then after sliding A it will still have a forward pointer to B. I think this may be more clear if you consider that we only do linear scan from start of the heap to the end, and slide an object at an earlier address before objects at later addresses. So a forward pointer never becomes a backward pointer (or the other way around) as a result of moving objects.

- Threading and then moving would not work. To see why, suppose I have this heap: (assume that we store pointers directly, without skewing)
  ```
  0x100: [ARRAY, ..., 0x200]
  ...
  0x200: [BLOB, ...]
  ```
  So object at 0x100 has a forward pointer to object at 0x200.

  If I thread fields of 0x100 now (before sliding it) it will look like this:

  ```
  0x100: [ARRAY, ..., BLOB]
  ...
  0x200: [0x100, ...]
  ```
  Now if I slide object at `0x100` to its new location:
  ```
  0x050: [ARRAY, ..., BLOB]
  ...
  0x200: [0x100, ...]
  ```
  Note that I have a pointer to random data in the header slot of `0x200` now. I can't unthread it anymore.

  If I move first, then thread, it looks like this:
  ```
  0x050: [ARRAY, ..., BLOB]
  ...
  0x200: [0x050, ...]
  ```
  Now when I visit object at 0x200 I can unthread it.

Does this make sense?

rts/motoko-rts/src/gc/mark_compact.rs
osa1: You're right. I updated this code to compare the field value (pointer) with the object address, instead of field address.

rts/motoko-rts/src/gc/mark_compact.rs
crusso: Yes, that makes sense. I was being stupid.

crusso: You're diagrams are always excellent btw

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -123,87 +126,89 @@ unsafe fn mark_object<M: Memory>(mem: &mut M, obj: SkewedPtr, heap_base: u32) {
...
 
-unsafe fn thread_roots(static_roots: SkewedPtr, heap_base: u32) {
-    // Static roots
-    let root_array = static_roots.as_array();
-    for i in 0..root_array.len() {
-        thread_obj_fields(root_array.get(i).unskew() as *mut Obj, heap_base);
-    }
-    // No need to thread closure table here as it's on heap and we already marked it
+        // Thread if backwards pointer
+        if field_value.unskew() < obj as usize {
```
crusso, line 144: I though I had commented on this, but perhaps I forgot to submit.

Is there any danger you thread the same field twice, if you've already visited the object?

crusso (approved):

rts/motoko-rts/src/gc/mark_compact.rs
osa1: We should not visit an object twice. This function is called on values popped from mark stack, and we mark objects when pushing to mark stack and do not push marked objects to the mark stack. So a field should only be threaded once.

PR #2650, opened 2021-07-05
Smarter GC scheduling

osa1:
We now avoid doing GC after every message and only do when

- Heap grows more than HEAP_GROWTH_FACTOR since last GC (1) and more than
  SMALL_HEAP_DELTA is allocated since last GC (2)

or

- Current heap size is more than MAX_HEAP_FOR_GC (3)

(1) is to make sure we don't do GC when heap size does not change much. On
small heaps `heap_size * HEAP_GROWING_FACTOR` will be a small amount, so (2) is
to avoid doing redundant GCs in such cases. (3) is to make sure we'll have
allocation space for the next message when heap is larger than MAX_HEAP_FOR_GC.

(Thanks to @ulan for the algorithm)

New moc flag `--force-gc` added to do GC after every update message in tests.
New scheduling is enabled by default.

## Benchmarks

Using CanCan backend benchmark

instructions

total_instructions

dirtied_host_pages

total_dirtied_host_pages

accessed_host_pages

total_accessed_host_pages

total_Wasm_pages_in_use

With the current parameters (HEAP_GROWING_FACTOR, SMALL_HEAP_DELTA,
MAX_HP_FOR_GC) we only do GC once, somewhere between calls 600 and 800. As a
result, instructions and dirtied pages are dramatically lower, but we allocate
more Wasm pages. After the GC we no longer need to grow the heap so total Wasm
pages is constant.

## Testing

Currently tests allocate very little and with the new scheduler we end up not
doing any GC in tests. To avoid reducing test coverage tests are currently run
with `--force-gc`. After #2684 I will run drun tests with both schedulers.
Currently the new scheduler is tested manually using the test suite and CanCan
backend.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,20 @@
 * Trap on attempt to upgrade when canister not stopped and there are outstanding callbacks.
   (This failure mode can be avoided by stopping the canister before upgrade.)
 
+* New garbage collection scheduling: previously Motoko runtime would do gc
+  after every upgrade message. We now schedule a gc when
+
+  1. Heap grows more than 50% and 10 MiB since the last GC, or
+  2. Heap size is more than 3 GiB
+
+  (1) is to make sure we don't do gc on tiny heaps, and for larger heaps we
+  allow small allocations and do gc on large allocations. (2) is to make sure
+  on large heaps we will have enough allocation space in the next message.
+
+  New scheduling reduces cycles significantly at the cost of more Wasm pages.
+
+  New flag `--force-gc` restores the old behavior.
+
 == 0.6.5 (2021-07-08)
 
 * Add alternative, _compacting_ gc, enabled with new moc flag `--compacting-gc`.
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,17 +3,16 @@
 * Trap on attempt to upgrade when canister not stopped and there are outstanding callbacks.
   (This failure mode can be avoided by stopping the canister before upgrade.)
 
-* New garbage collection scheduling: previously Motoko runtime would do gc
-  after every upgrade message. We now schedule a gc when
+* Vastly improved garbage collection scheduling: previously Motoko runtime would do GC
+  after every upgrade message. We now schedule a GC when
 
   1. Heap grows more than 50% and 10 MiB since the last GC, or
   2. Heap size is more than 3 GiB
 
-  (1) is to make sure we don't do gc on tiny heaps, and for larger heaps we
-  allow small allocations and do gc on large allocations. (2) is to make sure
-  on large heaps we will have enough allocation space in the next message.
+  (1) is to make sure we don't do GC on tiny heaps or after only small amounts of allocation. (2) is to make sure that
+  on large heaps we will have enough allocation space during the next message.
 
-  New scheduling reduces cycles significantly at the cost of more Wasm pages.
+  This scheduling reduces cycles substantially, but may moderately increase memory usage.
 
   New flag `--force-gc` restores the old behavior.
 
--- a/rts/motoko-rts/src/gc.rs
+++ b/rts/motoko-rts/src/gc.rs
@@ -18,7 +18,7 @@ unsafe fn should_do_gc() -> bool {
 
     // Regardless of other parameters (`HEAP_GROWTH_FACTOR`, `SMALL_HEAP_DELTA`) we want to do GC
     // if `HP` is more than this amount.
-    const MAX_HP_FOR_GC: u64 = 3 * 1024 * 1024 * 1024; // 3 GiB
+    const MAX_HP_FOR_GC: u64 = 1 * 1024 * 1024 * 1024; // 3 GiB
 
     let heap_limit = min(
         max(
--- a/rts/motoko-rts/src/gc.rs
+++ b/rts/motoko-rts/src/gc.rs
@@ -23,7 +23,7 @@ unsafe fn should_do_gc() -> bool {
     let heap_limit = min(
         max(
             (f64::from(LAST_HP) * HEAP_GROWTH_FACTOR) as u64,
-            SMALL_HEAP_DELTA.0,
+            u64::from(LAST_HP) + SMALL_HEAP_DELTA.0,
         ),
         MAX_HP_FOR_GC,
     );
--- a/rts/motoko-rts/src/gc.rs
+++ b/rts/motoko-rts/src/gc.rs
@@ -4,6 +4,7 @@ pub mod mark_compact;
 #[cfg(feature = "ic")]
 unsafe fn should_do_gc() -> bool {
     use crate::memory::ic::{HP, LAST_HP};
+    use crate::types::Bytes;
 
     use core::cmp::{max, min};
 
@@ -13,7 +14,7 @@ unsafe fn should_do_gc() -> bool {
     // On small heaps `last_hp * HEAP_GROWTH_FACTOR` will be quite small. To avoid doing redundant
     // collections in such cases we only check `last_hp * HEAP_GROWTH_FACTOR` if at least this
     // much is allocated.
-    const SMALL_HEAP_DELTA: u64 = 10 * 1024 * 1024; // 10 MiB
+    const SMALL_HEAP_DELTA: Bytes<u64> = Bytes(10 * 1024 * 1024); // 10 MiB
 
     // Regardless of other parameters (`HEAP_GROWTH_FACTOR`, `SMALL_HEAP_DELTA`) we want to do GC
     // if `HP` is more than this amount.
@@ -22,7 +23,7 @@ unsafe fn should_do_gc() -> bool {
     let heap_limit = min(
         max(
             (f64::from(LAST_HP) * HEAP_GROWTH_FACTOR) as u64,
-            SMALL_HEAP_DELTA,
+            SMALL_HEAP_DELTA.0,
         ),
         MAX_HP_FOR_GC,
     );
--- a/rts/motoko-rts/src/gc.rs
+++ b/rts/motoko-rts/src/gc.rs
@@ -8,20 +8,20 @@ unsafe fn should_do_gc() -> bool {
     use core::cmp::{max, min};
 
     // A factor of last heap size. We allow at most this much allocation before doing GC.
-    const HEAP_GROWING_FACTOR: f32 = 1.5;
+    const HEAP_GROWTH_FACTOR: f32 = 1.5;
 
-    // On small heaps `last_hp * HEAP_GROWING_FACTOR` will be quite small. To avoid doing redundant
-    // collections in such cases we only check `last_hp * HEAP_GROWING_FACTOR` if at least this
+    // On small heaps `last_hp * HEAP_GROWTH_FACTOR` will be quite small. To avoid doing redundant
+    // collections in such cases we only check `last_hp * HEAP_GROWTH_FACTOR` if at least this
     // much is allocated.
     const SMALL_HEAP_DELTA: u32 = 10 * 1024 * 1024; // 10 MiB
 
-    // Regardless of other parameters (`HEAP_GROWING_FACTOR`, `SMALL_HEAP_DELTA`) we want to do GC
+    // Regardless of other parameters (`HEAP_GROWTH_FACTOR`, `SMALL_HEAP_DELTA`) we want to do GC
     // if `HP` is more than this amount.
     const MAX_HP_FOR_GC: u32 = 3 * 1024 * 1024 * 1024; // 3 GiB
 
     let heap_limit = min(
         max(
-            (LAST_HP as f32 * HEAP_GROWING_FACTOR) as u32,
+            (LAST_HP as f32 * HEAP_GROWTH_FACTOR) as u32,
             SMALL_HEAP_DELTA,
         ),
         MAX_HP_FOR_GC,
--- a/rts/motoko-rts/src/gc.rs
+++ b/rts/motoko-rts/src/gc.rs
@@ -1,2 +1,31 @@
 pub mod copying;
 pub mod mark_compact;
+
+#[cfg(feature = "ic")]
+unsafe fn should_do_gc() -> bool {
+    use crate::memory::ic::{HP, LAST_HP};
+
+    use core::cmp::{max, min};
+
+    // A factor of last heap size. We allow at most this much allocation before doing GC.
+    const HEAP_GROWING_FACTOR: f32 = 1.5;
+
+    // On small heaps `last_hp * HEAP_GROWING_FACTOR` will be quite small. To avoid doing redundant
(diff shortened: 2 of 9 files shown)


rts/motoko-rts/src/gc.rs
```diff
@@ -1,2 +1,31 @@
...
+    // much is allocated.
+    const SMALL_HEAP_DELTA: u32 = 10 * 1024 * 1024; // 10 MiB
+
+    // Regardless of other parameters (`HEAP_GROWING_FACTOR`, `SMALL_HEAP_DELTA`) we want to do GC
+    // if `HP` is more than this amount.
+    const MAX_HP_FOR_GC: u32 = 3 * 1024 * 1024 * 1024; // 3 GiB
+
+    let heap_limit = min(
+        max(
+            (LAST_HP as f32 * HEAP_GROWING_FACTOR) as u32,
```
ulan, line 24: I think this can overflow u32 if we are close to 4GB. Can we do all operations in u64?

rts/motoko-rts/src/gc.rs
```diff
@@ -1,2 +1,31 @@
...
+
+    use core::cmp::{max, min};
+
+    // A factor of last heap size. We allow at most this much allocation before doing GC.
+    const HEAP_GROWING_FACTOR: f32 = 1.5;
+
+    // On small heaps `last_hp * HEAP_GROWING_FACTOR` will be quite small. To avoid doing redundant
+    // collections in such cases we only check `last_hp * HEAP_GROWING_FACTOR` if at least this
+    // much is allocated.
+    const SMALL_HEAP_DELTA: u32 = 10 * 1024 * 1024; // 10 MiB
```
ulan, line 16: Do we have a type for NumBytes? Alternatively let's reflect the units (bytes) in the name?

rts/motoko-rts/src/gc.rs
```diff
@@ -1,2 +1,31 @@
...
+
+    // Regardless of other parameters (`HEAP_GROWING_FACTOR`, `SMALL_HEAP_DELTA`) we want to do GC
+    // if `HP` is more than this amount.
+    const MAX_HP_FOR_GC: u32 = 3 * 1024 * 1024 * 1024; // 3 GiB
```
ulan, line 20: Should we start conservatively here? (e.g. <1GB) and increase it in subsequent releases?

osa1:
RTS stats tests are currently failing as the accounting changes after this PR. I'll fix those tests later.

ulan (approved):
Thanks for implementing this so quickly! The improvements are impressive.

I have comment about a potential overflow and nits.

rts/motoko-rts/src/gc.rs
```diff
@@ -1,2 +1,31 @@
...
+
+    let heap_limit = min(
+        max(
+            (LAST_HP as f32 * HEAP_GROWING_FACTOR) as u32,
+            SMALL_HEAP_DELTA,
+        ),
+        MAX_HP_FOR_GC,
+    );
```
osa1, line 28: Another implementation of this is

```rust
    let heap_limit =
        ((LAST_HP as f32 + HEAP_GROWING_FACTOR) as u32).clamp(SMALL_HEAP_DELTA, MAX_HP_FOR_GC);
```

Not sure if this is more readable though..

osa1:
I think it would be useful to have plots total instructions, total dirtied pages, and total accessed pages. I'll update my script add those as well.

rts/motoko-rts/src/gc.rs
rossberg: Doesn't this criterion mean that we'll GC on every message after we've reached MAX_HP_FOR_GC of live data? Shouldn't this still be taking the amount of allocation since last GC into account somehow (as a heuristic for amount of dead data)?

rossberg (commented):
Fantastic numbers!

Can you explain why there is only one GC? I'm a bit confused about this, shouldn't the first GC happen at around SMALL_HEAP_DELTA, and then roughly in logarithmically decreasing distances? Not sure I understand how the heap size grows to hundreds of pages first.

rts/motoko-rts/src/gc.rs
ulan: My suggestion is about incremental rollout of the change: make the change for smaller heaps first, see that it works as expected, and then apply to larger ones. The motivation is that the risk of OOM is higher for larger heaps.

In general we do what a similar heap growing for large heap, but need to be more careful.

rossberg: @ulan, ah, sorry, my comment wasn't in reply to yours, just on the same line of code. :)  I agree with rolling out in stages. But I believe a fixed upper bound is gonna be a massive (prohibitive?) performance cliff?

rts/motoko-rts/src/gc.rs
```diff
@@ -1,2 +1,31 @@
...
 pub mod mark_compact;
+
+#[cfg(feature = "ic")]
+unsafe fn should_do_gc() -> bool {
+    use crate::memory::ic::{HP, LAST_HP};
+
+    use core::cmp::{max, min};
+
+    // A factor of last heap size. We allow at most this much allocation before doing GC.
+    const HEAP_GROWING_FACTOR: f32 = 1.5;
```
crusso, line 11:
```suggestion
    const HEAP_GROWING_FACTOR: f32 = 1.5;
```
HEAP_GROWTH_FACTOR?

rts/motoko-rts/src/gc.rs
osa1: > Doesn't this criterion mean that we'll GC on every message after we've reached MAX_HP_FOR_GC of live data

Only as long as heap size is larger than `MAX_HP_FOR_GC`, but yes. So if after a message we free enough space we'll be back with `heap size * growth factor` check again.

> But I believe a fixed upper bound that forces back into per-message GC is gonna be a massive (prohibitive?) performance cliff?

This is possible, yeah. I guess we should try and see. The "massive" case is the default today so this should still be an improvement.

Motivations for `MAX_HP_FOR_GC` are:

- If we only use `heap size * growth factor` then a program that allocates little over many calls will fill the heap. So at some point we should do GC regardless of how much the last message allocated.
- The risk of OOM as mentioned by Ulan.

osa1: > Should we start conservatively here? (e.g. <1GB) and increase it in subsequent releases?

Done. I'll update the plots.

osa1: > I'll update the plots.

Done.

osa1:
> I think it would be useful to have plots total instructions, total dirtied pages, and total accessed pages. I'll update my script add those as well.

Done, new plots in the description.

osa1:
It turns out CanCan backend benchmark is not allocating enough so we still do one GC in the benchmark with new scheduling. Would be good to have a better benchmark that allocates larger amounts.

rossberg:
Hm, so after the GC we have ~33000 pages, which is around 2G. If there was no GC before, then almost all of these pages must have been allocated in a single message, or am I missing something? How is that possible?

osa1:
> Hm, so after the GC we have ~33000 pages, which is around 2G. If there was no GC before, then almost all of these pages must have been allocated in a single message, or am I missing something? How is that possible?

I don't know. Just confirmed again that we run GC once in this benchmark with the new scheduling.

crusso:
> 3\. Current heap size is less than MAX_HP_FOR_GC

Should that by `more than`?

osa1:
> >    3. Current heap size is less than MAX_HP_FOR_GC
> 
> Should that by more than?

Good catch, fixed.

rossberg:
Hm, something seems very fishy then. Maybe some memory growth computation in the GC or allocator is bogus?

osa1:
> Hm, so after the GC we have ~33000 pages, which is around 2G. If there was no GC before, then almost all of these pages must have been allocated in a single message, or am I missing something? How is that possible?

I will try to answer this question this week.

Another question is how to do GC after every message in tests. Some ideas:

1. Have a moc flag to disable smart GC scheduling and do GC always. When used, it could pass a parameter to GC functions that says GC should run in every call.

2. Same as (1), but set it with a prim instead of in compile time.

3. Implement prims for setting the parameters `HEAP_GROWING_FACTOR`, `SMALL_HEAP_DELTA`, and `MAX_HEAP_FOR_GC`.

4. Implement a prim to force GC after the current message.

(1) and (2) would probably not be used by users, but (3) and (4) could be used to tweak GC scheduling in user canisters.

Any opinions on this? @crusso @ggreif @rossberg

osa1:
> > Hm, so after the GC we have ~33000 pages, which is around 2G. If there was no GC before, then almost all of these pages must have been allocated in a single message, or am I missing something? How is that possible?
> 
> I will try to answer this question this week.

This revealed a bug which I fixed in #2673. Plots make sense after the fix. I'll update the plots in the PR description after merging #2673.

rossberg:
> This revealed a bug which I fixed in #2673.

Excellent, I'm glad you found it! This amount of memory usage would be very expensive for users.

If possible, we should try to lamd and ship both the fix and this PR this week, since there is already user concern about the GC cost of many small messages.

osa1:
> If possible, we should try to land and ship both the fix and this PR this week

I think the only blocker left (after #2673 which will be merged in a few mins) is https://github.com/dfinity/motoko/pull/2650#issuecomment-886402808. We should do GC after every message in tests otherwise test coverage will be greatly reduced.

rossberg:
For now, I would go for (1), which seems like the minimum functionality. A runtime GC interface is a much larger design question, and prone for misuse.

osa1:
Plots updated after #2673

osa1:
I added a flag to force GC and I'm using that in tests, but I'm not sure how to test new scheduling. With the new scheduling tests like `rts_stats2` fail because they check things like heap size, and there is currently no way to skip a test or use a different expected output file depending on compiler flags.

rossberg (approved):
Thanks, that looks great!

One more side question: What exactly do dirtied & accessed host pages count? The spikes in these graphs when GC happens are much higher than the number of "Wasm pages in use", how is that?

crusso:
> I added a flag to force GC and I'm using that in tests, but I'm not sure how to test new scheduling. With the new scheduling tests like `rts_stats2` fail because they check things like heap size, and there is currently no way to skip a test or use a different expected output file depending on compiler flags.

Maybe add a flag to disable schedukling and add it to set it in that one test? I think there is a comment pragma (like //MOC-FLAG) for setting extra args... (see drun/actor-import.mo).

osa1:
I didn't know `//MOC-FLAG`, thanks. I think we want to force GC in all tests, otherwise none of the tests do GC (they allocate very little) so we end up not testing GC at all. That will significantly reduce test coverage.

I think ideally we should be testing all canister tests with `--force-gc`, and some also without `--force-gc`. First part is already done, but for the second part we need to implement some way to skip a test when `--force-gc` is disabled (with an additional `--no-force-gc`) or not used at all. Normally it wouldn't be too difficult, but I found maintaining run.sh a bit difficult..

osa1:
> One more side question: What exactly do dirtied & accessed host pages count? The spikes in these graphs when GC happens are much higher than the number of "Wasm pages in use", how is that?

Dirtied host pages = modified host pages, accessed host pages = read host pages. On the benchmark system a page is 4KiB (reported by `getconf PAGESIZE`) so a Wasm page is 16 host pages. One of those spikes in the CSV files looks like this:

```
canister,method,instructions,accessed_pages,dirtied_pages,total_pages_in_use
...
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,26353018,2206,2202,166
...
```

Here accessed *Wasm* pages is something like ~137. One thing to note here is even though a Wasm page is 16 host pages on the benchmark system, reading a Wasm page does not make 16 host pages accessed or dirtied. Only the actually touched host page is dirtied/accessed. That explains why "dirtied pages" is not a multiple of 16 in the example above.

rossberg:
Ah, thanks.

rossberg:
Btw, this PR should probably include a prominent Changelog entry. :)

crusso (approved):

osa1:
I was talking to @ulan about this algorithm and we found a bug in the code: `SMALL_HEAP_DELTA` part needs to be `last_hp + SMALL_HEAP_DELTA`, otherwise on small heaps we may redundantly do GC. For example, on 15 MiB heap if I allocate 9 MiB no need to do GC. Correct heap limit should be `min(max(last_hp * HEAP_GROWTH_FACTOR, last_hp + SMALL_HEAP_DELTA), MAX_HEAP_FOR_GC)`.

This only changes GC behavior on tiny heaps (less than 20 MiB with the current `HEAP_GROWTH_FACTOR`) so it shouldn't matter in practice, but I'll repeat the benchmarks and confirm.

osa1:
`SMALL_HEAP_DELTA` part is only useful on tiny heaps, and given that we also consider static data I think that part of the calculation may be completely useless. I will compile some of the canister tests and check initial heap pointer ("heap base", static heap size) to have an idea of initial heap sizes.

osa1:
> This only changes GC behavior on tiny heaps (less than 20 MiB with the current HEAP_GROWTH_FACTOR and SMALL_HEAP_DELTA) so it shouldn't matter too much in practice, but I'll repeat the benchmarks and confirm.

No changes in the benchmarks.

rossberg:
Good catch.

osa1:
@ulan @rossberg @crusso could you please review the changelog entry? I wanted to avoid things like "amortizing gc costs with mutator allocations", "allocation patterns" etc. as that would be too technical. What do you think?

Changelog.md
```diff
@@ -3,6 +3,20 @@
...
 
+* New garbage collection scheduling: previously Motoko runtime would do gc
+  after every upgrade message. We now schedule a gc when
+
+  1. Heap grows more than 50% and 10 MiB since the last GC, or
+  2. Heap size is more than 3 GiB
+
+  (1) is to make sure we don't do gc on tiny heaps, and for larger heaps we
+  allow small allocations and do gc on large allocations. (2) is to make sure
+  on large heaps we will have enough allocation space in the next message.
```
rossberg, line 14:
```suggestion
  (1) is to make sure we don't do GC on tiny heaps or after only small amounts of allocation. (2) is to make sure that
  on large heaps we will have enough allocation space during the next message.
```

Changelog.md
```diff
@@ -3,6 +3,20 @@
...
+
+  New scheduling reduces cycles significantly at the cost of more Wasm pages.
```
rossberg, line 16:
```suggestion
  This scheduling reduces cycles substantially, but may moderately increase memory usage.
```

ulan:
The changelog lgtm. We could also mention that the "GC after 3GB" heuristic will be improved in the future. One option would be to set the limit at `(last_hp + 4GB) / 2` - i.e. the half-way to the max heap size.

crusso:
LGTM

PR #2653, opened 2021-07-06
Memory size investigation

crusso:


crusso:
This was just an investigation into a possible issue raised in a private communication on the forum.

PR #2660, opened 2021-07-13
Codegen: remove unused function

osa1:


ggreif (approved):
I think it is okay to remove this, it was needed for the DWARF gen. I can bring it back if it is still needed.

PR #2663, opened 2021-07-16
Allocate just one `ClosureTable` entry per self-send

### LINKED ISSUE #2640: [BUG] Possible memory leaks through closure table, induced by faulting callbacks.
Our one-shot callbacks are responsible for releasing their slots in the closure table. 
If a callback fails, it's slot won't be released, leading to a memory leak.

The System API now has some support for cleanups on callback failures, but I don't think Motoko exploits that yet.

NB: This certainly was an issue a while back and I think it probably still is. See #2106 for the first mention of this problem.

@nomeata?

(c.f
https://sdk.dfinity.org/docs/interface-spec/index.html#system-api-call
````
ic0.call_on_cleanup : (fun : i32, env : i32) -> ()
````

- If the a cleanup callback is specified for this call, it is executed if and only if the reply or the reject callback was executed and trapped (for any reason).
- During the execution of the cleanup function, only a subset of the System API is available (namely ic0.debug_print, ic0.trap and the ic0.stable_* functions). The cleanup function is expected to run swiftly (within a fixed, yet to be specified cycle limit) and serves to free resources associated with the callback.
- If this traps (e.g. runs out of cycles), the state changes from the cleanup function are discarded, as usual, and no further actions are taken related to that call. Canisters likely want to avoid this from happening.
- There must be at most one call to ic0.call_on_cleanup between ic0.call_new and ic0.call_perform.)

### PR DESCRIPTION
**Pull request by ggreif**
This PR cleans up the `ClosureTable` entries in the last-resort _cleanup callback_. Since potentially all call and reply state changes of a canister can be rolled back by carefully engineering the traps, the initiator of a message send must only allocate exactly one slot in the closure table. Formerly self-sends would leak a slot when the receiver trapped and the reject callback too. See https://github.com/dfinity/motoko/pull/2663#issuecomment-884974977.

Motivation: fixes #2640

### CODE REVIEW

**Comment by ggreif** [comment]
`await-without-yield` and `commit-on-await` (in `run-drun/`) already trigger the cleanup callbacks (when installed). No `one-shot` cleanup callback is ever seen to be triggered.

**Comment by ggreif** [comment]
We have several problems currently.
a) non-self-send `async` methods may invoke a reply/reject continuation that traps. The index allocated in the `ClojureTable` and what it points to will never be deallocated.
b) Explicit self-sends allocate `ClojureTable` index (for the _future_) that is supposed to be removed _on the other side_ of the call. If that traps, the sender-side becomes the owner of the future's slot. Then the reject callback gets invoked and can (additionally) pop it. The trouble starts when the reject callback _also_ traps. Then the cleanup callback will be invoked and must clean up two slots. But the problem is that if the reply callback trapped it only needs to clean-up one slot. But the cleanup callback has no information whatsoever what happened.

For a) the solution is easy, pop the one slot.
For b) we have to consolidate the two slots to one (pointing to a 3-element array) in which _the other side_ can peep to fetch the future.

**Comment by crusso** [comment]
To avoid serializing/deserialing the result of the future, if wonder if it would make sense to transfer it it in a fourth slot of that array?

**Comment by ggreif** [comment]
> To avoid serializing/deserialing the result of the future, if wonder if it would make sense to transfer it it in a fourth slot of that array?

Yep, I was thinking overwriting the third position of the array (i.e. the future) since that is not needed any more. But sending the payload back _in heap_ changes the semantics a bit, so I'd only do that for _single, small_ results.

File: src/codegen/compile.ml [hunk 1]
```diff
@@ -5795,23 +5796,102 @@ module FuncDec = struct
...
+        G.i (LocalGet (nr 0l)) ^^
+        ClosureTable.recall env ^^
+        set_arr ^^ get_arr ^^
+        Arr.load_field 0l ^^ (* get the reply closure *)
+        set_closure ^^
+        get_closure ^^
+
+        (* Deserialize reply arguments  *)
+        Serialization.deserialize env ts ^^
+        (*G.i Drop ^^
```
ggreif, line 5815: This is an idea to pass back (sufficiently small) results to the future's caller. Needs a follow-up issue.

src/codegen/compile.ml
```diff
@@ -5836,9 +5944,11 @@ module FuncDec = struct
       ignoring_callback env ^^
       compile_unboxed_zero ^^
       (* The reject callback *)
-      ignoring_callback env ^^
+      ignoring_callback env ^^ (* TODO: figure out why this one gets called in oneshot-callbacks and not the reply one! *)
```
ggreif, line 5947: Needs issue.

src/codegen/compile.ml
```diff
@@ -5836,9 +5944,11 @@ module FuncDec = struct
...
       compile_unboxed_zero ^^
       IC.system_call env "ic0" "call_new" ^^
+      (* NOT NEEDED! cleanup_callback env ^^ compile_unboxed_zero ^^
```
ggreif, line 5950: remove

src/codegen/compile.ml
```diff
@@ -5881,7 +5991,7 @@ module FuncDec = struct
         (* Deserialize and look up closure argument *)
         Serialization.deserialize env Type.[Prim Nat32] ^^
         BoxedSmallWord.unbox env ^^
-        ClosureTable.recall env ^^
+        ClosureTable.peek_future env ^^ (*HERE*)
```
ggreif, line 5994: clean up

src/codegen/compile.ml
```diff
@@ -6830,7 +6940,7 @@ and compile_exp (env : E.t) ae exp =
           let add_cycles = Internals.add_cycles env ae in
           code1 ^^ StackRep.adjust env fun_sr SR.Vanilla ^^
           set_meth_pair ^^
-          compile_exp_as env ae SR.Vanilla e2 ^^ set_arg ^^
+          compile_exp_vanilla env ae e2 ^^ set_arg ^^
```
ggreif, line 6943: unrelated, move to own PR

src/codegen/compile.ml
```diff
@@ -7516,24 +7626,25 @@ and compile_exp (env : E.t) ae exp =
...
-    set_closure_idx ^^
+    set_future ^^
 
-    compile_exp_as env ae SR.Vanilla exp_k ^^ set_k ^^
-    compile_exp_as env ae SR.Vanilla exp_r ^^ set_r ^^
+    compile_exp_vanilla env ae exp_k ^^ set_k ^^
+    compile_exp_vanilla env ae exp_r ^^ set_r ^^
 
-    FuncDec.ic_call env Type.[Prim Nat32] ts
+    FuncDec.ic_self_call env Type.[Prim Nat32] ts
```
ggreif, line 7641: no need to  pass ts1, it is an implementation detail

test/only.mk
```diff
@@ -0,0 +1,4 @@
+.PHONY: %.only
```
ggreif, line 1: remove this (see #2669)

test/run-drun/oneshot-callbacks.mo
```diff
@@ -0,0 +1,29 @@
...
+    Prim.debugPrint("go 0: " # debug_show Prim.rts_callback_table_count());
+    oneway_ping();
+    await async {Prim.debugPrint("go 1: " # debug_show Prim.rts_callback_table_count())};
+    await ping();
+    try {
+        ignore await (async {Prim.debugPrint("go 2: " # debug_show Prim.rts_callback_table_count()); assert false; 42})
+    } catch _ { ignore 42/0 }
+  };
+};
+await a.go(); //OR-CALL ingress go "DIDL\x00\x00"
```
ggreif, line 23: pass bool, to select `catch` or regular

rts/motoko-rts/src/closure_table.rs
```diff
@@ -92,6 +92,25 @@ pub unsafe fn remember_closure<M: Memory>(mem: &mut M, ptr: SkewedPtr) -> u32 {
...
+        rts_trap_with("peek_future_closure: Closure index out of range");
+    }
+
+    let ptr = TABLE.as_array().get(idx);
+
+    if ptr.0 & 0b1 != 1 {
+        rts_trap_with("peek_future_closure: Closure index not in table");
+    }
+
+    ptr.as_array().get(2)
```
ggreif, line 111: This magic needs documentation

rts/motoko-rts/src/types.rs
```diff
@@ -226,11 +226,13 @@ impl Array {
     }
 
     pub unsafe fn get(self: *mut Self, idx: u32) -> SkewedPtr {
+        debug_assert!(self.len() > idx);
```
ggreif, line 229: This is now #2675.

src/codegen/compile.ml
```diff
@@ -5795,23 +5796,102 @@ module FuncDec = struct
...
+      get_cb_index ^^
+      compile_unboxed_const (E.add_fun_ptr env (E.built_in env reject_name)) ^^
+      get_cb_index
+
   let ignoring_callback env =
     let name = "@ignore_callback" in
     Func.define_built_in env name ["env", I32Type] [] (fun env -> G.nop);
     compile_unboxed_const (E.add_fun_ptr env (E.built_in env name))
 
+  let _faulting_callback env =
```
ggreif, line 5866: maybe remove this function?

src/codegen/compile.ml
```diff
@@ -7516,24 +7626,25 @@ and compile_exp (env : E.t) ae exp =
...
-    FuncDec.ic_call env Type.[Prim Nat32] ts
+    FuncDec.ic_self_call env Type.[Prim Nat32] ts
       ( IC.get_self_reference env ^^
         IC.actor_public_field env (IC.async_method_name))
-      (get_closure_idx ^^ BoxedSmallWord.box env)
+      get_future
       get_k
-      get_r add_cycles
+      get_r
+      add_cycles (* why? *)
```
ggreif, line 7647: it doesn't make sense to send cycles to ourselves

src/codegen/compile.ml
ggreif: removed

src/codegen/compile.ml
ggreif: done

src/codegen/compile.ml
ggreif: Done: #2674.

src/codegen/compile.ml
ggreif: Done.

test/run-drun/oneshot-callbacks.mo
ggreif: Done.

src/codegen/compile.ml
ggreif: Removed.

src/codegen/compile.ml
ggreif: Written: #2679.

rts/motoko-rts/src/types.rs
ggreif: Merged.

test/only.mk
ggreif: Done.

src/codegen/compile.ml
ggreif: 5b76ab1663b21c5dfcc383b77050f1a25c7b7fdd

rts/motoko-rts/src/closure_table.rs
ggreif: Done in ca5451008.

Changelog.md
```diff
@@ -1,6 +1,10 @@
 # Motoko compiler changelog
 
-== 0.6.5 (2021-07-08)
+* == 0.6.6 (2021-0?-??)
```
crusso, line 3:
```suggestion
```
We'll add that line when we do the release.

test/run-drun/oneshot-callbacks.mo
```diff
@@ -0,0 +1,30 @@
...
+  };
+
+  public func ping() : async () {
+    Prim.debugPrint("ping-async! " # debug_show Prim.rts_callback_table_count());
+  };
+
+  public func go(trigger_cleanup : Bool) : async () {
+    Prim.debugPrint("go 0: " # debug_show Prim.rts_callback_table_count());
+    oneway_ping();
+    await async {Prim.debugPrint("go 1: " # debug_show Prim.rts_callback_table_count())};
```
crusso, line 16:
```suggestion
    await async {
       Prim.debugPrint("go 1: " # debug_show Prim.rts_callback_table_count())
    };
```

test/run-drun/oneshot-callbacks.mo
```diff
@@ -0,0 +1,30 @@
...
+    await ping();
+    try {
+        ignore await (async {Prim.debugPrint("go 2: " # debug_show Prim.rts_callback_table_count()); assert false; 42})
```
crusso, line 19:
```suggestion
        ignore await async { 
           Prim.debugPrint("go 2: " # debug_show Prim.rts_callback_table_count()); 
           assert false; 
           42
        }
```

test/run-drun/oneshot-callbacks.mo
```diff
@@ -0,0 +1,30 @@
...
+    } catch _ { if trigger_cleanup { ignore 42/0 } }
+  };
+};
+
+await a.go(true); //OR-CALL ingress go "DIDL\x00\x01\x7E\x01"
+await a.go(false); //OR-CALL ingress go "DIDL\x00\x01\x7E\x00"
+await a.go(true); //OR-CALL ingress go "DIDL\x00\x01\x7E\x01"
+
```
crusso, line 27: Could you add another (perhaps separate) test that uses recursion to have a closure count > 1?

src/codegen/compile.ml
```diff
@@ -7516,24 +7616,25 @@ and compile_exp (env : E.t) ae exp =
...
-    set_closure_idx ^^
+    set_future ^^
 
     compile_exp_vanilla env ae exp_k ^^ set_k ^^
     compile_exp_vanilla env ae exp_r ^^ set_r ^^
 
-    FuncDec.ic_call env Type.[Prim Nat32] ts
+    FuncDec.ic_self_call env ts
       ( IC.get_self_reference env ^^
         IC.actor_public_field env (IC.async_method_name))
```
crusso, line 7633:
```suggestion
      (IC.get_self_reference env ^^
       IC.actor_public_field env (IC.async_method_name))
```

src/codegen/compile.ml
```diff
@@ -5795,23 +5796,94 @@ module FuncDec = struct
       compile_unboxed_const (E.add_fun_ptr env (E.built_in env reject_name)) ^^
       get_cb_index
 
+  let closures_to_self_reply_reject_callbacks env ts =
```
crusso, line 5799: This code looks almost identical to the one above? I wonder if its possible to actually share the implementation, if, instead of passing in the array components, you just pass in the get_array code.  Then I think you would also wind up sharing the code for the success continuations (that do deserialization). But maybe I'm missing some essential difference... 
Try to refactor in a branch off this one to see?

src/codegen/compile.ml
```diff
@@ -5795,23 +5796,94 @@ module FuncDec = struct
...
+
+        message_cleanup env (Type.Shared Type.Write)
+      );
+
+    (* The (above) upper half of this function must not depend on the
+       get_future, get_k and get_r parameters, so hide them from above (cute trick) *)
+    fun get_future get_k get_r ->
+      let (set_cb_index, get_cb_index) = new_local env "cb_index" in
+      (* store the tuple away, future_array_index = 2, keep in sync with rts/closure_table.rs *)
+      Arr.lit env [get_k; get_r; get_future] ^^
```
crusso, line 5848:
```suggestion
      Arr.lit env [get_k; get_r; get_future] ^^
```
Continuing from my comment above: is this the only difference from the cloned code? If so, maybe refactor the original as:
```
fun get_array ->
      let (set_cb_index, get_cb_index) = new_local env "cb_index" in
      (* store the tuple away, future_array_index = 2, keep in sync with rts/closure_table.rs *)
      get_array ^^
...
```
and have the caller pass in the appropriate get_array code (constructed at the call site)

If possible, this will avoid duplicating some gnarly code (and share more code to boot).

src/codegen/compile.ml
```diff
@@ -5881,7 +5981,7 @@ module FuncDec = struct
         (* Deserialize and look up closure argument *)
         Serialization.deserialize env Type.[Prim Nat32] ^^
         BoxedSmallWord.unbox env ^^
-        ClosureTable.recall env ^^
+        ClosureTable.peek_future env ^^
```
crusso, line 5984: Just to clarify, if the future traps, when does the future itself get recalled? Is the answer as part of the reject handler of the self-call?

crusso (changes requested):
Nice work. Seems fine as is but would be really nice to get rid of the code duplication by refactoring (and maybe add another or extended test with recursion).

src/codegen/compile.ml
ggreif: If the future traps ("on the other side") then the reject handler will clean up. If that fails, the cleanup-callback is the last resort.

src/codegen/compile.ml
ggreif: I agree, this is grossly duplicated. Will look into it how to bring them together.

test/run-drun/oneshot-callbacks.mo
ggreif: good point: cdf94f6

rts/motoko-rts/src/closure_table.rs
```diff
@@ -92,6 +92,29 @@ pub unsafe fn remember_closure<M: Memory>(mem: &mut M, ptr: SkewedPtr) -> u32 {
     idx
 }
 
+// Position of the future in explicit self-send ClosureTable entries
```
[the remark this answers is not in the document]
ggreif: `ContinuationTable`?

test/run-drun/oneshot-callbacks.mo
```diff
@@ -2,6 +2,24 @@ import Prim "mo:⛔";
 
 actor a {
 
+  public func drill(rabbit_hole : Nat8) : async () {
```
crusso, line 5: Looks good, thanks!

rts/motoko-rts/src/closure_table.rs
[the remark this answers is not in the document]
osa1: `ContinuationTable` sounds good to me.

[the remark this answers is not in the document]
osa1: I think we should also rename "future closure" below to "continuation".

src/codegen/compile.ml
ggreif: Done. In a ~~few~~ bunch of commits!

src/codegen/compile.ml
ggreif: Yep, the `ic_call_threaded` now deals with code holes that are customisable. This way I could unify remote calls and self calls, which only differ in argument passing and how the continuations are pushed.

rts/motoko-rts/src/closure_table.rs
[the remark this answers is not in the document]
ggreif: Good, I'll do this as a follow-up PR.

src/codegen/compile.ml
```diff
@@ -5790,35 +5757,39 @@ module FuncDec = struct
...
+    let (set_cb_index, get_cb_index) = new_local env "cb_index" in
+    Arr.lit env closure_getters ^^
+    ClosureTable.remember env ^^
+    set_cb_index ^^
+
+    (* return arguments for the ic.call *)
+    compile_unboxed_const (E.add_fun_ptr env (E.built_in env reply_name)) ^^
+    get_cb_index ^^
+    compile_unboxed_const (E.add_fun_ptr env (E.built_in env reject_name)) ^^
+    get_cb_index
```
crusso:
```suggestion
      let (set_cb_index, get_cb_index) = new_local env "cb_index" in
      Arr.lit env closure_getters ^^
      ClosureTable.remember env ^^
      set_cb_index ^^

      (* return arguments for the ic.call *)
      compile_unboxed_const (E.add_fun_ptr env (E.built_in env reply_name)) ^^
      get_cb_index ^^
      compile_unboxed_const (E.add_fun_ptr env (E.built_in env reject_name)) ^^
      get_cb_index
```
(indentation)

crusso (commented):
LGTM. 

(Good you refactored the ic_(self)_call too - wasn't actually asking for that ;->)

crusso (approved):
LGTM. 

(Good you refactored the ic_(self)_call too - wasn't actually asking for that ;->)

(One formatting change only)

src/codegen/compile.ml
ggreif: Yeah, I am blaming my `emacs` mode for messing this up ;-)

Fixed.

rts/motoko-rts/src/closure_table.rs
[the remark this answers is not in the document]
ggreif: Here we go: #2690

nomeata:
Thanks for this! I had quite some trouble figuring out what’s happening based on the patch only, but the comment in https://github.com/dfinity/motoko/pull/2663#issuecomment-884974977 helped. Maybe worth expanding the code comments a bit, so that future readers (who don’t have the benefit of the PR discussion) can make sense of it? Probably one comprehensive GHC-style code node that explains how async self-calls works would be useful.

PR #2669, opened 2021-07-21
add makefile helper to only run a single test

ggreif:
Usage:
```
make -C test/run-drun commit-on-await.only
```
The test must be a Motoko one from an inner directory.
This is pretty neat when you repeatedly run your preferred test and don't want to waste time waiting on the others.

Otherwise this extension should be rather cheap, and won't interfere if you don't want to use it.

### DIFF
--- a/test/only.mk
+++ b/test/only.mk
@@ -1,0 +1,4 @@
+.PHONY: %.only
+
+%.only: %.mo
+	../run.sh $(RUNFLAGS) $<


nomeata (commented):
I always just used `../run.sh file/to/test` for that. Advantage is that it works also with multiple tests, and you can pass `-a`. And no risk that the test doesn't run because make thinks the “build product” is up to date 

Downside: you have to remember to pass `-d` sometimes. I was long wondering if this setting should be stored in a file in the directory so that `run.sh` always does the right thing.

Using make as a test runner always rubs me the wrong way (make is for building software, not running it). My recommendation is to keep the logic in run.sh and use make not at all, or only as a trivial list of common commands. And maybe reimplement run.sh in a less controversial language one day.

test/only.mk
```diff
@@ -0,0 +1,4 @@
+.PHONY: %.only
+
+%.only: %.mo
```
nomeata, line 3: Some tests have different file endings (`.sh`, or none if they are directories)

crusso:
Yeah, I'm not convinced this is worth it (agreeing with @nomeata)

ggreif:
Well, no problem at all, I can just leave that file around (out of `git`). Works as well for me. Thanks!

PR #2673, opened 2021-07-26
Mark-compact GC: update STACK_TOP when growing mark stack

osa1:
When growing mark stack we should update `STACK_TOP` otherwise subsequent
pushes will think that the stack is full and call `grow_stack` again.

Discovered while debugging strange allocation pattern in #2650 with the
compacting GC. It turns out once the mark stack is full ever subsequent push
(until stack is smaller than the initial size again) doubles the stack size,
causing redundant allocations.

Regression test added.

### DIFF
--- a/rts/motoko-rts/src/gc/mark_compact/mark_stack.rs
+++ b/rts/motoko-rts/src/gc/mark_compact/mark_stack.rs
@@ -13,13 +13,13 @@ pub const INIT_STACK_SIZE: Words<u32> = Words(64);
 static mut STACK_BLOB_PTR: *mut Blob = null_mut();
 
 /// Bottom of the mark stack
-static mut STACK_BASE: *mut usize = null_mut();
+pub static mut STACK_BASE: *mut usize = null_mut();
 
 /// Top of the mark stack
-static mut STACK_TOP: *mut usize = null_mut();
+pub static mut STACK_TOP: *mut usize = null_mut();
 
 /// Next free slot in the mark stack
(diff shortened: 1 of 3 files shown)


crusso (approved):
Is this a bug fix?

osa1:
@crusso

> Is this a bug fix?

Yes, updated the PR description.

rts/motoko-rts/src/gc/mark_compact/mark_stack.rs
```diff
@@ -13,13 +13,13 @@ pub const INIT_STACK_SIZE: Words<u32> = Words(64);
 static mut STACK_BLOB_PTR: *mut Blob = null_mut();
```
crusso, line 13: Keeping that one private deliberately?

crusso (approved):
Still LGTM.

Does this explain the weird GC cliff Andreas observed in your earlier PR? Probably not, but might be worth checking

rts/motoko-rts/src/gc/mark_compact/mark_stack.rs
osa1: It's not used in tests so no need to make it `pub`.  Should I make it `pub` for consistency?

osa1:
> Does this explain the weird GC cliff Andreas observed in your earlier PR? Probably not, but might be worth checking

This fixes that, yes.

crusso:
Yeah, sorry, that was actually clear from your updated PR description ;->

PR #2675, opened 2021-07-26
Bounds-check `Array` access for the debug runtime

ggreif:
This adds debug assertions that no access outside of `Array`'s length is attempted.

### DIFF
--- a/rts/motoko-rts/src/types.rs
+++ b/rts/motoko-rts/src/types.rs
@@ -226,11 +226,13 @@ impl Array {
     }
 
     pub unsafe fn get(self: *mut Self, idx: u32) -> SkewedPtr {
+        debug_assert!(self.len() > idx);
         let slot_addr = self.payload_addr() as usize + (idx * WORD_SIZE) as usize;
         *(slot_addr as *const SkewedPtr)
     }
 
     pub unsafe fn set(self: *mut Self, idx: u32, ptr: SkewedPtr) {
+        debug_assert!(self.len() > idx);
         let slot_addr = self.payload_addr() as usize + (idx * WORD_SIZE) as usize;
         *(slot_addr as *mut SkewedPtr) = ptr;
     }


rts/motoko-rts/src/types.rs
```diff
@@ -226,11 +226,13 @@ impl Array {
     }
 
     pub unsafe fn get(self: *mut Self, idx: u32) -> SkewedPtr {
```
ggreif, line 228: @osa1 I wonder if `*const Self` would be appropriate here...

osa1: @ggreif these methods currently all have `*mut Self` receiver because until recently it wasn't possible to call a `*const Self` method on a `*mut` value, see https://github.com/rust-lang/rust/issues/80258. Without the fix for that issue it's quite noisy to use these methods as we need casts `x as *const _` everywhere.

I fixed that issue in https://github.com/rust-lang/rust/issues/80258 so we just need to update rustc.

ggreif: Not many people seem to use raw pointers in Rust...

osa1: MMTk is also implemented in Rust, I'd expect it to use raw pointers in the GC code. Would be interesting to look at how they use raw pointers.

osa1 (approved):

rts/motoko-rts/src/types.rs
ggreif: Maybe we can prune this list then?
``` rust
// TODO (osa): Some of these are stabilized, we need to update rustc
#![feature(
    arbitrary_self_types,
    panic_info_message,
    assoc_char_funcs,
    core_intrinsics,
    ptr_offset_from
)]
```

osa1: Yeah most of those should also be gone with an update.

ggreif: Yay, [here it comes](https://blog.rust-lang.org/2021/07/29/Rust-1.54.0.html)!

PR #2677, opened 2021-07-26
Trap in canister_pre_upgrade when canister not stopped and callback count is non-zero

### LINKED ISSUE #2676: [BUG] pre_upgrade should trap when pending closure count > 0
At the moment, it is possible to "live" upgrade a Motoko actor that has pending requests, causing the (old) continuations of those requests to be incorrectly resumed in the upgraded actor. The old continuations are invalid in the new actor as they refer to (unstable) code pointers and will almost certainly fail, or worse, silently corrupt the heap.

To prevent this situation, we can modify the canister_pre_upgrade to trap if the Motoko closure_table is non-empty, i.e. when `rts.closure_count != 0`.

( Rejected alternative solution:
  make pending callbacks fail safely by upgrading the closure table, preserving the (index) domain, but replacing the codomain with callbacks that just immediately fail appropriately.
 Unfortunately, I can't see how to make this work when updating a foreign canister (or legacy Motoko canister) with a new  Motoko canister, as is there is no reliable way discover the retired closure table without cooperation from the retired canister (e.g. in Motoko, by writing the domain of the closure table to a stable variable).

@nomeata @rossberg what do you think?

# Related discussion copied from slack:

> Paul Liu 15 hours ago 
> I have a developer approaching me with a bug: canister upgrade happened when there was still outstanding inter-canister calls. This corrupted their canister state. Do we know that if this is indeed possible? @Akhi Singhania
> 
> Roman Kashitsyn  15 hours ago
> I don't think it's a bug. The specification doesn't require the canister to be stopped before the upgrade. In fact, with a bit of cleverness one can build a canister that can handle cross-canister calls spanning upgrades.
> However, it's generally a good idea to stop your canister before doing an upgrade. Both Rust CDK and Motoko aren't designed for live upgrades. (edited) 
> 
> Paul Liu  14 hours ago
> Say canister A was calling B, and B calling itself. When B was calling itself, it went through an upgrade mid way. In this case, am I right that B will not receive any error callback from calling itself, because B would have lost the calling context of the outgoing call.
> But A was still intact, so will will A receive an error callback?
> 
> Paul Liu  14 hours ago
> In fact, with a bit of cleverness one can build a canister that can handle cross-canister calls spanning upgrades.
> I'm afraid it is just a lot of boilerplate than cleverness: one has to use a global counter to keep track of out-standing inter-canister calls, use try/catch catch around every call to make sure this counter is updated accurately, and throw an error during preupgrade when this counter is not 0. It is quite bad.
> 
> Roman Kashitsyn  13 hours ago
>  one has to use a global counter to keep track of out-standing inter-canister calls, use try/catch catch around every call to make sure this counter is updated accurately, and throw an error during preupgrade when this counter is not 0
> I'm not sure I follow. What you describe prevents upgrades, not makes them "live".
> The "cleverness" I had in mind is to use a single gateway function with a stable index for all callback invocations (e.g., Motoko already does this anyway) and allocate callback environments in a way that is persisted across upgrades (e.g., in stable memory, or in a data structure that is marshalled through stable memory during upgrade. this typically implies "no async/await sugar for you").
>  am I right that B will not receive any error callback from calling itself, because B would have lost the calling context of the outgoing call.
> Callbacks are never lost, call contexts should survive the upgrade. In your case, B_upgraded will receive a callback set by B_old, which can easily lead to memory corruption in case of Rust CDK (we'll free memory that was never allocated by this version of B).
> The safe way to upgrade B is to stop if first, which will prevent future calls until all B's outstanding calls are closed. (edited) 
> 
> Paul Liu  13 hours ago
> It does look like stopping the canister before upgrades is safer than rolling a soluton on your own. I'll just tell the dev to stop canister first.
>  In your case, B_upgraded will receive a callback set by B_old, which can easily lead to memory corruption in case of Rust CDK
> I don't think motoko would be able to handle this either. Please correct me if I'm wrong. @Claudio Russo (edited) 
> 
> Roman Kashitsyn  13 hours ago
> > It does look like stopping the canister before upgrades is safer than rolling a soluton on your own
> Yes, definitely. I don't even think that it's possible to tell rustc to place a function at specific index in the Wasm binary it generates.
> There were a few proposals to make live upgrades more feasible, but they were never implemented:
> https://github.com/dfinity-lab/ic-ref/pull/183
> https://github.com/dfinity-lab/ic-ref/pull/195
> #183 Public Spec: Single entry point for callbacks
> <https://github.com/dfinity-lab/ic-ref|dfinity-lab/ic-ref>dfinity-lab/ic-ref | Sep 28th, 2020 | Added by GitHub
> #195 Public Spec: Callbacks as named, exported functions
> <https://github.com/dfinity-lab/ic-ref|dfinity-lab/ic-ref>dfinity-lab/ic-ref | Oct 6th, 2020 | Added by GitHub
> :+1:
> 1
> :thankyou:
> 1
> 
> 
> Akhi Singhania:spiral_calendar_pad:  3 hours ago
> @Paul Liu, it seems like Roman covered your question well.  It will be good to know that it was not the case that teh developer had stopped the canister but there were still callbacks outstanding.
> 
> Claudio Russo  2 hours ago
> (@Paul Liu you are correct, Motoko is no better at this)
> 
> Claudio Russo  2 hours ago
> For Motoko, I wonder if we could make pending callbacks fail safely by upgrading the closure table, preserving the (index) domain, but replacing the codomain with callbacks that just immediately fail appropriately.
> 
> Claudio Russo  2 hours ago
> Or just trap in pre_upgrade if the closure table is non-empty, to fail live upgrades, without further instrumentation.
> :100:
> 2
> 
> 
> Claudio Russo  28 minutes ago
> If I do that, won't a canister that goes into an infinite loop starve all upgrade requests? Or do we rely on the cycle limit to ensure that no such infinity loops are possible? Is that even true on, say, the nns subnet? (edited) 
> 
> Roman Kashitsyn  25 minutes ago
> Instruction limit takes care of infinite loops. It's still possible to starve "live" upgrades by constantly sending messages to other canisters or to self, but this can be "fixed" by stopping the canister before the upgrade. (edited) 
> :+1:
> 1
> 
> 
> Claudio Russo  25 minutes ago
> Just out of curiousity, what happens if we upgrade a canister with an open call context that will never get closed because of the upgrade - will the execution engine leak that context?
> 
> Roman Kashitsyn  22 minutes ago
> I'm not sure I understand how an upgrade can prevent closing of a call context. Call contexts survive upgrades. Given that we're guaranteed to receive a response for each request, the call context should eventually be closed (even if there was an upgrade between the request and the response). (edited) 
> 
> Claudio Russo  8 minutes ago
> I was thinking if the continuation is corrupted due to an upgrade, it won't call msg_reply or msg_reject, but I guess you are right that the system will nevertheless create a response. Thanks for the clarification!
> New
> 
> Roman Kashitsyn  4 minutes ago
> That's correct, if the callback doesn't call msg_reply/msg_reject, the system will generate an automatic reply and close the call context._

crusso:
Fixes #2676 

Although this appears to do the right thing for the replica (drun), it seems superfluous for ic-ref. 

@nomeata does ic-ref already wait for the call back queue to drain before doing an upgrade?


crusso:
Hmm, I wonder if the test might be flaky, because its output depends on the interleaving of `repeat` and upgrade....

src/codegen/compile.ml
```diff
@@ -3478,7 +3478,10 @@ module IC = struct
 
     let pre_upgrade_fi = E.add_fun env "pre_upgrade" (Func.of_body env [] [] (fun env ->
       Lifecycle.trans env Lifecycle.InPreUpgrade ^^
-      G.i (Call (nr (E.built_in env "pre_exp"))) ^^
+      ClosureTable.count env ^^
+      G.if_ []
+       (E.trap_with env "callback count non-zero")
```
rossberg, line 3483: Maybe make the message a bit more self-explanatory?

crusso:
@rossberg PTAL

ggreif:
Luc Mercier once voiced strong reservation against trapping in the `pre_upgrade` hook. I'll try to find the message (Slack?).

test/run-drun/live-upgrade.drun
```diff
@@ -0,0 +1,15 @@
...
+# wait for pending callbacks to clear
+ingress $ID wait "DIDL\x00\x00"
+# upgrade should succeed
+upgrade $ID live-upgrade/live-upgrade.mo ""
+query $ID check "DIDL\x00\x01\x7d\x02"
+ingress $ID repeat "DIDL\x00\x00"
+# upgrade should fail due to pending callbacks
+upgrade $ID live-upgrade/live-upgrade.mo ""
+ingress $ID wait "DIDL\x00\x00"
+query $ID check "DIDL\x00\x01\x7d\x02"
```
ggreif, line 15: please add NL

src/codegen/compile.ml
```diff
@@ -3478,7 +3478,10 @@ module IC = struct
 
     let pre_upgrade_fi = E.add_fun env "pre_upgrade" (Func.of_body env [] [] (fun env ->
       Lifecycle.trans env Lifecycle.InPreUpgrade ^^
-      G.i (Call (nr (E.built_in env "pre_exp"))) ^^
+      ClosureTable.count env ^^
+      G.if_ []
+       (E.trap_with env "canister_pre_upgrade trapped due to outstanding message callbacks (try stopping the canister before upgrade)")
```
ggreif, line 3483: "trap" is already contained in the execution level string "Canister rwlgt-iiaaa-aaaaa-aaaaa-cai _trapped_ explicitly", so we shouldn't repeat it. Suggesting:
```suggestion
       (E.trap_with env "canister_pre_upgrade attempted with outstanding message callbacks (try stopping the canister before upgrade)")
```

crusso:
> Luc Mercier once voiced strong reservation against trapping in the `pre_upgrade` hook. I'll try to find the message (https://dfinity.atlassian.net/browse/NNS1-318). The context is different, but worth some analysis. The [Slack thread](https://dfinity.slack.com/archives/C0175CHJ0P6/p1620561601038600) is added for completeness.

I think you have a point here. If have leak in the closure table, we won't be able to upgrade.

However, I think we still want to avoid the possibility of corruption, so  I suggest we trap only if the canister isn't stopped and has pending callbacks. That means stopping the canister first will always prevent this trap.

This should be our friend for that check:

https://sdk.dfinity.org/docs/interface-spec/index.html#system-api-canister-status

ggreif (approved):
Looks good. Just a few nits. Please be aware that adding such functionality in presence of leaked `ClosureTable` slots (a real bug, see #2663) will render affected canisters unupgradable when using some `dfx` version > 0.8.0. So this change must be synchronised with that fix when releasing.

crusso:
@ggreif PTAL. I've modified the code to avoid trapping when the canister is stopped, so that should make it less likely that we paint ourselves into a corner with a leaking closure table.

ggreif (approved):

src/codegen/compile.ml

Issue #2671, opened 2021-07-24
[Bug] Large Nat32 literal overflow in JS


chenyan-dfinity:
The following code raise an exception in `int_of_big_int` in Motoko Playground. 

```
actor {
    let a: Nat32 = 0x80000000; // raise int_of_big_int
    let b: Nat64 = 0x80000000; // okay
    let c: Nat32 = 0x7fffffff; // okay
};
```

Not necessarily a compiler bug, but would be nice if we can support this in `js_of_ocaml` somehow. The exception only happens when a `Nat32` literal is larger than 28 bits. The interpreter in JS works fine though. 

cc @ggreif

ggreif:
We don't have many uses of `int_of_big_int` in `moc`:
```
$ git grep int_of_big_int
src/codegen/compile.ml:          [ Int32.of_int (Big_int.int_of_big_int b) ] @ go a
src/mo_values/numerics.ml:  let to_int = int_of_big_int
src/mo_values/numerics.ml:    else power_big_int_positive_int x (int_of_big_int y)
src/mo_values/prim.ml:    let i = Big_int.int_of_big_int i in
```
I am trying to narrow down which one is causing this. There is a possibility that it comes from another library altogether, though not probable.

ggreif:
@chenyan-dfinity I wonder where the 28 bit arise. https://ocaml.org/api/Sys.html#VALint_size mentions 31 and 32 (on JS).

ggreif:
I tried Motoko Playground, and with your code deployment gets stuck after:
> [17:03:35] Compiling code...

I wonder where do you get the piece of information from the log that `int_of_big_int` is the culprit.

chenyan-dfinity:
You can see the error from console.log, e.g. Tools/Developer Tools in Chrome.

chenyan-dfinity:
> Also, 0x7fffffff seems like 31 trailing binary ones.

Ah, right. It's 31bit. Maybe your fix works, i.e. use Int32 instead of int.

chenyan-dfinity:
That can be caught by the frontend? By the time it reaches codegen, we are safe.

ggreif:
Yeah, sure, that is fine, but you still want `Nat32` in Motoko, so we should handle it. Btw. I see in the playground there is no `try`-`catch` around `const out = await worker.Moc({ type:"compile", file });`, is this by design?

chenyan-dfinity:
[...]

PR #2682, opened 2021-07-27
Use `int32_of_big_int` when computing bignum limbs

ggreif:
I suspect this will also resolve potential problems with trapping on 32-bit compilation platforms. I don't think it is related to #2671, but who knows?

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -2497,7 +2497,7 @@ module BigNumLibtommath : BigNumType = struct
         then []
         else
           let (a, b) = Big_int.quomod_big_int n twoto28 in
-          [ Int32.of_int (Big_int.int_of_big_int b) ] @ go a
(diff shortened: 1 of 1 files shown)


src/codegen/compile.ml
```diff
@@ -2497,7 +2497,7 @@ module BigNumLibtommath : BigNumType = struct
         then []
         else
           let (a, b) = Big_int.quomod_big_int n twoto28 in
```
chenyan-dfinity, line 2499: Curious why it's 28 bit here?

chenyan-dfinity (approved):

src/codegen/compile.ml
ggreif: @nomeata is this some MP thing?

nomeata: Yes, I think the internal representation of libtommath bigints uses 28 bit limbs

PR #2684, opened 2021-07-28
Rewrite test runner in OCaml

osa1:
This replaces run.sh with OCaml. Not ready, submitting to get early feedback.

The goal is everyone in the team should be able to read, understand, and modify test runner easily.

Currently only runs drun tests but does not compare outputs yet.

To test:

- `touch enable-internals`
- `nix-shell`
- `cd test/runner`
- `dune build moc_test.exe`
- `./_build/default/moc_test.exe`

Preprocessing of drun scripts is also done in OCaml code so this replaces drun-wrapper.sh and perl invocations.

My plan is to gradually replace run.sh, instead of replacing all of it in one PR. Currently I need more flexibility when running drun tests so I started with replacing drun tests.

## Notes, questions, and todos

- I'm using [alcotest](https://github.com/mirage/alcotest). As far as I can see from the README it doesn't allow parallelism or concurrency so I'm looking for an alternative that supports parallelism and/or concurrency. Since tests are mostly waiting on drun I think concurrency should be good enough (i.e. one OS thread but tests are interleaved as blocked).
- See `--help` for test options. By default prints in test runner are not shown and redirected to a file if you use the `-o` parameter.
- ~~I don't understand why test names are printed twice when running. In the example in alcotest's README they're printed once.~~ Fixed.
- OCaml code is formatted with ocamlformat. I'll update CI code to check formatting.
- The OCaml process is not responsive to `C-c`. It seems like signals are handled by the spawned processes instead of OCaml runtime, as it causes test to fail but the test runner keeps running.
- It seems like when multiple tests fail alcotest only shows one of the outputs? I currently see 4 failures, but only outputs of one of them is shown at the end of the test run. If I clean the tree with `git clean -xfd` and re-run the tests it shows output of another failing test. Update: I need to pass `-e` to see all failures. Why is this not default??
- Output files generated when using `-o` contain ANSI escapes for colors. This should be an alcotest bug, it doesn't make sense to write terminal escape codes to files.
- Try running with `-e -c`, the output is really neat.
- ~~I can't figure out how to run just a single test. `./_build/default/moc_test.exe list | grep reinstall` shows that the test "reinstall" has number 21, but `./_build/default/moc_test.exe test 21` doesn't do anything ("no tests to run, filter skipped everything").~~ Fixed.

## Inconsistencies and problems in run.sh

- Directives `//MOC-FLAG` and `//MOC-ENV` are only allowed in .mo files. So if a drun test needs `MOC-FLAG` or `MOC-ENV` directives it can't be implemented as a drun file and needs to be implemented as a single-file test.
- run.sh runs drun tests implemented as single .mo files with the interpreters, but the interpreters don't really do anything. I think we could skip interpreters.
- run.sh has an explicit type checking test (shown as `[tc]` in test outputs) when running .mo files, but for .drun we don't have such a step. It seems inconsistent. We already collect all .mo files referenced by a .drun file to compile, so we could have a `[tc]` step for those files.


test/runner/moc_test.ml
```diff
@@ -0,0 +1,190 @@
...
+              Printf.sprintf
+                "diff -a -u -N --label expected ../run-drun/ok/%s.drun.ok \
+                 --label actual %s/%s.drun.out"
+                test_name out_dir test_name
+            in
+
+            (* TODO: Somehow this line is printed after the `diff` output.
+               OCaml runtime probably has its own output buffer, we need to
+               flush it somehow. *)
+            Printf.printf "Outputs do not match, running %s\n" diff_cmd;
```
crusso, line 153: You can use format string `%!` to flush, I think https://ocaml.org/releases/4.07/htmlman/libref/Printf.html.

Reminds me of something I heard in school: "flush twice, it's a long way to the candy machine."

crusso: You're a star for doing this. What do we conclude if the ocaml is way longer than the bash? ;->

osa1:
I can now run test implemented as .drun files. I will now implement testing single-file drun tests (.mo files in drun/).

test/runner/moc_test.ml
kritzcreek: > What do we conclude if the ocaml is way longer than the bash?

I don't think people dislike bash because of its verbosity...

nomeata:
> * `cd test/runner`
> 
> * `dune build moc_test.exe`
> 
> * `./_build/default/moc_test.exe`

You should add it to `./bin` (add a symlink and extend `wrapper.sh`), then one can just run `moc_test`.

PR #2685, opened 2021-07-28
Remove own `default_uncaught_exception_handler`

ggreif:
we have OCaml 4.12 now, so `Printexc` provides it

### DIFF
--- a/src/mo_config/internal_error.ml
+++ b/src/mo_config/internal_error.ml
@@ -1,8 +1,9 @@
 
 let setup_handler () =
-  Printexc.record_backtrace true;
-  Printexc.set_uncaught_exception_handler (fun exn rb ->
+  let open Printexc in
+  record_backtrace true;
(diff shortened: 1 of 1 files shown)


osa1:
Does this change the error output in any way?

ggreif:
> Does this change the error output in any way?

No. But I so far haven't figured out how to get backtraces, so that's a point that remains to be clarified.

osa1:
> so far haven't figured out how to get backtraces

I think you need `OCAMLRUNPARAM=b` before running.

LGTM but if this changes the output it would be good to show before and after in the commit message.

osa1 (approved):

ggreif:
The function replacing ours is  https://github.com/ocaml/ocaml/pull/9263/files#diff-5fa438e4a308c8dfd6c75768649ae4fe17b72159d89f5ae3e7319c8eeaf2b790R273 and looks exactly the same.

PR #2686, opened 2021-07-28
Guard against Motoko 32 bit values overflowing the compiler's 31-bit arithmetic

### LINKED ISSUE #2671: [Bug] Large Nat32 literal overflow in JS

ggreif:
Since `NumType.of_int` returns an `int`, this has not enough precision (just 31 bits) when
the compiler is 32-bit, e.g. on JS (Motoko Playground)

This patch switches to using `Big_int` instead.

Fixes #2671.

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -6116,6 +6116,13 @@ end (* AllocHow *)
 
 (* The actual compiler code that looks at the AST *)
 
+(* wraps a bigint in range [0…2^32-1] into range [-2^31…2^31-1] *)
+let nat32_to_int32 n =
+  let open Big_int in
+  if ge_big_int n (power_int_positive_int 2 31)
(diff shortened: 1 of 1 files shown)


crusso:
Nice catch!

chenyan-dfinity (approved):
This works in the playground!

PR #2689, opened 2021-07-29
Tweak `Changelog.md`

ggreif:


Changelog.md
```diff
@@ -5,12 +5,12 @@
...
 
 * Vastly improved garbage collection scheduling: previously Motoko runtime would do GC
-  after every upgrade message. We now schedule a GC when
+  after every update message. We now schedule a GC when
 
   1. Heap grows more than 50% and 10 MiB since the last GC, or
   2. Heap size is more than 3 GiB
 
-  (1) is to make sure we don't do GC on tiny heaps or after only small amounts of allocation. (2) is to make sure that
+  (i) is to make sure we don't do GC on tiny heaps or after only small amounts of allocation. (ii) is to make sure that
```
rossberg, line 13: These refer to the (1) and (2) above, why confuse the reader by spelling them Roman here?

ggreif:
Because the renderer shows it as "i" and "ii". See image above.

ggreif:
But you are right, if somebody reads the unrendered `.md` it is confusing. Will the primary reader audience see the text or the rendering?

rossberg:
In that case, can't you change the 1./2. to Roman as well?

Changelog.md
ggreif: Since the Roman numerals come from GitHub's braindead CSS, the HTML should be fine elsewhere. So I went back to the original.

ggreif:
> In that case, can't you change the 1./2. to Roman as well?

Sadly, no. `  i. para` doesn't count as an ordered list item.

I reverted my change.

PR #2690, opened 2021-07-29
Rename: closure*table -> continuation*table

ggreif:
This is a wholesale rename, in the spirit of https://github.com/dfinity/motoko/pull/2663#discussion_r678530699


src/codegen/compile.ml
```diff
@@ -5745,7 +5745,7 @@ module FuncDec = struct
         (* Look up closure *)
         let (set_closure, get_closure) = new_local env "closure" in
```
ggreif, line 5746: Rename these too.

rts/motoko-rts-tests/src/closure_table.rs
```diff
@@ -1,14 +1,14 @@
...
+use motoko_rts::continuation_table::{closure_count, recall_closure, remember_closure};
 use motoko_rts::types::{SkewedPtr, Words};
 
 pub unsafe fn test() {
     println!("Testing closure table ...");
 
     assert_eq!(closure_count(), 0);
 
-    const N: usize = 2000; // >256, to exercise `double_closure_table`
+    const N: usize = 2000; // >256, to exercise `double_continuation_table`
```
ggreif: What is `double_continuation_table`?

osa1: Not on my dev computer right now so can't check, but there should be some code path for doubling the size of closure table. This number was chosen to exercise that. I guess there isn't a function with that name anymore, this should just say "> 256 to exercise growing continuation table".

ggreif: Ah, I see:
```
$ git grep double_continuation_table
rts/motoko-rts-tests/src/continuation_table.rs:    const N: usize = 2000; // >256, to exercise `double_continuation_table`
rts/motoko-rts/src/continuation_table.rs:unsafe fn double_continuation_table<M: Memory>(mem: &mut M) {
rts/motoko-rts/src/continuation_table.rs:        double_continuation_table(mem);
```

src/codegen/compile.ml
ggreif: No, turns out these are actually closures, already extracted from the continuation array.

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -94,12 +94,12 @@ unsafe fn mark_compact<M: Memory, SetHp: Fn(u32)>(
 
     mark_static_roots(mem, static_roots, heap_base);
 
-    if (*closure_table_ptr_loc).unskew() >= heap_base as usize {
-        // TODO: No need to check if closure table is already marked
-        mark_object(mem, *closure_table_ptr_loc, heap_base);
-        // Similar to `mark_root_mutbox_fields`, `closure_table_ptr_loc` is in static heap so it
-        // will be readable when we unthread closure table
-        thread(closure_table_ptr_loc);
+    if (*continuation_table_ptr_loc).unskew() >= heap_base as usize {
```
crusso, line 97: @osa1 Just to double check, your gc of the closure/continuation table doesn't assume anything about the Motoko object stored there right? It's just that they used to be scalars, closures or 2-element tuples and now may be scalars or 2 or 3 element tuples after @ggreif's recent change.

src/codegen/compile.ml
```diff
@@ -7264,11 +7264,11 @@ and compile_exp (env : E.t) ae exp =
 
     | OtherPrim "rts_callback_table_count", [] ->
```
crusso, line 7265: Ha! Maybe we should have called it the callback table! But let's stick with this for now.

crusso: Do we want to rename the rts_ functions too, or just leave them for now?

crusso (approved):
But see comments.

rts/motoko-rts/src/gc/mark_compact.rs
osa1: @crusso From gc point of view closure table is just a heap object that is also a root. We don't make any assumptions on the type of it, or how many elements it has, or whether elements are scalar or pointers.

src/codegen/compile.ml
ggreif: I'll leave them, this is an external interface.

PR #2691, opened 2021-07-29
Ill-typed upgrades causing data-loss (repro)

### LINKED ISSUE #2692: Ill-typed upgrades cause data-loss
Doing an ill-typed upgrade can cause stable data loss.

The problem is the recent change to candid to support defaulting of optional fields.

Motoko encodes stable variables in a record of options. 

Doing an upgrade to an incompatible type will cause those optional fields to default to null, discarding any data those fields might have held (and just running the initializer instead).

A conformance check prior to upgrade could warn about this potential data loss, but absent tooling for that, we are in the situation where users can easily lose their data on upgrade by doing a simple, and natural, mistake like adding a field to stable variable of record type (or, more commonly, map thereof).

#2691 contains a repro of a sequence of ugrades that should fail but don't.

https://forum.dfinity.org/t/questions-about-data-structures-and-migrations/822/15 is a real world example of a user being tripped up by this, trying to "upgrade" a stable array of tuples containing records by adding a field.
Instead of failing the upgrade, the stable variable is regarded as uninitialized and initialized from this initializer expression (losing the data).

Solution: I briefly considered just changing the record of options encoding to a record of variants encoding for stable variables, which would be fine, but isn't  backwards compatible with previously compiled code, so a non-starter.

I'll investigate whether we can just change the deserialization behaviour for "T.Memory" records to avoid the option defaulting instead.

@rossberg, @nomeata whaddya think? 

Of course, we could say we really need the static check, but I think a defensive implementation is called for anyway.

crusso:


luc-blaeser:

PR #2702, opened 2021-08-02
Implement write barriers for heap object updates

osa1:
This is just a PoC, not for merging. This PR adds write barriers to heap object
updates. The goal is to find all the places that will need a write barrier for
a generational gc.

## Testing

RTS code is updated to make it easy to find all the places that needs a write
barrier. After a gc we copy the dynamic heap. Before the next gc we compare the
current heap with the copy. For a heap location in the current heap that is
different than the copy, we check that the location was recorded with the write
barrier. If not, it means that we miss a write barrier somewhere in the
generated code or the RTS. Note that this test ignores new allocations as we
don't have those in the copy so we can't know before a gc whether a new
allocation is updated or not.

Currently all test pass (except a few ones that check RTS stats). This doesn't
mean we have write barriers in all the necessary places, but at least the
common cases should be covered.


crusso:
Did  you want me to review this?

osa1:
No need to review yet, this is not useful on its own.

PR #2706, opened 2021-08-04
WIP: Implement page allocation

osa1:
This PR implements a new memory allocator for Motoko heap. It will be used for
implementing generational GC and other GC schemes.

Note: text below will be copied to code as documentation in the final version
of this PR.

# Overview of the API

## Page allocator API

`PageAlloc` trait defines the interface for page allocators. Page size is
currently a constant and same as Wasm page size (64 KiB). This allows a simple
implementation where allocating a new page can be done with `memory.grow`.

**TODO** The test allocator takes a "page size" argument, but `Space` assume 64
KiB pages (uses `motoko_rts::page_alloc::PAGE_SIZE` constant). We should remove
the "page size" argument from test page allocator for now.

Usable part of a page is a few words smaller than the page size, because of the
metadata at the beginning of pages.

A requirement for page allocators is an efficient way of mapping an address to
the page metadata (to the start of the page). This is needed to be able to get
the bitmap of a page in mark-compact GC, when marking an object. Since Wasm
pages are 64 KiB-aligned, this is currently implemented as masking the low bits
of the address. This requirement comes from mark-compact GC which needs to
efficiently find bitmap of an object, when marking. Bitmap pointers are stored
in page headers.

A page allocator needs to be able to allocate multiple consecutive pages, to be
able to allocate objects larger than usable parts of a page (currently 64 KiB
minus a few words). However, we do not require "get page start" method (as
described in the previous paragraph) to work on multi-page allocations, for a
few reasons:

- We don't have interior pointers, so whenever we have a pointer to a large
  object, the pointer will be to the first page of the large object, so "get page
  start" will still work.

- We don't need to get page header of a large object, since we don't mark large
  object using page bitmaps.

Page allocator trait currently has two implementations:

- A zero-sized struct `IcPageAlloc`, used in the RTS linked with the generated
  code. Being zero-sized means this type does not exist in runtime. This type
  is implemented in `motoko_rts::page_alloc::ic`.

- An implementation for testing: `motoko_rts_tests::page_alloc::TestPageAlloc`.

The trait:

```rust
pub trait PageAlloc: Clone {
    type Page: Page;

    /// Allocate single page
    unsafe fn alloc(&self) -> Self::Page;

    /// Allocate given number of consecutive pages
    unsafe fn alloc_pages(&self, n_pages: u16) -> Self::Page;

    /// Free a page
    unsafe fn free(&self, page: Self::Page);

    /// Get the start address of the page that contains the given address. Only works on
    /// single-page pages! Can be used to get page header.
    unsafe fn get_address_page_start(&self, addr: usize) -> usize;
}
```

### Implementation

We maintain two free lists:

- `SIZE_SORTED`: a free list with pages sorted on page sizes (number of pages).
  This free list is used for efficient allocation of requested amout of pages.

  This list contains `(size class, page set)` pairs (actually a struct), sorted
  on `size class`. When an allocation requrested for `N` pages, we find the
  smallest size class larger than or equal to `N`, split the page into smaller
  pages if the size class has more pages than `N` and free the extra pages.

- `ADDR_SORTED`: a free list with pages sorted on page start addresses (or Wasm
  page numbers). This list is used for coalescing consecutive free pages when
  freeing a page.

  This list contains `(Wasm page number, number of pages)` pairs (actually a
  struct), sorted on `Wasm page number`. When freeing a page, we find
  consecutive pages before and after the freed page, and coalesce those pages
  into a single page.

**Note on the complexity analyses below:** In the implementation we currently
use vectors for size class sets and `ADDR_SORTED` free list, but in the final
version we will use a binary tree (or Rust's `BTree`) implementation where
insertion and deletion are `O(log(N))` operations. So in the analyses below we
assume `O(log(N))` insertion and deletion.

**Runtime complexity of freeing:** *O(log(N))* where N is the number of free
pages.

- Find index of the page being freed in `ADDR_SORTED` free list. *O(log(N))*

- Find pages before and after the page being freed, for coalescing. There will
  be at most one page before and after the page being freed, since the pages
  before and after are also coalesced when they were freed. *O(1)*

- Pages found in the previous step are removed from `ADDR_SORTED` free list and
  also from `SIZE_SORTED` free list. Both operations are *O(log(N))*.

- Coalesced page added to `SIZE_SORTED` and `ADDR_SORTED` lists. Both operations
  are *O(log(N))* (see below).

**Runtime complexity of allocation:** *O(log(N))* where N is the number of free
pages.

- Find size class in `SIZE_SORTED` free list. *O(log(N1))* where `N1` is the
  number of size classes.

- Remove a page from the size class. This will be *O(log(N2))* where `N2` is
  the number of pages in the size class.

- Since `N1 * N2 = N`, max. values of `N1` and `N2` will be `sqrt(N)`. So the
  complexity of steps so far will be `O(log(sqrt(N)) + log(sqrt(N)))` = `O(1/2
  log(N))` = `O(log(N))`.

- (Assuming allocation request is smaller than the page) Split the removed page
  into two smaller pages (one with the requested amount, one for the extra
  pages), free the extra page. As described above, freeing is `O(log(N))`
  operation.

## Page API

Page allocator trait has an associated "page" type. This type needs to support
a few operations, listed below:

```rust
pub trait Page: Clone {
    /// Get the start of this page
    unsafe fn start(&self) -> usize;

    /// Get the start of this page after the page header (e.g. linked list fields)
    unsafe fn contents_start(&self) -> usize;

    /// Get the end (exclusive) of this page
    unsafe fn end(&self) -> usize;

    /// Size of the page, excluding any headers
    unsafe fn size(&self) -> usize {
        self.end() - self.contents_start()
    }

    unsafe fn get_bitmap(&self) -> Option<*mut Bitmap>;

    unsafe fn set_bitmap(&self, bitmap: Option<Bitmap>);
}
```

These methods are used in `Space` implementation (see below), and by the
garbage collectors, to be able to scan a page.

Bitmap methods are used by mark-compact collector. When marking, we first get
the object's page using the page allocator's `get_address_page_start` method,
and coerce that address into a `Page` pointer and use `get_bitmap` to get the
bitmap. Effectively this is `O(1)` operation: `get_address_page_start` masks
low bits of the object's location, `get_bitmap` returns the first word of the
page since bitmap pointer is the first word of a page.

Page header layout is defined as:

```rust
pub struct PageHeader {
    // Bitmap used when marking the page for garbage collection
    pub bitmap: Option<Bitmap>,
}
```

If an object lives in heap location `p`, it is always safe to cast
`page_alloc.get_address_page_start(p)` to `*mut PageHeader`.

## Space API

Page allocator API is not concerned with how a page is used. Allocating inside
a page is handled by the `Space` type. `Space` is a concrete type, not a trait,
but it is parametric on `PageAlloc` trait, so it works on IC and also in tests.

`Space` type keeps track of pages that it allocated using the page allocator,
and maintains a "heap pointer" (or "allocation pointer"), which points to the
next free location in the current page.

It also provides the `alloc_words`, `alloc_blob`, and `alloc_array` methods
used by the mutator.

The API:

```rust
impl<P: PageAlloc> Space<P> {
    /// Allocate a new space. Pages of the space will be allocated using the
    /// given page allocator.
    pub unsafe fn new(page_alloc: P) -> Space<P> { ... }

    /// Total allocation done in this page so far. Note that unused space in
    /// pages will be considered as allocated and will be included in this
    /// number.
    pub fn total_alloc(&self) -> usize { ... }

    /// Iterate pages of this space. Used by the mark-compact collector to
    /// initialize (allocate) bitmaps for pages.
    pub fn iter_pages(&self) -> impl Iterator<Item = &P::Page> { ... }

    /// Get the first page in this space. Used by the mark-compact collector
    /// for scanning objects in this space.
    pub fn first_page(&self) -> PageIdx { ... }

    /// Get the last page in this space. Used by the mark-compact collector
    /// for scanning objects in this space.
    pub fn last_page(&self) -> PageIdx { ... }

    /// Get the index of the page we're currently allocating in. This is used
    /// in tests: when scanning a page for sanity we need to know where to stop.
    /// For the current page, we stop at "allocation pointer". For other pages,
    /// we scan the whole page.
    pub fn current_page_idx(&self) -> PageIdx { ... }

    /// Get the page with given index. This method assumes that the page will
    /// be in this space.
    pub fn get_page(&self, idx: PageIdx) -> Option<&P::Page> { ...}

    /// Get the allocation pointer. This is used when scavenging in copying GC
    /// and in tests.
    pub fn allocation_pointer(&self) -> usize { ... }

    /// Allocation entry point. Used by the mutator, and copying GC to
    /// evacuate objects.
    pub unsafe fn alloc_words(&mut self, n: Words<u32>) -> Value { ... }

    /// Free the space. All pages of the space will be freed.
    pub unsafe fn free(&mut self) { ... }

    /// Convenience method for the mutator.
    pub unsafe fn alloc_array(&mut self, len: u32) -> Value { ... }

    /// Convenience method for the mutator.
    pub unsafe fn alloc_blob(&mut self, size: Bytes<u32>) -> Value { ... }

    /// Get the page allocator of this space. Used by the mark-compact
    /// collector to get the page of an object, which is then used to
    /// get the page bitmap and mark the object.
    pub fn page_alloc(&self) -> &P { ... }
}
```

Mutators only use `alloc_array`, `alloc_blob`, and `alloc_words` methods. The
rest is to support garbage collectors and testing.

Allocation is O(1) operation in most cases. When the page is full, we allocate
a new page, which is O(log(N)) operation as analyzed in the previous section
(where N is the number of free pages).

## Mutator interface

Mutator interface is defined in module `motoko_rts::allocation_space`. This module
is only compiled when targeting the IC (not when testing). Summary:

```rust
pub static mut ALLOCATION_SPACE: Space<IcPageAlloc> = ...;

pub unsafe fn init() { ... }

#[no_mangle]
pub unsafe fn alloc_words(n: Words<u32>) -> Value { ... }

#[no_mangle]
pub unsafe fn alloc_blob(size: Bytes<u32>) -> Value { ... }

#[no_mangle]
pub unsafe fn alloc_array(len: u32) -> Value { ... }

// Used by copying GC to free from-space and update allocation space.
pub unsafe fn free_and_update_allocation_space(new: Space<IcPageAlloc>) { ... }
```

## Benchmarks

Current heap layout has no fragmentation, and allocation is always O(1).

With the page-segregated heap, it is not possible to avoid fragmentation, and
allocation will be O(log(N)) when the current page is full, or when allocating
large objects.

The main advantage is it will enable splitting the heap into generations and
allow generational GC.

Since this PR only implements page-segregated heap and not a new GC algorithm
(e.g. generational GC), it will most likely make things worse.

However, perhaps one win with this PR may be large object allocation. Currently
in master branch, all objects are copied, no matter the size. With this PR,
objects larger than 3/4 of a page size have their own pages, and they're not
copied. For programs with many large objects this can potentially improve
cycles.

---

ToDos:

- [ ] Run benchmarks (runtime, binary sizes)
- [ ] Fix scheduling, RTS stats
- [x] Fix compacting GC (sort pages before marking)
- [ ] Implement large object collection in compacting GC
- [x] Use tree-based data structure in free list operations, instead of vectors. `Vec` was used to keep things simple.


osa1:
I'm now trying to enable `alloc` crate and use `wee_alloc` to allocate metadata (bitmaps, mark queues, etc.). I can build the changes locally, but in nix-shell I'm getting

```
error[E0464]: multiple matching crates for `compiler_builtins`
  |
  = note: candidates:

error: aborting due to previous error
```

Would be great to update rustc and get rid of xargo..

osa1:
I managed to reproduce the issue outside of nix-shell. Files:

```
// Cargo.toml
[package]
name = "xargo_test"
version = "0.1.0"
edition = "2018"

// Xargo.toml
[dependencies.alloc]

[dependencies.core]
stage = 0

[dependencies.compiler_builtins]
stage = 1
version = "0.1.47"

// lib.rs
#![no_std]
```

Override toolchain

```
rustup override set nightly
```

Build with xargo:

```
xargo build
```

Output:

```
warning: Patch `rustc-std-workspace-alloc v1.99.0 (/home/omer/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/rustc-std-workspace-alloc)` was not used in the crate graph.
Patch `rustc-std-workspace-std v1.99.0 (/home/omer/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/rustc-std-workspace-std)` was not used in the crate graph.
Check that the patched package version and available features are compatible
with the dependency requirements. If the patch has a different version from
what is locked in the Cargo.lock file, run `cargo update` to use the new
version. This may also occur with an optional dependency that is not enabled.
   Compiling core v0.0.0 (/home/omer/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core)
   Compiling compiler_builtins v0.1.47
   Compiling rustc-std-workspace-core v1.99.0 (/home/omer/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/rustc-std-workspace-core)
   Compiling alloc v0.0.0 (/home/omer/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc)
    Finished release [optimized] target(s) in 8.77s
warning: Patch `rustc-std-workspace-alloc v1.99.0 (/home/omer/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/rustc-std-workspace-alloc)` was not used in the crate graph.
Patch `rustc-std-workspace-std v1.99.0 (/home/omer/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/rustc-std-workspace-std)` was not used in the crate graph.
Check that the patched package version and available features are compatible
with the dependency requirements. If the patch has a different version from
what is locked in the Cargo.lock file, run `cargo update` to use the new
version. This may also occur with an optional dependency that is not enabled.
    Finished release [optimized] target(s) in 0.00s
warning: Patch `rustc-std-workspace-alloc v1.99.0 (/home/omer/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/rustc-std-workspace-alloc)` was not used in the crate graph.
Patch `rustc-std-workspace-core v1.99.0 (/home/omer/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/rustc-std-workspace-core)` was not used in the crate graph.
Patch `rustc-std-workspace-std v1.99.0 (/home/omer/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/rustc-std-workspace-std)` was not used in the crate graph.
Check that the patched package version and available features are compatible
with the dependency requirements. If the patch has a different version from
what is locked in the Cargo.lock file, run `cargo update` to use the new
version. This may also occur with an optional dependency that is not enabled.
   Compiling compiler_builtins v0.1.47
    Finished release [optimized] target(s) in 1.12s
   Compiling xargo_test v0.1.0 (/home/omer/rust/xargo_test)
error[E0464]: multiple matching crates for `compiler_builtins`
  |
  = note: candidates:

error: could not compile `xargo_test` due to previous error
```

osa1:
Existing upstream issue: https://github.com/japaric/xargo/issues/214

It seems like we can't specify compiler_builtins dependency when also depending on alloc.

chenyan-dfinity:
I don't think we can use `wee_alloc` for our platform. There is "memory leak" for upgrade. `wee_alloc` put freed memory into its own free list, and never return them to the system. So if you keep upgrading the same Wasm, eventually replica will be OOM. We have to take out `wee_alloc` from the Rust CDK for this reason.

osa1:
> wee_alloc put freed memory into its own free list, and never return them to the system.

There's no other way to implement an allocator on Wasm because Wasm does not have a way of freeing pages. I don't understand what the issue is. How does the leak occur exactly? I suspect the leak in the CDK was Rust related, e.g. maybe Rust references to heap allocated objects got lost after an upgrade, causing leaks. It would leak even with the standard allocator (the one comes with `std`).

How did you fix the leaks?

chenyan-dfinity:
I don't understand the root cause either. All I know is that by switching to the std allocator, we don't have memory leak. The original message is here: https://dfinity.slack.com/archives/CH4CADCJX/p1615932677024900

To reproduce the leak, we can keep upgrading the same Wasm 50 times, then replica will be OOM.

osa1:
It's very interesting that switching to std allocator fixed the issue (mentioned in the linked thread).

> To reproduce the leak, we can keep upgrading the same Wasm 50 times, then replica will be OOM.

So you install the exact same Wasm? I think this shouldn't cause losing references, except maybe `lazy_static!`s run again after an installation and the old values are lost?

I will keep this issue in mind and test Motoko canister memory usage change with upgrades. I'm sure we can debug this, and either fix the issue (maybe in wee_alloc), or switch to std allocator. We could also implement our own allocator for small objects (on top of the page allocator I'm implementing in this PR).

The linked thread mentions the issue happens only with large blocks (> 64 KiB) so we may not be able to reproduce the issue here though. Currently we only allocate bitmaps for pages. A page is 64 KiB so it needs 16,384 bits, which is 2 KiB. That's how large a bitmap needs to be.

chenyan-dfinity:
I misremembered. It has leaks even with reinstall. Here is the repo: https://github.com/dfinity/sdk/pull/1502/commits/bbfb23f72c5619c11593ec423318d7a077d2ea51

Yeah, that's just be cautious about using `wee_alloc`. I think for large blocks, wee_alloc will just ask for pages from the system, even though there are available space in the free list.

nomeata:
> , and never return them to the system. So if you keep upgrading the same Wasm, eventually replica will be OOM.

That confuses me. 

* Wasm has no way to return memory to the system in the first place. On our platform you can upgrade, which from the wasm pov is just a completely new process.
* Because of that, how can upgrading leak? After an upgrade, the wasm process is completely freshly initialized, it should be the same to the allocator than a new canister.

ulan:
Great work, Ömer! This is moving in the right direction. I have a couple of high-level comments.

1) Wdyt about turning the page allocator into a lower-level concept like chunk/region/rage allocator so that it doesn't know the structure of the page and is only responsible for allocating unstructured address ranges: [start, end)?
It would manage address ranges on 64 KB granularity, so that it is the simplest possible abstraction of the underlying memory system of the platform (Wasm, OS, test). The benefit is that it can be reused for other allocations besides pages and you can share the page-related implementation between different platforms (i.e. Page doesn't need to be a trait).

2) I am not sure that we need to put pages into doubly-linked list. Would it make sense for the corresponding space/generation to own its pages and keep them in a vector?

3) So far we only need to produce a page out of thin air when marking an object. Would it make sense to define an unsafe function for that operation? I believe for all other operations, we can have a proper page reference borrowed from the owner (space or generation).

osa1:
Thanks @ulan for taking a look.

> Wdyt about turning the page allocator into a lower-level concept like chunk/region/rage allocator so that it doesn't know the structure of the page and is only responsible for allocating unstructured address ranges: [start, end)?
> It would manage address ranges on 64 KB granularity, so that it is the simplest possible abstraction of the underlying memory system of the platform (Wasm, OS, test). 

I think current interface very close to this, right? `PageAlloc` doesn't have anything specific to Wasm and does not have a fixed page size. Implementation for IC (`IcPageAlloc`) currently allocates Wasm pages, but that doesn't have to be the case. In the test suite we allocate different sized pages, using Rust's std allocator. (most of the tests are currently disabled, but in bigint tests we use 1 KiB pages) Simiarly in `IcPageAlloc` (`PageAlloc` implementation for IC) we could split Wasm pages into 64 1 KiB pieces.

> The benefit is that it can be reused for other allocations besides pages and you can share the page-related implementation between different platforms (i.e. Page doesn't need to be a trait).

Good point. I think we can remove the trait with a page type that knows its start and end address.

> I am not sure that we need to put pages into doubly-linked list. Would it make sense for the corresponding space/generation to own its pages and keep them in a vector?

Why do you prefer vectors? I think it would work, but each `Space` (and `MarkStack`) will need a vector for their pages, and pages will need to know their indices in the vectors.

> So far we only need to produce a page out of thin air when marking an object. Would it make sense to define an unsafe function for that operation? I believe for all other operations, we can have a proper page reference borrowed from the owner (space or generation).

There are two places where we allocate pages:

1. `Space`: This type is currently used to implement "nursery" for mutator allocations, and "to space" in mark-copy.
2. `MarkStack`: This type is used in mark-compact collector. The stack is implemented as a list of pages.

In both cases we share the page allocator with the types (Space and MarkStack), which then use it to allocate new pages when needed.

What is the concern here exactly?

ulan:
> I think current interface very close to this, right? `PageAlloc` doesn't have anything specific to Wasm and does not have a fixed page size.

Indeed, implementation-wise it is very close. I just wanted to make sure that we don't confuse GC pages (that have headers , marking bitmaps) from the lower-level OS/Wasm pages that are just regions of memory. Ideally we have two different names for these concepts. This distinction becomes important with large objects. There a GC page would consist of multiple low-level pages. We could keep the "page" for the low-level pages and find another name for GC page (maybe span or chunk?).

> Why do you prefer vectors? I think it would work, but each Space (and MarkStack) will need a vector for their pages, and pages will need to know their indices in the vectors.

The main intention is to avoid raw references to pages and to have a single owner for each page (vector is just an example of a container) and use proper rust references. Why would pages need to know their indices? My understand is that whenever we need to remove a page, we are iterating all pages anyway, so we can rebuild the container. 

> In both cases we share the page allocator with the types (Space and MarkStack), which then use it to allocate new pages when needed.

That sounds good. Sharing the page allocator is okay, but I would like to avoid sharing of pages where possible.

> What is the concern here exactly?

The concern is to limit unsafe code and references to a few known places (such as marking an object).

Shall we do a quick video call to discuss the details?

osa1:
> The main intention is to avoid raw references to pages and to have a single owner for each page (vector is just an example of a container) and use proper rust references.

I think this is a good idea, but it comes with some overheads. I'm currently implementing it, will update today.

We can't use `mut` references though as there will be aliasing in e.g. scavenging (where we scan the page we're allocating in) or compaction pass in mark-compact (where the page of the next live object and page we're currently moving objects to will be the same).

An alternative would be to use page indices (usize) instead of page references, but I'm not sure if that's any better than modifying pages with non-mut references, and it will add even more overhead because every read or write will have to index a page vector.

ulan:
> We can't use `mut` references though as there will be aliasing in e.g. scavenging (where we scan the page we're allocating in) or compaction pass in mark-compact (where the page of the next live object and page we're currently moving objects to will be the same).

I agree there. It will be very difficult to maintain the single `mut`-ref invariant. Especially considering that in the future we will need to deal with the mutator modifying objects while incremental GC is in progress. So I think it makes sense to allow modifications with non-mut references. Considering that we are not going to use concurrency, that's not a big loss IMO. That said we can still leverage ownership to manage allocation and freeing of the underlying memory.

> but it comes with some overheads.

I am curious to learn about these (my Rust is still at novice level :))

osa1:
> I am curious to learn about these (my Rust is still at novice level :))

It's not related to Rust, it's just that storing pages in a vector and avoiding raw pointers means an extra indexing in some cases. However I think we can still keep the fast path efficient as we store allocation pointer in `Space` which can be used directly to allocate.

osa1:
I'm currently trying to figure out how to allocate large objects (larger than a page, or some smaller amount that we specify).

When allocating large objects (> page size) we need to allocate consecutive pages, which we currently cannot do. An easy workaround is to use malloc (wee_alloc, or Rust's standard allocator), but because Wasm has no way of freeing pages, those allocators won't free any pages and add them to their free lists instead. This can potentially cause bad utilization when e.g. an app allocates lots of large objects on initialization but never allocates large objects again. The pages allocated for those large objects won't be utilized.

So I'm currently looking into hacking wee_alloc so that it will allocate pages with our page allocator and also release free pages. If x86 version of the library frees pages (`munmap`) then perhaps we can use that version with patched mmap and munmap.

In the worst case we may have to implement our own malloc-style allocator on top of the page allocator.

osa1:
Just to update my comment above: we thought using a malloc-style allocator (e.g. wee_alloc, or Rust's std allocator) won't work great as allocators implemented for Wasm do not release pages (Wasm does not have a way of releasing pages). So if a canister allocates large objects and then those get freed by the gc, the pages for those pages will never be used for small allocations, causing fragmentation.

So I updated the page allocator to allow allocating multiple pages (pages are coalesced when freeing). Large objects will be allocated using the page allocator.

wee_alloc is used to allocate bitmaps and page free lists. We can also use it for mark queues, but currently that's allocated using page allocator. In the future we may also want to allocate remembered sets using wee_alloc.

Next: reserve a bit in object headers for large objects. Those won't be copied, and will have a link field for chaining live large objects. We will use the bit in the header to distinguish large objects from others.

(We can also allocate a bit for a "is_static" field in object headers, but comparing object address with heap base seems to work fine for now)

osa1:
Storing a bit in headers for indicating whether an object is static broke mark-compact collector somehow, but I couldn't figure out why. I'll investigate more.

Copying GC currently works, but large object allocation is not complete yet. Page allocation for large objects is implemented, but the mutator interface rejects large objects for now. Should be easy to implement.

osa1:
> It's worth reviewing the object scan order to make sure "is backwards pointer" test uses the same ordering we use to scan object. In other words, if p1 to p2 is a backwards pointer, then we need to scan p2 before p1 in the second pass.

This currently does not hold. To see why, suppose we have two pages, first one  starts at 0x20000 and the other one starts at 0x10000. The current "is backwards pointer" test is just a pointer comparison, so if we use that we a pointer from e.g. 0x200a4 to 0x10012c will be a backwards pointer and we will thread that pointer in the mark pass. However the second pass will visit object at 0x200a4 before 0x10012c (because its page comes first in the page list), so the algorithm will break.

I see a few options:

- Use a more complicated "is backwards pointer" test that takes the page's index in the page list into account.
- Sort page lists based on start addresses in the second pass, and visit pages in sorted order
- Use a 3-pass mark-compact algorithm (current implementation is the third version, any of the previous two versions should work)

src/codegen/compile.ml
```diff
@@ -1620,7 +1630,7 @@ module BoxedSmallWord = struct
     then BitTagged.tag_const (Int64.of_int (Int32.to_int i))
     else
       E.add_static env StaticBytes.[
-        I32 Tagged.(int_of_tag Bits32);
+        I32 (Int32.logor Tagged.(int_of_tag Bits64) Tagged.static_bit_mask);
```
crusso, line 1633: Should that be Bits32?

osa1: Good catch, thanks. Fixed.

rts/motoko-rts-macros/src/lib.rs
```diff
@@ -6,26 +6,36 @@ use quote::quote;
...
+/// ) -> Value
+/// {
 ///     ...
 /// }
 /// ```
 ///
-/// This functions has a `Memory` parameter to be able to allocate on heap. To compile this
-/// function to use in generated code we need a monomorphic version, without a `Memory` parameter.
-/// This macro generates the monomorphic version. Macro expansion looks like this:
+/// This functions has a `Space` parameter to be able to allocate on heap. To compile this function
```
ggreif, line 19:
```suggestion
/// This function has a `Space` parameter to be able to allocate on heap. To compile this function
```

rts/motoko-rts/src/page_alloc.rs
```diff
@@ -0,0 +1,90 @@
...
+#[cfg(feature = "ic")]
+pub mod ic;
+
+pub const PAGE_SIZE: Bytes<u32> = WASM_PAGE_SIZE;
+
+// Cannot use Div and Mul implementations of Bytes in const context, so using WASM_PAGE_SIZE.
+// LARGE_OBJECT_THRESHOLD = (PAGE_SIZE / 3) * 4
+pub const LARGE_OBJECT_THRESHOLD: Bytes<u32> = Bytes((WASM_PAGE_SIZE.0 / 4) * 3);
+
+/// Trait for page allocators. A page is a unit of allocation from the underlying systme (Wasm, OS,
```
ggreif, line 17:
```suggestion
/// Trait for page allocators. A page is a unit of allocation from the underlying system (Wasm, OS,
```

rts/motoko-rts/src/page_alloc.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// Get the start address of the page that contains the given address. Only works on
+    /// single-page pages! Can be used to get page header.
+    unsafe fn get_address_page_start(&self, addr: usize) -> usize;
+}
+
+/// Trait for allocation units from the underlying system (Wasm, OS, ...). Page state can be held
+/// withing the pages, or externally (e.g. in an array indexed by a page), or a combination of
+/// both.
+pub trait Page: Clone {
+    // TODO FIXME
```
ggreif, line 43: why this?

rts/motoko-rts/src/space.rs
```diff
@@ -0,0 +1,211 @@
...
+    pub unsafe fn alloc_array(&mut self, len: u32) -> Value {
+        // Array payload should not be larger than half of the memory
+        if len > 1 << (32 - 2 - 1) {
+            // 2 for word size, 1 to divide by two
+            rts_trap_with("Array allocation too large");
+        }
+
+        let skewed_ptr = self.alloc_words(size_of::<Array>() + Words(len));
+
+        // NB. Cannot use as_array() here as the header is not written yet
```
ggreif, line 187: Such comments are usually a sign that the API could be improved. See #2998.

osa1:
Currently CanCan benchmark reports +40% instructions in this branch, compared to master. Some ideas to profile this:

- We can compile test suite to a native 32-bit platform (see rts/README.md, "Debugging the RTS" section), and use a tool like Linux's "perf" to generate profile the test suite.

  - We should disable all tests other than GC tests while doing this, to avoid noise. Just commenting out all tests other than GC in `main.rs` should be enough.

  - We can also implement a Rust benchmark program that crates a large heap and runs GC, the same way we create heaps and run GCs in the test suite.

- Implement a wasmtime wrapper, print instructions used after running a program.

  As far as I can see, wasmtime (executable) doesn't have a flag or setting to print number of instructions executed. However, we should be able to write a new executable that uses wasmtime API to run a program, and print instructions executed afterwards.

  (Note: replica currently uses instrumentation to count instructions, but Ulan says wasmtime API could also be used)

- Generate .wasms for the RTS, in master and this branch, and manually compare.

- Run [Motoko playground](https://github.com/dfinity/motoko-playground) locally.

  Motoko playground can generate flamegraphs. We need to run it locally to be able to run it with a different moc. Generate the flamegraphs using master branch for a benchmark, then do the same using this branch.

rts/motoko-rts/src/page_alloc.rs
osa1: This is unused, removed.

PR #2707, opened 2021-08-05
RTS: Introduce heap objects for unused space

osa1:
Two new objects OneWordFiller and FreeSpace introduced to mark unused
spaces in heap.

These are mainly needed in #2706 and for new gc experiments, but they
are also useful to mark the slop when shrinking blobs in principal id
functions.

Previously when shrinking a blob we were filling the unused space with
zeros, and the gc was skipping the zeros when scanning the heap for e.g.
debug prints or sanity checks.

With OneWordFiller and FreeSpace, for 1 and 2 words unused space we do
the same amount of writes (header or header + length), for anything
larger we do 2 writes, instead of one write per word.

### DIFF
--- a/rts/motoko-rts/src/debug.rs
+++ b/rts/motoko-rts/src/debug.rs
@@ -226,6 +226,13 @@ pub(crate) unsafe fn print_boxed_object(buf: &mut WriteBuf, p: usize) {
                 (*concat).text2.0
             );
         }
+        TAG_ONE_WORD_FILLER => {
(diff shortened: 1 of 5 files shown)


ulan (approved):
lgtm, thanks!

The distinction between "filler" and "free_space" will become more clear once we get free lists. The free list entries will get the "free_space" tag and areas that are too small for the free list will be fillers.

rossberg (approved):
Rubberstamp.

PR #2708, opened 2021-08-06
Optimise field access by computing the minimal positional index

ggreif:
Formerly we did a linear search from field № 0 until finding the right hash. But frequently the _positional index_ where we find the hash can be given a static lower bound:
- when we have a supertype approximation of the actual memory object, we can sort the field hashes and anticipate (some of) the layout
- then statically search for the accessed field's hash to obtain the lower bound for starting the linear search.

The only drawback I see is that some (generated template) code is duplicated, by getting new functions for each lower bound. So basically this is a tradeoff. Maybe this optimisation should reside behind a performance flag.
Code duplication also means that we potentially get more cache-cold code on the nodes, but I think that is not reflected in our cost model.

Next move: generate stubs `<lower_bound, hash>` that call a common worker.

--------
## Background

On the surface (Motoko) you index fields by name (which is by hash value in the machine model). Under the hood we have two _parallel arrays_: `[field_hashes_sorted]` and `[field_values]`. See the ASCII art below `module Object` (in `compile.ml`).

Example:
```
| field | hash |
+-------+------+
|  foo  |  789 |
|  bar  |  613 |
|  baz  |  915 |
```
So when you want to access field `(o : { foo : Int, bar : Int, baz : Int}).foo` with `o` in reality having more fields potentially than the (super)type indicates, you'll find that hash array will be `[613, .., 789, .., 915]` because of sorting. We statically know that the lowest _positional index_ for `foo` is 1, since hash `613` already occupies position 0.


crusso:
@nomeata Gabor is trying to optimize record access exploiting sortedness to commence the linear search from a higher index than 0, but we noticed that Motoko uses 31-bit label hashed while Candid uses 32-bit? Do you recall why? We are now worried that we might have some lurking issues in the deserialization code which is using Motoko hashes to search for Candid hashes, which won't always be the same...

But actually, looking at files idlHash.ml and Hash.ml, I think there's a bug in Hash.ml which is using Int32 not UInt32 arithmetic.
Gabor, would that account for your oberserved behaviour?

crusso:
Why is Hash.ml discarding top bit?

nomeata:
> We are now worried that we might have some lurking issues in the deserialization code which is using Motoko hashes to search for Candid hashes, which won't always be the same...

This shouldn’t happen; they are different hashes (just happen to be similar), and one should use one for the other, but always calculate it from the actual label.

> which is using Int32 not UInt32 arithmetic. Gabor, would that account for your obeserved behaviour?

Isn’t that isomorphic if your only operations are addition and multiplication? But not sure at 11pm :)

crusso:
> > We are now worried that we might have some lurking issues in the deserialization code which is using Motoko hashes to search for Candid hashes, which won't always be the same...
> 
> This shouldn’t happen; they are different hashes (just happen to be similar), and one should use one for the other, but always calculate it from the actual label.
> 
> > which is using Int32 not UInt32 arithmetic. Gabor, would that account for your obeserved behaviour?
> 
> Isn’t that isomorphic if your only operations are addition and multiplication? But not sure at 11pm :)

Yeah, the arithmetic is probably fine, you're right, but masking the last bit in one and not the other is not ;->

src/codegen/compile.ml
```diff
@@ -604,6 +604,26 @@ let compile_while cond body =
...
+      ) (
+        mk_body get_i ^^
+
+        get_i ^^
+        compile_add_const 1l ^^
+        set_i
+      )
+
 (* Expects a number on the stack. Iterates from zero to below that number. *)
 let from_0_to_n env mk_body =
```
ggreif, line 628: define this in terms of the above

src/codegen/compile.ml
```diff
@@ -2706,17 +2726,21 @@ module Object = struct
 
 
   (* Returns a pointer to the object field (without following the indirection) *)
-  let idx_hash_raw env =
-    Func.share_code2 env "obj_idx" (("x", I32Type), ("hash", I32Type)) [I32Type] (fun env get_x get_hash ->
+  let idx_hash_raw env m = (* HERE *)
+    let name = match m with
+      | 0 -> "obj_idx"
```
ggreif, line 2731: this is just for backwards "compatibility" (`diff`-wise)

src/codegen/compile.ml
```diff
@@ -2731,33 +2755,63 @@ module Object = struct
             G.i Return
           ) G.nop
       ) ^^
-      E.trap_with env "internal error: object field not found"
+
+      get_h_ptr ^^
```
ggreif, line 2759: remove this (I needed it for debugging my brainfart)

src/codegen/compile.ml
```diff
@@ -2731,33 +2755,63 @@ module Object = struct
...
+            G.i (Store {ty = I32Type; align = 2; offset = 0l; sz = None}) ^^
+      get_h_ptr ^^
+      compile_unboxed_const 9l ^^
+      print_ptr_len env ^^
+      E.trap_with env (Printf.sprintf "internal error: object field not found")
     )
 
   (* Returns a pointer to the object field (possibly following the indirection) *)
-  let idx_hash env indirect =
+  let idx_hash env m indirect =
```
ggreif, line 2773:
```suggestion
  let idx_hash env low_bound indirect =
```

src/codegen/compile.ml
```diff
@@ -3756,7 +3810,7 @@ module Serialization = struct
   let sort_by_hash fs =
     List.sort
       (fun (h1,_) (h2,_) -> Lib.Uint32.compare h1 h2)
-      (List.map (fun f -> (Idllib.Escape.unescape_hash f.Type.lab, f)) fs)
+      (List.map (fun f -> ((match f with Type.{ typ = Typ _; _} -> assert false | _ -> Idllib.Escape.unescape_hash f.Type.lab), f)) fs)
```
ggreif, line 3813: drop this

test/run-drun/idx.mo
```diff
@@ -0,0 +1,27 @@
+actor a {
+    type O1 = { field : Int }; // "\ba\94\93\00"
+    type O2 = { a: Int; field : Int }; // "a\00\00\00\ba\94\93\00"
+    type O3 = { a: Int; field : Int; other : Int }; // "a\00\00\00\ba\94\93\00\d0fv6"
+    type O4 = { a: Int; foo : Int; field : Int; other : Int }; // "a\00\00\00\06\c7M\00\ba\94\93\00\d0fv6"
```
ggreif, line 5: add a type component here just to test the filtering works

src/codegen/compile.ml
ggreif: I'll remove

src/codegen/compile.ml
```diff
@@ -2731,33 +2734,50 @@ module Object = struct
             G.i Return
           ) G.nop
       ) ^^
-      E.trap_with env "internal error: object field not found"
+      E.trap_with env (Printf.sprintf "internal error: object field not found")
```
ggreif, line 2737:
```suggestion
      E.trap_with env "internal error: object field not found"
```

src/codegen/compile.ml
ggreif:
```suggestion
      (List.map (fun f -> (Idllib.Escape.unescape_hash f.Type.lab, f)) fs)
```

test/run-drun/idx.mo
```diff
@@ -0,0 +1,27 @@
...
+
+    func go1(o : O1) : async () = async inner o;
+    func go2(o : O2) : async () = async inner o;
+    func go3(o : O3) : async () = async inner o;
```
crusso, line 9: Why are these async? That seems irrelevant to this test, I think. Make them sync and then you can also remove the awaits below

test/run-drun/idx.mo
```diff
@@ -0,0 +1,27 @@
...
+    public shared func go4(o : O4) : async () { assert o.a == 25; assert o.field == 42; assert o.other == 83; assert o.foo == 8; inner o };
+
+    func inner(o : O1) { assert o.field == 42 };
+
+    public func go() : async () {
+        await go1({ field = 42 }); // field: 9671866
+        await go2({ a = 25; field = 42 });
+        await go3({ a = 25; field = 42; other = 83 });
+        await go4({ a = 25; foo = 8; field = 42; other = 83 })
+    }
```
crusso, line 19: Can you add a test that projects a field from a value of type None (generated with loop `{}`). That would test code generated for impossible values.

src/codegen/compile.ml
```diff
@@ -2735,29 +2738,46 @@ module Object = struct
...
-    else idx_hash_raw env
+    else idx_hash_raw env m
 
   (* Determines whether the field is mutable (and thus needs an indirection) *)
   let is_mut_field env obj_type s =
     let _, fields = Type.as_obj_sub [s] obj_type in
     Type.is_mut (Type.lookup_val_field s fields)
 
+  (* Computes a lower bound for the linear index of a field in an object *)
+  let field_lower_bound env obj_type s =
```
crusso, line 2758: I wonder if you can actually assume the fields are sorted and simplify this code a bit.
Might be worth checking if check_ir enforces sortedness. 


(But if in any doubt, just keeping sorting as is)

crusso:
Did you resolve the mystery of the different hash functions? I'm still concerned that serialization code might be confusing the Motoko and Candid hash functions, and should, instead, be computing the appropriate hash from the label. Ok to leave to that another PR too.

ggreif:
@osa1 @crusso @nomeata I can't assign you as reviewers right now (GitHub won't add the assignments). PTAL. https://github.com/dfinity/motoko/pull/2708

src/codegen/compile.ml
ggreif: Good point, since I took the sorting lines from `Serialization` when I started with this, it might be redundant here! Will test.

src/codegen/compile.ml
```diff
@@ -2735,29 +2738,46 @@ module Object = struct
...
     )
-    else idx_hash_raw env
+    else idx_hash_raw env m
 
   (* Determines whether the field is mutable (and thus needs an indirection) *)
   let is_mut_field env obj_type s =
     let _, fields = Type.as_obj_sub [s] obj_type in
     Type.is_mut (Type.lookup_val_field s fields)
 
+  (* Computes a lower bound for the linear index of a field in an object *)
```
osa1, line 2757: I found "linear index" part a bit confusing. This is computing where to start the linear search for a field, right? I'd say "linear search" instead of "linear index". Maybe "Computes where to start the linear search for a field". 

It would also be good to say here why we can start from an index other than 0, or at least point the reader to the right places for learning about that. (I don't know the answer to this myself)

osa1 (commented):
I don't understand how can we start linear search anywhere other than the start of an object. Could you say a few words about this in the PR description? How are the fields laid out exactly?

src/codegen/compile.ml
ggreif: Well, on the surface (Motoko) you index fields by name (which is by hash value in the machine model). Under the hood we have two _parallel arrays_: `[field_hashes_sorted]` and `[field_values]`. See the ASCII art below `module Object`.

Example:
```
| field | hash |
+-----+-----+
|  foo |  789 |
|  bar | 613 |
|  baz | 915 |
```
So when you want to access field `(o : { foo : Int, bar : Int, baz : Int}).foo` with `o` in reality having more fields potentially than the (super)type indicates, you'll find that hash array will be `[613, .., 789, .., 915]` because of sorting. We statically know that the lowest _linear index_ for `foo` is 1, since hash `613` already occupies position 0.

I could say _positional index_ above. Is that okay with you?

src/codegen/compile.ml
```diff
@@ -2735,29 +2738,46 @@ module Object = struct
...
+  let field_lower_bound env obj_type s =
+    let open Type in
+    let _, fields = as_obj_sub [s] obj_type in
+    let fields = List.filter (function {typ = Typ _; _} -> false | _ -> true) fields in
```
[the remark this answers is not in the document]
ggreif: In the IR the object's type can have _type fields_, which are erased from the memory layout of the object. To not mess up the positional index calculation we'll erase them. A test for this feature is forthcoming.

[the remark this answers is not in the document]
osa1: Thanks for the explanation. I think ideally it would be good to have different views of records, one for type checking, one for code generation (and maybe others for other stuff, like interpretation) would be good. But that kind of thing will require some refactoring..

src/codegen/compile.ml
ggreif: If you look at all the surrounding  code using `idx*` as "finding a field by hash", you'll probably reconsider that in this context :-)

src/codegen/compile.ml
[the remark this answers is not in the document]
ggreif: Turns out that there is a pass on IR that already have removed the type fields when they hit codeine, so yes I'll remove this and assert only.

src/codegen/compile.ml
ggreif: They are sorted by label string! Which is different from hash magnitude order... See the failures in 27438402e.

test/run-drun/idx.mo
ggreif: Due to the mid-end type erasure this is not a real issue.

src/codegen/compile.ml
```diff
@@ -2706,17 +2708,18 @@ module Object = struct
 
 
   (* Returns a pointer to the object field (without following the indirection) *)
-  let idx_hash_raw env =
-    Func.share_code2 env "obj_idx" (("x", I32Type), ("hash", I32Type)) [I32Type] (fun env get_x get_hash ->
+  let idx_hash_raw env m = (* HERE *)
```
nomeata, line 2711:
```suggestion
  let idx_hash_raw env m =
```

src/codegen/compile.ml
```diff
@@ -2706,17 +2708,18 @@ module Object = struct
...
+    let name = Printf.sprintf "obj_idx<%d>" m  in
```
nomeata, line 2712: Hmm, quite wasteful duplicating that code for every possible starting index, when the body of the function really just differs in one consant.

I suggest to write
```
Func.share_code3 env "obj_idx
```
and pass the starting index as the second parameter instead.

src/codegen/compile.ml
```diff
@@ -2735,36 +2738,52 @@ module Object = struct
     )
 
   (* Returns a pointer to the object field (possibly following the indirection) *)
-  let idx_hash env indirect =
+  let idx_hash env m indirect =
     if indirect
-    then Func.share_code2 env "obj_idx_ind" (("x", I32Type), ("hash", I32Type)) [I32Type] (fun env get_x get_hash ->
+    then
+      let name = Printf.sprintf "obj_idx_ind<%d>" m in
+      Func.share_code2 env name (("x", I32Type), ("hash", I32Type)) [I32Type] (fun env get_x get_hash ->
```
nomeata, line 2745: Same here, too much wasm code duplication, I’d say.

src/codegen/compile.ml
```diff
@@ -2735,36 +2738,52 @@ module Object = struct
...
+    let open Type in
+    let _, fields = as_obj_sub [s] obj_type in
+    List.iter (function {typ = Typ _; _} -> assert false | _ -> ()) fields;
+    let sorted_by_hash =
+      List.sort
+        (fun (h1, _) (h2, _) -> Lib.Uint32.compare h1 h2)
+        (List.map (fun f -> Lib.Uint32.of_int32 (E.hash env f.lab), f) fields) in
+    match List.find_opt
+      (fun (_, {lab; _}) -> lab = s)
+      (List.mapi (fun i e -> (i, snd e)) sorted_by_hash) with
```
nomeata, line 2768: Maybe this code would be clearer if you'd have
```
Lib.List.index_of : 'a -> 'a list -> int option
```
then this becomes simply 
```
match Lib.List.index_of s sorted_by_hash with …
```
and the intent is much clearer.
```

File: test/run-drun/idx.mo [hunk 17]
```diff
@@ -0,0 +1,32 @@
+import CO4 "ok/idx-class"
```
nomeata, line 1: Why is this a `run-drun` test? If there is no Internet Computer required, I would suggest to make tests `run` tests whenever possible.

src/codegen/compile.ml
ggreif: Sure, this is a leftover bookmark :-)

src/codegen/compile.ml
ggreif: I totally get what you are saying, but I am afraid I don't agree... I posted this concern in the description. For small Wasm we'd have wasteful duplication, but who cares, we don't cross the message limit. For big programs it is likely that a lot of field indexing is happening, so the 2 bytes of Wasm _for each call site_ will start to become relevant, potentially outweighing the penalty in Wasm coming  from the duplication. Of course we could have short functions for each start index which then (tail-)call the `obj_idx` that you propose. This adds some double-call overhead.

test/run-drun/idx.mo
ggreif: Will try to move. Where should I put dependency modules? Currently I stash it under `ok/`, but that looks like an abomination.

ggreif: Ah, when in `run/` don't need a module any more, I can inline the class instead!

src/codegen/compile.ml
nomeata: Hmm, I see.

Still only gut feeling, but a common function called from everywhere is probably also better for the runtime’s inliner: It’ll be hot code faster.

Also, why stop here? Why not include the `hash` (which is also statically known) in the function name and have static functions for all combinations.

test/run-drun/idx.mo
nomeata: Why did you need  a module in the first place? You can have inline classes just fine

src/codegen/compile.ml
ggreif: 3872bc7b2 is it. I could also unwrap to eliminate the `match`.

src/codegen/compile.ml
ggreif: Next PR :-) Good points about many small stubs.

test/run-drun/idx.mo
ggreif: That's what I mean. Remove the module wrapping.

ggreif: ad403730f

src/codegen/compile.ml
crusso: That's weird, I though I had made the same suggestion as @nomeata about abstracting on the start index to avoid duplication but the comment seems to have disappeared. GH acting strangely or did I forget to commit?

test/run-drun/idx.mo
ggreif: Class with type field added for good measure...

src/codegen/compile.ml
ggreif: Done, description above extended.

test/run-drun/idx.mo
ggreif: Isn't that dead code, so that codegen won't insert the field access, but `unreachable` instead?

src/codegen/compile.ml
ggreif: GH's fault :->

test/run-drun/idx.mo
ggreif: Uhm, I added
``` swift
debug {
    let l : O4 = loop {};
    go4(l)
}
```
at the end, which compiles, but doesn't terminate!

ggreif: Is this what you mean?

src/codegen/compile.ml
ggreif: I am working on a follow-up PR to minimise duplication, based on the ideas and feedback. I prefer to merge this as-is.

test/run-drun/idx.mo
crusso: No, I meant  something more direct like  `(loop {}).foo`, where the unreachable code is directly project from. These kind of examples have bitten us before. I'm sure it'll be fine but worth checking.

crusso:
If you want to verify your hunch about wasm blowup, I think the qr code example use lots of records...

nomeata:
> If you want to verify your hunch about wasm blowup using duplicated vs shared code, I think the qr code example use lots of records...

Another way of thinking about this is: Would you consider implementing this transformation as a separate pass. Consider someone comes along and writes a pass that
 * for each wasm function, checks if one of the function’s argument is always a constant
 * the duplicate the function for each possible value of that constant, specializing it to that constant
I would assume you’d hesitate to add that “optimization”.  So if you’d also hesitate in the general case, then why not in this particular case?

ggreif:
Okay, let's present some numbers:
- `qr-oldmoc.wasm` is `test/perf/qr.mo` compiled by a fresh `master`
- `qr-newmoc.wasm` the same compiled by `gabor/idx` tip of branch (2f83444)
```
[nix-shell:~/motoko]$ ls -l *moc.wasm
-rw-r--r-- 1 ggreif staff 258162 Aug 13 12:19 qr-newmoc.wasm
-rw-r--r-- 1 ggreif staff 257898 Aug 13 12:19 qr-oldmoc.wasm

[nix-shell:~/motoko]$ expr 258162 \* 10000 / 257898
10010
```
Looks like 0.1% size increase to me.

src/codegen/compile.ml
```diff
@@ -2706,17 +2708,18 @@ module Object = struct
...
 
       get_x ^^ Heap.load_field hash_ptr_field ^^ set_h_ptr ^^
 
       get_x ^^ Heap.load_field size_field ^^
       (* Linearly scan through the fields (binary search can come later) *)
-      from_0_to_n env (fun get_i ->
+      from_m_to_n env (Int32.of_int low_bound) (fun get_i ->
         get_i ^^
-        compile_mul_const Heap.word_size  ^^
+        compile_mul_const Heap.word_size ^^
```
crusso, line 2722: Hmm, another comment seems to have been lost by GH.
I think I suggested turning the for loop into a while loop, using pointer addition instead of multplication on each round.
Something like

let cur = htp_ptr + low_bound * n
while (*hp_ptr <> hash) {
   hp_ptr += 4
};
return hp_ptr

ggreif: Yes, I have seen this. Strength reduction would be nice here. Also there is really no need for the end condition, since the hash is necessarily present if our type system is sound :-)

ggreif:
> This PR does not affect the produced WebAssembly code.

@nomeata, this must be some hiccup with the measurement.

nomeata:
> > This PR does not affect the produced WebAssembly code.
> 
> @nomeata, this must be some hiccup with the measurement.

Yes, this is broken; tracked at https://github.com/dfinity/motoko/issues/2631

src/codegen/compile.ml
crusso: Again, that could be left as an optimization to a separate PR.

ggreif:
> I would assume you’d hesitate to add that “optimization”. So if you’d also hesitate in the general case, then why not in this particular case?

@nomeata I would hesitate in the general case, but in a special case there might be additional constraints/specialities that would make me hesitate less. E.g. the number of copies is bounded by the max number of fields any of the objects has:
```
$ wasm2wat qr-newmoc.wasm | grep obj_idx | grep '>' | grep 'func '

  (func $obj_idx<0> (type 11) (param $x i32) (param $hash i32) (result i32)
  (func $obj_idx<1> (type 11) (param $x i32) (param $hash i32) (result i32)
  (func $obj_idx<2> (type 11) (param $x i32) (param $hash i32) (result i32)
```
versus the number of calls to these:
```
$ wasm2wat qr-newmoc.wasm | grep obj_idx | grep '>' | grep 'call ' | wc -l
65
```

src/codegen/compile.ml
ggreif: Oh, just see you say the same thing :-)

crusso:
> > > This PR does not affect the produced WebAssembly code.
> > 
> > 
> > @nomeata, this must be some hiccup with the measurement.
> 
> Yes, this is broken; tracked at #2631. Oh, and a fix has been waiting in #2643 for 5 weeks. Not a great way to treat aspiring external contributers ;-)
> I suggest to merge that one, and then update this PR to see if it helps.

Sorry, that was me asleep at the wheel.

crusso:
> > I would assume you’d hesitate to add that “optimization”. So if you’d also hesitate in the general case, then why not in this particular case?
> 
> @nomeata I would hesitate in the general case, but in a special case there might be additional constraints/specialities that would make me hesitate less. E.g. the number of copies is bounded by the max number of fields any of the objects has:
> 
> ```
> $ wasm2wat qr-newmoc.wasm | grep obj_idx | grep '>' | grep 'func '
> 
>   (func $obj_idx<0> (type 11) (param $x i32) (param $hash i32) (result i32)
>   (func $obj_idx<1> (type 11) (param $x i32) (param $hash i32) (result i32)
>   (func $obj_idx<2> (type 11) (param $x i32) (param $hash i32) (result i32)
> ```
> 
> versus the number of calls to these:
> 
> ```
> $ wasm2wat qr-newmoc.wasm | grep obj_idx | grep '>' | grep 'call ' | wc -l
> 65
> ```

What happens to the size of the binary though? If you have that handy?

src/codegen/compile.ml
crusso: Probably only worth considering when the measurements are back in place (and totally fine as a separate PR).

crusso:
> Okay, let's present some numbers:
> 
> * `qr-oldmoc.wasm` is `test/perf/qr.mo` compiled by a fresh `master`
> * `qr-newmoc.wasm` the same compiled by `gabor/idx` tip of branch ([2f83444](https://github.com/dfinity/motoko/commit/2f83444962756270e708d68185dc617221440122))
> 
> ```
> [nix-shell:~/motoko]$ ls -l *moc.wasm
> -rw-r--r-- 1 ggreif staff 258162 Aug 13 12:19 qr-newmoc.wasm
> -rw-r--r-- 1 ggreif staff 257898 Aug 13 12:19 qr-oldmoc.wasm
> 
> [nix-shell:~/motoko]$ expr 258162 - 257898
> 264
> [nix-shell:~/motoko]$ expr 258162 \* 10000 / 257898
> 10010
> ```
> 
> Looks like 0.1% size increase to me. And @dfinity-ci confirms above :-)

That's comparing master against this version with specialized searches, right?
What if you compare this version with a single shared search function?

nomeata:
I think number wise it doesn't matter greatly either way. The argument that a single hot function is better in general (sooner to be JITed, better cache usage when loading code) stands, but unsubstantiated with numbers. So it remains a question of taste and gut feeling and (maybe) consistency. I am not convinced, but won't veto

src/codegen/compile.ml
crusso: Dooh, sorry about that.

test/run-drun/idx.mo
crusso: If the non-termination is a problem, just put it behind a lambda to test code gen doesn't fail, without actually invoking it in the test.

crusso:
If we ever move to separate compilation, this could will get repeated across modules. If its just a single function, we could move it into the shared runtime system. Just a thought, since we are miles from separate compilation (and will have similar problems with equality and show I guess).

crusso (approved):
I'm ok with this for now - its easy enough to revert.

ggreif:
A follow-up with better spacial characteristics is coming.

ggreif:
> What if you compare this version with a single shared search function?

I don't have that code ready yet. Will report.

test/run-drun/idx.mo
ggreif: PTAL: #2727

src/codegen/compile.ml
ggreif: This is merged as #2722.

PR #2711, opened 2021-08-09
src/Makefile: Create/delete source_id/generated.ml

nomeata:
since #2587, the generated file is `source_id/generated.ml`, and no
longer `source_id/source_id.ml`. The latter is a normal file that `make
clean` should not delete. So let’s tell `make` about this.

(It seems that either nobody runs `make clean`, or everybody expects
weird behaviour and isn’t surprised that `make clean` deletes committed
files.)


crusso (approved):
I rarely do make clean...

PR #2712, opened 2021-08-09
nix-shell: Do not include nix-built candid-tests

nomeata:
so that ./bin/candid-tests is used and incremental development is
possible, just as for the other binaries built “by us” (`moc` etc.)


crusso:
(more of a rubber stamp, but welcome back!)

crusso (approved):

PR #2719, opened 2021-08-12
Use deterministic hashers in RTS test suite

osa1:
By default Rust's HashMap and HashSet hashers are non-deterministic,
which is a problem when debugging as the tests can fail differently in
consecutive runs. I even observed that a buggy test passes when I run
the test suite repeatedly a few times.

This PR switches to `fxhash`s HashMap and HashSet, which is just std map
and set with a deterministic hasher. In the future we should be careful
to not use std maps and sets and always use fxhash.

`maplit` is removed as the `hashmap` and `hashset` macros only allow
generating std maps and sets with non-det hashers.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -167,7 +167,7 @@ rec {
         name = "motoko-rts-deps";
         src = subpath ./rts;
         sourceRoot = "rts/motoko-rts-tests";
-        sha256 = "1vr9mvjrddjv7xga6hhzq39x8qzdqsnhwic76apv7ksfkh0psfx2";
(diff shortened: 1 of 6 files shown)


crusso (approved):
(rubber stamped)

chenyan-dfinity:
If you don't want an extra dependency, you can use `BTreeMap` and `BTreeSet`, which are deterministic.

osa1:
Adding a dependency in the test suite is not a problem, and `fxhash` is tiny (just a hasher and a few type synonyms) so I just added it.

PR #2721, opened 2021-08-12
Fix handling of self references in mark-compact GC, imrpove tests

osa1:
One line fix to thread self references during marking in mark-compact GC.

We had a test with self references but it accidentally passed. The problem was
none of the objects below the object with self reference were dead, so the
object was not moved, and as a result the field with the self reference had the
correct value even though we didn't update it.

The test is updated to add a dead object before the object with self ref. It
fails without the fix.

Test code is also updated to use a vector for the heap description, instead of
a map. This allows specifying object order in heap. With the previous HashMap
based implementation there is no way to reason about the object orders, so for
example the test with a backward pointer might or might not have a backward
pointer, depending on the HashMap (and hasher) implementation. With the vector
representation allocation is done in the same order as the test description,
allowing testing backward and forward pointers.

Thanks to @ulan for the catching the bug.

### DIFF
--- a/rts/motoko-rts/src/gc/mark_compact.rs
+++ b/rts/motoko-rts/src/gc/mark_compact.rs
@@ -211,7 +211,7 @@ unsafe fn update_refs<SetHp: Fn(u32)>(set_hp: SetHp, heap_base: u32) {
 /// Thread forwards pointers in object
 unsafe fn thread_fwd_pointers(obj: *mut Obj, heap_base: u32) {
     visit_pointer_fields(obj, obj.tag(), heap_base as usize, |field_addr| {
-        if (*field_addr).unskew() > field_addr as usize {
+        if (*field_addr).unskew() > obj as usize {
             thread(field_addr)
         }
     });
--- a/rts/motoko-rts/src/gc/mark_compact.rs
+++ b/rts/motoko-rts/src/gc/mark_compact.rs
@@ -149,8 +149,8 @@ unsafe fn mark_fields<M: Memory>(mem: &mut M, obj: *mut Obj, obj_tag: Tag, heap_
         let field_value = *field_addr;
         mark_object(mem, field_value, heap_base);
 
-        // Thread if backwards pointer
-        if field_value.unskew() < obj as usize {
+        // Thread if backwards or self pointer
+        if field_value.unskew() <= obj as usize {
             thread(field_addr);
         }
     });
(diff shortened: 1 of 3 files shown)


rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -149,8 +149,8 @@ unsafe fn mark_fields<M: Memory>(mem: &mut M, obj: *mut Obj, obj_tag: Tag, heap_
         let field_value = *field_addr;
         mark_object(mem, field_value, heap_base);
 
-        // Thread if backwards pointer
-        if field_value.unskew() < obj as usize {
+        // Thread if backwards or self pointer
+        if field_value.unskew() <= obj as usize {
```
ulan, line 153: Would it make sense to use field_addr to be symmetric with the forward thread in line 214?

ulan (approved):
lgtm, thanks for fixing both the bug and the test!

rts/motoko-rts/src/gc/mark_compact.rs
osa1: Good point. I'm not sure if we should make this one like line 214, or make line 214 like this one. Personally I prefer comparing with `obj` instead of field address, as we don't allow internal pointers and I think comparing with `obj` makes this clear. WDYT?

ulan: Comparing with `obj` sounds good as well as long as we are consistent

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -211,7 +211,7 @@ unsafe fn update_refs<SetHp: Fn(u32)>(set_hp: SetHp, heap_base: u32) {
 /// Thread forwards pointers in object
 unsafe fn thread_fwd_pointers(obj: *mut Obj, heap_base: u32) {
     visit_pointer_fields(obj, obj.tag(), heap_base as usize, |field_addr| {
-        if (*field_addr).unskew() > field_addr as usize {
+        if (*field_addr).unskew() > obj as usize {
```
osa1, line 214: Refactored for consistency with the comparison above

rts/motoko-rts/src/gc/mark_compact.rs
crusso: Even better, a 1-character fix!

PR #2722, opened 2021-08-13
Improve on #2708 by applying strength reduction and abolishing termination checks

ggreif:
Pick up an idea to reduce the computational complexity of the loop searching for the field hash:
- don't use multiplication, use addition instead
- trust the type system that the hash is present, and scrap the termination check

This gives us a nice 0.5% reduction in benchmarks for cycles usage as well as shaving off the 0.1% size increase that we suffered with #2708.

------------------
### Measurements

this gives us a comparison in binary size for 65 calls:
```
[nix-shell:~/motoko]$ wasm2wat qr-sharemoc.wasm | grep obj_idx | grep 'call ' | wc -l
65
```
there is a single `func $obj_idx`
```
[nix-shell:~/motoko]$ ls -l *moc.wasm
-rw-r--r-- 1 ggreif staff 257898 Aug 13 12:19 qr-oldmoc.wasm <<< previous `master`
-rw-r--r-- 1 ggreif staff 258162 Aug 13 12:19 qr-newmoc.wasm <<< current `master`
-rw-r--r-- 1 ggreif staff 258039 Aug 13 14:36 qr-sharemoc.wasm  <<< this commit (2569509)
-rw-r--r-- 1 ggreif staff 258027 Aug 17 09:37 qr-shavemoc.wasm <<< rearchitected loop
```
Looks like calling a nullary wrapper that calls the bottleneck with pushed bound will improve things (worth exploring, but not done in this PR).

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -2711,12 +2711,26 @@ module Object = struct
   let idx_hash_raw env low_bound =
     let name = Printf.sprintf "obj_idx<%d>" low_bound  in
     Func.share_code2 env name (("x", I32Type), ("hash", I32Type)) [I32Type] (fun env get_x get_hash ->
+      let set_x = G.i (LocalSet (nr 0l)) in
       let (set_h_ptr, get_h_ptr) = new_local env "h_ptr" in
 
       get_x ^^ Heap.load_field hash_ptr_field ^^ set_h_ptr ^^
 
-      get_x ^^ Heap.load_field size_field ^^
+        (*get_x ^^ Heap.load_field size_field ^^*)
       (* Linearly scan through the fields (binary search can come later) *)
+        (* unskew h_ptr and advance both to low bound *)
+        get_h_ptr ^^ compile_add_const Int32.(of_int (1 + low_bound * to_int Heap.word_size)) ^^ set_h_ptr ^^
+          get_x ^^ compile_add_const Int32.(of_int ((3 + low_bound) * to_int Heap.word_size)) ^^ set_x ^^
+    G.loop_ [] (
+        get_h_ptr ^^ load_unskewed_ptr ^^
+        get_hash ^^ G.i (Compare (Wasm.Values.I32 I32Op.Eq)) ^^
+        G.if_ [] (get_x ^^ G.i Return)
+          (get_h_ptr ^^ compile_add_const Heap.word_size ^^ set_h_ptr ^^
+             get_x ^^ compile_add_const Heap.word_size ^^ set_x
+          ^^ G.i (Br (nr 1l)))
+    ) ^^ G.i Unreachable
+
+(*
       from_m_to_n env (Int32.of_int low_bound) (fun get_i ->
         get_i ^^
         compile_mul_const Heap.word_size ^^
@@ -2735,6 +2749,8 @@ module Object = struct
           ) G.nop
       ) ^^
       E.trap_with env "internal error: object field not found"
+ *)
+
     )
 
   (* Returns a pointer to the object field (possibly following the indirection) *)
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -2712,14 +2712,17 @@ module Object = struct
     let name = Printf.sprintf "obj_idx<%d>" low_bound  in
     Func.share_code2 env name (("x", I32Type), ("hash", I32Type)) [I32Type] (fun env get_x get_hash ->
       let set_x = G.setter_for get_x in
-      let (set_h_ptr, get_h_ptr) = new_local env "h_ptr" in
+      let set_h_ptr, get_h_ptr = new_local env "h_ptr" in
 
       get_x ^^ Heap.load_field hash_ptr_field ^^
 
       (* Linearly scan through the fields (binary search can come later) *)
       (* unskew h_ptr and advance both to low bound *)
-      compile_add_const Int32.(of_int (1 + low_bound * to_int Heap.word_size)) ^^ set_h_ptr ^^
-      get_x ^^ compile_add_const Int32.(of_int ((3 + low_bound) * to_int Heap.word_size)) ^^ set_x ^^
+      compile_add_const Int32.(add ptr_unskew (mul Heap.word_size (of_int low_bound))) ^^
+      set_h_ptr ^^
+      get_x ^^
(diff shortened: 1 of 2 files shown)


src/codegen/compile.ml
```diff
@@ -2711,30 +2711,25 @@ module Object = struct
...
+      (* unskew h_ptr and advance both to low bound *)
+      compile_add_const Int32.(of_int (1 + low_bound * to_int Heap.word_size)) ^^ set_h_ptr ^^
+      get_x ^^ compile_add_const Int32.(of_int ((3 + low_bound) * to_int Heap.word_size)) ^^ set_x ^^
+      G.loop_ [] (
+          get_h_ptr ^^ load_unskewed_ptr ^^
+          get_hash ^^ G.i (Compare (Wasm.Values.I32 I32Op.Eq)) ^^
+          G.if_ []
+            (get_x ^^ G.i Return)
+            (get_h_ptr ^^ compile_add_const Heap.word_size ^^ set_h_ptr ^^
+             get_x ^^ compile_add_const Heap.word_size ^^ set_x ^^
```
crusso, line 2729: Is there a reason to advance the both ptrs in lock step rather than just do one subtract and add at the end? I guess you are hoping that the fast path is to not iterate at all, and this is then better?
Anyway, this is nice an clear, just asking.

crusso: (I'd change the title of the PR to reflect what it's now doing)

ggreif: Yes, the reason is that `diff`-`add` is penalising the _fast path_ when the speculation hits. My gut feeling says also that there are less data dependencies this way. But I'll try to code up the other solution too, and benchmark.

src/codegen/compile.ml
```diff
@@ -2711,30 +2711,25 @@ module Object = struct
   let idx_hash_raw env low_bound =
     let name = Printf.sprintf "obj_idx<%d>" low_bound  in
     Func.share_code2 env name (("x", I32Type), ("hash", I32Type)) [I32Type] (fun env get_x get_hash ->
+      let set_x = G.i (LocalSet (nr 0l)) in
```
ggreif, line 2714: this should be something like `let set_x = setter_for get_x`

src/codegen/compile.ml
```diff
@@ -2711,30 +2711,25 @@ module Object = struct
...
-            compile_mul_const Heap.word_size ^^
-            get_x ^^
-            G.i (Binary (Wasm.Values.I32 I32Op.Add)) ^^
-            G.i Return
-          ) G.nop
-      ) ^^
-      E.trap_with env "internal error: object field not found"
+      (* unskew h_ptr and advance both to low bound *)
+      compile_add_const Int32.(of_int (1 + low_bound * to_int Heap.word_size)) ^^ set_h_ptr ^^
+      get_x ^^ compile_add_const Int32.(of_int ((3 + low_bound) * to_int Heap.word_size)) ^^ set_x ^^
```
ggreif, line 2722: `s/3/header_size/1` or such

src/codegen/compile.ml
```diff
@@ -2711,30 +2711,25 @@ module Object = struct
...
+      G.loop_ [] (
+          get_h_ptr ^^ load_unskewed_ptr ^^
+          get_hash ^^ G.i (Compare (Wasm.Values.I32 I32Op.Eq)) ^^
+          G.if_ []
+            (get_x ^^ G.i Return)
+            (get_h_ptr ^^ compile_add_const Heap.word_size ^^ set_h_ptr ^^
+             get_x ^^ compile_add_const Heap.word_size ^^ set_x ^^
+             G.i (Br (nr 1l)))
```
ggreif, line 2730: All other examples in the internets say this should be `0`, but we seem to apply a numbering beginning from `1`.

src/codegen/compile.ml
ggreif: applicable elsewhere too

ggreif: Done in e3dbcd6bf07a62f178a0a285ac3a055dc2ee8eb1.

src/codegen/compile.ml
ggreif: Done in 0317bd9662df98cfbc2254a169a818015f173321.

src/codegen/compile.ml
```diff
@@ -1516,8 +1516,8 @@ module Word64 = struct
     let name = prim_fun_name Type.Nat64 "wpow_nat" in
     Func.share_code2 env name (("n", I64Type), ("exp", I64Type)) [I64Type]
       (fun env get_n get_exp ->
-        let set_n = G.i (LocalSet (nr 0l)) in
-        let set_exp = G.i (LocalSet (nr 1l)) in
+        let set_n = G.setter_for get_n in
```
crusso, line 1519: Is it really worth doing this rather than just creating a new local? I would (perhaps naively) expect the jit to optimize the local away and identify with the parameter.

If being able to assign to parameters is generally useful, I wonder if we should adjust the arguments to shared_codeX to be setter/getter pairs in general.

src/codegen/compile.ml
crusso: Did you resolve this?

src/codegen/compile.ml
crusso: Ah, I see that .setter_for also cleans up some existing code (serialization below); got for it.

crusso (approved):
LGTM

src/codegen/compile.ml
ggreif: I didn't, but the loop I am replacing is doing the same thing and since it passes the validator, it must  be correct, right ;-)
It was more like a bait for @nomeata, I accept it as-is.

src/codegen/compile.ml
ggreif: Exactly, I prefer having immutable parameters, but here I see justification to mutate. The extra local would inflate instruction count and probably cycles balances too, since instrumenting counts instructions too.

src/codegen/compile.ml
ggreif: Mystery resolved: The `if` also establishes an exit label, this one gets assigned `0` in good old DeBruijn fashion. So if we want to reenter the `loop`, we must jump to label `1`.

PR #2723, opened 2021-08-13
[test] StableMemory: Game of Life using StableMemory API

crusso:
This is a WIP attempt to exercise the stable memory API but extending the life-mut test with another version using stable memory to store the generations of life.

~While the initial upgrade to the new version seems to work and preserved state, I'm still seeing differences in the output that I can't explain.~ It turns out the original test was broken an using  shared state for the grids that, now mutable, should have been kept separate. life-v1 was ok by accident (it did an additional allocation of the state in the Grid constructor), life-v2.mo was broken an used the same backing area for the two grid buffers (only spotted by visualizing the output). life-v3.mo was correct a along.
 
The test output was generated using an additional install sequence of life-v2, to set output expections, later replace by life-v3.mo, trying to match the same output.

I found an insidious bug in #2626 in the process.

A sample should be much cleaner than this and rewritten based on the code in the dfinity/examples repro.


test/run-drun/life-mut/life-v3.mo
```diff
@@ -0,0 +1,206 @@
...
+    assert (readBit(offset, index) == v);
+  };
+
+  type Cell = Bool;
+
+  type State = {
+    #v1 : [[var Cell]];
+    #v2 : {size : Nat; bits : [var Nat64]};
+    #v3 : {size : Nat; offset : Nat32}
+  };
```
matthewhammer, line 50: @crusso Just reading this carefully now.  Very interesting.  But, to sustain this pattern indefinitely, the (static) variant type's representation grows without bound, correct?  (with ever more variants)

That seems intuitively unavoidable to me, but does seem to create a tension with the desire not to grow an unbounded history of past representations within the current one (paraphrasing what I imagine @rossberg to mean when we talk about "not storing full history" in our stable memory story).

WDYT, out of curiosity?

rossberg: Actually, I believe you could remove old tags in the type just fine once all canisters running this code have been successfully upgraded.

In any case, this is linear in the _version_ history, which I assume is several orders of magnitude smaller than the call/data history that we need to worry about wrt scalability.

matthewhammer: > Actually, I believe you could remove old tags in the type just fine once all canisters running this code have been successfully upgraded.

I see, good point.  Notably, that's a little different than with a pubic API and our Candid upgrade story, where we must assume that some old clients always lurk out in the wild, right?

> In any case, this is linear in the version history, which I assume is several orders of magnitude smaller than the call/data history that we need to worry about wrt scalability.

Yes, absolutely.  My original thought a concern about the *concept* of growing "static history", not the practice of it, which I agree is not concerning.

rossberg: >Notably, that's a little different than with a pubic API and our Candid upgrade story, where we must assume that some old clients always lurk out in the wild, right?

Yes, exactly. For the public Candid interface, you generally do not know the set of all clients. For the private stable interface, your future self is the only client.

crusso:
Oops - completely missed the discussion here - GH never gives me good notification anymore, at least,not where I can see them

PR #2724, opened 2021-08-16
Support building RTS to native for better debug experience

osa1:
Currently it's a bit hacky: there are 3 commands to run, but the last
command does not work in nix-shell.


crusso (approved):
Rubber stamping. Don't know how to fix the nix issue either....

PR #2726, opened 2021-08-17
nix: create dedicated ic-ref-run derivation

nomeata:
this breaks the run-time dependency from our tests via `ic-ref-run` to things like Haskell library files
and documentation, by extracting just the relevant executable from the haskell package build, and should
speed up builds an `nix-shell` by downloading less stuff.


ggreif (approved):
Reducing closure size for the win!

PR #2727, opened 2021-08-17
Add tests to check for compilation of bottom (as object)

ggreif:
This is a follow-on to #2708, addressing a [review comment](https://github.com/dfinity/motoko/pull/2708#discussion_r686132804).

The resulting Wasm is as expected:
``` wasm
  (func $absurdities (type 8) (param $clos i32) (param $inp i32)
    local.get $inp
    i32.const 2
    call $B_eq
    if  ;; label = @1
      loop  ;; label = @2
        br 0 (;@2;)
      end
      unreachable
    end
...
```
the rest is analogous.

### DIFF
--- a/test/run/idx.mo
+++ b/test/run/idx.mo
@@ -34,6 +34,11 @@ func absurdities(inp : Int) {
     if (inp == 2) go2(loop {});
     if (inp == 3) go3(loop {});
     if (inp == 4) go4(loop {});
+
+    if (inp == 1) (loop {}).field;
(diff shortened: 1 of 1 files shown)


ggreif:
I wonder if a `FileCheck` test would be nice here.

crusso:
Looks good, I think checking the compiler doesn't fall over is good enough. We've had few cases where this hit us before.

crusso (approved):

PR #2733, opened 2021-08-18
Implement `Int` ⇔ `Float` conversion as primitives

ggreif:
See https://forum.dfinity.org/t/motokos-type-system-converting-between-types.

The language-level change in this PR is that we add two new functions to `prim.mo`:
- `floatToInt : Float -> Int`
- `intToFloat : Int -> Float`

`floatToInt` cannot fail, at least in theory. In the interpreter I have seen failures for `floatToInt(1.7e308)`, while the compiler works well. Tests go up to `1e100`. And then there are `inf` and `NaN`, which are not even tested.

If `intToFloat` overflows, the result is `(+/-)inf`.

Of course these conversions are lossy in general. Caveat emptor!

### Base library implications

Let's merge this PR after soaking a few days/weeks.

Then module `Float.fromInt` would call the new primitives instead of going the `Int` -> `Int64` -> `Float` route (which does trap for big numbers). The fix need to go into the `motoko-base` lib at a later stage. This is the place:
``` swift
  /// Conversion from Int via Int64. May trap.
  public let fromInt : Int -> Float =
    func (x : Int) : Float = fromInt64(Prim.intToInt64(x));
```


rts/motoko-rts/src/bigint.rs
```diff
@@ -246,6 +246,26 @@ unsafe extern "C" fn bigint_of_int64(j: i64) -> SkewedPtr {
     persist_bigint(i)
 }
 
+#[cfg(feature = "ic")]
+#[no_mangle]
+unsafe extern "C" fn bigint_of_float64(j: f64) -> SkewedPtr {
+    let mut i = tmp_bigint();
+    //mp_set_i64(&mut i, 3);
+    mp_set_double(&mut i, j);
+    persist_bigint(i)
```
ggreif, line 255: perform the _small bignum_ optimisation first

rts/motoko-rts/src/bigint.rs
```diff
@@ -246,6 +246,25 @@ unsafe extern "C" fn bigint_of_int64(j: i64) -> SkewedPtr {
     persist_bigint(i)
 }
 
+#[cfg(feature = "ic")]
+#[no_mangle]
+unsafe extern "C" fn bigint_of_float64(j: f64) -> SkewedPtr {
+    let mut i = tmp_bigint();
+    mp_set_double(&mut i, j);
```
ggreif, line 253: check result here

src/mo_values/prim.ml
```diff
@@ -52,17 +52,21 @@ Wrapping numeric conversions are all specified uniformly by going through bigint
 (* Trapping conversions (the num_conv_t1_t2 prim used in prelude/prelude.ml) *)
 let num_conv_trap_prim t1 t2 =
   let module T = Type in
-  match (t1, t2) with
-  | T.Nat, (T.Nat8|T.Nat16|T.Nat32|T.Nat64)
-  | T.Int, (T.Int8|T.Int16|T.Int32|T.Int64)
-  | (T.Nat8|T.Nat16|T.Nat32|T.Nat64), T.Nat
-  | (T.Int8|T.Int16|T.Int32|T.Int64), T.Int
+  match t1, t2 with
+  | T.Nat, T.(T.Nat8|Nat16|Nat32|Nat64)
```
ggreif, line 56: `T.Nat8` is  necessary to suppress a warning.

src/mo_values/prim.ml
```diff
@@ -52,17 +52,21 @@ Wrapping numeric conversions are all specified uniformly by going through bigint
...
+  | T.(Nat8|Nat16|Nat32|Nat64), T.Nat
+  | T.(Int8|Int16|Int32|Int64), T.Int
   | T.Nat32, T.Char
   -> fun v -> of_big_int_trap t2 (as_big_int t1 v)
 
   | T.Float, T.Int64 -> fun v -> Int64 (Int_64.of_big_int (Big_int.big_int_of_int64 (Wasm.I64_convert.trunc_f64_s (as_float v))))
   | T.Int64, T.Float -> fun v -> Float (Wasm.F64_convert.convert_i64_s (Big_int.int64_of_big_int (Int_64.to_big_int (as_int64 v))))
-  | t1, t2 -> raise (Invalid_argument ("Value.num_conv_trap_prim: " ^ T.string_of_typ (T.Prim t1) ^ T.string_of_typ (T.Prim t2) ))
+
+  | T.Float, T.Int -> fun v -> Int (Int.of_big_int (Big_int.big_int_of_int64 (Wasm.I64_convert.trunc_f64_s (as_float v)))) (* we could divide-and-conquer *)
```
ggreif, line 66: Using `Wasm.I64_convert.trunc_f64_s` is not exactly correct (as in the `Int64` case), but enough for the interpreter IMHO.

test/run/float-ops.mo
```diff
@@ -90,3 +90,19 @@ assert (Prim.floatToInt64(-pi) == (-3 : Int64));
 
 assert (Prim.int64ToFloat(42) == 42.0);
 assert (Prim.int64ToFloat(-42) == -42.0);
+
+Prim.debugPrint (debug_show Prim.floatToInt(-pi));
```
ggreif, line 94: clean up

test/run/float-ops.mo
```diff
@@ -90,3 +90,19 @@ assert (Prim.floatToInt64(-pi) == (-3 : Int64));
...
+Prim.debugPrint (debug_show Prim.intToFloat(-42_000_000_000));*/
+
+assert (Prim.floatToInt(pi) == 3);
+assert (Prim.floatToInt(-pi) == -3);
+
+assert (Prim.intToFloat(42) == 42.0);
+assert (Prim.intToFloat(-42) == -42.0);
+
+
+assert (Prim.intToFloat(42_000_000_000) == 42000000000.0);
```
ggreif, line 107: test the compact-to-heap bignum boundary (add Rust debug assertions).

rts/Makefile
```diff
@@ -45,6 +46,7 @@ TOMMATH_FLAGS = \
   -DMP_MEMSET=0 \
   -DMP_FIXED_CUTOFFS \
   -DMP_NO_FILE \
+  -D__STDC_IEC_559__ \
```
ggreif, line 49: @rossberg is this a reasonable thing to define? I am not terribly informed about the float format employed in Wasm.

ggreif: Background: without this, `mp_set_double` is not defined by `LibTomMath`.

rts/motoko-rts/src/bigint.rs
```diff
@@ -246,6 +246,25 @@ unsafe extern "C" fn bigint_of_int64(j: i64) -> SkewedPtr {
...
+    persist_bigint(i)
+}
+
+#[cfg(feature = "ic")]
+#[no_mangle]
+unsafe extern "C" fn bigint_to_float64(p: SkewedPtr) -> f64 {
```
ggreif, line 259: is `unsafe` needed?

ggreif: Yes, since I am calling another `unsafe` function. Is transitivity a curse of a blessing? :-)

rts/motoko-rts/src/bigint.rs
```diff
@@ -246,6 +246,25 @@ unsafe extern "C" fn bigint_of_int64(j: i64) -> SkewedPtr {
...
+    if p.is_tagged_scalar() {
+        (p.0 as i32 >> 1) as f64
```
ggreif, line 261: make a method for this: `as_tiny()`

rts/motoko-rts/src/bigint.rs
ggreif: Done: f3c7454bc.

test/run/float-ops.mo
ggreif: Done with f3c7454bc.

ggreif: In Rust we can add a debugging assert to `persist_bigint`, which checks that heap allocated bignums are never tiny.

rts/motoko-rts/src/bigint.rs
```diff
@@ -246,6 +246,28 @@ unsafe extern "C" fn bigint_of_int64(j: i64) -> SkewedPtr {
     persist_bigint(i)
 }
 
+#[cfg(feature = "ic")]
+#[no_mangle]
+unsafe extern "C" fn bigint_of_float64(j: f64) -> SkewedPtr {
+    if j < 1073741824.0 && j > -1073741825.0 {
+        return SkewedPtr(((j as i32) << 1) as usize);
```
nomeata, line 253: The `as` operator is a rounding conversion, not just a type cast, right? Does it round the right and same way as the libtommath code? In particular at .5?

src/mo_values/prim.ml
nomeata: Turn that into a code comment?

nomeata (approved):
Phone review, but looks reasonable

rts/motoko-rts/src/bigint.rs
ggreif: The tests have fractional parts, and the various technologies coincide, so I am hopeful (pretty certain) that `as` does the right thing!

ggreif: Checked: https://stackoverflow.com/questions/37506672/convert-float-to-integer-in-rust

rts/motoko-rts/src/bigint.rs
ggreif: Done: 94fb93dac2f0c611dd239972380d2370726e7222!

rts/Makefile
osa1: I think we could just define `MP_HAS_SET_DOUBLE`, no? See https://github.com/libtom/libtommath/blob/44ee82cd34d0524c171ffd0da70f83bba919aa38/tommath_private.h#L167-L172

osa1 (commented):
So the language-level change in this PR is we add two new functions to prim:

- floatToInt : Float -> Int
- intToFloat : Int -> Float

I'm guessing `floatToInt` is cannot fail, right? What happens if `intToFloat` overflows?

Would be good to mention language-level changes in the PR description.

As evidenced by `SkewedPtr::of_tiny` which is creating a scalar value and not a skewed pointer, `SkewedPtr` is a bit poorly named. I'm also having problems with it in my PRs and I was considering changing it. I think we should rename it to something like `struct FieldValue(u32)`. We could then have view method `fn get_value(&self) -> PtrOrScalar` with `enum PtrOrScalar { Ptr(usize), Scalar(u32) }`. Not in this PR of course. @ggreif WDYT?

rts/Makefile
osa1: > is this a reasonable thing to define? I am not terribly informed about the float format employed in Wasm.

Wasm uses IEEE754: https://webassembly.github.io/spec/core/syntax/types.html#number-types

src/mo_values/prim.ml
osa1: If `T.Nat8` is not supposed to be handled but only here to suppress a warning I think it would be better to turn that case into a panic.

rts/motoko-rts/src/bigint.rs
osa1: > is unsafe needed?

Good question. `unsafe` in this context means "memory safety not guaranteed". Dereferencing a raw pointer is unsafe because raw pointers are not borrow checked so you can read e.g. freed memory, or memory being accessed in another thread etc. We deal with raw pointers a lot in the RTS, it was easier to mark most functions as "unsafe". But if there are cases where any potential unsafety is checked (for example, by bounds checking in array access functions) then we can remove "unsafe" from the function type and use an unsafe block in the function body.

In the context of language runtimes, perhaps "safe if not buggy" is safe enough and we can move the `unsafe` from function types to function bodies, I'm not sure. `unsafe` does not affect code gen, it's just a signal to the programmer.

src/mo_values/prim.ml
ggreif: It is supposed to be there, I don't want to change the semantics. It is purely about type ambiguity. There are two constructors in different types called `Nat8`, and the compiler can disambiguate them, but warns. This kind-of hits it on the nose. No idea why the `T.(...)` scope extension won't give `T.Nat8` a preferred status.

rts/Makefile
ggreif: I didn't know about `MP_HAS_SET_DOUBLE`, thanks!

ggreif: Done: 89189ed5d.

rts/motoko-rts/src/bigint.rs
ggreif: Well, it works this way, so I'll leave it as-is. I was just wondering, and missing the intuition. I am fine now.

ggreif:
I am testing the the overflow cases and see strange things. For later investigation...
``` swift
Prim.debugPrint(debug_show(Prim.intToFloat(17 * 10 ** 307)));
// Adding the ^^ above line causes:
/* +16999999999999999388307957886599817433334607430407587450277311919353772917816056586433009178758470798857226246798318891bigint function errorinternal error: unexpected state entering Idle)pattern failednegative powerassertion failed at float-ops.mo:118.1-118.60
*/
```

test/run/float-ops.mo
ggreif: I'll leave this to another PR.

rts/Makefile
ggreif: I had to reopen, as tests failed. The commit had to be reverted.

ggreif:
I agree that `SkewedPtr` is the wrong name for the beast. But we know what it is :-) Not urgent to change.

My suggestion list is {`VanillaValue`, `VanillaRepresentation`}, in this order. We use the term _vanilla_ a lot in `compile.ml`.

rts/Makefile
ggreif: If `__STDC_IEC_559__` is only tested in `mp_set_double.c` then we _should_ be okay, right?

osa1: > If __STDC_IEC_559__ is only tested in mp_set_double.c then we should be okay, right?


I'm a bit confused, why do you want `__STDC_IEC_559__`? Do the tests pass with `__STDC_IEC_559__` but not `MP_HAS_SET_DOUBLE`? I don't think that's possible, `STDC_IEC_559` is only used to set `MP_HAS_SET_DOUBLE`.

osa1: @ggreif OK, figured it out. The version I linked above is quite different than the version we're using. The version we're using needs `__STDC_IEC_559__`:

```
#if defined(__STDC_IEC_559__) || defined(__GCC_IEC_559)
mp_err mp_set_double(mp_int *a, double b) ...
```

If the code assumes IEC 559 then it won't work even if we enable it.

osa1: I compared the implementations and they seem identical:

```diff
~ $ diff -u /nix/store/19iw4ws6xhqla72blr8w1g9x11gd4rpd-libtommath-src/bn_mp_set_double.c opt/libtommath/mp_set_double.c
--- /nix/store/19iw4ws6xhqla72blr8w1g9x11gd4rpd-libtommath-src/bn_mp_set_double.c       1970-01-01 02:00:01.000000000 +0200
+++ opt/libtommath/mp_set_double.c      2021-01-02 11:47:36.755469002 +0300
@@ -1,9 +1,9 @@
 #include "tommath_private.h"
-#ifdef BN_MP_SET_DOUBLE_C
+#ifdef MP_SET_DOUBLE_C
 /* LibTomMath, multiple-precision integer library -- Tom St Denis */
 /* SPDX-License-Identifier: Unlicense */

-#if defined(__STDC_IEC_559__) || defined(__GCC_IEC_559)
+#if defined(MP_HAS_SET_DOUBLE)
 mp_err mp_set_double(mp_int *a, double b)
 {
    uint64_t frac;
@@ -16,7 +16,7 @@
    cast.dbl = b;

    exp = (int)((unsigned)(cast.bits >> 52) & 0x7FFu);
-   frac = (cast.bits & ((1uLL << 52) - 1uLL)) | (1uLL << 52);
+   frac = (cast.bits & (((uint64_t)1 << 52) - (uint64_t)1)) | ((uint64_t)1 << 52);

    if (exp == 0x7FF) { /* +-inf, NaN */
       return MP_VAL;
@@ -30,7 +30,7 @@
       return err;
    }

-   if (((cast.bits >> 63) != 0uLL) && !MP_IS_ZERO(a)) {
+   if (((cast.bits >> 63) != 0u) && !mp_iszero(a)) {
       a->sign = MP_NEG;
    }
```

Without knowing anything about IEC 559, I'm guessing since the implementations are identical and `__STDC_IEC_559__` is not used elsewhere in the version we're using, it should be OK to define `__STDC_IEC_559__`. Alternatively we could update libtommath and define `MP_HAS_SET_DOUBLE`.

The documentation does not mention any of these macros, I'm guessing they define those in the configure script.

osa1:
> In the interpreter I have seen failures for floatToInt(1.7e308), while the compiler works well

Is this not a problem? Shouldn't they agree?

rts/Makefile
ggreif: Yes, I guessed so. Maybe I'll write an issue, so we don't commit errors when upgrading the `libTomMath`.

ggreif:
> > In the interpreter I have seen failures for floatToInt(1.7e308), while the compiler works well
> 
> Is this not a problem? Shouldn't they agree?

@osa1 nope I don't think it is a problem, but it is annoying when it happens in tests, as they start spewing out weird stuff. I think this is some infelicity with OCaml's libs.

ggreif:
> > I am testing the the overflow cases and see strange things. For later investigation...
> > ```swift
> > Prim.debugPrint(debug_show(Prim.intToFloat(17 * 10 ** 307)));
> > // Adding the ^^ above line causes (in compilation mode):
> > /* +16999999999999999388307957886599817433334607430407587450277311919353772917816056586433009178758470798857226246798318891
> > bigint function errorinternal error: unexpected state entering Idle)pattern failednegative powerassertion failed at float-ops.mo:118.1-118.60
> > */
> > ```
> 
> Looks like `debug_show` uses `float_fmt` which by default prints the float in full into a buffer of size 120, and in case of overflow it prints whatever happens to be in memory after that. This happens also with e.g. a direct constant `debug_show(1e140)`.

Well, yeah that sucks. Thanks for the analysis! I didn't look into this, and when writing `float_fmt` I didn't consider that the exponents could be expanded. I will increase the buffer, so that all the digits will fit in! Good catch!

osa1:
> nope I don't think it is a problem, but it is annoying when it happens in tests, as they start spewing out weird stuff. I think this is some infelicity with OCaml's libs.

But they execute the same language. For a deterministic program if they output differently, then either one (or both) of them are buggy, or the program has undefined behavior.

[OCaml float documentation][1] says OCaml floats are 64-bit IEEE 754 numbers (same as Wasm `f64`s) so I think it should work. If not then there seems to be a bug to be fixed.

[1]: https://ocaml.org/api/Float.html

ggreif:
@cristianoc the `debug_show(1e140)` issue should be fixed now, PTAL.

nomeata:
> My suggestion list is {`VanillaValue`, `VanillaRepresentation`}, in this order. We use the term _vanilla_ a lot in `compile.ml`.

+1 for `VanillaValue`

osa1:
> In the interpreter I have seen failures for floatToInt(1.7e308), while the compiler works well

Here's the OCaml version of tommath mp_set_double:

```ocaml
(* tommath mp_set_double *)
let bigint_of_double (f : Wasm.F64.t) : Big_int.big_int =
  let bits = Wasm.F64.to_bits f in

  (* 0b111...1: 11 bits *)
  let mask_11_bits = Int64.of_int 0x7FF in

  (* Exponent part of IEEE 754 double, 11 bits *)
  let exp = Int64.(logand (shift_right_logical bits 52) mask_11_bits) in

  (* Fraction part of IEEE 754 double, 52 bits, but this also sets 53rd bit for
     some reason. tommath code: 

     frac = (cast.bits & (((uint64_t)1 << 52) - (uint64_t)1)) | ((uint64_t)1 << 52); *)
  let frac = Int64.(logor (shift_right_logical (shift_left bits 12) 12) (shift_left (of_int 1) 52)) in

  if Int64.(equal exp mask_11_bits) then
    (* Exponent is fully set: NaN or inf *)
    raise (Invalid_argument "bigint_of_double: argument is NaN or inf");

  (* I don't understand this part, copied from tommath mp_set_double
     exp -= 1023 + 52; *)
  let exp = Int64.(sub exp (of_int (1023 + 52))) in

  let a = Big_int.big_int_of_int64 frac in

  let a = if Int64.(compare exp (of_int 0)) < 0 then
    Big_int.(shift_right_big_int a (- (Int64.to_int exp)))
  else
    Big_int.(shift_left_big_int a (Int64.to_int exp))
  in

  if Int64.shift_right_logical bits 63 = Int64.of_int 1 && a <> Big_int.zero_big_int then
    Big_int.minus_big_int a
  else
    a
```

With this I get the same outputs for `floatToInt(1.7e308)`. I left inline comments for parts that I don't understand.

osa1:
> I left inline comments for parts that I don't understand

Figured it out. ([source][1]) The fraction part represents this number `1.xxxxx...` (in binary) where `x`s are the bits in fraction part of the double. That `1` is not represented in the 52 bits, it's implicit. That's the 53rd bit `mp_set_double` is setting.

In `exp -= 1023 + 52`, `1023` is "bias". So for example, value 1023 in exponent part of the double represents 0, 1022 represents -1 and so on (we don't have sign bit in exponent). `- 52` is for the lost fraction point in the `frac`. We represent `1.xxx...` as `1xxx...`, so need `-52` is for the missing fraction point there.

[1]: https://en.wikipedia.org/wiki/Double-precision_floating-point_format#IEEE_754_double-precision_binary_floating-point_format:_binary64

osa1:
Here's a better documented version of the same function:

```ocaml
let bigint_of_double (f : Wasm.F64.t) : Big_int.big_int =
  let bits = Wasm.F64.to_bits f in

  (* A bit pattern with 11 least significant bits set *)
  let bits_11 = Int64.of_int 0x7FF in

  (* Exponent part of IEEE 754 double, 11 bits *)
  let exp = Int64.(logand (shift_right_logical bits 52) bits_11) in

  (* Fraction part of IEEE 754 double, 52 bits from the float, with an implicit
     1 at the 53rd bit *)
  let frac = Int64.(logor (shift_right_logical (shift_left bits 12) 12) (shift_left (of_int 1) 52)) in

  if Int64.(equal exp bits_11) then
    (* Exponent is fully set: NaN or inf *)
    raise (Invalid_argument "bigint_of_double: argument is NaN or inf");

  (* Actual exponent value: subtract bias, and 52 to take the missing fraction
     dot in `frac`. Reminder: if fractional part is `xxx...` (binary) then
     actual fraction is `1.xxx...`, which we represent as `1xxx...` in `frac`.
     `- 52` here is to take that lost fraction point into account. *)
  let exp = Int64.(sub exp (of_int (1023 + 52))) in

  let a = Big_int.big_int_of_int64 frac in

  let a = if Int64.(compare exp (of_int 0)) < 0 then
    (* Exponent < 0, shift right *)
    Big_int.(shift_right_big_int a (- (Int64.to_int exp)))
  else
    (* Exponent >= 0, shift left *)
    Big_int.(shift_left_big_int a (Int64.to_int exp))
  in

  (* Negate the number if sign bit is set (double is negative) *)
  if Int64.shift_right_logical bits 63 = Int64.of_int 1 && a <> Big_int.zero_big_int then
    Big_int.minus_big_int a
  else
    a
```

I think this function could be in OCaml Big_int module.

rts/Makefile
ggreif: I have filed #2739 to track this.

src/mo_values/prim.ml
ggreif: Use @nomeata's code here!

test/run/float-ops.mo
```diff
@@ -90,3 +90,35 @@ assert (Prim.floatToInt64(-pi) == (-3 : Int64));
...
+assert (Prim.floatToInt(1073741823.5) == 1073741823);
+assert (Prim.floatToInt(1073741823.1) == 1073741823);
+assert (Prim.floatToInt(1073741822.9) == 1073741822);
+assert (Prim.floatToInt(-1073741823.9) == -1073741823);
+assert (Prim.floatToInt(-1073741824.1) == -1073741824);
+assert (Prim.floatToInt(-1073741824.5) == -1073741824);
+assert (Prim.floatToInt(-1073741824.8) == -1073741824);
+assert (Prim.floatToInt(-1073741825.1) == -1073741825);
+
+assert (Prim.floatToInt(1.7e308) == 169_999_999_999_999_993_883_079_578_865_998_174_333_346_074_304_075_874_502_773_119_193_537_729_178_160_565_864_330_091_787_584_707_988_572_262_467_983_188_919_169_916_105_593_357_174_268_369_962_062_473_635_296_474_636_515_660_464_935_663_040_684_957_844_303_524_367_815_028_553_272_712_298_986_386_310_828_644_513_212_353_921_123_253_311_675_499_856_875_650_512_437_415_429_217_994_623_324_794_855_339_589_632);
```
ggreif, line 109: Thanks to `bigint_of_double` this works now!

rts/motoko-rts/src/bigint.rs
ggreif: Done in 06c038331. But I see several other places where we don't check yet, e.g. `mp_set_i64/u64`. @osa1

src/mo_values/prim.ml
osa1: Which `bigint_of_double` do you mean? I see one in this thread which is mine.

src/mo_values/numerics.ml
```diff
@@ -27,6 +27,43 @@ let group_num s =
...
+
+     frac = (cast.bits & (((uint64_t)1 << 52) - (uint64_t)1)) | ((uint64_t)1 << 52); *)
+  let frac = Int64.(logor (shift_right_logical (shift_left bits 12) 12) (shift_left (of_int 1) 52)) in
+
+  if Int64.(equal exp mask_11_bits) then
+    (* Exponent is fully set: NaN or inf *)
+    raise (Invalid_argument "bigint_of_double: argument is NaN or inf");
+
+  (* I don't understand this part, copied from tommath mp_set_double
+     exp -= 1023 + 52; *)
```
osa1, line 51: @ggreif I have an updated version of this function in the thread with all this stuff explained. Could you update this code please? (removed the "automerge" label)

rts/Makefile
osa1: Would be good to add a comment here saying that tommath's implementation works fine with IEEE 754. No idea why they called it `IEC_559`.

src/mo_values/prim.ml
ggreif: Done: daa22fd.

rts/motoko-rts/src/bigint.rs
osa1: I checked the documentation and the source, I think `mp_set_i64` and `mp_set_u64` cannot fail.

src/mo_values/numerics.ml
ggreif: Oh, sorry I messed this up. I wanted to copy your code but ended up with the wrong one :-( Thanks for double checking!

I'll fix it.

ggreif:
> > this is a commented version of `mp_set_double` transcribed and commented by @nomeata and further refined with explanations by @osa1
> 
> Both of these were done by me, not Joachim.

Riiiiiight! I shouldn't do commits when on vacation 😞 . Another fix in the works!!!!

ggreif:
comment added in 73ec00225fed7eb18d277af9e64ea9c8cbad6552,
we can clean this up in the course of #2739.

ggreif:
That's great, thanks for checking. I guess the only problem is when
tomMath  cannot allocate, but that would be a trap.

osa1 (approved):
LGTM

PR #2738, opened 2021-08-20
Refactor SkewedPtr type

osa1:
SkewedPtr is used for values in heap slots (object fields). The name is
misleading because slots can have scalars or pointers (indicated by the least
significant bit of the value). This caused confusion in a few PRs before.

In addition, we currently have the skewing (for pointers) and shifting (for
scalars) duplicated in a few places.

This PR renames the type to Value. Representation of scalars and pointers are
now hidden from the callers and Values are analyzed with `View::get`, which
returns a view of the value that allows distinguishing values from pointers.
Methods `View::get_scalar` and `View::get_ptr` are used when we know that a
value is a scalar or pointer, respectively. Functions that assume that a value
is a scalar or a pointer always check that in debug mode.

Escape hatches `View::get_raw` and `View::from_raw` are used in two cases:

- In mark-compact GC we temporarily write values in slots that are not scalars
  or pointers.

- In debug functions we want to show the raw representation of a value

- In performance critical code when we know more about the `Value`
  representation than the optimizer (for example, for a value to be `Scalar(0)`
  it must be represented as literal `0`).

These functions should be avoided whenever possible.

API summary:

    /// A value in a heap slot
    #[derive(Clone, Copy, PartialEq, Eq)]
    pub struct Value(...);

    /// A view of `Value` for analyzing the slot contents.
    pub enum PtrOrScalar {
        /// Slot is a pointer to a boxed object
        Ptr(usize),

        /// Slot is an unboxed scalar value
        Scalar(u32),
    }

    impl Value {
        /// Create a value from a pointer
        pub const fn from_ptr(ptr: usize) -> Self { ... }

        /// Create a value from a scalar
        pub const fn from_scalar(value: u32) -> Self { ... }
     
        /// Create a value from raw representation. Useful when e.g.
        /// temporarily writing invalid values to object fields in garbage collection.
        pub const fn from_raw(raw: u32) -> Self { ... }

        /// Analyzes the value.
        pub fn get(&self) -> PtrOrScalar { ... }

        /// Get the raw value. Should only be used when debugging.
        pub fn get_raw(&self) -> u32 { ... }

        /// Is the value a scalar?
        pub fn is_scalar(&self) -> bool { ... }

        /// Is the value a pointer?
        pub fn is_ptr(&self) -> bool { ... }

        /// Assumes that the value is a scalar and returns the scalar value. In
        /// debug mode panics if the value is not a scalar.
        pub fn get_scalar(&self) -> u32 { ... }

        /// Assumes that the value is a pointer and returns the pointer value.
        /// In debug mode panics if the value is not a pointer.
        pub fn get_ptr(self) -> usize { ... }

        /// Get the object tag. In debug mode panics if the value is not a
        /// pointer.
        pub unsafe fn tag(self) -> Tag { ... }

        /// Get the pointer as `Obj`. In debug mode panics if the value is not
        /// a pointer.
        pub unsafe fn as_obj(self) -> *mut Obj { ... }

        /// Get the pointer as `Array`. In debug mode panics if the value is
        /// not a pointer or the pointed object is not an `Array`.
        pub unsafe fn as_array(self) -> *mut Array { ... }

        /// Get the pointer as `Concat`. In debug mode panics if the value is
        /// not a pointer or the pointed object is not an `Concat`.
        pub unsafe fn as_concat(self) -> *mut Concat { ... }

        /// Get the pointer as `Blob`. In debug mode panics if the value is not
        /// a pointer or the pointed object is not an `Blob`.
        pub unsafe fn as_blob(self) -> *mut Blob { ... }

        /// Get the pointer as `BigInt`. In debug mode panics if the value is
        /// not a pointer or the pointed object is not an `BigInt`.
        pub unsafe fn as_bigint(self) -> *mut BigInt { ... }
    }

TODOs:

- [ ] Assertion failures in debug mode. Release mode seems to work fine.
- [ ] const fns currently do not check the arguments as in the current Rust version we're using const fns cannot panic. Updating rustc will fix this.
- [ ] Check performance


osa1:
"VanillaValue" came up before but I don't like it too much because the "vanilla" part does not indicate anything (i.e. we don't have non-vanilla names). Value is shorter and gives same amount of information.

nomeata:
> we don't have non-vanilla names

We have on the stack, where vanilla is one of a number of ways to represent something on the stack. But for conciseness Value works too, I don't mind either way

ggreif:
> "VanillaValue" came up before but I don't like it too much because the "vanilla" part does not indicate anything (i.e. we don't have non-vanilla names). Value is shorter and gives same amount of information.

I am not sure I can follow. We have values of all flavours lying on the stack, `{0, 1}` booleans, pointers, `i32`s. These all are not _vanilla_. Vanilla is the flavour that appeals to most "audiences", i.e. it can accommodate all data types. All I want to say is that `Value` would be a misnomer IMO.

osa1:
> I am not sure I can follow. We have values of all flavours lying on the stack, {0, 1} booleans, pointers, i32s. These all are not vanilla. Vanilla is the flavour that appeals to most "audiences", i.e. it can accommodate all data types. All I want to say is that Value would be a misnomer IMO.

As far as RTS concerned there's only one type of value. Other representations used by the code gen are not used anywhere in the RTS.

If we want to represent other kinds of values in the RTS for some reason (maybe in the future if we do GC in the middle of a message) then I'd still like something else than "vanilla" because that word is not informative (at least to me, though I realize that this is also a familiarity thing). I like HeapValue more, it makes it clear that the representation is for uses in heap, and I think it would mean the same to other readers (unlike "vanilla"). Similarly we could use StackValue for the values on the stack. Another idea is SlotValue as they're used in heap "slots" (object fields). FieldValue would be OK too, I think.

Using a common vocabulary in both the compiler and RTS is a good idea, but I'm just not sure that using "vanilla" for heap values is a good naming. Maybe we can rename uses in the compiler too? :-)

nomeata:
HeapValue is misleading as well, as these 32 bits are not always on the heap (they can be on the stack) nor do they always point to the heap (can be scalars).

A name that is not idiocratic, but correct, might be UniformRep, as it's the the uniform representation (a common technical term) that we use to implement polymorphism etc.

I'm personally not bothered by idiosyncrasies like Vanilla. They may even be better than terms like HeapValue that may lead the uninformed to jump to wrong conclusions (e.g. that these are always heap allocated).

nomeata (approved):
The refactoring is a great improvement, and bikeshedding the name should not hold it up

ulan (approved):
Great work, lgtm!

Out of all naming options mentioned I like "Value" the most, but I don't have a strong opinion here.

rts/motoko-rts/src/types.rs
```diff
@@ -124,51 +124,136 @@ impl From<Words<u32>> for Bytes<u32> {
...
+    /// Get the pointer as `Blob`. In debug mode panics if the value is not a pointer or the
+    /// pointed object is not an `Blob`.
     pub unsafe fn as_blob(self) -> *mut Blob {
         debug_assert_eq!(self.tag(), TAG_BLOB);
-        self.unskew() as *mut Blob
+        self.get_ptr() as *mut Blob
     }
 
+    /// Get the pointer as `BigInt`. In debug mode panics if the value is not a pointer or the
+    /// pointed object is not an `BigInt`.
```
crusso, line 244:
```suggestion
    /// pointed object is not a `BigInt`.
```

rts/motoko-rts/src/types.rs
```diff
@@ -124,51 +124,136 @@ impl From<Words<u32>> for Bytes<u32> {
...
+    /// Get the pointer as `Concat`. In debug mode panics if the value is not a pointer or the
+    /// pointed object is not an `Concat`.
     pub unsafe fn as_concat(self) -> *mut Concat {
         debug_assert_eq!(self.tag(), TAG_CONCAT);
-        self.unskew() as *mut Concat
+        self.get_ptr() as *mut Concat
     }
 
+    /// Get the pointer as `Blob`. In debug mode panics if the value is not a pointer or the
+    /// pointed object is not an `Blob`.
```
crusso, line 237:
```suggestion
    /// pointed object is not a `Blob`.
```

rts/motoko-rts/src/types.rs
```diff
@@ -124,51 +124,136 @@ impl From<Words<u32>> for Bytes<u32> {
...
+    /// Get the pointer as `Array`. In debug mode panics if the value is not a pointer or the
+    /// pointed object is not an `Array`.
     pub unsafe fn as_array(self) -> *mut Array {
         debug_assert_eq!(self.tag(), TAG_ARRAY);
-        self.unskew() as *mut Array
+        self.get_ptr() as *mut Array
     }
 
+    /// Get the pointer as `Concat`. In debug mode panics if the value is not a pointer or the
+    /// pointed object is not an `Concat`.
```
crusso, line 230:
```suggestion
    /// pointed object is not a `Concat`.
```

rts/motoko-rts/src/text_iter.rs
```diff
@@ -70,13 +66,13 @@ pub unsafe fn text_iter<M: Memory>(mem: &mut M, text: SkewedPtr) -> SkewedPtr {
...
-pub unsafe extern "C" fn text_iter_done(iter: SkewedPtr) -> u32 {
+pub unsafe extern "C" fn text_iter_done(iter: Value) -> u32 {
     let array = iter.as_array();
-    let pos = array.get(ITER_POS_IDX).0 >> 2;
+    let pos = array.get(ITER_POS_IDX).get_scalar();
     let blob = array.get(ITER_BLOB_IDX).as_blob();
     let todo = array.get(ITER_TODO_IDX);
 
-    if pos >= blob.len().0 as usize && todo == SkewedPtr(0) {
+    if pos >= blob.len().0 && matches!(todo.get(), PtrOrScalar::Scalar(0)) {
```
crusso, line 75: Is the cost of this test really trivial or would it worth interoducing a specialized is_Zero function?

rts/motoko-rts/src/text_iter.rs
```diff
@@ -85,17 +81,17 @@ pub unsafe extern "C" fn text_iter_done(iter: SkewedPtr) -> u32 {
...
     let blob = iter_array.get(ITER_BLOB_IDX).as_blob();
-    let pos = (iter_array.get(ITER_POS_IDX).0 >> 2) as u32;
+    let pos = iter_array.get(ITER_POS_IDX).get_scalar();
 
     // If we are at the end of the current blob, find the next blob
     if pos >= blob.len().0 {
         let todo = iter_array.get(ITER_TODO_IDX);
 
-        if todo == SkewedPtr(0) {
+        if matches!(todo.get(), PtrOrScalar::Scalar(0)) {
```
crusso, line 94: ditto

rts/motoko-rts/src/types.rs
```diff
@@ -124,51 +124,136 @@ impl From<Words<u32>> for Bytes<u32> {
...
+    /// Create a value from raw representation. Useful when e.g. temporarily writing invalid values
+    /// to object fields in garbage collection.
+    pub const fn from_raw(raw: u32) -> Self {
+        Value(raw)
+    }
+
+    /// Analyzes the value.
+    pub fn get(&self) -> PtrOrScalar {
+        if self.0 & 0b1 == 0b1 {
+            PtrOrScalar::Ptr(unskew(self.0 as usize))
```
crusso, line 175: I assume this is just stack allocation in Rust, right?

crusso (approved):
Looks very nice! Minor comments. I think Value is a good name.

rts/motoko-rts/src/types.rs
osa1: Yes.

rts/motoko-rts/src/text_iter.rs
osa1: I will check that. We also have a similar pattern in `continuation_table::table_initialized`.

osa1: LLVM or rustc cannot optimize one of these

```
// 1
!matches!(TABLE.get(), PtrOrScalar::Scalar(0))
// 2
TABLE.get() == PtrOrScalar::Scalar(0)
```

into

```
TABLE.0 == 0
```

Which makes sense, because without knowing that `Scalar` will always have 0 in the least significant bit (an invariant we maintain in the constructors) it cannot assume that for `value >> 1` to be `0`, `value` needs to be `0`.

Doing this optimization manually reduces binary sizes and should improve performance. I will commit this change now.

osa1:
It seems like perf delta script is buggy. Last commit changed binary sizes and should also improve performance, but the comment is not updated.

nomeata:
If we look at the corresponding job (https://github.com/dfinity/motoko/pull/2738/checks?check_run_id=3408389610), it clearly ran and updated the comment, but it updated it with the same text, so github doesn’t register that as a change. It looks like the change didn’t change it measurably. I’ll include the commits in the message to avoid such confusion…

nomeata:
https://github.com/dfinity/motoko/pull/2746 should help with that confusion

osa1:
>  In terms of gas, 3 tests improved and the mean change is -0.0%.
> In terms of size, 3 tests improved and the mean change is -0.0%.

OK, this is ready for merging now.

I'm a bit surprised that rustc/LLVM cannot optimize `pointer_to_dynamic_heap`: https://godbolt.org/z/9TdqrrEfo

In function `bad` after inlining `Value::get` there's no reason to compute `value >> 1` as that expression has no side effects (cannot fail, cannot change heap) and is not used, but the generated code computes both `value >> 1` and `value + 1` and then selects one based one `value & 0b1`. So we have to do this optimization manually.

osa1:
After the last commit all uses of `Value::get` other than those in assertions are removed because the generated code is inefficient compared to what we can write with `get_raw` and `unskew`. Unfortunate..

nomeata:
>  but the generated code computes both `value >> 1` and `value + 1` and then selects one based on `value & 0b1`. So we have to do this optimization manually.

I’m not surprised. LLVM optimizes for runtime, not cycle count, and in that case it can produce branchless (and thus likely much faster) code by computing both values and selecting the right one using `select`, rather than using `if`. On our target platform, the cost model is different and `if` might be better, but LLVM doesn’t do that yet.

Issue #2568, opened 2021-06-09
CI: Publish reports


nomeata:
Our old CI was able to publish reports of various kinds (profiling reports, rendered documentation etc.) for each build. It even had a stable URL for “latest build report on branch X”, which we could link from the top-level README.

It would be good to recreate that using our Github Actions CI somehow. I am not sure what the best way is, though, so this needs some good ideas/research.

nomeata:
@krpeacock [recommends](https://twitter.com/kylpeacock/status/1429165452094152710) netlify, which we already use in the agent-* repos.

If someone in the org can configure the `NETLIFY_AUTH_TOKEN` and `NETLIFY_SITE_ID` secrets for the `motoko` repo, I can set up the deployment.

nomeata:
I guess with https://github.com/dfinity/motoko/pull/3113 I can close this. We could also use netlify if we want per-PR reports, but so far that is not a strong demand, it seems.

Issue #2565, opened 2021-06-09
Public CI metaissue


nomeata:
This issue tracks various refinements we want to do to our CI infrastructure post-open source. 

 * [x] https://github.com/dfinity/motoko/issues/2544 Pick a public CI system (Github actions + cachix for now)
 * [x] https://github.com/dfinity/motoko/pull/1224 Set it up
 * [x] https://github.com/dfinity/motoko/issues/2566 Automatically create releases and publish build artifacts there
 * [x] https://github.com/dfinity/motoko/issues/2567 Allow public access to `drun`, so that all tests can be run on CI
       This blocks most of the following items:
 * [ ] https://github.com/dfinity/motoko/issues/2568 Publish reports (profiling, documentation preview) from GitHub Action runs
 * [x] #2740 Report performance changes in comment to PR (maybe using [this Github Action](https://github.com/marketplace/actions/create-or-update-comment#where-to-find-the-id-of-a-comment))
 * [x] #2744 Switch branch protection to Github Action rather than hydra
 * [ ] #2771 Stop using the internal CI system.

osa1:
"CI: Publish reports" this is not something we do today, right?

I'm thinking what's left to do before stopping using Hydra. I think once we port posting perf and size changes to GH CI we should be able ditch it?

nomeata:
> "CI: Publish reports" this is not something we do today, right?

Yes, we have direct links to the latest flame graph, documentation, etc. for each PR. Would be a shame to lose that, but maybe ok temporarily until we have a good idea. I should research what kind of services are there out there.

Besides that it's just the perf changes comments, which I can look into this week maybe

osa1:
> Yes, we have direct links to the latest flame graph, documentation, etc. for each PR.

Interesting, I didn't know that. Where can I see these reports?

nomeata:
You didn't know? Anyways, see https://github.com/dfinity/motoko/pull/2562 for some of the links, which we removed to keep up appearances.

nomeata:
Flame graphs at https://hydra.dfinity.systems/job/dfinity-ci-build/motoko/tests.profiling-graphs/latest/download/1/index.html I assume (I can't check, of course). One can construct similar urls for each PR. I hope to preserve that feature, or improve it (easier discoverability in particular of course). But needs not block turning off hydra.

nomeata:
[...]

PR #2741, opened 2021-08-21
CI: Report performance delta via comment

### LINKED ISSUE #2740: CI: Create performance delta comment via Github Action
The performance delta report is a useful feature that is currently implemented via Hydra, which we want to decommission (#2565).

I’ll play around with https://github.com/marketplace/actions/create-or-update-comment#where-to-find-the-id-of-a-comment.

nomeata:
fixes #2740. See CI.md for an overview of how it works.


nomeata:
This looks plausible. With this merged, I think turning off hydra becomes a viable option.

nomeata:
foo

CI.md
```diff
@@ -118,6 +118,13 @@ Performance changes are known
 For every PR, the developer is told about performance changes relative to the
 merge point, via an continuously updated comment on the PR.
 
+**Implementation (external):**
+ * Steps in the Github Actoin calculates the correct merge base using
```
crusso, line 122:
```suggestion
 * Steps in the Github Action calculates the correct merge base using
```

crusso (approved):
More of rubberstamp than a review. Perhaps @osa1 should take a look.

.github/workflows/test.yml
```diff
@@ -1,5 +1,8 @@
 name: "build"
-on: push
+on:
+  push:
+     branches: [ master ]
```
osa1, line 4: Where do we see the results for master? It seems like the code below is only adding PR comments (`github.event_name == 'pull_request'`) so results for master will be lost?

.github/workflows/test.yml
```diff
@@ -24,3 +32,41 @@ jobs:
...
+        path: ./perf-delta.txt
+
+    - name: Find performance comment
+      if: runner.os == 'Linux' && github.event_name == 'pull_request'
+      uses: peter-evans/find-comment@v1
+      id: fc
+      with:
+        issue-number: ${{ github.event.pull_request.number }}
+        comment-author: 'github-actions[bot]'
+        body-includes: '<!-- perf comment -->'
```
osa1, line 61: So this updates an existing perf comment, but where do we create a comment with this line for the first time?

I think adding a new comment might also be useful to see changes with commits.

osa1 (approved):
LGTM. Added some questions to understand this better.

Should we remove ci-pr.nix now? It seems like `perf-delta.nix` is doing a similar thing. Do we need both?

nomeata:
> Should we remove ci-pr.nix now? It seems like perf-delta.nix is doing a similar thing. Do we need both?

I’d suggest to just pull the plug on hydra, and then remove all hydra files.

osa1:
Even if we keep hydra, is there a need to have two bots adding the same comment?

.github/workflows/test.yml
nomeata: Differential CI needs something to compare changes against, i.e. a base branch, and this is about commenting on PRs, which doesn’t exist for pushes to master, so this stuff only happens for PRs.

You can of course always run `nix-build -A tests.perf` on any commit to get the base data, i.e. the CSV file:
```
~/dfinity/motoko $ cat $(nix-build -A tests.perf)/stats.csv
gas/assetstorage;14442
size/assetstorage;165585
gas/qr;1237055836
size/qr;220465
gas/reversi;1305950
size/reversi;160031
```

We may want to publish the data along with other reports (e.g. flamegraphs), see #2568 for that.

.github/workflows/test.yml
nomeata: This Github action creates a commit if there is none, or updates one if there is one – no change in behavior here (if everything works as intended)

nomeata:
> Even if we keep hydra, is there a need to have two bots adding the same comment?

No, if we keep hydra we would remove the code from `ci-pr.nix`. Do we want to keep hydra?

osa1:
@nomeata let's remove ci-pr.nix then, no need to have two bots add the same comment.

nomeata:
Ok, see https://github.com/dfinity/motoko/pull/2745.

But someone (you?) should still tell the IDX team (@nmattia maybe?) that we no longer need hydra, and they can turn it off. And _then_ we can remove all hydra-related files (`ci.nix`, `ci-pr.nix`, maybe others, will check) from the repo.

osa1:
I thought we still need Hydra until flamegraph generation is ported to GitHub CI? https://github.com/dfinity/motoko/issues/2565#issuecomment-900927515

nomeata:
The graphs are still generated, just not published at a easily accessible url. Given that not even you knew about that feature, and it's not used often, it doesn't seem to be that important, so I suggest we don't wait for that.

osa1:
I think there's a problem with this PR. In #2738 I made a change that changes binary sizes and should also affect performance, but the comment (https://github.com/dfinity/motoko/pull/2738#issuecomment-903910303) is not updated.

nomeata:
Too bad we didn’t keep the other bot for a while to compare…

osa1:
I will run the nix command locally to compare, but the comment is updated 16 hours ago so definitely not updated after my last commit which I pushed about an hour ago.

nomeata:
oh, parallel conversation threads. Let’s continue at https://github.com/dfinity/motoko/pull/2738#issuecomment-904432467. I think all is well.

nmattia:
I've pinged IDX for you!

PR #2742, opened 2021-08-21
RTS: Shift scalar values by one instead of two

osa1:
Shifting by one is enough to make a scalar value.
This was omitted in 365b4b8d471b14c18e5c525c757ec7135c1bc171


nomeata (approved):
Well spotted! This was an omission by me when I did 365b4b8d471b14c18e5c525c757ec7135c1bc171

PR #2746, opened 2021-08-24
CI: Include revisions in perf change comment

nomeata:
this makes it easier to see if it actually has updated yet, and may
avoid confusion about what te base commit it.


osa1 (approved):
This looks much better, thanks.

PR #2747, opened 2021-08-24
CI: Use nix-build-uncached to speed up CI jobs

nomeata:
before, the `nix-build` step would build or download everything defined in the
`default.nix` file. But there is no point in downloading stuff from the
cache here; if it is there, all is well.

The [`nix-build-uncached` tool](https://github.com/Mic92/nix-build-uncached) supposedly fixes that. So let’s try.


crusso (commented):
Rubberstamping

crusso (approved):
Approving

PR #2749, opened 2021-08-25
nix: Try different patch to fix -pie issues

nomeata:
in #2532 we added a patch related to static building of ocaml packages,
submitted to nixpkgs as https://github.com/NixOS/nixpkgs/pull/124498,
but it was never merged upstream.

Supposedly a patch from https://github.com/NixOS/nixpkgs/pull/135619
fixes it as well (and maybe more properly). So let’s try that!


nomeata:
Probably obsolete due to #2954

PR #2753, opened 2021-08-27
[feat] StableMemory: load/store implementations for numeric types beyond Nat32

crusso:
- [x] Nat8
- [x] Nat16
- [x] Nat32
- [x] Nat64
- [x] Float
- [x] Int8
- [x] Int16
- [x] Int32
- [x] Int64


nomeata:
These operations look awfully useful when doing binary data formats. Can we maybe create a “mutable blob” type, with ways to create them backed by stable memory or normal memory, and have a common interface?

(Maybe later of cause)

crusso:
> These operations look awfully useful when doing binary data formats. Can we maybe create a “mutable blob” type, with ways to create them backed by stable memory or normal memory, and have a common interface?
> 
> (Maybe later of cause)

Yeah, I've been wondering about adding a mutable blob type or perhaps we should optimize arrays more generally.

 I'm not sure we want to have each operation dispatch on the type of address, though do we? Also, the load operations would make sense on both mutable and immutable blobs, but I don't think we have any way to abstract on that except overloading (as with our two array flavours).
```
loadNat8 : ({#Blob : Blob; #VarBlob : VarBlob, #Mem}, offset) : Nat8 
storeNat8 : ({#Blob : Blob; #VarBlob : VarBlob, #Mem}, offset, value : Nat8) : Nat8 
```

Looks a bit clunky to me and even less efficient.

Perhaps we should just treat StableMemory as a singleton object and have a class mapping (mutable) Blobs to object with various checked (store/)load  operations.

```
class Reader(b : Blob) {
   loadNat8 : (offset : Nat32) -> Nat8
  ...
}
class ReaderWriter(b : VarBlob) {
 loadNat8 : (offset : Nat32) -> Nat8
 storeNat8 : (offset : Nat32, value: Nat8) -> ()
};
}
```
Which also begs the question of whether to use `Nat32` or `Nat` or maybe just `Nat64` for offsets.

crusso:
Kinda makes you wish you had type classes...

rossberg (approved):
Cool, LGTM. Maybe @ggreif can have a second look at the codegen.

src/codegen/compile.ml
```diff
@@ -7784,16 +8006,86 @@ and compile_exp (env : E.t) ae exp =
...
+      SR.Vanilla,
+      compile_exp_as env ae SR.UnboxedWord32 e ^^
+      StableMem.load_word8 env ^^
+      TaggedSmallWord.msb_adjust Type.Int8
+
+    | OtherPrim ("stableMemoryStoreNat8"), [e1; e2] ->
+      SR.unit,
+      compile_exp_as env ae SR.UnboxedWord32 e1 ^^
+      compile_exp_as env ae SR.Vanilla e2 ^^ TaggedSmallWord.lsb_adjust Type.Nat8 ^^
+      StableMem.store_word8 env
```
ggreif, line 8036: I wonder if you could sink the `m/lsb_adjusts` into the `StableMem.store_word8`.

src/prelude/prim.mo
```diff
@@ -292,6 +292,54 @@ func stableMemoryGrow(pages : Nat32) : Nat32 =
...
+
+func stableMemoryStoreNat16(offset : Nat32, val : Nat16) : () =
+  (prim "stableMemoryStoreNat16" : (Nat32, Nat16) -> ()) (offset, val);
+
+func stableMemoryLoadNat64(offset : Nat32) : Nat64 =
+  (prim "stableMemoryLoadNat64" : Nat32 -> Nat64) offset;
+
+func stableMemoryStoreNat64(offset : Nat32, val : Nat64) : () =
+  (prim "stableMemoryStoreNat64" : (Nat32, Nat64) -> ()) (offset, val);
+
```
ggreif, line 315: Nit: extra line?

src/codegen/compile.ml
```diff
@@ -3980,6 +4095,113 @@ module StableMem = struct
...
+        (("offset", I32Type)) [F64Type]
+        (fun env get_offset  ->
+          get_offset ^^
+          compile_unboxed_const 8l ^^
+          guard_range env ^^
+          get_offset ^^
+          read_float64 env)
+    | _ -> assert false
+
+  let store_float64 env =
```
ggreif, line 4191: Around line 3936 there is a similar function? What is the semantic difference between `__stablemem_store_float64` and `__stablemem_write_float64`?

src/codegen/compile.ml
```diff
@@ -3832,6 +3832,120 @@ module StableMem = struct
...
+            get_temp_ptr ^^
+            G.i (Load {ty = F64Type; align = 0; offset = 0l; sz = None })))
+    | _ -> assert false
+
+  (* write stable mem offset and value, operands on stack *)
+  let write_float64 env =
+    match E.mode env with
+    | Flags.ICMode | Flags.RefMode ->
+      Func.share_code2 env "__stablemem_write_float64"
+        (("offset", I32Type),("value", F64Type)) []
```
ggreif, line 3940:
```suggestion
        (("offset", I32Type), ("value", F64Type)) []
```
Nit, there are a few more above.

src/codegen/compile.ml
```diff
@@ -3980,6 +4095,113 @@ module StableMem = struct
...
     | _ -> assert false
+
+  let load_word8 env =
+    match E.mode env with
+    | Flags.ICMode | Flags.RefMode ->
+      Func.share_code1 env "__stablemem_load_word8"
+        (("offset", I32Type)) [I32Type]
+        (fun env get_offset  ->
+          get_offset ^^
+          guard env ^^
```
ggreif, line 4106: `guard env` could leave the offset on the stack, then the next line is not necessary.

src/codegen/compile.ml
ggreif: I presume the `load`/`store` ones operate on guarded ranges?

src/codegen/compile.ml
```diff
@@ -3832,6 +3832,120 @@ module StableMem = struct
...
 
+  (* read word8 from stable mem offset on stack *)
+  let read_word8 env =
+    match E.mode env with
+    | Flags.ICMode | Flags.RefMode ->
+      Func.share_code1 env "__stablemem_read_word8"
+        ("offset", I32Type) [I32Type]
+        (fun env get_offset ->
+          Stack.with_words env "temp_ptr" 1l (fun get_temp_ptr ->
+            get_temp_ptr ^^ get_offset ^^  compile_unboxed_const 1l ^^
```
ggreif, line 3843:
```suggestion
            get_temp_ptr ^^ get_offset ^^ compile_unboxed_const 1l ^^
```
Nit, there are a few around.

ggreif (approved):
Codegen looks good, a few suggestions and nits only.

src/codegen/compile.ml
ggreif: Then some of these could be collapsed using `or` patterns.

src/prelude/prim.mo
crusso:
```suggestion
```

src/codegen/compile.ml
crusso: Yeah, I'm not sure Its worth it . I don't want to retest the string lit and don't want to rely on the type of e2 (that could, in principal, be `Non`). I'll just leave it, but thanks for the suggestion.

src/codegen/compile.ml
crusso: That's just pushing the load elsewhere and I'd like to keep the code similar to guard_range...

PR #2757, opened 2021-08-30
Remove didc

### LINKED ISSUE #2755: Do we still need didc here?
This repository still ships `didc`, as documented in https://github.com/dfinity/motoko/blob/master/design/DFX-Interface.md#compiling-idl-files-to-js. But I believe that the Rust-implemented `didc` supports this as well, and probably better.

Does that mean we can and should remove `didc` from https://github.com/dfinity/motoko/blob/master/design/DFX-Interface.md?

If so, what else uses it? 

@chenyan-dfinity

nomeata:
this tool has been superseded by the `didc` program built as part of the
Rust Candid library,
<https://github.com/dfinity/candid/blob/master/tools/didc/src/main.rs>.

Closes #2755


design/IDL-Motoko.md
```diff
@@ -312,26 +312,14 @@ just suggestions.
...
   and check that `e(t) <: t_spec` (using IDL subtyping).
 
-* Converting IDL types to Motoko types
+* Importing IDL types from the Motoko compiler
 
-  If `foo.did` a textual IDL file, then
+  If `path/to/foo.did` a textual IDL file, then a declaration
 
-      didc foo.did -o foo.mo
+      import Foo "path/to/foo"
```
chenyan-dfinity, line 319: This is not supported yet?

test/run.sh
```diff
@@ -443,34 +438,6 @@ do
...
-    tc_succeeded=$?
-    normalize $out/$base.tc
-    diff_files="$diff_files $base.tc"
-
-    if [ "$tc_succeeded" -eq 0 ];
-    then
-      $ECHO -n " [pp]"
-      didc --pp $base.did > $out/$base.pp.did
-      sed -i 's/import "/import "..\//g' $out/$base.pp.did
-      didc --check $out/$base.pp.did > $out/$base.pp.tc 2>&1
```
chenyan-dfinity, line 460: hmm, this test is still useful. Motoko needs the candid parser and it's good if we can test it somehow.

design/IDL-Motoko.md
nomeata: Yeah, but it's an one design doc and thus was already there

test/run.sh
nomeata: Yes, this is the only thing we lose. We should import the new rust based didc and use it here, wdyt? Might be convenient to have that tool in the nix shell anyways.

chenyan-dfinity: We want to test the parser from `idllib` which is used by import. Maybe we can wrap the .did file in the test as a  motoko program and test from moc --check

nomeata: Ah, this isn't the line where we check the output of `moc -idl`. Sorry.

We have existing tests that exercise the parser via `import` which is what matters. Should we go through the `.did` tests and see which are valuable and make them import tests?

chenyan-dfinity: Sure, we can put all the `.did` file into the `import` test. I think they are all useful in testing the parser.

nomeata:
I removed the automerge label, I guess we want to scavenge the `.did` tests first and see which one we want to preserve as input tests.

PR #2760, opened 2021-09-01
Update linker in prep for updating LLVM

osa1:
- Implement extended name section
  (https://github.com/WebAssembly/extended-name-section/blob/master/proposals/extended-name-section/Overview.md)

- Handle `start` functions in the RTS module in linker

These changes are needed to update to LLVM 12: #2542


nomeata (approved):
Thanks

PR #2761, opened 2021-09-01
Update Rust

osa1:
This PR bumps rustc version and removes xargo. std dependencies are now built
with cargo's new `-Zbuild-std` parameter.

To work around rust-lang/wg-cargo-std-aware#23, we implement a tool
"vendor-rust-std-deps" that reads `Cargo.lock` of a Rust toolchain std and
manually vendors all the dependencies.

These dependencies are then merged with the Rust RTS dependencies before the
building the RTS in nix.

RTS README updated with instructions to bump rustc.

New rustc will enable more const functions, new API functions, stabilizations,
and features for other RTS PRs.


nomeata (approved):
Looks simpler. Great!

osa1:
We need to add `compiler_builtins` as a dep so that it will be vendored, but no matter what version of `compiler_builtins` I add, `cargo build` tries to build both 0.1.50 and 0.1.49 and fails because apparently you can't link with two different versions of `compiler_builtins` (unlike other libraries). Without a `compiler_builtins` dependency it builds 0.1.49 so I think adding `compiler_builtins = "0.1.49"` should work, but it doesn't. `cargo build` tries to build `0.1.50` when I do that.

osa1:
If I use `compiler_builtins = "=0.1.49"` then it uses 0.1.49 as expected, and I don't see multiple versions of `compiler_builtins` being built in `cargo build` output, but it still fails with

```
error[E0465]: multiple rlib candidates for `compiler_builtins` found
  |
  = note: candidate #1: /home/omer/dfinity/motoko/rts/motoko-rts/target/wasm32-unknown-emscripten/release/deps/libcompiler_builtins-da9a0c66847c05a7.rlib
  = note: candidate #2: /home/omer/dfinity/motoko/rts/motoko-rts/target/wasm32-unknown-emscripten/release/deps/libcompiler_builtins-122ca795d65a1505.rlib
```

osa1:
> We need to add compiler_builtins as a dep so that it will be vendored

I wonder if this is a `cargo vendor` bug. I shouldn't need to add deps of deps to my crates for it to work.

osa1:
I'm convinced that this is a `cargo vendor` bug and filed an issue to cargo. compiler_builtins is being downloaded from crates.io so it should be vendored.

ggreif:
> filed an issue to cargo

Can you add a link here?

osa1:
A cargo dev confirmed that the problem with `cargo vendor` mentioned above is a bug, but it doesn't seem like it's going to be fixed soon. We will probably need to work around it somehow.

osa1:
I wonder if we could manually download the packages that `cargo vendor` fails to vendor, in preBuild step.

osa1:
Nix team pointed out that this empty checksum file disables checking hashes of files and the package in cargo: https://github.com/nix-community/naersk/blob/ee7edec50b49ab6d69b06d62f1de554efccb1ccd/build.nix#L126

With this we should be able to manually vendor dependencies. (previously this didn't work because I couldn't figure out how to generate hashes of files, it's not just `sha256sum <file>`)

osa1:
Good point. The reason why we need `--build-std` is we need PIC relocation model when building the libraries, `-Crelocation-model=pic`, otherwise we can't link the generated code (see below). If fenix allows us to configure Rust build system's `config.toml` then we should be able to build the libraries with PIC relocation model and not use `--build-std`.

---

I just tested if we still need PIC relocation model with the more recent Rust build, and got the same error output as before:

```
wasm-ld: warning: creating shared libraries, with -shared, is not yet stable
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_TABLE_INDEX_SLEB cannot be used against symbol core::fmt::num::imp::_$LT$impl$u20$core..fmt..Display$u20$for$u20$u32$GT$::fmt::hdd03639995f8d7ed; recompile with -fPIC
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_MEMORY_ADDR_SLEB cannot be used against symbol .Lanon.41584def8eb6324459d82d576c638e33.176; recompile with -fPIC
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_TABLE_INDEX_SLEB cannot be used against symbol core::fmt::num::imp::_$LT$impl$u20$core..fmt..Display$u20$for$u20$u32$GT$::fmt::hdd03639995f8d7ed; recompile with -fPIC
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_TABLE_INDEX_SLEB cannot be used against symbol core::fmt::num::imp::_$LT$impl$u20$core..fmt..Display$u20$for$u20$u32$GT$::fmt::hdd03639995f8d7ed; recompile with -fPIC
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_MEMORY_ADDR_SLEB cannot be used against symbol .Lanon.41584def8eb6324459d82d576c638e33.251; recompile with -fPIC
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_TABLE_INDEX_SLEB cannot be used against symbol core::fmt::num::imp::_$LT$impl$u20$core..fmt..Display$u20$for$u20$u32$GT$::fmt::hdd03639995f8d7ed; recompile with -fPIC
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_TABLE_INDEX_SLEB cannot be used against symbol core::fmt::num::imp::_$LT$impl$u20$core..fmt..Display$u20$for$u20$u32$GT$::fmt::hdd03639995f8d7ed; recompile with -fPIC
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_MEMORY_ADDR_SLEB cannot be used against symbol .Lanon.41584def8eb6324459d82d576c638e33.253; recompile with -fPIC
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_TABLE_INDEX_SLEB cannot be used against symbol core::fmt::num::imp::_$LT$impl$u20$core..fmt..Display$u20$for$u20$u32$GT$::fmt::hdd03639995f8d7ed; recompile with -fPIC
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_MEMORY_ADDR_SLEB cannot be used against symbol .Lanon.41584def8eb6324459d82d576c638e33.101; recompile with -fPIC
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_TABLE_INDEX_SLEB cannot be used against symbol core::fmt::num::imp::_$LT$impl$u20$core..fmt..Display$u20$for$u20$u32$GT$::fmt::hdd03639995f8d7ed; recompile with -fPIC
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_MEMORY_ADDR_SLEB cannot be used against symbol .Lanon.41584def8eb6324459d82d576c638e33.256; recompile with -fPIC
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_TABLE_INDEX_SLEB cannot be used against symbol core::fmt::num::imp::_$LT$impl$u20$core..fmt..Display$u20$for$u20$u32$GT$::fmt::hdd03639995f8d7ed; recompile with -fPIC
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_MEMORY_ADDR_SLEB cannot be used against symbol .Lanon.41584def8eb6324459d82d576c638e33.169; recompile with -fPIC
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_MEMORY_ADDR_SLEB cannot be used against symbol .Lanon.41584def8eb6324459d82d576c638e33.101; recompile with -fPIC
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_TABLE_INDEX_SLEB cannot be used against symbol core::ops::function::FnOnce::call_once::ha3741e923a2931b4; recompile with -fPIC
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_TABLE_INDEX_SLEB cannot be used against symbol core::ops::function::FnOnce::call_once::ha3741e923a2931b4; recompile with -fPIC
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_MEMORY_ADDR_SLEB cannot be used against symbol .Lanon.41584def8eb6324459d82d576c638e33.101; recompile with -fPIC
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_MEMORY_ADDR_SLEB cannot be used against symbol .Lanon.41584def8eb6324459d82d576c638e33.131; recompile with -fPIC
wasm-ld: error: _build/wasm/libmotoko_rts.a(core-3e9a2fb490781a64.core.b5f0d450-cgu.0.rcgu.o): relocation R_WASM_MEMORY_ADDR_SLEB cannot be used against symbol .Lanon.41584def8eb6324459d82d576c638e33.225; recompile with -fPIC
```

We also pass `-Ccodegen-units=1`, but that's not strictly necessary. It just makes debugging easier because it generates less `.o` files (and IIRC some of the symbols have shorter names).

osa1:
Actually I'm not sure that building Rust with a custom `config.toml` will help. I vaguely remember that I made this work somehow with a custom build, but I checked `config.toml` again and I don't see any relevant fields. I also tried building rustc with `RUSTFLAGS=-Crelocation-model=pic` and it didn't make any difference. So I think we need to make `cargo build -Zbuild-std` work somehow.

osa1:
> Python scripts and nix lines for the cargo vendor but are copied from baloo/nixos-firmware@e108c50. I will ask the author for permission on using the code. If they don't give permission I will implement the workaround myself.

I implemented the workaround myself, removed copied code.

osa1:
This is ready now @crusso @ggreif @nomeata

ggreif (approved):
LGTM!

Thanks for doing this! First time I _months_ that I can build the RTS on the Mac with ToT.

default.nix
```diff
@@ -32,9 +32,15 @@ let
...
-    xargo
     wasmtime
     rust-bindgen
+    python3
+    tree
+  ] ++ pkgs.lib.optional pkgs.stdenv.isDarwin [
+    libiconv
+    # The following are usually also necessary:
+    # pkgs.darwin.apple_sdk.frameworks.Security
+    # pkgs.darwin.apple_sdk.frameworks.CoreFoundation
```
nomeata, line 43: Well, are they necessary or are they not necessary? :)

default.nix
```diff
@@ -152,46 +158,104 @@ in
...
+        name = "cargo-vendor-tools";
+        src = ./rts/cargo-vendor-tools;
+        cargoSha256 = "0zi3fiq9sy6c9dv7fd2xc9lan85d16gfax47n6g6f5q5c1zb5r47";
+      };
+
+      # Path to vendor-rust-std-deps, provided by cargo-vendor-tools
+      vendorRustStdDeps = "${cargoVendorTools}/bin/vendor-rust-std-deps";
+
+      # Vendors Rust std deps in the output directory
+      rustStdDepsVendor = rust: sha256: nixpkgs.stdenvNoCC.mkDerivation {
```
nomeata, line 172: This function only seems to be used in line 194. So why not just define here
```suggestion
      # Vendors Rust std deps in the output directory
      rustStdDeps = nixpkgs.stdenvNoCC.mkDerivation {
```
without passing `rust` and `rustStdDepsHash`?

default.nix
```diff
@@ -152,46 +158,104 @@ in
...
+          buildCommand = ''
+            mkdir -p $out
+            cp -r ${rtsDeps}/* $out/
+            cp -r ${rustStdDeps}/* $out/
+          '';
+        };
+
+      # All dependencies needed to build the RTS, including Rust std deps, to
+      # allow `cargo -Zbuild-std`. (rust-lang/wg-cargo-std-aware#23)
+      allDeps = mergeDeps rtsDepsUnpacked rustStdDeps;
```
nomeata, line 228: As above, worth defining a function? Why not simply define a derivation and use `rtsDepsUnpacked` and `rustStdDeps` directly?

default.nix
```diff
@@ -32,9 +32,15 @@ let
     llvmPackages_12.bintools
     rustc-nightly
     cargo-nightly
-    xargo
     wasmtime
     rust-bindgen
+    python3
+    tree
```
nomeata, line 38: Leftover from earlier revision?

nomeata (approved):
Great work, thanks for your perseverence!

Didn't look closely at the new rust code, but have some comments about the nix code.

In `nix/drun.nix` we have instructions on how to use `nix-update` to automate updating the sha256s. Not in this PR, but maybe I’ll figure out how to do that with all the hashes we have everywhere, and maybe even write a CI check that updates them.

The new code in `./rts/cargo-vendor-tools` was written by you, not copied from somewhere else? Would it be worth turning that into a separate independent projects, or better not?

default.nix
osa1: nix team said these are often necessary so wanted to leave a hint here. Removed now.

default.nix
osa1: Only reason this is a function is because I don't understand nix. Removed redundant functions now.

default.nix
osa1: Done.

default.nix
osa1: I was using this for debugging. Removed now.

osa1:
> The new code in ./rts/cargo-vendor-tools was written by you, not copied from somewhere else? Would it be worth turning that into a separate independent projects, or better not?

Yeah it's my code. I started implementing it as a reusable package (https://github.com/osa1/cargo-vendor-tools) but then decided to include it in RTS code to give control to Motoko devs. We could still make it a package, but I think it would need to live in dfinity organization and be maintained by the Motoko team.

osa1:
It seems like there's a intermittent failure in some of the tests. Ubuntu job failed with this after the last commit:

```
  tailpositions-actor:skipped (no drun)
   [actor1.drun.comp] [actor2.drun.comp] [drun]
  upgrade-hooks:skipped (no drun)
   [upgrade-hook.drun.comp] [upgrade-hook0.drun.comp] [upgrade-hook1.drun.comp] [upgrade-hook2.drun.comp] [drun]
  upgrades:skipped (no drun)
   [upgrade.drun.comp] [upgrade0.drun.comp] [upgrade1.drun.comp] [upgrade2.drun.comp] [drun]
  Some tests failed:
  stable-mem-nat32.mo stable-mem-nat32.mo
  make: *** [Makefile:4: all] Error 1
  make: Leaving directory '/build/test-run-drun-src/run-drun'
```

The error message is not helpful because it doesn't show the relevant parts. I tried running it a few times locally but it worked fine. I now restarted the job, but it would be good to reproduce the issue and see what's going wrong.

nomeata:
> The error message is not helpful because it doesn't show the relevant parts. I tried running it a few times locally but it worked fine. I now restarted the job, but it would be good to reproduce the issue and see what's going wrong.

Still waiting for https://github.com/Mic92/nix-build-uncached/issues/40 to fix the log display issue :-(

If push comes to shove, we may have to revert #2747. At least for running the tests.

PR #2764, opened 2021-09-06
RTS: remove repr(packed) attributes from RTS structs

osa1:
This is to remove some of the potential undefined behaviors (UB). It will also
remove some of the syntactic noise in #2761.

Rust has alignment restrictions for types and fields beyond the hardware
limitations. This means even on Wasm (which has no alignment restrictions) we
have some restrictions to deal with.

Recently the compiler started to check for some of the obvious sources of UB
(https://github.com/rust-lang/rust/issues/82523). One of these is when taking a
pointer or reference to a `packed` struct. Since `packed` means no padding
between fields, the fields may be unaligned, in which case getting a reference
to them would be UB.

To fix this, this PR removes `packed` attributes and refactors the `Bits64`
type to make sure that the layout is as before. It turns out for types other
than `Bits64` we don't need `packed`: all fields are word sizes so the compiler
does not add any padding. For `Bits64`, we had a `u64` field which is aligned
on 64-bit boundary. To avoid this we now split 64-bit payload into two 32-bit
fields.

A new module `static_checks` added to make sure struct sizes are as expected.
Assertions in this module are checked in compile time.

No extra work needed to align objects. All objects need 4 bytes alignment
(checked with `core::mem::align_of`). The compiler already [aligns static
objects][2], and in runtime we only allocate whole words. So both static and
dynamic objects are always aligned. Debug mode assertions added in GC to check
object alignment.

[1]: https://doc.rust-lang.org/stable/reference/
[2]: https://github.com/dfinity/motoko/blob/59ddaa4520793b6e7038b60e3c30ff267d8415f6/src/codegen/compile.ml#L453


RalfJung:
> Since packed means no padding between fields

It means that, *and* it means that the required alignment of the struct itself is 1. So e.g.
```
#[repr(packed)]
struct My64(u64);
```
Here, `My64` has a required alignment of 1, so if you put it in another struct, no padding will be added.

> Note that we also need to make sure all references and pointers need to be
aligned (should be on word boundary, but I can't find this mentioned in
anywhere in the reference). 

Alignment is not on word boundaries, alignment is a per-type property. So a `&u16` has to always be aligned for the alignment requirement of a `u16` (which is target-specific and can be obtained via `align_of::<u16>`; on x86-64 Linux this is 2), but a `&u64` has to be aligned for the alignment requirement of a `u64` (which will typically be higher; on x86-64 Linux this is 8). All `static` are aligned according to the requirement of their type, similar for all `let`-bound variables and all `Box`. When you create your own heap allocation, the `Layout` passed to the allocation function defines both the required size and alignment of the new allocation.

 When you only write safe code, the compiler already ensures that this is always the case. The documentation of unsafe operations all explicitly say that alignment is required.  If you think there is a place in the reference that should talk more explicitly about alignment, please open an issue!

rts/motoko-rts/src/types.rs
```diff
@@ -540,31 +528,35 @@ impl Concat {
...
 }
 
-#[repr(packed)]
 pub struct Bits64 {
     pub header: Obj,
-    pub bits: u64,
+    // We have two 32-bit fields instead of one 64-bit to avoid aligning the fields on 64-bit
+    // boundary.
+    bits_hi: u32,
+    bits_lo: u32,
```
RalfJung, line 540: An alternative here might be to use `[u8; 8]`, and then do the load with something like `bits.as_ptr().cast::<u64>().read_unaligned()`; that should translate to a single wasm load.

osa1: > I am a bit surprised to see a new getter but not a new setter/constructor that properly encodes data into these fields

These objects are allocated by the generated code and the field is [written using Wasm `i64.store`][1]. We don't need to allocate them in the RTS so we don't have setter or allocation function for it.

Wasm is little endian so we have low bits first, then high bits.

[1]: https://github.com/dfinity/motoko/blob/cd0c3f8b04cf10361e6df7aef476bdf00caaa139/src/codegen/compile.ml#L1484

osa1:
> Alignment is not on word boundaries, alignment is a per-type property. You are already familiar with size being a per-type property: every type has a size, which can differ depending on the target. Exactly the same is true with alignment: every type has an alignment requirement, which can differ depending on the target.
> 
> So a &u16 has to always be aligned for the alignment requirement of a u16 (which is target-specific and can be obtained via align_of::<u16>; on x86-64 Linux this is 2), but a &u64 has to be aligned for the alignment requirement of a u64 (which will typically be higher; on x86-64 Linux this is 8).

I understand that alignmend is type specific, but for pointer types, at least in my testing, alignment seems to be on the word boundary. Example: (types extracted from types.rs)

```rust
use std::mem::align_of;

type Tag = u32;

struct Value(u32);

struct Obj {
    pub tag: Tag,
}

struct Variant {
    header: Obj,
    tag: u32,
    field: Value,
}

fn main() {
    dbg!(align_of::<Obj>());
    dbg!(align_of::<&Obj>());
    dbg!(align_of::<*const Obj>());
    dbg!(align_of::<Variant>());
    dbg!(align_of::<&Variant>());
    dbg!(align_of::<*const Variant>());
    dbg!(align_of::<u8>());
    dbg!(align_of::<&u8>());
    dbg!(align_of::<*const u8>());
    dbg!(align_of::<u16>());
    dbg!(align_of::<&u16>());
    dbg!(align_of::<*const u16>());
    dbg!(align_of::<u32>());
    dbg!(align_of::<&u32>());
    dbg!(align_of::<*const u32>());
    dbg!(align_of::<u64>());
    dbg!(align_of::<&u64>());
    dbg!(align_of::<*const u64>());
}
```

Output when compiled to Wasm:

```
[src/main.rs:18] align_of::<Obj>() = 4
[src/main.rs:19] align_of::<&Obj>() = 4
[src/main.rs:20] align_of::<*const Obj>() = 4
[src/main.rs:21] align_of::<Variant>() = 4
[src/main.rs:22] align_of::<&Variant>() = 4
[src/main.rs:23] align_of::<*const Variant>() = 4
[src/main.rs:24] align_of::<u8>() = 1
[src/main.rs:25] align_of::<&u8>() = 4
[src/main.rs:26] align_of::<*const u8>() = 4
[src/main.rs:27] align_of::<u16>() = 2
[src/main.rs:28] align_of::<&u16>() = 4
[src/main.rs:29] align_of::<*const u16>() = 4
[src/main.rs:30] align_of::<u32>() = 4
[src/main.rs:31] align_of::<&u32>() = 4
[src/main.rs:32] align_of::<*const u32>() = 4
[src/main.rs:33] align_of::<u64>() = 8
[src/main.rs:34] align_of::<&u64>() = 4
[src/main.rs:35] align_of::<*const u64>() = 4
```

We have a one word struct, a 3-word struct, int types from u8 to u64, and all references and pointers need 4 byte alignment. So it seems like pointers and references need to be on word boundary regardless of the pointed type.

Are there any cases where a pointer or reference type needs different alignment depending on the pointed type?

osa1:
> Alignment is not on word boundaries, alignment is a per-type property. You are already familiar with size being a per-type property: every type has a size, which can differ depending on the target. Exactly the same is true with alignment: every type has an alignment requirement, which can differ depending on the target.
>
> So a &u16 has to always be aligned for the alignment requirement of a u16 (which is target-specific and can be obtained via align_of::<u16>; on x86-64 Linux this is 2), but a &u64 has to be aligned for the alignment requirement of a u64 (which will typically be higher; on x86-64 Linux this is 8).

Thanks for pointing this out again. I updated PR description for what we need to do. Object we allocate all need 4 byte alignment, so we need to make sure static objects are aligned on word boundary (4 bytes). In runtime we only allocate words, so once we align heap base on word boundary all allocations will be aligned.

RalfJung:
> So it seems like pointers and references need to be on word boundary regardless of the pointed type.

You deleted that comment so maybe it is not worth replying any more, but just to clear up the understanding: the alignment of `T` is what *pointers/references to `T`* have to uphold. So the alignment of `&i32` is relevant for `&&i32`.

> Object we allocate all need 4 byte alignment, so we need to make sure static objects are aligned on word boundary (4 bytes). In runtime we only allocate words, so once we align heap base on word boundary all allocations will be aligned.

So I take it your allocation happens somewhere outside of the Rust code? Then yes you have to ensure that all your structs are happy with whatever alignment that allocator provides.

osa1:
> You deleted that comment so maybe it is not worth replying any more, but just to clear up the understanding: the alignment of T is what pointers/references to T have to uphold. So the alignment of &i32 is relevant for &&i32. (Or maybe I misunderstood and this is what you meant.)

Yeah I deleted it because I realized because I realized that I misunderstood this part.

> So I take it your allocation happens somewhere outside of the Rust code? Then yes you have to ensure that all your structs are happy with whatever alignment that allocator provides.

Correct.

nomeata:
Hi @RalfJung, how nice to see you here! What brought you to this part of the internet? Dfinity didn't poach you, did they? :-)

rts/motoko-rts/src/types.rs
```diff
@@ -540,31 +528,35 @@ impl Concat {
...
 
-#[repr(packed)]
 pub struct Null {
     pub header: Obj,
 }
 
-#[repr(packed)]
 pub struct Bits64 {
     pub header: Obj,
-    pub bits: u64,
```
crusso, line 551: So was the RTS code actually broken before? Or did the repr(packed) attribute save us?

osa1: It had undefined behavior because alignment of this field was not as expected by the compiler. We don't know any bugs that this caused so far.

`repr(packed)` is to avoid padding between fields, so it potentially breaks alignment. For this object we don't want any padding between `header` and the payload so we either need `repr(packed)` or to split it into two `u32` fields. The former causes undefined behavior because `u64` fields need to be aligned on 8 byte boundaries (as reported by `core::mem::align_of::<u64>()`), so we now use two `u32` fields.

crusso: Ok, thanks for the explanation!

RalfJung:
I don't think they have. ;) @osa1 linked to https://github.com/rust-lang/rust/issues/82523 above and I saw the backlink.

osa1:
Any volunteers to review this? Please check the updated PR description first.

rts/motoko-rts/src/types.rs
osa1: > We don't know any bugs that this caused so far.

I've added some more assertions in the last commits. It turns out all objects are already aligned as expected so we don't have any bugs.

rts/motoko-rts/src/static_checks.rs
```diff
@@ -0,0 +1,53 @@
...
+// TODO: I don't understand why I get a "unused constant" warning for this. Removing it causes
+// compilation failures as expected.
+#[allow(unused)]
+const WORD_SIZE: usize = crate::constants::WORD_SIZE as usize;
+
+// Check platform word size
+const_assert_eq!(size_of::<usize>(), size_of::<u32>());
+const_assert_eq!(size_of::<usize>(), WORD_SIZE);
+
+// Check that sizes of structs are as expected by the compiler
```
nomeata, line 18: Where is the compiler expecting these sizes? Maybe extend the comment with a pointer to where these expectations are currently encoded.

Or is this just a sanity check so that we definitely notice changes to these object sizes, but nothing would break if someone changes the size of one of these?

nomeata (approved):
Looks reasonable to me

rts/motoko-rts/src/static_checks.rs
osa1: Some of these may be implicit in the compiler but for example https://github.com/dfinity/motoko/blob/ede201c60664c3f1ec4e51824ea4e3981b4bd03b/src/codegen/compile.ml#L1446-L1451 expects `Closure`s to be 3 words and https://github.com/dfinity/motoko/blob/ede201c60664c3f1ec4e51824ea4e3981b4bd03b/src/codegen/compile.ml#L2633 expects `Object` header (the `Object` struct in the RTS) to be 3 words.

The idea is if we change the structs in the RTS we'll get a compile error and either update this and the compiler or revert the change. So this checks for accidental changes.

Ideally I think we would have a file that defines object sizes and field offsets, and some kind of preprocessor to generate OCaml and Rust code from that. In GHC we use CPP header files for this. Some other compilers use custom file format and preprocessor.

(I just realized we don't check size of `Object` in here, I'll add it)

nomeata: Ok, maybe a vague
```suggestion
// Check that sizes of structs are as expected by the compiler
// (Expectations are all over the place, e.g. `header_size` definitions in `compile.ml`, calls to `static_closure`, etc.)
```
is helpful.

osa1: Once we update rustc we can also check that fields offsets are as expected by the compiler, using https://crates.io/crates/memoffset. This crate currently cannot be compiled with the rustc version we're using if we want to use it in `const` context.

crusso (approved):
Rubberstamping

PR #2765, opened 2021-09-06
Releasing 0.6.8

ggreif:


ggreif:
@nomeata Looks like the release script has a problem: https://github.com/dfinity/motoko/actions/runs/1206228744

PR #2766, opened 2021-09-07
Remove unnecessary attribute passing in release expression

ggreif:
This might resolve the glitch that threw the last release from the rails: https://github.com/dfinity/motoko/pull/2765#issuecomment-913717691

There was no receive-side match for the argument `internal`, thus the nix call failed.


osa1 (approved):

osa1:
`internal` argument was removed in e440ff21

PR #2768, opened 2021-09-07
Refactor zoo of read/write, load/store StableMem operations to share more code.

crusso:


crusso:
@ggreif I decide to refactor all those (unguarded) read/write and guarded load/store operations to factor out the common bits. See what you think and I'll merge it in to the parent branch.

src/codegen/compile.ml
```diff
@@ -3805,147 +3805,63 @@ module StableMem = struct
...
-        ("offset", I32Type) [I32Type]
-        (fun env get_offset ->
-          Stack.with_words env "temp_ptr" 1l (fun get_temp_ptr ->
-            get_temp_ptr ^^ get_offset ^^ compile_unboxed_const 1l ^^
-            IC.system_call env "ic0" "stable_read" ^^
-            get_temp_ptr ^^
-            G.i (Load {ty = I32Type; align = 0; offset = 0l; sz = Some Wasm.Types.(Pack8, ZX)})))
-    | _ -> assert false
+  let read_word32 env =
+    read env "word32" I32Type 4l load_unskewed_ptr
```
ggreif, line 3836: The common usage pattern is `read env "thingy4xN" ⟨wasmtype⟩ N ⟨dest⟩`, you can exploit that if you want...

ggreif (approved):
Abstraction FTW!

src/codegen/compile.ml
crusso: Life is too short man. 

Was wondering about about pulling the range_check in to avoid the nested function call...

PR #2776, opened 2021-09-10
feat: Stable Log test, example and documentation

crusso:
Simple logging example using StableMemory.mo, for eventual inclusion in doc.


crusso:
@rossberg, @matthewhammer I'm trying to come up with a simple StableMemory example (first written as a motoko test sans base) for use in the doc. See what you think. I like to keep them very short.

test/run-drun/stable-log.mo
```diff
@@ -0,0 +1,73 @@
...
+//SKIP run-low
+//SKIP run-ir
+// too slow on ic-ref-run:
+//SKIP comp-ref
+
+//CALL ingress populate "DIDL\x00\x00"
+//CALL ingress readAll "DIDL\x00\x00"
+//CALL upgrade ""
+//CALL ingress populate "DIDL\x00\x00"
+//CALL ingress readAll "DIDL\x00\x00"
```
crusso, line 73:
```suggestion
//CALL ingress readAll "DIDL\x00\x00"

```

test/run-drun/stable-log.mo
```diff
@@ -0,0 +1,73 @@
...
+    StableMemory.storeBlob(base + 4, blob);
+    base := next;
+  };
+
+  public query func read(offset : Nat32) : async ?(Text, nextOffset : Nat32) {
+    if (offset >= base) return null;
+    let size = StableMemory.loadNat32(offset);
+    let blob = StableMemory.loadBlob(offset + 4, Prim.nat32ToNat(size));
+    let ?text = Prim.decodeUtf8(blob);
+    ?(text, offset + 4 + size)
```
rossberg, line 31: Would be nice to avoid the warning. I think you can use `do?`:
```suggestion
    do? { (Prim.decodeUtf8(blob)!, offset + 4 + size) }
```

test/run-drun/stable-log.mo
```diff
@@ -0,0 +1,73 @@
...
+import StableMemory "stable-mem/StableMemory";
+
+actor {
+
+  func ensure(offset : Nat32) {
+    let pages = (offset + 65536) >> 16;
+      if (pages > StableMemory.size()) {
+        let oldsize = StableMemory.grow(pages - StableMemory.size());
+        assert (oldsize != 0xFFFF);
+     };
```
rossberg, line 11:
```suggestion
    if (pages > StableMemory.size()) {
      let oldsize = StableMemory.grow(pages - StableMemory.size());
      assert (oldsize != 0xFFFF);
    };
```

test/run-drun/stable-log.mo
```diff
@@ -0,0 +1,73 @@
...
+  };
+
+  stable var base : Nat32 = 0;
```
rossberg, line 14: It's funny you're storing this one value in a stable var, instead of the first word of stable memory. :)

test/run-drun/stable-log.mo
```diff
@@ -0,0 +1,73 @@
...
+    let blob = Prim.encodeUtf8(t);
+    let size = Prim.natToNat32(blob.size());
+    let next = base + 4 + size;
+    ensure(next);
+    StableMemory.storeNat32(base, size);
+    StableMemory.storeBlob(base + 4, blob);
+    base := next;
+  };
+
+  public query func read(offset : Nat32) : async ?(Text, nextOffset : Nat32) {
```
rossberg, line 26: This strikes me as unsafe API, since the offset might be completely unaligned. Better be private? (For a public method, I'd rather use an index, or some abstract iteration API).

rossberg (approved):
Great test!

test/run-drun/stable-log.mo
crusso: Right, but how do you enforce abstraction? We don't have any and, for a query, you can't maintain any state on the server.

test/run-drun/stable-log.mo
crusso: That was deliberate, to show you can use both.

test/run-drun/stable-log.mo
crusso: Done

test/run-drun/stable-log.mo
crusso: @rossberg can you think of any solutions?

rossberg: It doesn't matter for this test, I'd just make it private.

For a real-world lib, it would probably need an indirection through an index. The simplest way would be a stable var with a map entry# -> address. But that would be kind of cheating, since then upgrade is O(N). To avoid that, you'd have to fold the map's tree into the stable memory as well. I could imagine ways of doing that, e.g., with index blocks of a fixed size are followed by their actual messages. Each such block can have two successor blocks When the current block fills up, you find the next unused successor in a breadth-first traversal and append a new block. I think this could be done with amortised O(1) append and O(log N) index lookup. Something like that, just thinking aloud...

crusso: Queries can't be private, and that would defeat the point of having a quickly readable log (from the outside at least).
Maybe I could make it a stack and let you query the last n entries...

crusso: I want the example to be simple.

crusso:
@rossberg, @matthewhammer kept the example the same for now, added some doc with caveat.

I could make the api safer if we are embarrassed about that (I'm not really) by just returning a suffix of the log, up to some user provided bound,

rossberg (commented):
Aw damn, wrote a reply yesterday but forgot to submit the review.

crusso:
I'll rewrite to return n last messages, safely. 

However, given that this is a query, apart from reading gibberish, the raw pointer can't actually screw up anything (beyond perhaps executing too many cycles).

rossberg:
Thanks.

> However, given that this is a query, apart from reading gibberish, the raw pointer can't actually screw up anything

Well, yes, until somebody adopts this sort of pattern while also storing other data in stable memory. All that's needed then is a bug in the range check. Would be tragic if we encouraged this dangerous sort of thing.

crusso:
@rossberg One last hurrah, PTAL.

rossberg (approved):
Cool!

test/run-drun/stable-log.mo
matthewhammer: > If you want simple, maybe have read take an index and iterate through the log, and accept that it's O(N) (maybe have the index reversed, as you suggest)?

I believe that is a reasonable cost for the simplicity that @crusso understandably wants here.  Wanting more seems unreasonable without having a more complex representation, which would undermine the value of the example.

> I'd definitely not showcase an unsafe API design that leaks (and accepts!) verbatim backing store addresses.

> For a real-world lib, it would probably need an indirection through an index. 

Yes, I agree.

Just thinking aloud (I expect similar thoughts to others): I would expect that a monotonic log's API would identify "space" in the log using either numbers for the entries, and/or would permit "global offsets" as if the log where a giant text file.

In both cases, the numbers are logical (in terms of the text content), and do not mention any "addresses"; each addressing scheme can be used to build more useful indexes, stored separately, for keyword search, or date search, etc.

(For this simple example, we would accept that those extra features aren't going to exist, and the ability to inspect the log is limited to looking at its tail, or paying more, and looking in O(N) time back an arbitrary `N` entries.)

@rossberg -- Regarding your thoughts toward a more clever (tree) representation where we store more meta data as a balanced tree of some kind.  Rather than design something new or choose something that requires re-balancing at all (this is stable memory and it would be nice to avoid that), my conception of that would be to use a cartesian tree data structure that is balanced in expectation as a it grows, and never rebalances.

- https://en.wikipedia.org/wiki/Cartesian_tree
- I've already [implemented them in Motoko, as a vessel library](https://github.com/matthewhammer/motoko-sequence/blob/e57b88cf4aa4852c7f66b9150692e256911c1425/src/Sequence.mo#L7); we used these in CanCan, though we never tested the scaling properties of them there.
- I learned about these from a [POPL 1989 paper about representations for caching and sharing data](http://matthewhammer.org/courses/csci7000-s17/readings/Pugh89.pdf); the paper calls them "level trees", rather than "cartesian trees".

It's like a balanced rope, except that by adding some extra meta data up front to each internal node, we avoid ever needing to rebalance at all.  Then, we get the time bounds that Andreas mentioned above:
- random query/update accesses from the root are O(log n) time, 
- and query/update to the end can be made in (I think?), amortized O(1) time [by using a variation of the tree that is opened like a zipper](https://arxiv.org/abs/1608.06009).

test/run-drun/stable-log.mo
```diff
@@ -0,0 +1,84 @@
...
+//SKIP run-ir
+// too slow on ic-ref-run:
+//SKIP comp-ref
+
+//CALL ingress populate "DIDL\x00\x00"
+//CALL ingress readAll "DIDL\x00\x00"
+//CALL upgrade ""
+//CALL ingress populate "DIDL\x00\x00"
+//CALL ingress readAll "DIDL\x00\x00"
+//CALL ingress readMore "DIDL\x00\x00"
```
matthewhammer, line 84: I'm curious --- what is the role of `readMore` in the test?

Is the purpose of this to try to read "more than all" and ensure the correct output?

In the test output, the call to `readAll` and `readMore` seem to produce the same list, since `readMore` is reading twice as many entries as ever exist (twice as many as `readAll`).

FWIW, if it were called "readExtra" I would have guessed this intention with more confidence.

matthewhammer (approved):
LGTM.

doc/modules/language-guide/examples/StableLog.mo
```diff
@@ -18,22 +18,28 @@ actor StableLog {
...
+      offset -= 4;
+      let size = StableMemory.loadNat32(offset);
+      offset -= size;
+      let blob = StableMemory.loadBlob(offset, Nat32.toNat(size));
+      switch (Text.decodeUtf8(blob)) {
+        case (?t) { a[k] := t };
+        case null { assert false };
+      }
+    };
+    return Array.tabulate<Text>(count, func i { a[i] });
```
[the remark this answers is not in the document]
crusso: Refactored to have common exit.

test/run-drun/stable-log.mo
crusso: Yup. Renamed to readExtra

crusso:
We'll hold off merging this until StableMemory is, err, stable.

crusso:
Eventual changelog message:

* moc

  * Add runtime support for low-level, direct access to 32-bit IC stable memory (#2626)

* motoko-base

  * Add StableMemory library, exposing 32-bit IC stable memory (#280)

crusso:
@rossberg  PTAL (revised for 64-bit API)

doc/modules/language-guide/pages/stablememory.adoc
```diff
@@ -0,0 +1,112 @@
+# The `StableMemory` library
```
rossberg, line 1: Shouldn't this still say `ExperimentalStableMemory`?

crusso:
```suggestion
# The `ExperimentalStableMemory` library
```

doc/modules/language-guide/pages/stablememory.adoc
```diff
@@ -0,0 +1,112 @@
...
+
+WARNING: The `StableMemory` library is experimental, subject to change and may be
```
crusso, line 3:
```suggestion
WARNING: The `ExperimentalStableMemory` library is experimental, subject to change and may be
```

crusso: Oh god, there's loads ...

PR #2780, opened 2021-09-15
Prim.trap: Expose ic0.trap

nomeata:
the System API provides canisters with a way to explicitly trap with an
error message. This is very useful for developing, e.g. to get useful
messages in code you think is unreachable, or when checking invariants.

This exposes this as `Prim.trap`. Later this can be exposed in
`motoko-base` via `Debug.trap` maybe, and also the odd `Prelude.nyi`,
`Prelude.xxx` and `Prelude.unreachable` can be implemented that way in a
more helpful way.

Semantically this is like the existing `assert false`, just with better
control over the error message and a more helpful type (return type
`None`).

### DIFF
--- a/src/mo_values/prim.ml
+++ b/src/mo_values/prim.ml
@@ -184,6 +184,7 @@ let prim =
                                           | code -> Wasm.Utf8.encode [code]
                                in k (Text str)
   | "print" -> fun _ v k -> Printf.printf "%s\n%!" (as_text v); k unit
+  | "trap" -> fun _ v k -> Printf.printf "%s\n%!" (as_text v); raise (Invalid_argument "explicit trap")
   | "rts_version" -> fun _ v k -> as_unit v; k (Text "0.1")
   | "rts_heap_size" -> fun _ v k -> as_unit v; k (Int (Int.of_int 0))
   | "rts_total_allocation" -> fun _ v k -> as_unit v; k (Int (Int.of_int 0))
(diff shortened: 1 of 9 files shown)


rossberg (approved):

crusso:
Weird, I though I had filed an issue for that. @chenyan-dfinity had requested it a while ago...

nomeata:
Yes, I also thought there was an issue, but couldn't find it

crusso:
Shall I make a next-moc pr? I wouldn't mind cutting a release soon with the new ExperimentalStableMemory Api....

crusso:
Maybe the issue is in base...

nomeata:
> Shall I make a next-moc pr? I wouldn't mind cutting a release soon with the new ExperimentalStableMemory Api....

Yes, please

src/prelude/prim.mo
```diff
@@ -50,6 +50,10 @@ func debugPrintNat(x : Nat) { debugPrint (@text_of_Nat x) };
 func debugPrintInt(x : Int) { debugPrint (@text_of_Int x) };
 func debugPrintChar(x : Char) { debugPrint (charToText x) };
 
+// Trapping
+
+func trap(x : Text) { (prim "trap" : Text -> None) x };
```
crusso, line 55: Shouldn't that be: 
func trap(x : Text) : None { (prim "trap" : Text -> None) x };

crusso:
I think there's a bug...

src/prelude/prim.mo
nomeata: Yes, sorry, my bad

src/mo_values/prim.ml
```diff
@@ -184,6 +184,7 @@ let prim =
                                           | code -> Wasm.Utf8.encode [code]
                                in k (Text str)
   | "print" -> fun _ v k -> Printf.printf "%s\n%!" (as_text v); k unit
+  | "trap" -> fun _ v k -> Printf.printf "%s\n%!" (as_text v); raise (Invalid_argument "explicit trap")
```
crusso, line 187: Hmm, why is this printing and not embedding the message in the exception....?

PR #2787, opened 2021-09-16
Minor refactoring in variant tags in the backend

osa1:
- Rename tag_field -> variant_tag_field, to make it clear that this is
  not the object tag but the variant tag.

- Typo fix: "top of the stacks" -> "top of the stack"

Co-authored-by: Gabor Greif <gabor@dfinity.org>

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -1387,8 +1387,8 @@ module Variant = struct
 
   *)
 
-  let tag_field = Tagged.header_size
-  let payload_field = Int32.add Tagged.header_size 1l
+  let variant_tag_field = Tagged.header_size
+  let payload_field = Int32.add variant_tag_field 1l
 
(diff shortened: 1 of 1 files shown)


nomeata (approved):

src/codegen/compile.ml
```diff
@@ -1387,21 +1387,21 @@ module Variant = struct
 
   *)
 
-  let tag_field = Tagged.header_size
-  let payload_field = Int32.add Tagged.header_size 1l
+  let variant_tag_field = Tagged.header_size
```
ggreif, line 1390: I understood that by putting these identifiers into a `module Variant` namespace we don't need the `variant_` prefix. But I am not requesting a revert as @nomeata is fine with this.

osa1: Even in this namespace we have two notions of tag so I think this clarifies things.

PR #2788, opened 2021-09-16
Use repr(C) in RTS struct

osa1:
See the node added with this commit and
https://github.com/dfinity/motoko/pull/2761#issuecomment-920765586 for
the problem this solves.

### DIFF
--- a/rts/motoko-rts/src/types.rs
+++ b/rts/motoko-rts/src/types.rs
@@ -330,7 +330,7 @@ pub const TAG_ONE_WORD_FILLER: Tag = 16;
 pub const TAG_FREE_SPACE: Tag = 17;
 
 // Common parts of any object. Other object pointers can be coerced into a pointer to this.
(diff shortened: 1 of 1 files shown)


rts/motoko-rts/src/types.rs
```diff
@@ -309,6 +330,7 @@ pub const TAG_ONE_WORD_FILLER: Tag = 16;
 pub const TAG_FREE_SPACE: Tag = 17;
 
 // Common parts of any object. Other object pointers can be coerced into a pointer to this.
+#[repr(C)] // See the note at the beginnign of this module
```
ggreif, line 333:
```suggestion
#[repr(C)] // See the note at the beginnig of this module
```

ggreif: ditto everywhere!

nomeata (approved):
Kudos for a great note

PR #2790, opened 2021-09-17
Add a word for gc metadata in object headers

osa1:
This will be a part of #2706. cc @ggreif

Submitted to get feedback.

Currently life-mut and stable-mutable fail. Other tests pass.

I compared Wasms for life-mut using master and this branch. Changes seem fine, so I think we may need to update a few more places in the compiler or RTS.

The idea is to have a byte in object headers for GC metadata. Initially this will be used to distinguish large object from others in #2706.

I will be moving unrelated changes (refactoring etc.) to separate PRs to keep #2706 as small as possible.


crusso:
Why is the PR still failing?

src/codegen/compile.ml
```diff
@@ -4488,7 +4491,8 @@ module Serialization = struct
           (* Second time we see this *)
```
crusso, line 4491: I think the correct thing to do is to just remove the preceding Sanity checks as they will now have false positives.

nomeata:
Annoying that `nix-build-uncached` does not print the full log files. Reported https://github.com/Mic92/nix-build-uncached/issues/42

src/codegen/compile.ml
```diff
@@ -4488,7 +4491,8 @@ module Serialization = struct
...
           (* Calculate relative offset *)
           let (set_offset, get_offset) = new_local env "offset" in
-          get_tag ^^ get_data_buf ^^ G.i (Binary (Wasm.Values.I32 I32Op.Sub)) ^^
+          get_x ^^ Heap.load_field Tagged.tag_field ^^
+          get_data_buf ^^ G.i (Binary (Wasm.Values.I32 I32Op.Sub)) ^^
           set_offset ^^
           (* A sanity check *)
           get_offset ^^ compile_unboxed_const 0l ^^
```
crusso, line 4498: 'office' below should be `relative offset`

src/codegen/compile.ml
```diff
@@ -4488,7 +4491,8 @@ module Serialization = struct
           (* Second time we see this *)
           (* Calculate relative offset *)
           let (set_offset, get_offset) = new_local env "offset" in
-          get_tag ^^ get_data_buf ^^ G.i (Binary (Wasm.Values.I32 I32Op.Sub)) ^^
+          get_x ^^ Heap.load_field Tagged.tag_field ^^
```
crusso, line 4494: Should the initial load aboce `get_x ^^ Tagged.load ^^ set_tag` also load the full header and read `get_x ^^ Heap.load_field Tagged.tag_field`? 

Note that the offsets need no be even or aligned in any way, AFAIK, since these are just offsets into the blob, not aligned heap addresses, right? Serialized objects are in no way aligned on word boundaries.

osa1: > Note that the offsets need no be even or aligned in any way, AFAIK, since these are just offsets into the blob, not aligned heap addresses, right? Serialized objects are in no way aligned on word boundaries.

Right, Joachim confirmed this in https://github.com/dfinity/motoko/pull/2793#issuecomment-923251202

osa1:
> Why is the PR still failing?

There were compile errors in tests, which I fixed, but compacting gc is still not quite right. Tests in run-drun fail with compacting gc. I'm currently debugging.

src/codegen/compile.ml
osa1: https://github.com/dfinity/motoko/pull/2810

src/codegen/compile.ml
osa1: Maybe I'm confused, but I think those assertions are right. In the size calculation we replace mutable objects with `StableSeen`, so in this pass we shouldn't see mutable objects. All mutable objects should be turned into `StableSeen`s, or their headers should be updated with their offsets in the serialization buffer.

osa1:
This StableSeen business is really tricky, but I think I figured it out. We can't make StableSeen a one-byte tag, it needs to be 32-bit -1 as before. Otherwise we won't be able to distinguish it from offsets, because small values can be valid offsets. 0xFFFFFFFF (-1) cannot be an offset because we can't write an object to that location, for two reasons: (1) that's the end of 4 GiB address space, but we can't allocate a 4 GiB serialization buffer as that's the entire heap and we always have others stuff in the heap (at the very least, the RTS stack). (2) we can't have an object at that address because objects are at least 2 words large.

Keeping StableSeen as 32-bit is fine as the objects overwritten with StableSeen (and later with offsets) won't be traced by the GC, so losing the other fields in the headers is fine. We already rely on this property, nothing new to this PR.

PR #2793, opened 2021-09-20
fix: remove sanity checks as they might trigger false positives

ggreif:
When the reference being written is near the very start of the stable memory
it *could be* `6l` (corresponding to `MutBox`).

Such is case is probably very rare (for realistic programs), but could be engineered.

This PR avoids the conflating of heap tag constants with absolute stable memory
offsets. For references the heap tag field has either been already changed to hold
the absolute offset (by `write_alias`), or has the `StableSeen` tag value as written by `size_alias`.

This is a preparation for #2790 (but also an independent fix).

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -4478,13 +4478,6 @@ module Serialization = struct
         begin
           (* This is a reference *)
           write_byte (compile_unboxed_const 1l) ^^
-          (* Sanity Checks *)
-          get_tag ^^ compile_eq_const Tagged.(int_of_tag MutBox) ^^
-          E.then_trap_with env "unvisited mutable data in serialize_go (MutBox)" ^^
-          get_tag ^^ compile_eq_const Tagged.(int_of_tag ObjInd) ^^
-          E.then_trap_with env "unvisited mutable data in serialize_go (ObjInd)" ^^
-          get_tag ^^ compile_eq_const Tagged.(int_of_tag Array) ^^
-          E.then_trap_with env "unvisited mutable data in serialize_go (Array)" ^^
           (* Second time we see this *)
           (* Calculate relative offset *)
           let (set_offset, get_offset) = new_local env "offset" in


osa1:
I'm a bit confused about this patch. It seems like the assumption here is that for `MutBox` (and `ObjInd`, `Array`) should not be seen here as those would be turned into `StableSeen` in the previous pass, which is what we do in

https://github.com/dfinity/motoko/blob/423b0923eb6f609358778cd0072cc25bf097e475/src/codegen/compile.ml#L4320-L4326

So these assertions look correct to me.

osa1:
However, as you mentioned on Slack, we should make sure that the value of `StableSeen` is not a valid offset to stable memory. I think the only way to make sure is by making it a value that is not a multiple of word size as objects are word aligned.

nomeata:
> However, as you mentioned on Slack, we should make sure that the value of `StableSeen` is not a valid offset to stable memory.

I think that’s why it is (was) `-1`.

(And, to be precise, I think it’s not the offset into stable memory, but rather into the Candid serialization, which is itself 4 bytes into stable memory.)

>  I think the only way to make sure is by making it a value that is not a multiple of word size as objects are word aligned.

The offsets can be arbitrary bytes, as Candid serialization isn’t word-size. So a negative number is probably the safest.

nomeata:
Actually, what we write into the tag isn’t the offset into the candid date buffer either, but actually the _absolute_ memory position of a position in the buffer. We get the offset using
```
          get_tag ^^ get_data_buf ^^ G.i (Binary (Wasm.Values.I32 I32Op.Sub)) ^^
```

So we’d have a problem if a pointer into that dynamically allocated data blob could be the same as these heap tags, which is not possible right now (they’ll always point to the stack).

osa1:
> I think that’s why it is (was) -1.
> ...
> The offsets can be arbitrary bytes, as Candid serialization isn’t word-size. So a negative number is probably the safest.
> ...
> what we write into the tag isn’t the offset into the candid date buffer either, but actually the absolute memory position of a position in the buffer

32-bit -1 is the same as 4294967295 which is a valid address (last location in 4 GiB address space). So I think the reason why this works is because that memory location cannot be used for a `MutBox`, `ObjInd`, or `Array`, as there won't be enough space for the rest of those objects after the header.

nomeata:
Ok you caught me, I admit I willingly disregarded the case that the candid data buffer extends to precisely the last byte in memory, and that a mutable object of size one byte ends up being out precisely there :-)

But I acknowledge that while extremely unlikely, it's not really nice to do it that way.

Which means that small numbers (that end up pointing to the stack area) are actually safer.

ggreif:
Yeah, now I think (went back and examined `advance_data_buf`) that the code is safe. The data buffer lives in the heap (memory `0x0` and at address `> 0xFFFF`) so any absolute address saved into the 32 bits will be distinct from `MutBox`, `ObjInd`, or `Array`'s code. My error was that I thought that the addresses point into a different memory (e.g. stable memory).

Closing.

ggreif:
That said, the sanity checks should be only inserted for "debug instrumentation" as they will never trigger for a correct serialiser. Also they eat up cycles. And they will trip #2790, if it ever gets put into practice.

PR #2794, opened 2021-09-20
Example: Certified (counter) var.

matthewhammer:
@rossberg @crusso @nomeata 

In our video discussion this morning, we discussed an example that used the basic system functionality directly to illustrate a counter.  This is my first attempt at doing that.  PTAL.

### DIFF
--- a/doc/modules/language-guide/examples/CounterCertified.mo
+++ b/doc/modules/language-guide/examples/CounterCertified.mo
@@ -2,33 +2,36 @@
 /// implement the counter value as a certified variable.
 import CD "mo:base/CertifiedData";
 import Blob "mo:base/Blob";
+import Nat8 "mo:base/Nat8";
+import Nat32 "mo:base/Nat32";
 
 actor Counter {
 
   var value : Nat32 = 0;
 
   /// Helper; should be in base?
-  func blobOfNat64(n : Nat32) : Blob { 
+  func blobOfNat32(n : Nat32) : Blob {
     let byteMask : Nat32 = 0xff;
+    func byte(n : Nat32) : Nat8 { Nat8.fromNat(Nat32.toNat(n)) };
     Blob.fromArray(
-      [byteMask << 0 & value,
-       byteMask << 8 & value,
-       byteMask << 16 & value,
-       byteMask << 24 & value])    
+      [byte(byteMask << 0 & value),
+       byte(byteMask << 8 & value),
+       byte(byteMask << 16 & value),
+       byte(byteMask << 24 & value)])
   };
 
   /// Update counter and certificate (via system).
-  public func inc() : async Nat64 {
+  public func inc() : async Nat32 {
     value += 1;
-    CD.set(value);
-    return value; 
+    CD.set(blobOfNat32(value));
+    return value;
   };
-  
+
   /// Returns the current counter value,
   /// and an unforgeable certificate (from the system) about its authenticity.
-  public query func get() : async ?{ value : Nat64; certificate : Blob } {
+  public query func get() : async ?{ value : Nat32; certificate : Blob } {
     do ? {
-      (value, CD.getCertificate()!)
+      { value; certificate = CD.getCertificate()! }
     }
   };
 }
--- a/doc/modules/language-guide/examples/CounterCertified.mo
+++ b/doc/modules/language-guide/examples/CounterCertified.mo
@@ -1,0 +1,34 @@
+/// Simple counter (see `Counter.mo`), but uses `mo:base/CertifiedData` to
+/// implement the counter value as a certified variable.
+import CD "mo:base/CertifiedData";
+import Blob "mo:base/Blob";
+
+actor Counter {
+
+  var value : Nat32 = 0;
+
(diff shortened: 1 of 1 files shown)


doc/modules/language-guide/examples/CounterCertified.mo
```diff
@@ -0,0 +1,37 @@
...
+import CD "mo:base/CertifiedData";
+import Blob "mo:base/Blob";
+import Nat8 "mo:base/Nat8";
+import Nat32 "mo:base/Nat32";
+
+actor Counter {
+
+  var value : Nat32 = 0;
+
+  /// Helper; should be in base?
```
nomeata, line 12: You can use Claudio’s new stable memory functions for this :-D

doc/modules/language-guide/examples/CounterCertified.mo
```diff
@@ -0,0 +1,37 @@
...
+    return value;
+  };
+
+  /// Returns the current counter value,
+  /// and an unforgeable certificate (from the system) about its authenticity.
+  public query func get() : async ?{ value : Nat32; certificate : Blob } {
+    do ? {
+      { value; certificate = CD.getCertificate()! }
+    }
+  };
```
nomeata, line 36: I’d do explicit error handling here, to help readers understand when `getCertificate` returns `none`.

In fact, for best practice we want `get` to be callable via inter-canister call / update call, and then `getCertificate` fails, but it’s not really an error.

So maybe simply
```suggestion
  /// Returns the current counter value,
  /// and, if available, an unforgeable certificate (from the system) about its authenticity.
  /// When called via update call or inter-canister call, no certificate is present (and not needed,
  /// as in these cases the system already certifies the response)
  public query func get() : async { value : Nat32; certificate : ?Blob } {
    return { value; certificate = CD.getCertificate() }
  };
```

Sorry for not getting to show off `do ? {…}` :-)

crusso (commented):
Looks good. 

I guess it's not actually tested in the examples folder, thought there might be ./check.sh that you can run to see if it type checks.

Once you've got working, maybe rewrite as a test in run-drun (probably removing dependency on base unless we now allow that) and run with ./run.sh -d <name>.mo

I usually start the other way round. Write the test and then turn it into an example... so I can run while developing.

nomeata:
BTW, this is only half as interesting without the frontend. I thought we have a separate GitHub repo for full example projects, including TypeScript and HTML code? Maybe it should live there?

matthewhammer:
> I guess it's not actually tested in the examples folder, thought there might be ./check.sh that you can run to see if it type checks.

Yes, the type-check runs, but no testing.  I had wondered if it should start as a test, or otherwise how to run the examples.

> I usually start the other way round. Right the test and then turn it into an example... so I can run while developing.

Okay, makes sense.  Will do.

matthewhammer:
> BTW, this is only half as interesting without the frontend. I thought we have a separate GitHub repo for full example projects, including TypeScript and HTML code? Maybe it should live there?

Yeah, in this case, even the tests that I do won't be sufficient without inspecting the certificate (can our test system do that?)

@nomeata Would you prefer that this be entirely removed and moved to the separate examples repo, or have a more developed version there, and a less developed (sans frontend) version here?

nomeata:
> can our test system do that?

Nope. I think the only validation code in existence are the agents (JS agent and rust agent), and the Haskell code in `ic-hs`.

> Would you prefer that this be entirely removed and moved to the separate examples repo, or have a more developed version there, and a less developed (sans frontend) version here?

Dunno. I think the full-stack version is more interesting, as there is little value in learning just the Motoko side. Especially as we have this tight coupling here due to the application-specific certification attempt. The Motoko manual can certainly include snippets of that and explain that and maybe point to the full thing. But https://github.com/dfinity/examples/ seems like a more natural home.

matthewhammer:
> But https://github.com/dfinity/examples/ seems like a more natural home.

I see your other points about it not belonging where it is, but I don't see that that repo (currently) seems natural (yet) -- If so, where?

For instance, when I look [here (under `motoko`)](https://github.com/dfinity/examples/tree/master/motoko), I don't see any existing examples with accompanying JS or frontend code.  Though, maybe there _should_ be some!

doc/modules/language-guide/examples/CounterCertified.mo
matthewhammer: Sure, but that's yet another experimental feature.

And perhaps just as important, I've been told repeatedly that these features are to be thought of as orthogonal, so we best not show that even our simplest examples mix them, lest we confuse people about how they interact?

FWIW, no strong feeling on my part about this.

nomeata:
Yeah, the examples repo is probably not perfect. Ideally, one generates a nice browsable documentation from there, with nicely annotated and explained code, integration into Motoko playground etc. Maybe hire some technical writers and tech doc site programmers…

crusso:
The motoko repo doc could just include snippets of code and link to the dfinity/examples repo.

matthewhammer:
Moving to https://github.com/dfinity/examples/pull/119

PR #2804, opened 2021-09-22
Use real tags in mark stack tests

osa1:
Previously for the tags in mark stack tests we were using `value - 1`
where `value` is the value being pushed to the stack.

Using actual tags in these tests is more resilient to changes so we now
use actual tags in mark stack tests.


nomeata (approved):
Rubber stamp

PR #2809, opened 2021-09-23
Refactor ascriptions (`as`) in RTS, avoid accessing Bytes/Words fields

osa1:
Ascriptions can be lossy, so every usage of `as` can potentially cause
problems as we refactor the code.

So in this PR we reduce the usage of `as`:

- For `Bytes` and `Words` values, instead of `x.0 as usize` we now use
  `x.as_usize()`. `usize` has the same size as `u32` on Wasm (which is
  the only target we support), so `as_usize()` method uses `as` without
  any checks in debug mode.

- For casting `u32` to `u64`, use `u64::from(x)`. `from` methods cannot
  fail, and for numeric types they are only implemented for casting a
  smaller type to a larger type.

- To aid readbility this PR also introduces `Words::as_u32` and
  `Bytes::u32` and uses these methods instead of accessing the unnamed
  field of `Words` and `Bytes`.

- Also remove a redundant unwrapping/wrapping of `Bytes` value when
  dividing it by a scalar. (`Bytes(bytes.0 / x)` -> `bytes / x`)

### DIFF
--- a/rts/motoko-rts-tests/src/gc.rs
+++ b/rts/motoko-rts-tests/src/gc.rs
@@ -167,7 +167,7 @@ fn check_dynamic_heap(
             check_continuation_table(object_offset, continuation_table, heap);
             offset += (size_of::<Array>() + Words(continuation_table.len() as u32))
                 .to_bytes()
-                .0 as usize;
+                .as_usize();
             continue;
         }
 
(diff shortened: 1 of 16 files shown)


crusso:
@ggreif, can you review? I'm busy with doc today.

PR #2811, opened 2021-09-23
Codegen: remove unused instructions in size_alias

osa1:


osa1:
I think this line was a copy-paste error

nomeata (commented):
Hmm, that's the argument to the Or two lines down, isn't it?

osa1:
Hm, you are right. Interestingly I didn't get any errors when running the Wasm generated with this change. I would expect this to break the stack by popping one more value than before.

nomeata:
That is odd indeed, worth investigating. Are we exiting this via return? … No. Weird.

ggreif:
I checked this code a few days ago, it is okay. One could perform 4x `get_tag ^^ compile_eq_const` and the 3x `Or`. But I think it is fine as-is.

PR #2816, opened 2021-09-24
Use convenience synonyms for simple constants in `compile.ml`

ggreif:


nomeata (commented):
I know this is a matter of taste, but I would actually advice against them (despite having introduced them). They are not much shorter, and I find it clearer to see `0` or `1` there. Recently I tried to use `compile_unboxed_zero` only when the intuition is not really a ”number”  (such as “what do we compile the value `null` to). But this is all just unhelpful mental overhead, so maybe simply get rid of `_zero` and `_one`.

To combat line noise, we could try to find a better convention for “definition that geneates wasm”  than `compile_` (Maybe `c_`?), and then maybe write something like `c_i32` instead? The whole `compile_` stuff is quite noisy anyways.

(But don’t consider this a veto if you think it improves things, you probably have to work with the code more than I do now…)

ggreif:
What about `s/compile_unboxed_zero/compile_unboxed_0/g` in a subsequent PR? @nomeata

nomeata:
Sure, why not

luc-blaeser:

PR #2821, opened 2021-09-27
Reproducibility: Make assertion error messages reproducible

crusso:
Our builds are not reproducible because the filepaths reported in assert messages are build environment dependent.

A simple fix is to only report the basename of the file in the assertion message, though this loses some information and may be ambiguous if a  project imports two distinct files with the same basename (this is possible, but not that likely, given the current scale of motoko projects.)

A better approach my be to relativize the source file to the current main file or package, but this probably requires pervasive changes to our Source.ml region representation, or other measures, like storing the root dir in the compilation unit.

TODO:

- [ ] normalize to base name in interpretation of AssertE (both source and IR interpreters), if we care about that level of fidelity.

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -3655,6 +3655,12 @@ module IC = struct
     Tuple.from_stack env 2
 
   let fail_assert env at =
+    let open Source in
+    let at = {
+        left = {at.left with file = Filename.basename at.left.file};
+        right = {at.right with file = Filename.basename at.right.file}
+      }
+    in
     E.trap_with env (Printf.sprintf "assertion failed at %s" (string_of_region at))
 
   let async_method_name = "__motoko_async_helper"
--- a/test/run-drun/assert.mo
+++ b/test/run-drun/assert.mo
@@ -1,0 +1,9 @@
+import Lib "assert/Lib";
+actor a {
+
+  public func go() : async () {
+    Lib.fail();
+  };
+
+};
+a.go(); //OR-CALL ingress go "DIDL\x00\x00"
--- a/test/run-drun/assert/Lib.mo
+++ b/test/run-drun/assert/Lib.mo
@@ -1,0 +1,8 @@
+module {
+
+  public func fail() : () {
+    assert false;
+  };
+
+};
+
--- a/test/run-drun/ok/assert.diff-ir.ok
+++ b/test/run-drun/ok/assert.diff-ir.ok
@@ -1,0 +1,5 @@
+--- assert.run
++++ assert.run-ir
+@@ -1 +1 @@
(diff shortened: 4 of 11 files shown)


nomeata (approved):
Reasonable way forward. Relative to project or package root is probably better, but maybe not worth the complexity at this point.

osa1:
> A better approach my be to relativize the source file to the current main file or package

I agree. How do we show paths from imported libraries currently? I'm guessing something like  `../../../motoko-base/src/Hash.mo:123:456 ...`? If that's the case then I think this PR could be a step backwards, because just showing `Hash.mo:...` will probably be confusing: user may have a file with the same name in the project, or may get confused about where that file is.

I wonder if we could also refactor the `region` type a little bit:

```
type pos = {file : string; line : int; column : int}
type region = {left : pos; right : pos}
```

The problem is this allows invalid region representations where start and end of the region have different files.

rossberg (approved):
Agreed this is a regression yet okayish for now. I suspect that a nicer fix would be rather non-trivial. (It's perhaps representative of the kind of rabbit holes you get into with the goal of reproducible builds. ;) )

crusso:
> > A better approach my be to relativize the source file to the current main file or package
> 
> I agree. How do we show paths from imported libraries currently? I'm guessing something like `../../../motoko-base/src/Hash.mo:123:456 ...`? If that's the case then I think this PR could be a step backwards, because just showing `Hash.mo:...` will probably be confusing: user may have a file with the same name in the project, or may get confused about where that file is.
> 
> I wonder if we could also refactor the `region` type a little bit:
> 
> ```
> type pos = {file : string; line : int; column : int}
> type region = {left : pos; right : pos}
> ```
> 
> The problem is this allows invalid region representations where start and end of the region have different files.

Yeah, I'm not sure why its like this - the code was copied from the wasm reference implementation I think, except perhaps to avoid even more allocation when constructing a region or projecting a position.

rossberg:
I reused the definition of source region from an older project of mine. The definition is so because there might have been a file inclusion mechanism, such that left and right of a region may indeed be in separate files.

oggy-dfin:
@crusso confirmed that this fixes the issue compared to the `moc` from the old master

skilesare:
Copying this from a conversation with Claudio on the Forum

Privacy issue about the motoko compiler. If you go to Principal wm2oc-maaaa-aaaah-qb47a-cai  on https://ic.rocks/principal/wm2oc-maaaa-aaaah-qb47a-cai and look at the getFiles function, it will fail. The error message leaks information about the person who compiled the module. Who is nandit? I don’t know, but perhaps they do not want their name in the error message? You can move to a different folder without personal info, but maybe people don’t know that. 

Some kind of relative path make sense here...bonus if it makes it deterministic for all compilers.

PR #2825, opened 2021-10-01
Implement type union/intersection

### LINKED ISSUE #1841: Add syntax for type intersection and union
Motoko supports width subtyping, but no way to incrementally express object or variant types. In practice, that means that parts of type definitions have to be repeated:
```
type Point = {
  x : Float;
  y : Float;
};

type ColoredPoint = {
  x : Float;
  y : Float;
  color : Color;
};
```

It would be desirable to have a shorthand for expressing an extension of another object or variant type. For example:
```
type Point = {
  x : Float;
  y : Float;
};

type ColoredPoint = Point and {
  color : Color;
};
```
The same mechanism ought to be available for variants:
```
type Color = RGB or {#black; #white};
```
Note that extension creates a subtype for objects but a supertype for variants.

The natural way to enable this is to provide operators expressing type intersection (`and`) and union (`or`), which compute the glb and lub of two type, respectively. In order to avoid full-blown intersection/union types, they would have to be restricted to concrete (or equivalent) types, i.e., `X and {y : Nat}` would not be allowed if `X` was an abstract type variable.

rossberg:
Fixes #1841.

### DIFF
--- a/doc/modules/language-guide/pages/language-manual.adoc
+++ b/doc/modules/language-guide/pages/language-manual.adoc
@@ -605,6 +605,8 @@ Type expressions are used to specify the types of arguments, constraints (a.k.a 
   ( ((<id> :)? <typ>),* )                       tuple
   Any                                           top
   None                                          bottom
+  <typ> and <typ>                               intersection
+  <typ> or <typ>                                union
   Error                                         errors/exceptions
   ( <typ> )                                      parenthesized type
 
@@ -901,6 +903,22 @@ Type `None` is the _bottom_ type, a subtype of all other types.
 No value has type `None`.
 
 As an empty type, `None` can be used to specify the impossible return value of an infinite loop or unconditional trap.
+
+[[and-type]]
+=== Intersection type
+
+The type expression `<typ1> and <typ2>` denotes the syntactic _intersection_ between its two type operands, that is, the greatest type that is a subtype of both. If both types are incompatible, the intersection is `None`.
+
+The intersection is _syntactic_, in that it does not consider possible instantiations of type variables. The intersection of two type variables is `None`, unless they are equal, or one is declared to be a (direct or indirect) subtype of the other.
+
+
+[[or-type]]
+=== Union type
+
+The type expression `<typ1> or <typ2>` denotes the syntactic _union_ between its two type operands, that is, the smallest type that is a supertype of both. If both types are incompatible, the intersection is `Any`.
+
+The union is _syntactic_, in that it does not consider possible instantiations of type variables. The union of two type variables is the union of their bounds, unless the variables are equal, or one is declared to be a (direct or indirect) subtype of the other.
+
 
 [[paren-type]]
 === Parenthesized type
--- a/src/mo_def/arrange.ml
+++ b/src/mo_def/arrange.ml
@@ -178,6 +178,8 @@ and typ t = match t.it with
   | TupT ts -> "TupT" $$ List.concat_map typ_item ts
   | FuncT (s, tbs, at, rt) -> "FuncT" $$ [func_sort s] @ List.map typ_bind tbs @ [ typ at; typ rt]
   | AsyncT (t1, t2) -> "AsyncT" $$ [typ t1; typ t2]
+  | AndT (t1, t2) -> "AndT" $$ [typ t1; typ t2]
+  | OrT (t1, t2) -> "AndT" $$ [typ t1; typ t2]
   | ParT t -> "ParT" $$ [typ t]
   | NamedT (id, t) -> "NamedT" $$ [Atom id.it; typ t]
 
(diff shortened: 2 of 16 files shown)


src/mo_def/arrange.ml
```diff
@@ -178,6 +178,8 @@ and typ t = match t.it with
   | TupT ts -> "TupT" $$ List.concat_map typ_item ts
   | FuncT (s, tbs, at, rt) -> "FuncT" $$ [func_sort s] @ List.map typ_bind tbs @ [ typ at; typ rt]
   | AsyncT (t1, t2) -> "AsyncT" $$ [typ t1; typ t2]
+  | AndT (t1, t2) -> "AndT" $$ [typ t1; typ t2]
+  | OrT (t1, t2) -> "AndT" $$ [typ t1; typ t2]
```
crusso, line 182:
```suggestion
  | OrT (t1, t2) -> "OrT" $$ [typ t1; typ t2]
```

crusso (commented):
This looks lovely, but haven't had a chance to review properly yet. Maybe tomorrow.

doc/modules/language-guide/pages/language-manual.adoc
```diff
@@ -902,6 +904,22 @@ No value has type `None`.
...
+
+The type expression `<typ1> and <typ2>` denotes the syntactic _intersection_ between its two type operands, that is, the greatest type that is a subtype of both. If both types are incompatible, the intersection is `None`.
+
+The intersection is _syntactic_, in that it does not consider possible instantiations of type variables. The intersection of two type variables is `None`, unless they are equal, or one is declared to be a (direct or indirect) subtype of the other.
+
+
+[[or-type]]
+=== Union type
+
+The type expression `<typ1> or <typ2>` denotes the syntactic _union_ between its two type operands, that is, the smallest type that is a supertype of both. If both types are incompatible, the intersection is `Any`.
```
crusso, line 918:
```suggestion
The type expression `<typ1> or <typ2>` denotes the syntactic _union_ between its two type operands, that is, the smallest type that is a supertype of both. If both types are incompatible, the union is `Any`.
```

src/mo_frontend/typing.ml
```diff
@@ -476,6 +476,28 @@ and check_typ' env typ : T.typ =
       (List.map (fun (field : typ_field) -> field.it.id) fields);
     let fs = List.map (check_typ_field env sort.it) fields in
     T.Obj (sort.it, List.sort T.compare_field fs)
+  | AndT (typ1, typ2) ->
+    let t1 = check_typ env typ1 in
+    let t2 = check_typ env typ2 in
+    let t = T.glb t1 t2 in
```
crusso, line 482: Can't this lead to non-termination if t1 or t2 is sibling recursive type that isn't known to be contractive yet?

What does:
 "type A = A and/or A"
 produce?

crusso: (Type elaboration used to be syntactic with some checks, now it ain't.)

rossberg: Good question. Intersection and union are not productive themselves, so this produces:
```
type error [M0157], block contains non-productive definition A
```
I think this works fine because we do that check before elaborating the types. I added a couple of simple tests.
> (Type elaboration used to be syntactic with some checks, now it ain't.)
True, is that a problem?

rossberg: Ah, okay, this example happens to work, because in the prepass a named type is simply represented as its abstract name and union/intersection shortcuts the reflexive case.

Here is a problematic case:
```
type T = U and V; type U = {}; type V = {};
```
This crashes with an assertion failure, because glb computation needs to expand the type id and runs into type `Pre` in the prepass. Hm...

crusso: Bingo, thanks for persevering.

rossberg: Yeah, I thought about making them more primitive, but then I'd rather bite and make them proper intersection and union types.

Another option I considered is to check that the operands aren't Pre, and if they are, give an "cannot compute intersection on forward types" error, similar to what we do for type inference in similar cases.

rossberg: Okay, I implemented the latter. Extending our type checker's "incompleteness" that way seems okay given that we're already incomplete in similar ways. And it's conservative, so we can do smarter later if this ever turns out to be relevant in practice.

rossberg:
Thanks! Changed error message to "recursive or forward references", since the error doesn't actually depend on recursion (cf. my counter example). Also added a couple of recursive examples to test.

FWIW, order does matter here, even in the recursive case, because type definitions are still computed in order. This is rejected:
```
type B = A and {}; type A = {a : B};
```
while this is accepted:
```
type A = {a : B}; type B = A and {};
```
I know, it's a bit unsatisfactory... Hope we get to turn these into proper union/intersection types eventually.

jzxchiang1:
Can this be added to the official SDK [documentation](https://smartcontracts.org/docs/language-guide/overview.html)? It's a really useful feature.

PR #2829, opened 2021-10-06
RTS build: don't vendor Rust deps

osa1:
Vendoring dependencies is causing a lot of problems in #2761 and other
PRs, because `cargo vendor` currently has a bug and cannot vendor
standard dependencies (deps of alloc and std, see
https://github.com/rust-lang/wg-cargo-std-aware/issues/23).

For a long time I thought vendoring is a necessity and tried to work
around this issue by vendoring dependencies manually, or using
https://github.com/nix-community/naersk.

However, it turns out vendoring is not necessary, and we can download
dependencies in build step just fine. I also don't see any advantages of
vendoring the dependencies. So in this PR we remove vendoring.

Unblocks #2761.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -37,6 +37,7 @@ let
     rust-bindgen
     rustfmt
     cacert
+    git
   ];
 
   llvmEnv = ''
--- a/default.nix
+++ b/default.nix
@@ -176,6 +176,7 @@ rec {
       preBuild = ''
         export XARGO_HOME=$PWD/xargo-home
         export CARGO_HOME=$PWD/cargo-home
+        export CARGO_HTTP_MULTIPLEXING=false
 
         ${llvmEnv}
         export TOMMATHSRC=${nixpkgs.sources.libtommath}
--- a/default.nix
+++ b/default.nix
@@ -36,6 +36,7 @@ let
     wasmtime
     rust-bindgen
     rustfmt
+    cacert
   ];
 
   llvmEnv = ''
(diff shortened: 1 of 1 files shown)


osa1:
Weird, CI passed on macos, but not on Ubuntu. Maybe a problem with Ubuntu CI configuration?

nomeata:
Downloading stuff in nix builds is of course a no-go; how else would you achieve hermeticity - the content of the URL could change, making the build nonreproducable.

It may appear to work in some system configurations that don't sandbox the build (e.g. on Darwin CI, it seems), but it (rightfully) will not work on properly sandboxed builds. There is a reason naersk etc exists and does this step where it downloads the dependencies first to a “fixed output derivation”, where the content (hash) is declared ahead of time, and which therefore are allowed to do unsafe things.

osa1:
> Downloading stuff in nix builds is of course a no-go; how else would you achieve hermeticity - the content of the URL could change, making the build nonreproducable.

We already download stuff from crates.io today. It just happens in `preBuild` step rather than `build` step. If a crate contents change in crates.io (which I think should never happen, and checksums in `Cargo.lock` are to check this) we will have all of the same problems in any case.

Why is it a problem to download in `build` step but not in `preBuild` step?

nomeata:
I would be very surprised if that makes a difference, these steps are just different parts of essentially one big shell script. . Are you sure it's not the difference between a normal build and a fixed-output build (like the normal vendoring build step)?

nomeata:
Some docs: https://nixos.org/manual/nix/stable/#fixed-output-drvs

osa1:
> But the vendoring build has, as part of it's build instructions, the hash of the result

OK, this makes sense.

I wonder if naersk uses `cargo vendor`. If it does then we can't use it to work around the `cargo vendor` bug.

PR #2831, opened 2021-10-07
feat: transform `for`-loops on arrays into `while`-loops

ggreif:
## How it works

- we transform AST `ForE` loops by `Construct.whileE` _while_ lowering to IR
- the transformation kicks in when a `DotE` expression is applied to a value with an array type, by projecting out the `vals` accessor
- we get hold of the expression in the `DotE` node and use that (i.e. its constituents) as the starting point to obtain the array's size, and then index into it (for `keys` accessor the indexing is trivial)
- we build a classic `while` loop with an index running from zero below the size
- we don't distinguish between immutable/mutable arrays and also cater for potentially effectful loop bodies, as well as array expressions (and unit expressions passed to the call)
- we always CBV-bind the array expression to a variable (even if it is already a `VarE`, otherwise we end up in name capture hell)

## Pain points

We lose a bunch of source locations in the process of conversion to IR. This is not really a problem now, but will make the debugging experience less enjoyable some day.
 
## TODOs:
- [x] `FileCheck` on final IR
- [x] `async` (talk to @crusso)
- [x] `S.`-only transformation?
- [x] Are parens interfering? — No
- [x] test all combinations
- [ ] check why we have a perf regression

- eliminate unknown call following `call $@(im)mut_array_size` — See #2882
- use `u32` arithmetic and indexing (currently _bignum_ calls)
- [x] we could optimise `DotE(..., "keys")` similarly — DONE

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,7 +2,7 @@
 
 * `for` loops over arrays are now converted to more efficient
   index-based iteration (#2831). This can result in significant cycle
-  savings for tight loops.
+  savings for tight loops, as well as slightly less memory usage.
 
 * Add type union and intersection. The type expression
 


src/ir_def/construct.ml
```diff
@@ -600,14 +607,40 @@ let loopWhileE exp1 exp2 =
         )
     )
 
+let countingE binding arr expSize expIndexing expWorker =
```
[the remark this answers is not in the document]
ggreif: of course if that is _already_ a `VarE`, we could reuse it.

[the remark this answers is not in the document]
ggreif: The latter is implemented in 2d00637.

test/run/ok/iter-no-alloc.wasm-run.ok
```diff
@@ -2,7 +2,7 @@ Allocation per iteration (Nat): 0
 Allocation per iteration (Nat16): 0
 Allocation per iteration (Nat32): 0
 Allocation per iteration (?Nat, all null): 0
-Allocation per iteration (FixOpt, all ?null): 8
-Allocation per iteration (FixOpt, all ??null): 8
+Allocation per iteration (FixOpt, all ?null): 0
```
ggreif, line 5: Allocation is removed, good!

test/run-drun/ok/optimise-for-array-eff.ic-ref-run.ok
```diff
@@ -0,0 +1,18 @@
+→ update create_canister(record {settings = null})
+← replied: (record {hymijyo = principal "cvccv-qqaaq-aaaaa-aaaaa-c"})
```
ggreif, line 2: @nomeata just out of curiosity: why `hymijyo`?

nomeata: https://github.com/nomeata/haskell-candid/blob/master/src/Codec/Candid/Hash.hs#L24-L45

ggreif: Fun :-) Thanks!

nomeata:
> There is a failure that Hydra caught, but not the GitHub action. I'd be more comfortable to disable Hydra when the observations match.

Oh, is there? That’s odd… what is run by hdyra that’s not run by the other CI? Do let me know; I can’t see hydra logs.

src/lowering/desugar.ml
```diff
@@ -239,6 +242,77 @@ and lexp' = function
   | S.IdxE (e1, e2) -> I.IdxLE (exp e1, exp e2)
   | _ -> raise (Invalid_argument ("Unexpected expression as lvalue"))
 
+and rewrite_for_to_while p arr proj c0 c1 c2 e1 e2 =
```
crusso, line 245: I think this code would be much simpler and maintainable if you could use the helpers in Construct.ml that compute the types and effect for you. Helpers like Construct.letP,  letD, callE, blockE etc. Is there any reason you can't?

src/lowering/desugar.ml
```diff
@@ -198,6 +198,9 @@ and exp' at note = function
   | S.WhileE (e1, e2) -> (whileE (exp e1) (exp e2)).it
   | S.LoopE (e1, None) -> I.LoopE (exp e1)
   | S.LoopE (e1, Some e2) -> (loopWhileE (exp e1) (exp e2)).it
+  | S.(ForE (p, ({it=CallE (({it=DotE (arr, proj); _} as c0), c1, c2); _} as e1), e2))
+      when T.is_array arr.note.S.note_typ && proj.it = "vals"
+    -> (exp (rewrite_for_to_while p arr proj c0 c1 c2 e1 e2)).it
```
crusso, line 203: Why call exp on the result of rewrite_for_to_while, which requires you to generate Source and not just directly call rewrite_for_to_while and have it generate the IR using the helpers in Construct.ml (plus any more you want to add).

crusso: (in the style of S.While (e1, e2) -> (whileE (exp e1) (exp e2)).it above).

The final .it project is ugly but only a small price to pay.

src/ir_def/construct.ml
```diff
@@ -91,6 +91,7 @@ let primE prim es =
     | ICStableWrite _ -> T.unit
     | IcUrlOfBlob -> T.text
     | ActorOfIdBlob t -> t
+    | BinPrim (t, _) -> t
```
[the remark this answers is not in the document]
ggreif: Well, it is back, because this code is now invoke due to building IR by `primE`, and I am generating addition nodes in IR now. We get an assertion failure when taking this out:
```
Fatal error: exception "Assert_failure ir_def/construct.ml:99:11"
Raised at Ir_def__Construct.primE in file "ir_def/construct.ml", line 99, characters 11-23
Called from Lowering__Desugar.sequentialForE.body in file "lowering/desugar.ml", line 353, characters 15-146
Called from Lowering__Desugar.sequentialForE in file "lowering/desugar.ml", line 361, characters 11-89
Called from Lowering__Desugar.exp' in file "lowering/desugar.ml", line 204, characters 147-202
Called from Lowering__Desugar.typed_phrase' in file "lowering/desugar.ml", line 45, characters 16-30
Called from Lowering__Desugar.dec' in file "lowering/desugar.ml", line 616, characters 22-29
Called from Lowering__Desugar.phrase' in file "lowering/desugar.ml", line 41, characters 16-34
Called from Lowering__Desugar.dec in file "lowering/desugar.ml", line 613, characters 14-30
Called from Stdlib__list.map in file "list.ml", line 92, characters 20-23
Called from Lowering__Desugar.transform_unit_body in file "lowering/desugar.ml", line 936, characters 26-35
Called from Lowering__Desugar.transform_unit in file "lowering/desugar.ml", line 974, characters 14-38
Called from Pipeline.desugar_unit in file "pipeline/pipeline.ml", line 542, characters 4-22
Called from Pipeline.compile_unit in file "pipeline/pipeline.ml", line 626, characters 16-43
Called from Pipeline.compile_files in file "pipeline/pipeline.ml", line 647, characters 14-53
Called from Diag.bind in file "lang_utils/diag.ml", line 32, characters 27-30
Called from Diag.bind in file "lang_utils/diag.ml", line 32, characters 27-30
Called from Moc.process_files in file "exes/moc.ml", line 152, characters 37-82
Called from Moc in file "exes/moc.ml", line 214, characters 4-23
```
Nothing nefarious going on here.

src/lowering/desugar.ml
ggreif: I can, and it turns out to be much less hassle. I think when I started out this way I shouldn't have ripped the logic into two parts (in two source files), then I wouldn't have felt the urge to do the transformation in syntax-land only. PTAL.

src/lowering/desugar.ml
ggreif: Done.

ggreif:
I am looking at the perf data:
``` C
gas/assetstorage;13806
size/assetstorage;166100
gas/qr;1238450504
size/qr;221189
gas/reversi;976259
size/reversi;159625
```
On `master` we have
``` C
gas/assetstorage;13806
size/assetstorage;166100
gas/qr;1238449915
size/qr;221010
gas/reversi;1301469
size/reversi;160571
```

src/codegen/compile.ml
```diff
@@ -6053,7 +6053,7 @@ module Internals = struct
     | Some (VarEnv.Const(_, Const.Fun mk_fi)) ->
        compile_unboxed_zero ^^ (* A dummy closure *)
        G.i (Call (nr (mk_fi ())))
-    | _ -> assert false
+    | _ -> failwith var
```
nomeata, line 6056: If you leave in `failwith`, maybe with a better error message?

src/ir_def/construct.ml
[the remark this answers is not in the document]
nomeata: Yes, my comment was misguided, I thought you were introducing a `BinE`; `BinPrim` is fine.

src/ir_def/construct.ml
```diff
@@ -600,22 +607,48 @@ let loopWhileE exp1 exp2 =
         )
     )
 
+let countingE binding arr expSize expIndexing expWorker =
+  (* when e1 = arr.vals()
+
+     for p in e1 e2
```
nomeata, line 613:
```suggestion
     for (p in e1) e2
```

src/ir_def/construct.ml
```diff
@@ -600,22 +607,48 @@ let loopWhileE exp1 exp2 =
...
+     ~~>
+     let arr = ... ;
+     var indx = 0 ;
+     let size = arr.size() ;
+     label l loop {
+       if indx < size
+       then { let p = arr[indx]; e2; indx += 1 }
+       else { break l }
+     } *)
+  let indx = fresh_var "indx" T.(Mut nat) in
```
nomeata, line 623: Maybe assert that `arr` has the expected type, as safeguard and documentation?

src/ir_def/construct.ml
```diff
@@ -600,22 +607,48 @@ let loopWhileE exp1 exp2 =
...
+  let size = fresh_var "size" T.nat in
+  blockE [varD indx (natE Numerics.Int.zero)
+        ; letD size expSize]
+    (whileE (primE (RelPrim (T.nat, Operator.LtOp))
+               [varE indx; varE size])
+       (blockE [letP binding (expIndexing indx)
+              ; letP wildP expWorker]
```
nomeata, line 630:
```suggestion
              ; expD expWorker]
```

src/ir_def/construct.ml
```diff
@@ -600,22 +607,48 @@ let loopWhileE exp1 exp2 =
...
+          (assignE indx
+             (primE (BinPrim (T.nat, Operator.AddOp))
+                [ varE indx
+                ; natE (Numerics.Int.of_int 1)]))))
```
nomeata, line 634:
```suggestion
                ; natE (Numerics.Nat.of_int 1)]))))
```

src/ir_def/construct.mli
```diff
@@ -60,6 +60,7 @@ val projE : exp -> int -> exp
 val optE : exp -> exp
 val tagE : id -> exp -> exp
 val blockE : dec list -> exp -> exp
+val natE : Mo_values.Numerics.Int.t -> exp
```
nomeata, line 63: Why not
```suggestion
val natE : Mo_values.Numerics.Nat.t -> exp
```
(resolves to the same type anways)

src/lowering/desugar.ml
```diff
@@ -198,6 +198,9 @@ and exp' at note = function
   | S.WhileE (e1, e2) -> (whileE (exp e1) (exp e2)).it
   | S.LoopE (e1, None) -> I.LoopE (exp e1)
   | S.LoopE (e1, Some e2) -> (loopWhileE (exp e1) (exp e2)).it
+  | S.ForE (p, ({it=S.CallE (({it=S.DotE (arr, proj); _} as c0), c1, c2); _} as e1), e2)
+      when T.is_array arr.note.S.note_typ && proj.it = "vals"
```
nomeata, line 202: We should handle `.keys()` as well, for consistency

src/lowering/desugar.ml
```diff
@@ -198,6 +198,9 @@ and exp' at note = function
...
+    -> (rewrite_for_to_while (pat p) arr proj c0 c1 c2 e1 (exp e2)).it
```
nomeata, line 203: I find it fishy that you are passing `c0` here when you are already passing `arr` and `proj`, as similar for other components. Is that necessary?

Maybe simply match that  `c2` is `TupE []` here. And `c1 : inst` can probably be dropped? Isn’t that always “empty” anyways?

I see you are using them in `rewrite_for_to_while` in the call to `size`. That seems a bit hackish somehow.

src/lowering/desugar.ml
```diff
@@ -239,6 +242,53 @@ and lexp' = function
...
+       else { break l }
+     } *)
+  let arr_typ = arr.note.note_typ in
+  let triv t = { note_typ = t; note_eff = T.Triv } in
+  let body arrb =
+    let size_exp =
+      exp { e1 with note = triv T.nat;
+                    it = CallE ({ c0 with note = triv T.(Func (Local, Returns, [], [], [nat]));
+                                          it = DotE (arrb, { proj with it = "size" }) },
+                                c1, c2) } in
```
nomeata, line 263: Assuming `c2` definitely has no effect, then this could be
```suggestion
    let size_exp = array_dotE arr_typ "size" arb in
```

src/lowering/desugar.ml
```diff
@@ -239,6 +242,53 @@ and lexp' = function
...
+    let indx = fresh_var "indx" T.(Mut nat) in
+    let indexing_exp = exp
+           { note = triv T.(as_immut (as_array arr_typ));
+             at = arr.at;
+             it = IdxE (arrb, { arr with note = triv T.nat;
+                                         it = VarE { it = id_of_var indx;
+                                                     note = ();
+                                                     at = arr.at } }) } in
```
nomeata, line 271: Can’t you produce generate `I.PrimE (I.IdxPrim, […, …])` directly here?

src/lowering/desugar.ml
```diff
@@ -239,6 +242,53 @@ and lexp' = function
...
+                  [ varE indx
+                  ; natE (Numerics.Int.of_int 1)])))) in
+  let arr_ir = exp arr in
+  match arr_ir.it with
+  | I.VarE _ -> body arr
+  | _ ->
+    let arrv = fresh_var "arr" arr_typ in
+    letE arrv arr_ir
+      (body { arr with it = VarE { it = id_of_var arrv; note = (); at = arr.at };
+                       note = triv arr.note.note_typ })
```
nomeata, line 290: This (let-bind if not trivial) is too much unrelated logic; I suggest to put that in `construct.ml` as
```
letNonTrivialE : exp -> (var -> exp) -> exp
```
or something like that.

And then maybe simply write
```
rewrite_for_to_while p arr proj c0 c1 c2 e1 e2 =
  letNonTrivialE (exp arr) (fn arr_ir ->
    …
  )
```
i.e. wrap it all in `letNonTrivialE`. This way code order matches evaluation order, is closer to the pseudocode in the comment, and the `arr_ir` is in scope, i.e. `body` doesn’t have to be a function.

test/run/optimise-for-array.mo
```diff
@@ -0,0 +1,53 @@
...
+// CHECK-NEXT: if
+// CHECK-NEXT: loop
+// CHECK-NEXT: br 0
+// CHECK-NEXT: end
+// CHECK-NEXT: unreachable
+// CHECK-NEXT: else
+// bottom iteration expression is treated fairly
+var c = 42;
+if (c == c + 1) {
+    for (check4 in (loop {}).vals()) { Prim.debugPrint check4 }
```
nomeata, line 52: I don’t think this will trigger your transformation becaues `T.is_type T.None = false`. Probably something like
```suggestion
    for (check4 in ((loop {} : [Int])).vals()) { Prim.debugPrint check4 }
```
would work better

src/codegen/compile.ml
ggreif: Nope, this is a change I did for debugging only. I'll revert!

ggreif:
```suggestion
    | _ -> assert false
```

src/ir_def/construct.ml
ggreif: It gets implicitly asserted in `rewrite_for_to_while`. This function is gone now.

src/lowering/desugar.ml
ggreif: Yes, the name gets captured and the type mismatch detected by the IT lint:
```
test/run/optimise-for-array.mo:75.16-75.22: IR type error [M0000], subtype violation:
  Text
  [var Text]
```
I'll see if I can guard against this case.

nomeata: Don’t worry about explicitly guarding, just introduce a fresh name. We are doing that all over the place, e.g. in the await translation. If this becomes a problem we can introduce a separate pass that cleans up the code again, or be smarter in the backend. (But not before introducing a proper non-recursive `let` to the IR…)

ggreif: Heh, note to self: never argue with a red-team guy...

I agree that variable path compression should be eventually be done in the backend, but we don't have walkable def-use chains yet.

Consider me convinced.

src/lowering/desugar.ml
```diff
@@ -239,6 +242,53 @@ and lexp' = function
...
+                  [ varE indx
+                  ; natE (Numerics.Nat.of_int 1)])))) in
+  let arr_ir = exp arr in
+  match arr_ir.it with
+  | I.VarE v when p.it <> I.VarP v -> body arr
+  | _ ->
+    let arrv = fresh_var "arr" arr_typ in
+    letE arrv arr_ir
+      (body { arr with it = VarE { it = id_of_var arrv; note = (); at = arr.at };
+                       note = triv arr.note.note_typ })
```
crusso, line 290:
```suggestion
      (body (varE arrv))
```
or maybe better to abstract `body` on a `var`, not `exp`, so you know the effect is trivial (this is call-by-value) and call:

```
body arrv
````

**Review by crusso — COMMENTED** [review]
Ping me for the final review. Looking much better.

File: src/lowering/desugar.ml [hunk 20]
```diff
@@ -239,6 +242,47 @@ and lexp' = function
...
+  let arr_typ = arr.note.note_typ in
+  let triv t = { note_typ = t; note_eff = T.Triv } in
+  let body arrb =
+    let size_exp =
+      exp { e1 with note = triv T.nat;
+                    it = CallE ({ c0 with note = triv T.(Func (Local, Returns, [], [], [nat]));
+                                          it = DotE (arrb, { proj with it = "size" }) },
+                                c1, c2) } in
+    let indx = fresh_var "indx" T.(Mut nat) in
+    let indexing_exp = (primE Ir_def.Ir.IdxPrim [exp arrb; varE indx]) in
```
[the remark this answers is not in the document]
ggreif: It should. THX!

src/lowering/desugar.ml
ggreif: This should wrap it up: d8242d132. No more variable shortcuts.

src/lowering/desugar.ml
```diff
@@ -239,6 +242,47 @@ and lexp' = function
...
+  let size_exp =
+    let arr_bind =
+      { arr with it = VarE { it = id_of_var arrv; note = (); at = arr.at };
+                 note = { note_typ = arr.note.note_typ; note_eff = T.Triv } } in
+    exp { e1 with note = { note_typ = T.nat; note_eff = c2.note.note_eff };
+                  it = CallE ({ note = { note_typ = T.(Func (Local, Returns, [], [], [nat]));
+                                         note_eff = c2.note.note_eff };
+                                it = DotE (arr_bind, { proj with it = "size" });
+                                at = c0 },
+                              c1, c2) } in
```
crusso, line 267: Ok, I was wrong, and missed the fact that you are invoking `exp` on source.
But I still think that's a little indirect and hard to follow.
I would delete `arr_bind` and just insert  a `expD (exp c2)` and the simpler size_exp suggested by @nomeata
(see below)

src/lowering/desugar.ml
```diff
@@ -239,6 +242,47 @@ and lexp' = function
...
+  let indx = fresh_var "indx" T.(Mut nat) in
+  let indexing_exp = match proj.it with
+    | "vals" -> primE I.IdxPrim [varE arrv; varE indx]
+    | "keys" -> varE indx
+    | _ -> assert false in
+  let size = fresh_var "size" T.nat in
+  letE arrv (exp arr)
+    (blockE [ letD size size_exp
+            ; varD indx (natE Numerics.Nat.zero)]
```
crusso, line 276:
```suggestion
  letE arrv (exp arr)
    (blockE [
            let arrv (exp arr);
            expD (exp c2); 
            letD size (array_dotE arr_typ "size" (varE arrv));
            varD indx (natE Numerics.Nat.zero)]
```
Isn't that simpler? Or am I missing something.

crusso, line 276:
```suggestion
    (blockE [
            letD arrv (exp arr);
            expD (exp c2); 
            letD size (array_dotE arr_typ "size" (varE arrv));
            varD indx (natE Numerics.Nat.zero)]
```
Isn't that simpler? Or am I missing something.

src/lowering/desugar.ml
```diff
@@ -239,6 +242,47 @@ and lexp' = function
...
+       (whileE (primE (I.RelPrim (T.nat, LtOp))
+                  [varE indx; varE size])
+          (blockE [ letP p indexing_exp
+                  ; expD e2]
```
crusso, line 280:
```suggestion
                  ; expD (exp e2)]
```
(and remove exp from argument at call site)

crusso, line 280:
```suggestion
                  ; expD (exp e2)]
```
(and remove exp from argument at call site)

src/lowering/desugar.ml
```diff
@@ -239,6 +242,47 @@ and lexp' = function
...
+  (* for (p in e1) e2 when e1 = ⟨...array-typed...⟩.vals()
+     ~~>
+     let arr = ⟨...array-typed...⟩ ;
+     let size = arr.size() ;
+     var indx = 0 ;
+     label l loop {
+       if indx < size
+       then { let p = arr[indx]; e2; indx += 1 }
+       else { break l }
+     } *)
```
crusso, line 255:
```suggestion
  (* for (p in e1) e2 when e1 = e3.vals e4 (when e3 : [ var? ... ])
     ~~>
     let arr = e3;
     e4;
     let size = arr.size() ;
     var indx = 0 ;
     label l loop {
       if indx < size
       then { let p = arr[indx]; e2; indx += 1 }
       else { break l }
     } *)
```

PR #2827, opened 2021-10-02
Stable signatures: frontend

crusso:
WIP: Baby steps towards parsing and emitting stable sigs.

Notes so far: 

* the Memory types we compute have optioned, immutable fields (enough info to drive deserialization)
* name generation for type constructors may get confused by stamping
* Primitive type decs need to be discovered and elided.
* reference to type constructors appears to be unfolded once. Not wrong, but odd.

PR #2838, opened 2021-10-14
Propagate metadata to wasm custom sections

crusso:
Implements #2837 (Propagating candid and stable vars to custom section) (by building on #2827).

I think this can be reviewed in parallel since its largely independent of the work in #2827.


chenyan-dfinity (commented):
LGTM. We also want to update the `--print-deps` flag to identify if an import is an actor class: https://github.com/dfinity/motoko/pull/2837/files#diff-930fd92d888e5ab8dd9ac1b49d0f091a012c26fcfe9dc294ace3d2f9f7231d69R142

src/wasm-exts/customModuleDecode.ml
```diff
@@ -684,6 +684,41 @@ let custom_section (name_pred : int list -> bool) (f : int -> stream -> 'a) (def
...
+    let opt = icp_name n name in
+    begin
+    match opt with
+    | Some b ->
+      (* this is the right custom section *)
+      let x = f sec_end s in
+      require (pos s = sec_end) s sec_start "custom section size mismatch";
+      Some (b, x)
+    | None -> begin (* wrong custom section, rewind *)
+      s.pos := p;
```
rossberg, line 715: Hm, this rewind is a bit ugly (you could start a new stream, skipping over sections isn't that costly). Could you at least abstract that into a function at the top of the module?

src/wasm-exts/customModuleDecode.ml
```diff
@@ -760,27 +795,48 @@ let is_name n = (n = Utf8.decode "name")
...
 
-let motoko_section_subsection (ms : motoko_section) s =
+let motoko_section_subsection (ms : motoko_sections) s =
   match u8 s with
-  | 0 -> (* module name *)
-    let labels = sized (fun _ -> vec string) s in
-    { labels = ms.labels @ labels }
+  | 0 ->
+     let labels = sized (fun _ -> vec string) s in
+     { ms with labels = ms.labels @ labels }
```
rossberg, line 804:
```suggestion
    let labels = sized (fun _ -> vec string) s in
    { ms with labels = ms.labels @ labels }
```

src/wasm-exts/customModuleDecode.ml
```diff
@@ -760,27 +795,48 @@ let is_name n = (n = Utf8.decode "name")
...
+  let service = icp_custom_section "candid:service" string_content None s in
+  let args = icp_custom_section "candid:args" string_content None s in
+  { service; args }
 
 (* Other custom sections *)
 
-let is_unknown n = not (is_dylink n || is_name n)
+let candid_service_name = icp_name "candid:service"
+let candid_args_name = icp_name "candid:args"
+let motoko_stable_types_name = icp_name "candid:stable-types"
```
rossberg, line 829:
```suggestion
let motoko_stable_types_name = icp_name "motoko:stable-types"
```

rossberg (approved):

src/wasm-exts/customModuleDecode.ml
crusso: Done; and added a todo to redo the work for all custom sections. I was merely adapting the custom_section code already present in this file, but I think we should probably just read all custom sections in the manner you suggest, rather than being clever about it (as Joachim was trying to be for the dylink and names sections).

src/codegen/compile.ml
```diff
@@ -3,7 +3,7 @@ This module is the backend of the Motoko compiler. It takes a program in
 the intermediate representation (ir.ml), and produces a WebAssembly module,
 with Internet Computer extensions (customModule.ml). An important helper module is
 instrList.ml, which provides a more convenient way of assembling WebAssembly
-instruction lists, as it takes care of (1) source locations and (2) labels.
+instruction lists, as it takes care of (1) source locations and (2) labels
```
crusso, line 6:
```suggestion
instruction lists, as it takes care of (1) source locations and (2) labels.
```

src/codegen/compile.ml
```diff
@@ -255,10 +255,16 @@ module E = struct
     static_memory_frozen : bool ref;
       (* Sanity check: Nothing should bump end_of_static_memory once it has been read *)
     static_roots : int32 list ref;
-      (* GC roots in static memory. (Everything that may be mutable.) *)
+    (* GC roots in static memory. (Everything that may be mutable.) *)
```
crusso, line 258:
```suggestion
      (* GC roots in static memory. (Everything that may be mutable.) *)
```

PR #2841, opened 2021-10-16
add missing closed bracket

bytesun:
it cause the link can't be open (404 error)

PR #2843, opened 2021-10-17
Documentation corrections

ggreif:
Several fixes, originally from #2841.


ggreif:
This also fails the CI.

PR #2845, opened 2021-10-19
Add the motoko-base library to releases

pikajude:
There doesn't seem to be any way to associate a specific motoko release with a version of base that's guaranteed to work with it, so this at least guarantees that we publish the same version that the testsuite passes with.


nomeata:
Looks reasonable. Is this part of the plan to not use nix anymore to get Motoko artifacts into the dfx build?

PR #2846, opened 2021-10-20
add file remove and rename in moc.js

chenyan-dfinity:


crusso (approved):
Is this to support the playground?

PR #2855, opened 2021-10-22
Eliminate bignum calls from `for`-iteration on arrays

ggreif:
This is an afterburner on top of #2831, which implemented the `while` optimisation for `for` loops. Even with that the resulting code was littered with `call`s to {`$B_add`, `$@(im)mut_array_size`, `$B_lt`, `$Array.idx_bigint`}, which hide significant cycle counts and contain redundant conditionals. (Also, with future versions of the cycle counting algorithm the `call` instructions will become heavier.) Now we switch away from _expensive_ (that is _checked_) bignum arithmetic to very cheap operations (`i32.shl` and `i32.add`) since the index space of arrays is so limited that it cannot cause overflows into proper heap-allocated bignums. The underlying type of indexing operations remains `Nat`.

I have implemented a handful of primitives that we now use when lowering `for`-to-`while`. In the case of `vals` loop we increment by offset distances of 4 bytes, such that we can simply add the offset to the array pointer and dereference. This saves a bunch of multiplications. We get cycle savings in the ballpark of 7%, which is not too shabby.

## The new primitives

We introduce four new primitive into the IR, which allow to express the use cases of iterating `vals` and `keys` sequences of arrays:

| Name | Purpose | Used in iteration over | Type |
|------|----------|-----------------------|------|
| `GetPastArrayOffset` | obtain the iteration limit | `vals`, `keys` | `[A] -> Nat` |
| `ValidArrayOffset` | determine whether the iteration can continue | `vals`, `keys` | `Nat -> Nat -> Bool` |
| `NextArrayOffset` | advance iteration state | `vals`, `keys` | `Nat -> Nat` |
| `DerefArrayOffset ` | obtain element from array corresponding to the iteration state  | `vals` | `[A] -> Nat -> A` |

`GetPastArrayOffset` and `NextArrayOffset` need a _spacing hint_ which is needed by the backend only to know by which number to increment the iteration state.

The elimination of expensive bignum arithmetic (Wasm calls) grounds on the fact that arrays have less than 2^30 elements. This makes sure that the iteration state is always
- a compact bignum for `keys` iteration
- a `u32` which always has the two lowest bits cleared (and thus appears as a valid compact signed bignum, a.k.a. `Int`) for `vals` iteration.

The iteration state is always a GC compatible object and thus won't cause problems when being heap allocated by the `await`-transformation.

### How these primitives work with iteration state

In the realm of IR all these primitives are strictly adhering to their types.

There is only one case where the iteration state is not a bona fide `Nat`. Namely when _running Wasm and iterating over an array's_ `vals`. In this case the iteration limit (as well as the iteration variable) can technically become a compact "`Nat`" with the _sign bit_ set (so basically a negative compactly-represented `Int`). Here is the scenario how this can happen:

- assume an array with size `1000_000_000` (i.e. `< 2^30`, since no array can occupy more than 4 GB with each element occupying 4 bytes)
- `GetPastArrayOffset` will give `1000000000 << 2` == `0xee6b2800` as `Vanilla`, pointer bit cleared
- `0xee6b2800` is an "illegal" `Nat`, but a legal `Int` with the value -147483648

Since the value is only consumed by the primitives `ValidArrayOffset`, `NextArrayOffset` and `DerefArrayOffset`, which don't care for this quirk, everything is fine. Even if the loop takes part in the `await` transformation (or any other constraint that moves Wasm locals out into heap-allocated data), the `Vanilla` value has no pointer bit set, and thus behaves correctly w.r.t. GC (since GC won't take Motoko's type information into account due to type erasure).

## TODOs
- [x] IR typing rule for `GetPastArrayOffset` still to be filled in
- [x] ~~don't use `2l` as a magic for `1 : Nat`~~ added comment
- [x] verify that the biggest arrays still give compact bignums by `i32.shl 1` (not relevant for `vals`)  — DONE, see subsection above
- [ ] can we use `OtherPrim` (from `lowering`)?
- [x] improve `OtherPrim "array_len"` (suggested for removal, see #2882)
- [ ] find better names for the primitives (without resorting to `Offset`?)

## Further Opportunities
- shift `n_fields` when building `Array`?
- single-`get` variables can be eliminated (if effects not block), by moving the computation formerly before the `set` to replace the `get`
- `for`-loops shouldn't materialise `i32.const 0` on the stack as the result, as units can be readily materialised anywhere (if ever needed)
- [x] lower `DotE(arr, "size")` to `OtherPrim ("array_len", arr)` — see #2882


nomeata:
> there is only one get, could be eliminated along with the local!

Is this worth worrying about? I was under the impression that the production VMs will do that kind of optimization better than us. (That leaves code size, of course, but there we have probably much lower hanging fruit). Nothing wrong with fixing it if it is easy, but I would hesitate to pay with much code complexity for that.

ggreif:
> but I would hesitate to pay with much code complexity for that

I am just collecting the infelicities that I encounter. If there are Wasm->Wasm optimisers that perform this task, then I am absolutely happy.

src/codegen/compile.ml
```diff
@@ -7462,12 +7467,38 @@ and compile_exp (env : E.t) ae exp =
...
       compile_exp_vanilla env ae e2 ^^ (* idx *)
       Arr.idx_bigint env ^^
       load_ptr
+    | NextArrayOffset purpose, [e] ->
+      let spacing = if purpose = "vals" then Arr.element_size else 2l in
+      SR.Vanilla,
+      compile_exp_vanilla env ae e ^^ (* previous byte offset to array *)
+      compile_add_const spacing
+    | ValidArrayOffset, [e1; e2] ->
+      let sr, code = compile_relop env Type.(Prim Nat16) Operator.LtOp in
```
ggreif, line 7482: this is not very transparent

src/ir_def/ir.ml
```diff
@@ -110,6 +110,10 @@ and prim =
   | ActorDotPrim of Type.lab          (* actor field access *)
   | ArrayPrim of mut * Type.typ       (* array constructor *)
   | IdxPrim                           (* array indexing *)
+  | NextArrayOffset of string         (* advance array offset *)
```
ggreif, line 113: `type Spacing = One | ElementSize`

test/run/optimise-for-array.mo
```diff
@@ -88,7 +109,7 @@ num8 := 25;
 // indexing variable itself
 for (check8 in ["hello", "keyed", "world"].keys()) { ignore (check8 + num8) };
 
-
+// polymorphic arrays should still work
 func f9<A>(array : [A]) {
 for (check9 in array.keys()) { }
```
ggreif, line 114:
```suggestion
  for (check9 in array.keys()) { }
```

src/codegen/compile.ml
ggreif: Resolved by just evaluating as `Vanilla` and doing an unsigned comparison.

src/codegen/compile.ml
```diff
@@ -7338,6 +7338,11 @@ and compile_exp (env : E.t) ae exp =
 
     begin match p, es with
     (* Calls *)
+    | CallPrim _, [{it=PrimE (CallPrim _, [{it=VarE ("@immut_array_size" | "@mut_array_size");_}; e1]);_}; {it=PrimE (TupPrim, []);_}] ->
```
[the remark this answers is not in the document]
ggreif: I moved the transformation into the desugarer with another PR, #2882. This PR will just take care of the new primitives.

src/codegen/compile.ml
```diff
@@ -7462,12 +7467,40 @@ and compile_exp (env : E.t) ae exp =
...
+      SR.bool,
+      compile_exp_vanilla env ae e1 ^^
+      compile_exp_vanilla env ae e2 ^^
+      G.i (Compare (Wasm.Values.I32 I32Op.LtU))
+    | DerefArrayOffset, [e1; e2] ->
+      SR.Vanilla,
+      compile_exp_vanilla env ae e1 ^^ (* skewed pointer to array *)
+      compile_exp_vanilla env ae e2 ^^ (* byte offset *)
+      G.i (Binary (Wasm.Values.I32 I32Op.Add)) ^^
+      Arr.(load_field 0l)              (* loads the element at the byte offset *)
```
crusso, line 7494:
```suggestion
    | DerefArrayOffset, [e1; e2] ->
      SR.Vanilla,
      compile_exp_vanilla env ae e1 ^^ (* skewed pointer to array *)
      compile_exp_vanilla env ae e2 ^^ (* byte offset *)
      G.i (Binary (Wasm.Values.I32 I32Op.Add)) ^^
      Arr.load_field 0l            (* loads the element at the byte offset *)
```
Is there something clever going on here, because it seems odd. Is it safe to just add to the skewed pointer without unskewing first? And why return the 0th field of the resulting array on the stack?

src/ir_def/check_ir.ml
```diff
@@ -473,6 +473,19 @@ let rec check_exp env (exp:Ir.exp) : unit =
...
+        (try T.as_array_sub t1 with
+         | Invalid_argument _ ->
+           error env exp1.at "expected array type, but expression produces type\n  %s"
+             (T.string_of_typ_expand t1));
+      T.nat <: t
+    | NextArrayOffset _, [exp1] ->
+      typ exp1 <: T.nat
+    | ValidArrayOffset, [exp1; exp2] ->
+      typ exp1 <: T.nat;
+      typ exp2 <: T.nat
```
crusso, line 488: So, according to your comment, sometime the Nat is actually an Int? So is the ir-checker actually fibbing? If so, that makes me little nervous....

src/ir_def/construct.ml
```diff
@@ -89,7 +89,11 @@ let primE prim es =
     | ICCallerPrim -> T.caller
     | ICStableRead t -> t
     | ICStableWrite _ -> T.unit
-    | IdxPrim -> T.(as_immut (as_array (List.hd es).note.Note.typ))
+    | IdxPrim
+    | DerefArrayOffset -> T.(as_immut (as_array (List.hd es).note.Note.typ))
+    | NextArrayOffset _ -> T.nat
+    | ValidArrayOffset -> T.bool
+    | GetPastArrayOffset _ -> T.nat
```
crusso, line 96: The return type of GetPastArrayOffset _ is a lie?

ggreif: It is not a lie while in the scope of IR. _After_ leaving IR (entering Wasm, most types being erased), not a problem any more. I have described this in the PR commentary.

src/ir_def/check_ir.ml
ggreif: Here it is always `Nat`. It changes meaning (slightly) in the Wasm realm.

src/codegen/compile.ml
ggreif: `Arr.load_field` implicitly unskews and hops over the administrative fields, then loads the element. Unskewing and hopping is done by an `add`, which commutes with the `add` of the gate offset. So yes, it is mildly clever :-)

src/ir_def/construct.ml
ggreif: There is not a single IR typing lie. The new primitives all play by the typing rules.

src/codegen/compile.ml
crusso: maybe just add a comment to explain this?

src/codegen/compile.ml
```diff
@@ -7462,12 +7467,40 @@ and compile_exp (env : E.t) ae exp =
...
+    | GetPastArrayOffset spacing, [e] ->
+      let shift =
+        match spacing with
+        | ElementSize -> compile_shl_const 2l (* effectively a multiplication by word_size *)
+        | One -> BigNum.from_word30 env in    (* make it to compact bignum *)
```
crusso, line 7499:
```suggestion
        | One -> BigNum.from_word30 env in    (* make it a compact bignum *)
```

src/ir_def/check_ir.ml
```diff
@@ -473,6 +473,19 @@ let rec check_exp env (exp:Ir.exp) : unit =
...
+    | GetPastArrayOffset _, [exp1] ->
+      let t1 = T.promote (typ exp1) in
+      ignore
+        (try T.as_array_sub t1 with
+         | Invalid_argument _ ->
+           error env exp1.at "expected array type, but expression produces type\n  %s"
+             (T.string_of_typ_expand t1));
+      T.nat <: t
+    | NextArrayOffset _, [exp1] ->
+      typ exp1 <: T.nat
```
crusso, line 485:
```suggestion
      typ exp1 <: T.nat;
      T.nat <: t
```
Shouldn't you check the return type T.nat?

src/ir_def/check_ir.ml
crusso:
```suggestion
    | GetPastArrayOffset _, [exp1] ->
      let t1 = T.promote (typ exp1) in
      ignore
        (try T.as_array_sub t1 with
         | Invalid_argument _ ->
           error env exp1.at "expected array type, but expression produces type\n  %s"
             (T.string_of_typ_expand t1));
      T.nat <: t
    | NextArrayOffset _, [exp1] ->
      typ exp1 <: T.nat
    | ValidArrayOffset, [exp1; exp2] ->
      typ exp1 <: T.nat;
      typ exp2 <: T.nat;
      T.bool <: t
```
Doesn't this need to return a Boolean?

src/ir_def/construct.ml
```diff
@@ -89,7 +89,11 @@ let primE prim es =
     | ICCallerPrim -> T.caller
     | ICStableRead t -> t
     | ICStableWrite _ -> T.unit
-    | IdxPrim -> T.(as_immut (as_array (List.hd es).note.Note.typ))
+    | IdxPrim
+    | DerefArrayOffset -> T.(as_immut (as_array (List.hd es).note.Note.typ))
```
crusso, line 93: I wonder if this code will actually fail (and used to fail for IdxPrim) if the array argument has type None...

crusso (changes requested):
LGTM apart from suggestions. Nice boost actually!

src/ir_def/construct.ml
crusso: (or is a generic type bounded by an array type)

src/ir_def/check_ir.ml
ggreif: Yes, indeed! Thanks for catching!

ggreif: Line 485 has the same deficiency. Fixing too!

src/ir_def/construct.ml
ggreif: I remember writing a transformation a few weeks ago that called `is_array` on a value of type `None` and it returned false. I'll look it up. I assume that `as_array` would throw in that context.

src/ir_def/check_ir.ml
ggreif: Hah, you spotted it: f77424577

src/ir_def/construct.ml
ggreif: Do you mean, that the right semantics for `as_array` should return `None` for a `None` input?

ggreif: `as_array_sub` does precisely that: 78b8517f5

crusso (approved):
Hurrah!

PR #2867, opened 2021-10-29
Stable signatures: implement moc arguments --compatible <pre> <post> and --stable-types

crusso:
- `--stable-types emits a (stable var)` `.sig` file containing the contents of the compilations stable-type custom section.

- `--compatible <pre> <post>` checks whether the signatures in files `<pre>` and `<post>` are upgrade compatible.
   This reports any static errors and exits with 0 (true) or 1 (false) so you can use it as a test in scripts.

Currently assumes `pre` and `post` are `.sig` files containing just stable variables, but we could imagine generalizing that to allow:

- wasm files, extracting the stable-vars section and/or candid sections and checking compatibility of those.
- unextended basenames, using the derived `.did` and `.sig` files and checking those.

We currently can't check compatibility of did files in moc because we don't have a subtyping check implement in the OCaml candid implementation.

Ideally, it would be nice to add a `--compatibleWith <pre>` option to check the current compilation unit remains compatible with `<pre>` (instead of just checking two standalone sigs).

Are we happy with extension `.sig`? I also thought of `.moi`, which is kinda cute if you speak french.

TODO: 

- [x] regression tests
- [x] better error messages than simple yes or no.


chenyan-dfinity (approved):
LGTM

chenyan-dfinity:
This is good enough for dfx. Checking did file can be done in dfx via didc, and we have better error messages there.

What's the syntax for `.sig` file? Do they follow the subtyping in Motoko or Candid?

src/exes/moc.ml
```diff
@@ -27,13 +27,21 @@ let link = ref true
...
 let explain_code = ref ""
+let stable_types = ref false
 
 let argspec = [
   "-c", Arg.Unit (set_mode Compile), " compile programs to WebAssembly";
   "-g", Arg.Set Flags.debug_info, " generate source-level debug information";
   "-r", Arg.Unit (set_mode Run), " interpret programs";
   "-i", Arg.Unit (set_mode Interact), " run interactive REPL (implies -r)";
   "--check", Arg.Unit (set_mode Check), " type-check only";
+  "--compatible",
```
rossberg, line 38: Perhaps rather call this `stable-compatible`? It does not check compatibility of the actor interface itself, so the name would be a bit misleading.

src/exes/moc.ml
```diff
@@ -106,6 +114,10 @@ let argspec = [
     (fun () -> Flags.sanity := true),
   " enable sanity checking in the RTS and generated code";
 
+  "--stable-types",
+  Arg.Unit (fun () -> stable_types := true),
+  " emit signature of stable types to .sig file";
```
rossberg, line 119: Can we use a more specific file extension? How about `.stable`, or `.most`?

src/mo_frontend/stability.ml
```diff
@@ -17,11 +17,22 @@ let parse_with mode lexer parser name : Syntax.sig_ Diag.result =
...
 let parse_sig s name  =
   let open Diag.Syntax in
-  let mode = {Lexer.privileged = true} in (* TODO: make false *)
+  let mode = {Lexer.privileged = false} in
   let lexer = Lexing.from_string s in
   let parse = Parser.Incremental.parse_sig in
   let* sig_ = parse_with mode lexer parse name in
   Diag.return sig_
 
+let parse_sig_from_file filename : Syntax.sig_ Diag.result =
```
rossberg, line 26:
```suggestion
let parse_stable_sig_from_file filename : Syntax.sig_ Diag.result =
```

src/pipeline/pipeline.ml
```diff
@@ -227,6 +226,15 @@ let prelude, initial_stat_env0 =
 let internals, initial_stat_env =
   check_builtin "internals" Prelude.internals initial_stat_env0
 
+(* Compatibility *)
+
+let compatible pre post : bool Diag.result =
```
rossberg, line 231:
```suggestion
(* Stable compatibility *)

let stable_compatible pre post : bool Diag.result =
```

rossberg (approved):
LGTM, but as in the previous PR, I'd suggest a slightly less generic naming scheme.

src/exes/moc.ml
crusso: I liked .moi ;->

src/exes/moc.ml
crusso: Well, I was hoping we would extend --compatible to support other inputs as well, based on file extension.
*.wasm  compare both candid:service and motoko:stable-types, extracted from
*.did compare service (ignoring any service arguments)
*.sig compare sig
* (no extension) compare both *.did and *.sig
But I guess we could add the general option later.

src/pipeline/pipeline.mli
```diff
@@ -15,6 +15,8 @@ val print_deps: string -> unit
 val check_files  : string list -> unit Diag.result
 val check_files' : parse_fn -> string list -> unit Diag.result
 
+val compatible : string -> string -> bool Diag.result
```
crusso, line 18:
```suggestion
val stable_compatible : string -> string -> bool Diag.result
```

src/exes/moc.ml
rossberg: Oh yes, I'd like that. But as long as it doesn't do that perhaps better to not call it that.

src/exes/moc.ml
crusso: Changed  to .most in full PR #2887

PR #2874, opened 2021-11-01
Fix "no drun" message in run.sh when ic-ref-run is missing

osa1:


crusso (approved):
Thanks!

nomeata (approved):
Thanks for fixing my 🤦🏻

PR #2876, opened 2021-11-01
Candid decoder: Trap if a `principal` value is too large

### LINKED ISSUE #2875: Candid decoder: Fail if principals are large
The public spec says that  [principals are at most 29  bytes](https://sdk.dfinity.org/docs/interface-spec/index.html#principal) in  length, but the [Motoko Candid decoder](https://github.com/dfinity/motoko/blob/6feb52696bd7c446a7fe5231519045dd6d248bdc/src/codegen/compile.ml#L5031) does not check that.

This may lead to DOS vulnerabilities where a developer does not expect that attackers can send very large values (up to the message size limit) here.

Also applies to the `principal` in a service or function reference.

Suggested fix: add a `read_principal` helper function (like `read_blob`) that traps if the size is more than 29 bytes.

nomeata:
this fixes #2875


src/codegen/compile.ml
```diff
@@ -4775,6 +4775,20 @@ module Serialization = struct
         get_x
       in
 
+      let read_principal () =
+        let (set_len, get_len) = new_local env "len" in
+        let (set_x, get_x) = new_local env "x" in
+        ReadBuf.read_leb128 env get_data_buf ^^ set_len ^^
+
+        get_len ^^ compile_unboxed_const 23l ^^ G.i (Compare (Wasm.Values.I32 I32Op.LeU)) ^^
```
ggreif, line 4783: I feel some disconnect here. In #2875 you say that principals are at most 29 bytes, but you set the limit to 23 in this line.

nomeata: Thanks for catching! The number 23 entered my mind somehow between writing the issue and writing the code

ggreif (approved):
Great, thanks!

PR #2882, opened 2021-11-03
Improve lowering of calls to `DotE (_, "size")`

ggreif:
When lowering `DotE (_, "size")`, previously `@(im)mut_array_size` helper would be invoked on an array. But `@(im)mut_array_size` would return a closure of type `() -> Nat`, and in almost all cases this closure will be invoked with a unit to obtain the `Nat`. This results in an indirect call in the Wasm, which is overkill.

Now we lower the saturated `DotE` in a more efficient way by using the `PastEndArrayOffset` primitive (introduced in #2855) to basically reduce the whole shebang to a single memory load.

## TODOs
- What happens when `DotE (arr, "size")` is not saturated? Test this!


luc-blaeser:

PR #2884, opened 2021-11-04
Stable signatures: test suite for --compatible .sig comparison

crusso:

PR #2887, opened 2021-11-05
[feat] stable signatures: frontend, metadata, command-line args and tests

crusso:
Implements all support for stable signatures (plus candid:service and candid:arg custom sections)

Rolls up PRs

- #2884 (final integration plus fixes)
- #2867
- #2838 
- #2827 

You can either review this massive PR in one go or the others in reverse order (note the last one #2884 includes some fixes to the compatibility relation, which was buggy in #2827 and final refactorings).

This work is all based on the original design here:

https://github.com/dfinity/motoko/blob/master/design/Stable.md#stable-signatures

Also implements #2837


crusso:
@rossberg PTAL

rossberg:
Great stuff!

crusso:
@rossberg  PTAL, perhaps at this PR

chenyan-dfinity:
How do we upgrade actor class at the moment? If we change the code, `moc` gives us a new Wasm for the actor class, but there is no way to upgrade the already deployed actor class instances, right? The developer will have to somehow write the upgrade logic in the main actor? We cannot even use `postupgrade` hooks, because canister calls are not allowed there.

crusso:
> How do we upgrade actor class at the moment? If we change the code, `moc` gives us a new Wasm for the actor class, but there is no way to upgrade the already deployed actor class instances, right? The developer will have to somehow write the upgrade logic in the main actor? We cannot even use `postupgrade` hooks, because canister calls are not allowed there.

You mean for an imported actor class?

Yeah, the user will need to develop a protocol for this. Given we can't upgrade more than one actor atomically, I'm not sure how that can be avoided - we don't even provide a programmatic way for doing an upgrade, just an install (tried, but couldn't locate the issue that discusses this). 

I was wondering if we should check the relation pointwise for each actor class in the program, but there's no real way to identify the actor classes anyway - doing it by name of the class seems arbitrary, especially since it's the program that determines the name on import, not the name of the file.

crusso:
@chenyan-dfinity what's the status of access the custom sections I now emit for an installed canister on the ic? Do I need to change anything or shall we just merge what is here and change it later if needed?

chenyan-dfinity:
> there's no real way to identify the actor classes anyway

Right. We don't even know the canister id for each instantiated actor class. So I'm not sure what we can check for upgrade.

> what's the status of access the custom sections

Let me test it.

chenyan-dfinity:
`./moc a.mo --public-metadata "candid:service"` produced a wasm module, but the exit code is 1

crusso:
> `./moc a.mo --public-metadata "candid:service"` produced a wasm module, but the exit code is 1

I'll ivestigate, but what I meant is what the status of the IC allowing access to custom sections of installed canisters  via http or IC ManagementCanister?

chenyan-dfinity:
@nomeata has an implementation of ic-ref that supports metadata from the state tree. Not from the management canister yet.

I will investigate tomorrow, somehow I always get empty `candid:service` content with the latest ic-ref and moc.

src/wasm-exts/customModuleDecode.ml
```diff
@@ -45,6 +51,7 @@ let peek s = if eos s then None else Some (read s)
 let get s = check 1 s; let b = read s in skip 1 s; b
 let get_string n s = let i = pos s in skip n s; String.sub s.bytes i n
 
+let checkpoint s = let p = !(s.pos) in fun () -> s.pos := p
```
rossberg, line 54: Nice!
```suggestion
let checkpoint s = let p = !(s.pos) in fun () -> s.pos := p

```

rossberg (approved):

src/wasm-exts/customModuleDecode.ml
crusso: Actually that whole module could use a rewrite since it's a bit precious about where custom sections can occur. I opened an issue for that #2890

crusso:
> `./moc a.mo --public-metadata "candid:service"` produced a wasm module, but the exit code is 1

Now fixed.

chenyan-dfinity:
With the greet example, here is what I read from the state tree:
candid:service ")service : {\n  greet: (text) -> (text);\n}\n"
candid:args "\u{2}()"
motoko:stable-types "\nactor {};\n"

The first character of candid section seems to be off?

crusso:
> With the greet example, here is what I read from the state tree: candid:service ")service : {\n greet: (text) -> (text);\n}\n" candid:args "\u{2}()" motoko:stable-types "\nactor {};\n"
> 
> The first character of candid section seems to be off?

Bugger, I'll take a look. I'd like to add some tests for this stuff but couldn't figure out a decent way to it. Maybe we just need a directory of custom bash tests.

crusso:
Another possibility is that the first chars are the (redundant) size of the string. I'll need to check the encoding used... Ascii 41, 2, 10 seem to encode the length.

crusso:
> With the greet example, here is what I read from the state tree: candid:service ")service : {\n greet: (text) -> (text);\n}\n" candid:args "\u{2}()" motoko:stable-types "\nactor {};\n"
> 
> The first character of candid section seems to be off?

Yeah, I checked the encoding I used and its uleb128 representation of a 32 bit string length, followed by raw bytes of utf8 string contents.

We could probably infer the size of the string from the size of the section and the size of the section name, but that seems a bit hacky and I prefer to stay compositional, even if its wasts a few bytes.

I do wonder if I should be double-checking checking the bytes are valid utf-8 before serializing though...

crusso:
@chenyan-dfinity 

In the notation of the wasm spec, I think the payload of these section is vec(bytes) = n:u32;byte^n, same as 
https://webassembly.github.io/spec/core/binary/values.html#names

chenyan-dfinity:
hmm, the size seems redundant, because the custom section has a size already. From the tooling perspective, it's better to remove the size in the content, otherwise the receiver needs a decoder for LEB128, which makes it harder to pipe with CLI tools, such as `didc` or `dfx`.

crusso:
> hmm, the size seems redundant, because the custom section has a size already. From the tooling perspective, it's better to remove the size in the content, otherwise the receiver needs a decoder for LEB128, which makes it harder to pipe with CLI tools, such as `didc` or `dfx`.

Makes sense, I'll take it out now unless @rossberg objects

crusso:
@chenyan-dfinity @rossberg PTAL (now using raw utf8 not size encode string for metadata sections)

src/wasm-exts/customModuleEncode.ml
```diff
@@ -841,16 +841,20 @@ let encode (em : extended_module) =
...
       section 0 (vec string) labels (labels <> [])
 
-    let stable_types_body s =
-      string s
+    let utf8 bs =
+      try
+        let _ = Wasm.Utf8.decode bs in
+        put_string s bs
+      with Wasm.Utf8.Utf8 ->
+        assert false
```
rossberg, line 849: Nit: This assertion is somewhat overcomplicated. Why not
```suggestion
      ignore (Wasm.Utf8.decode bs);  (* assert well-formedness *)
      put_string s bs
```

rossberg (approved):
Makes sense to me, the contents should be raw UTF-8, not an encoded string.

src/wasm-exts/customModuleEncode.ml
crusso: Don't I just get an uncaught exception then? Rather than a report of where the assertion failed. But ok.

chenyan-dfinity:
`moc --stable-types main.mo -o test.most` works correctly. But if I use `-o test.moi`, it still generates `test.most` for stable types, and `test.moi` becomes the Wasm module.

chenyan-dfinity:
Is is possible that the compatibility check fails, but the user can still use `preupgrade` and `postupgrade` hooks to prevent data loss?

crusso:
@rossberg @chenyan-dfinity I've added some documentation for this stuff. PTAL.  I'm wondering if I should add the slides I prepared as a separate section, but can do that later some time.

crusso:
> `moc --stable-types main.mo -o test.most` works correctly. But if I use `-o test.moi`, it still generates `test.most` for stable types, and `test.moi` becomes the Wasm module

For --stable-types, -o only affects the name of the binary. I produce stable-types as a by-product of compilation, not a separate process like `--idl` does. I don't actually understand why --idl isn't done the same way. At the moment, don't you need top type-check the code twice to produce .did and .wasm?

crusso:
> Is is possible that the compatibility check fails, but the user can still use `preupgrade` and `postupgrade` hooks to prevent data loss?

The postupgrade hook runs after we've deserialized the data and initialized stable field, IIRC, at which point the data will already be lost.

I'm not sure how preupgrade could be used - the loss happens on the other side of the upgrade. Have you got an example in mind?

chenyan-dfinity:
> For --stable-types, -o only affects the name of the binary. I produce stable-types as a by-product of compilation, not a separate process like --idl does.

But if `-o` is a `.most` file, it doesn't produce the Wasm. Not a problem for dfx, but it seems inconsistent with `--idl`. Maybe `--idl` should be a by-product as well. Yes, currently we run type check twice.

chenyan-dfinity:
> I'm not sure how preupgrade could be used - the loss happens on the other side of the upgrade. Have you got an example in mind?

This is probably a big can of worm. I was hoping this to work

v1:
```
  stable var counter : Nat = 0;
  stable var tmp : Nat = 0;
// inc counter before upgrade
```

v2:
```
stable var counter : [Nat] = [];
stable var tmp : Nat = 0;
system func preupgrade() {
  tmp := counter;  // This doesn't type check. But preupgrade is run before the upgrade, so counter should still have type Nat instead of [Nat]?
};
system func postupgrade() {
  counter := [tmp];
};
```

crusso:
> > For --stable-types, -o only affects the name of the binary. I produce stable-types as a by-product of compilation, not a separate process like --idl does.
> 
> But if `-o` is a `.most` file, it doesn't produce the Wasm. Not a problem for dfx, but it seems inconsistent with `--idl`.

I think it does produce the wasm .most and then overwrites it with the stable-sig .most file

> Maybe `--idl` should be a by-product as well. Yes, currently we run type check twice.

Yeah, I think we should fix that unless dfx is somehow relying on that behaviour. I wasn't sure.

--idl currently produces the full idl (with service args if any). 

I wonder if we should add --candid-service (sans args) and --candid-args that just spits out the custom section content, but to what files?

crusso:
> > I'm not sure how preupgrade could be used - the loss happens on the other side of the upgrade. Have you got an example in mind?
> 
> This is probably a big can of worm. I was hoping this to work
> 
> v1:
> 
> ```
>   stable var counter : Nat = 0;
>   stable var tmp : Nat = 0;
> // inc counter before upgrade
> ```
> 
> v2:
> 
> ```
> stable var counter : [Nat] = [];
> stable var tmp : Nat = 0;
> system func preupgrade() {
>   tmp := counter;  // This doesn't type check. But preupgrade is run before the upgrade, so counter should still have type Nat instead of [Nat]?
> };

Yeah, but this is the preupgrade for the *next* upgrade (from v2 to v3), not the previous one, so why should that work?

It's all pretty ugly to be honest. 

I find the `stable let` particularly bad since it suggests an equality that doesn't always on upgrade. For example, in-lining the let is completely broken.

 `stable var` is a bit more defensible.

doc/modules/language-guide/pages/upgrades.adoc
```diff
@@ -118,6 +106,70 @@ Note that the type of `+entries+`, being just an array of `+Text+` and `+Nat+` p
...
+* discard Motoko stable state (due to an incompatible change in stable declarations).
+
+A {proglang} canister upgrade is safe provided:
+
+* the canister's Candid interface evolves to a Candid subtype; and
+* the canister's Motoko stable signature evolves to a _stable-compatible_ one.
+
+Upgrade safety does not guarantee that the upgrade process will succeed (it can still fail due to resource constraints). However, it should at least ensure that a successful upgrade will not break Candid type compatibility with existing clients or unexpectedly lose data that was marked `stable`.
+
+TIP: You can check valid Candid subtyping between two services described in `.did` files, `cur.did` and `nxt.did` (containing Candid types), using  the `didc` tool with option  `--check nxt.did cur.did`.
```
crusso, line 157:
```suggestion
TIP: You can check valid Candid subtyping between two services described in `.did` files, `cur.did` and `nxt.did` (containing Candid types), using  the `didc` tool with argument `check nxt.did cur.did`.
```

chenyan-dfinity:
> Yeah, but this is the preupgrade for the next upgrade (from v2 to v3), not the previous one, so why should that work?

Make sense. I can move `preupgrade` to v1, but then the upgrade fails in IDL decoding. This is good, it means if compatibility check fails, upgrade will either fail or lose data.

> I think it does produce the wasm .most and then overwrites it with the stable-sig .most file

ah, I see.

> Yeah, I think we should fix that unless dfx is somehow relying on that behaviour. I wasn't sure.

Yep, we should fix `--idl` then. It makes `dfx` logic simpler. No need to run type checker twice. I think it's fine to output the full did file including the init args. dfx can parse the service constructor properly.

crusso:
> Yep, we should fix `--idl` then. It makes `dfx` logic simpler. No need to run type checker twice. I think it's fine to output the full did file including the init args. dfx can parse the service constructor properly.

In this PR? If so, does candid actually use the -o argument to put the candid file somewhere special, not just next to the wasm file?

chenyan-dfinity:
> If so, does candid actually use the -o argument to put the candid file somewhere special, not just next to the wasm file?

It is using the `-o` argument in dfx, but it's putting the did file next to the Wasm file with the same filename. So it's pretty much our default option.

crusso:
> > If so, does candid actually use the -o argument to put the candid file somewhere special, not just next to the wasm file?
> 
> It is using the `-o` argument in dfx, but it's putting the did file next to the Wasm file with the same filename. So it's pretty much our default option. This is a breaking change though.

Yeah, I was just looking at the code. How is it a breaking change? You mean old dfx won't be able to use a new compiler?

Another option would be to keep --idl, but introduce --candid --candid:service and --candid:args going forward, or just always emit the files by default. Yuck.

chenyan-dfinity:
> You mean old dfx won't be able to use a new compiler?

Yes. But `--stable-types` and `--idl` have different behavior, so it's worth fixing. If we have a Wasm parser, we don't even need to generate these files.

crusso:
If you come up with a plan, let me know and I'll try to implement it tomorrow. Or we could just punt for now, but I guess it might be better to clean up now.

chenyan-dfinity:
If it's easy to do, we can make `--idl` the same behavior as `--stable-types`, then we only need to call `moc` once in `dfx build`.

crusso:
@rossberg could you please take a look at the adoc and changelog changes?

doc/modules/language-guide/pages/upgrades.adoc
```diff
@@ -118,6 +106,70 @@ Note that the type of `+entries+`, being just an array of `+Text+` and `+Nat+` p
...
+  stable var y : Int;
+  stable z : [var Nat];
+};
+....
+
+It specifies the names, types and mutability of the actor's stable fields, possibly preceded by relevant Motoko type declarations.
+
+TIP: You can emit the stable signature of the main actor or actor class to a `.most` file using `moc` compiler option `--stable-types`. You should never need to author your own `.most` file.
+
+A stable signature, `<stab-sig1>`, is _stable-compatible_ with signature `<stab-sig2>`, if, and only,
```
rossberg, line 129: Might help to clarify here which sig is assumed to be the new and which the old one.

doc/modules/language-guide/pages/upgrades.adoc
```diff
@@ -118,6 +106,70 @@ Note that the type of `+entries+`, being just an array of `+Text+` and `+Nat+` p
...
+* every immutable field `stable <id> : T` in `<stab-sig1>` has a matching field `stable <id> : U` in `<stab-sig2>` with `T <: U`.
+* every mutable field `stable var <id> : T` in `<stab-sig1>` has a matching field `stable var <id> : U` in `<stab-sig2>` with `T <: U`.
+
+Note that `<stab-sig2>` may contain additional fields.
+
+The subtyping condition on stable fields ensures that the final value of some field can be consumed as the initial value of that field in the upgraded code.
+
+TIP: You can check the stable-compatiblity of two `.most` files, `cur.most` and `nxt.most` (containing stable signatures), using `moc` compiler option `--stable-compatible cur.most nxt.most`.
+
+NOTE: The _stable-compatible_ relation is quite conservative. In future, it may be relaxed to accommodate a change in field mutability
```
rossberg, line 140:
```suggestion
NOTE: The _stable-compatible_ relation is quite conservative. In the future, it may be relaxed to accommodate a change in field mutability
```

rossberg (approved):
Looks great!

doc/modules/language-guide/pages/upgrades.adoc
crusso:
```suggestion
A stable signature, `<stab-sig1>`, is _stable-compatible_ with signature `<stab-sig2>` (the upgrade), if, and only,
```

doc/modules/language-guide/pages/upgrades.adoc
```diff
@@ -118,6 +106,70 @@ Note that the type of `+entries+`, being just an array of `+Text+` and `+Nat+` p
...
+  stable var y : Int;
+  stable z : [var Nat];
+};
+....
+
+It specifies the names, types and mutability of the actor's stable fields, possibly preceded by relevant Motoko type declarations.
+
+TIP: You can emit the stable signature of the main actor or actor class to a `.most` file using `moc` compiler option `--stable-types`. You should never need to author your own `.most` file.
+
+A stable signature, `<stab-sig1>`, is _stable-compatible_ with signature `<stab-sig2>` (the upgrade), if, and only,
```
crusso, line 129:
```suggestion
A stable signature `<stab-sig1>` is _stable-compatible_ with signature `<stab-sig2>`, if, and only,
```

doc/modules/language-guide/pages/upgrades.adoc
```diff
@@ -118,6 +106,70 @@ Note that the type of `+entries+`, being just an array of `+Text+` and `+Nat+` p
...
+
+* every immutable field `stable <id> : T` in `<stab-sig1>` has a matching field `stable <id> : U` in `<stab-sig2>` with `T <: U`.
+* every mutable field `stable var <id> : T` in `<stab-sig1>` has a matching field `stable var <id> : U` in `<stab-sig2>` with `T <: U`.
+
+Note that `<stab-sig2>` may contain additional fields.
```
crusso, line 134:
```suggestion
Note that `<stab-sig2>` may contain additional fields.
Typically, `<stab-sig1>` is the signature of an older version while `<stab-sig2>` is the signature of a newer version. 
```

src/exes/moc.ml
```diff
@@ -27,14 +27,31 @@ let link = ref true
...
+  "--stable-compatible",
+    Arg.Tuple [
+      Arg.String (fun fp -> Flags.pre_ref := Some fp);
+      Arg.String (fun fp -> Flags.post_ref := Some fp);
+      Arg.Unit (set_mode StableCompatible);
+      ],
+    "<pre> <post> test upgrade compatibility between stable-type signatures <pre> and <post>";
+  "--idl", Arg.Unit (fun () ->
+    idl := true;
+    set_mode Compile ()), (* HACK to support dfx 0.8.3 or earlier *)
```
crusso, line 53:
```suggestion
    set_mode Compile ()), (* similar to --stable-types *)
```

PR #2889, opened 2021-11-08
Backport random heap tests from osa1/page_alloc

osa1:
This implements a very simple random heap generation that caught a lot of bugs
in #2706. The idea is similar to QuickCheck/proptest style property testing: we
generate random heaps using a pseudo-random number generator and a fixed seed,
and then run the GC tests as usual (run both collectors, check GC properties).

Difference from QuickCheck/proptest-style testing is we don't implement
shrinking. Implementing shrinking manually is not a lot of work, but I believe
it won't work in practice, because checking subsets of the nodes/edges of the
graph will take exponential time (N nodes will have 2^N subsets of nodes), and
we really need to check all subsets as these graphs won't have any properties
that allows us to shrink more efficiently.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -191,7 +191,7 @@ rec {
         name = "motoko-rts-deps";
         src = subpath ./rts;
         sourceRoot = "rts/motoko-rts-tests";
-        sha256 = "129gfmn96vm7di903pxirg7zybl83q6nkwiqr3rsy7l1q8667kxx";
+        sha256 = "07i8mjky9w0c9gadxzpfvv9im40hj71v69a796q7vgg2j6agr03q";
         copyLockfile = true;
       };
 
--- a/rts/motoko-rts-tests/Cargo.lock
+++ b/rts/motoko-rts-tests/Cargo.lock
@@ -84,6 +84,7 @@ dependencies = [
  "fxhash",
  "libc",
  "motoko-rts",
+ "oorandom",
  "proptest",
 ]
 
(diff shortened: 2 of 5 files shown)


nomeata:
It won't be necessary exponential, as shrinking tends to remove one element at a time, and only progress if the smaller test case is still a falling case. So it may take quadratic time if the failure can be shrunk this way, or it will fail after linear time if it can't be shrunk this way. 

But the benefit on debugging is huge. So if shrinking can be implemented I suggest to try this. Different PR though

crusso (approved):
(rubber stamp from me)

osa1:
> It won't be necessary exponential, as shrinking tends to remove one element at a time, and only progress if the smaller test case is still a falling case

I think this could work, yes. I'm guessing for a large heap that triggers a bug, most of the nodes will be irrelevant, and if we simply remove those nodes one by one we may be able to simplify the graph quite a bit.

We could also try removing edges, one edge at a time, instead of nodes, while keeping object sizes the same (this is currently not supported, we can't have an object with scalar fields, but that's easy to fix).

This way of shrinking won't be able to shrink the original repro when removing one node hides the bug, if you remove two particular nodes it fails again. I don't know any concrete example for this, but I wouldn't be surprised if this is common.

I think we won't be able to use proptest for implementing this way of shrinking, because [its shrinking API][1] assumes a linear search space where buggy-ness changes monotonically (i.e. it wants to do binary search). It shouldn't be too hard to implement this manually though.

[1]: https://altsysrq.github.io/rustdoc/proptest/1.0.0/proptest/strategy/trait.ValueTree.html#tymethod.simplify

PR #2891, opened 2021-11-09
Make mark-compact the default collector, improve scheduling

osa1:
This PR makes mark-compact the default GC strategy and improves GC scheduling.

New flag `--copying-gc` added to enable copying GC. `--compacting-gc` is now
enabled by default.

# Motivation

Mark-compact GC uses more cycles than copying GC (see below), but in return it
almost halves dirtied pages, and allows us to use a more efficient GC scheduler
as it works with less scratch space. When both of these are taken into account,
it becomes clear that mark-compact GC is much more efficient than copying GC,
and should be the default.

Furthermore, when used with the mark-compact collector, a new GC scheduler
makes it possible for the tested canister to run same number of calls without
getting stuck, while doing less GC.

## Benchmarking

**Comparing GC cycles and dirtied pages**: We run an old [CanCan backend][1]
canister, compiled using `--force-gc`. drun script contains 1,000 calls to
`createProfile`, with different arguments:

    ingress rwlgt-iiaaa-aaaaa-aaaaa-cai createProfile "DIDL\x02n\x01m{\x02q\x00\x05test0\x00"

For drun we use dfinity `d3b184f303`, which has a modified replica that
generates a `canister_perf.csv` file after each run with cycles, read pages,
dirtied pages, and total Wasm pages.

Plots are then generated using the `canister_perf.csv` files generated by the
previous step using [generate_plots][2].

Results of the last call, comparing copying GC and compacting GC:

- Instructions: +34.7% (+12,052,682 cycles)
- Accessed pages: -47.5% (-296 pages)
- Dirtied pages: -47.8% (-296 pages)
- Total Wasm pages: -43.9% (-18 pages)

Execution team estimates that a single dirtied page costs between 10,000 and
100,000 cycles to the replica. If we assume 10,000, cycle costs of compacting
GC in the last call becomes 22% more (+9,092,682 cycles). If we assume 100,000,
then compacting GC uses 18% *less* cycles (-17,547,318 cycles). For this
canister, in the last call, compacting GC becomes more efficient in terms of
cycles when a dirtied page costs 40,719 or more cycles.

Below are plots comparing instructions, dirtied pages, read pages, and Wasm
pages for copying and compacting collectors.

<details>

<summary>Instructions, dirtied host pages, accessed host pages, total Wasm pages</summary>

instructions

total_instructions

dirtied_host_pages

total_dirtied_host_pages

accessed_host_pages

total_accessed_host_pages

total_Wasm_pages_in_use

</details>

## Testing and benchmarking new scheduler

For benchmarking the new scheduler we use a [simulation][3].

We compare 3 schedulers:

1. The current one

2. The current one, except `MAX_HP_FOR_GC` is set 2 GiB.

   Reminder: `MAX_HP_FOR_GC` is the point where we start doing GC after every
   update call. Current value for it is 1 GiB.

3. `limit = min(live * 1.5, (live + max_live) / 2)` where `live` is the live
   amount after the last GC, and `max_live` is the max. live data allowed by the collector.

   For the copying GC, `max_live` is 2 GiB. For the mark-compact GC, `max_live`
   is 4,164,816,771 bytes. To calculate this number, we ignore the mark stack
   size. Mark stack size depends on the heap contents rather than size. In
   CanCan backend benchmark, it's 1024 words large for 1 GiB heap. To keep
   things simple we assume that the size is constant and negligible.

   Bitmap size can be at most 130,150,524 and the rest will be the heap. This
   is calculated with

       x + x/32 = 4 GiB, x = 4,164,816,771, x/32 = 130,150,524

   Reminder: one word in bitmap addresses 32 words in heap.

(Reminder: in the benchmarking section above, we use `--force-gc`, so the
scheduler is disabled and we do GC after every call)

The canister used allocates 100,000 bytes in each call, 50% of which survive
the GC. Unless specified, all GC parameters (`GROWTH_FACTOR`) are unchanged.

Metrics:

- Calls: number of calls the canister is able to make before getting stuck
  (Reminder: canister keeps allocating, and 50% of allocations are live)

- GCs: number of GCs scheduled

**Results of scheduler (1):**

|               | Calls           | GCs              |
| ------------- | --------------- | ---------------- |
| Copying       | 42,950          | 21,508           |
| Compacting    | 83,296          | 61,854           |
| Diff          | +40,346, +93.9% | +40,346, +187.5% |

Summary: the canister lives much longer with the compacting GC. Cycle usage
when this scheduler is used is as shown in the previous section.

<details>

<summary>Plots</summary>

copying_old_1gb

compacting_old_1gb

</details>

**Results of scheduler (2):**

|               | Calls           | GCs               |
| ------------- | --------------- | ----------------- |
| Copying       | 42,949          | 36                |
| Compacting    | 83,296          | 40,383            |
| Diff          | +40,347, +93.9% | +40,347, +112075% |

Summary: Setting `MAX_HEAP_FOR_GC` 2 GiB, does not make any difference in how
long the canister can run before getting stuck. However, we do less GC.
Especially with copying GC, this scheduler performs extremely well.

<details>

<summary>Plots</summary>

copying_old_2gb

compacting_old_2gb

</details>

**Results of scheduler (3):**

|               | Calls           | GCs               |
| ------------- | --------------- | ----------------- |
| Copying       | 42,950          | 74                |
| Compacting    | 83,296          | 78                |
| Diff          | +40,346, +93.9% | +4, +5.4%         |

Summary: this scheduler yields the best results when used on both collectors.
For copying GC, it's the same as (2). For mark-compact GC, the canister can run
for the same amount of calls, but number of GCs is significantly lower.

<details>

<summary>Plots</summary>

copying_new

compacting_new

</details>

[1]: https://github.com/dfinity/cancan-archived/
[2]: https://github.com/osa1/generate_plots
[3]: https://github.com/osa1/mmsim_2

### DIFF
--- a/rts/motoko-rts/src/gc/copying.rs
+++ b/rts/motoko-rts/src/gc/copying.rs
@@ -1,4 +1,4 @@
-use crate::constants::WORD_SIZE;
+use crate::constants::{WASM_HEAP_SIZE, WORD_SIZE};
 use crate::mem_utils::{memcpy_bytes, memcpy_words};
 use crate::memory::Memory;
 use crate::types::*;
@@ -7,9 +7,12 @@ use motoko_rts_macros::ic_mem_fn;
 
 #[ic_mem_fn(ic_only)]
 unsafe fn schedule_copying_gc<M: Memory>(mem: &mut M) {
-    const MAX_LIVE: Bytes<u64> = Bytes(2 * 1024 * 1024 * 1024); // 2 GiB
+    // Half of the heap.
+    // NB. This expression is evaluated in compile time to a constant.
+    let max_live: Bytes<u64> =
+        Bytes(u64::from((WASM_HEAP_SIZE / 2).as_u32()) * u64::from(WORD_SIZE));
 
-    if super::should_do_gc(MAX_LIVE) {
+    if super::should_do_gc(max_live) {
         copying_gc(mem);
     }
 }
--- a/rts/motoko-rts/src/gc/copying.rs
+++ b/rts/motoko-rts/src/gc/copying.rs
@@ -7,7 +7,9 @@ use motoko_rts_macros::ic_mem_fn;
 
 #[ic_mem_fn(ic_only)]
 unsafe fn schedule_copying_gc<M: Memory>(mem: &mut M) {
-    if super::should_do_gc() {
+    const MAX_LIVE: Bytes<u64> = Bytes(2 * 1024 * 1024 * 1024); // 2 GiB
+
+    if super::should_do_gc(MAX_LIVE) {
         copying_gc(mem);
     }
 }
--- a/rts/motoko-rts/src/gc/copying.rs
+++ b/rts/motoko-rts/src/gc/copying.rs
@@ -1,4 +1,4 @@
-use crate::constants::{WASM_HEAP_SIZE, WORD_SIZE};
+use crate::constants::WORD_SIZE;
 use crate::mem_utils::{memcpy_bytes, memcpy_words};
 use crate::memory::Memory;
 use crate::types::*;
@@ -10,7 +10,7 @@ unsafe fn schedule_copying_gc<M: Memory>(mem: &mut M) {
     // Half of the heap.
     // NB. This expression is evaluated in compile time to a constant.
     let max_live: Bytes<u64> =
-        Bytes(u64::from((WASM_HEAP_SIZE / 2).as_u32()) * u64::from(WORD_SIZE));
+        Bytes(u64::from((crate::constants::WASM_HEAP_SIZE / 2).as_u32()) * u64::from(WORD_SIZE));
 
     if super::should_do_gc(max_live) {
         copying_gc(mem);
--- a/rts/motoko-rts/src/gc/mark_compact.rs
+++ b/rts/motoko-rts/src/gc/mark_compact.rs
@@ -20,13 +20,13 @@ use motoko_rts_macros::ic_mem_fn;
 unsafe fn schedule_compacting_gc<M: Memory>(mem: &mut M) {
     // 512 MiB slack for mark stack + allocation area for the next message
     let slack: u64 = 512 * 1024 * 1024;
-    // x + slack + x/32 = 4 GiB, x = 3,644,214,675, x/32 = 113,881,708
+    // x + slack + x/32 = heap_size
     // (Reminder: one word in bitmap covers 32 words in heap)
     // NB. `max_live` below is evaluated in compile time to a constant
     let heap_size_bytes: u64 = u64::from(WASM_HEAP_SIZE.as_u32()) * u64::from(WORD_SIZE);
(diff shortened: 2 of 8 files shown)


osa1:
TODO: We should say at least a few words on how these results change when we change the alloc or survival rates.

rts/motoko-rts/src/gc/copying.rs
```diff
@@ -7,7 +7,9 @@ use motoko_rts_macros::ic_mem_fn;
 
 #[ic_mem_fn(ic_only)]
 unsafe fn schedule_copying_gc<M: Memory>(mem: &mut M) {
-    if super::should_do_gc() {
+    const MAX_LIVE: Bytes<u64> = Bytes(2 * 1024 * 1024 * 1024); // 2 GiB
```
ulan, line 10: Don't we have a constant for max-wasm-memory-size? If not, can we introduce it?

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -18,7 +18,11 @@ use motoko_rts_macros::ic_mem_fn;
 
 #[ic_mem_fn(ic_only)]
 unsafe fn schedule_compacting_gc<M: Memory>(mem: &mut M) {
-    if super::should_do_gc() {
+    // Mark stack ignored. Max. bitmap can be 130,150,524 bytes.
+    // (x + x / 32 = 4 GiB, x = 4,164,816,771, x/32 = 130,150,524)
+    const MAX_LIVE: Bytes<u64> = Bytes(4_164_816_771);
```
ulan, line 23: Please express this in terms of max-wasm-memory-size and max-marking-bitmap-size.

Shall we also add a some slack for subsequent message execution and marking stack? The copying GC doesn't have this problem because the mutator can continue allocating in the second half of the heap.

To be on the safe side, let's leave 500MB-1GB of slack? 

`MAX_LIVE = MAX_WASM_MEMORY - MAX_MARKING_BITMAP - SLACK`

ulan:
lgtm from my side. Thanks for benchmarking!

ulan (approved):

rossberg (approved):
Excellent!

rts/motoko-rts/src/gc/mark_compact.rs
crusso: Hmm, maybe we could put all but the first page of markstack in stable memory in future, paging out/in in stack order....

crusso (approved):
Will this change mess with our perf diffs in CI? I guess just once ;->

rts/motoko-rts/src/gc/mark_compact.rs
osa1: I reserved 500 MiB for the mark stack.

osa1:
@ulan I addressed your comments, could you review again please?

ulan:
The latest commit looks good, thanks!

PR #2893, opened 2021-11-10
example showing serialization stack overflow during upgrade

crusso:
Demonstrate failure to upgrade due to candid serialization failure on deeply nested data structure (stack overflow).

### DIFF
--- a/test/run-drun/message-overflow.mo
+++ b/test/run-drun/message-overflow.mo
@@ -10,7 +10,7 @@ actor a {
    };
 
    public func go () : async () {
-      let n = toN(10000, null);
+      let n = toN(50000, null);
       await m(n);
    };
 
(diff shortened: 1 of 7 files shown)


crusso:
@rossberg @ulan  another nail in the coffin for candid stabilization

rossberg:
He he. But isn't this a problem for our Candid implementation in general? Shouldn't it be impossible to send such a list to another canister?

crusso:
> He he. But isn't this a problem for our Candid implementation in general? Shouldn't it be impossible to send such a list to another canister?

Yes, it is, but given the 3MB message limit, it might be hard to exercise it. I'm not sure if it affects the Rust implementation but we should check. @chenyan-dfinity did you try this with Rust?

nomeata:
With deeply nested single field records (which don't cost extra bytes in the data array) it should be easy to get very deeply nested data structures in 2M, at least quadratic in input size, so maybe a billion deep.

luc-blaeser:
Solved by enhanced orthogonal persistence

PR #2895, opened 2021-11-11
Pass trap continuation to prim interpreter, make call sites tail recursive

### LINKED ISSUE #2844: BUG: Overflow in Nat.toText?
https://forum.dfinity.org/t/arithmetic-overflow/7955

osa1:
This commit makes calls to prim interpreter (Mo_values.Prim.prim) tail
recursive by passing a "trap" continuation to it.

Previously call sites looked like this:

    try Prim.prim s env v k
    with Invalid_argument s -> trap exp.at "%s" s

This code is not tail recursive, so recursive calls to the interpreter from the
prim interpreter causes stack overflow.

With this PR we now pass a continuation for failing:

    Prim.prim { Prim.trap = trap exp.at "%s" } s env v k

The new type `trap` introduced to allow rank-2 error function. In Haskell
syntax, type of the function is

    forall a . String -> a

Without rank-2 error function, the code in prim interpreter fails to type check
as we expect the result of `trap` applications to have different types.

In the reproducer for #2844, stack overflow manifests itself as different
errors. In most cases it fails as "execution error, arithmetic overflow", but
I've also seen it fail with segfault. I don't know if this is supposed to
happen, or a bug in OCaml RTS.

Fixes #2844


osa1:
I was looking for something simple to work on and wanted to do this (the issue is assigned to @ggreif so I asked Gabor first).

I will try to add a test. I will add an upper bound to the loop, it should loop enough to overflow the stack without this PR.

Something interesting that I discovered is, in this repro:

```
import Prim "mo:⛔";

loop {
    Prim.debugPrint(debug_show(136 : Nat32 / 32));
}
```

If I run this with `-iR` (IR interpreter), both in master and in this branch, it doesn't print anything and exits with 0. Source interpreter works fine. I think this is another bug.

ggreif (approved):
This looks really cool. The way we had to catch exceptions in the middle always irked me, and resorting to a rank-2 continuation is an elegant way of eliminating those. There are two things that surprised me
- the low number of call sites where the trap continuation needs to be passed down (I expected a lot more)
- that rank-2 types need to be wrapped in a one-field record (would it be possible to define `type trap = 'a. string -> 'a`?

osa1:
> that rank-2 types need to be wrapped in a one-field record (would it be possible to define type trap = 'a. string -> 'a?

As far as I understand for rank-2 polymorphism we need records or objects: https://ocaml.org/manual/polymorphism.html#s:higher-rank-poly

src/mo_values/prim.ml
```diff
@@ -43,22 +45,22 @@ let of_big_int_wrap = function
...
+let num_conv_trap_prim trap t1 t2 =
   let module T = Type in
   match t1, t2 with
   | T.Nat, T.(T.Nat8|Nat16|Nat32|Nat64)
   | T.Int, T.(Int8|Int16|Int32|Int64)
   | T.(Nat8|Nat16|Nat32|Nat64), T.Nat
   | T.(Int8|Int16|Int32|Int64), T.Int
   | T.Nat32, T.Char
-  -> fun v -> of_big_int_trap t2 (as_big_int t1 v)
+  -> fun v -> of_big_int_trap trap t2 (as_big_int trap t1 v)
```
rossberg, line 63:
```suggestion
  | T.Nat32, T.Char ->
    fun v -> of_big_int_trap trap t2 (as_big_int trap t1 v)
```

src/mo_values/prim.ml
```diff
@@ -185,7 +187,7 @@ let prim =
                                in k (Text str)
   | "print" -> fun _ v k -> Printf.printf "%s\n%!" (as_text v); k unit
   | "trap" -> fun _ v k ->
-    raise (Invalid_argument ("explicit trap: "^ (as_text v)))
+    trap.trap ("explicit trap: " ^ (as_text v))
```
rossberg, line 190:
```suggestion
  | "trap" -> fun _ v k -> trap.trap ("explicit trap: " ^ (as_text v))
```

rossberg (approved):
Nice!

crusso (commented):
Nice!

Are you sure we need that rank 2 type? We don't need need it for the success continuation k (which returns unit) and I wonder if re-expressing the num_conv prims to take both a success and a trap continuation would let you avoid it.

crusso:
The ir interpreter is also written in cps and could use a similar improvement I expect...

osa1:
Claudio points out the the `raise (Invalid_argument ...)` calls in the prim interpreter are actually for the cases where we detect a compiler (maybe type checker) bug. I think we don't want to catch those exceptions in the call sites to convert them into traps, right? We could catch them at the top level in the compiler and ask the user to report it.

If we do that then I think the prim interpreter won't need exception handling because as far as I can see all the error sites there (`raise (Invalid_argument ...)`) are for compiler bugs.

nomeata:
This has approvals. Should it be merged?

ggreif (changes requested):
@crusso pointed out that this is not optimal yet. Let's not merge at this point.

nomeata:
> @crusso pointed out that this is not optimal yet. Let's not merge at this point.

So, what’s not optimal about it? :-)

PR #2896, opened 2021-11-12
Physical equality for `Variant`

ggreif:
## How it works

There are cases when two objects being compared for equality are located in the same memory locations. Then, trivially every subobject (bytes) is the same, so the comparison will yield `true`.

### Analysis

It makes sense to analyse which shape of objects (in memory) is unlikely to exhibit pointer equality while being structurally equal. In such cases the address equality test is mostly done in vain.

When we have a branching structure in memory then it is unlikely that pointer comparison succeeds because the object could have been built in many ways. This basically means that product-like types don't tend to exhibit pointer equality when being equal.

OTOH coproduct-like types (options, variants) have tree-like type shapes, but tend to be compact in memory, so these are our candidates. Turns out that option types already have a nice and efficient memory layout, such that we don't attempt to optimise more. For variants, however, we can further take the tag (resp. hash) equality into account, which can negate the structural equality. For flat (i.e. enum-like) variants we can totally optimise away the costly switch on the pair of inputs, which is linear in the alternatives and currently even allocates memory.

## Primitives added

We add two primitives:

| Primitive | Type | Constraint |
|--|--|--|
| `SameReferece` | `<A> A -> A -> Bool` | |
| `SameVariantTag ` | `<A> A -> A -> Bool` | `A` must be variant type |

## Things to figure out

- are variants (with flat alts) coming out of deserialise are pointer-equal?
- if not, can we internalise in a way that we reuse objects from the static heap?

- `func $to_2_tuple` allocates (called by diagonalised `Variant` equality)
- [x] for _enum-like_ variant types only compare hashes (in the fallback after pointer-equality)
- [x] conversely, a differing `Variant` hash is a proof of inequality
- [x] don't mention flat alternatives in the diagonal switch, instead have a catch-all that just compares the tags


src/ir_def/check_ir.ml
```diff
@@ -489,8 +489,9 @@ let rec check_exp env (exp:Ir.exp) : unit =
       typ exp2 <: T.nat;
       T.bool <: t
     | SameReference, [exp1; exp2] ->
-      (*FIXME typ exp1 <: T.nat;
-      typ exp2 <: T.nat;*)
+      let t1, t2 = typ exp1, typ exp2 in
+      t1 <: t2;
```
ggreif, line 493: One could put the type in `SameReference` as an argument, and test both expressions against it here. Suggestion by @crusso.

src/ir_def/check_ir.ml
```diff
@@ -488,6 +488,15 @@ let rec check_exp env (exp:Ir.exp) : unit =
       typ exp1 <: T.nat;
       typ exp2 <: T.nat;
       T.bool <: t
+    | SameReference, [exp1; exp2] ->
+      let t1, t2 = typ exp1, typ exp2 in
+      t1 <: t2;
+      t2 <: t1; (* i.e. t1 = t2 *)
+      T.bool <: t
+    | SameVariantTag tv, [exp1; exp2] ->
+      typ exp1 <: tv;
```
ggreif, line 497: I guess we should `assert` that `tv` is a variant type here.

src/ir_def/ir.ml
```diff
@@ -133,6 +133,9 @@ and prim =
   | ValidArrayOffset                  (* verify array offset *)
   | DerefArrayOffset                  (* array offset indexing *)
   | GetPastArrayOffset of spacing     (* array offset past the last element *)
+  (* Material/Variant equality *)
```
ggreif, line 136: "Physical"
```suggestion
  (* Physical/Variant equality *)
```

src/ir_passes/eq.ml
```diff
@@ -134,16 +146,22 @@ let eq_for : T.typ -> Ir.dec * T.typ list = fun t ->
...
+    define_eq_variant t (trueE ()),
+    List.map (fun (f : T.field) -> T.normalize f.T.typ) fs
   | T.Variant fs ->
-    define_eq t (
+    let flat, deep = List.partition is_flat_alt fs in
+    assert (deep <> []);
+    define_eq_variant t (
       (* switching on the diagonal *)
       { it = SwitchE
         ( tupE [arg1E t; arg2E t],
```
crusso, line 167: To avoid the tuple, would it be better to generate nested switches?

src/ir_passes/eq.ml
```diff
@@ -78,6 +88,8 @@ let array_eq_func_body : T.typ -> Ir.exp -> Ir.exp -> Ir.exp -> Ir.exp = fun t f
 
 (* Returns the new declarations, as well as a list of further types it needs *)
 
+let is_flat_alt f = T.(is_unit (normalize f.typ))
```
crusso, line 91: Does this generalize to all singleton types, not just unit? Like the empty record, null, and singleton variants of singleton types? Not that anyone in their right mind would be using any of those I guess (except perhaps null).

crusso (commented):
Do you have a perf measurement to show this is worthwhile? Getting rid of the tupling would be nice.

src/ir_def/check_ir.ml
crusso: Yeah, I think so. It's a bit of a shame this isn't actually something you can encode in the type, but there's no common upper bound for variants as there is for records, right, just a lower one (and vice versa).

src/ir_def/check_ir.ml
crusso: Yeah, I think I really would.

src/ir_passes/eq.ml
ggreif: Nope, I think we could use multi-value (`SR.UnboxedTuple 2`) to pass around pairs on stack and keep them in variables.

src/ir_passes/eq.ml
ggreif: _Very_ good point. We have a `Type.is_singleton` predicate already!

src/ir_def/check_ir.ml
ggreif: I do it with the other primitive. SMOP.

src/ir_def/check_ir.ml
ggreif: Can't we say `{#} <: type exp1` etc.?

ggreif:
Quick idea, without thinking it through now: what about loading the heap tag _and_ the variant tag at the same time and doing a 64-bit comparison?

nomeata:
Lacking context, I think. Why even load the heap tag when looking at variants?

ggreif:
> Lacking context, I think. Why even load the heap tag when looking at variants?

Only for 1 reason I can think of: currently the hashing function is is `name -> 31 bits` (as opposed to the Candid hash that has full 32 bits). If we extend the heap tag for variants to be two possibilities, then we can push that missing I bit into it.

nomeata:
I think we don’t need to worry about that one bit; Motoko had 31 bit hashes before we even did Candid, I vaguely remember.

luc-blaeser:

PR #2905, opened 2021-11-17
Make sure that one-byte element optimised iteration still works

ggreif:
Maybe later we'll have some compression here and better have a regression test in place, just in case...


ggreif:
The SHA-256 folks will come after us if this ever breaks.

nomeata (approved):

PR #2907, opened 2021-11-17
Make bignum equality a bit more performant

ggreif:
... by doing a pointer comparison first, and when not equal
proceed as usual (except when both are compact, the result
is `false`, as we already checked for identity).

Well, I conjecture that this helps, and the actual cycle counts w.r.t. our benchmarks indicate an improvement of 0.3%, so I guess this could go in.

The discussion happened in JIRA [IC-622], but the gist is captured in the PR.

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -1584,7 +1584,7 @@ module Word64 = struct
         get_n ^^ get_exp ^^ compile_unsigned_pow env
       )
 
-  let compile_eq env = G.i (Compare (Wasm.Values.I64 I64Op.Eq))
+  let _compile_eq env = G.i (Compare (Wasm.Values.I64 I64Op.Eq))
   let compile_relop env i64op = G.i (Compare (Wasm.Values.I64 i64op))
 
 end (* BoxedWord64 *)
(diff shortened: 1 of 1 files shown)


ggreif:
``` wasm
  (func $B_eq (type 11) (param $a i32) (param $b i32) (result i32)
    local.get $a
    local.get $b
    i32.eq
    if (result i32)  ;; label = @1
      i32.const 1
    else
      local.get $a
      local.get $b
      i32.or
      i32.const 1
      i32.and
      if (result i32)  ;; label = @2
        local.get $a
        i32.const 1
        i32.and
        if (result i32)  ;; label = @3
          local.get $a
        else
          local.get $a
          i64.extend_i32_s
          i64.const 1
          i64.shr_s
          call $bigint_of_int64
        end
        local.get $b
        i32.const 1
        i32.and
        if (result i32)  ;; label = @3
          local.get $b
        else
          local.get $b
          i64.extend_i32_s
          i64.const 1
          i64.shr_s
          call $bigint_of_int64
        end
        call $bigint_eq
      else
        i32.const 0
      end
    end)
```

rossberg (approved):
Cool, thanks!

PR #2908, opened 2021-11-18
Align semantics of --idl and --stable-types command line options

crusso:
@chenyan-dfinity 

My attempt to converge the --idl and --stable-types so they are consistent without breaking dfx or requiring two compiles.

In dfx. just add --idl to the second invocation, delete the first, and do the sensible things with the error settings.

Fingers crossed.

If this is ok, and works for you, I'll merge it into the main PR (#2887) and pull the trigger.

### DIFF
--- a/src/exes/moc.ml
+++ b/src/exes/moc.ml
@@ -10,14 +10,14 @@ let usage = "Usage: " ^ name ^ " [option] [file ...]"
 
 (* Argument handling *)
 
-type mode = Default | Check | StableCompatible | Compile | Run | Interact | Idl | PrintDeps | Explain
+type mode = Default | Check | StableCompatible | Compile | Run | Interact | PrintDeps | Explain
 
 let mode = ref Default
 let args = ref []
 let add_arg source = args := !args @ [source]
 
 let set_mode m () =
-  if !mode <> Default then begin
+  if !mode <> Default && !mode <> m then begin
     eprintf "moc: multiple execution modes specified"; exit 1
   end;
   mode := m
@@ -28,6 +28,7 @@ let interpret_ir = ref false
 let gen_source_map = ref false
 let explain_code = ref ""
 let stable_types = ref false
+let idl = ref false
 
 let valid_metadata_names =
     ["candid:args";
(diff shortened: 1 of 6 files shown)


rossberg (approved):
LGTM, but @chenyan-dfinity probably understands the subtlety of the CLI better than I.

src/exes/moc.ml
```diff
@@ -46,8 +47,11 @@ let argspec = [
       Arg.String (fun fp -> Flags.post_ref := Some fp);
       Arg.Unit (set_mode StableCompatible);
       ],
-   "<pre> <post> test upgrade compatibility between stable-type signatures <pre> and <post>";
-  "--idl", Arg.Unit (set_mode Idl), " generate IDL spec";
+    "<pre> <post> test upgrade compatibility between stable-type signatures <pre> and <post>";
+  "--idl", Arg.Unit (fun () ->
+    idl := true;
+    set_mode Compile ()), (* HACK to support dfx 0.8.3 or earlier *)
```
chenyan-dfinity, line 53: Why is this a hack? We need to set `moc` to compile mode for `--idl`. 

Is there a flag check to disallow `--idl` in interpreter or wasi mode?

src/exes/moc.ml
```diff
@@ -262,7 +269,7 @@ let () =
   Internal_error.setup_handler ();
   Arg.parse_expand argspec add_arg usage;
   if !mode = Default then mode := (if !args = [] then Interact else Compile);
-  Flags.compiled := (!mode = Compile || !mode = Idl);
+  Flags.compiled := !mode = Compile;
```
chenyan-dfinity, line 272: Nice! One more reason we need to fix this.

src/exes/moc.ml
crusso: Well, it's a hack in make the first invocation of old dfx versions to work, by forcing compilation

crusso: But I can remove the comment...

src/exes/moc.ml
crusso: Unfortunately, removing the Pipeline.generate_idl broke moc.js so I just restored it...

crusso: Two steps forwards, one step back I guess

chenyan-dfinity (approved):
Let's merge, and I will try with dfx in the main branch.

crusso:
> Let's merge, and I will try with dfx in the main branch.

Sure, I'm just waiting for CI to finish...

crusso:
Got impatient and merged. Will need to do a full CI anyway,...

PR #2913, opened 2021-11-19
Enable tentative release artefact builds on `master`

ggreif:
This enables building of release products on push-events for the `master` branch. It neither sends them to `cachix`, nor attempts to upload them to any release. As previously those activities only happen when a new `git tag` is created.

I took the liberty to streamline the logic a bit, further opportunities exist (commented in PR) but you have to beat me to it.


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: "build"
 on:
   push:
      branches: [ master ]
-  pull_request: {}
+  # pull_request: {}
```
ggreif, line 5: Revert me!

.github/workflows/release.yml
```diff
@@ -15,15 +17,15 @@ jobs:
...
       id: get_version
-      run: echo ::set-output name=version::${GITHUB_REF/refs\/tags\//}
+      run: echo ::set-output name=version::${{ github.ref_name }}
 
     - name: Extract changelog
       id: read_changelog
+      if: github.ref != 'refs/heads/gabor/release-tinker'
       run: |
-        export VERSION='${{ steps.get_version.outputs.VERSION }}'
+        export VERSION='${{ steps.get_version.outputs.version }}'
```
ggreif, line 28: this is overkill now that we have `github.ref_name`

.github/workflows/release.yml
```diff
@@ -6,6 +6,8 @@ on:
   push:
     tags:
     - '*'
+    branches:
+    - 'gabor/release-tinker'
```
ggreif, line 10:
```suggestion
    - 'master'
```

.github/workflows/release.yml
```diff
@@ -51,24 +53,31 @@ jobs:
...
-  # build product to the nix cache, as we cannot build the darwin products on
+  # build product to the cachix cache, as we cannot build the darwin products on
   # linux
   release:
+    if: startsWith(github.ref, 'refs/tags/')
     runs-on: 'ubuntu-latest'
-    needs:
-    - changelog
-    - build
+    needs: [ changelog, build ]
```
ggreif, line 80: `build` should be enough?
```suggestion
    needs: build
```

ggreif: nope, it needs the version from the `changelog`. We could use `${{ github.ref_name }}` instead!

ggreif: Actually `needs.changelog.outputs.release_body` is a prerequisite too.

.github/workflows/release.yml
ggreif: but I'll leave it, it works...

nomeata (approved):
Sure

PR #2921, opened 2021-11-22
Fix documentation for `range` and `revRange`

### LINKED ISSUE #2920: Bug: Documentation of 'range'
` range(x : Nat, y : Int)` allows `Int` as the upper bound. There’s probably a reason that makes it easier to use. But anyway, in other parts of the documentation (here: https://smartcontracts.org/docs/language-guide/control-flow.html) it says the upper bound is `Nat`:
 As a constructor function, `range` has a function type:
 
 `(lower:Nat, upper:Nat) -> Iter<Nat>`

ggreif:
Corrects:
- type of the `upper` bound of `range` constructor function
- spelling of `revRange`
- types of the bounds of `revRange` constructor function
- return type of the `revRange` constructor function

fixes #2920

### DIFF
--- a/doc/modules/language-guide/pages/control-flow.adoc
+++ b/doc/modules/language-guide/pages/control-flow.adoc
@@ -178,7 +178,7 @@ As a constructor function, `range` has a function type:
 
 [source.no-repl, motoko]
 ....
-(lower:Nat, upper:Nat) -> Iter<Nat>
+(lower : Nat, upper : Int) -> Iter<Nat>
 ....
 
 Where `Iter<Nat>` is an iterator object type with a `next` method that produces optional elements, each of type `?Nat`:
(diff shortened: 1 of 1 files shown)


doc/modules/language-guide/pages/control-flow.adoc
```diff
@@ -196,17 +196,17 @@ The value `null` indicates that the iteration sequence has terminated.
 Until reaching `null`, each non-`null` value, of the form ``?``__n__ for some number _n_, contains the next successive element in the iteration sequence.
 
 [[intro-revrange]]
-== Using `revrange`
+== Using `revRange`
 
-Like `range`, the function `revrange` is a `class` that constructs iterators (each of type `Iter<Nat>`).
+Like `range`, the function `revRange` is a `class` that constructs iterators (each of type `Iter<Int>`).
```
crusso, line 201: Hmm, I wonder why it's not symmetric with Range in motoko-base but at least this documents what we've got, strange as it is. Thanks for fixing.

crusso (approved):

PR #2926, opened 2021-11-22
revise add_tag.sh to match dfx 0.8.4 onwards

crusso:
From 0.8.4 onwards, dfx ships the motoko binaries, but still includes base manually for some reason. This is my sad attempt to update the tagging script we use to update motoko and motoko-base base repos accordingly.

Is it actually ok to define a tag using a tag?
e.g.
```
git tag sdk-0.8.4 0.6.11
```
I had git complaining that I was redefining a tag I had used previously using a commit to one using a tag.


add-tags.sh
```diff
@@ -15,7 +15,8 @@ set -e
 
 git -C "$1" fetch --tags
 # 0.5.10 was the first where we have motoko-base via motoko
-for tag in $(git -C "$1" tag --contains 0.5.10 | sort -V)
+# 0.8.4 is the first where motoko is removed from sources.json (and we need to ddo something else)
```
ggreif, line 18:
```suggestion
# 0.8.4 is the first where motoko is removed from sources.json (and we need to do something else)
```

nomeata:
> Is it actually ok to define a tag using a tag?

Yes, that should be fine.

PR #2927, opened 2021-11-22
Use a translated bitmap to enable more efficient marking

### LINKED ISSUE #2892: Marking in mark-compact GC can be optimized (one-two instructions per live object)
Currently we convert an address to a bitmap index with `(address - heap_base) / word_size`: https://github.com/dfinity/motoko/blob/939a539539d234c601ced48da949004c46774580/rts/motoko-rts/src/gc/mark_compact.rs#L133

It's possible to eliminate two instructions (local load, subtraction) in this operation by doing `address / word_size` instead. This requires allocating slightly larger bitmaps for the heap below `heap_base`. 

One Wasm page needs 8,192 bytes (8 KiB) in the bitmap. `heap_base` is often 1 Wasm page (Rust/C stack) + a few more bytes for the literals (strings, bytes) in the canister.

Thanks @ulan for the idea.

ggreif:
Since the mark-and-compact GC is currently less performant than the copying one, it makes sense to tweak its performance. Here I take [the idea from @ulan](https://github.com/dfinity/motoko/issues/2892) and run with it. Basically, we now pass the object's absolute word number to `g/set_bit` and thus avoid some arithmetic and passing the `heap_base` along.

The marking was sped up by 4 instructions. This is compounded by the fact that now we don't pass the heap base to the marking routine `mark_object` any more:
- `mark_object`: 120 now vs. old 123 instructions
-  every callsite to `mark_object` 1 instruction less.

Also speeds up the iteration over the bitmap after marking (`BitmapIter::next` has now 92 instructions compared to formerly 110). In particular, the inner loop is now more compact.

Fixes #2892.

Benchmarks will be included in #2927 when available.

- [ ] show that due to 2-word object allocation the bitmap can be halved (not clear how)
- [ ] pass tag (and first word?) to object allocator
- [ ] when marking, use the skewed pointer and use pointer shifting, buffer padding tricks
- [ ] inner loop in `BitmapIter::next` [can be eliminated](https://github.com/dfinity/motoko/pull/2927#discussion_r760394082).

### DIFF
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -40,10 +40,10 @@
         "homepage": "",
         "owner": "dfinity",
         "repo": "ic-hs",
-        "rev": "e8f86abe7d27293f5f5140f844c2e61c96e0a486",
-        "sha256": "0cn8yf17jc9zl4v313b2cwb5jg1s4ka5avx5g8m96hrjhk346k23",
+        "rev": "07cad5267dcf2ac4d0c6d540130e56d080223814",
+        "sha256": "0y79sr39jq3rkx1c5czbr9xag3i8xr0bng9hj1187yi99v6nsad4",
         "type": "tarball",
-        "url": "https://github.com/dfinity/ic-hs/archive/e8f86abe7d27293f5f5140f844c2e61c96e0a486.tar.gz",
+        "url": "https://github.com/dfinity/ic-hs/archive/07cad5267dcf2ac4d0c6d540130e56d080223814.tar.gz",
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     },
     "libtommath": {
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -27,10 +27,10 @@
         "homepage": "",
         "owner": "dfinity",
         "repo": "ic",
-        "rev": "35dd8f93dec82662ed4df35664a9c0be6dbf203a",
-        "sha256": "0w6rz1s43i87abwf3fhi121ym2qxgirvbycsg5ykcz1019gdz803",
+        "rev": "6c2da54fe215bfbd534d3f932a179dc7d3faf877",
+        "sha256": "19ascaryrpcn8wv1y44l1zjmqarsvnabrjfivz7n834wb7inf8q4",
         "type": "tarball",
-        "url": "https://github.com/dfinity/ic/archive/35dd8f93dec82662ed4df35664a9c0be6dbf203a.tar.gz",
+        "url": "https://github.com/dfinity/ic/archive/6c2da54fe215bfbd534d3f932a179dc7d3faf877.tar.gz",
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     },
     "ic-hs": {
@@ -64,10 +64,10 @@
         "homepage": null,
         "owner": "dfinity",
         "repo": "motoko-base",
-        "rev": "4f46158ce7797a77a14a4543039c6c177aec2ded",
-        "sha256": "0wkj106h2v4kmlrh29qm0lr07nkfhw4gny00qynhdfgp1gfpv98n",
+        "rev": "c226be95be41d2bd69f2a3e671579d3ab24330f9",
+        "sha256": "0yl9zq9xvbvx2dhz0ylq17jw8ajcn150jz2l6r023s5d47k3j6qx",
         "type": "tarball",
-        "url": "https://github.com/dfinity/motoko-base/archive/4f46158ce7797a77a14a4543039c6c177aec2ded.tar.gz",
+        "url": "https://github.com/dfinity/motoko-base/archive/c226be95be41d2bd69f2a3e671579d3ab24330f9.tar.gz",
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     },
     "motoko-matchers": {
--- a/perf-delta.nix
+++ b/perf-delta.nix
@@ -36,7 +36,7 @@ let
         do
           # ignore all errors
           echo -n $file
-          if timeout 10s moc $file --compacting-gc -no-check-ir -ref-system-api -o $file.wasm 2>/dev/null
+          if timeout 10s moc $file --force-gc --compacting-gc -no-check-ir -ref-system-api -o $file.wasm 2>/dev/null
           then echo " failed (ignored)"
           else echo " ok"
           fi
--- a/perf-delta.nix
+++ b/perf-delta.nix
@@ -36,7 +36,7 @@ let
         do
           # ignore all errors
           echo -n $file
-          if timeout 10s moc $file -no-check-ir -ref-system-api -o $file.wasm 2>/dev/null
+          if timeout 10s moc $file --compacting-gc -no-check-ir -ref-system-api -o $file.wasm 2>/dev/null
           then echo " failed (ignored)"
           else echo " ok"
           fi
--- a/rts/motoko-rts-tests/src/bitmap.rs
+++ b/rts/motoko-rts-tests/src/bitmap.rs
@@ -2,7 +2,7 @@ use crate::memory::TestMemory;
 
 use motoko_rts::constants::WORD_SIZE;
 use motoko_rts::gc::mark_compact::bitmap::{
-    alloc_bitmap, get_bit, iter_bits, set_bit, BITMAP_ITER_END,
+    alloc_bitmap, free_bitmap, get_bit, iter_bits, set_bit, BITMAP_ITER_END,
 };
 use motoko_rts::memory::Memory;
 use motoko_rts::types::{Bytes, Words};
@@ -68,7 +68,7 @@ fn test_set_get<M: Memory>(mem: &mut M, mut bits: Vec<u16>) -> Result<(), String
     unsafe {
         alloc_bitmap(
             mem,
-            Bytes((u32::from(*bits.iter().max().unwrap()) + 1) * WORD_SIZE),
+            Bytes((u32::from(*bits.iter().max().unwrap()) + 1) * WORD_SIZE), 0 // TODO: test with static heap
         );
 
         for bit in &bits {
@@ -98,6 +98,8 @@ fn test_set_get<M: Memory>(mem: &mut M, mut bits: Vec<u16>) -> Result<(), String
 
             last_bit = Some(bit);
         }
+
+	free_bitmap()
     }
 
     Ok(())
@@ -111,7 +113,7 @@ fn test_bit_iter<M: Memory>(mem: &mut M, bits: HashSet<u16>) -> TestCaseResult {
     .to_bytes();
 
     unsafe {
-        alloc_bitmap(mem, heap_size);
+        alloc_bitmap(mem, heap_size, 0); // TODO: test with static heap
 
         for bit in bits.iter() {
             set_bit(u32::from(*bit));
@@ -154,6 +156,8 @@ fn test_bit_iter<M: Memory>(mem: &mut M, bits: HashSet<u16>) -> TestCaseResult {
                 .into(),
             ));
         }
+
+	free_bitmap()
     }
 
     Ok(())
--- a/rts/motoko-rts-tests/src/bitmap.rs
+++ b/rts/motoko-rts-tests/src/bitmap.rs
@@ -69,7 +69,7 @@ fn test_set_get<M: Memory>(mem: &mut M, mut bits: Vec<u16>) -> Result<(), String
         alloc_bitmap(
             mem,
             Bytes((u32::from(*bits.iter().max().unwrap()) + 1) * WORD_SIZE),
-            1, // TODO: test with static heap
+            0,
         );
 
         for bit in &bits {
@@ -114,7 +114,7 @@ fn test_bit_iter<M: Memory>(mem: &mut M, bits: HashSet<u16>) -> TestCaseResult {
     .to_bytes();
 
     unsafe {
-        alloc_bitmap(mem, heap_size, 1); // TODO: test with static heap
+        alloc_bitmap(mem, heap_size, 0);
 
         for bit in bits.iter() {
             set_bit(u32::from(*bit));
--- a/rts/motoko-rts-tests/src/bitmap.rs
+++ b/rts/motoko-rts-tests/src/bitmap.rs
@@ -68,7 +68,8 @@ fn test_set_get<M: Memory>(mem: &mut M, mut bits: Vec<u16>) -> Result<(), String
     unsafe {
         alloc_bitmap(
             mem,
-            Bytes((u32::from(*bits.iter().max().unwrap()) + 1) * WORD_SIZE), 0 // TODO: test with static heap
+            Bytes((u32::from(*bits.iter().max().unwrap()) + 1) * WORD_SIZE),
+            0, // TODO: test with static heap
         );
 
         for bit in &bits {
@@ -99,7 +100,7 @@ fn test_set_get<M: Memory>(mem: &mut M, mut bits: Vec<u16>) -> Result<(), String
             last_bit = Some(bit);
         }
 
-	free_bitmap()
+        free_bitmap()
     }
 
     Ok(())
@@ -157,7 +158,7 @@ fn test_bit_iter<M: Memory>(mem: &mut M, bits: HashSet<u16>) -> TestCaseResult {
             ));
         }
 
-	free_bitmap()
+        free_bitmap()
     }
 
     Ok(())
--- a/rts/motoko-rts-tests/src/bitmap.rs
+++ b/rts/motoko-rts-tests/src/bitmap.rs
@@ -69,7 +69,7 @@ fn test_set_get<M: Memory>(mem: &mut M, mut bits: Vec<u16>) -> Result<(), String
         alloc_bitmap(
             mem,
             Bytes((u32::from(*bits.iter().max().unwrap()) + 1) * WORD_SIZE),
-            0, // TODO: test with static heap
+            1, // TODO: test with static heap
         );
 
         for bit in &bits {
@@ -114,7 +114,7 @@ fn test_bit_iter<M: Memory>(mem: &mut M, bits: HashSet<u16>) -> TestCaseResult {
     .to_bytes();
 
     unsafe {
-        alloc_bitmap(mem, heap_size, 0); // TODO: test with static heap
+        alloc_bitmap(mem, heap_size, 1); // TODO: test with static heap
 
         for bit in bits.iter() {
             set_bit(u32::from(*bit));
--- a/rts/motoko-rts-tests/src/gc.rs
+++ b/rts/motoko-rts-tests/src/gc.rs
@@ -119,8 +119,8 @@ fn test_gc(
         heap.continuation_table_ptr_offset(),
     );
 
-    for _ in 0..0 {
-        gc.run(heap.clone()); /* FIXME: not alignment-preserving! */
+    for _ in 0..3 {
+        gc.run(heap.clone());
 
         let heap_base_offset = heap.heap_base_offset();
         let heap_ptr_offset = heap.heap_ptr_offset();
--- a/rts/motoko-rts-tests/src/gc.rs
+++ b/rts/motoko-rts-tests/src/gc.rs
@@ -119,8 +119,8 @@ fn test_gc(
         heap.continuation_table_ptr_offset(),
     );
 
-    for _ in 0..3 {
-        gc.run(heap.clone());
+    for _ in 0..0 {
+        gc.run(heap.clone()); /* FIXME: not alignment-preserving! */
 
         let heap_base_offset = heap.heap_base_offset();
         let heap_ptr_offset = heap.heap_ptr_offset();
--- a/rts/motoko-rts-tests/src/gc/heap.rs
+++ b/rts/motoko-rts-tests/src/gc/heap.rs
@@ -204,12 +204,14 @@ impl MotokoHeapInner {
             map.len(),
         );
 
+        // The Worst-case unalignment w.r.t. 32-byte alignment is 28 (assuming
+        // that we have general word alignment). So we over-allocate 28 bytes.
         let mut heap: Vec<u8> = vec![0; heap_size + 28];
 
         // MarkCompact assumes that the dynamic heap starts at a 32-byte multiple
         let realign = match gc {
             GC::Copying => 0,
-            GC::MarkCompact => (32 - &heap[static_heap_size_bytes] as *const u8 as usize % 32) % 32,
+            GC::MarkCompact => (32 - heap.as_ptr() as usize % 32) % 32,
         };
         assert_eq!(realign % 4, 0);
 
--- a/rts/motoko-rts-tests/src/gc/heap.rs
+++ b/rts/motoko-rts-tests/src/gc/heap.rs
@@ -212,7 +212,8 @@ impl MotokoHeapInner {
         // Each object will be 3 words per object + one word for each reference. Static heap will
         // have an array (header + length) with one element, one MutBox for each root. +1 for
         // continuation table pointer.
-        let static_heap_size_bytes = (2 + roots.len() + (roots.len() * 2) + 1) * WORD_SIZE;
+        let static_heap_size_bytes_unaligned = (2 + roots.len() + (roots.len() * 2) + 1) * WORD_SIZE;
+	let static_heap_size_bytes = (static_heap_size_bytes_unaligned + 31) / 32 * 32;
 
         let dynamic_heap_size_without_continuation_table_bytes = {
             let object_headers_words = map.len() * 3;
@@ -226,7 +227,6 @@ impl MotokoHeapInner {
                 .as_usize();
 
         let total_heap_size_bytes = static_heap_size_bytes + dynamic_heap_size_bytes;
-	let static_heap_size_aligned_bytes = (static_heap_size_bytes + 31) / 32 * 32;
 	
         let heap_size = heap_size_for_gc(
             gc,
@@ -235,24 +235,23 @@ impl MotokoHeapInner {
             map.len(),
         );
 
-        let mut heap: Vec<u8, Aligned32Bytes> = Vec::with_capacity_in(heap_size + (static_heap_size_aligned_bytes - static_heap_size_bytes), Aligned32Bytes {});
-        //let mut heap: Vec<u8> = vec![0; heap_size];
-	heap.resize(heap_size + (static_heap_size_aligned_bytes - static_heap_size_bytes), 0);
-	//assert_eq!(heap.len(), heap_size);
-	assert_eq!(&mut heap[static_heap_size_aligned_bytes] as *mut u8 as usize % 32, 0);
+        let mut heap: Vec<u8, Aligned32Bytes> = Vec::with_capacity_in(heap_size, Aligned32Bytes {});
+	heap.resize(heap_size, 0);
+	assert_eq!(heap.len(), heap_size);
+	assert_eq!(&mut heap[static_heap_size_bytes] as *mut u8 as usize % 32, 0);
 
         // Maps `ObjectIdx`s into their offsets in the heap
         let object_addrs: FxHashMap<ObjectIdx, usize> =
-            create_dynamic_heap(map, continuation_table, &mut heap[static_heap_size_aligned_bytes..]);
+            create_dynamic_heap(map, continuation_table, &mut heap[static_heap_size_bytes..]);
 
         // Closure table pointer is the last word in static heap
-        let continuation_table_ptr_offset = static_heap_size_aligned_bytes - WORD_SIZE;
+        let continuation_table_ptr_offset = static_heap_size_bytes - WORD_SIZE;
         create_static_heap(
             roots,
             &object_addrs,
             continuation_table_ptr_offset,
             static_heap_size_bytes + dynamic_heap_size_without_continuation_table_bytes,
-            &mut heap[..static_heap_size_aligned_bytes],
+            &mut heap[..static_heap_size_bytes],
         );
 
 	let boxed_slice = heap.into_boxed_slice();
@@ -261,7 +260,7 @@ impl MotokoHeapInner {
 	assert_eq!(heap_start % 32, 0);
         MotokoHeapInner {
             heap: boxed_slice,
-            heap_base_offset: static_heap_size_aligned_bytes,
+            heap_base_offset: static_heap_size_bytes,
             heap_ptr_offset: total_heap_size_bytes,
             static_root_array_offset: 0,
             continuation_table_ptr_offset: continuation_table_ptr_offset,
--- a/rts/motoko-rts-tests/src/gc/heap.rs
+++ b/rts/motoko-rts-tests/src/gc/heap.rs
@@ -6,10 +6,8 @@ use motoko_rts::gc::mark_compact::mark_stack::INIT_STACK_SIZE;
 use motoko_rts::memory::Memory;
 use motoko_rts::types::*;
 
-use std::alloc::*;
 use std::cell::{Ref, RefCell};
 use std::convert::TryFrom;
-use std::ptr::NonNull;
 use std::rc::Rc;
 
 use fxhash::{FxHashMap, FxHashSet};
@@ -92,7 +90,7 @@ impl MotokoHeap {
     }
 
     /// Get the heap as an array. Use `offset` values returned by the methods above to read.
-    pub(crate) fn heap(&self) -> Ref<Box<[u8], Aligned32Bytes>> {
+    pub fn heap(&self) -> Ref<Box<[u8]>> {
         Ref::map(self.inner.borrow(), |heap| &heap.heap)
     }
 
@@ -110,35 +108,10 @@ impl MotokoHeap {
     }
 }
 
-pub(crate) struct Aligned32Bytes {}
-
-unsafe impl Allocator for Aligned32Bytes {
-    fn allocate(&self, layout_unaligned: Layout) -> Result<NonNull<[u8]>, AllocError> {
-        //assert_eq!(layout_unaligned.size(), 232);
-        assert_eq!(layout_unaligned.align(), 1);
-        //let overhead = Layout::for_value(&[1; 0]);
-        let overhead = Layout::from_size_align(16, 8).unwrap();
-        assert_eq!(overhead.size(), 16);
-        assert_eq!(overhead.align(), 8);
-        //let layout = Layout::from_size_align(layout_unaligned.size() + 32 - overhead.size(), 32).unwrap();
-        let layout = layout_unaligned.align_to(32).unwrap();
-        unsafe {
-            let ptr = std::alloc::alloc(layout); //.add(overhead.size());
-            assert_eq!(ptr as usize % 32, 0);
-            let slice = std::slice::from_raw_parts_mut(ptr, layout_unaligned.size());
-            assert_eq!(slice.len(), layout_unaligned.size());
-            Ok(NonNull::new(slice).unwrap())
-        }
-    }
-    unsafe fn deallocate(&self, ptr: NonNull<u8>, layout: Layout) {
-        dealloc(ptr.as_ptr(), layout.align_to(32).unwrap())
-    }
-}
-
 struct MotokoHeapInner {
     /// The heap. This is a boxed slice instead of a vector as growing this wouldn't make sense
     /// (all pointers would have to be updated).
-    heap: Box<[u8], Aligned32Bytes>,
+    heap: Box<[u8]>,
 
     /// Where the dynamic heap starts
     heap_base_offset: usize,
@@ -209,9 +182,7 @@ impl MotokoHeapInner {
         // Each object will be 3 words per object + one word for each reference. Static heap will
         // have an array (header + length) with one element, one MutBox for each root. +1 for
         // continuation table pointer.
-        let static_heap_size_bytes_unaligned =
-            (2 + roots.len() + (roots.len() * 2) + 1) * WORD_SIZE;
-        let static_heap_size_bytes = (static_heap_size_bytes_unaligned + 31) / 32 * 32;
+        let static_heap_size_bytes = (2 + roots.len() + (roots.len() * 2) + 1) * WORD_SIZE;
 
         let dynamic_heap_size_without_continuation_table_bytes = {
             let object_headers_words = map.len() * 3;
@@ -233,17 +204,21 @@ impl MotokoHeapInner {
             map.len(),
         );
 
-        let mut heap: Vec<u8, Aligned32Bytes> = Vec::with_capacity_in(heap_size, Aligned32Bytes {});
-        heap.resize(heap_size, 0);
-        assert_eq!(heap.len(), heap_size);
-        assert_eq!(
-            &mut heap[static_heap_size_bytes] as *mut u8 as usize % 32,
-            0
-        );
+        let mut heap: Vec<u8> = vec![0; heap_size + 28];
+
+        // MarkCompact assumes that the dynamic heap starts at a 32-byte multiple
+        let misalign = match gc {
+            GC::Copying => 0,
+            GC::MarkCompact => (32 - &heap[static_heap_size_bytes] as *const u8 as usize % 32) % 32,
+        };
+        assert_eq!(misalign % 4, 0);
 
         // Maps `ObjectIdx`s into their offsets in the heap
-        let object_addrs: FxHashMap<ObjectIdx, usize> =
-            create_dynamic_heap(map, continuation_table, &mut heap[static_heap_size_bytes..]);
+        let object_addrs: FxHashMap<ObjectIdx, usize> = create_dynamic_heap(
+            map,
+            continuation_table,
+            &mut heap[static_heap_size_bytes + misalign..heap_size + misalign],
+        );
 
         // Closure table pointer is the last word in static heap
         let continuation_table_ptr_offset = static_heap_size_bytes - WORD_SIZE;
@@ -252,19 +227,15 @@ impl MotokoHeapInner {
             &object_addrs,
             continuation_table_ptr_offset,
             static_heap_size_bytes + dynamic_heap_size_without_continuation_table_bytes,
-            &mut heap[..static_heap_size_bytes],
+            &mut heap[misalign..static_heap_size_bytes + misalign],
         );
 
-        let boxed_slice = heap.into_boxed_slice();
-        let heap_start = boxed_slice.as_ptr() as usize;
-        //assert_eq!(heap_start, 0);
-        assert_eq!(heap_start % 32, 0);
         MotokoHeapInner {
-            heap: boxed_slice,
-            heap_base_offset: static_heap_size_bytes,
-            heap_ptr_offset: total_heap_size_bytes,
-            static_root_array_offset: 0,
-            continuation_table_ptr_offset: continuation_table_ptr_offset,
+            heap: heap.into_boxed_slice(),
+            heap_base_offset: static_heap_size_bytes + misalign,
+            heap_ptr_offset: total_heap_size_bytes + misalign,
+            static_root_array_offset: misalign,
+            continuation_table_ptr_offset: continuation_table_ptr_offset + misalign,
         }
     }
 
--- a/rts/motoko-rts-tests/src/gc/heap.rs
+++ b/rts/motoko-rts-tests/src/gc/heap.rs
@@ -211,7 +211,7 @@ impl MotokoHeapInner {
         // MarkCompact assumes that the dynamic heap starts at a 32-byte multiple
         let realign = match gc {
             GC::Copying => 0,
-            GC::MarkCompact => (32 - heap.as_ptr() as usize % 32) % 32,
+            GC::MarkCompact => (32 - (heap.as_ptr() as usize + static_heap_size_bytes) % 32) % 32,
         };
         assert_eq!(realign % 4, 0);
 
--- a/rts/motoko-rts-tests/src/gc/heap.rs
+++ b/rts/motoko-rts-tests/src/gc/heap.rs
@@ -207,17 +207,17 @@ impl MotokoHeapInner {
         let mut heap: Vec<u8> = vec![0; heap_size + 28];
 
         // MarkCompact assumes that the dynamic heap starts at a 32-byte multiple
-        let misalign = match gc {
+        let realign = match gc {
             GC::Copying => 0,
             GC::MarkCompact => (32 - &heap[static_heap_size_bytes] as *const u8 as usize % 32) % 32,
         };
-        assert_eq!(misalign % 4, 0);
+        assert_eq!(realign % 4, 0);
 
         // Maps `ObjectIdx`s into their offsets in the heap
         let object_addrs: FxHashMap<ObjectIdx, usize> = create_dynamic_heap(
             map,
             continuation_table,
-            &mut heap[static_heap_size_bytes + misalign..heap_size + misalign],
+            &mut heap[static_heap_size_bytes + realign..heap_size + realign],
         );
 
         // Closure table pointer is the last word in static heap
@@ -227,15 +227,15 @@ impl MotokoHeapInner {
             &object_addrs,
             continuation_table_ptr_offset,
             static_heap_size_bytes + dynamic_heap_size_without_continuation_table_bytes,
-            &mut heap[misalign..static_heap_size_bytes + misalign],
+            &mut heap[realign..static_heap_size_bytes + realign],
         );
 
         MotokoHeapInner {
             heap: heap.into_boxed_slice(),
-            heap_base_offset: static_heap_size_bytes + misalign,
-            heap_ptr_offset: total_heap_size_bytes + misalign,
-            static_root_array_offset: misalign,
-            continuation_table_ptr_offset: continuation_table_ptr_offset + misalign,
+            heap_base_offset: static_heap_size_bytes + realign,
+            heap_ptr_offset: total_heap_size_bytes + realign,
+            static_root_array_offset: realign,
+            continuation_table_ptr_offset: continuation_table_ptr_offset + realign,
         }
     }
 
--- a/rts/motoko-rts-tests/src/gc/heap.rs
+++ b/rts/motoko-rts-tests/src/gc/heap.rs
@@ -116,20 +116,17 @@ unsafe impl Allocator for Aligned32Bytes {
     fn allocate(&self, layout_unaligned: Layout) -> Result<NonNull<[u8]>, AllocError> {
         //assert_eq!(layout_unaligned.size(), 232);
         assert_eq!(layout_unaligned.align(), 1);
-	//let overhead = Layout::for_value(&[1; 0]);
-	let overhead = Layout::from_size_align(16, 8).unwrap();
+        //let overhead = Layout::for_value(&[1; 0]);
+        let overhead = Layout::from_size_align(16, 8).unwrap();
         assert_eq!(overhead.size(), 16);
         assert_eq!(overhead.align(), 8);
         //let layout = Layout::from_size_align(layout_unaligned.size() + 32 - overhead.size(), 32).unwrap();
-	let layout = layout_unaligned.align_to(32).unwrap();
+        let layout = layout_unaligned.align_to(32).unwrap();
         unsafe {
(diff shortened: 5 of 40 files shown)


rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -22,19 +23,29 @@ pub unsafe fn alloc_bitmap<M: Memory>(mem: &mut M, heap_size: Bytes<u32>) {
...
+    BITMAP_BASE = 0
+}
+
+/// Move the bitmap base such that accessing `idx`
+/// will touch the beginning of the bitmap.
+pub unsafe fn translate_bitmap(idx: u32) {
+    let (byte_idx, bit_idx) = (idx / 8, idx % 8);
+    assert_eq!(bit_idx, 0);
+    BITMAP_BASE = byte_idx as usize;
+    BITMAP_PTR = BITMAP_PTR.sub(BITMAP_BASE);
```
osa1, line 35: I'm confused about this code. `BITMAP_PTR` after allocating the bitmap points to a blob payload. Now you subtract a few bytes from it. How do you make sure this won't overwrite the blob header, or dynamic objects allocated before the blob?

osa1:
As far as I understand, the idea here is to move bitmap address *back* so that when we compute the bit byte index of an object with `object_addr / WORD_SIZE` and add that to bitmap address, we will have a byte in the bitmap.

This way we don't want to allocate bits for the static heap, as suggested in #2892. It's faster, but doesn't come with space overhead. I like it.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
osa1: I think I understand this now, see my top-level comment.

ggreif: I have integrated this into the `alloc_bitmap` code.

ggreif:
Currently I am fighting with the heap's alignment when doing the tests. Here is the principal problem:
When `heap_base` is not divisible by 32 then adding `heap_base % WORD_SIZE` (effectively) to the mark location (`set_mark`'s argument) can cause two effects
  1. marking can happen beyond the allocated mark blob
  2. mark bits appear in shifted positions when iterated over in 64-bit chunks.
The latter will cause the sweep phase to reclaim the wrong objects (or reach into the middle of them, looking for heap tags). (The former could be compensated easily by bumping the mark blob's size by one.)

Aligning the `heap_base` on the IC is pretty much trivial, but in the test environment it is trickier, since a `Vec[u8]`'s payload can end up being allocated at any address that is divisible by 8.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -4,37 +4,47 @@ use crate::types::{size_of, Blob, Bytes, Obj};
...
+static mut BITMAP_COMPENSATION: usize = 0;
+static mut BITMAP_SIZE: u32 = 0;
 
-pub unsafe fn alloc_bitmap<M: Memory>(mem: &mut M, heap_size: Bytes<u32>) {
+pub unsafe fn alloc_bitmap<M: Memory>(mem: &mut M, heap_size: Bytes<u32>, heap_prefix_words: u32) {
     // We will have at most this many objects in the heap, each requiring a bit
     let n_bits = heap_size.to_words().as_u32();
+    // Additional bits needed to account for
+    // unaligned heap start address
+    let prefix_bits = heap_prefix_words % 8;
```
ggreif, line 15: assert that `heap_prefix_words % 8 == 0`, i.e. the heap base is 32-byte aligned.

ggreif: then this can go away for good

crusso:
Looking good! Thanks for grasping the nettle!

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
ggreif: Gone in 0aaaeeb.

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -24,13 +24,18 @@ pub(crate) static mut LAST_HP: u32 = 0;
 
 // Provided by generated code
 extern "C" {
-    pub(crate) fn get_heap_base() -> u32;
+    fn get_heap_base() -> u32;
```
ggreif, line 27: @osa1 which macro generates this function? Maybe we ca generate `get_aligned_heap_base` as well?

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -4,37 +4,45 @@ use crate::types::{size_of, Blob, Bytes, Obj};
 
 /// Current bitmap
 static mut BITMAP_PTR: *mut u8 = core::ptr::null_mut();
+static mut BITMAP_COMPENSATION: usize = 0;
+static mut BITMAP_SIZE: u32 = 0;
 
-pub unsafe fn alloc_bitmap<M: Memory>(mem: &mut M, heap_size: Bytes<u32>) {
+pub unsafe fn alloc_bitmap<M: Memory>(mem: &mut M, heap_size: Bytes<u32>, heap_prefix_words: u32) {
+    assert_eq!(heap_prefix_words % 8, 0);
```
ggreif, line 11: this can be `debug_` since we `assert!` the harder fact in rts/motoko-rts/src/gc/mark_compact.rs:75 already

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -4,37 +4,45 @@ use crate::types::{size_of, Blob, Bytes, Obj};
 
 /// Current bitmap
 static mut BITMAP_PTR: *mut u8 = core::ptr::null_mut();
+static mut BITMAP_COMPENSATION: usize = 0;
+static mut BITMAP_SIZE: u32 = 0;
```
osa1, line 8: You will be the maintainer so feel free to ignore, but documenting these variables would be good. For example, that the `BITMAP_SIZE` is only used for sanity checking.

I find it a bit difficult to explain what `BITMAP_COMPENSATION` is, maybe a visualization (in ASCII) would help.

I think it's also worth documenting why you choose to implement this with these set of variables. Here's an alternative implementation:

- BITMAP_PTR: pointer to the blob payload. Used in `bitmap_iter`. Saves us two cycles when skipping a bitmap word in `BitmapIter::next`.
- BLOB_PTR: pointer to the blob header. Used in sanity checks to get the blob size.
- BITMAP_PTR_SHIFTED: (not sure about the name) the base address used in `set_bit` and `get_bit`.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -4,37 +4,45 @@ use crate::types::{size_of, Blob, Bytes, Obj};
...
 
 pub unsafe fn get_bit(idx: u32) -> bool {
-    let byte_idx = idx / 8;
+    let (byte_idx, bit_idx) = (idx / 8, idx % 8);
+    debug_assert!(byte_idx as usize >= BITMAP_COMPENSATION);
+    debug_assert!(BITMAP_COMPENSATION + BITMAP_SIZE as usize > byte_idx as usize);
     let byte = *BITMAP_PTR.add(byte_idx as usize);
-    let bit_idx = idx % 8;
-    (byte >> bit_idx) & 0b1 == 0b1
+    (byte >> bit_idx) & 0b1 == 0b1 // TODO IDEA: != 0 (faster, also below!)
```
osa1, line 38: This is a trivial change, why leave a TODO instead of just doing it?

Why do you think `(byte >> bit_idx) & 0b1 != 0` is faster that `(byte >> bit_idx) & 0b1 == 0b1`? I'd expect the compiler to do this transformation if it's obviously faster.

What does "also below!" mean?

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -100,7 +109,9 @@ impl BitmapIter {
 
             // Inner loop iterates bits in the current word
             while self.current_word != 0 {
-                if self.current_word & 0b1 == 0b1 {
+                if self.current_word & 0b1 == 0b1
+                /* TODO */
```
osa1, line 113: I guess this is what "also below!" above points to? If `!= 0` is really faster (why?) let's fix this as well, the change for it will be smaller than adding the TODO comments.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -117,8 +128,10 @@ impl BitmapIter {
             if self.current_word_idx == self.size {
                 return BITMAP_ITER_END;
             }
-            self.current_word =
-                unsafe { *(BITMAP_PTR as *const u64).add(self.current_word_idx as usize) };
+            self.current_word = unsafe {
+                *(BITMAP_PTR.add(BITMAP_COMPENSATION) as *const u64)
+                    .add(self.current_word_idx as usize)
```
osa1, line 133: `add(BITMAP_COMPENSATION)` part can be avoided if we cache the blob payload location, as before. See my comment at the beginning of this file.

rts/motoko-rts/src/memory/ic.rs
osa1: This function is generated by the codegen.

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -24,13 +24,18 @@ pub(crate) static mut LAST_HP: u32 = 0;
...
     pub(crate) fn get_static_roots() -> Value;
 }
 
+pub(crate) unsafe fn get_aligned_heap_base() -> u32 {
+    // align to 32 bytes
+    (get_heap_base() + 31) / 32 * 32
+}
```
osa1, line 34: Again, I won't be maintaining this code so feel free to ignore, but it will be very confusing to readers why we need to align the heap base, so documentation here would be good.

Yesterday we talked about adding one more byte to the bitmap when heap base is not aligned, did that not work out? Why do you choose to align heap base instead of adding one more byte to the bitmap when it's not already aligned?

osa1 (commented):
Thanks @ggreif, I really like this idea.

Since the whole point here is performance, it would be good to see benchmark results.

One thing that is worse in this version than the previous is iteration of bitmap words. It can be improved as explained in my inline comments. I think the performance may still be better than master as it is, but bitmap can be a few MiBs (1 MiB of bitmap addresses 32 MiB of heap). For a heap near full, it will be near 134M, which requires ~16k 64-bit word iterations. So I think improving cycles there may worth it.

rts/motoko-rts/src/memory/ic.rs
ggreif: Documentation is coming.

It did not work out, see https://github.com/dfinity/motoko/pull/2927#issuecomment-977006471.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
ggreif: I believe LLVM would easily optimise this. f1b9fb7, because it is shorter.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
ggreif: ditto

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
ggreif: This PR is still draft and evolving. I used the current schema, because it was the first I got working. I am aware of the pointer pair representation, and will probably switch. Nice ASCII graphics coming.

osa1: Ah, sorry, didn't realize this is draft.

rts/motoko-rts/src/memory/ic.rs
ggreif: PTAL: ea18318

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -2,39 +2,90 @@ use crate::mem_utils::memzero;
...
+with the Motoko static heap.
+
+## The alignment caveat
+
+For this scheme to work, it is essential that the start of the DH is an address that
+is divisible by 32 (implying that `heap_prefix_words % 8 == 0`). Otherwise the
+`/%8`-operation on the DH's starting address will not yield the least significant bit
+in the BM, and thus the sweep operation will be off.
+
+## Example calculation:
```
ggreif, line 36:
```suggestion
## Example calculation
```

perf-delta.nix
```diff
@@ -36,7 +36,7 @@ let
         do
           # ignore all errors
           echo -n $file
-          if timeout 10s moc $file -no-check-ir -ref-system-api -o $file.wasm 2>/dev/null
+          if timeout 10s moc $file --compacting-gc -no-check-ir -ref-system-api -o $file.wasm 2>/dev/null
```
ggreif, line 39: this will be removed

test/profile-report.sh
```diff
@@ -51,7 +51,7 @@ __END__
 for file in perf/*.mo; do
   base="$(basename "$file" .mo)"
   echo "Profiling $base..."
-  moc -g "$file" -o "_profile_build/$base.wasm"
+  moc --compacting-gc -g "$file" -o "_profile_build/$base.wasm"
```
ggreif, line 54: will be reverted

nix/sources.json
```diff
@@ -25,12 +25,12 @@
         "builtin": false,
         "description": "Internet Computer blockchain source: the client/replica software run by nodes",
         "homepage": "",
-        "owner": "dfinity",
+        "owner": "ggreif",
```
ggreif, line 28: this is just an experiment for @crusso, hopefully unblocking him. This will be reverted before merge. Please ignore.

perf-delta.nix
ggreif:
```suggestion
          if timeout 10s moc $file --force-gc --compacting-gc -no-check-ir -ref-system-api -o $file.wasm 2>/dev/null
```

rts/motoko-rts-tests/src/gc.rs
```diff
@@ -119,8 +119,8 @@ fn test_gc(
         heap.continuation_table_ptr_offset(),
     );
 
-    for _ in 0..3 {
-        gc.run(heap.clone());
+    for _ in 0..0 {
+        gc.run(heap.clone()); /* FIXME: not alignment-preserving! */
```
ggreif, line 123: `clone()` needs to be aware of the dynamic heap's 32-byte alignment constraint when doing compacting GC.

rts/motoko-rts-tests/src/gc/heap.rs
```diff
@@ -217,15 +227,15 @@ impl MotokoHeapInner {
...
+            &mut heap[realign..static_heap_size_bytes + realign],
         );
 
         MotokoHeapInner {
             heap: heap.into_boxed_slice(),
-            heap_base_offset: static_heap_size_bytes,
-            heap_ptr_offset: total_heap_size_bytes,
-            static_root_array_offset: 0,
-            continuation_table_ptr_offset: continuation_table_ptr_offset,
+            heap_base_offset: static_heap_size_bytes + realign,
```
ggreif, line 235: maybe pass `realign` separately, to help cloning?

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -2,55 +2,113 @@ use crate::mem_utils::memzero;
 use crate::memory::{alloc_blob, Memory};
 use crate::types::{size_of, Blob, Bytes, Obj};
 
+/* How the Wasm-heap maps to the bitmap
+
+  +---- Rts stack ----+---- Motoko statics ----+---- Dynamic heap --~~--+ Heap end
+                  (prefix words)               | BM |   <- bitmap lives here
```
ggreif, line 8: I think BM is the _last_ object allocated in the dynamic heap!

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -90,36 +151,36 @@ pub const BITMAP_ITER_END: u32 = 1024 * 1024 * 1024;
 impl BitmapIter {
     /// Returns the next bit, or `BITMAP_ITER_END` if there are no more bits set.
     pub fn next(&mut self) -> u32 {
-        debug_assert!(self.current_word_idx <= self.size);
+        debug_assert!(self.current_bit_idx <= self.size);
 
         // Outer loop iterates 64-bit words
         loop {
-            if self.current_word == 0 && self.current_word_idx == self.size {
+            if self.current_word == 0 && self.current_bit_idx == self.size {
```
ggreif, line 158: I have the feeling that this can still be tweaked, maybe just `self.current_bit_idx >= self.size`?

rts/motoko-rts-tests/src/bitmap.rs
```diff
@@ -98,6 +99,8 @@ fn test_set_get<M: Memory>(mem: &mut M, mut bits: Vec<u16>) -> Result<(), String
 
             last_bit = Some(bit);
         }
+
+        free_bitmap()
```
ggreif, line 103: done by `mark_compact`

rts/motoko-rts-tests/src/bitmap.rs
```diff
@@ -154,6 +157,8 @@ fn test_bit_iter<M: Memory>(mem: &mut M, bits: HashSet<u16>) -> TestCaseResult {
                 .into(),
             ));
         }
+
+        free_bitmap()
```
ggreif, line 161: dito

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -90,36 +151,36 @@ pub const BITMAP_ITER_END: u32 = 1024 * 1024 * 1024;
...
-            self.current_word_idx += 1;
-            if self.current_word_idx == self.size {
+            self.current_bit_idx += self.leading_zeros;
+            if self.current_bit_idx == self.size {
                 return BITMAP_ITER_END;
             }
             self.current_word =
-                unsafe { *(BITMAP_PTR as *const u64).add(self.current_word_idx as usize) };
-            self.bits_left = 64;
+                unsafe { *(BITMAP_PTR as *const u64).add(self.current_bit_idx as usize / 64) };
```
ggreif, line 182:
```suggestion
                unsafe { *(BITMAP_FORBIDDEN_PTR.add(self.current_bit_idx as usize / 8) as *const u64) };
```

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -90,36 +151,37 @@ pub const BITMAP_ITER_END: u32 = 1024 * 1024 * 1024;
...
                     self.current_word >>= shift_amt;
-                    self.bits_left -= shift_amt;
+                    self.current_bit_idx += shift_amt;
                 }
             }
 
             // Move on to next word
-            self.current_word_idx += 1;
-            if self.current_word_idx == self.size {
+            self.current_bit_idx += self.leading_zeros;
```
ggreif, line 177: explain why this always works, though it looks mildly clever :-)

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
ggreif: Okay, we have renamed variables, PTAL.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
ggreif: The iteration code is bigly reworked, should be moderately more efficient too.

rts/motoko-rts/src/debug.rs
```diff
@@ -31,21 +31,21 @@ pub unsafe fn dump_heap(
     print_heap(heap_base, hp);
 }
 
-pub(crate) unsafe fn print_continuation_table(closure_tbl_loc: *mut Value) {
+pub(crate) unsafe fn print_continuation_table(continuation_tbl_loc: *mut Value) {
```
ggreif, line 34: only spelling changes in this file

rts/motoko-rts/src/gc/copying.rs
```diff
@@ -23,7 +23,7 @@ unsafe fn copying_gc<M: Memory>(mem: &mut M) {
 
     copying_gc_internal(
         mem,
-        ic::get_heap_base(),
+        ic::get_aligned_heap_base(),
```
ggreif, line 26: no need to align here

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -90,36 +151,37 @@ pub const BITMAP_ITER_END: u32 = 1024 * 1024 * 1024;
...
+        debug_assert!(self.current_bit_idx <= self.size);
 
         // Outer loop iterates 64-bit words
         loop {
-            if self.current_word == 0 && self.current_word_idx == self.size {
+            if self.current_word == 0 && self.current_bit_idx == self.size {
                 return BITMAP_ITER_END;
             }
 
             // Inner loop iterates bits in the current word
```
ggreif, line 162: BitmapIterator-next
The inner loop is a few instructions shorter, and has only `store` traffic.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -2,55 +2,115 @@ use crate::mem_utils::memzero;
...
+## The alignment caveat
+
+For this scheme to work, it is essential that the start of the DH is an address that
+is divisible by 32 (implying that `heap_prefix_words % 8 == 0`). Otherwise the
+`/%8`-operation on the DH's starting address will not yield the least significant bit
+in the BM, and thus the sweep operation will be off.
+
+## Example calculation
+
+Assume the DH is at 0x80000. The BM thus could be at 0x80004.
```
ggreif, line 40: redo with more realistic numbers

rts/motoko-rts/src/gc/copying.rs
ggreif: c3b3b185e

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -2,55 +2,125 @@ use crate::mem_utils::memzero;
 use crate::memory::{alloc_blob, Memory};
 use crate::types::{size_of, Blob, Bytes, Obj};
 
+/* How the Wasm-heap maps to the bitmap
+
+  +---- Rts stack ----+---- Motoko statics ----+---- Dynamic heap --~~--+ Heap limit
```
ggreif, line 7:
```suggestion
  +---- RTS stack ----+---- Motoko statics ----+---- Dynamic heap --~~--+ Heap limit
```

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -2,55 +2,125 @@ use crate::mem_utils::memzero;
...
+
+       +---- BITMAP_FORBIDDEN_PTR         +---- BITMAP_PTR
+       v                                  v
+       ,          (forbidden)             ,...................... bitmap ..........~~....,
+       |   heap_prefix_words / 8 bytes    |               heap_size / 32 bytes           |
+       |   get_bitmap_forbidden_size()    |                   BITMAP_SIZE                |
+       ^                                  ^                                              ^
+      /           ^^^^^^^^^^^              \                                             |
+     /            corresponds               \                 ^^^^^^^^^^^               /
+    /                                        \                corresponds              /
```
ggreif, line 31: say that the forbidden are overlaps with the dynamic heap

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -85,41 +158,42 @@ pub unsafe fn iter_bits() -> BitmapIter {
...
-            self.current_word_idx += 1;
-            if self.current_word_idx == self.size {
+            self.current_bit_idx += self.leading_zeros;
+            if self.current_bit_idx == self.size {
                 return BITMAP_ITER_END;
             }
-            self.current_word =
-                unsafe { *(BITMAP_PTR as *const u64).add(self.current_word_idx as usize) };
-            self.bits_left = 64;
+            self.current_word = unsafe {
```
crusso, line 193: Would it be slightly faster to compute the field values in locals and refer to them via those rather than reading from memory, e.g. for computing self.current_word from the local (hopefully still in a register) rather than self.current_bit_idx? Same for computing self.leading_zeroes  from self.current_word.

ggreif: Below is the function in full, I marked the area you are referring to with comments
``` wasm
  (func $motoko_rts::gc::mark_compact::bitmap::BitmapIter::next::hd19ec51ccf587428 (type 8) (param i32) (result i32)
    (local i32 i32 i32 i64 i64)
    block  ;; label = @1
      local.get 0
      i32.load offset=12
      local.tee 1
      local.get 0
      i32.load offset=8
      local.tee 2
      i32.eq
      br_if 0 (;@1;)
      i32.const 67696
      i32.const 15036
      i32.add
      i32.load
      local.set 3
      local.get 0
      i64.load
      local.set 4
      loop  ;; label = @2
        block  ;; label = @3
          local.get 4
          i64.eqz
          br_if 0 (;@3;)
          block  ;; label = @4
            loop  ;; label = @5
              local.get 4
              i32.wrap_i64
              i32.const 1
              i32.and
              br_if 1 (;@4;)
              local.get 0
              local.get 4
              local.get 4
              i64.ctz
              local.tee 5
              i64.shr_u
              local.tee 4
              i64.store
              local.get 0
              local.get 1
              local.get 5
              i32.wrap_i64
              i32.add
              local.tee 1
              i32.store offset=12
              local.get 4
              i64.eqz
              br_if 2 (;@3;)
              br 0 (;@5;)
            end
          end
          local.get 0
          local.get 1
          i32.const 1
          i32.add
          i32.store offset=12
          local.get 0
          local.get 4
          i64.const 1
          i64.shr_u
          i64.store
          local.get 1
          return
        end
        local.get 0
        local.get 1
        local.get 0
        i32.load offset=16
        i32.add
        local.tee 1
        i32.store offset=12
        local.get 1
        local.get 2
        i32.eq
        br_if 1 (;@1;)
        local.get 0 ;; self
        local.get 3 ;; BITMAP_FORBIDDEN_PTR
        local.get 1 ;; self.current_bit_idx
        i32.const 3
        i32.shr_u ;; / 8
        i32.add ;; .add
        i64.load ;; *
        local.tee 4
        i64.store ;; .current_word =
        local.get 0 ;; self
        local.get 4 ;; self.current_word
        i64.clz;; .leading_zeros()
        i64.store32 offset=16 ;; .leading_zeros =
        br 0 (;@2;)
      end
    end
    i32.const -1)
```
As you can see, the compiler already places all fields in locals:
- `self`: 0
- `current_bit_idx`: 1
- `size`: 2
- `current_word`: 4
- `BITMAP_FORBIDDEN_PTR`: 3

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
ggreif: Graphic updated.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
ggreif: this is better now

rts/motoko-rts-tests/src/gc.rs
ggreif: Seems to work o my Mac, so I reverted this for now.

crusso:
The PR looks good to me, but I think I'll need to defer to someone who actually speaks rust.

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -38,7 +38,7 @@ unsafe fn compacting_gc<M: Memory>(mem: &mut M) {
 
     compacting_gc_internal(
         mem,
-        ic::get_heap_base(),
+        ic::get_aligned_heap_base(),
```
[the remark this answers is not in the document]
ggreif: I now pass `aligned: bool` to init. I pass `true` for `MarkCompact` GC. This avoids any heap size regression for copying GC. See: 52da3088a.

rts/motoko-rts-tests/src/gc/heap.rs
ggreif: Looks like not an issue any more.

rts/motoko-rts-tests/src/gc.rs
ggreif: Looks like this was a red herring.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
ggreif: Yeah, LLVM does a good job here.

perf-delta.nix
```diff
@@ -36,7 +36,7 @@ let
         do
           # ignore all errors
           echo -n $file
-          if timeout 10s moc $file -no-check-ir -ref-system-api -o $file.wasm 2>/dev/null
+          if timeout 10s moc $file --force-gc --compacting-gc -no-check-ir -ref-system-api -o $file.wasm 2>/dev/null
```
ggreif, line 39: @osa1 should we leave this `--compacting-gc` in? I guess it will be the default, so we should test it with priority...

ggreif:
> The PR looks good to me, but I think I'll need to defer to someone who actually speaks rust.

@ulan can you have a look?

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -2,55 +2,126 @@ use crate::mem_utils::memzero;
...
+
+Debug assertions guard the forbidden bytes from access, as this area potentially
+physically overlaps with the Motoko dynamic heap.
+
+## The alignment caveat
+
+For this scheme to work, it is essential that the start of the DH is an address that
+is divisible by 32 (implying that `heap_prefix_words % 8 == 0`). Otherwise the
+`/%8`-operation on the DH's starting address will not yield the least significant bit
+in the BM, and thus the sweep operation will be off.
```
osa1, line 46:
```suggestion
is divisible by 32 (`heap_prefix_words % 8 == 0`). Otherwise the `address / 8`
and `address % 8` operations on the DH's starting address will not yield the
least significant bit in the BM.
```

Changes:

- "implying that" part gone, to keep it short
- Removed unfamiliar `%/` syntax
- "sweep operation will be off" we don't do sweep. This is important as sweep thouches unreachable objects, but we do compaction and touch reachable objects.
- Formatted with vim's `gq`

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -2,55 +2,126 @@ use crate::mem_utils::memzero;
...
+    BITMAP_SIZE = 0;
 }
 
 pub unsafe fn get_bit(idx: u32) -> bool {
-    let byte_idx = idx / 8;
-    let byte = *BITMAP_PTR.add(byte_idx as usize);
-    let bit_idx = idx % 8;
-    (byte >> bit_idx) & 0b1 == 0b1
+    let (byte_idx, bit_idx) = (idx / 8, idx % 8);
+    #[cfg(debug_assertions)]
```
osa1, line 96:
```suggestion
```

`debug_assert` already has this guard, it only runs in debug builds.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -2,55 +2,126 @@ use crate::mem_utils::memzero;
...
+    debug_assert!(byte_idx as usize >= get_bitmap_forbidden_size());
+    #[cfg(debug_assertions)]
```
osa1, line 98:
```suggestion
```

Same as above.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -2,55 +2,126 @@ use crate::mem_utils::memzero;
...
+
+For this scheme to work, it is essential that the start of the DH is an address that
+is divisible by 32 (implying that `heap_prefix_words % 8 == 0`). Otherwise the
+`/%8`-operation on the DH's starting address will not yield the least significant bit
+in the BM, and thus the sweep operation will be off.
+
+## Example calculation
+
+Assume the DH is at 0x80000. Assume heap limit being at 0xB0000. Then the BM thus
+could be placed at 0xB0004. Since the heap_prefix_words is 0x20000,
```
osa1, line 51: Shouldn't the bitmap start at `0xB0008`? It has two words in the header, not one.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -2,55 +2,126 @@ use crate::mem_utils::memzero;
...
+BITMAP_FORBIDDEN_PTR = 0xB0004 - 0x20000 / 8 = 0xAC004.
+
+Now let's mark the address 0x80548 in the DH. Its absolute word number is 0x20152.
+The `(0x20152 / 8, 0x20152 % 8)`-rule gives a bit position 2 with byte offset 0x402A,
+thus we mark bit 2 in byte 0xAC004 + 0x402A = 0xB002E, which is physically in the BM.
```
osa1, line 56:
```suggestion
thus we mark bit 2 in byte 0xAC004 + 0x402A = 0xB002E, which is in the BM.
```

Same as above. Maybe it's me being obsessive, but when I read a qualifier like "physically" here, I think about what else it could be. In this case, I think about what does being "virtually in the BM" mean (opposite of "virtually in the BM"). I think this distinction doesn't make sense, so the qualifier is confusing to me.

As usual, feel free to ignore.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -2,55 +2,126 @@ use crate::mem_utils::memzero;
...
+    let byte = *BITMAP_FORBIDDEN_PTR.add(byte_idx as usize);
+    (byte >> bit_idx) & 0b1 != 0
 }
 
 pub unsafe fn set_bit(idx: u32) {
-    let byte_idx = idx / 8;
-    let byte = *BITMAP_PTR.add(byte_idx as usize);
-    let bit_idx = idx % 8;
+    let (byte_idx, bit_idx) = (idx / 8, idx % 8);
+    #[cfg(debug_assertions)]
```
osa1, line 106:
```suggestion
```

Same as above.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -2,55 +2,126 @@ use crate::mem_utils::memzero;
...
+    debug_assert!(byte_idx as usize >= get_bitmap_forbidden_size());
+    #[cfg(debug_assertions)]
```
osa1, line 108:
```suggestion
```

Same as above.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -85,41 +159,42 @@ pub unsafe fn iter_bits() -> BitmapIter {
...
+            if self.current_bit_idx == self.size {
                 return BITMAP_ITER_END;
             }
-            self.current_word =
-                unsafe { *(BITMAP_PTR as *const u64).add(self.current_word_idx as usize) };
-            self.bits_left = 64;
+            self.current_word = unsafe {
+                *(BITMAP_FORBIDDEN_PTR.add(self.current_bit_idx as usize / 8) as *const u64)
+            };
+            self.leading_zeros = self.current_word.leading_zeros();
```
osa1, line 197: I'm a bit confused about why we need a state (`self.leading_zeros`) here. When we move on to the next word, we need to update the bit index with `self.current_bit_idx += self.current_word.leading_zeros()`. After this the leading zeros won't be needed anymore so the state can be removed. Am I missing anything?

It also seems to me like the code above is buggy, because it updates `current_bit_idx` *before* moving on to the next word.

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -28,9 +28,18 @@ extern "C" {
     pub(crate) fn get_static_roots() -> Value;
 }
 
+pub(crate) unsafe fn get_aligned_heap_base() -> u32 {
+    // align to 32 bytes
+    (get_heap_base() + 31) / 32 * 32
```
osa1, line 33:
```suggestion
    ((get_heap_base() + 31) / 32) * 32
```

We also have a "rounding up division" function in Rust core now, but it has a [strange name](https://doc.rust-lang.org/beta/std/primitive.u32.html#method.unstable_div_ceil) for [complicated reasons](https://github.com/rust-lang/rust/issues/88971).

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -28,9 +28,18 @@ extern "C" {
...
+
 #[no_mangle]
-unsafe extern "C" fn init() {
-    HP = get_heap_base() as u32;
+unsafe extern "C" fn init(align: bool) {
+    HP = if align {
+        get_aligned_heap_base()
+    } else {
+        get_heap_base()
+    };
```
osa1, line 42: I wonder if we should align always and keep this code (and the code below) simple. It will waste at most 7 bytes when copying GC is used. Not sure if it's worth making things more complicated. This would simplify tests too.

rts/motoko-rts-tests/src/gc/heap.rs
```diff
@@ -217,15 +227,15 @@ impl MotokoHeapInner {
...
+            heap_ptr_offset: total_heap_size_bytes + realign,
+            static_root_array_offset: realign,
+            continuation_table_ptr_offset: continuation_table_ptr_offset + realign,
```
osa1, line 238: This shifts the entire static heap, but that's not what we do in the compiler. In the compiler the padding is added at the end of the static heap. Should they be in sync?

rts/motoko-rts-tests/src/gc/heap.rs
```diff
@@ -204,11 +204,21 @@ impl MotokoHeapInner {
             map.len(),
         );
 
-        let mut heap: Vec<u8> = vec![0; heap_size];
+        let mut heap: Vec<u8> = vec![0; heap_size + 28];
```
osa1, line 207: Document the magic number please.

rts/motoko-rts-tests/src/gc/heap.rs
```diff
@@ -204,11 +204,21 @@ impl MotokoHeapInner {
...
+
+        // MarkCompact assumes that the dynamic heap starts at a 32-byte multiple
+        let realign = match gc {
+            GC::Copying => 0,
+            GC::MarkCompact => (32 - &heap[static_heap_size_bytes] as *const u8 as usize % 32) % 32,
```
osa1, line 212: Something like this should be simpler:

```suggestion
            GC::MarkCompact => (32 - heap.as_ptr() as usize % 32) % 32,
```

osa1 (commented):
Thanks @ggreif. Added inline comments.

rts/motoko-rts/src/memory/ic.rs
ggreif: It wastes at most 28 bytes (if we assume that the HP is word-aligned).

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
ggreif: I had compile errors when I removed this.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
ggreif: `self.leading_zeros` is needed because inner loop doesn't handle the leading zeros of the `self.current_word`. So when the topmost bit is _not_ set, then `self.current_bit_idx` won't be incremented to the next 64-bit boundary _by the loop_ and we have to do it separately. Also if the inner loop never gets entered (`self.current_word == 0`) then we have to jump to the next word boundary by adding 64 bits. In this case `self.current_bit_idx += self.leading_zeros` will do the right thing, because `self.leading_zeros == 64`, and we can make progress.

`self.leading_zeros` needs to be in the `self` state (and recorded before the inner loop), because the shifting of `self.current_word` would increase the leading zeros in it.

I don't think we have a bug here, hopefully the above makes it clear why.

rts/motoko-rts/src/memory/ic.rs
ggreif: That name is indeed ugly. Will remember this, and might change when available regularly.

rts/motoko-rts-tests/src/gc/heap.rs
ggreif: Good observation, but not really relevant for the inner workings, I think. I'll try to only realign the dynamic portion. Hopefully that won't break relative references from the dynamic heap into the static one (they are never followed by GC, but they might get accessed otherwise?).

perf-delta.nix
ggreif: I'll leave this in and revisit it as soon as #2930 gets ripe.

rts/motoko-rts/src/memory/ic.rs
ggreif: Somebody relying on being able to use the utmost bit might be tripped over by even such a minor regression. I wouldn't be very happy about receiving such a bug report :-)

osa1: > Somebody relying on being able to use the utmost bit might be tripped over by even such a minor regression

I don't understand what you mean here. What is "utmost bit"? How can aligning heap base always cause a bug?

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
osa1: You need to remove `debug_assertions` guard from `get_bitmap_forbidden_size` too.

Expansion of `debug_assert` is something like

```rust
if cfg!(debug_assertions) {
    ... assertion ...
}
```

The condition is compiled to `true` or `false`, but the body is still type checked, so `get_bitmap_forbidden_size` should be defined.

rts/motoko-rts/src/memory/ic.rs
ggreif: Oh, if you increase the initial heap pointer then you reduce the available dynamic heap space. You _can_ (but very seldom will) run out of heap space and get a GC earlier than before. Not that it is a sound strategy to rely on it, but some may actually.

osa1: I doubt anyone will notice that the usable heap size is 27 bytes smaller in some cases, but OK.

rts/motoko-rts-tests/src/gc/heap.rs
ggreif: Oops that should be
``` diff
            GC::MarkCompact => (32 - heap.as_ptr().add(static_heap_size_bytes) as usize % 32) % 32,
```

ggreif: And that seems to work nicely.

rts/motoko-rts-tests/src/gc/heap.rs
ggreif: Done in cf98977.

rts/motoko-rts-tests/src/gc/heap.rs
ggreif: Hmm, I tried to do this, but the necessary gap needs to go into the `heap_size` computation and which is needed for the `vec!` allocation, which then determines the necessary gap size. The circularity introduced would make this very fragile and much harder to understand.

So I won't do this now.

That said, I could map
``` rust
        |gap| heap_size_for_gc(
            gc,
            static_heap_size_bytes + gap,
            dynamic_heap_size_bytes,
            map.len(),
        )
```
over `[0, 4, 8, .. 28]` and get the heap size from there. But this speculative approach is also not worth it, IMO.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
ggreif: I imagine. I cleaned this up in f37b9af30098496ae6edd8c541a9f8c76d59c6b4.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
ggreif: Done.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
ggreif: I think this is done.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
osa1: I found a bug while working on simplifying this method. When `self.leading_zeros` is 0, this line

```rust
            self.current_bit_idx += self.leading_zeros;
```

does not increment current bit idx, and cause the iterator to yield the same bit index twice.

I'm surprised that this isn't causing heap corruption. I realized this case in the RTS tests, but the tests somehow pass..

osa1: We should be able to easily extend `test_bit_iter` test function to check for visited bits, and fail if the iterator yields the same bit multiple times. @ggreif do you want to do this? I can do this if you prefer that. We can then merge it, merge master to this branch, and make sure we won't yield same bit index multiple times in this branch.

ggreif: I wonder how that could happen, let's analyse! When entering `BitmapIter::next` we have 2 cases:
- `self.leading_zeros` is 0 (implying `current_word` is `0x[8-F].......`)
- `self.leading_zeros` is `> 0` (implying `current_word` is `0x[0-7].......`)
In the first case we enter the inner loop and find a bit:
- either the LSBit -> increments the `self.current_bit_idx` by one
- a string of trailing zeros -> increments the `self.current_bit_idx` by at least one
In the other case we don't enter the loop and `self.current_bit_idx` gets incremented by nonzero `self.leading_zeros`.

I still can't see a bug.

osa1: OK, that makes sense. Thanks for the explanation.

osa1 (approved):
LGTM, but would be good to see benchmarks. Marking should be more efficient now, but I can't tell if the new bitmap iterator will be faster or slower, and what will be the effect of it to the overall GC.

It would be helpful to see the changes in the PR description. For example, number of instructions in marking or one iteration of bitmap iterator, or even better, actual benchmark results.

What happens to TODOs in the PR description? If they're for another PR should we remove them? They will be in the commit message when this is merged.

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -85,41 +154,42 @@ pub unsafe fn iter_bits() -> BitmapIter {
...
+                    let bit_idx = self.current_bit_idx;
                     self.current_word >>= 1;
-                    self.bits_left -= 1;
+                    self.current_bit_idx += 1;
                     return bit_idx;
                 } else {
                     let shift_amt = self.current_word.trailing_zeros();
                     self.current_word >>= shift_amt;
-                    self.bits_left -= shift_amt;
+                    self.current_bit_idx += shift_amt;
```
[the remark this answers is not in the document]
ggreif: Here is a quick implementation of the concept for posterity.
```
diff --git a/rts/motoko-rts/src/gc/mark_compact/bitmap.rs b/rts/motoko-rts/src/gc/mark_compact/bitmap.rs
index 2607a4baa..234f74de8 100644
--- a/rts/motoko-rts/src/gc/mark_compact/bitmap.rs
+++ b/rts/motoko-rts/src/gc/mark_compact/bitmap.rs
@@ -167,8 +167,8 @@ impl BitmapIter {
 
         // Outer loop iterates 64-bit words
         loop {
-            // Inner loop iterates bits in the current word
-            while self.current_word != 0 {
+            // Inner conditional examines the least significant bit(s) in the current word
+            if self.current_word != 0 {
                 if self.current_word & 0b1 != 0 {
                     let bit_idx = self.current_bit_idx;
                     self.current_word >>= 1;
@@ -177,7 +177,10 @@ impl BitmapIter {
                 } else {
                     let shift_amt = self.current_word.trailing_zeros();
                     self.current_word >>= shift_amt;
-                    self.current_bit_idx += shift_amt;
+                    self.current_word >>= 1;
+                    let bit_idx = self.current_bit_idx + shift_amt;
+                    self.current_bit_idx = bit_idx + 1;
+                    return bit_idx;
                 }
             }
 ```
I won't commit this, as the double shift looks ugly. Maybe I can come up with a more elegant way.

**Comment by ggreif** [comment]
> LGTM, but would be good to see benchmarks. Marking should be more efficient now, but I can't tell if the new bitmap iterator will be faster or slower, and what will be the effect of it to the overall GC.

Yes, there can be surprises, although unlikely. Benchmarks will be added to the MR.

> It would be helpful to see the changes in the PR description. For example, number of instructions in marking or one iteration of bitmap iterator, or even better, actual benchmark results.

Added instruction balances.

> What happens to TODOs in the PR description? If they're for another PR should we remove them? They will be in the commit message when this is merged.

I nuked the TODOs as these were meant to track my progress, and are now irrelevant.

**Comment by ggreif** [comment]
As promised, here come the performance counters for the change
- baseline is release 0.6.16
- compared to with a7868cd1f1378f0b2c7f30c2626b46ac765e9651 reversed

```
[nix-shell:~/motoko]$ tail compacting-baseline_perf.csv compacting-a7868cd1f13_perf.csv  
==> compacting_perf.csv <==
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,29548225,222,54,16
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,29572634,222,54,16
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,29318871,224,37,16
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,29305196,223,36,16
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,29326984,223,36,16
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,29186706,223,26,16
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,29178490,225,23,16
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,29159337,224,22,16
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,29173610,224,22,16
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,32425028,224,200,16

==> compacting-a7868cd1f13_perf.csv <==
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,30624396,222,54,16
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,30649781,222,54,16
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,30397508,224,37,16
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,30384817,223,36,16
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,30407620,223,36,16
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,30268366,223,26,16
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,30261620,225,23,16
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,30243489,224,22,16
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,30258762,224,22,16
rwlgt-iiaaa-aaaaa-aaaaa-cai,canister_update createProfile,33511297,224,200,16
```
----------------
A few quick percentages:
``` shell
[nix-shell:~/motoko]$ ghc -e '33511297/32425028*100'
103.35009425435192

[nix-shell:~/motoko]$ ghc -e '30258762/29173610*100'
103.71963565702016

[nix-shell:~/motoko]$ ghc -e '30624396/29548225*100'
103.6420834077174
```
We would regress more that 3% (on performing pure GC) with this PR reverted!

PR #2930, opened 2021-11-24
Minor tweaks to CI and documentation

ggreif:


ggreif:
@nomeata @osa1 I guess we should use `--force-gc` for the benchmarks, otherwise the GC scheduler will avoid unnecessary work. Eventually we should benchmark 3 ways
- default gc and scheduler:
- `--force-gc --copying-gc`
- `--force-gc --compacting-gc`

ggreif:
@nomeata where are the cycle counts generated? Do we call `drun` to do it for us? THX!

osa1:
> Eventually we should benchmark 4 ways:

I agree, the ways you show makes sense.

Is it difficult to implement this in this PR?

> where are the cycle counts generated? Do we call drun to do it for us? THX!

That's done by drun, yes. Relevant files/lines:

- https://github.com/dfinity/motoko/blob/235c994a8e8f5cce72024732d9e40b1e517bd753/test/drun.json5
- https://github.com/dfinity/motoko/blob/235c994a8e8f5cce72024732d9e40b1e517bd753/test/drun-wrapper.sh#L53
- https://github.com/dfinity/motoko/blob/235c994a8e8f5cce72024732d9e40b1e517bd753/test/run.sh#L350-L353

ggreif:
> Is it difficult to implement this in this PR?
> 
I prefer not to do it here, as consolidating the benchmarks from different runs can be tricky. I'll have to discuss that with @nomeata.

crusso (approved):

ggreif:
> And GitHub actions doesn't? The matrix is being calculated in `default.nix`, is it, and thus used by both?

No idea, I'll have a closer look ext week!

PR #2931, opened 2021-11-25
feat: StableMemory Library, 64 bit

crusso:
Internal updgrade to 64 bit StableMemory API, leaving external 32-bit API unchanged.

@rossberg could you take a quick look before I move on to the surface primitives and libraries?


src/codegen/compile.ml
```diff
@@ -5610,33 +5641,34 @@ module Stabilization = struct
            fs
           in
           StableMem.get_mem_size env ^^
-          E.then_trap_with env "StableMem.mem_size non-zero" ^^
+          G.i (Test (Wasm.Values.I64 I64Op.Eqz)) ^^
```
crusso, line 5644: if0?

src/codegen/compile.ml
```diff
@@ -3838,25 +3853,20 @@ module StableMem = struct
...
           get_size ^^ G.i (Convert (Wasm.Values.I64 I64Op.ExtendUI32)) ^^
           G.i (Binary (Wasm.Values.I64 I64Op.Add)) ^^
-          compile_sub64_const 1L ^^
-          compile_const_64 (Int64.of_int page_size_bits) ^^
-          G.i (Binary (Wasm.Values.I64 I64Op.ShrU)) ^^
-          G.i (Convert (Wasm.Values.I32 I64Op.WrapI64)) ^^
+          set_sum ^^
+          get_sum ^^
+          get_offset ^^
+          G.i (Compare (Wasm.Values.I64 I32Op.LtU)) ^^
```
rossberg, line 3865:
```suggestion
          G.i (Compare (Wasm.Values.I64 I64Op.LtU)) ^^
```

src/codegen/compile.ml
```diff
@@ -3838,25 +3853,20 @@ module StableMem = struct
     match E.mode env with
     | Flags.ICMode | Flags.RefMode ->
       Func.share_code2 env "__stablemem_guard_range"
-        (("offset", I32Type), ("size", I32Type)) []
+        (("offset", I64Type), ("size", I32Type)) []
```
rossberg, line 3856: Should size not be i64 as well?

src/codegen/compile.ml
```diff
@@ -5610,33 +5637,34 @@ module Stabilization = struct
            fs
           in
           StableMem.get_mem_size env ^^
-          E.then_trap_with env "StableMem.mem_size non-zero" ^^
+          G.i (Test (Wasm.Values.I64 I64Op.Eqz)) ^^
+          E.else_trap_with env "StableMem.mem_size non-zero" ^^
```
rossberg, line 5641: eqz is performing a not, followed by a conditional with only an else branch. So this looks like unnecessary double negation.

src/codegen/compile.ml
```diff
@@ -5672,33 +5701,33 @@ module Stabilization = struct
...
-              compile_add_const 4l ^^
+              compile_add64_const 4L ^^
               set_offset
             end
             begin
               (* assert mem_size == 0 *)
               StableMem.get_mem_size env ^^
-              E.then_trap_with env "unexpected, non-zero stable memory size" ^^
+              G.i (Test (Wasm.Values.I64 I64Op.Eqz)) ^^
+              E.else_trap_with env "unexpected, non-zero stable memory size" ^^
```
rossberg, line 5711: Same here.

rossberg (approved):
LGTM, only nits.

src/codegen/compile.ml
crusso: Hmm, I wonder why the tests didn't fail - this should probably produce illegal code

crusso: Oh, no its fine - I only ever use this with i32 sizes

src/codegen/compile.ml
crusso: Actually, I don't see how to improve this. `if` consumes an i32 and I've got an i64 on the stack. Is there a better way?

rossberg: Ah, you're right, I overlooked that this is an i64. Scratch my comment then.

PR #2932, opened 2021-11-25
Temporarily vendor `drun` from a fork of `dfinity/ic`

ggreif:
This will cause conflicts with `master`, which we'll iron out  later...

This is not intended to hit `master`. We had to fork `dfinity/ic` so that we could fix a bug in `drun` that snuck in why no CI was watching it. A fresh, working `drun` is essential for authoring the 64-bit stable memory functionality in Motoko.


osa1:
What is the motivation here?

crusso:
> What is the motivation here?

To get drun both building and working again with recent drops of the ic, so we can access new functionality.

osa1:
> To get drun both building and working again with recent drops of the ic, so we can access new functionality.

OK. Would be good to copy this to PR description so that it will become the commit message and next person looking at these lines can look at the commit and figure out why we're using a fork and not the upstream version.

nomeata:
Instead of using a fork, I'd suggest to use nix's ability to apply patches to paths, and store the path here. This way it's clear what has changed, what the base revision is, and (for small changes) the auto updater will continue to work.

Or maybe it's faster to get the patch upstream :-)

Anyways, once CI is green we can do that

ggreif:
> 
> Or maybe it's faster to get the patch upstream :-)
> 

You must be joking! ;-) This is the turkey-apocalipse weekend!

When we have the fork working, I can try to distill a diff and apply it in `nix`. But this is probably easier (albeit tedious) for now.

crusso:
Ok, I'm beginning to the think the test failures might be due to scheduler changes in the ic. I wonder if they every merge Akhi's optimization for messages self, so they could be scheduled in the same block.

crusso:
I've change a couple of tests that were affecting by the scheduler changes. Let's see if it lights up green now.

ggreif:
Feel free to merge this one int your branch and mark that as _DO-NOT-MERGE_ until we get the official `ic` repo.

ggreif:
The `drun` fix has been ported upstream: https://gitlab.com/dfinity-lab/core/ic/-/merge_requests/1415, hopefully it will be mirrored to https://github.com/dfinity/ic

nomeata:
The `ic` code dump suddenly also contains all the nix infrastructure. So we _could_ go back to the old way of pulling in `drun`, based on the nix derivation in that repo. Hard to predict which one will work better in the future. But pulling in the nix derivation means no manual updating of cargo sha256 hashes, so maybe that's a good reason?

nomeata:
Nevermind, the `common` repo isn't open source yet. 🤦

ggreif:
This can be closed now, as https://github.com/dfinity/motoko/pull/2779 just landed

ggreif:
I have deleted https://github.com/ggreif/ic now, as it is not needed any more.

Issue #2937, opened 2021-11-26
Trap in callback makes canister unupgradeable


nomeata:
Claudio mentioned on the forums that Motoko canister will not upgrade when it believes it has outstanding callbacks. This is well-intended and prevents one kind of footgun (callbacks from old versions reaching the upgraded canister), but it introduces another huge problem: If a canister _ever_ traps in a callback handler, the canister's internal count ouf outstanding calls will never reach zero, and the canister will never be upgradeable, _even if stopped_. This change was https://github.com/dfinity/motoko/pull/2677

To mitigate this, we should get around to implement the `on_cleanup` callback, and in that callback remove the entry from the table (but don't do GC, or anything else that could possible cause a trap). I am not sure if this is a bullet-proof mitigation; maybe we can make that code so predicatbly small that we can guarantee its running within the cycle limit for the callback? But really, it just has to do a bit of index calculation and write one word to memory, maybe that's safe enough?

origyndev:
Hmmm....ok...trying to understand this....@nometa has seen our code....I think we are doing a ton of this.  Right? When we our vesting canister calls our vesting canister and asks it to do some work, if the governance canister returns and then we trap in the vesting canister then the count would be off?  Or is it when the governance canister traps and doesn't return to the vesting canister?  I'm trying to visualize how this happens. If our canisters are currently unupgradeable that would be good to know as we need to add some hydration code to complement our backup code.

nomeata:
> When we our vesting canister calls our vesting canister and asks it to do some work, if the governance canister returns and then we trap in the vesting canister then the count would be off?

Yes.

I think in your case, you don't _expect_ the canisters to trap, so with some luck, you won't have a problem yet, just a risk, and likely hopefully will be able to upgrade just fine until a fix lands. And if not, you can still reinstall (instead of upgrade) and use your backups, just like with any other possible disaster.

It is a more serious problem if you call untrustworthy canisters and don't have a an off-chain backup.

nomeata:
False alarm; the necessary fixes to use `on_cleanup` have been implemented (#2663). For some reason Github didn't show anything when I used it's search with `cleanup`, but maybe I misspelt something. Sorry for the confusion!

I'll still create a PR with the test case I was writing in preparation for fixing this.

PR #2939, opened 2021-11-26
Test case for upgrade-after-trap-in-callback

nomeata:
sorry for shouting wolf in #2937. At least we get more tests out of
this.


crusso:
Don't worry about calling wolf - turns out no-one was listening. But yes, @ggreif implemented the on_cleanup logic a while ago so we should be good.

Any idea why CI is failing on ubuntu?

crusso (approved):
Thanks for the test

PR #2942, opened 2021-11-27
Use an invalid table index as the callback for one-way functions

### LINKED ISSUE #2938: Upgradeable compilation of one-shot calls
Canisters that are upgradeable with zero downtime are probably quite important for some uses cases, and while it seems not possible for code that uses `await`, it _might_ just be possible for canisters that only use one-way messages.

Luckily, Motoko was created when we thought we would have them on the system level, and so is Candid, so we know which shared method calls are meant to be fire-and-forget.

For these, instead of registering do-nothing callbacks, I suggest the following trick: We pass `-1` as the table index for the callbacks. When the callback comes back, this will trap, but that is fine - the trap is just as good as the do-nothing callback. But the advantage is that it will do that reliably even after an upgrade (assuming no canister accepted by the system will ever have a table of size 2^32 ). So a developer who writes their canisters in the plain actor model, with only one-way calls, will be able to safely upgrade the canister, even with the current System API (i.e. no need to wait for [changes to the system api](https://forum.dfinity.org/t/prioritize-safe-instantaneous-canister-upgrades/9039/2)).

Developers still need to manually figure out if stopping is needed or not, but at least it's one way foward, await from the current, quite unsatisfying "stop-it-first" regime.

nomeata:
this fixes #2938, and allows developers to write canisters that are
zero-downtime upgradeable, as long as they stick to making one-way calls
onlys (i.e. plain actor model)


crusso:
Very nicce but we should double check this doesn't somehow interfere with the work that @ggreif did with cleanup of the closure table after a trapping callback. Probably fine, but best to double check.

nomeata:
Double checking is appreciated. I believe that the callback table wasn't used at all for one-ways, and no cleanup callback needed or used. (This also means that the “is it safe to upgrade” check was incomplete, as it would ignore outstanding callbacks to one-way calls.)

crusso (commented):
Looks ok to me, but I don't quite see why the error message has changed in the test? @ggreif have you taken a look at this?

nomeata:
Because with this change, when the callback comes back, the system now complains about the callback is to be invalid. that's expected (and drun is even silent about this)

ggreif:
Sorry for the delay, I am a bit behind. I'll have a look soon.

nomeata:
Ping, @ggreif?

ggreif:
This got buried again :-(, sorry!

I have double-checked, #2663 is compatible with this change, as it will try to access the continuation table in the same way as the other callbacks.

ggreif (approved):
LGTM!

PR #2969, opened 2021-12-08
add stable var checking and gcflag to moc.js

chenyan-dfinity:
Tested in https://github.com/dfinity/motoko-playground/pull/91

### DIFF
--- a/src/js/common.ml
+++ b/src/js/common.ml
@@ -50,6 +50,13 @@ let js_run list source =
   let list = List.map Js.to_string list in
   ignore (Pipeline.run_stdin_from_file list (Js.to_string source))
 
+let js_candid source =
+  js_result (Pipeline.generate_idl [Js.to_string source])
+    (fun prog ->
+      let code = Idllib.Arrange_idl.string_of_prog prog in
+      Js.some (Js.string code)
+    )
+
 let js_stable_compatible pre post =
   js_result (Pipeline.stable_compatible (Js.to_string pre) (Js.to_string post)) (fun _ -> Js.null)
 
--- a/src/js/common.ml
+++ b/src/js/common.ml
@@ -50,28 +50,35 @@ let js_run list source =
   let list = List.map Js.to_string list in
   ignore (Pipeline.run_stdin_from_file list (Js.to_string source))
 
-let js_candid source =
-  js_result (Pipeline.generate_idl [Js.to_string source])
-    (fun prog ->
-      let code = Idllib.Arrange_idl.string_of_prog prog in
-      Js.some (Js.string code)
-    )
+let js_stable_compatible pre post =
+  js_result (Pipeline.stable_compatible (Js.to_string pre) (Js.to_string post)) (fun _ -> Js.null)
 
 let js_compile_wasm mode source =
+  let source = Js.to_string source in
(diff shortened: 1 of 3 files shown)


chenyan-dfinity:
@crusso Is there an easy way to get the stable signature from type checking or desugar? In the playground, it would be nice if we can get the signature before producing the Wasm.

crusso:
There is the --stable-type types option, but that writes to <foo>.most. Or you can read it from the custom section.

src/js/common.ml
```diff
@@ -57,21 +57,35 @@ let js_candid source =
...
+  js_result (Pipeline.compile_files mode true [source])
+    (fun (idl_prog, m) ->
+      let open CustomModule in
+      let sig_ = match m.motoko.stable_types with
+        | Some (_, txt) -> Js.some (Js.string txt)
+        | _ -> Js.null in
+      let candid = Idllib.Arrange_idl.string_of_prog idl_prog in
       let (_, wasm) = CustomModuleEncode.encode m in
       let constructor = Js.Unsafe.global##._Uint8Array in
       let code = constructor##from
```
crusso, line 80: I know this old code, but is the code returned as a stream, not just a JS bytearray? Doesn't that make reading each byte really slow?

src/js/common.ml
```diff
@@ -109,3 +123,10 @@ let set_actor_aliases entries =
   let aliases = Flags.actor_aliases in
   aliases := Flags.M.of_seq (Array.to_seq entries)  
 
+let gc_flags option =
+  match Js.to_string option with
+  | "force" -> Flags.force_gc := true
+  | "scheduling" -> Flags.force_gc := false
+  | "copying" -> Flags.gc_strategy := Mo_config.Flags.Copying
+  | "marking" -> Flags.gc_strategy := Mo_config.Flags.MarkCompact
```
crusso, line 131: Would it make sense to use the same names as moc compiler options, just for sanity?

src/js/common.ml
```diff
@@ -57,21 +57,35 @@ let js_candid source =
...
     | "wasi" -> Flags.WASIMode
     | "ic" -> Flags.ICMode
     | _ -> raise (Invalid_argument "js_compile_with: Unexpected mode")
   in
-  js_result (Pipeline.compile_files mode true [Js.to_string source])
-    (fun (_, m) ->
+  js_result (Pipeline.compile_files mode true [source])
+    (fun (idl_prog, m) ->
+      let open CustomModule in
+      let sig_ = match m.motoko.stable_types with
```
crusso, line 74: You can also get the arg-less service type this way, should you need it.

src/js/common.ml
chenyan-dfinity: It's returning a bytearray/Uint8Array. The constructor is slow, but it's kind of the only way to produce a bytearray to JS. Other "faster" method breaks the content if it's not utf8 compatible.

src/js/common.ml
chenyan-dfinity: I think this is close enough. There is no sanity in JS...I just spent the last two days figuring out that `await setX(42)` in React doesn't guarantee x is 42 after the await...

PR #2971, opened 2021-12-08
feat: Implement heartbeat support

### LINKED ISSUE #2728: Support the heartbeat
@roman-kashitsyn [leaked on the forum](https://forum.dfinity.org/t/ic-cron-lets-schedule-some-tasks-bois/6506/6?u=nomeata) that the `canister_heartbeat` entry point is now supported by application subnets. The Interface spec doesn’t list it yet, but once it’s there we should probably support it from Motoko.

Unless we take the stance (from the discussion on the feature in the `ic-ref` at https://github.com/dfinity-lab/ic-ref/issues/245, which I can’t read any more) that `canister_heartbeat` is a low-level feature meant to be used mainly by canisters providing cron services to other canisters, and that Motoko canisters should not use this (inefficient, inflexible) interface, but should rather use dedicated cron-as-a-service-canisters.

But the pragmatic thing is probably to expose the `canister_heartbeat` entry point.

ninegua:
Closes #2728 

- [x] Basic support for heartbeat as a system function
- [x] Proper async type and type checking of the heartbeat function
- [x] Test cases

### DIFF
--- a/.github/workflows/niv-updater-trial.yml
+++ b/.github/workflows/niv-updater-trial.yml
@@ -28,7 +28,7 @@ jobs:
       - name: niv-updater-action
         uses: knl/niv-updater-action@60f23607814cf4f2e80a1e32ee74f8323897d09e
         with:
-          whitelist: 'nixpkgs,musl-wasi'
+          whitelist: 'nixpkgs,musl-wasi,ic'
           labels: |
             autoclose
           keep_updating: true
--- a/.github/workflows/niv-updater.yml
+++ b/.github/workflows/niv-updater.yml
@@ -23,7 +23,7 @@ jobs:
         uses: knl/niv-updater-action@60f23607814cf4f2e80a1e32ee74f8323897d09e
         with:
           # might be too noisy
-          blacklist: 'nixpkgs,ic-ref,musl-wasi'
+          blacklist: 'nixpkgs,ic-ref,musl-wasi,ic'
           labels: |
             automerge-squash
           keep_updating: true
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,11 @@
 # Motoko compiler changelog
 
+## 0.6.19 (2022-01-05)
+
+* motoko-base
+
+  * Fixed a bug in the `RBTree.size()` method.
+
 ## 0.6.18 (2021-12-20)
 
 * moc
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # Motoko compiler changelog
 
+* motoko
+
+  * Implement support for `heartbeat` system methods (thanks to Paul Liu) (#2971)
+
 ## 0.6.19 (2022-01-05)
 
 * motoko-base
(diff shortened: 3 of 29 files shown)


ninegua:
At the moment it seems to work. But I am trying to change heartbeat to return `()` instead of `async ()`, then I got this error:

```
+heartbeat.mo:12.16-12.31: type error [M0037], misplaced async expression; try enclosing in an async function
```

Not exactly sure how to proceed. So I'm asking help from @crusso

crusso:
I pulled your branch locally and played around with changing the return type to (), but couldn't quite make it work.

To do that, you'd really want the heartbeat to be a shared oneway function so your example would wind up looking something like this:

```
import Prim "mo:⛔";

actor {
  var count = 0;
  public shared func inc() : async () {
    count := count + 1;
    Prim.debugPrint("count = " # debug_show(count));
  };

/*
  system func heartbeat() : async () {
    if (count < 10) {
      Prim.debugPrint("heartbeat");
      ignore inc();
    }
  };
 */

  system shared func heartbeat() : () {
     ignore async {
      if (count < 10) {
        Prim.debugPrint("heartbeat");
        ignore inc();
      }
    }
  };

};
```

Looking at that now, I'm not sure that's preferable to just declaring the `async ()` type and using a local function.

crusso:
I've made the IR typeable, but I'm still a little worried that this code will have the hearbeat message generating a response, even though the spec says a heartbeat message cannot reply or reject. I guess that might generate a silent trap at the end of the heartbeat message or maybe the ic-ref and the replica aren't actually enforcing this.

crusso:
Note to self: the way to fix this is probably to introduce a new CallHeartbeat primitive that passed dummy success/fail continuations (instead of ones that reply/reject).

crusso:
Another small issue with this PR is that I believe it always exports the hearbeat method, even if it has trivial work. That's probably not what we want since every Motoko canister will no opt in to hearbeat notification, stressing the heart as it were.

crusso:
Apart from the issue of always registering a heartbeat, I think I've convinced myself that this code is actually ok. I was worried it was attempting to reply or reject, but in fact, it just calls the helper method and discards it, without every registering a reply/reject continations which is probably ok - the body will run asynchronously but never respond.

crusso:
I've redone a bit of the lowering/codegen and added some documentation. @ninegua PTAL if you care, otherwise I'll defer to @chenyan-dfinity or @ggreif for review.

(This implementation will do an extra async call for each heartbeat, but that's really a problem with our implementation of async functions and will be fixed by fixing that. I didn't have the stomach to circumvent this by adding a hack. An alternative would be to also support ()-returning heartbeats that are executed inline, but cannot send, but that seems gross too).

crusso:
@ninegua is seem to have confused git with my inept merging of branches - sorry about that.

crusso:
PR moved to #3039 to improve perf of CI

ggreif:
As #3039 has landed, I think this can be closed now.

ggreif:
Please reopen if something is still missing.

PR #2976, opened 2021-12-10
`nix`: don't patch `nixpkgs` if we don't have patches

nomeata:
This should speed up `nix-shell` etc. when we don't apply any patches,
without removing the code to do so.

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -10,12 +10,22 @@ let
     system = builtins.currentSystem;
   };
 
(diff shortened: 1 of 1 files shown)


ggreif (approved):
Cool, now I understand! THX!

ggreif:
@Mergifyio refresh

ggreif:
This seems very effective! Previously we had:
Screenshot 2021-12-10 at 19 55 55
And now:
Screenshot 2021-12-10 at 19 56 41

PR #2981, opened 2021-12-10
Don't run `Update nix hashes` action on certain pushes

ggreif:


nomeata (approved):

.github/workflows/update-hash.yml
```diff
@@ -2,6 +2,10 @@ name: Update nix hashes
 
 on:
   push:
+    branches-ignore:
+      - master
+    tags-ignore:
+      - '*'
```
ggreif, line 8:
```suggestion
      - '**'
```
Matches all characters, also `/`. @nomeata Probably we'll never have slashes in tag names, but this may be more correct.

PR #2999, opened 2021-12-15
RTS: trap in stubbed libc functions

### LINKED ISSUE #2248: RTS stub functions should probably panic
In RTS we have 3 libc functions that we define on our own instead of using from musl, in [stubs.c][1]:

- `int wctomb(char *s, wchar_t wc)`: "convert a wide character to a multibyte sequence", our version just terminates the input string
- `int fputs(const char *s, FILE *stream)`: writes the string to the stream and returns nonnegative on success. Our version just returns 0 without doing anything.
- `char *strerror(int errnum)`: returns string describing the argument. Our version returns empty string.

(we also define `abort` but I think the implementation of it is correct)

I think the idea with these definitions is: we know these are not going to be used and we don't want to build and link large portions of musl for these, which makes sense. However I also think we should just panic in these functions instead of silently doing a wrong thing. That way we'll know if our assumption that these shouldn't be used doesn't hold.

[1]: https://github.com/dfinity/motoko/blob/master/rts/stubs.c

osa1:
See the comments in the code.

Fixes #2248


rts/stubs.c
```diff
@@ -3,15 +3,15 @@
...
+// stubs below that just trap when called.
 
-// NOTE (osa): This function previously did not return even though it should
-// (return type declared as `int`). Returning `0` here for now but we should
-// probably make sure is never used (maybe trap).
-int wctomb(char *s, wchar_t wc) { *s = '\0'; return 0; }
+int wctomb(char *s, wchar_t wc) { rts_trap("wctomb", 6); }
 
-int fputs(const char *s, void *f) { return 0; }
+int fputs(const char *s, void *f) { rts_trap("fputs", 5); }
```
ggreif, line 13: Maybe we should only do this for debug mode? When installing canisters, every byte of "ballast" might cross the limit.

osa1: > When installing canisters, every byte of "ballast" might cross the limit.

Sorry I couldn't follow this. What happens when installing a canister that can execute this code?

ulan (approved):

rts/stubs.c
ggreif: It is about the IC install message size that carries the Wasm binary. When there is dead code in there that pulls in more strings the Wasm might pass a limit and won't install any more. Here it is just a few bytes, that is probably okay. It was a general comment. Feel free to ignore.

ggreif: Looking at https://github.com/emscripten-core/emscripten/issues/8525, my suggestion is to just have a `__builtin_trap` here which is supposed to lower to the Wasm `unreachable`.

ggreif: That change (481597c) gives
``` wasm
  (func $wctomb (type 2) (param i32 i32) (result i32)
    unreachable
    unreachable)
  (func $fputs (type 2) (param i32 i32) (result i32)
    unreachable
    unreachable)
  (func $strerror (type 3) (param i32) (result i32)
    unreachable
    unreachable)
```

rts/stubs.c
```diff
@@ -3,15 +3,16 @@
...
 
-// NOTE (osa): This function previously did not return even though it should
-// (return type declared as `int`). Returning `0` here for now but we should
-// probably make sure is never used (maybe trap).
-int wctomb(char *s, wchar_t wc) { *s = '\0'; return 0; }
+// musl implementation uses system calls so we provide our own implementation
+void abort(void) { rts_trap("abort", 5); }
 
-int fputs(const char *s, void *f) { return 0; }
+// Functions below should not be called so we trap when they're called.
```
ggreif, line 12:
```suggestion
// Functions below should not be called, so we trap when they're called.
```

ggreif (approved):
I am happy with this.

PR #3003, opened 2021-12-15
RTS: enable assertions in const fns Value::{from_ptr, from_scalar}

osa1:


ulan (approved):

ggreif (approved):

ggreif:
@osa1 the +3.5% gas increase is worrying!

ggreif (changes requested):
Maybe this should be `debug_assert`?

ggreif (approved):

PR #3004, opened 2021-12-15
Apply the macOS `nix-2.4` GC treatment

ggreif:
`nix-2.4` has a (memory, Boehm) GC bug on macOS, this PR works around it.

See also #2913.

Also, for background, https://github.com/NixOS/nix/issues/4893.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,6 +36,8 @@ jobs:
     - run: cachix watch-store ic-hs-test &
 
     - name: "nix-build"
+      env:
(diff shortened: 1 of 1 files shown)


ggreif:
@osa1 this seems to fix the issue you are seeing in the other CI runs.

crusso:
Are there more .yml s to fix?

crusso (approved):

ggreif:
> Are there more .yml s to fix?

Possibly. All `nix-build`s are probably affected. We'll see. I still hope for a proper fix coming in `nix` itself.

PR #3005, opened 2021-12-15
Remove TODOs in mark-compact GC

osa1:
This removes all the TODOs in mark-compact GC as they are invalid or not
necessary.

**Line 109:** Here we call `mark_object` on the continuation table. `mark_object`
checks if the argument is already marked, but continuation table in this call
site will be unmarked, so that check is redundant. We could have another
function like `mark_object` that omits the "already marked" check, and call it
here, to save a few instructions.

Removed: This will only save us 3-4 instructions per GC, so it's probably not
worth adding more code for this and this TODO can be removed.

**Line 176:** This code marks fields of `MutBox`es in static heap. If we know
that a static `MutBox` can only point to dynamic heap, then the
`pointer_to_dynamic_heap` can be omitted.

Removed: This TODO is invalid. At least initially `MutBox`es need to point to
static data or have scalars, as Motoko canisters start with empty dynamic heap.

**Line 178:** The comment explains this nicely. If we know objects pointed by
static `MutBox`es are not aliased, then we can use a `mark_object` variant
without a "already marked" check, similar to (1) above.

Removed: I don't think we guarantee that objects pointed by static `MutBox`es
are not aliased.


ulan (approved):

ggreif (approved):
We can revisit these if marking becomes a bottleneck!

ggreif:
@Mergifyio refresh

ggreif:
@Mergifyio rebase

ggreif:
@nomeata I am a bit worried that this simple PR consistently fails on Linux. Any ides what this could be?

PR #3006, opened 2021-12-15
feat: add max_stable_pages check and flag

crusso:
Adds command line flag to limit ExperimentalStableMemory pages to compile-time maximum (defaults to 65536 / 2^16).
This is to reserves headroom for stable variable serialization of (currently) 4GB.


src/mo_config/flags.ml
```diff
@@ -24,6 +24,8 @@ let check_ir = ref true
 let package_urls : string M.t ref = ref M.empty
 let actor_aliases : string M.t ref = ref M.empty
 let actor_idl_path : string option ref = ref None
+let max_stable_pages_default = 65536
```
rossberg, line 27: Is this abstraction useful? It's not used except on the line that follows, is it?

rossberg (approved):

src/mo_config/flags.ml
crusso: Its used in the compiler flag documentation, to report the default.

PR #3010, opened 2021-12-16
Evolution white paper

rossberg:
More of a draft of a braindump than a polished paper. I may make another pass tomorrow.

### DIFF
--- a/design/WhitePaper.md
+++ b/design/WhitePaper.md
@@ -148,13 +148,6 @@ Another shortcoming in text processing is the inability to efficiently extract s
 Finally, it would be desirable to have extended syntax for multi-line text literals, ideally with interpolation, perhaps based on a generic toText function (see [shared generics](#shared-generics)).
 
 
-#### Floats
-
-While Motoko is the first blockchain language to feature floating point numbers, their functionality is still limited to the primitives that Wasm provides. The base library should include common math functionality like transcendental functions. Since these have to be implemented in software (as part of the runtime), any imprecision in such computations remains deterministic.
-
-Again, it would be desirable to implement tree shaking for the runtime to minimise the size of produced Wasm binaries.
-
-
 ### Type System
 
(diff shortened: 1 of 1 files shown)


design/WhitePaper.md
```diff
@@ -0,0 +1,713 @@
...
+Adding these requires including Unicode libraries and tables into the generated Wasm code, which can easily be several MB in size. Presumably, supporting Unicode would hence depend on some tree shaking ability of the compiler.
+
+Another shortcoming in text processing is the inability to efficiently extract substrings. Motoko provides iterators over strings, but they cannot be used as arguments to other operations, because their type is structural. A solution to this is needed, e.g., by extending text iterator objects in some way. Details tbd.
+
+Finally, it would be desirable to have extended syntax for multi-line text literals, ideally with interpolation, perhaps based on a generic toText function (see [shared generics](#shared-generics)).
+
+
+#### Floats
+
+While Motoko is the first blockchain language to feature floating point numbers, their functionality is still limited to the primitives that Wasm provides. The base library should include common math functionality like transcendental functions. Since these have to be implemented in software (as part of the runtime), any imprecision in such computations remains deterministic.
```
crusso, line 153:
```suggestion
While Motoko is the first blockchain language to feature floating point numbers, their functionality is still limited to the primitives that Wasm provides. The base library should include common math functionality like transcendental functions. Since these have to be implemented in software (as part of the runtime), any imprecision in such computations remains deterministic.
```
I think we do already - are we missing some? https://github.com/dfinity/motoko-base/blob/master/src/Float.mo

rossberg: Ugh, indeed, I forgot. Sorry Gabor :) Removed.

crusso: I guess there's the question if we want to support float32 since some AI algorithms don't mind the loss of precision.

crusso (approved):
Rubberstamping

PR #3019, opened 2021-12-20
CI: Build tests with nix-build to show logs

nomeata:
we are now building them twice, but with caching, so only a problem if
they actually fail.

Also include tests.qc in the default test run. This reverts 41b4fbecc6c78996df146bf26cdda583a5c57ee. It’s against the principle of least surprise if `nix-build` does not build something that `nix-build -A tests` builds. If the QC test runs too long, we should make it run fast by default (few iterations) and maybe add a separate job (daily?) for a long run.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,6 +40,15 @@ jobs:
         GC_DONT_GC: 1 # otherwise `nix` crash on macOS
       run: nix-build-uncached --max-jobs 4 -A all-systems-go
 
+    # The above doesn’t show nice logs, due to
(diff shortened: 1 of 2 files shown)


nomeata:
Odd, why does this build anything that the previous step in the CI job didn't already build? Needs a closer look, but not now.

ggreif:
> Odd, why does this build anything that the previous step in the CI job didn't already build? Needs a closer look, but not now.

@nomeata is this understood now?

ggreif (approved):

nomeata:
Yes, `tests.qc` was not built, but is now

PR #3042, opened 2022-01-07
feat: support 128 bit cycles API

crusso:
Upgrades Motoko to use new 128 bit cycle transfer API.

- Uses `Nat` and overflow checks to represent 128 bit values.
- Adds test `basic-cycles.mo` to check transfer work (complicated by fact that topping up of `drun` canisters is limited to 64-bit values).
- Rewrites existing tests to avoid old 64 bit prims.
- bumps `motoko-base` to use revamped `ExperimentalCycles.mo` with same external API, but new internal implementation. 

(Stumbled across restriction that multivalue elimination doesn't support anything but i32. Otherwise, could have combined `push_high` and `push_low` into single `push_high_low` operation. But that can come later).

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -2350,7 +2350,7 @@ module MakeCompact (Num : BigNumType) : BigNumType = struct
 
   let fits_unsigned_bits env n =
     try_unbox I32Type (fun _ -> match n with
-        | 32 | 64 | 128 -> G.i Drop ^^ Bool.lit true
+        | 32 | 64 -> G.i Drop ^^ Bool.lit true
         | 8 | 16 ->
           compile_bitand_const Int32.(logor 1l (shift_left minus_one (n + 1))) ^^
           G.i (Test (Wasm.Values.I32 I32Op.Eqz))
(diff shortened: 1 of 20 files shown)


src/codegen/compile.ml
```diff
@@ -6800,6 +6802,47 @@ let compile_lit_as env sr_out lit =
...
+      BigNum.compile_relop env Lt ^^
+      E.else_trap_with env "cycles out of bounds")
+
+  let push_high env =  Func.share_code1 env "__cycles_push_high" ("val", I32Type) [I64Type]
+    (fun env get_val ->
+      get_val ^^
+      (* shift right 64 bits *)
+      compile_lit_as env SR.Vanilla (Ir.NatLit (Numerics.Nat.of_big_int
+        (Big_int.power_int_positive_int 2 64))) ^^
+      BigNum.compile_unsigned_div env ^^ (* TODO: use shift right instead *)
```
ggreif, line 6836: Let's have an issue for these shifts!

test/run-drun/basic-cycles.mo
```diff
@@ -0,0 +1,101 @@
...
+      amount = amount_})
+  };
+
+
+  public func test(amount : Nat) : async () {
+     Prim.debugPrint(debug_show({available=available(); amount = amount}));
+     assert (available() == amount);
+  };
+
+  // test overflow of 128 bit cylce amounts
```
ggreif, line 32:
```suggestion
  // test overflow of 128 bit cycle amounts
```

ggreif (approved):
LGTM!

crusso:
@ggreif @nomeata I wonder if we should consider this a breaking change or not (and bump the minor version)? 

Real clients of base that don't access Prim shouldn't break since ExperimentalCycles was using Nat already, but I've changed the typing of the Motoko compiler prims, so the new compiler won't work with old versions of base.

src/codegen/compile.ml
crusso: #3060

ggreif:
I don't think that it is a breaking change. Everything behind ⛔ is a black box.

PR #3049, opened 2022-01-10
Provide basis for obtaining stable variable memory footprint

ggreif:
### Algorithm

- ~~get controllers list~~ (would need to be its own controller)
- if caller is ~~one of the controllers~~ (or itself) — trap otherwise
  - if actor specifies `system func preupgrade`
    - perform a GC
    - invoke `preupgrade`
  - compute the serialised length of the object containing all stable variables
    (this step will trap if `Nat32` wraps around)
  - convert to `Nat64`
  - add the type descriptor's length (converted to `Nat64`)
  - return it

### Rationale

A GC helps expanding the usable heap so that `preupgrade` won't be hampered by fragmentation.
Before calling the user-supplied `preupgrade` hook, we insert a GC new primitive (`ICPerformGC`, non-user-accessible). This also happens in the upgrade scenario, now.

Invoking `preupgrade` is necessary, otherwise custom logic to fill the stable variables would be ignored.

Getting the _serialised length_ of the stable variables is performed by the highly destructive primitive `ICStableSize` (also non-user-accessible) which leaves the heap in a (potentially) damaged state. No heap reads should be performed after it returns.

The endpoint for getting the stable variable footprint must be a `query`, since it may damage the heap, and needs a state rollback.

A primitive (`stableVarInfo`, user-accessible) is provided to get the query entry point such that the canister can obtain its own stable variable statistics.

`Nat64` is the result type because potentially the serialised stable variables may take more than 4 GB space. This is due to sharing, and the fact that serialisation flattens DAGs to trees and then to a buffer. **Caveat**: Internally all computation currently is done as `u32`, and an assertion makes sure that instead of wrapping a trap occurs on overflow. This means that every call to the primitive `stableVarInfo` should be wrapped in a `try` ... `catch` block.

###  Internal endpoint

The name of the internal query endpoint is `__motoko_stable_var_size`. In the future we should suppress the user-defined endpoints that have a `__` prefix (#3117).

### Lowering
The lowering of the endpoint logic is modelled like `$__get_candid_interface_tmp_hack`.

## TODOs
- [x] test explosion by sharing — e11b6fa58, `query-footprint-overflow.mo` tests when 4GB is exceeded.
- [x] ~~add flag whether to generate footprint query endpoint~~ — Moot!
- [ ] check for name conflicts: added method should not already be part of the interface — deterred to #3117
- [ ] should the query check whether it is called by the [canister's controller](https://forum.dfinity.org/t/how-to-get-the-controller-inside-the-canister-with-rust-motoko/5102)? (see also https://github.com/dfinity/cdk-rs/issues/16#issuecomment-1014052172) — Moot: only controllers can call `canister_status` :-(
- [x] perform a GC before calling the user-defined `preupgrade`?
- [x] test whether the query appears in the Candid (should not!)
    - generated IDL (manually)
    - `$__get_candid_interface_tmp_hack` reply — 476391687fbc
- [x] make the query self-callable (higher-order primitive) — 48df8d1fc
- [x] test the `u32` overflow case
- [ ] we currently drop the count of all references that might be contained in the stable variables (is this a problem?)
- [ ] should we also return other GC-related stats? — Not yet. The result record is extensible.


src/codegen/compile.ml
```diff
@@ -3581,6 +3581,18 @@ module IC = struct
...
+    let fi = E.add_fun env "canister_query stable-variable-footprint"
+      (Func.of_body env [] [] (fun env ->
+        G.i (Call (nr (E.built_in env "heartbeat_exp"))) ^^
+        E.collect_garbage env))
+    in
+    E.add_export env (nr {
+      name = Wasm.Utf8.decode "canister_heartbeat";
+      edesc = nr (FuncExport (nr fi))
+    })
+
```
ggreif, line 3595:
```suggestion
```

src/codegen/compile.ml
```diff
@@ -7922,6 +7934,17 @@ and compile_exp (env : E.t) ae exp =
       SR.Vanilla,
       ContinuationTable.size env ^^ Prim.prim_word32toNat env
 
+    | OtherPrim "rts_stable_vars_size", [e] ->
```
ggreif, line 7937:
```suggestion
    | OtherPrim "☠️rts_stable_vars_size☠️", [e] ->
```

src/codegen/compile.ml
```diff
@@ -8821,6 +8846,9 @@ and main_actor as_opt mod_env ds fs up =
     let as_local n = not (Freevars.S.mem n captured) in
     let ae1 = VarEnv.add_arguments env ae0 as_local arg_names in
 
+    (* Add footprint query as an synthetic exported field *)
+    
+
```
ggreif, line 8851:
```suggestion
```

src/codegen/compile.ml
```diff
@@ -8831,6 +8859,12 @@ and main_actor as_opt mod_env ds fs up =
 
     (* Export the public functions *)
     List.iter (export_actor_field env ae2) fs;
+    (*if true (*flags.footprint*) then
+      begin
+        let field' : Ir_def.Ir.field' = { name = "stable-variable-footprint"; var = "go"} in
+        let footprint_field = Type.{ it = field'; at = no_region; note = Func(Shared Query, Promises, [](*binds*), [], [Prim Nat64]) } in
+        export_actor_field env ae2 footprint_field
+      end;*)
```
ggreif, line 8867:
```suggestion
```

src/ir_def/construct.ml
```diff
@@ -101,7 +101,8 @@ let primE prim es =
     | RelPrim _ -> T.bool
     | SerializePrim _ -> T.blob
     | SystemCyclesAvailablePrim
-    | SystemCyclesAcceptPrim -> T.nat64
+    | SystemCyclesAcceptPrim
+    | OtherPrim "rts_stable_vars_size" -> T.nat64
```
ggreif, line 105:
```suggestion
    | OtherPrim "☠️rts_stable_vars_size☠️" -> T.nat64
```

src/lowering/desugar.ml
```diff
@@ -334,16 +334,34 @@ and export_interface txt =
...
+  let binds = [scope_bind] in
+  let typ = Func (Shared Query, Promises, binds, [], [nat64]) in
+
+  let scope_con = Con.fresh "T" (Abs ([], scope_bound)) in
+  let scope_con2 = Con.fresh "T2" (Abs ([], Any)) in
+  let bind  = typ_arg scope_con Scope scope_bound in
+  let bind2 = typ_arg scope_con2 Scope scope_bound in
+  ([ letD (var v typ) (
+    funcE v (Shared Query) Promises [bind] [] [nat64] (
+      asyncE bind2 (primE (I.OtherPrim "rts_stable_vars_size") [expr]) (Con (scope_con, []))
```
ggreif, line 364:
```suggestion
      asyncE bind2 (primE (I.OtherPrim "☠️rts_stable_vars_size☠️") [expr]) (Con (scope_con, []))
```

src/lowering/desugar.ml
```diff
@@ -386,8 +404,26 @@ and build_actor at ts self_id es obj_typ =
...
+                 at = no_region;
+                 note = f.T.typ }
+             ) fields vs)
+          ty) in
+
+  let footprint_d, footprint_f = export_footprint stable_vars in
+  I.ActorE (interface_d @ footprint_d @ ds', interface_f @ footprint_f @ fs,
      { meta;
        I.preupgrade =
        (let vs = fresh_vars "v" (List.map (fun f -> f.T.typ) fields) in
```
ggreif, line 429: reuse from `stable_vars`

test/run-drun/ok/query-footprint.drun-run.ok
```diff
@@ -0,0 +1,4 @@
+ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
+ingress Completed: Reply: 0x4449444c0000
+Ok: Reply: 0x4449444c0001780000000000000000
+Error: "Line 4: Illegal method name: stable-variable-footprint."
```
ggreif, line 4: this needs a fix (but where?)

test/run-drun/query-footprint.mo
```diff
@@ -0,0 +1,18 @@
+actor footprint = {
+    stable var s : Nat64 = 42;
+    stable var hello : Text = "Hey You!";
+    public query func go() : async Nat64 {
```
ggreif, line 4: add a non-stable variable too

src/lowering/desugar.ml
```diff
@@ -386,8 +404,26 @@ and build_actor at ts self_id es obj_typ =
...
+            (seqE (List.map (fun (i,t) -> optE (varE (var i t))) ids))])
+      (newObjE T.Memory
+          (List.map2 (fun f v ->
+               { it = {I.name = f.T.lab; I.var = id_of_var v};
+                 at = no_region;
+                 note = f.T.typ }
+             ) fields vs)
+          ty) in
+
+  let footprint_d, footprint_f = export_footprint stable_vars in
```
ggreif, line 425: this should depend on a flag

test/run-drun/query-footprint.mo
```diff
@@ -0,0 +1,19 @@
...
+    stable var s : Nat64 = 42;
+    stable var hello : Text = "Hey You!";
+    var lost : Text = "This will be lost";
+    public query func go() : async Nat64 {
+        s - 42
+    };
+    public func stay() : async Nat64 {
+        await go()
+    }
+};
```
ggreif, line 11: Add a `system func preupgrade` so that it also gets tested.

ggreif: Here: 071dc9d09.

src/lowering/desugar.ml
ggreif: here: f73a0d789

src/lowering/desugar.ml
ggreif: Yes: de549f8

crusso:
Re: test of controller. I'm not sure the System API let's you get at all the controllers at the moment, just one, so it might be hard to check this reliably without some song and dance.

It would probably also make sense to allow self to call the method, not just the controller, so the canister can detect low memory conditions.

ggreif:
> It would probably also make sense to allow self to call the method, not just the controller, so the canister can detect low memory conditions.

_self-callable_ is a TODO already

src/lowering/desugar.ml
```diff
@@ -386,32 +403,39 @@ and build_actor at ts self_id es obj_typ =
...
     I.{ candid = candid;
         sig_ = T.string_of_stab_sig sig_} in
-  let (interface_d, interface_f) = export_interface candid.I.service in
-  I.ActorE (interface_d @ ds', interface_f @ fs,
+  let interface_d, interface_f = export_interface candid.I.service in
+  let with_stable_vars wrap =
+    let vs = fresh_vars "v" (List.map (fun f -> f.T.typ) fields) in
+    blockE
+      ((match call_system_func_opt "preupgrade" es with
+        | Some call -> [ expD (primE (I.OtherPrim "⏰rts_perform_gc⏰") [])
```
ggreif, line 411: This also triggers a GC for the _upgrade scenario_. I guess this is what we actually want.

test/run-drun/query-footprint.mo
```diff
@@ -0,0 +1,41 @@
...
+    let e2 = (e1, e1);
+    let e3 = (e2, e2);
+    let e4 = (e3, e3);
+    let e5 = (e4, e4);
+    let e6 = (e5, e5);
+    let e7 = (e6, e6);
+    let e8 = (e7, e7);
+    let e9 = (e8, e8);
+    let e10 = (e9, e9);
+    stable var expl = e10;
```
ggreif, line 18: We can go _deeper_. Do we want to?

ggreif:
Looks like the [interface spec](https://smartcontracts.org/docs/interface-spec/index.html) has this:
```
  canister_status : (record {canister_id : canister_id}) -> (record {
      status : variant { running; stopping; stopped };
      settings: definite_canister_settings;
      module_hash: opt blob;
      memory_size: nat;
      cycles: nat;
  });
```
with
```
type definite_canister_settings = record {
  controllers : vec principal;
  compute_allocation : nat;
  memory_allocation : nat;
  freezing_threshold : nat;
};
```
is all we need.

nomeata:
But only controllers an call `canister_status` :-D

ggreif:
> But only controllers an call `canister_status` :-D

Yeah, that sucks. But the possibly better idea is to restrict the footprint-`query` to _self_ and let another update call deal with the authorisation and then invoke the `query`.

test/run-drun/query-footprint.mo
ggreif: Nope, it gets slow, and at `e13` (dc19ba63e02c113aa28bc0f1c8138714475e9741) the `motoko:stable_types` section overflows the 64kB limit.

src/codegen/compile.ml
```diff
@@ -8006,14 +8006,33 @@ and compile_exp (env : E.t) ae exp =
...
+      let ts = [e.note.Ir_def.Note.typ] in
+      let tydesc = Serialization.type_desc env ts in
+      let tydesc_len = Int32.of_int (String.length tydesc) in
+      compile_exp_vanilla env ae e ^^
+      Serialization.buffer_size env (Type.seq ts) ^^
+      G.i Drop ^^
+      compile_add_const tydesc_len  ^^
+      G.i (Convert (Wasm.Values.I64 I64Op.ExtendUI32))
+
+    | OtherPrim "⏰rts_perform_gc⏰", [] ->
```
crusso, line 8024: I don't think this should be a primitive, otherwise it could easily be called with a non-empty stack and collect data that is actually live by only being reachable from the stack.

I wonder if it would be safter to add the call to collect not to the body the preupgrade Motoko system method, but to the wasm only canister_preupgrade (that calls the former).

But perhaps I'm being anal about this and just not exposing it in `Prim` is enough if we are careful.

src/prelude/internals.mo
```diff
@@ -411,3 +411,17 @@ func @create_actor_helper(wasm_module_ : Blob, arg_ : Blob) : async Principal =
 func @call_raw(p : Principal, m : Text, a : Blob) : async Blob {
   await (prim "call_raw" : (Principal, Text, Blob) -> async Blob) (p, m, a);
 };
+
+// stable memory footprint
+func @query_stable_var_footprint(self : actor {}, method : ?Text) : async Nat64 {
```
crusso, line 416: Why not just fix the method name to some well known-name like `__stable_variable_footprint` - I guess you still need to use the dynamic call because you can't use express this method name as a valid identifier - we should really add some way of escaping identifiers so we can interact with foreign canisters with funny names.

src/lowering/desugar.ml
crusso: Yes, I think so, though I'd prefer it if that call to perform gc was actually in canister_preupgrade, and never reachable from motoko code (that might have something on the stack - actually, aren't the values in `vs` on the stack here - I'd be nervous that `rts_perform_gc` might collect them!)

src/lowering/desugar.ml
```diff
@@ -386,32 +407,39 @@ and build_actor at ts self_id es obj_typ =
...
+                       ; expD call]
+        | None -> []) @
+         [letP (seqP (List.map varP vs)) (* dereference any mutable vars, option 'em all *)
```
crusso, line 418: Note this will only do a GC if we have a preupgrade system method, not for every size computation.

crusso (commented):
Added some comments.

crusso:
I think if I were doing this I would extend try to move the dodgy gc step into the backend by:

- putting the pure size computation into another field of the IR ActorE's upgrade record (next to `pre` and `post` fields, maybe called `footprint` or just `size`, perhaps adding a primitive ICStableSize (similar to ICStableWrite, but doing half the work).
- In compile.ml, emit a well-known canister_query method that does the GC and then calls the size computation appropriately, as we emit calls to "pre_exp" and "post_exp"do for canister_preupgrade and canister_postupgrade,  modeled after __async_helper.

And I'd try that in separate branch off this one - in case it leads to a dead end ;->

test/run-drun/self-query-footprint.mo
```diff
@@ -0,0 +1,17 @@
+//MOC-FLAG --stable-var-footprint-query=stable-variable-footprint
+
+import { queryStableVarFootprint } "mo:⛔";
```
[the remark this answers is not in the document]
ggreif: Yeah! This `Prim.` prefix is ugly like hell...

src/prelude/internals.mo
ggreif: I was thinking of adding another primitive that gets the `?<footpring-query-name>`, and I need a switch anyway to only create it when necessary (the user wants it).

src/lowering/desugar.ml
ggreif: That is intended. Why lose time when size computation doesn't care about fragmentation, and the heap (potentially damaged) will be jettisoned anyway? We might find an argument for doing a GC _after_ having the stable variables ready for serialisation, but that is another PR.

src/lowering/desugar.ml
ggreif: I don't see yet how Motoko code can reach this. I am constructing a top-level actor method programmatically here.

src/lowering/desugar.ml
crusso: Oh, you're right. I was assuming we might need to allocate but we don't.

src/codegen/compile.ml
crusso: I think it protection would be worth it. I'm sure someone will find this and try to invoke it to force gcs at the wrong time.
Alternatively, to avoid the prim, refactor as I suggest below.

src/prelude/internals.mo
crusso: I'd just always create it, or just create it on demand if the code requires it - asking the user to set a flag for this seems awkward,  just to save a few bytes in the image. It's not that many, right?

crusso:
You were right about __motoko_async_helper being accessible (#3117, #3118).

I would just call the new method `__motoko_stable_var_size` (or whatever) and we can add a restriction that user defined shared functions cannot be declared with `__` prefix, going forward.

test/run-drun/query-footprint.mo
ggreif: > to avoid it hitting disk at all

But `e10` isn't occupying a lot of space, or is it?

ggreif:
> You were right about __motoko_async_helper being user-definable (#3117, #3118).
> 
> I would just call the new method `__motoko_stable_var_size` (or whatever) and we can add a restriction that user defined shared functions cannot be declared with `__` prefix, going forward.

We should forbid the `__motoko_` prefix

src/ir_def/ir.ml
```diff
@@ -158,6 +158,8 @@ and prim =
   (* backend stuff *)
   | CPSAwait of Type.typ
   | CPSAsync of Type.typ
+  | ICPerformGC
+  | ICStableVarsSize
```
crusso, line 162:
```suggestion
  | ICStableSize of Type.typ
```
For uniformity, I'd add a type argument, drop Var, and move to the end of the datatype, next to StableWrite and StableSize. Then you don't need to use the argument to infer the type later.

ggreif:
These are the `OtherPrim`s that give useful information when called _after_ the `preupgrade` hook has run:
- `rts_heap_size`
- `rts_memory_size`
- `rts_total_allocation`
- `rts_reclaimed`
- `rts_max_live_size`

test/run-drun/ok/query-footprint.drun-run.ok
crusso: What's this about?

src/lowering/desugar.ml
```diff
@@ -334,20 +334,39 @@ and export_interface txt =
...
+  let typ = Func (Shared Query, Promises, binds, [], [nat64]) in
+
+  let scope_con = Cons.fresh "T" (Abs ([], scope_bound)) in
+  let scope_con2 = Cons.fresh "T2" (Abs ([], Any)) in
+  let bind  = typ_arg scope_con Scope scope_bound in
+  let bind2 = typ_arg scope_con2 Scope scope_bound in
+  ([ letD (var v typ) (
+       funcE v (Shared Query) Promises [bind] [] [nat64] (
+           (asyncE bind2
+              (blockE [expD (assertE (primE (I.RelPrim (caller, Operator.EqOp))
```
crusso, line 364: The assertion message produced below is a bit obscure - maybe call OtherPrim "trap" with `"__motoko_stable_var_size: not a self-call"` or similar to what `__motoko_async_helper` produces (see compile.ml)

test/run-drun/query-footprint.mo
```diff
@@ -0,0 +1,39 @@
+import { queryUpgradeFootprint } "mo:⛔";
```
crusso, line 1:
```suggestion
import { stableVarInfo} "mo:⛔";
```

test/run-drun/query-footprint.mo
```diff
@@ -0,0 +1,39 @@
...
+    let e5 = [e4, e4];
+    let e6 = [e5, e5];
+    let e7 = [e6, e6];
+    let e8 = [e7, e7];
+    let e9 = [e8, e8];
+    let e10 = [e9, e9];
+    stable var expl = e10;
+
+    public func delegate() : async Nat64 {
+        let { stable_vars } = await queryUpgradeFootprint(footprint);
```
crusso, line 21:
```suggestion
        let { size } = await stableVarInfo(footprint);
```

test/run-drun/query-footprint.mo
```diff
@@ -0,0 +1,39 @@
...
+        stable_vars
```
crusso, line 22:
```suggestion
        size
```

crusso (commented):
Some more suggestions.

test/run-drun/ok/query-footprint.drun-run.ok
ggreif: Gone.

src/ir_def/check_ir.ml
```diff
@@ -663,6 +663,11 @@ let rec check_exp env (exp:Ir.exp) : unit =
       T.unit <: t
     | GetCertificate, [] ->
       T.Opt T.blob <: t
+    | ICPerformGC, [] ->
+      T.unit <: t
+    | ICStableVarsSize, [e1] ->
+      (* check e1 for being a record *)
```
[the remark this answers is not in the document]
ggreif: Done.

src/lowering/desugar.ml
ggreif: Hmm, I am reluctant to add user-friendliness for things that are clearly off-limits to users. It only adds weight, but no value. Who calls (manages to call) this will be able to disassemble it (or find the generating code).

src/lowering/desugar.ml
```diff
@@ -334,20 +334,39 @@ and export_interface txt =
   let open T in
   let name = "__get_candid_interface_tmp_hack" in
   let v = "$__get_candid_interface_tmp_hack"  in
-  let binds = [T.scope_bind] in
+  let binds = [scope_bind] in
   let typ = Func (Shared Query, Promises, binds, [], [text]) in
 
-  let scope_con = Cons.fresh "T" (Abs ([], T.scope_bound)) in
+  let scope_con1 = Cons.fresh "T" (Abs ([], scope_bound)) in
```
crusso, line 340:
```suggestion
  let scope_con1 = Cons.fresh "T1" (Abs ([], scope_bound)) in
```

crusso (approved):
Preapproving since you're pretty close. (One tiny suggestion if you need to rebuild anyway)...

PR #3050, opened 2022-01-10
Add `Prim.principalOfBlob()`

nomeata:
It’s silly to have `Principal.toBlob()`, but not the other direction.

### DIFF
--- a/src/prelude/prim.mo
+++ b/src/prelude/prim.mo
@@ -255,6 +255,7 @@ func time() : Nat64 = (prim "time" : () -> Nat64) ();
 // Principal
 
 func blobOfPrincipal(id : Principal) : Blob = (prim "cast" : Principal -> Blob) id;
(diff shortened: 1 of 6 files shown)


ggreif:
I assume all `Principal`s are checked for well-formedness on the system boundary entry?

crusso:
Are there really no restrictions on the binary format of a principal that we should check?

nomeata:
The only meaningful might be the length (at most 29 bytes), but so far we treat them completely opaque, including in our other intro form so far, namely `actor "…"`. And for all principals from the system we trust that the limit holds.

This is ok; too long principals are a bit like non-existent principals?

Interestingly, we _do_ check the length in Candid decoding. But that’s because there it’s untrustworthy input.

We could make it an invariant in Motoko that `Principal` is short. Not sure if it’s worth it.

crusso:
In that case, fire away.

crusso (approved):
LGTM (FWIW)

Issue #3057, opened 2022-01-12
Segmentation fault: 11 when trying to define a method which returns a new “instance” of a class


paulyoung:
I’m sure I brought this up before but I ran into it again and couldn’t find anything about it.

Here’s a fairly minimal example that reproduces the issue when run with `--check`:

```swift
module {
  public class Id<A>(x : A) {
    public func map<B>(f : A -> B) : Id<B> {
      Id(f(x));
    };
  };
};
```

ggreif:
Heh, if the good old identity functor fails, then all bets are off! ;-)

ggreif:
I bet this is a mutual recursion problem. I have seen OCaml segfaulting when exhausting the stack several times.

crusso:
Interesting...

paulyoung:
If anyone has any suggestions on how to tackle this I can try to take a look.

ggreif:
@paulyoung is this a curiosity problem ("let's code up some Haskell idioms in Motoko") or appearing in your production code?

ggreif:
@nomeata you have squashed "not lazy enough reference" bugs in the past. Maybe this rings a bell?

nomeata:
Not in the typechecker… :-)

paulyoung:
> is this a curiosity problem ("let's code up some Haskell idioms in Motoko") or appearing in your production code?

@ggreif I’m trying to do the former to see if there’s something interesting to write about and perhaps promote Motoko, but bugs like this are preventing me from reaching the latter.

ggreif:
> I think this type should actually be rejected as expansive since it involves polymorphic recursion.

By which definition of _polymorphic recursion_? It would be polymorphic recursion when the type variable (`A` for `B`) were wrapped (nested) in a different type constructor, like `Id<?B>` or `Id<(B, A)>`. But `Id<>` applied to a tyvar isn't polymorphic recursion.

PR #3058, opened 2022-01-12
Reproduce #3057

ggreif:
Reproduces the infinite recursion.

Also speeds up the normalisation by passing various `-e` to `sed`.


ggreif:
@crusso I am a bit split whether this should live in `tests/fail`, but I think this is not _intended_ to fail. Rather it fails due to a bug now, but should be fixed some time.

crusso (approved):
Should we merge this or just leave it open until #3507 is fixed?

ggreif:
I'll merge. When fixing #3507 this must be touched. Null problemo.

PR #3059, opened 2022-01-13
Switch `rustc` to a more fresh nightly

ggreif:
> [`rustc` v1.59.0 is now beta](https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html), let's use it!

Well, the stable `rustc` (just like beta) doesn't support certain features yet, so we _really_ want a nightly.

We also update all rust-related dependencies to the most recent ones.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -166,7 +166,7 @@ rec {
       vendorRustStdDeps = "${cargoVendorTools}/bin/vendor-rust-std-deps";
 
       # SHA256 of Rust std deps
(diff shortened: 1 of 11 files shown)


crusso (approved):
I assumed there's a reason we can't use a stable release?

ggreif:
> I assumed there's a reason we can't use a stable release?

Yes, we use the `-Z` flag to activate nightly-only features, and on beta we get this error:
```
error: the `-Z` flag is only accepted on the nightly channel of Cargo, but this is the `beta` channel
```

PR #3066, opened 2022-01-20
test: test --max-stable-pages

crusso:
- wanted to test on > 4GB size, but can't on my VM, so testing on 1GB instead.


ggreif (approved):
LGTM!

PR #3067, opened 2022-01-20
Bug: trap during  IDL deserialization of large array?

crusso:
Repro for Issue https://github.com/dfinity/motoko/issues/3068

The test result shows the bug, and should be changed with a proper fix.


crusso:
It's very odd that CI isn't passing the test correctly. @ggreif, @nomeata, could you try the new tests locally to see if they pass with the expected, but wrong, result I checked in the IDL buffer overrun.

I'm feeling gaslighted now...

nomeata:
It looks like you did not commit `upgrade-bug-immut.mo`

nomeata:
But yes, oddly `nix-build -A tests.drun` succeeds here.

My guess is that we are hitting size limits on our cheap Github builders that we get for free? Maybe a good time to finally ask for a bit more resources here (and maybe switch to Hercules CI to get better support for nix-based CI)

nomeata:
The CI build log says `nix build failed: signal: killed`. Yes, this smells like OOM.

nomeata:
BTW, `nix-build -A tests.drun-dbg` fails here too:

```
--- upgrade-bug.drun-run.ret (expected)
+++ upgrade-bug.drun-run.ret (actual)
@@ -0,0 +1 @@
+Return code 101
--- upgrade-bug.drun-run (expected)
+++ upgrade-bug.drun-run (actual)
@@ -1,4 +1,7 @@
 debug.print: pre
-ingress Err: IC0503: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped explicitly: IDL error: byte read out of buffer
+thread 'main' panicked at 'Ingress message did not finish executing within 10000 batches, panicking', drun/src/lib.rs:356:5
+note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
+debug.print: attempt to add with overflow
+debug.print: panic occurred in file 'src/memory/ic.rs' at line 79
```

crusso:
Looks like a missing overflow check in rts/memory/alloc_words

crusso:
@osa1 I think I've found an overflow bug in alloc_words - can you confirm?

src/codegen/compile.ml
```diff
@@ -4392,8 +4392,15 @@ module Serialization = struct
       compile_unboxed_const 0l ^^ set_ref_size ^^
 
       let inc_data_size code =
+        let (set_sum, get_sum) = new_local env "sum" in
```
ggreif, line 4395: OCaml doesn't mandate the parens.
```suggestion
        let set_sum, get_sum = new_local env "sum" in
```

src/codegen/compile.ml
```diff
@@ -4392,8 +4392,15 @@ module Serialization = struct
...
         get_data_size ^^ code ^^
         G.i (Binary (Wasm.Values.I32 I32Op.Add)) ^^
+        set_sum ^^
+        get_sum ^^
+        get_data_size ^^
+        G.i (Compare (Wasm.Values.I32 I32Op.LtU)) ^^
+        E.then_trap_with env "buffer_size overflow" ^^
```
ggreif, line 4402: this string doesn't appear in the error log...

ggreif: maybe use the exponential tower like in #3049? But somewhat higher…

crusso: Yeah, I was considering that too, but this isn't the real bug anyway, just a lurking one...

src/codegen/compile.ml
crusso: When in Rome - do as the Romans - most of the other code make the tuple explicit...

crusso:
It's annoying that Rust does overflow checking in debug mode, but not in release. I've had to hoist the overflow check to produce the same output in debug and release builds.

crusso:
replaced by #3077

PR #3071, opened 2022-01-24
test ic-ref implementation of heartbeat

crusso:
- [ ] add dummy ingress messages to drive ic-ref heartbeat (which only occurs on ingress)


luc-blaeser:
Obsolete, as ic-ref is discontinued.

PR #3076, opened 2022-01-27
feat: generalized imports supporting pattern matching and selective field import

### LINKED ISSUE #2354: Allow explicit imports
**Story**

As a developer, I want to be able to explicitly import unqualified types and functions from other modules.

**Motivation**

Currently we can only import modules as a whole with a qualified identifier.
This leads to a lot of `Result.Result` which looks silly, and makes people create module local private aliases.
This is bad because:

1.  They're adding boilerplate to work around compiler limitations
2.  They don't get hyperlinked documentation because their type signatures point to module local private type definitions.

**Demo**
I want something like this to compile
``` Motoko
    import Result "Result";
    // Showcases both type as well as value level imports
    import { type Result, mkErr } "Result";
    
    module {
      public fun fromOption<Ok, Err>(err: Err, opt: ?Ok) : Result<Ok, Err> {
        switch opt {
          case null { mkErr(err) };
          case (?ok) { Result.mkOk(ok) };
        };
      };
    }
```

**Design**

This doesn't *just work* if we allow object patterns, because we can't project type components that way.

It might also be nice to have a more compact format for the "import both as qualified and explicit" pattern, as shown above.

ggreif:
Resolves #2354, the bulk of it.

We can now import subsets of modules:
``` Motoko
import { cons; nil = empty } = "lib/ListM";

let s = cons(1, empty());
```
Yay!

These will be done separately:
- importing of `type` fields explicitly,
- dealing with explicit imports in the `languageServer` (#3078).


src/mo_def/compUnit.ml
```diff
@@ -32,6 +32,12 @@ let comp_unit_of_prog as_lib (prog : prog) : comp_unit =
     | {it = LetD ({it = VarP n; _}, ({it = ImportE (url, ri); _} as e)); _} :: ds' ->
       let i : import = { it = (n, url, ri); note = e.note.note_typ; at = e.at } in
       go (imports @ [i]) ds'
```
ggreif, line 34: Use the _prepend/reverse_ trick here to avoid quadratic runtime.

doc/modules/language-guide/examples/grammar.txt
```diff
@@ -299,6 +299,7 @@
 
 <imp> ::= 
     'import' <id>? '='? <text>
+    'import' '{' <list(<pat_field>, ';')> '}' '=' <text>
```
crusso, line 302: why not just <pat> = text?

doc/modules/language-guide/examples/grammar.txt
```diff
@@ -299,6 +299,7 @@
 
 <imp> ::= 
     'import' <id>? '='? <text>
```
crusso, line 301: I never undestood why we allo anonymous imports.... especially if they have to be pure...

doc/modules/language-guide/examples/grammar.txt
rossberg: Agreed the syntax here ought to simply be `<pat>` (or `<pat_nullary>`).

doc/modules/language-guide/examples/grammar.txt
rossberg: I believe anonymous imports are a leftover from when imports still were expressions.

doc/modules/language-guide/examples/grammar.txt
ggreif: Doesn't that allow tuple patterns and all the other things (wildcard and literals)?
What would those even mean?
Or should we just delegate those to the type-checker and accept `_` as a NOP import?

ggreif: 8d7fb3cae

doc/modules/language-guide/examples/grammar.txt
ggreif: these are gone in 8d7fb3cae!

test/fail/not-static.mo
```diff
@@ -1 +0,0 @@
-import "lib/not-static"
```
ggreif, line 1: We should leave this!
``` Motoko
import _ = "lib/not-static"
```

test/fail/self-import.mo
```diff
@@ -1 +0,0 @@
-import "self-import";
```
ggreif, line 1: ditto

src/mo_def/compUnit.ml
```diff
@@ -30,8 +30,16 @@ let comp_unit_of_prog as_lib (prog : prog) : comp_unit =
...
-      let i : import = { it = (n, url, ri); note = e.note.note_typ; at = e.at } in
+      let i : import = { it = (Surface n, url, ri); note = e.note.note_typ; at = e.at } in
       go (imports @ [i]) ds'
+    | {it = LetD ({it = ObjP []; _}, _); _} :: ds' ->
+      go imports ds'
+    | ({it = LetD (({it = ObjP (f :: fs); _} as letd), ({it = ImportE (url, ri); _} as es)); _} as imp) :: ds' ->
+      let m, tfs = Type.as_obj es.note.note_typ in
+      assert (m = Type.Module);
+      let field_id = f.it.id in
+      let i : import = { it = (Bulk (field_id, es.note), url, ri); note = Type.lookup_val_field field_id.it tfs; at = es.at } in
```
ggreif, line 41: Instead of `Bulk (field_id, es.note)` just capture the entire binding group (`{it = ObjP fs; _}`) and send it to the other side!

src/mo_frontend/parser.mly
```diff
@@ -871,9 +871,8 @@ class_body :
 (* Programs *)
 
 imp :
-  | IMPORT xf=id_opt EQ? f=TEXT
-    { let _, x = xf "import" $sloc in
-      let_or_exp true x (ImportE (f, ref Unresolved)) (at $sloc) }
+  | IMPORT p=pat_nullary EQ? f=TEXT
+    { LetD(p, ImportE(f, ref Unresolved) @? at $sloc)  @? at $sloc }
```
ggreif, line 875:
```suggestion
    { LetD(p, ImportE(f, ref Unresolved) @? at $sloc) @? at $sloc }
```

src/mo_def/compUnit.ml
ggreif: Done in c32448fca.

src/docs/extract.ml
```diff
@@ -55,7 +55,7 @@ let un_prog prog =
   let imports =
     List.map
       (fun i ->
-        let alias, path, _ = i.it in
+        let Syntax.{ it = VarP alias; _ }, path, _ = i.it in
```
ggreif, line 58: this needs some love

test/fail/self-import.mo
ggreif: 2e9df2578

test/fail/not-static.mo
ggreif: 2e9df2578

src/mo_def/compUnit.ml
ggreif: Nice idea: 0eff5af79de19d96a7985fec74a4957906ed4c45

test/repl/ok/double-import.stdout.ok
```diff
@@ -8,14 +8,12 @@ Motoko compiler (source XXX)
 -- Interpreting empty.mo:
 -- Interpreting stdin:
 -- Finished stdin:
-let anon-import-1.1 : module {...}
```
ggreif, line 11: I guess these are gone, because the wildcard pattern has a side-effect-less RHS.

ggreif:
@kritzcreek I need some guidance how to deal with explicit imports of (a list of) module symbols in th language server. It is mostly about how I can tell `languageServer` to drill into a parsed module and pick out a public binding, instead of binding an import-side binding to an entire module. I'll open an issue soon, feel free to add feedback there or here.

src/docs/extract.ml
ggreif: This now works, but ignores explicit symbol imports -> #3078 written.

doc/modules/language-guide/examples/grammar.txt
crusso: You could insist the pattern is irrefutable, I think, as with shared function args.

rossberg: > Or should we just delegate those to the type-checker (will error out) and accept _ as a NOP import?

Yes. Don't special-case syntax where types are sufficient to handle it.

ggreif: Cool, that's what I do. I have also added tests for it.

ggreif: @crusso the pattern matches arising from imports have this form
``` scheme
  (LetD (ObjP (cons (VarP cons)) (nil (VarP nil))) (VarE file$lib/ListM.mo))
```
but we have little control about the RHS `(VarE file$lib/ListM.mo)`. What we could do is, when in `check_ir`, detect ids like `file$*` and run `check (Ir_utils.is_irrefutable p)` on the LHS pattern. Is this what you mean?

ggreif: @crusso what about 4c94e004f?

test/run/import-module-explicit.mo
```diff
@@ -1,5 +1,5 @@
-import { cons; nil } = "lib/ListM";
+import { cons; nil = empty } = "lib/ListM";
```
ggreif, line 1: @crusso indeed, this works! 🥳

doc/modules/language-guide/pages/overview.adoc
```diff
@@ -69,6 +69,11 @@ Inspirations: Java, JavaScript, C#, Swift, Pony, ML, Haskell
 
 (`import MyLib "src/MyLib"` imports a library from the local file system.)
 
+Specific bindings can be imported from the module using object patterns
+....
+  import { cons; nil } "mo:base/List";
```
ggreif, line 74:
```suggestion
  import { cons; nil } = "mo:base/List";
```

test/fail/ok/refutable-import.tc.ok
```diff
@@ -0,0 +1,4 @@
+refutable-import.mo:1.8-1.20: warning [M0145], this pattern of type
```
ggreif, line 1: Nice feedback to the user that something dodgy happens here.

doc/modules/language-guide/pages/language-manual.adoc
```diff
@@ -406,7 +406,10 @@ The syntax of an _import_ `<imp>`  is as follows:
   "canister:<name>"                 import external actor by <name>
 ```
 
-An import introduces a resource named '<id>?' referring to a local source module, module from a package of modules, or canister (imported as an actor).
+An import introduces a resource named '<pat>' referring to a local source module, module from a package of modules, or canister (imported as an actor).
```
crusso, line 409:
```suggestion
An import introduces a resource referring to a local source module, module from a package of modules, or canister (imported as an actor).
The contents of the resource are bound to `<pat>`. 

Though typically a simple identifier, `<id>`, `<pat>` can also be any composite pattern binding selective components of the resource.

The pattern must be irrefutable.
```

doc/modules/language-guide/pages/language-manual.adoc
```diff
@@ -406,7 +406,10 @@ The syntax of an _import_ `<imp>`  is as follows:
...
+
+The pattern is bound to a module and explicit bindings can be established
```
crusso, line 411:
```suggestion
```

doc/modules/language-guide/pages/language-manual.adoc
```diff
@@ -1231,6 +1234,7 @@ The case sensitivity of file references depends on the host operating system so
 (Remark: when building multi-canister projects with the {sdk-short-name}, {proglang} programs can typically import canisters by alias (e.g. `import C "canister:counter"`), without specifying low-level canister ids (e.g. `import C "ic:lg264-qjkae"`).
 The SDK tooling takes care of supplying the appropriate command-line arguments to the {proglang} compiler).
 
+(Remark: sensible choices for `<pat>` are identifiers, such as `Array`, or object patterns like `{ cons; nil = empty }`, which allow explicit importing of specified symbols only, including the possibility of changing the binding name).
```
crusso, line 1237:
```suggestion
(Remark: sensible choices for `<pat>` are identifiers, such as `Array`, or object patterns like `{ cons; nil = empty }`, which allow selective importing of individual fields, under original or other names).
```

src/docs/extract.ml
```diff
@@ -51,16 +51,18 @@ and class_doc = {
...
+  let { imports; body } = comp_unit.it in
   let imports =
-    List.map
+    List.concat_map
       (fun i ->
-        let alias, path, _ = i.it in
-        (alias.it, path))
+        match i.it with
+        | { it = VarP alias; _ }, path, _ -> [ (alias.it, path) ]
+        | _ -> []) (* FIXME: explicit imports *)
```
crusso, line 61: Add a GH Issue number?

src/ir_def/check_ir.ml
```diff
@@ -1057,7 +1057,10 @@ and check_dec env dec  =
   | LetD (pat, exp) ->
     ignore (check_pat_exhaustive env pat);
     check_exp env exp;
-    typ exp <: pat.note
+    typ exp <: pat.note;
+    check env pat.at
+      Ir_utils.(expected_irrefutable dec ==> is_irrefutable pat)
+      "refutable pattern in import"
```
crusso, line 1063: I  don't think you should be checking this here but in typing.ml, where all other pats are checked.
```suggestion
```
And then remove the expected_irrefutable hack below

src/ir_def/ir_utils.ml
```diff
@@ -7,3 +7,8 @@ let rec is_irrefutable p = match p.it with
   | WildP | VarP _ -> true
   | TagP _ | LitP _ | OptP _ -> false
 
+
+let expected_irrefutable : dec -> bool = function
```
crusso, line 12: This should be unnecessary, as along we can check refutability upfront.

test/run/import-module-explicit.mo
crusso: Yeah, that's nice. But what if they want to refer to the List.List type? 

Absent type fields, we could solve that with simple alias " <pat> as <id>`patterns (a la OCaml) or more general `<pat> and <pat>` patterns, I guess. That can come later.

crusso (commented):
There's probably an informal doc section that needs updating too.

https://smartcontracts.org/docs/language-guide/modules-and-imports.html

crusso:
To do the exhaustiveness check in typing.ml, not check_ir, it might be enough to just enforce that for a  LetD whose RHS is ImportE, the pat must be exhaustive.

src/docs/extract.ml
ggreif, line 61:
```suggestion
        | _ -> []) (* FIXME: explicit imports #3708 *)
```

src/ir_def/check_ir.ml
ggreif: I think I a bit dense today. `typing.ml` looks at AST, and the `LetD` is being added to IR.
But we can see the pattern and the `ImportE` in the AST, which should be good enough. Is this what you mean?

doc/modules/language-guide/pages/language-manual.adoc
ggreif: Somehow your eraser doesn't work :-(

test/run/import-module-explicit.mo
ggreif: The will need a second `import` for now. Is that posssible? I guess so.

doc/modules/language-guide/pages/language-manual.adoc
```diff
@@ -406,7 +406,14 @@ The syntax of an _import_ `<imp>`  is as follows:
...
 
-An import introduces a resource named '<id>?' referring to a local source module, module from a package of modules, or canister (imported as an actor).
+An import introduces a resource referring to a local source module, module from a package of modules, or canister (imported as an actor).
+The contents of the resource are bound to `<pat>`. 
+
+Though typically a simple identifier, `<id>`, `<pat>` can also be any composite pattern binding selective components of the resource.
+
+The pattern must be irrefutable.
+
+by using an object pattern.
```
ggreif, line 416:
```suggestion
```

src/ir_def/check_ir.ml
```diff
@@ -1057,7 +1057,7 @@ and check_dec env dec  =
   | LetD (pat, exp) ->
     ignore (check_pat_exhaustive env pat);
     check_exp env exp;
-    typ exp <: pat.note
+    typ exp <: pat.note;
```
ggreif, line 1060:
```suggestion
    typ exp <: pat.note
```

src/mo_frontend/typing.ml
```diff
@@ -2446,6 +2444,11 @@ and infer_block_valdecs env decs scope : Scope.t =
     ) (env, scope) decs
   in scope'
 
+and is_import d =
+  match d.it with
+  | LetD (_, {it = ImportE _; _}) -> true
```
ggreif, line 2449: all patterns allowed here

doc/modules/language-guide/examples/grammar.txt
ggreif: Okay now I reuse the coverage check from `typing.ml` but make it an error when its is happening in an error. PTAL!

test/run/import-module-explicit.mo
ggreif: Like this:
``` Motoko
import { cons; nil = empty } = "lib/ListM";
import L = "lib/ListM";

type stack = L.List<Int>;
```
and this works.

test/fail/ok/refutable-import.tc.ok
ggreif: And now this is an error.

doc/modules/language-guide/pages/language-manual.adoc
```diff
@@ -406,7 +406,10 @@ The syntax of an _import_ `<imp>`  is as follows:
   "canister:<name>"                 import external actor by <name>
 ```
 
-An import introduces a resource named '<id>?' referring to a local source module, module from a package of modules, or canister (imported as an actor).
+An import introduces a resource named '<pat>' referring to a local source module, module from a package of modules, or canister (imported as an actor).
+
+The pattern is bound to a module and explicit bindings can be established
+by using an object pattern.
```
ggreif:
```suggestion
```

ggreif:
>

ggreif:
> There's probably an informal doc section that needs updating too.

But that is not in this repo, right?

src/mo_frontend/typing.ml
```diff
@@ -2468,7 +2471,7 @@ and infer_dec_valdecs env dec : Scope.t =
     Scope.{empty with val_env = T.Env.singleton id.it obj_typ}
   | LetD (pat, exp) ->
     let t = infer_exp {env with pre = true} exp in
-    let ve' = check_pat_exhaustive warn env t pat in
+    let ve' = check_pat_exhaustive (if is_import dec then local_error else warn) env t pat in
```
crusso, line 2474: perfect!

doc/modules/language-guide/pages/language-manual.adoc
```diff
@@ -1231,6 +1236,7 @@ The case sensitivity of file references depends on the host operating system so
 (Remark: when building multi-canister projects with the {sdk-short-name}, {proglang} programs can typically import canisters by alias (e.g. `import C "canister:counter"`), without specifying low-level canister ids (e.g. `import C "ic:lg264-qjkae"`).
```
crusso, line 1236:
```suggestion
(Remark: when building multi-canister projects with the {sdk-short-name}, {proglang} programs can typically import canisters by alias (e.g. ``import C "canister:counter"``), without specifying low-level canister ids (e.g. ``import C "ic:lg264-qjkae"``.)
```

doc/modules/language-guide/pages/language-manual.adoc
```diff
@@ -1231,6 +1236,7 @@ The case sensitivity of file references depends on the host operating system so
...
 The SDK tooling takes care of supplying the appropriate command-line arguments to the {proglang} compiler).
 
+(Remark: sensible choices for `<pat>` are identifiers, such as `Array`, or object patterns like `{ cons; nil = empty }`, which allow selective importing of individual fields, under original or other names).
```
crusso, line 1239:
```suggestion
(Remark: sensible choices for `<pat>` are identifiers, such as ``Array``, or object patterns like ``{ cons; nil = empty }``, which allow selective importing of individual fields, under original or other names.)
```

doc/modules/language-guide/pages/language-manual.adoc
ggreif: Let's change the remark above too?

ggreif: 3e8226ce041e941839e53d4803bdda4d8ce29298

src/mo_def/compUnit.ml
```diff
@@ -88,14 +89,15 @@ let obj_decs obj_sort at note id_opt fields =
...
-      note = { note_typ = note; note_eff = Type.Triv } }) imports
+  let import_decs =
+    List.map (fun { it = (pat, fp, ri); at; note} ->
+        { it = LetD (pat,
+                     { it = ImportE (fp, ri);
+                       at;
+                       note = { note_typ = note; note_eff = Type.Triv} });
+          at;
+          note = { note_typ = note; note_eff = Type.Triv } }
+      ) imports
```
crusso, line 100: Use the identation style on the left - I think that's what Andreas was advocating... (at least my understanding of it).

src/mo_frontend/typing.ml
crusso: thanks!

src/mo_def/compUnit.ml
ggreif: This is what `emacs` gave me, will change.

test/repl/ok/double-import.stdout.ok
crusso: I think its gone because we don't have the optional <id>? production which would introduce a gensym-ed, anonymous binding (not a wildcard).

src/mo_def/compUnit.ml
ggreif: b5ed7d031ceb01527c84dfe0107c596e815616a8

doc/modules/language-guide/examples/grammar.txt
```diff
@@ -298,7 +298,7 @@
     <obj_body>
 
 <imp> ::= 
-    'import' <id>? '='? <text>
+    'import' <pat_nullary> '='? <text>
```
crusso, line 301: Is there any reason this isn't just `<pat>` as for `let`?

ggreif: We don't want `(?(#foo))` here, do we?

crusso: Yes, but I don't think we should enforce typing properties using the grammar. 
Also, it keeps the grammar closer to the informal language reference.
Plus, you might want a type annotation on the pattern without having to parenthesize it.

crusso (approved):
For safetey, I add a quick test/run-drun test to see this works on imports to compiled actors - it should be fine, but the code for compilation units is more complicated than it should be and something could be going wrong.

ggreif:
> For safety, I would add a quick test/run-drun test to see this works on imports to compiled actors - it should be fine, but the code for compilation units is more complicated than it should be and something could be going wrong.

I get errors: af6941a96, but that is fine, I think.

doc/modules/language-guide/examples/grammar.txt
ggreif: Can I do this tomorrow in a new PR? I am totally burnt out, sorry.

crusso:
> > For safety, I would add a quick test/run-drun test to see this works on imports to compiled actors - it should be fine, but the code for compilation units is more complicated than it should be and something could be going wrong.
> 
> I get errors: [af6941a](https://github.com/dfinity/motoko/commit/af6941a96aeb7ca78c11bf506d050803f75313a8), but that is fine, I think.

Right, but that won't exercise the codegen path. 

Maybe add another or different test (Like your List client in test/run above, but enclode in `actor {...}`), that actually compiles and runs.


BTW, do we actor patterns and module patterns, if you specify the keyword? Indeed, it's a bit odd that we can use a record pattern to deconstruct something that is declared as a module (shouldn't the full pattern syntax be `<objsort> {field_pat *}`, but never mind...)

doc/modules/language-guide/examples/grammar.txt
crusso: Of course.

crusso: Tomorrow is saturday. Monday!

crusso:
(wrote you a drun test in #3081 - to merge into this branch)

doc/modules/language-guide/examples/grammar.txt
ggreif: JFTR, `<pat>` gives
```
      menhir mo_frontend/parser.{ml,mli} (exit 1)
(cd _build/default && /nix/store/ih98div9nccad7gjvksxwzhyl0jag1m7-ocaml4.12.0-menhir-20211012/bin/menhir --table --inspection -v --strict mo_frontend/parser.mly --base mo_frontend/parser)
Error: one state has shift/reduce conflicts.
Warning: one shift/reduce conflict was arbitrarily resolved.
```

doc/modules/language-guide/examples/grammar.txt
ggreif: It doesn't work: https://github.com/dfinity/motoko/pull/3076#discussion_r795963018

Issue #3072, opened 2022-01-27
Provide syntax for functional record update operations


ggreif:
From the `README.md` of a certain project:

> Technically the record types of Motoko form a _meet sub-semilattice_
> of the global type lattice. Unfortunately the value side of things is less
> convenient, and Motoko doesn't (yet) provide the _record extension_
> (concatenation) and record _field update_ operations (along with
> correspondingly intuitive syntax). In contrast, moving up in the meet
> semilattice is accomplished by record subsumption (i.e. forgetting
> fields is always automatic) already.

Sadly, the lack of these shows in the Motoko code, we have to copy fields a lot. With the record features the thing would be much more compact.

@krpeacock also asked for this in the past.

This all needs some design work, e.g. should record concatenation be biased to the left when duplicated fields occur, etc.
(My suggestion would be to error on non-empty joins, and require users to upcast in order to forget fields.)

## Strawman

Let's explore the linguistic construct like "I want sports seats for my existing car". That means I have `existing_car` and get a new version of it such that `revised_car.seats == #sport` holds. So it would be neat to do such revisions in Motoko with the syntax
``` Motoko
let revised_car = { seats = #sport for existing_car };
```
-----------------
So I suggest:

(Repeated) field extension/update could be written with the `for` keyword:
``` Motoko
{ new_field = 42; another = "Easy" for existing_record }
```
When `existing_record` already has a mentioned field then it would be updated, otherwise extended. Semicolon would be optional before the `for`. One could have `and` instead of `for`, which would give us nice record concatenation syntax too:
``` Motoko
let settings = get_settings(key, settings_db);
let funds = get_funds(key, finances_db);
return { settings and funds }
```
It would also parallel how record type meets are written. This syntax is currently forbidden by Motoko:
```
> { settings and funds };
stdin: syntax error [M0001], unexpected token 'and', expected one of token or <phrase> sequence:
  }
  <annot_opt>
  ; seplist(<exp_field>,<semicolon>)
  = <exp(ob)>
  <annot_opt> = <exp(ob)>
```
Iterated concatenation (also combined with field extension/update) is possible:
``` Motoko
{ timestamp = now() and settings and funds }
```

#### Punning

What about field name punning?
``` Motoko
let timestamp = now();
{ timestamp for settings and funds }
```
This would work, but below not:
``` Motoko
let timestamp = now();
{ timestamp and settings and funds }
```
Here `timestamp` wants to be a punned field, not a record.

#### Presence of `var` fields

[...]

PR #3084, opened 2022-01-29
feat: Field extensions/updates to existing objects

ggreif:
Related to #3072.

## Functional record updates

This adds the ability to overwrite a set of fields (in a potentially type-changing manner) relative to a (set of) base(s).

The syntax is
``` Motoko
{ base_1 and <more-bases> with a = val; <more-fields> }
```
The resulting object has type `{ a : val_ty;  <more-fields> } and base_1_ty and <more-bases_ty>` as long as the bases and the new fields are field-disjoint. In general the overriding fields are removed from the bases and the resulting type is formed from the overwritten fields' types plus the rump bases' field types.

The full syntax is a Swiss army knife of thing, allowing
- record construction from fields
- object field extension
- object field type changing
- object concatenation
- (when opted-in with `--experimental-field-aliasing`) ability to create entangled `var` fields (internally: aliased mutable cells).

## Some fine points
- When the field list is omitted, `{ base_1 and base_2 and base_3 }` syntax is object concatenation, analogous to the type-level `and` operator.
- The `and` operator (when surrounded by braces) puts the bases into object context. This is a different flavour of product construction from the `Bool` (monoidal) operation (also a product).
- When any two base is overlapping in a common field, then an explicit overwrite of that field is required to remove any ambiguity.
- [EXPERIMENTAL, `--experimental-field-aliasing` must be specified] All non-overwritten `var` fields from bases will show up in the resulting object as `var` fields that alias the original field from its base. This behaviour can be disabled by adding an overwrite `base with var field = base.field`. Note that this is analogous to the aliasing behaviour of shared `[var]` arrays.

## TODOs
- [x] how to exclude `{ in {} }` in the parser? FIXED: 47ebf5a11e1cae94dc37bc830658a385ac4509be
- [x] do we want to exclude semi in `{ a = 25; in { b = 42 } }`? Yes, FIXED: fbf5ade8f
- [x] field replacement and subtyping?
- [x] `IN base=exp(ob)` is too permitting (future?) — FIXED: 3120aa12737918919ca3238494ff38aa93b06301
- [ ] multi-value fetching from base?
- [x] documentation
- [ ] constness inference should see bases
- [x] compiler tests for `var` field aliasing (works in interpreter)
- [x] effect ordering: fields vs. bases
- [ ] consider `module { M and N };` where `M`, `N` are modules
- [x] tests with `await`
- [x] bases that have bounded polymorphic types (see f4d8bbe and #3405)
- [x] add tests ruling out `Text` etc. (the special member hacks) — See https://github.com/dfinity/motoko/pull/3084/files#r947909810, however
- [ ] consolidate `RefD` to prim + `VarD` — write issue!
- [ ] `Changelog.md`
- [x] Check interaction with type fields (claudio)
- [x] type fields aren't ambiguous when they are the same
- [x] don't let values overwrite types
- [x] test `{ t : type } and { t : Val }` — what to do? They live in different namespaces!


crusso:
I think this needs some ascii typing and evaluation rules before any more coding.

Re typing, I think you are taking the glb of the bases so won't

`{{x=false} in {x=0}}` have type `{x : None}`?

That's not sound, I think.

ggreif:
> That's not sound, I think.

Yeah, right that is `field replacement and subtyping?` checkbox above. When I replace a field, that has to be removed from the bases and re-added according to the `exp_field`.

ggreif:
@crusso I have added a "Formalisation" section to my proposal in #3072.

rossberg:
Sorry, not a fan of this choice of syntax. Something like "exp with fields" is more common and reads more intuitively.

Is there a reason we cannot use `<exp> with {...}`, so that it can be nicely generalised to full record composition later?

src/lowering/desugar.ml
```diff
@@ -92,8 +92,12 @@ and exp' at note = function
       (varP v) (varE v) ty).it
   | S.ObjBlockE (s, dfs) ->
     obj_block at s None dfs note.Note.typ
-  | S.ObjE efs ->
+  | S.ObjE (efs, []) ->
     obj note.Note.typ efs
```
ggreif, line 96: Later this can become `obj_extend note.Note.typ efs []` and `obj` then becomes redundant.

krpeacock:
A common use case in the JS/TS language is to modify multiple fields at once using the spread operator. That system uses sequence to determine which value to use when keys override. I've found this to be a very friendly and concise pattern.

For example:

```ts
const apartment = {
  rooms: 4,
  baths: 2,
  type: "apartment",
  backyard: true,
};

// Very similar home
const house = { ...apartment, type: "single family" };

// multiple changes
const condo = {
  ...apartment,
  rooms: 2,
  baths: 1,
  type: "condominium"
};
```

The condo retains the backyard, but then overrides the keys labeled "rooms", "baths", and "type". Multiple modifications can take place at once

ggreif:
> A common use case in the JS/TS language is to modify multiple fields at once using the spread operator.

My suggestion covers multiple field additions/changes:
``` Motoko
let apartment = {
  rooms = 4;
  baths = 2;
  typ = "apartment";
  backyard = true
};

let condo = {
  rooms = 2;
  baths = 1;
  typ = "condominium"
  in apartment
};
```

test/run/objects-order.mo
```diff
@@ -5,4 +5,4 @@ let _ = {a = x := 1; b = x := 2};
 assert (x == 2);
 
 let _ = {b = x := 3; a = x := 4};
-assert (x == 4);
+//assert (x == 4);
```
ggreif, line 8: This is a regression, needs a fix!

ggreif: It also needs tests for the extension syntax.

kentosugama:
@ggreif commenting

src/lang_utils/error_codes.ml
```diff
@@ -96,7 +96,7 @@ let error_codes : (string * string option) list =
     "M0090", None; (* Actor reference must have an actor type *)
     "M0091", None; (* Mutable array expression cannot produce expected type *)
     "M0092", None; (* Async cannot produce scope *)
-    (* "M0093" Unallocated *)
+    "M0093", None; (* Bases of record extensions must be objects *)
```
ggreif, line 99: or modules

ggreif:
```suggestion
    "M0093", None; (* Bases of record extensions must be objects or modules *)
```

test/repl/object-ext.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+# Tests that fields completely override base fields
+moc -i <<__END__
+{ x = -25 in { x = 42 : Nat } };
+
+module X {public let a = 7};
+{ y = -25 in { y = 3; z = "H" } and X };
```
nomeata, line 7: Out curiosity, why is this a repl test?

ggreif: Will be rewritten, as it is not testing the compiler.

ggreif: It could be a `test/fail` one

crusso:
The interpeter at least is broken, due to the problem with a subject reduction.

Here, `r` statically doesn't contain `l`, but dynamically does, so the merge in `m` drops the field from the record.

You might get this to work if the interpreter merges based on the static type of the value (which may be what the desugar is doing (I haven't checked yet).
```
[nix-shell:~/motoko/src]$ rlwrap moc
Motoko compiler (source 0.6.29-95-g4b1b8856d)
> let r : {} = {l = 0};
let r : {} = {l = 0}
> let s = {l = true};
let s : {l : Bool} = {l = true}
> let m = {r and s};
let m : {l : Bool} = {}
> m.l
  ;
> (unknown location): execution error, unbound identifier l in domain
```

nomeata:
Are the IR changes just for optimizing the repeated object calculation, or are they actually needed for correctness?

ggreif:
> Are the IR changes just for optimizing the repeated object calculation, or are they actually needed for correctness?

I assume you are referring to my changes I d437fc9e81a277aef2bb9aede304c2d06f654f56. The IR didn't provide a way to reference mutboxes. When a base `var` field is reused (i.e. not being overwritten), then the mutbox is shared between the original record and the new one. This ensures that `var` fields from bases don't cons and it introduces a new feature of _aliased fields_.

src/ir_def/freevars.ml
```diff
@@ -138,6 +138,7 @@ and cases cs : f = unions case cs
 and dec d = match d.it with
   | LetD (p, e) -> fd_of_defs (pat p) +++ exp e
   | VarD (i, t, e) -> fd_of_defs (M.singleton i t) +++ exp e
+  | RefD (i, t, e) -> fd_of_defs (M.singleton i t) +++ lexp e
```
nomeata, line 141: The idea is that it produces, like VarD, a pointer to a mutable box, but unlike VarD, it is not allocated and filled with the given value, but instead read off an object.

This can only work of the lexp is not an array index, right? So maybe check that in Check_IR. Or maybe cleaner: change the syntax to take an expression and a field name?

src/ir_def/ir.ml
```diff
@@ -95,7 +95,7 @@ and meta = {
   }
 
 and field = (field', Type.typ) Source.annotated_phrase
-and field' = {name : Type.lab; var : id} (* the var is by reference, not by value *)
+and field' = {name : Type.lab; var : id; byref : bool} (* the var is by reference, not by value *)
```
nomeata, line 98: Do you really need this field? The var was read by reference before already (to make DefineD work), so this ought to work fine for your aliasing with RefD just fine?

src/ir_def/freevars.ml
nomeata: Maybe instead of introducing RefD, we could get rid of VarD, use just LetD here, but introduce NewE to create new varbox and ObjRefE to read the mutable box off an object? This would also subsume DeclareD if we have a NewUninitializedE or something like that.

In all cases the result would be a pointer to a MutBox (type Mut t) - quite nice.

src/ir_def/ir.ml
ggreif: The problem is that `is_mut` always used to allocate a fresh mutbox for a field. This flag makes it more like an immutable initialisation. I'd be super happy to get rid of it, but have no  better idea right now :-)

src/ir_def/freevars.ml
ggreif: Yes, `RefD` having a `lexp` in it is too liberal. This was just my first push to get something that works.

src/ir_def/ir.ml
nomeata: Is it? The comment right there says something else, and I think in order to support await in record construction, at the IR level, all mutable fields are always initialized by reference, not by creating a new box.

I should look at the code on the laptop, though.

nomeata: I have an idea, will create a new issue for it… https://github.com/dfinity/motoko/issues/3351

ggreif: I didn't touch the comment. But I *think* it means "read the value off that variable", i.e. indirect. The construction of the outbox happens in `compile.ml:9160` where we forego the `Var.get_aliased_box` when that flag is set.

ggreif: Aaahh, if we could trick `get_aliased_box` into seeing a `HeapInd I` then it would do the trick for us, and `byref` would become redundant. I need to read up on ``HeapInd`!

nomeata: Exactly. And it’s not “tricking” – I think your `RefD` will invariably create something that is stored on the heap, as everything else would be wrong™.

So I think you don’t need `byref`, but “just” have to make sure that `AllocHow` does the right thing.

nomeata: Hmm, it’s fishy that `get_aliased_box` doesn’t just _always_ find a mutable box there. This relative dynamic alias-or-not is a bit dubious…

ggreif: I think there is a disconnect. `AllocHow.StoreHeap` (which makes `HeapInd`-accessed variables) is _not_ the correct notion here. Why? Because `HeapInd`-variables expect the trifecta of alloc/store-in/read-from of _mutboxes_, but what we want here is to have plain locals here that _deal with_ mutbox pointers! So `RefD` should default to `LocalImmut` (as it contains a plain pointer), but `get_aliased_box` should be prepared to deal with that by doing a `Var.get_val`.

nomeata: That catch all is bogus, see my other PR.

How do RefD and VarD differ? Don't both introduce a name of type Mut t that's represented by a pointer (in a local, of course) to a MutBox on the heap? That pointer is created differently (new box vs. reading a record), but afterwards I see no difference yet.

ggreif: I pushed my WIP as 19bbf3072. It contains vestiges of my `AllocHow.StoreHeap` experiment, which went wrong. Will cleanup later.

default.nix
```diff
@@ -782,6 +782,7 @@ rec {
           nixpkgs.niv
           nixpkgs.nix-update
           nixpkgs.rlwrap # for `rlwrap moc`
+          nixpkgs.difftastic
```
ggreif, line 785: back this out

src/ir_def/ir.ml
nomeata: Ok, but isn't that the same thing? 
The settable location from VarD is represented as a local with a pointer to a MutBox. I can read from the Box (VarE) or get the address of the box (NewObjE). Is that not enough?

Yes, it is a change of perspective: if you think of the RefD's _value_ as the VarD's _address_ they are different. But for the RefD you only care about it's address for putting it into an object, right? (or is this where I am confused?)

ggreif: Yes, `RefD` is just a pointer to a mutbox that wants to be stashed into a `var` field of a new record.

default.nix
nomeata: But thanks for the recommendation!

src/ir_def/ir.ml
nomeata: Right, and so is a `VarD` :-)

src/codegen/compile.ml
```diff
@@ -9459,20 +9470,32 @@ and compile_dec env pre_ae how v2en dec : VarEnv.t * G.t * (VarEnv.t -> scope_wr
...
+      alloc_code,
+      (fun ae -> let sr, code = Var.set_val env ae name in compile_exp_as env ae sr e ^^ code),
+      unmodified
+    )
+  | RefD (name, _, e) ->
+    assert AllocHow.(M.find_opt name how = Some (StoreHeap));
+    let pre_ae1, alloc_code = AllocHow.add_local env pre_ae how name in
+
+    ( pre_ae1,
+      alloc_code,
```
nomeata, line 9485: This may be a problem. You might want to allocate differently than for VarD (a new local, but not a new box). Didn't check closely though.

src/codegen/compile.ml
```diff
@@ -9459,20 +9466,34 @@ and compile_dec env pre_ae how v2en dec : VarEnv.t * G.t * (VarEnv.t -> scope_wr
...
-        (fun ae -> let sr, code = Var.set_val env ae name in compile_exp_as env ae sr e ^^ code),
-        unmodified
-      )
+    ( pre_ae1,
+      alloc_code,
+      (fun ae -> let sr, code = Var.set_val env ae name in compile_exp_as env ae sr e ^^ code),
+      unmodified
+    )
+  | RefD (name, _, e) ->
+    assert AllocHow.(M.find_opt name how = Some (StoreHeap));
```
ggreif, line 9477: not needed

src/ir_def/ir.ml
ggreif: Yes, but `VarD` will create unique mutboxes, right? `RefD` avoids  that. Anyway, the effective changes in `compile.ml` are now down to the two additions about `RefD` and a few `compile_lexp` additions that will go away (when I change `RefD`'s `lexp` to `(lab, exp)`. The goodies are in 06674bae8. PTAL!

src/codegen/compile.ml
```diff
@@ -9459,20 +9466,34 @@ and compile_dec env pre_ae how v2en dec : VarEnv.t * G.t * (VarEnv.t -> scope_wr
...
+    let pre_ae1, _ = AllocHow.add_local env pre_ae how name in
+
+    ( pre_ae1,
+      G.nop,
+      (fun ae ->
+        let _, prepare_raw, sr, _ = compile_lexp env ae e in
+        assert (sr = Vanilla);
+        match VarEnv.lookup_var ae name with
```
ggreif, line 9485: This should be factored out to `Var.capture_aliased_box`

ggreif: done in 1ad46fa50

src/codegen/compile.ml
ggreif: 8b3021925

src/codegen/compile.ml
```diff
@@ -8094,21 +8104,24 @@ let compile_load_field env typ name =
 (* compile_lexp is used for expressions on the left of an assignment operator.
    Produces some preparation code, an expected stack rep, and some pure code taking the value in that rep*)
 let rec compile_lexp (env : E.t) ae lexp =
-  (fun (code, sr, fill_code) -> (G.with_region lexp.at code, sr, G.with_region lexp.at fill_code)) @@
+  (fun (code, ref_code, sr, fill_code) -> G.(with_region lexp.at code, with_region lexp.at ref_code, sr, with_region lexp.at fill_code)) @@
   match lexp.it with
   | VarLE var ->
      let sr, code = Var.set_val env ae var in
-     G.nop, sr, code
+     G.nop, G.nop, sr, code
```
ggreif, line 8111:
```suggestion
     G.nop, G.nop(* FIXME *), sr, code
```

src/codegen/compile.ml
```diff
@@ -9459,20 +9472,31 @@ and compile_dec env pre_ae how v2en dec : VarEnv.t * G.t * (VarEnv.t -> scope_wr
...
-        (fun ae -> let sr, code = Var.set_val env ae name in compile_exp_as env ae sr e ^^ code),
-        unmodified
-      )
+    ( pre_ae1,
+      alloc_code,
+      (fun ae -> let sr, code = Var.set_val env ae name in compile_exp_as env ae sr e ^^ code),
+      unmodified
+    )
+  | RefD (name, _, e) ->
+    let pre_ae1, _ = AllocHow.add_local env pre_ae how name in
```
nomeata, line 9483: Once this works I'd rather not use a function and throw away a result, but just do whats need. Maybe add a `add_local_for_alias` or something. Or maybe there is already a function that does exactly what's needed here.

src/codegen/compile.ml
```diff
@@ -9459,20 +9472,31 @@ and compile_dec env pre_ae how v2en dec : VarEnv.t * G.t * (VarEnv.t -> scope_wr
...
+
+    ( pre_ae1,
+      G.nop,
+      (fun ae ->
+        let _, prepare_raw, sr, _ = compile_lexp env ae e in
+        prepare_raw ^^
+        Var.capture_aliased_box env ae sr name),
```
nomeata, line 9490: I am worried about sr showing up here. We are writing a pointer to a local, there cannot be any stackreps, because no values are represented! Probably an artifact of using lexp here, which is not quite right. Maybe just handle aliasing object fields, and write that directly?

src/codegen/compile.ml
ggreif: Right, `alloc_code` is gone now.

src/codegen/compile.ml
ggreif: Yes, what is now in `compile_lexp` will appear here inlined.

src/codegen/compile.ml
ggreif: 404c24fd8!

src/codegen/compile.ml
ggreif: Does this look better? aa4e131f5

ggreif:
> The interpeter at least is broken, due to the problem with subject reduction alluded to in #3072. Your compilation via desugaring might still be ok though (yet to check that).

Okay, I did the type-guided base pruning in 5325ade92. The previous commit added you test.

src/mo_frontend/typing.ml
```diff
@@ -1419,7 +1460,7 @@ and check_exp' env0 t exp : T.typ =
   | TupE exps, T.Tup ts when List.length exps = List.length ts ->
     List.iter2 (check_exp env) ts exps;
     t
-  | ObjE exp_fields, T.Obj(T.Object, fts) ->
+  | ObjE (exp_fields, []), T.Obj(T.Object, fts) -> (* FIXME: bases? *)
```
ggreif, line 1463: Can we offer better error messages than the standard `M0096` (below) if the bases are known?

src/ir_def/check_ir.ml
```diff
@@ -1105,6 +1108,14 @@ and gather_dec env scope dec : scope =
     let val_info = {typ = T.Mut t; const = false; loc_known = env.lvl = TopLvl} in
     let ve = T.Env.add id val_info scope.val_env in
     { val_env = ve }
+  | RefD (id, t, lexp) ->
```
crusso, line 1111: It might be nicer if this was similar to VarD and used `t` to describe the contents of the mutable cell, not the type of the cell. Then you could also share the code of the two branches quite nicely for gather_dec

crusso:
Gabor, I think you allow mixtures of objects and modules but I would restrict to just objects (records) for now, otherwise we should impose static restrictions when constructing modules (for consistency)

doc/md/language-manual.md
```diff
@@ -1654,6 +1654,24 @@ Such an object literal, sometimes called a *record*, is equivalent to the object
 
 Object expressions support *punning* for concision. A punned field `<id>` is shorthand for `<id> = <id>`; Similarly, a typed, punned field `<id> : <typ>` is short-hand for `<id> = <id> : <typ>`. Both associate the field named `<id>` with the value of the identifier `<id>`.
 
+#### Object reuse
```
ggreif, line 1657: maybe "combination"?

src/mo_frontend/typing.ml
```diff
@@ -1048,11 +1051,49 @@ and infer_exp'' env exp : T.typ =
...
+                info env h_exp.at "field also present in base, here (consider overwriting)"
+              end in
+          iter (fun (b_t, b) -> avoid_labels b (map (fun {T.lab; _} -> lab) (T.as_obj b_t |> snd))) t in
+        iter avoid (T.as_obj h |> snd);
+        disjoint t in
+    disjoint (map2 (fun b_t b -> b_t, b) stripped_bases exp_bases);
+
+    (* TODO: var in stripped_bases? *)
+    let t_base = T.(fold_left glb (Obj (Object, [])) stripped_bases) in
+    T.(glb t_base (Obj (Object, sort T.compare_field fts)))
```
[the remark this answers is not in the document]
ggreif: So that I get a sorted result :-)

ggreif:
@crusso as an addendum to the huddle we had Monday morning, I dumped the IR by commit 36c0f4f
``` Motoko
  (LetD
    (VarP c0)
    (BlockE
      (VarD c Int (LitE (IntLit 0)))
      (LetD
        (VarP incr)
        (FuncE
          incr
          Local
          Returns
          ()
          (AssignE
            (VarLE c)
            (PrimE (BinPrim Int AddOp) (VarE c) (LitE (IntLit 1)))
          )
        )
      )
      (NewObjE Object (incr incr) (c c) {var c : Int; incr : () -> ()})
    )
  )
```
This shows that methods accessing `var` fields are actually closing over a _local_ `var` and then both the resulting mutcell and the method function are used to construct the object. I had the erroneous impression that the method would close over the resulting object and use `DotE` to grab the `var`. This means that methods should be accessing variables rather efficiently already, and having inlined mutable fields would not make things run faster (though would consume less memory), but the compilation strategy would need a rework.

So I am pretty happy with the current aliasing semantics of the `var` fields coming from bases.

doc/md/language-manual.md
```diff
@@ -1654,6 +1654,24 @@ Such an object literal, sometimes called a *record*, is equivalent to the object
...
+- supplying fresh fields,
+- changing field types, and
+- breaking up entanglement of `var` fields.
+
+The resulting type is determined by the bases' (and explicitly given fields') static type.
+
+`var` fields from the bases (when not disambiguated) will show up in the result as `var` fields that *share* the original contents,
+i.e. an assignment to such a field of either the base or the result will change the shared content. If such a behaviour is not desired,
+then the field must be explicitly overwritten in the explicit field list.
+
```
nomeata, line 1674: Does that mean I cannot alias a var field from one base when the same name appears in another base?

ggreif: No, you can always _upcast_ the base with the undesired field.

nomeata: Ah, ok. A bit unwieldy, I guess, as you have to write out the full type. But maybe rare enough to be ok in practice.

ggreif:
@crusso, another aspect I alluded to yesterday was the fact that we can still allow-list a lot of `var` fields in records for inlining as long as a field doesn't participate in the record extension ever. This could be done as a static analysis on the lowered IR by forbid-listing the `RefD` accesses.

doc/md/language-manual.md
ggreif: Yes, that is the spirit! :-)

ggreif: You can avoid repeating type signatures by
``` Motoko
type Smaller = { nicefield : Nat, prettyfield : Char };
type Bigger = Smaller and { uglyfield : Text };
let big : Bigger = ...;
// then cast to Smaller:
{ (base : Smaller) and otherbase }
```

src/mo_frontend/typing.ml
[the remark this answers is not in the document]
crusso: Ah ok, its because you remember the types of the bases as object types, not list of fields, so need to merge them using glb. Ok.
The other option would be to just remember the list of fields taken from each object, flatten, append tfs, and sort.

src/mo_frontend/typing.ml
```diff
@@ -1048,11 +1051,49 @@ and infer_exp'' env exp : T.typ =
...
+                  "ambiguous field in base%a"
+                  display_lab lab;
+                info env h_exp.at "field also present in base, here (consider overwriting)"
+              end in
+          iter (fun (b_t, b) -> avoid_labels b (map (fun {T.lab; _} -> lab) (T.as_obj b_t |> snd))) t in
+        iter avoid (T.as_obj h |> snd);
+        disjoint t in
+    disjoint (map2 (fun b_t b -> b_t, b) stripped_bases exp_bases);
+
+    (* TODO: var in stripped_bases? *)
```
crusso, line 1094: Is this still a todo?

src/mo_frontend/typing.ml
```diff
@@ -1048,11 +1051,49 @@ and infer_exp'' env exp : T.typ =
...
+  | ObjE (exp_bases, exp_fields) ->
+    let open List in
     check_ids env "object" "field"
-      (List.map (fun (ef : exp_field) -> ef.it.id) exp_fields);
-    let fts = List.map (infer_exp_field env) exp_fields in
-    T.Obj (T.Object, List.sort T.compare_field fts)
+      (map (fun (ef : exp_field) -> ef.it.id) exp_fields);
+    let fts = map (infer_exp_field env) exp_fields in
+    let fls = map (fun {T.lab; _} -> lab) fts in
+    let bases = map (fun b -> infer_exp env b, b) exp_bases in
```
crusso, line 1060:
```suggestion
    let bases = map (fun b -> infer_exp_promote env b, b) exp_bases in
```
This should promote bases of varialbe types to their bounds, letting you merge a base expression whose type is generic, but bounded by a record. The typeing code for DotE does similar.

src/mo_frontend/typing.ml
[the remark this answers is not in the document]
ggreif: Right

src/mo_frontend/typing.ml
```diff
@@ -1419,7 +1460,7 @@ and check_exp' env0 t exp : T.typ =
   | TupE exps, T.Tup ts when List.length exps = List.length ts ->
     List.iter2 (check_exp env) ts exps;
     t
-  | ObjE exp_fields, T.Obj(T.Object, fts) ->
+  | ObjE ([], exp_fields), T.Obj(T.Object, fts) -> (* FIXME: bases? *)
```
crusso, line 1463: I guess you could do better here by inferring the non-empty bases but pushing the expected fields types into the fields, but that might duplicate a lot of code from the synthesis case above....

doc/md/language-manual.md
ggreif: this is expanded now

src/mo_frontend/typing.ml
ggreif: This improved things: a973e4220

doc/md/language-manual.md
```diff
@@ -1654,6 +1654,24 @@ Such an object literal, sometimes called a *record*, is equivalent to the object
 
 Object expressions support *punning* for concision. A punned field `<id>` is shorthand for `<id> = <id>`; Similarly, a typed, punned field `<id> : <typ>` is short-hand for `<id> = <id> : <typ>`. Both associate the field named `<id>` with the value of the identifier `<id>`.
 
+#### Object reuse and extension
+
+Objects can be combined and/or extended using the `and` and `with` keywords.
+
+A record expression `{ <exp_post> (and <exp_post>)* with <exp-field>;* }` merges the objects (or modules) specified as *bases* which can be
```
crusso, line 1661:
```suggestion
A record expression `{ <exp> (and <exp>)* (with <exp-field>;+)? }` merges the objects (or modules) specified as *bases* which can be
```
(For better or worse, the manual deals only with abstract syntax, not concrete syntax, so I'd skip the distinction between <exp_post> and <exp>).

doc/md/language-manual.md
```diff
@@ -1654,6 +1654,24 @@ Such an object literal, sometimes called a *record*, is equivalent to the object
...
+postfixed expressions, and augments the result to also contain the specified fields. The field list can be omitted when at least two bases appear and none have common field labels.
```
crusso, line 1662:
```suggestion

```
(For better or worse, the manual deals only with abstract syntax, not concrete syntax, so I'd skip the distinction between <exp_post> and <exp>).
```suggestion
A record expression `{ <exp> (and <exp>)* (with <exp-field>;+)? }` merges the objects (or modules) specified as *base* expressions, and augments the result to also contain the specified fields. The `with <exp-field>;+` clause can be omitted when at least two bases appear and none have common field labels.
```

doc/md/language-manual.md
```diff
@@ -1654,6 +1654,24 @@ Such an object literal, sometimes called a *record*, is equivalent to the object
...
+Thus the field list serves as
+- disambiguation for field labels occurring more than once in the bases,
+- supplying fresh fields,
+- changing field types, and
+- breaking up entanglement of `var` fields.
```
crusso, line 1667:
```suggestion
- breaking up entanglement of `var` fields.
```
```suggestion
- adding new `var` field or to avoid aliasing with an existing `var` field from some base.
```

src/mo_frontend/typing.ml
ggreif: Nope!

src/mo_frontend/typing.ml
ggreif: Yeah, I guess if I had a test case that potentially gave a better error message then I'd be inclined more going down this path, but right now it looks like a lot of sweat for little gain...

doc/md/language-manual.md
crusso: The `{ <exp> (and <exp>)* (with <exp-field>;+)? }` should also be added to the figure defining expression syntax earlier in the document. (Search for `<exp> ::=`  I guess).

It would be nice if you could describe the typing rule and evaluation rules for the expression, but I think that might be quite a mouthfull. Perhaps I could have a go later sometime.

src/mo_frontend/typing.ml
ggreif:
```suggestion
  | ObjE ([], exp_fields), T.Obj(T.Object, fts) -> (* TODO: infer bases? *)
```

crusso (commented):
Added some comments.

src/ir_passes/await.ml
```diff
@@ -429,6 +441,7 @@ and declare_dec dec exp : exp =
   match dec.it with
   | LetD (pat, _) -> declare_pat pat exp
   | VarD (id, typ, exp1) -> declare_id id (T.Mut typ) exp
+  | RefD (id, typ, exp1) -> declare_id id typ exp (* FIXME *)
```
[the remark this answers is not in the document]
ggreif: @claudio should we change to `_exp1` here and the line above?

test/repl/object-ext.sh
ggreif: I'll leave it for now, it is somewhat informative.

src/mo_frontend/typing.ml
```diff
@@ -1048,11 +1051,48 @@ and infer_exp'' env exp : T.typ =
...
-    T.Obj (T.Object, List.sort T.compare_field fts)
+      (map (fun (ef : exp_field) -> ef.it.id) exp_fields);
+    let fts = map (infer_exp_field env) exp_fields in
+    let fls = map (fun {T.lab; _} -> lab) fts in
+    let bases = map (fun b -> infer_exp_promote env b, b) exp_bases in
+
+    (* removing explicit fields from the bases *)
+    let strip (base_t, base) =
+      let s, tfs =
+        try T.as_obj base_t with Invalid_argument _ ->
```
crusso, line 1065: I think this is fine and there is no need use as_obj_sub - there is no minimal type to infer is some base has type 'None' and you should just error.

test/fail/ok/object-ext-pseudo.tc.ok
```diff
@@ -0,0 +1,2 @@
+object-ext-pseudo.mo:1.16-1.27: type error [M0093], expected object type, but expression produces type
```
ggreif, line 1: I am expecting 2 error messages!

maybe I am calling the wrong error reporting function!?

ggreif: I'm inclined to experiment with this in a separate PR.

src/ir_passes/await.ml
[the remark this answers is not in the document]
ggreif: Done!

src/mo_frontend/typing.ml
ggreif: I am getting the error (#3405) for `<A <: { a : Nat }>(base : A)`, I don't think we have a `None` in this game...

test/fail/ok/object-ext-pseudo.tc.ok
crusso: It could be local_error vs error (IIRC)

crusso: It could be because the error isn't an an analysis context, but in inference, so it can't recover and proceed. If you add a type annotation, do you get more errors?

rossberg:
> All non-overwritten var fields from bases will show up in the resulting object as var fields that alias the original field from its base.

Is this a good idea? AFAICS, this essentially makes the unfortunate extra indirection in the current implementation observable for var fields. So it's painting us into a corner with a costly representation that the compiler should rather try to get rid of.

ggreif:
> Is this a good idea? AFAICS, this essentially makes the unfortunate extra indirection in the current implementation observable for var fields. So it's painting us into a corner with a costly representation that the compiler should rather try to get rid of.

@rossberg I am discussing this with @crusso since Monday. See my response(s) beginning with https://github.com/dfinity/motoko/pull/3084#issuecomment-1216508340.

The bottom line is that our compilation strategy favours heavily the mutcell approach. Object methods right now don't capture the homing object, but the `var`-cells directly. This allows to migrate methods seamlessly from one object to another and GC away the old object. (Otherwise you'd get space leaks, retaining old objects.)

We still not lose the ability to inline `var` fields (in cases where not captured by methods or not participating as base in functional updates). We have to do an IR-level usage analysis and allow-list those fields that are eligible.

crusso:
> > All non-overwritten var fields from bases will show up in the resulting object as var fields that alias the original field from its base.
> 
> Is this a good idea? AFAICS, this essentially makes the unfortunate extra indirection in the current implementation observable for var fields. So it's painting us into a corner with a costly representation that the compiler should rather try to get rid of.
> 
> (At that point you could just as well introduce a first-class `var` type, ML-style. The whole point of the more restrictive design in Motoko was to be able to avoid the runtime costs that come with that.)

Yeah, I was rather worried about that too, but, even with record combination, the optimization is currently impeded by the way we stabilize mutable data. I don't think Joachim's trick that we use to preserve sharing of invariant mutable data during stabilzation works when you start inlining mutable fields into the enclosing object. You'd have to share the entire object, but because some fields may be immutable and allow variance at the type level, I don't think you could just serialize the first value you come across and reuse it at different occurrences, because not all aliases will be viewed at the same type.

crusso:
To avoid painting ourselve into a corner, what we could consider doing is restricting the operation to only take immutable fields from the base objects and allow the addition of new mutable fields in the `with` part, but that might be too restrictive.

src/lowering/desugar.ml
```diff
@@ -581,18 +581,20 @@ and obj obj_typ efs bases =
...
+      else
+        let id = fresh_var lab typ in
+        let [@warning "-8"] [base_var] = concat_map ((|>) lab) pickers in
+        let d =
+          if T.is_mut typ then
+            refD id { it = I.DotLE(varE base_var, lab); note = typ; at =  no_region }
+          else
+            letD id (dotE (varE base_var) lab typ) in
+        let f = { it = I.{ name = lab; var = id_of_var id }; at = no_region; note = typ } in
+        [d, f] in
```
crusso, line 597: You'll still need the type fields fs' to contains both type fields and value fields so I think you need to refine this a bit.
Maybe don't use map but a loop that accummulates the decs and type level fields separately, with different lengths.

ggreif: I guess you look ate a _late IR_, as directly after desugaring I see
``` Motoko
  (LetD
    (VarP ac)
    (BlockE
      (LetD (VarP $base/0) (VarE a))
      (LetD (VarP $base/1) (VarE c))
      (LetD (VarP $x/0) (PrimE (DotPrim x) (VarE $base/0)))
      (LetD (VarP $y/0) (PrimE (DotPrim y) (VarE $base/1)))
      (NewObjE
        Object
        (y $y/0)
        (x $x/0)
        {type T = Int; type X = Int; x : Int; y : () -> X}
      )
    )
  )
```

crusso: Ah not, I misunderstood your code and thought fs were the field types, not the field values. Sorry. Looks good, I think.

rossberg:
@ggreif:
> The bottom line is that our compilation strategy favours heavily the mutcell approach. Object methods right now don't capture the homing object, but the var-cells directly. This allows to migrate methods seamlessly from one object to another and GC away the old object. (Otherwise you'd get space leaks, retaining old objects.)

Yeah, but so far it's an implementation detail that could (and eventually should) be changed. I wouldn't count it as a space leak when an extracted method keeps an object alive. That's what most people would probably expect anyway (and as soon as the method is remotely interesting it is not unlikely to call another method, which keeps more of the objects alive anyway – I think the ideal case you have in mind is not particularly realistic in practice).

> We still not lose the ability to inline var fields (in cases where not captured by methods or not participating as base in functional updates).

I don't think you'd have much of a chance, since you wouldn't know locally whether a downstream function aliases a field. You'd need very sophisticated whole program flow analysis for that.

@crusso:
> I don't think Joachim's trick that we use to preserve sharing of invariant mutable data during stabilzation works when you start inlining mutable fields into the enclosing object.

Probably not, or at least not as easily. But that's another implementation detail that I wouldn't take for granted.

So, I agree that it is probably best to be conservative, i.e., not allow var for the time being, and at least get some user experience with a more limited version before deciding to go wild. It might not generally be a good idea to use functional update on mutable records anyway – either somebody designs their data structure as imperative and updates in place, or as immutable and updates by copying. Mixing the two qualifies as a code smell, probably begging for subtle bugs.

FWIW, I'd also pick a syntax that's more consistent with the type level (with no braces around). Why not simply a (binary) `with` operator on records, that overrides right to left, similar to what Scala does? Having both multiple `and` and `with` in a single construct seems a bit baroque to me...

crusso:
The other alternative to forbidding var fields is to just adopt the Ocaml semantics and do a shallow copy of var fields. That drops the aliasing and doesn't paint us into the corner. I.e. expect both the aliasing and non-aliasing semantics have their uses and abuses. The other advantage of this is that one can explain the construct as type directed syntactic sugar (which can't explain the aliasing).

Eventually, I'd also like to introduce  a first class SML style ref type so that users can efficiently express first class cells with aliasing, without the silly overhead of an additional object wrapper plus field lookup - though I guess we could try to optimise to get the same result.

Rant: I regret our simple minded encoding of objects as records of values. I wish we'd been more daring and just gone for something more codata like: a collection of functional observations of some encapsulated state. I expect that might have yielded a more efficient encoding where all instances of the same class have an environment and a shared, single pointer to the same vector of behaviours, rather than representing each instance as a bloated vector of values and closures.

crusso:
FTR, Joachim made a very interesting observation about removing the field indirection and upgrades:

"Because the candid format allows different records to have aliased fields, and we need to decode those, we cannot stop supporting aliased fields or upgrades can break."

Since stable variable serialization can't encode aliasing of records, it aliases only their mutable fields and deserialization will need to preserve this aliasing to preserve the semantics.

ggreif:
> It's strange that something peculiar like aliasing copying exists only as part of a “large” feature. It seems better if this new syntax is really just sugar.

I don't see why we can't fix this in the future, e.g. by adding reference assignments `foo.mutfield := var bar.mutfield`.

crusso:
Ok, I propose the following compromise.

We allow bases to have mutable fields but require them to be overriden in the with clause, with no support for aliasing.

If users complain that this is not enough, we can relax the requirement and introduce the aliasing semantics (or even implicit shallow copy, if that's what users actually expect).

I'm fine with the syntax as is - it's a bit late to complain about it.

crusso:
> Ok, I propose the following compromise.
> 
> We allow bases to have mutable fields but require them to be overridden in the with clause, with no support for aliasing.
> 
> If users complain that this is not enough, we can relax the requirement and introduce the aliasing semantics (or even implicit shallow copy, if that's what users actually expect). Any future relaxation would not be a breaking change.
> 
> I'm fine with the syntax as is - it's a bit late to complain about it.
> 
> @rossberg @nomeata sound reasonable?

@nomeata and I discussed this:

me: Gabor has implemented my suggestion of requiring users to override inherited var fields, preventing aliasing (in a way that we can relax later, sans breaking anyone). For or against? Indifferent?
nomeata: For. It means the syntax is just simple sugar around a plain record construction, with the inherited fields just copying values out of the appropriate parent, right?
me: Yep, but you need to be explicit about what to do with var fields, the only option being redefining them (or forgetting them via prior annotation)
nometa: Right

doc/md/language-manual.md
```diff
@@ -1654,6 +1654,23 @@ Such an object literal, sometimes called a *record*, is equivalent to the object
...
+- disambiguation for field labels occurring more than once in the bases,
+- supplying fresh fields,
+- changing field types, and
+- adding new `var` field or to avoid aliasing with an existing `var` field from some base.
+
+The resulting type is determined by the bases' (and explicitly given fields') static type.
+
+`var` fields from the bases (when not disambiguated) will show up in the result as `var` fields that *share* the original contents,
+i.e. an assignment to such a field of either the base or the result will change the shared content. If such a behaviour is not desired,
+then the field must be explicitly overwritten in the explicit field list.
```
crusso, line 1672: Any `var` field from some base must be overwritten in the explicit field list, to avoid introducing aliasing of var fields.

src/ir_def/ir.ml
```diff
@@ -181,6 +181,7 @@ and dec = dec' Source.phrase
 and dec' =
   | LetD of pat * exp                          (* immutable *)
   | VarD of id * Type.typ * exp                (* mutable *)
+  | RefD of id * Type.typ * lexp               (* reference *)
```
crusso, line 184:
```suggestion
  | RefD of id * Type.typ * lexp               (* reference - only required for flag --experimental_field_aliasing *)
```

test/run/object-extend.mo
```diff
@@ -0,0 +1,71 @@
...
+func mix<A <: A0, B <: B0>(a : A, b : B) : A0 and B0 and { c : Text } =
+    { a and b with c = "Waay to go!" };
+
+func mox<A <: { a : Int }, B <: { b : Char }>(a : A, b : B) : { a : Int } and { b : Char } and { c : Text } =
+    { a and b with c = "Right" };
+
+func mux<A <: { a : Int }, B <: { b : Char }>(a : A, b : B) : { a : Int; b : Char; c : Text } =
+    { a and b with c = "Yeah" }
+*/
+
```
crusso, line 68: See #3417 for a fix

crusso (commented):
See doc changes and #3417

doc/md/language-manual.md
crusso, line 1672:
```suggestion
Any `var` field from some base must be overwritten in the explicit field list. This avoids introducing aliasing of `var` fields.
```

src/mo_frontend/typing.ml
crusso: Should be fixed now

crusso (approved):
But don't forget my suggestion to the doc. You can also put the other wording in an html comment for preservation.

test/run/object-extend.mo
ggreif: Merged!

rossberg:
The restriction makes sense to me.

Still curious why you had to make this construct so complex, though. What can it express that a simple binary one can not?

ggreif:
> Still curious why you had to make this construct so complex, though. What can it express that a simple binary one can not?

@rossberg a few reasons:
- type-level `and` is unbiased, it is nice to have an unbiased value-level construct too
- a binary `with` would have different performance characteristics when applied to three or more bases (depending on choice of association and size of the bases)
- iterated binary merges would also create unnecessary garbage in heap

crusso:
> The restriction makes sense to me.
> 
> Still curious why you had to make this construct so complex, though. What can it express that a simple binary one can not?

@rossberg Isn't the simplest example one where there are two fields that clash and you want to take the first from the left, but the second from the right record? Though I guess you could forget one field using subtyping before doing the merge (or, perhaps better, add a `\ x` operator for removing fields by name).

I agree the complexity is a bit high, especially when you try to describe it in prose.... The binary version isn't that much better in the sense that it still breaks subject reduction and can (equally) only be described as a type based elaboration.

rossberg:
@crusso:
> Isn't the simplest example one where there are two fields that clash and you want to take the first from the left, but the second from the right record? 

True, although you can achieve that with a type constraint or more orthogonal means, as you say. In general, bidirectional merging is only practically useful if you have actual mixins (records with abstract fields), short of that, it may not justify the extra complexity.

@ggreif, the performance and garbage arguments are a bit unconvincing. It should be fairly easy for a compiler to syntactically recognise the cases that correspond to the complex construct and compile them the same way.

As for syntax and symmetry, I'd have thought that the obvious incoherence between `A and B` on the type level and `{A and B}` on the term level would be more concerning. ;)

nomeata:
I am confused to still see IR changes here. I thought the final design did not introduce aliasing of mutable fields? I understand that _maybe_ we’ll get some form of aliasing in the future, but until then it’s dead code and hence dangerous. (unless I am missing something).

Maybe revert it in a separate PR, where we can easily take it from if we need it again?

ggreif:
@nomeata The aliasing feature is still available, but behind an opt-in flag `--experimental-field-aliasing`, which we don't disclose in the changelog. It is being tested.

PR #2998, opened 2021-12-15
Gabor/tag on alloc

ggreif:

PR #3085, opened 2022-01-29
perf: optimize overflow check in RTS `alloc_words`, fix and update tests

crusso:
#3077 added an expensive, but necessary overflow check to `alloc_words`, to prevent allocating addresses outside the 32-bit heap.

* This PR optimizes that check to avoid the conditional test by doing the page diff arithmetic in u64s, 
relying on `grow_memory` to trap when asked to grow beyond 2^16 pages.
* fixes test `run-drun/oom-update` to actually use an update (not query) method as advertised.

I wonder why `grow_memory` is marked `inline_never`? Seems like we'd save a function call on each allocation...

### DIFF
--- a/rts/motoko-rts/src/memory/ic.rs
+++ b/rts/motoko-rts/src/memory/ic.rs
@@ -95,7 +95,7 @@ unsafe fn grow_memory(ptr: u64) {
     debug_assert!(ptr <= 2 * u64::from(core::u32::MAX));
     let page_size = u64::from(WASM_PAGE_SIZE.as_u32());
     let total_pages_needed = ((ptr + page_size - 1) / page_size) as usize;
-    let current_pages = wasm32::memory_size(0) as usize;
+    let current_pages = wasm32::memory_size(0);
     if total_pages_needed > current_pages {
         if wasm32::memory_grow(0, total_pages_needed - current_pages) == core::usize::MAX {
             rts_trap_with("Cannot grow memory");
--- a/rts/motoko-rts/src/memory/ic.rs
+++ b/rts/motoko-rts/src/memory/ic.rs
@@ -79,10 +79,10 @@ impl Memory for IcMemory {
         let old_hp = u64::from(HP);
         let new_hp = old_hp + delta;
 
-
         // Grow memory if needed
         grow_memory(new_hp);
 
+        debug_assert!(new_hp <= u64::from(core::u32::MAX));
         HP = new_hp as u32;
 
         Value::from_ptr(old_hp as usize)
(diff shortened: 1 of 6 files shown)


crusso:
> Comparing from [5f678e9](https://github.com/dfinity/motoko/commit/5f678e9e3e6a996b448efe8de8c12cf5719a305b) to [f6d70fb](https://github.com/dfinity/motoko/commit/f6d70fb4a289e31be8aafbab4bb2b9c2eec539f4): In terms of gas, 3 tests improved and the mean change is -2.2%. In terms of size, 3 tests regressed and the mean change is +0.4%.

Seems worth it, maybe.

ggreif:
See also https://github.com/dfinity/motoko/pull/2998, where I was wondering about the _no-inline_ too...

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -72,29 +72,29 @@ impl Memory for IcMemory {
...
 
         Value::from_ptr(old_hp as usize)
     }
 }
 
 /// Page allocation. Ensures that the memory up to, but excluding, the given pointer is allocated.
 #[inline(never)]
-unsafe fn grow_memory(ptr: usize) {
+unsafe fn grow_memory(ptr: u64) {
+    debug_assert!(ptr <= 2 * u64::from(core::u32::MAX));
```
ulan, line 95: We can use the same condition as in the assert in line 85: `ptr <= u64::from(core::u32::MAX)` (i.e. compare against 4GiB, not 8GiB)

ulan (approved):

ggreif (approved):
The logic didn't change, only the u64 conversions moved a bit, and the trapping is consolidated. LGTM!

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -72,29 +72,29 @@ impl Memory for IcMemory {
...
-        if overflow {
-            rts_trap_with("Out of memory");
-        }
-        HP = new_hp;
+        let old_hp = u64::from(HP);
+        let new_hp = old_hp + delta;
 
         // Grow memory if needed
-        grow_memory(new_hp as usize);
+        grow_memory(new_hp);
```
ggreif, line 83: @ulan commenting remark in the meeting just now. The fast path should be predicated on the fact that the allocation remains on the last page, i.e. `old_hp >> 16 == new_hp >> 16`.

rts/motoko-rts/src/memory/ic.rs
crusso: No, I'm deliberately comparing against 8GB because I want to test the overflow detection later on (otherwise we will get different faults on debug and release builds).

rts/motoko-rts/src/memory/ic.rs
crusso: Right, but I'lll leave that to #3090.

PR #3086, opened 2022-01-29
feat: add prim call_raw : (Principal, Text, Blob) -> async Blob 

### LINKED ISSUE #2703: Raw Calls - What is necessary?
I've recently been frustrated by Motoko's inability to do raw calls as seen in the cycle wallet(https://github.com/dfinity/cycles-wallet/blob/5725370020593bd744ddbc589c117d76d6ca8ae4/wallet/src/lib.rs#L625).

To be able to do this I would imagine motoko would need to:

- expose the raw call
- enable some kind of type reflection
- formalize some kind of serialization/deserialization

I've seen some mention of these in some of the open/closed meta-issue issues here and I thought I'd stand on top of the building yelling to hopefully get these some attention. :)

What I don't want to do is to have to create a function on every motoko pass through canister called __forwardCall(principal: Principal; method: text, args: [Nat8]/WorkSpace/[DataZones]) and have to have every canister handle a __handleForwardCall that deserializs all the parameters manually into known types and organize the community to universally add these to their canisters.  perhaps there is a compiler way to automate this kind of thing?  Seems overkill when the underlying functions can just be exposed some how.

crusso:
Fixes #2703 (by lowering ourselves to the level of Rust).

Adds a prim to dynamically invoke a method by name with an already serialized blob and get the raw (undeserialized) blob back asynchronously:

``` Motoko
call_raw : (canister : Principal, function_name : Text, arg : Blob) -> async Blob
````

The function can only be called in an asynchronous context and this is enforced by the type system.

There is no assumption that the contents of either blob is Candid, so this could also be used to talk to non-Candid endpoints.

This should be sufficient to implement the call-forwarding functionality of the Rust cycles wallet.

- [x] Determine whether the method name must be (rope)-normalized before use. Currently it is not. @nomeata,  what's the representation invariant for ordinary shared functions - I see they are (Principal,Text) pairs, but is the Text normalized? 
- [ ] Any ideas for better name: `request`, `send`, `call`, `invoke`, `call_dynamic` spring to mind

### CODE REVIEW

File: src/codegen/compile.ml [hunk 1]
```diff
@@ -6362,8 +6362,22 @@ module FuncDec = struct
      deserialization); the reject callback function is unique.
   *)
 
-  let closures_to_reply_reject_callbacks env ts =
-    let reply_name = "@callback<" ^ Typ_hash.typ_hash (Type.Tup ts) ^ ">" in
+  let closures_to_reply_reject_callbacks_aux env ts_opt =
+    let arity, reply_name, from_arg_data =
```
crusso, line 6366: Make these parameters, don't switch on ts_opt.

src/prelude/prim.mo
```diff
@@ -357,3 +357,5 @@ func stableMemoryLoadBlob(offset : Nat64, size : Nat) : Blob =
 
 func stableMemoryStoreBlob(offset : Nat64, val :  Blob) : () =
   (prim "stableMemoryStoreBlob" : (Nat64, Blob) -> ()) (offset, val);
+
+let call_raw = @call_raw;
```
crusso, line 361:
```suggestion
let call_raw = @call_raw;

```

src/codegen/compile.ml
crusso: Actually, I now prefer the original.

nomeata:
> Determine whether the method name must be (rope)-normalized before use. Currently it is not. @nomeata, what's the representation invariant for ordinary shared functions - I see they are (Principal,Text) pairs, but is the Text normalized?

The new `call_raw` takes a `Text`, and you certainly need to make that contiguous before passing it to the system API. I'm not sure how shared functions representation play a role here?

nomeata (approved):
Looks reasonable

nomeata:
Ah, I see why, because you construct such a function ref pair in the backend.

I don't think it's clearly defined what the representation is yet, presumably, so far only flat text values occurred. So the value needs to flattened at the right place…

src/codegen/compile.ml
```diff
@@ -8234,7 +8261,21 @@ and compile_exp (env : E.t) ae exp =
...
+    | ICCallRawPrim, [p;m;a;k;r] ->
+      SR.unit, begin
+      let (set_meth_pair, get_meth_pair) = new_local env "meth_pair" in
+      let (set_arg, get_arg) = new_local env "arg" in
+      let (set_k, get_k) = new_local env "k" in
+      let (set_r, get_r) = new_local env "r" in
+      let add_cycles = Internals.add_cycles env ae in
+      compile_exp_vanilla env ae p ^^
+      compile_exp_vanilla env ae m ^^
+      Tuple.from_stack env 2 ^^ set_meth_pair ^^
```
nomeata, line 8273: Here you need to turn the text into a blob, because above we have
```
(* The method name *)
      get_meth_pair ^^ Arr.load_field 1l ^^ Blob.as_ptr_len env ^^
```

crusso: Excellent, thanks. That's precisely what I was worried about.

crusso: Fixed!

test/run-drun/call-raw.mo
```diff
@@ -0,0 +1,110 @@
...
+        let res = await P.call_raw(p,"trapInt", arg);
+        assert false;
+      }
+      catch e {
+        P.debugPrint(P.errorMessage(e));
+      }
+    };
+
+    do {
+      let m = "super"#"cali"#"fragilisticexpialidocious";
```
ggreif, line 93: Just being curious, did you see this crashing before it started working?

src/prelude/internals.mo
```diff
@@ -406,3 +406,8 @@ func @create_actor_helper(wasm_module_ : Blob, arg_ : Blob) : async Principal =
   });
   return canister_id_;
 };
+
+// raw calls
+func @call_raw(p : Principal, m : Text, a : Blob) : async Blob {
```
ggreif, line 411: Why to define this in `prim.mo`?

src/ir_def/check_ir.ml
```diff
@@ -593,6 +593,14 @@ let rec check_exp env (exp:Ir.exp) : unit =
          error env exp1.at "expected function type, but expression produces type\n  %s"
            (T.string_of_typ_expand t1)
       end
+      (* TODO: T.unit <: t ? *)
```
ggreif, line 596: still relevant?

src/ir_def/check_ir.ml
```diff
@@ -593,6 +593,14 @@ let rec check_exp env (exp:Ir.exp) : unit =
...
+    | ICCallRawPrim, [exp1; exp2; exp3; k; r] ->
+      typ exp1 <: T.principal;
+      typ exp2 <: T.text;
```
ggreif, line 599: why not make this `T.blob` and normalise `Text` -> `Blob` in Motoko?

rossberg:
> There is no assumption that the contents of either blob is Candid, so this could also be used to talk to non-Candid endpoints.

For the record, I'm really concerned about the slippery slope this creates towards leaking and breaking more abstractions.

skilesare:
Maybe there is a way to close the abstraction later? This is exactly what we need right now to develop some things on our road map using motoko that have been blocked since launch. Thank you!

test/run-drun/call-raw.mo
crusso: yes, I tried it locally before fixing

src/prelude/internals.mo
crusso: I tried, but our checker complains about await's in prim, so I just moved it here instead. Sneaky.

src/ir_def/check_ir.ml
crusso: Yeah, I just noticed the missing check

crusso:
> > There is no assumption that the contents of either blob is Candid, so this could also be used to talk to non-Candid endpoints.
> 
> For the record, I'm really concerned about the slippery slope this creates towards leaking and breaking more abstractions.


Objection noted. 

However, we had promised a solution for ages and never delivered one, so offering this as an experimental feature while we come up with something better seems reasonable to me.  Given that you can already achieve this (modulo a change in caller) by bouncing off Rust, I don't really see it compromising more than we have already across the company.

PR #3090, opened 2022-01-31
perf: allocator fast path

ggreif:
This introduces a XOR trick to detect whether the recent allocation stays on the same (already) allocated page, and only runs the expensive `grow_memory` route if a page boundary is crossed. It also reserves the last page as unallocated for essentials (like a potential future emergency allocation of the stable variable streaming buffer). The unallocated page can also serve as a guard page against stack overflow of the C/Rust RTS stack.

To make the trick work, a slight change in page allocation is necessary. When the new `HP` points to exactly the start of the page (i.e. has form `0xNNNN0000` for some `NNNN`) we have to eagerly allocate the page corresponding to all `0xNNNNxxxx`, because for subsequent in-page allocations the fast-path is in effect, and assume that the page is already allocated. Notably this rule _doesn't_ extend to new `HP = 0xFFFF0000`, as the last page is taboo.

### TODO
- [ ] make a test that successfully allocates till `0xFFFF0000` — dubious if this can be accomplished without a new primitive
- [ ] make a test that fails allocating till `0xFFFF0004`
- [x] consider `/ WASM_PAGE_SIZE` instead of `>> 16`
- [ ] try to increase cycles to make `oom-query.mo` work again


rts/motoko-rts/src/memory/ic.rs
```diff
@@ -93,8 +97,8 @@ impl Memory for IcMemory {
 #[inline(never)]
```
ggreif, line 97: The above comment needs revision!

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -93,8 +97,8 @@ impl Memory for IcMemory {
...
 unsafe fn grow_memory(ptr: u64) {
     debug_assert!(ptr <= 2 * u64::from(core::u32::MAX));
-    let page_size = u64::from(WASM_PAGE_SIZE.as_u32());
-    let total_pages_needed = ((ptr + page_size - 1) / page_size) as usize;
+    debug_assert_eq!(2 * u64::from(WASM_PAGE_SIZE.as_u32()) >> 16, 2);
+    let total_pages_needed = (ptr >> 16) as usize + 1;
```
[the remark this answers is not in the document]
ggreif:
```suggestion
    let total_pages_needed = ((ptr >> 16) as usize + 1) & 0xFFFF;
```

rts/motoko-rts/src/memory/ic.rs
ggreif: dcaedba3c!

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -41,6 +41,8 @@ unsafe extern "C" fn init(align: bool) {
         get_heap_base()
     };
     LAST_HP = HP;
+    // make sure that the page HP points to us allocated too
```
ggreif, line 44:
```suggestion
    // make sure that the page HP points to is allocated too
```

crusso:
Might be good to review and merge https://github.com/dfinity/motoko/pull/3101 first so we have a few more tests in run-drun-non-ci.

test/run-drun/ok/oom-query.drun-run.ok
```diff
@@ -1,3 +1,3 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
-Err: IC0503: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped explicitly: RTS error: Cannot grow memory
+Err: IC0502: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped: heap or table out of bounds
```
crusso, line 3: Wait, why is this message changing? That suggests the overflow detection isn't kicking in and we are trapping due to corruption.

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -89,12 +93,19 @@ impl Memory for IcMemory {
...
+/// with the slight exception of not allocating the extra page for 0x100_000_000. For all
+/// `ptr > 0x100_000_000` the call to `wasm32::memory_grow` will trap.
+/// Invariant: ptr must be a reasonable pointer (i.e. less than 0x200_000_000).
 #[inline(never)]
 unsafe fn grow_memory(ptr: u64) {
-    debug_assert!(ptr <= 2 * u64::from(core::u32::MAX));
-    let page_size = u64::from(WASM_PAGE_SIZE.as_u32());
-    let total_pages_needed = ((ptr + page_size - 1) / page_size) as usize;
+    debug_assert!(ptr < 2 * u64::from(core::u32::MAX));
+    debug_assert_eq!(2 * u64::from(WASM_PAGE_SIZE.as_u32()) >> 16, 2);
```
crusso, line 103:
```suggestion
    debug_assert_eq!(u64::from(WASM_PAGE_SIZE.as_u32()) >> 16, 1);
```
Also  - why isn't this just a static assert - I don't think it depends on anything dynamic

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -89,12 +93,19 @@ impl Memory for IcMemory {
     }
 }
 
-/// Page allocation. Ensures that the memory up to, but excluding, the given pointer is allocated.
+/// Page allocation. Ensures that the memory up to, and including, the given pointer is allocated,
+/// with the slight exception of not allocating the extra page for 0x100_000_000. For all
+/// `ptr > 0x100_000_000` the call to `wasm32::memory_grow` will trap.
+/// Invariant: ptr must be a reasonable pointer (i.e. less than 0x200_000_000).
```
crusso, line 99:
```suggestion
/// Page allocation. Ensures that the memory up to, and including, the given pointer is allocated,
/// with the slight exception of not allocating the extra page for 0x1_0000_0000. For all
/// `ptr > 0x100_000_000` the call to `wasm32::memory_grow` will trap.
/// Invariant: ptr must be a reasonable pointer (i.e. less than 0x2_0000_0000).
```
Can you explain why this has to change? I'm a bit confused.

ggreif: I think this would be a good place to exclude the last page from being allocated, i.e. 0xFFFF_0000 should be the limit address for `HP`. Then we have a last-resort headroom for e.g. serialisation buffer (stable variables; caveat: do bignums allocate beyond that?)

rts/motoko-rts/src/memory/ic.rs
ggreif: Probably.

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -75,21 +77,30 @@ impl Memory for IcMemory {
...
+    debug_assert!(ptr < 2 * u64::from(core::u32::MAX));
+    debug_assert_eq!(2 * u64::from(WASM_PAGE_SIZE.as_u32()) >> 16, 2);
+    // Note: the `+ 1` below makes sure that there is a page allocated
+    //       for ptr (even when it points exactly to its start) and
+    //       the correction `ptr >> 32` is a branchless way to undo this
+    //       behaviour when ptr points beyond the first 4 GiB of memory.
+    let total_pages_needed = (ptr >> 16) as usize + 1 - (ptr >> 32) as usize;
     let current_pages = wasm32::memory_size(0);
     if total_pages_needed > current_pages {
         #[allow(clippy::collapsible_if)] // faster by 1% if not colapsed with &&
```
ggreif, line 106:
```suggestion
        #[allow(clippy::collapsible_if)] // faster by 1% if not collapsed with &&
```

rts/motoko-rts/src/memory/ic.rs
ggreif: This would also render #3782's stack canary potentially unnecessary. Or it could be reduced to only be present in `--check` builds.

test/run-drun/ok/oom-query.drun-run.ok
ggreif: These tests don't run on CI, and for some reason I can't run them locally. @luc-blaeser could you run them and accept the new output?

ggreif: Now I get
``` diff
--- oom-query.drun-run (expected)
+++ oom-query.drun-run (actual)
@@ -1,3 +1,3 @@
-Err: IC0502: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped: heap or table out of bounds
+Err: IC0522: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai exceeded the instruction limit for single message execution.
Some tests failed:
oom-query.mo
```

rts/motoko-rts/src/memory/ic.rs
ggreif: this code is different now

rts/motoko-rts/src/memory/ic.rs
ggreif: Last page is free restriction is introduced now.

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -75,26 +77,28 @@ impl Memory for IcMemory {
...
 #[inline(never)]
 unsafe fn grow_memory(ptr: u64) {
-    debug_assert!(ptr <= 2 * u64::from(core::u32::MAX));
-    let page_size = u64::from(WASM_PAGE_SIZE.as_u32());
-    let total_pages_needed = ((ptr + page_size - 1) / page_size) as usize;
+    if ptr > 0xFFFF_0000 {
+        // spare the last wasm memory page
+        rts_trap_with("Cannot grow memory")
+    };
+    let total_pages_needed = (ptr >> 16) as usize + (ptr < 0xFFFF_0000) as usize;
```
crusso, line 99:
```suggestion
    let total_pages_needed = (ptr >> 16) as usize + (ptr < 0xFFFF_0000) as usize;
```
Why the conditional +1? 


Can we add a debug_assert that the memory_size is < 65536?

.github/workflows/test.yml
```diff
@@ -49,13 +49,13 @@ jobs:
     - name: Read performance delta
```
crusso, line 49: Why is the yml changing in this PR?

test/run-drun-non-ci/ok/oom-query.drun-run.ok
```diff
@@ -1,3 +1,3 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
-Err: IC0503: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped explicitly: RTS error: Cannot grow memory
+Err: IC0502: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped: heap or table out of bounds
```
crusso, line 3: This is still worrying me.

rts/motoko-rts/src/memory/ic.rs
ggreif: The `+1` is explained in the comment above.

Yes, sure I can add an assert.

test/run-drun-non-ci/ok/oom-query.drun-run.ok
ggreif: Nope this now gives "cycles exhausted". I didn't update it yet.

ggreif: You prolly get it on `master` too.

.github/workflows/test.yml
ggreif: It is gone now, as it was already also on `master`.

test/run-drun-non-ci/ok/oom-query.drun-run.ok
ggreif: Looks like the replica doesn't give the same cycle budget to queries any more.

test/run-drun/ok/oom-query.drun-run.ok
crusso: Ok, I get that too now.
(on my PR branch)
Need to check master.

crusso:
This is growing on me, especially with the perf regression of #3788 which this should mitigate nicely!

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -75,26 +77,29 @@ impl Memory for IcMemory {
         let new_hp = old_hp + delta;
 
         // Grow memory if needed
-        grow_memory(new_hp);
+        if (old_hp ^ new_hp) >> 16 != 0 {
```
luc-blaeser, line 80: Just an idea (to be tested for performance):

Maybe, we could introduce a constant `PAGE_SIZE: usize = 65 * 1024;` and derive the check and the last page boundary from that. 

`if (new_hp - old_hp) / PAGE_SIZE > 0 {`

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -75,26 +77,29 @@ impl Memory for IcMemory {
...
-/// Page allocation. Ensures that the memory up to, but excluding, the given pointer is allocated.
+/// Page allocation. Ensures that the memory up to, and including, the given pointer is allocated,
+/// with the slight exception of not allocating the extra page for address 0xFFFF_0000.
+/// Enforced precondition: `ptr` must be a pointer less than or equal to 0xFFFF_0000.
 #[inline(never)]
 unsafe fn grow_memory(ptr: u64) {
-    debug_assert!(ptr <= 2 * u64::from(core::u32::MAX));
-    let page_size = u64::from(WASM_PAGE_SIZE.as_u32());
-    let total_pages_needed = ((ptr + page_size - 1) / page_size) as usize;
+    if ptr > 0xFFFF_0000 {
```
luc-blaeser, line 95: This could then become:
`if ptr > usize::MAX - PAGE_SIZE + 1`

luc-blaeser: (Most probably, the compiler will optimize all these constant calculations to the same as the bitmask.)

rts/motoko-rts/src/memory/ic.rs
luc-blaeser: (Most probably, the compiler will optimize the division to the bit shift.)

ggreif: See also `WASM_PAGE_SIZE` in line 4. I'll do this experiment.

ggreif:
``` rust
if (old_hp ^ new_hp) >> 16 != 0 {
```
already produces
``` wasm
  (func $alloc_words (type 8) (param i32) (result i32)
    (local i32 i64 i64)
    block  ;; label = @1
      global.get $__memory_base
      i32.const 71584
      i32.add
      i32.load
      local.tee 1
      i64.extend_i32_u
      local.tee 2
      local.get 0
      i32.const 2
      i32.shl
      i64.extend_i32_u
      i64.add
      local.tee 3
      local.get 2
      i64.xor
      i64.const 65536
      i64.lt_u
      br_if 0 (;@1;)
      local.get 3
      call $motoko_rts::memory::ic::grow_memory::hb0efe958cfedcb7b
    end
    global.get $__memory_base
    i32.const 71584
    i32.add
    local.get 3
    i64.store32
    local.get 1
    i32.const -1
    i32.add)
```
When having
``` rust
        if (old_hp ^ new_hp) / WASM_PAGE_SIZE != 0 {
```
I get
``` wasm
  (func $alloc_words (type 8) (param i32) (result i32)
    (local i32 i64 i64)
    block  ;; label = @1
      global.get $__memory_base
      i32.const 71584
      i32.add
      i32.load
      local.tee 1
      i64.extend_i32_u
      local.tee 2
      local.get 0
      i32.const 2
      i32.shl
      i64.extend_i32_u
      i64.add
      local.tee 3
      local.get 2
      i64.xor
      i64.const 65536
      i64.lt_u
      br_if 0 (;@1;)
      local.get 3
      call $motoko_rts::memory::ic::grow_memory::hb0efe958cfedcb7b
    end
    global.get $__memory_base
    i32.const 71584
    i32.add
    local.get 3
    i64.store32
    local.get 1
    i32.const -1
    i32.add)
```

test/run-drun-non-ci/ok/oom-query.drun-run.ok
ggreif: I am still seeing the problem: 0bd3f620d. I'd like to see the "heap out of bounds", but don't have an idea how to get there.

rts/motoko-rts/src/memory/ic.rs
ggreif: See 8a943f358. I prefer this for now.

test/run-drun/ok/oom-query.drun-run.ok
ggreif: Okay, I think I have a solution for this. I'll introduce a `prim` to advance the `HP` to a certain address, marking the heap segment as `TAG_FREE_SPACE`. This is super cheap, and gives us the tool to increase heap pressure. No need for expensive 3 GB arrays any more.
What do you think @luc-blaeser @crusso?
I'll do a separate PR for this.

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -38,6 +38,8 @@ unsafe extern "C" fn init(align: bool) {
         get_heap_base()
     };
     LAST_HP = HP;
+    // make sure that the page HP points to is allocated too
+    grow_memory(u64::from(HP))
```
crusso, line 42: I wonder if we could avoid the need for this pre-allocation by using a separate global, `last_grow_memory_ptr`, as a cache of the last `ptr` passed to grow_memory. perhaps initialized to 0xFFFF_0000.
and then replacing the test to grow at line 80 with (something like)

```
if (last_grow_memory(ptr) ^ new_hp) / WASM_PAGE_SIZE.as_u32() as u64 != 0 {
            // Page boundary crossed, make sure that pages are allocated
            grow_memory(new_hp);
            last_grow_memory(ptr) = new_hp;
        }
```
Or would that allocate pages too late?

rts/motoko-rts/src/memory/ic.rs
crusso: Except luc was asking for a clearer test. sans XOR, I think:

`if (new_hp - old_hp) / PAGE_SIZE > 0 {`

rts/motoko-rts/src/memory/ic.rs
ggreif: But this line is almost never visited. I don't think we have to optimise this.

test/run-drun/ok/oom-query.drun-run.ok
ggreif: But that wouldn't permit precise adjustment of the `HP`, which is what the first two TODO items ask for :-(

rts/motoko-rts/src/memory/ic.rs
ggreif: Except that wouldn't work :-) E.g. `old_hp == 0xA_FFF0` and `new_hp == 0xB_0008`, the `new_hp - old_hp` is then `0x18` and if we divide by 0x1_0000` we get 0.

luc-blaeser: Oh, yes, true, sorry for the confusion.
It would be `if (new_hp / PAGE_SIZE > old_hp / PAGE_SIZE)`
But this is maybe slower?

ggreif: I am pretty sure that is slower :-)

rts/motoko-rts/src/memory/ic.rs
crusso: It wasn't so much to optimize it but to avoid using the heap_ptr itself as the cach and to avoid preallocating the first page (if this works at all). No biggie though.

luc-blaeser:
I performed GC benchmark measurements:

## Performance Improvement

**11% faster allocations**

0.8% faster on GC benchmark (copying GC).

### Micro-Benchmark

Instantiation of a class object with a Nat field, instructions:

| Unoptimized   | Optimized | Speedup      |
|---------------|-----------|------------- |
| 90           | 80        | 1.125        |     

### GC Benchmark

GC benchmark, total instructions, average across all benchmark cases:

| GC           | Unoptimized  | Optimized   | Speedup      |
|------------- | ------------ | ----------- | ------------ |
| Copying      | 2.41e10      | 2.39e10     | 1.0083       |

rts/motoko-rts/src/memory/ic.rs
ggreif: It might be worth it. It needs to be initialised to 0x0000_FFFF and then a simple comparison against `new_hp` would suffice. Also the global would be monotonic and thus would avoid many calls to `grow_memory` when `HP` get decreased by GC. Thanks for the idea, I'll try this in a separate follow-up.

crusso: Yes,I was actually wondering if it would be even better to just compare against the high-water mark (didn't realise this suggestion might also give you that.)

test/run-drun-non-ci/ok/oom-query.drun-run.ok
ggreif: And for me the whole directory seems bit-rotten. It might be my machine but I see (on `master`!)
``` diff
make: Entering directory '/Users/ggreif/motoko/test/run-drun-non-ci'
../run.sh -d *.mo *.drun
inc-oom: [tc] [comp] [valid] [drun-run]
oom-query: [tc] [comp] [valid] [drun-run]
--- oom-query.drun-run (expected)
+++ oom-query.drun-run (actual)
@@ -1,3 +1,3 @@
-Err: IC0503: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped explicitly: RTS error: Cannot grow memory
+Err: IC0522: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai exceeded the instruction limit for single message execution.
oom-update: [tc] [comp] [valid] [drun-run]
preupgrade-oom: [tc] [comp] [valid] [drun-run]
upgrade-bug-immut: [tc] [comp] [valid] [drun-run]
--- upgrade-bug-immut.drun-run (expected)
+++ upgrade-bug-immut.drun-run (actual)
@@ -1,4 +1,5 @@
 debug.print: pre
-ingress Err: IC0503: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped explicitly: RTS error: Cannot grow memory
+debug.print: post
+ingress Err: IC0503: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped explicitly: deliberate trap
upgrade-bug: [tc] [comp] [valid] [drun-run]
--- upgrade-bug.drun-run (expected)
+++ upgrade-bug.drun-run (actual)
@@ -1,4 +1,5 @@
 debug.print: pre
-ingress Err: IC0503: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped explicitly: RTS error: Cannot grow memory
+debug.print: post
+ingress Err: IC0503: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped explicitly: deliberate trap
File *.drun does not exist.
Some tests failed:
oom-query.mo upgrade-bug-immut.mo upgrade-bug.mo *.drun
make: *** [Makefile:4: all] Error 1
make: Leaving directory '/Users/ggreif/motoko/test/run-drun-non-ci'
```

ggreif: So this part also needs some TLC in a follow-up!

rts/motoko-rts/src/memory/ic.rs
crusso: Err, why don't we just compare against memorysize × 65536 directly? No need for a cache unless expensive.

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -75,26 +77,31 @@ impl Memory for IcMemory {
...
+        rts_trap_with("Cannot grow memory")
+    };
+    let total_pages_needed = (ptr >> 16) as usize + (ptr < 0xFFFF_0000) as usize;
     let current_pages = wasm32::memory_size(0);
     if total_pages_needed > current_pages {
-        #[allow(clippy::collapsible_if)] // faster by 1% if not colapsed with &&
-        if wasm32::memory_grow(0, total_pages_needed - current_pages) == core::usize::MAX {
-            rts_trap_with("Cannot grow memory");
-        }
+        wasm32::memory_grow(0, total_pages_needed - current_pages);
```
crusso, line 104: actually, I'm not sure you should remove the test on the result of memory grow. What if fails because the replica decides to disallow all pages (as it does for older version of moc and may for later ones too). Then we should probably still error rather than wait for an access violation;

rts/motoko-rts/src/memory/ic.rs
crusso: FYI #3790 does the experiment of not pre-allocating the first page and testing against memory size instead of the xor trick.

test/bench/palindrome.mo
```diff
@@ -1,6 +1,6 @@
 // The 2 palindrome implementations from
 // "There and Back Again", by Olivier Danvy and Mayer Goldberg
-//
+//MOC-FLAG --force-gc
```
crusso, line 3: I think the heap usage is too small for two runs have much effect. In #3090 I add another bench test, alloc.mo, that does more allocation to reveal a slight win on the second run (but really very small)

crusso: Oh crap, I pushed to you branch by mistake. I think I must have fail to switch branch. Will fix.

crusso, line 3:
```suggestion
//
```

test/bench/palindrome.mo
```diff
@@ -83,3 +83,4 @@ actor Palindrome {
 }
 
 //CALL ingress go 0x4449444C0000
+//CALL ingress go 0x4449444C0000
```
crusso, line 86:
```suggestion
//CALL ingress go 0x4449444C0000
```

test/bench/palindrome.mo
crusso: Now fixed with my two commits below.

rts/motoko-rts/src/memory/ic.rs
ggreif, line 104:
```suggestion
        let got = wasm32::memory_grow(0, total_pages_needed - current_pages);
        if got == core::usize::MAX {
             rts_trap_with("Cannot grow memory");
         }
```

ggreif: Re-added this. But maybe we should change the line 99 message to "Cannot allocate memory" so that we have some differentiating hint when it happens.

test/bench/palindrome.mo
ggreif: I have killed the additional commits.

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -103,9 +103,9 @@ unsafe fn grow_memory(ptr: u64) {
     if total_pages_needed > current_pages {
         let got = wasm32::memory_grow(0, total_pages_needed - current_pages);
         if got == core::usize::MAX {
```
crusso, line 105:
```suggestion
        if wasm32::memory_grow(0, total_pages_needed - current_pages) == core::usize::MAX {
```
was nicer, IMO

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -75,26 +77,34 @@ impl Memory for IcMemory {
...
+/// Enforced precondition: `ptr` must be a pointer less than or equal to 0xFFFF_0000.
 #[inline(never)]
 unsafe fn grow_memory(ptr: u64) {
-    debug_assert!(ptr <= 2 * u64::from(core::u32::MAX));
-    let page_size = u64::from(WASM_PAGE_SIZE.as_u32());
-    let total_pages_needed = ((ptr + page_size - 1) / page_size) as usize;
+    debug_assert_eq!(0xFFFF_0000, usize::MAX - WASM_PAGE_SIZE.as_usize() + 1);
+    if ptr > 0xFFFF_0000 {
+        // spare the last wasm memory page
+        rts_trap_with("Cannot allocate memory")
```
crusso, line 99: Do we really want to distinguish the message from "Cannot grow memory" - It's kind of an implementation detail and might confuse people when they start seeing this when they formerly got that.

ggreif: Well, there are distinct root causes, so I think it is worth differentiating. But I am not dogmatic about it.

crusso (approved):
My PR will probably change it all again, but oh well...

PR #3091, opened 2022-01-31
feat: add `motoko:compiler` ICP custom section with version info

crusso:
* version is report as UTF8 release or, if unavailable, git revision, matching moc --version.
* section defaults to private.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,7 +4,7 @@
 
 * motoko (`moc`)
 
-  * Emit new ICP metadata custom section 'motoko:compiler' with compiler release or revision in UTF8 (e.g. "0.6.21"). Default is `icp:private`ax (#3091).
+  * Emit new ICP metadata custom section 'motoko:compiler' with compiler release or revision in UTF8 (e.g. "0.6.21"). Default is `icp:private` (#3091).
   * Generalized `import` supporting pattern matching and selective field imports (#3076).
   * Fix: insert critical overflow checks preventing rare heap corruptions
     in out-of-memory allocation and stable variable serialization (#3077).
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,7 @@
 
 * motoko (`moc`)
 
+  * Emit new ICP custom section 'motoko:compiler' with compiler release or revision in UTF8 (e.g. "0.6.21"). Default is private.
   * Generalized `import` supporting pattern matching and selective field imports (#3076).
   * Fix: insert critical overflow checks preventing rare heap corruptions
     in out-of-memory allocation and stable variable serialization (#3077).
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,7 +4,7 @@
 
 * motoko (`moc`)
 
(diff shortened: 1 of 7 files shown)


nomeata:
Why not using the “producer” custom section for that https://github.com/WebAssembly/tool-conventions/blob/main/ProducersSection.md

Changelog.md
```diff
@@ -4,6 +4,7 @@
 
 * motoko (`moc`)
 
+  * Emit new ICP custom section 'motoko:compiler' with compiler release or revision in UTF8 (e.g. "0.6.21"). Default is private (#3091).
```
ggreif, line 7: what do you mean by "private"?

crusso: ICP sections can be public or private. Private means only accessible from controller (but the system can still peek at it I assume). Otherwise people could search for canisters compiled with moc versions with known vulnerabilities.

crusso: https://github.com/dfinity/interface-spec/blob/master/spec/index.adoc#canister-information

crusso: Yeah, ok, `icp:private`

crusso:
> Why not using the “producer” custom section for that https://github.com/WebAssembly/tool-conventions/blob/main/ProducersSection.md

Hmm, that looks a bit complicated for what this is for - giving the IC a way to detect the compiler (in case we need to work-around nasty heap corruption bugs, for example). 

I guess the producers section can be made accessible to other canisters, even if we wanted to, given the way the public spec has gone (I always thought that was an ugly approach and we should have allow-listed instead).

nomeata:
But isn’t that what that section is for? And wouldn’t it be good to know also the involved other tools (e.g. linkers, rust compilers etc.)? But I don't feel strongly, reinventing the wheel is what we do :-)

Changelog.md
```diff
@@ -4,6 +4,7 @@
 
 * motoko (`moc`)
 
+  * Emit new ICP metadata custom section 'motoko:compiler' with compiler release or revision in UTF8 (e.g. "0.6.21"). Default is `icp:private`ax (#3091).
```
ggreif, line 7:
```suggestion
  * Emit new ICP metadata custom section 'motoko:compiler' with compiler release or revision in UTF8 (e.g. "0.6.21"). Default is `icp:private` (#3091).
```

ggreif (approved):
Looks good, and the version shows up:
```
$ llvm-objdump test/perf/_out/qr.wasm -p -s --section='icp:private motoko:compiler'

test/perf/_out/qr.wasm:	file format wasm
Program Header:
Version: 0x1
Contents of section icp:private motoko:compiler:
 0000 302e362e 32302d31 392d6735 34396466  0.6.20-19-g549df
 0010 34306631                             40f1
```

On `master` it is now
```
Contents of section icp:private motoko:compiler:
 0000 302e362e 3231                        0.6.21
```

ggreif:
> What is the -19- after the version number anyway?

Probably a dashed `git` commit.

chenyan-dfinity:
> Why not using the “producer” custom section for that https://github.com/WebAssembly/tool-conventions/blob/main/ProducersSection.md

We need the `icp:` prefix to expose the custom section. So `producer` section cannot be exposed in the HTTP endpoint.

PR #3108, opened 2022-02-08
test: oom-guard

crusso:
A simple test for triggering the oom guard (not for merging)


luc-blaeser:

PR #3109, opened 2022-02-09
bugfix: Issue 3103, confusing eponymous types from separate candid files

### LINKED ISSUE #3103: Namespace in candid imports
https://forum.dfinity.org/t/motoko-multiple-canister-imports-compile-error/10491

```
import B "canister:B”;
import C "canister:C”;
```

If B and C has the same type variable, one of them will be overwritten in the typing environment.

crusso:
Fixes #3103
- [x] repro
- [x] fix 

The problem was using global state to map used type definitions in a candid import to their Motoko translations, for subsequent inclusion as type fields in the imported Motoko actor type.

I find this inclusion of type fields problematic, as I expect a candid author to be able to rename/elide equivalent candid type definitions without breaking the compilation of Motoko clients, but that ship has sailed and users (Enzo's QR code generator) for one have started relying on accessing type components of imported actors, so I'm wary of removing this now.

That still leaves once question though: the actor only includes type components that are reference from the body, not the full candid type environment which would be easy to construct and perhaps expected by the user. Is there any reason for this subtlety or should I just clean up the code to add the full environment of the candid file to the actor's type fields (not just the referenced types). I don't think this will break many client, and might be more natural. The only breakage I can envision is if some unreferenced Candid type definition involved Motoko-unsupported float32 types that would now fail to import as Motoko type fields.

### DIFF
--- a/src/mo_idl/idl_to_mo.ml
+++ b/src/mo_idl/idl_to_mo.ml
@@ -97,6 +97,7 @@ and check_meth env (m: typ_meth) =
   M.{lab = Idllib.Escape.escape_method m.it.var.at m.it.var.it; typ = check_typ env m.it.meth; depr = None}
 
 let check_prog (env: typ I.Env.t) actor : M.typ =
+  m_env := M.Env.empty;
   match actor with
   | Some {it=ServT ms; _} ->
      let fs = List.map (check_meth env) ms in
--- a/src/mo_idl/idl_to_mo.ml
+++ b/src/mo_idl/idl_to_mo.ml
@@ -116,6 +116,6 @@ let check_prog (env: typ I.Env.t) actor : M.typ =
   | None -> assert false
   | _ -> assert false
 
-let occ = ref M.Env.empty t (* HACK *)
+let occs = ref M.Env.empty (* HACK *)
 let check_typ env t = check_typ' env occs t
 let check_typs env t = check_typs' env occs t
--- a/src/mo_idl/idl_to_mo.ml
+++ b/src/mo_idl/idl_to_mo.ml
@@ -4,7 +4,6 @@ open Source
 module M = Mo_types.Type
 module I = Idllib.Typing
 
-let m_env = ref M.Env.empty
 
 let check_prim at p =
   match p with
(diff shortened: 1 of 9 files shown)


src/mo_idl/idl_to_mo.ml
```diff
@@ -49,65 +48,73 @@ let is_tuple fs =
...
+  M.{lab = Idllib.Escape.escape_method m.it.var.at m.it.var.it; typ = check_typ' env occs m.it.meth; depr = None}
 
 let check_prog (env: typ I.Env.t) actor : M.typ =
+  let occs = ref M.Env.empty in
   match actor with
   | Some {it=ServT ms; _} ->
-     let fs = List.map (check_meth env) ms in
+     let fs = List.map (check_meth env occs) ms in
+     (* TODO: why do we only check and include the mentioned types (occs),
+        and not all of the .did declared ones (available to the caller, if not here? *)
```
chenyan-dfinity, line 104: I don't have a preference here, but if the service didn't mention this type, why does the user need to use that type? `did` file is kind of transparent to the user here: it's stored inside `.dfx` directory, not part of the source code.

test/run-drun/issue-3103.mo
```diff
@@ -0,0 +1,19 @@
+/* test actor import of eponymous types in separate candid files */
+//MOC-FLAG --actor-idl issue-3103
+
+import A "ic:lg264-qjkae";
+import B "ic:aaaaa-aa";
+
+actor a {
+  public func go() : async () {
+     let _ : Bool = await A.getBool(); // rejected, should be accepted
```
chenyan-dfinity, line 9:
```suggestion
     let _ : Bool = await A.getBool();
```

chenyan-dfinity (approved):
Thanks for the fix!

nomeata:
Is this just about adding type constructors “internally” to some environment, with an internal name, so that the type checker is happy? Or can the importing code actually explicitly refer to the imported `T`? (The test case nor the referenced issue doesn’t, so I’m not sure).

If it’s just internal I think it doesn’t matter whether one only looks as the occuring type defs, or all.

If it’s not just the internal ones, I’d expect _all_ to be uniformly available. (But how? Type members?)

crusso:
> Is this just about adding type constructors “internally” to some environment, with an internal name, so that the type checker is happy? Or can the importing code actually explicitly refer to the imported `T`? (The test case nor the referenced issue doesn’t, so I’m not sure).
> 
> If it’s just internal I think it doesn’t matter whether one only looks as the occuring type defs, or all.

No, its not just internal, they become type fields of the actor and thus accessible, which I find disturbing, particularly if its only those that are reference by the service type, not all of the ones that are declared by the .did.

> If it’s not just the internal ones, I’d expect _all_ to be uniformly available. (But how? Type members?)

Right, me too.

Unfortunately, enzo's QR code generator seems to be relying on this feature (Or at least, one of our tests based on that code uses it). If he's relied on it, I wonder if other users have too (hence my suggestion to add a deprecation).

crusso:
I'd be keen to know @rossberg's opinion on this one.

nomeata:
> No, it's not just internal, they become type fields of the actor and thus accessible, which I find disturbing

But then what was the bug? I thought there was a clash? Ah, was it that they were missing or wrongly typed if eponymous because of a clash in the internal data structure (but not a clash in the sense of conflicting definitions)?

crusso:
User relying on this in the wild:

https://github.com/enzoh/motoko-qr/blob/master/app/Demo.mo

crusso:
Merging the fix for now, can revisit what to do about type defs later.

PR #3110, opened 2022-02-11
Clean up `src = subpath` attrs

ggreif:


nomeata (approved):

ggreif:
I tried to understand how the definition `let subpath = import ./nix/gitSource.nix; in` works. This is just curried, right?

So `src = subpath ./doc;` will reduce to `src = import ./nix/gitSource.nix ./doc;`, but I can't grok that. @nomeata can you hit me with a clue-stick?

nomeata:
> This is just curried, right?

Not just curried; `import gitSource.nix` already does some work (reading `git ls-files`), and returns a function.

PR #3112, opened 2022-02-16
Shift primitives for `Nat`

ggreif:
Introduce two new primitives {`shiftLeft`, `shiftRight`}, both with signature `(Nat, Nat32) -> Nat`. This is hopefully helpful for #3175.

-------

- [x] `grep` for "TODO: use shift right instead"
- [x] `grep` for "TODO: use shift left instead"
- [ ] QuickChecks (but the tests are rather exhaustive already)
- [x] make the prims into calls? — 1672aa0680613e526932c5c71791969ee16a65f2

----------

- restructure the `lsh` code to reuse the 64-bit shift result to create the bignum ([see comment below](https://github.com/dfinity/motoko/pull/3112#discussion_r893264718))
- `clz >= 33` as a criterion for compact


rts/Makefile
```diff
@@ -19,7 +19,8 @@ TOMMATHFILES = \
    mp_init_size mp_exch mp_clear mp_copy mp_count_bits mp_mul_2d mp_rshd mp_mul_d mp_div_2d mp_mod_2d \
    s_mp_balance_mul s_mp_toom_mul s_mp_toom_sqr s_mp_karatsuba_sqr s_mp_sqr_fast s_mp_sqr s_mp_karatsuba_mul \
    s_mp_mul_digs_fast s_mp_mul_digs mp_init_multi mp_clear_multi mp_mul_2 mp_div_2 mp_div_3 mp_lshd mp_incr \
-   mp_decr mp_add_d mp_sub_d
+   mp_decr mp_add_d mp_sub_d \
+   mp_lshd mp_rshd
```
ggreif, line 23: not needed!

src/codegen/compile.ml
```diff
@@ -2045,6 +2045,7 @@ sig
   val compile_unsigned_div : E.t -> G.t
   val compile_unsigned_rem : E.t -> G.t
   val compile_unsigned_pow : E.t -> G.t
+  val compile_lshd : E.t -> G.t
```
ggreif, line 2048:
```suggestion
  val compile_lsh : E.t -> G.t
```

ggreif: 118f5a3a2

src/codegen/compile.ml
```diff
@@ -2350,6 +2353,49 @@ module MakeCompact (Num : BigNumType) : BigNumType = struct
...
+    set_amount ^^
+    set_n ^^ get_n ^^
+    BitTagged.if_tagged_scalar env [I32Type]
+      begin
+        get_n ^^
+        get_amount ^^
+        G.i (Binary (Wasm.Values.I32 I32Op.ShrU)) ^^
+        compile_bitand_const 0xFFFFFFFEl
+      end
+      (get_n ^^ get_amount ^^ Num.compile_rsh env)
```
ggreif, line 2397: check for tiny?

src/codegen/compile.ml
ggreif: Done: 3719c7c42159bf0444595cb4751d6694f11e9fa3

src/codegen/compile.ml
```diff
@@ -2350,6 +2353,59 @@ module MakeCompact (Num : BigNumType) : BigNumType = struct
...
+        G.i (Convert (Wasm.Values.I64 I64Op.ExtendUI32)) ^^
+        G.i (Binary (Wasm.Values.I64 I64Op.Shl)) ^^
+        let set_remember, get_remember = new_local64 env "remember" in
+        set_remember ^^ get_remember ^^
+        G.i (Convert (Wasm.Values.I32 I32Op.WrapI64)) ^^
+        let set_res, get_res = new_local env "res" in
+        set_res ^^ get_res ^^
+        G.i (Convert (Wasm.Values.I64 I64Op.ExtendSI32)) ^^ (* exclude sign flip *)
+        get_remember ^^
+        G.i (Compare (Wasm.Values.I64 I64Op.Eq)) ^^
```
ggreif, line 2376: One could structure it in a way that we still reuse the 64-bit result (by going to heap from there) even if it fails to truncate to 32 bits.

src/codegen/compile.ml
```diff
@@ -2344,6 +2347,65 @@ module MakeCompact (Num : BigNumType) : BigNumType = struct
           get_res
       end)
 
+
+  let compile_lsh env =
+    Func.share_code2 env "B_lsh" (("n", I32Type), ("amount", I32Type)) [I32Type]
+    (fun env get_n get_amount ->
+    (*let set_amount, get_amount = new_local env "amount" in
```
ggreif, line 2354: Make this a note

src/codegen/compile.ml
```diff
@@ -2344,6 +2347,57 @@ module MakeCompact (Num : BigNumType) : BigNumType = struct
           get_res
       end)
 
+
+  let compile_lsh env =
+    Func.share_code2 env "B_lsh" (("n", I32Type), ("amount", I32Type)) [I32Type]
+    (fun env get_n get_amount ->
+      get_n ^^
+      BitTagged.if_tagged_scalar env [I32Type]
+        ( (* non-ptr, shift amount <= 42: signed i32 -> i64; shift left, remember, trunc to i32, extend, compare with recalled, if same, done! *)
```
ggreif, line 2356: Make this a note

src/codegen/compile.ml
```diff
@@ -2344,6 +2347,57 @@ module MakeCompact (Num : BigNumType) : BigNumType = struct
...
+    (fun env get_n get_amount ->
+      get_n ^^
+      BitTagged.if_tagged_scalar env [I32Type]
+        begin
+          get_n ^^
+          get_amount ^^
+          G.i (Binary (Wasm.Values.I32 I32Op.ShrU)) ^^
+          compile_bitand_const 0xFFFFFFFEl ^^
+          get_amount ^^ compile_rel_const I32Op.LeU 31l ^^
+          G.i (Binary (Wasm.Values.I32 I32Op.Mul))
```
crusso, line 2389: What is this doing. for the arithmetically challenged like me?

test/run/nat-shift.mo
```diff
@@ -0,0 +1,37 @@
+import { debugPrint; shiftLeft; shiftRight } = "mo:⛔"
```
crusso, line 1: Instead of producing lots of output and relying on diff and visual inspection. why not just compare the result against the naive implementation using multiplication/division by recursively computed power of 2?

src/codegen/compile.ml
```diff
@@ -2344,6 +2347,57 @@ module MakeCompact (Num : BigNumType) : BigNumType = struct
...
+          G.i (Convert (Wasm.Values.I64 I64Op.ExtendSI32)) ^^ (* exclude sign flip *)
+          get_remember ^^
+          G.i (Compare (Wasm.Values.I64 I64Op.Eq)) ^^
+          get_amount ^^ compile_rel_const I32Op.LeU 42l ^^
+          G.i (Binary (Wasm.Values.I32 I32Op.And)) ^^
+          G.if1 I32Type
+            get_res
+            (get_n ^^ compile_shrS_const 1l ^^ Num.from_word30 env ^^ get_amount ^^ Num.compile_lsh env)
+        )
+        (get_n ^^ get_amount ^^ Num.compile_lsh env))
```
crusso, line 2376: I assume the Num operations do the compaction of result when necessary, right?

crusso (commented):
Generally looks good to me.

What this PR doesn't do is implement the shift operators, but I guess that can come later. I expect those will have type (Nat, Nat) -> Nat so might have to a range checked on the second argument.

Doesn't it also make sense to have shifts on Ints?

src/codegen/compile.ml
ggreif: Since Wasm` `bool` is {0, 1} this is a branch-free `if`.

src/codegen/compile.ml
ggreif: No. Any proper bignum shifted to the left will remain a bignum. No compaction can happen here.

src/codegen/compile.ml
```diff
@@ -8469,6 +8525,18 @@ and compile_prim_invocation (env : E.t) ae p es at =
   | OtherPrim "blob_iter_next", [e] ->
     SR.Vanilla, compile_exp_vanilla env ae e ^^ Blob.iter_next env
 
+  | OtherPrim "lshd", [e1; e2] ->
+    SR.Vanilla,
+    compile_exp_vanilla env ae e1 ^^
+    compile_exp_as env ae SR.UnboxedWord32 e2 ^^
+    BigNum.compile_lsh env
+
+  | OtherPrim "rshd", [e1; e2] ->
```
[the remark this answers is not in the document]
ggreif: I was wondering myself. I took this nomenclature from libTomMath.

test/run/nat-shift.mo
ggreif: Yeah, good idea. I started out wanting to see the results, but asserting makes a lot of sense now.

src/codegen/compile.ml
ggreif: Written out this is `if amount <= 31 then bit_clear(n >> amount, 0) else 0`.

ggreif: The  `bit_clear` being necessary to have a solar result.

src/codegen/compile.ml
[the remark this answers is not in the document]
ggreif: I'll change it.

test/run/nat-shift.mo
ggreif: Done.

src/codegen/compile.ml
ggreif: > Doesn't it also make sense to have shifts on Ints?

I am not entirely sure about that yet. Bit manipulations presume a bit pattern representation and we never say what that is for arbitrary precision `Int`s. Internally it is two's complement for compact `Int`s, but TomMathLib uses a different format (1 bit sign + absolute mantissa). We _could_ define the shift operators as-if `Int` was `base * 2 ** exp`.

ggreif: Also if we have a shift operator for `Int`, somebody will appear and ask for rotation operators :-)

src/codegen/compile.ml
ggreif: Done!

src/codegen/compile.ml
ggreif: Github is acting up again. The last two answers should have gone into the review comment thread. Oh well.

src/codegen/compile.ml
crusso: wow, ok. Maybe add that as a comment!

crusso (approved):
LTME - thanks for making the changes

src/codegen/compile.ml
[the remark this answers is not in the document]
ggreif: Done: 96d11c4e9f7e5754568e2988885917e7321c4366

PR #3113, opened 2022-02-17
CI: Publish reports to Github Actions

nomeata:
This publishes our various reports to https://dfinity.github.io/motoko/. So far, only for `master` (github action doesn’t support PR deployments like netlify would).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -91,9 +91,11 @@ jobs:
         edit-mode: replace
 
   reports:
+    # if: github.ref == 'refs/heads/master'
     needs: tests
     concurrency: ci-${{ github.ref }}
     runs-on: ubuntu-latest
+    steps:
     - uses: actions/checkout@v2
     - uses: cachix/install-nix-action@v16
       with:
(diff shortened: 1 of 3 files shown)


nomeata:
I thought maybe I can fix #2568 without waiting for anyone.

To boot, this PR creates a new derivation that combines all our “build reports” in one tree. This is a mild regression for what we had in hydra (there, you still could look at one report even though another might have failed), but probably ok.

Next I wanted to add a github action step that publishes that to a netlify page at <https://motoko-builds.netlify.app/>. But I stopped following that track because I need to be repo admin to configure secrets used by the build, but more importantly, [it seems](https://answers.netlify.com/t/per-site-access-tokens/52144) I use a netlify Personal Access Token which would give anyone with commit permissions on this repo the ability to control _any_ of my netlify. So maybe netlify isn’t great either, at least not in this mode.

We could use Github Pages if we don’t mind a `gh-pages` branch polluting our repo. I’ll explore this once I can change the repo settings again.

ggreif:
Why not upload to an asset canister on the IC?

nomeata:
Ok, deploy to github action works:

* https://dfinity.github.io/motoko/
* https://dfinity.github.io/motoko/flamegraphs/

Maybe start with this?

ggreif (approved):
LGTM! Some links give 404 errors, but we can check that later.

PR #3123, opened 2022-02-24
Fix #3121

### LINKED ISSUE #3121: Selective Field Import
It seems you can not import a file that uses selective field imports.
- #3076 by @ggreif

---

```tree
src
├── X.mo
└── Y.mo
```

```motoko
// X.mo
import { Array_init = init } = "mo:⛔";
```

```motoko
// Y.mo
import X "X";
```

### Error
```
src/X.mo:1.8-1.29: type error [M0015], only trivial patterns allowed in static expressions
```

### Context
```
$(vessel bin)/moc --check $(vessel sources) src/Y.mo
```
- `vessel 0.6.2`
- `Motoko compiler 0.6.21 (source wahvzdqw-grg3wqkh-0n8s7fj6-f4rxx0rh)`

ggreif:
Product patterns like `ObjP` should be allowed to be static (side-effect-free when matching in CBV) so allow them also. Analogous to `TupP`.

### DIFF
--- a/src/mo_frontend/static.ml
+++ b/src/mo_frontend/static.ml
@@ -104,9 +104,10 @@ and triv m p = match p.it with
   patterns here.
   *)
   | TupP ps -> List.iter (triv m) ps
+  | ObjP fs -> List.iter (fun (f : pat_field) -> triv m f.it.pat) fs
 
   (* TODO:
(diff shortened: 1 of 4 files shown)


nomeata (commented):
Rename `triv` to `pat` maybe?

nomeata (approved):

PR #3125, opened 2022-02-24
Correct the release date of 0.6.22

ggreif:


ggreif:
@nomeata I remembered that dates for releases were checked, but this one went though without CI protesting.

PR #3126, opened 2022-02-24
bump rust toolchain

ggreif:


ggreif:
Ubuntu tests seem to fail due to missing `libstdc++.so.6`. No idea how to debug this. @nomeata?

nomeata:
It happens in the `checkPhase`, else it builds nicely.

Ah, bumping to april seems to solve the problem?

ggreif (commented):
Cool, thanks for looking into this! The only thing that makes me worry is the 0.5%  size increase for the benchmarks...

nomeata (approved):

PR #3130, opened 2022-03-01
bugfix: fix pretty printing of types and #3128

### LINKED ISSUE #3128: bug: --stable-types generation does not emit types referenced only by used type definitions
See here for a real world example:
https://github.com/dfinity/examples/pull/164#issuecomment-1051011320

Probably a failure in Type.cons not implementing transitive closure by recursing into bodies of referenced definitions.

crusso:
* Fixes #3128 (by collecting constructors  *transitively* before emitting a .most file)
* Modifies type pretty printer to produce parsable types:
 * Print eponymous type variables using `<name>_<i>`, (no longer `<name>.<i>`), where `<i>` is the nth binding occurrence of this name (counting inwards), omitting `_<i>` when `<i>` is 0 and no eponymous constructor mentioned.
 * Print constructor stamps using `<name>__<stamp>` (no longer `<name>/<i>`) (using *double* underscore), omitting `_<i>` only when 0.
* Fixes bug in code computing index of shadowing type variables.
* Check stable sig parses and is compatible with self on every compile (a sanity check to detect ill-formed signatures early)
* Emit version number to stable signature (`// Version: 1.0.0`) for future versioning.
* Make type pretty printing more configurable by specifying separators for constructor stamps and type parameter indices.
* Always index type parameters  when occurring in same type/kind etc as some eponymous constructor (fixing an old hack).

Apart from rare clashes with user names involving `_`, this should ensure that generated type constructors and variable name are correctly distinguished yet parseable when emitted to `.most` files (or cut-n-paste by users) (depending on configuration).

This is not a perfect solution, but much better than what we have so far and should fix the problem the playground was encountering producing ill-formed .most files. I'm open to better suggestions.


chenyan-dfinity:
I think `T/2` feels more natural/readable than `T__2` in the error message. Do you consider only changing the type names in the generated `.most` file?

crusso:
> I think `T/2` feels more natural/readable than `T__2` in the error message. Do you consider only changing the type names in the generated `.most` file?

I thought about that and it wouldn't be hard to generalise the show_stamps pretty printer option to control this.

The other option is to use single underscore for both stamps and type parameters but use numeric  indices for stamps and alphabetic indices for parameters (List_1 and A_a).

I don't really like forms that aren't parseable, since it makes cut n paste scenarios painful.

Another option is to hide stamps but lazily reindex constructors from 0 for display of each error message, suppressing most stamps, most of the time, but I'm not sure how that would pan out.

crusso:
@rossberg any opinions on this?

crusso:
@ggreif any preferences?

chenyan-dfinity (approved):
Yeah, let's get in the fix first. I'm fine with the current format.

crusso:
@YanChen do you mean I should merge with the new format (using (double and single) underscores)? Or the old one using / and .?

chenyan-dfinity:
I mean the new underscore format.

crusso:
> I mean the new underscore format.

Great. I'll merge #3131 and #3132 into this branch and then squash.

PR #3134, opened 2022-03-03
experiment: diagnosing and addressing Issue 3057

crusso:
Don't merge this - its just a branch for investigating issue #3057.
* adds another repro
* limits recursion depth for subtyping and bi_matching to generate useful short stack-traces.
* hack definition of expansiveness (totally broken though). Needs some thought on how to recognize polymorphic recursion via a bound type parameter.


luc-blaeser:

PR #3139, opened 2022-03-06
bugfix: fix bug in ParseableStamps leading to duplicate type defs in sigs

### LINKED ISSUE #3140: bug: bogus identification of distinct type constructors in --stable-types section and .most file.
See here 

https://forum.dfinity.org/t/persistent-stable-interface-compatibility-check-failed-error/11281/6?u=claudio

repro:
```
actor {

  type A = Nat;
  module X = { public type A = Bool};
  module Y = { public type A = Text};
  type T = (A, X.A, Y.A);

  stable var t = (0, true, "abc") : T;

}
```
produces error: 
```
stable-types:3.6-3.7: type error [M0051], duplicate definition for type A in block
```

(the problem is that the type pretty printer isn't set to emit stamps in the signature file, leading to the identification of all types called `A`.)

crusso:
Fixes #3140.

```
actor {

  type A = Nat;
  module X = { public type A = Bool};
  module Y = { public type A = Text};
  type T = (A, X.A, Y.A);

  stable var t = (0, true, "abc") : T;

}
```
produces error: 
```
stable-types:3.6-3.7: type error [M0051], duplicate definition for type A in block
```

(the problem is that the type pretty printer isn't set to emit stamps in the signature file, leading to the identification of all types called `A`.)


nomeata (approved):
Rubber stamp

PR #3143, opened 2022-03-07
bugfix: fix for #3140 continued

crusso:
Continues the saga of fixing #3140 by being more careful about deciding how to elide type bindings for prelude defined prims. 

c.f. https://forum.dfinity.org/t/persistent-stable-interface-compatibility-check-failed-error/11281/9?u=claudio


nomeata (approved):
Rubber stamp

Issue #2994, opened 2021-12-13
Necessary tasks to make `--compacting-gc` the default


ggreif:
(YOU CAN LOOK, BUT THINKOS MAY BE INSIDE...)

8459514233 reverted the GC strategy to `copying`, this is the commit that needs to be backed out.

## Methods with pros/cons

So I propose to do the dumbest thing that works, for now. From what I understood we have two alternatives:
* allocation-time check with 3GB limit (pros: simple; cons: Cycles overhead per allocation, dead data may trigger rejection, queries could be penalised unnecessarily)
* post-GC check with 3 GB limit (pros: cheap, dead data not counted, query-fairness; cons: stable variables’ footprint may overshoot the (de)serialisation limits of 1 GB when streaming).

## Hazards

Several hazards lurk in the upgrade process
- DAG-like stable variable data gets serialised to trees, blowing up size
- serialisation buffer may overflow due to huge stable variable volume (FIXED: #3149)
- stable memory may get exhausted (if many regions allocated already)
- deserialisation buffer eats up too much heap space to successfully unpack
- streaming back stable variables overshoots the heap

The danger of overshooting the serialisation buffers is intuitively tallied in the following table.

(TABLE TO BE FINALISED🔴🟠🟡🟢)
| method | serialisation | ser-hazard | deserialisation | deser-hazard | prob. of heap overshoot on upgrade |
|---|---|---|---|---|---|
| alloc-check | heap buffered | 1 GB buffer too small 🟡 | heap buffered | 1 GB buffer too small 🟠 | middle |
| alloc-check | streamed | stable memory exhausted 🟢 | heap buffered | 1 GB buffer too small 🟠  | low |
| alloc-check | heap buffered | 1 GB buffer too small 🟡 | streamed | 3 GB heap too small 🟢 | middle |
| alloc-check | streamed | stable memory exhausted 🟢 | streamed | 3 GB heap too small 🟠 | middle |
| post-GC-check | heap buffered | <1 GB buffer too small 🟠 | heap buffered | >3 GB heap too small 🟢 | middle |
| post-GC-check | streamed | stable memory exhausted 🟡 | heap buffered | buffer and heap > 4 GB 🔴 | high |
| post-GC-check | heap buffered | <1 GB buffer too small 🟠 | streamed | >3 GB heap too small 🟢 | low |
| post-GC-check | streamed | stable memory exhausted 🟡 | streamed | 4 GB heap too small 🔴 | high |

[...]

PR #3149, opened 2022-03-09
Streaming `Stabilization` (emission)

### LINKED ISSUE #2909: GC: Reserve Wasm memory for upgrading canisters
The new mark-compact GC will try to use all 4GB of Wasm memory before giving up and failing with out-of-memory. This potentially leaves no space for pre-upgrade hook to run, so the canister can get into a state that is not upgradable.

Note that the old copying GC doesn't have the issue because it can allocate at most 2GB in a semi-space.

The solution would be to reserve space for upgrading and fail normal messages with OOM earlier. For example, reserve 1GB and fail normal messages with OOM after the heap size exceeds 3GB.

ggreif:
## Purpose

Streaming stabilisation is a necessary prerequisite for the compacting GC. This is explained in #2909 and #2994.

## Approach

We build (and extend) on the `Stream` abstraction introduced in #3161 and instantiate a new module that is capable of caching streaming data into a blob. When the cache limit is reached, an outputter bottleneck function will be called to move out byte ranges to a backing store. This is `BlobStream`.

`Stabilization` then further extends this idiom to set up the bottleneck function to directly write big chunks into stable memory.

Fixes #2909.

## Fine points

This section details some of the more tricky issues that arose while implementation, and how they have been mitigated. With some more thought, these may be completely eliminated in the future.

### Linking against `ic::stable64_write` from Rust

The stream bottleneck is also defined in Rust (in `stream.rs`), so it wants to call `stable64_write` from the `ic` module. This  will cause linker errors for the WASI mode. Instead we call `stable64_write_moc` from `env` (which `moc` provides). In turn `moc` will generate a trapping `stable64_write_moc` for WASI mode or re-export the proper `stable64_write`.

### The contract of `Stream::reserve`

When calling `Stream::reserve` it is the caller's responsibility to immediately fill in the reserved region. Note that the returned pointer is unaligned, so either write bytes or use `align = 0` in Wasm.

### Finalising a `Stream` (OCaml)

Some streams like to write out bytes eagerly, while others may need to cache all bytes until `Stream.terminate`. At this point the length of _all_ written bytes is returned, along with an optional pointer to the start of the data. For eager streams it might be impossible to return a pointer to the 0th byte, because it is not in the cache any more. In this case `0` must be returned. Should a value `!= 0` be returned, then the stream's user has the obligation to consume the bytes (freeing is _not_ necessary/possible). A helper is provided from the `Stream` API: `Stream.finalize_buffer` takes a code snippet that the user provides that consumes the length/dataptr, but the `Stream` implementation will decide whether the code will be inserted or ignored (because all bytes are already at their destination, e.g. in the backing store).
To be fully `Stream`-agnostic, either the dataptr returned from `Stream.terminate` must be checked at runtime, or the compile-time `Stream.finalize_buffer` facility must be used.

-------
## TODOs

Not all of these are strictly necessary (at this point), but nice to have.

- [x] Move boring refactors in a prerequisite PR! — #3182
- more `Stream` tests
- [x] outputting to stable
  - track totals (see `absolute_offset`)
  - support new SM layout too
- check every non-constant `stream_reserve` for `blob` overflow
  - is overflow bad here?
  - currently benign, but I left `FIXME`s in place – follow-on PR will be written
- [ ] consider moving a BigNum -> (S)LEB128 converter into `Stream` to alleviate the "`reserve` by unknown amount" issue
- make `as_blob` return a `*const` (others too?)
- [ ] [define and use](https://github.com/dfinity/motoko/pull/3182#discussion_r843605469) `compile_extend64`
- Should we use different `Stream`s for `Externalization` depending on the GC strategy?
- [ ] decouple the `StableMemoryStream` from the `N+4` start address


rts/motoko-rts/src/stream.rs
```diff
@@ -0,0 +1,24 @@
...
+//! fill up an in-heap buffer and then copy that wholesale into stable memory. This can be
+//! disadvategous for two reasons:
+//!  - double copying
+//!  - heap congestion (especially for the compactig collector)
+//!
+//! Instead now we'll 
+//!
+//!
+
+// Layout of a stream node:
```
crusso, line 13: Looks like you are introducing a new object tag an type for this, but I wonder if we can get away with something much simpler. Shall we chat?

ggreif: Nope, I'll reuse the `blob` tag

rts/motoko-rts-tests/src/main.rs
```diff
@@ -21,7 +22,8 @@ fn main() {
     }
 
     unsafe {
-        bigint::test();
+        stream::test();
```
ggreif, line 25: undo this!

ggreif:
@nomeata I get
```
ingress Err: IC0503: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped explicitly: Stable memory error: Aliased at wrong type, expected: VN
```
when I use BlobStream instead of the `BumpStream`.  The relative offsets are the same I have checked this. What is this type hash? Maybe I am writing it out wrongly?

rts/Makefile
```diff
@@ -223,8 +223,8 @@ $(TOMMATH_BINDINGS_RS): | _build
 
 
 $(RTS_RUST_WASM_A): $(TOMMATH_BINDINGS_RS) $(RTS_RUST_FILES) $(RTS_CARGO_FILES) | _build/wasm
-	cd motoko-rts && cargo build --release --target=wasm32-unknown-emscripten -Zbuild-std=core,alloc
-	cp motoko-rts/target/wasm32-unknown-emscripten/release/libmotoko_rts.a $@
+	cd motoko-rts && cargo build --target=wasm32-unknown-emscripten -Zbuild-std=core,alloc
```
ggreif, line 226: undo

rts/motoko-rts/src/debug.rs
```diff
@@ -8,7 +8,7 @@ use core::fmt::Write;
 /// Print an object. The argument can be a skewed pointer to a boxed object, or a tagged scalar.
 #[cfg(feature = "ic")]
 #[no_mangle]
-unsafe extern "C" fn print_value(value: Value) {
+pub unsafe extern "C" fn print_value(value: Value) {
```
ggreif, line 11: function is never used, but I think this should be public

rts/motoko-rts/src/debug.rs
```diff
@@ -199,8 +199,12 @@ pub(crate) unsafe fn print_boxed_object(buf: &mut WriteBuf, p: usize) {
             );
         }
         TAG_BLOB => {
-            let blob = obj as *const Blob;
-            let _ = write!(buf, "<Blob len={:#x}>", (*blob).len.as_u32());
+            let blob = obj.as_blob();
+            let _ = write!(buf, "<Blob len={:#x}>{:#x}{:#x}{:#x}{:#x}\n", blob.len().as_u32(), *blob.payload_addr(), *blob.payload_addr().add(1), *blob.payload_addr().add(2), *blob.payload_addr().add(3));
```
ggreif, line 203: undo below

rts/motoko-rts-tests/src/stream.rs
```diff
@@ -0,0 +1,57 @@
...
+            .stash(&chunk[0], Bytes(chunk.len() as u32));
+    }
+    assert_eq!(stream.as_blob().get(24), 10);
+    assert_eq!(stream.as_blob().get(25), 1);
+    assert_eq!(stream.as_blob().get(33), 9);
+    assert_eq!(stream.as_blob().get(34), 10);
+    assert_eq!(stream.as_blob().get(6023), 9);
+    let blob = stream.as_stream().split();
+    assert_eq!(blob.as_blob().len(), Bytes(6000));
+
```
ggreif, line 53: `reserve` also needs tests!

rts/motoko-rts/src/stream.rs
```diff
@@ -0,0 +1,133 @@
...
+            (self as *mut Blob).set((*self).filled.as_u32(), byte);
+            (*self).filled += Bytes(1)
+        }
+    }
+
+    #[export_name = "stream_reserve"]
+    pub fn reserve(self: *mut Self, bytes: Bytes<u32>) -> *mut u8 {
+        unsafe {
+            if (*self).filled + bytes > (*self).header.len {
+        assert!(false);
```
ggreif, line 112: remove!

nomeata:
> What is this type hash? 

https://github.com/dfinity/motoko/blob/master/src/mo_types/typ_hash.ml

So `VN` is `[var Nat]`.

nomeata:
Hmm, our extended Candid format follows indirections for aliased data, at least during decoding. And actually mutates the input data to leave traces… This might make streaming _reads_ hard. Did you plan to do that, or is this just about streaming writes for now?

ggreif:
> is this just about streaming writes for now?

Yes, only streaming writes. I have it basically working, only trouble is calling  `ic0::stable64_write` from Rust, which causes linker errors when compiling for WASI. Unfortunately the mere presence of the `extern` in the RTS will  cause the error even if the symbol that calls it (`send_to_stable`) isn't referenced by the final executable :-(

I am looking for solutions that aren't gross  hacks...

ggreif:
> And actually mutates the input data to leave traces… This might make streaming _reads_ hard.

I guess one could abstract away the `Blob` mutations and record those in a `StreamPos -> (memo, type_hash)`-style map instead. Then the past portion of the stream can be forgotten. But this is very much future work.

src/codegen/compile.ml
```diff
@@ -4312,6 +4301,19 @@ module RTS_Exports = struct
     E.add_export env (nr {
       name = Wasm.Utf8.decode "rts_trap";
       edesc = nr (FuncExport (nr rts_trap_fi))
+    });
+
+    let stable64_write_moc_fi =
+      if E.mode env = Flags.WASIMode then
+        E.add_fun env "stable64_write_moc" (
+            Func.of_body env ["to", I64Type; "from", I64Type; "len", I64Type] [] (fun env ->
+                G.nop
```
nomeata, line 4310: Maybe trap here? This shouldn’t ever be called, right?

nomeata:
> when on WASI, define stable64_write_moc, otherwise just re-export

I saw your email; but this would have been my suggestion as well. Good job reusing the function number, this way it’s even zero-cost.

src/codegen/compile.ml
ggreif: Totally!

ggreif: Done: 519b8646d

crusso (commented):
Some preliminary comments - will do a thorough review tomorrow.

src/codegen/compile.ml
```diff
@@ -5757,10 +5909,51 @@ module Stabilization = struct
 
   let extend64 code = code ^^ G.i (Convert (Wasm.Values.I64 I64Op.ExtendUI32))
 
+  module StableMemoryStream : Stream = struct
+    include BlobStream
```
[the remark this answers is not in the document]
ggreif: I should!

rts/motoko-rts/src/stream.rs
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    #[inline]
+    pub unsafe fn as_blob_mut(self: *mut Self) -> *mut Blob {
+        self as *mut Blob
+    }
+
+    /// make sure that the cache is empty
+    #[inline]
+    fn flush(self: *mut Self) {
```
ggreif, line 78:
```suggestion
    /// make sure that the cache is empty
    fn flush(self: *mut Self) {
```

rts/motoko-rts/src/stream.rs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+extern "C" {
+    // generated by `moc`
+    fn stable64_write_moc(to: u64, ptr: u64, n: u64);
+}
+
+impl Stream {
+    #[inline]
+    pub unsafe fn cache_addr(self: *const Self) -> *const u8 {
```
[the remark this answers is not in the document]
ggreif: I don't get  it:
```
error: cannot pass `self` by raw pointer
  --> src/stream.rs:90:30
   |
90 |     fn no_backing_store(*mut self, _ptr: *const u8, _n: Bytes<u32>) {
   |                              ^^^^ cannot pass `self` by raw pointer
```

src/codegen/compile.ml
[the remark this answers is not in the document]
ggreif: done: ee4b9e4cc

rts/motoko-rts/src/stream.rs
```diff
@@ -0,0 +1,184 @@
...
+        self.add(1) as *const u8 // skip closure header
+    }
+
+    #[inline]
+    pub unsafe fn as_blob_mut(self: *mut Self) -> *mut Blob {
+        self as *mut Blob
+    }
+
+    /// make sure that the cache is empty
+    #[inline]
```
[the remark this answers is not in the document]
ggreif: Removed in 9ff260a78379d6c851fdc8d4c5da0fb7b5e5cbdb

src/codegen/compile.ml
```diff
@@ -5744,6 +5829,86 @@ end (* MakeSerialization *)
...
+
+  let write_text env get_token get_x =
+    write_word_leb env get_token (get_x ^^ Text.size env) ^^
+    get_token ^^ get_x ^^
+    E.call_import env "rts" "stream_write_text"
+
+  let write_bignum_leb env get_token get_x =
+    get_token ^^
+    get_x ^^ BigNum.compile_data_size_unsigned env ^^
+    E.call_import env "rts" "stream_reserve" ^^ (* FIXME: might overflow! *)
```
ggreif, line 5897: This problem needs a resolution. Maybe we can define it away by saying that writing past the end of the cache is benign.

ggreif: Or somehow (debug mode?) check if the `HEAP_PTR` has advanced after the stream's allocation.

ggreif: Actually `reserve` could do this for us.

rts/motoko-rts/src/stream.rs
```diff
@@ -0,0 +1,184 @@
...
+        unsafe {
+            (*self).ptr64 = start;
+            (*self).limit64 = limit;
+            (*self).outputter = Self::send_to_stable;
+        }
+    }
+
+    /// Ingest a number of bytes into the stream.
+    #[export_name = "stream_write"]
+    pub fn stash(self: *mut Self, ptr: *const u8, n: Bytes<u32>) {
```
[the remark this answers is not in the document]
ggreif: e5c3e18bd. I get a strange build error with this, that I don't understand.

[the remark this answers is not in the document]
ggreif: Are you okay with `cache_bytes`?

ggreif:
Just for comparison, CDK has something that sounds similar: https://github.com/dfinity/cdk-rs/pull/243. I haven't looked at it yet.

rts/motoko-rts/src/stream.rs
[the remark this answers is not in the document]
ggreif: I figured out the build error: it comes because of https://doc.rust-lang.org/std/ptr/fn.write_bytes.html already being defined by the Rust lib.

rts/motoko-rts/src/stream.rs
```diff
@@ -0,0 +1,184 @@
...
+                (*self).filled += n;
+                assert!((*self).filled <= (*self).header.len);
+                memcpy_bytes(dest as usize, ptr as usize, n);
+            }
+        }
+    }
+
+    /// Ingest a single byte into the stream.
+    #[export_name = "stream_write_byte"]
+    pub fn stash8(self: *mut Self, byte: u8) {
```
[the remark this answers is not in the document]
ggreif: It  became `cache_byte` in a2ed6a17e653e9fec4b4fb2736bd4996a80fa01d.

src/codegen/compile.ml
```diff
@@ -5779,10 +6012,13 @@ module Stabilization = struct
...
-        extend64 get_dst ^^
-        extend64 get_len ^^
-        IC.system_call env "stable64_write"
+        StableMemoryStream.finalize_buffer
+          begin
+            compile_const_64 4L ^^
+            extend64 get_dst ^^
+            extend64 get_len ^^
+            IC.system_call env "stable64_write"
+          end
```
crusso, line 6021:
```suggestion
          begin
            compile_const_64 4L ^^
            extend64 get_dst ^^
            extend64 get_len ^^
            IC.system_call env "stable64_write"
          end
```
Shouldn't this code be delete now?

crusso (commented):
Added interim comments

rts/motoko-rts/src/stream.rs
```diff
@@ -0,0 +1,184 @@
...
+            }
+            self.as_blob_mut().set((*self).filled.as_u32(), byte);
+            (*self).filled += Bytes(1)
+        }
+    }
+
+    /// Return a pointer to a reserved area of the cache and advance the
+    /// fill indicator beyond it.
+    #[export_name = "stream_reserve"]
+    pub fn reserve(self: *mut Self, bytes: Bytes<u32>) -> *mut u8 {
```
crusso, line 151: What happens if you reserve some space, returning its pointer, but the next write flushes the cache. Wouldn't it be bad to now use that pointer? Is this guaranteed to never happen?

rts/motoko-rts/src/stream.rs
```diff
@@ -0,0 +1,184 @@
...
+    /// Split the stream object into two `Blob`s, a front-runner (small) one
+    /// and a latter one that comprises the current amount of the cached bytes.
+    /// Lengths are adjusted correspondingly.
+    #[export_name = "stream_split"]
+    pub unsafe fn split(self: *mut Self) -> Value {
+        (*self).header.len = INITIAL_STREAM_FILLED - size_of::<Blob>().to_bytes();
+        (*self).filled -= INITIAL_STREAM_FILLED;
+        let blob = (self.cache_addr() as *mut Blob).sub(1);
+        (*blob).header.tag = TAG_BLOB;
+        debug_assert_eq!(blob.len(), (*self).filled);
```
crusso, line 174: Is this actually correct. Isn't the extent of a blob the len rounded up to the nearest word? If blob.len is less than that (because we only filled a small portion of the cache, then there will be some gap in the heap, I think.

src/codegen/compile.ml
```diff
@@ -4304,6 +4345,33 @@ module type Stream = sig
...
+     Arguments:      env    token  size *)
+  val check_filled : E.t -> G.t -> G.t -> G.t
+
+  (* Pushes the stream's current absolute byte offset on stack.
+     The requirement is that the difference between two two uses
+     of this method must give a correct _relative_ offset.
+     Arguments:         env    token *)
+  val absolute_offset : E.t -> G.t -> G.t
+
+  (* Finishes the stream, performing consistence checks.
```
crusso, line 4362:
```suggestion
  (* Finishes the stream, performing consistency checks.
```

src/codegen/compile.ml
```diff
@@ -4304,6 +4345,33 @@ module type Stream = sig
...
+     on the Stream.
+     Arguments:   env    token  size   header_size *)
+  val terminate : E.t -> G.t -> G.t -> int32 -> G.t
+
+  (* Executes code to eliminate the residual buffer
+     that `terminate` returns (if at all) *)
+  val finalize_buffer : G.t -> G.t
+
+  (* Builds a unique name for a name seed and a type *)
+  val name_for : string -> string -> string
```
crusso, line 4373:
```suggestion
  val name_for : string -> T.typ -> string
```
and then do the hash internally?

src/codegen/compile.ml
```diff
@@ -4315,9 +4383,34 @@ end
...
+    G.i (Compare (Wasm.Values.I32 I32Op.Eq)) ^^
+    E.else_trap_with env "data buffer not filled"
+
+  let terminate env get_data_buf get_data_size header_size =
+    get_data_buf ^^ compile_sub_const header_size ^^
+    get_data_size ^^ compile_add_const header_size
+
+  let finalize_buffer code = code
+
+  let name_for seed typ_name = "@" ^ seed ^ "<" ^ typ_name ^ ">"
```
crusso, line 4407:
```suggestion
  let name_for  fn_name typ = "@" ^ fn_name ^ "<" ^ (typ_hash typ) ^ ">"
```

src/codegen/compile.ml
```diff
@@ -4732,7 +4825,7 @@ module MakeSerialization (Strm : Stream) = struct
   let rec serialize_go env t =
     let open Type in
     let t = Type.normalize t in
-    let name = "@serialize_go<" ^ typ_hash t ^ ">" in
+    let name = Strm.name_for "serialize_go" (typ_hash t) in
```
crusso, line 4828:
```suggestion
    let name = Strm.name_for "serialize_go" t in
```

src/codegen/compile.ml
```diff
@@ -5501,7 +5594,7 @@ module MakeSerialization (Strm : Stream) = struct
 
   let serialize env ts : G.t =
     let ts_name = typ_seq_hash ts in
-    let name = "@serialize<" ^ ts_name ^ ">" in
+    let name = Strm.name_for "serialize" ts_name in
```
crusso, line 5597:
```suggestion
    let name = Strm.name_for "serialize" ts in
```
(and delete previous let ts_name (hopefully unused now))

src/codegen/compile.ml
```diff
@@ -5744,6 +5829,86 @@ end (* MakeSerialization *)
...
+  let terminate env get_token _get_data_size _header_size =
+    get_token ^^ E.call_import env "rts" "stream_split" ^^
+    let set_blob, get_blob = new_local env "blob" in
+    set_blob ^^
+    get_blob ^^ Blob.payload_ptr_unskewed ^^
+    get_blob ^^ Blob.len env
+
+  let finalize_buffer code = code
+
+  let name_for seed typ_name = "@Bl_" ^ seed ^ "<" ^ typ_name ^ ">"
```
crusso, line 5853:
```suggestion
  let name_for fn_name ts = "@" ^ fn_name ^ "<" ^ (typ_seq_hash ts) ^ ">"
```

src/codegen/compile.ml
ggreif: Sadly both `BumpStream` and `BlobStream` don't work for `Stabilisation` right now :-(
I'll investigate.

rts/motoko-rts/src/stream.rs
ggreif: I am pretty sure that `len` can be 1. Otherwise how would you have `"a" : Text`?

rts/motoko-rts/src/types.rs
```diff
@@ -488,6 +496,15 @@ impl Blob {
     }
 }
 
+#[repr(C)] // See the note at the beginning of this module
+pub struct Stream {
+    pub header: Blob,
```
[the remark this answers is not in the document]
ggreif: Nope, it needs a `tag` too, otherwise it cannot be passed around as `Value`.

src/codegen/compile.ml
```diff
@@ -5757,10 +5922,78 @@ module Stabilization = struct
...
+    let name_for seed typ_name = "@Sm_" ^ seed ^ "<" ^ typ_name ^ ">"
+
+    let create env get_data_size set_token get_token header =
+      create env (compile_unboxed_const 0x8000l) set_token get_token header ^^
+        (* TODO: push header directly? *)
+
+      get_data_size ^^
+      compile_add_const (Int32.of_int (String.length header)) ^^
+      let (set_len, get_len) = new_local env "len" in
+      set_len ^^
```
crusso, line 5942:
```suggestion
      let (set_len, get_len) = new_local env "len" in
      get_data_size ^^
      compile_add_const (Int32.of_int (String.length header)) ^^
      set_len ^^
```
I think that's the style used elsewhere... declare, then define.

crusso (commented):
The hacky treatment of `N`(+4)in the stable memory version of `create` is a bit concerning. 
Would it be possible to refactor `create` to take an (optional) destination address (perhaps 64bit so wide enough for both heap and stable memory) and then use that appropriately?

Anyway, need to go to a meeting now...

rts/motoko-rts/src/stream.rs
crusso: Right, but I don't think the blob's physical size should be more than 3-bytes longer than len, right? Perhaps it doesn't matter at the moment, but I think it might create gaps in the heap that don't contain objects.

rts/motoko-rts/src/types.rs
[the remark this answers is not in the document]
crusso: So type Blob has the size of a Blob object tag + len?

rts/motoko-rts/src/stream.rs
ggreif: The code is correct. Before:
```
+---- *mut Stream
|
V
┌────────────┬─────┬───────┬─────────┬───────────┬────────┬──────────┐
│ tag (blob) │ len │ ptr64 │ limit64 │ outputter │ filled │ cache... │
└────────────┴─────┴───┴───┴────┴────┴───────────┴────────┴──────────┘
```
After
```
┌────────────┬─────┬───────┬─────────┬────────────┬────────┬─────────┐
│ tag (blob) │ 16  │ ptr64 │ limit64 │ tag (blob) │  len   │ viscera │
└────────────┴─────┴───┴───┴────┴────┴────────────┴────────┴─────────┘
                                     ^
                                     |
                                     +------- result blob
```

ggreif: Aaaah, I see what you mean. You are concerned about the empty space _after_ `viscera`, which could indeed be shorter than the original thing. It would be easy to fill that with another _valid_ heap object too.
_But_ after `Stabilisation` the heap gets erased anyway...

rts/motoko-rts/src/types.rs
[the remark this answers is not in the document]
ggreif: Yes the Rust `size_of<Blob>` will give you 8 back for Wasm.

rts/motoko-rts-tests/src/stream.rs
ggreif: We have them now.

rts/motoko-rts/src/stream.rs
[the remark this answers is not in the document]
ggreif: That is exactly what I demonstrated above, that the `rustc` rejects it :-(

ggreif:
> The hacky treatment of `N`(+4)in the stable memory version of `create` is a bit concerning.
> Would it be possible to refactor `create` to take an (optional) destination address (perhaps 64bit so wide enough for both heap and stable memory) and then use that appropriately?

I don't think so. The code generated by `create` ends up in a Wasm function body (with fixed argument count), so passing in local variable getters or extra stack arguments is not possible. Constants and globals are fine. `N` is a global, so that works. We basically have no choice but to recompute `N + 4`. Why do you think that this is a hack?

src/codegen/compile.ml
ggreif: Update: found the problem. f0d6437c0 fixes it. Now I can use either `BumpStream`, `BlobStream` or `StableMemoryStream` for `Externalization`.

crusso: Cool.

crusso:
> > The hacky treatment of `N`(+4)in the stable memory version of `create` is a bit concerning.
> > Would it be possible to refactor `create` to take an (optional) destination address (perhaps 64bit so wide enough for both heap and stable memory) and then use that appropriately?
> 
> I don't think so. The code generated by `create` ends up in a Wasm function body (with fixed argument count), so passing in local variable getters or extra stack arguments is not possible. Constants and globals are fine. `N` is a global, so that works. We basically have no choice but to recompute `N + 4`. Why do you think that this is a hack?

Why not just add an argument to create? It can be unused for the blob version, if necessary. The reason I find it hacky its that mixing the implementations of two things that could perhaps be kept a bit more separate.

rts/motoko-rts/src/stream.rs
crusso: yes, I guess what I was trying to say is that the physical size of the second blob will be BLOB+len'+3/4 (as seen by the GC) and thus there may be some junk in the heap (your ###). What I don't know is whether that is problematic or not - do we have a one word GC object that we could use to fill the space? I imagine if we implement a mark sweep algorithm, the junk might be problematic.

ggreif: Yep, I guess we can reuse this: https://github.com/dfinity/motoko/blob/5319cd5dd87bc9a43a27a048b44cd5eea7336fd7/rts/motoko-rts/src/types.rs#L476

crusso: perfect.

ggreif: Okay, bfee93a62 should take care of this. PHAL.

crusso (changes requested):
I think the main things remaining to do are:

Clean up the StableMemory serializer so it doesn't refer to global get_mem_size, but perhaps stores the 64-base address in the rust stream as an additional constant field, passed as an argument to `create` and consulted in `finalize` and `relative` address calculations.

Then the existing stabilization code can be cleaned up to look a bit more like it used to, perhaps just pushing the early serialization into the branches  of the conditional, creating appropriate stream only once you've determined the value of 'N'.

For the possibly overflowing StableMemory serialization of Nat/Int, I'd prefer to see a check that memory that is written beyond the stream cache is actually allocated via mem_grow (perhaps 'reserve' can do this).

I think the proper way to avoid this would be to actually add a dedicated bigint encode method to the Rust stream type, that writes bytes to the cache in a tight loop, flushing when necessary. That should just be a matter of cut-n-pasting and adapting the two functions in bigint.rs, as far as I can tell.

Making these changes would (in future) pave the way for allowing streaming serialization (of pure, never extended, candid) to stable memory at arbitrary offsets (say as an extension of the StableMemory API).

src/codegen/compile.ml
```diff
@@ -5779,10 +6015,13 @@ module Stabilization = struct
...
-        extend64 get_dst ^^
-        extend64 get_len ^^
-        IC.system_call env "stable64_write"
+        Externalization.Strm.finalize_buffer
+          begin
+            compile_const_64 4L ^^
+            extend64 get_dst ^^
+            extend64 get_len ^^
+            IC.system_call env "stable64_write"
+          end
```
crusso, line 6024:
```suggestion
        Externalization.Strm.finalize_buffer
          begin
            compile_const_64 4L ^^
            extend64 get_dst ^^
            extend64 get_len ^^
            IC.system_call env "stable64_write"
          end
```
Is it possible to just refactor this to:
```
 Externalization.Strm.finalize_buffer get_dst get_len
```
and hid the details in the implementation of finalize_buffer?

src/codegen/compile.ml
```diff
@@ -5744,6 +5830,86 @@ end (* MakeSerialization *)
...
+    get_x ^^
+    BigNum.compile_store_to_data_buf_unsigned env ^^
+    G.i Drop
+
+  let write_bignum_sleb env get_token get_x =
+    get_token ^^
+    get_x ^^ BigNum.compile_data_size_signed env ^^
+    E.call_import env "rts" "stream_reserve" ^^ (* FIXME: might overflow! *)
+    get_x ^^
+    BigNum.compile_store_to_data_buf_signed env ^^
```
crusso, line 5908: I think if you make bignum encoding a method a the rust stream then you wouldn't need to compute size and reserve space, saving some cycles! And I know you like to save cycles....

src/codegen/compile.ml
ggreif: I consider that as a layering violation. If `BumpStream` had a `finalize_buffer`, how should that be implemented? I cannot do `stable64_write`.
/me scratches head

src/codegen/compile.ml
```diff
@@ -5804,11 +6043,15 @@ module Stabilization = struct
...
-        IC.system_call env "stable64_write" ^^
+        (* copy data to following stable memory *)
+        Externalization.Strm.finalize_buffer
+          begin
+            get_N ^^
+            compile_add64_const 4L ^^
+            extend64 get_dst ^^
+            extend64 get_len ^^
+            IC.system_call env "stable64_write"
+          end ^^
```
crusso, line 6054:
```suggestion
        (* copy data to following stable memory *)
          get_N ^^
          compile_add64_const 4L ^^
          Externalization.Strm.finalize_buffer get_dst get_len ^^
```
Ditto.

src/codegen/compile.ml
crusso: (the nop implementaton would just pop the stack)

src/codegen/compile.ml
ggreif: Added a TODO in the PR description.

rts/motoko-rts-tests/src/main.rs
ggreif: Done: 2d812e74f

crusso (approved):
I'd prefer to fix the issues before we make a release. But if you want to merge as is I can approve that as long as we don't release. (edited) 

I'm happy fixing adding the N fields and fixing the BigInt encoder myself next week, before I do another release.

ggreif:
> I'm happy fixing adding the N fields and fixing the BigInt encoder myself next week, before I do another release.

I guess I can do this too.

PR #3153, opened 2022-03-14
feat: candid subtype check, during deserialization (tests)

crusso:
builds on #3151 
bumps candid dependency to see which tests are failing, if any.

### DIFF
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -6,10 +6,10 @@
         "homepage": "",
         "owner": "dfinity",
         "repo": "candid",
-        "rev": "a555d77704d691bb8f34e21a049d44ba0acee3f8",
-        "sha256": "0vn171lcadpznrl5nq2mws2zjjqj9jxyvndb2is3dixbjqyvjssx",
+        "rev": "0c8e620444ad6e2a4105ddc87b0e6292063b26cc",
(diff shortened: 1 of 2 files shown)


crusso:
> NB: I don't think the candid test suite has been updated to [dfinity/candid#311](https://github.com/dfinity/candid/pull/311), which I believe you are implementing.

Yeah, @chenyan-dfinity warned me but suggested I try bumping anyway.

> Maybe open a PR against candid so that you can fix the test suite as you go?

Yeah, makes sense. But I need to do some more plumbing first (like the failing, not trapping) and other fixes. I think I'll use a global type table for all types involved in subtype checks, and then construct a fresh cache  for each outer call to deserialize, shared between inner calls to sub type checks (to speed up the check for e.g. vectors of functions)

> Let me know if you need help running the test suite locally for faster iteration.

I'm doing nix-build -A tests.candid? Is there another way?

nomeata:
> I'm doing nix-build -A tests.candid? Is there another way?

Yes! `nix-build` is usually only what you want on CI, or maybe for final checks. For incremental development it’s probably too slow.

Just run
```
candid-tests -i ../path/to/candid/tests
```

* if you make changes to motoko, run `make -C src moc`
* if you make changes to the tests directory, `candid-tests` will pick them up

You can use `candid-tests -p "too long"` or run just a few tests, for quick iteration

This way you don’t have to recompile more than needed as you develop, and only need to `git commit` at the end.

PR #3155, opened 2022-03-14
Operations `to_candid` and `from_candid` as surface syntax.

matthewhammer:
Provide `to_candid` and `from_candid` operations in Motoko, for converting to and from Candid encoded `Blob`s.

Related to https://github.com/dfinity/motoko/pull/1754, which focuses on the IR and backend parts of this feature.

Redoes parts of https://github.com/dfinity/motoko/pull/1652, which did something very similar.

However there are a couple of subtle differences from the existing and past attempts:
 - `to_candid` uses the type inference to infer the types of its literal tuple arguments.  So, it often requires no explicit type annotations (unlike current serialization prim).
 - `from_candid` also using bidirectional typing to avoid annotations.  But it only works in type *checking* contexts (not type synthesizing ones), and always produces a tuple that matches the expected type of the context around the `Blob` to be deserialized.  It's an error to provide a non-`shared`, (or non-`?`) context for checking `from_candid`.
 - `from_candid` produces `null` for any type mis-match errors for otherwise valid Candid, and produces an optional result on success.  It traps for invalid Candid.  The earlier primitive that this one enhances here trapped in all error situations, and did not produce an optional result.

Steps to do
 - [x] typing.  See [@nomeata's PR that did this](https://github.com/dfinity/motoko/pull/1652/files#diff-3f553f492c5fe9acdbed07b9024025f63fffac45f2ce15ef36f2bd7e2f376dc2R773).
 - [x] use existing codegen for `SerializePrim` to codegen for `to_candid`.
 - [x] use existing codegen for `DeserializePrim` to codegen for `from_candid`, wrapping with an option, but still trapping on failure.
 - **Future PR**: adapt existing codegen `Serialization.deserialize_from_blob`, so that it produces an option, and does not trap on failure.


src/ir_def/check_ir.ml
```diff
@@ -528,9 +528,9 @@ let rec check_exp env (exp:Ir.exp) : unit =
       check (Show.can_show ot) "show is not defined for operand type";
       typ exp1 <: ot;
       T.text <: t
-    | SerializePrim ots, [exp1] ->
+    | SerializePrim ots, es ->
       check (T.shared (T.seq ots)) "debug_serialize is not defined for operand type";
-      typ exp1 <: T.seq ots;
+      List.iter2 (fun t e -> typ e <: t) ots es;
```
matthewhammer, line 533: @crusso FYI -- this is how the change to use a literal tuple in the ASTs is playing out at the moment.  In particular, I found myself generalizing this IR typing rule to permit any number of arguments.  If this change (or any other one) feels like the wrong direction, please let me know.  Seems okay to me, but I'm using your explanation of the backend code still, without knowing firsthand how it works yet.

matthewhammer: cc @ggreif

src/lowering/desugar.ml
```diff
@@ -142,10 +154,11 @@ and exp' at note = function
       I.PrimE (I.CastPrim (T.seq ts1, T.seq ts2), [exp e])
     | _ -> assert false
     end
-  | S.CallE ({it=S.AnnotE ({it=S.PrimE "serialize";_}, _);note;_}, _, e) ->
+  | S.CallE ({it=S.AnnotE ({it=S.PrimE "serialize";_}, _);note;_}, _,
+             {it=S.TupE es; _}) ->
```
matthewhammer, line 158: @crusso related to my other comment.  To make the IR type checker work for expansions of `to_candid` that become `serialize`, I tweaked the syntax and IR typing rule for `serialize` to be an n-ary expression node, like we chose for `to_candid`.

matthewhammer:
Hmm.  While it seems like the literal tuple syntax is fine for `to_candid` (since it is new syntax that we are defining), I still cannot seem to get the existing parsing and typing rules for `serialize` to work now, since they had expected to take a single value in terms of their typing and syntax.  Trying to adjust them is more challenging than I originally thought.

One problem is that we parse tuples and them forget them as tuples if they have a single member.  That's one headache, and I think that there are more too.

test/run/idl.mo
```diff
@@ -7,10 +7,10 @@ func deserUnit(x : Blob) : () = (prim "deserialize" : Blob -> ()) x;
 func serNats(x: Nat, y: Nat, z: Nat) : Blob = (prim "serialize" : (Nat,Nat,Nat) -> Blob) (x,y,z);
 func deserNats(x: Blob) : (Nat, Nat, Nat) = (prim "deserialize" : Blob -> (Nat,Nat,Nat)) x;
 
-func serText(x: Text) : Blob = (prim "serialize" : Text -> Blob) x;
+func serText(x: Text) : Blob = (prim "serialize" : Text -> Blob) (x,);
```
matthewhammer, line 10: Now learned that the parsing of `x` and `(x,)` is the same by virtue of [our parser ignoring these parens for single member tuples](https://github.com/dfinity/motoko/blob/cacf7742bd9e2c40f985fd29049c8c28847ac822/src/mo_frontend/parser.mly#L556).

src/ir_def/check_ir.ml
crusso: Instead of rewriting the implementations of SerializePrim, which might actually be fraught with peril, would it not be easier to just picky-back off the existing code and  
desugar the to_candid arguments sequence to a single argument with an explicit tuple (perhaps eliding the tuple for a unary argument (if necessary)).

src/lowering/desugar.ml
```diff
@@ -68,6 +68,20 @@ and exp' at note = function
     I.PrimE (I.RelPrim (!ot, o), [exp e1; exp e2])
   | S.ShowE (ot, e) ->
     I.PrimE (I.ShowPrim !ot, [exp e])
+  | S.ToCandidE (ot, es) ->
+    begin match !ot with
+    | T.Tup ts1 ->
+      I.PrimE (I.SerializePrim ts1, exps es)
+    | _ -> assert false
+    end
```
crusso, line 76:
```suggestion
  | S.ToCandidE (ot, es) ->
      I.PrimE (I.SerializePrim (List.map (typ es)), [tup es])
    end
```
Does this not work instead - also suggesting we don't need `ot`.

matthewhammer: I appreciate the suggestion.  I'm happy to try it out.  Thanks!

(I don't have enough intuition yet about the existing design to know what path is best overall.  Hopefully in time I will have a better sense of that for myself.)

crusso: Just because `ot` is declared in the pattern but never used in the rhs of the match.

matthewhammer: This is the only place where it's used, I guess?  Okay, I'll try.  Thanks!

src/ir_def/check_ir.ml
matthewhammer: Finally trying now.

src/lowering/desugar.ml
matthewhammer: Unfortunately, the suggestion does not compile, since `typ` and `tup` are not defined there.  I'm guessing that you mean `tupE` for `tup`, but I don't see any other place in that file that uses a function like `typ` (going from expressions to their types).  Rather, I had thought that was the role of `ot` in that constructor (to hold the types needed here).

src/ir_def/check_ir.ml
matthewhammer: So after reverting this check back to how it was, I recall why I changed it, to avoid this:

```
internals:420.24-420.58: IR type error [M0000], PrimE (SerializePrim)
 does not work with 0 arguments
Raised at Ir_def__Check_ir.error.(fun) in file "ir_def/check_ir.ml", line 95, characters 30-92
```

This comes quickly after building the compiler, after it's run on the `base` library.  I'm still not that familiar with all of the testing and making sub-steps, so I'm not certain how to sidestep this yet best, other than to simply generalize the line to handle the empty list.

src/lowering/desugar.ml
crusso: There is such a 'typ' function in `Ir_effect.ml`, and my suggestions should have been:
```
let es' = exps es in
I.PrimE (I.SerializePrim (List.map Ir_effect.typ es'), [tupE (exps es)])
```
Or similar.

Both source and ir expressions are type annotated, and you can access the type using `e.note.Note.typ` or similar (grep around for other examples).

Sorry I didn't see this earlier - just slack me if you get stuck on something.

src/lowering/desugar.ml
```diff
@@ -68,6 +68,20 @@ and exp' at note = function
...
   | S.ShowE (ot, e) ->
     I.PrimE (I.ShowPrim !ot, [exp e])
+  | S.ToCandidE (ot, es) ->
+    begin match !ot with
+    | T.Tup ts1 ->
+      I.PrimE (I.SerializePrim [T.Tup ts1], [tupE (exps es)])
+    | _ -> assert false
+    end
+  | S.FromCandidE (ot, e) ->
+    begin match !ot with
```
crusso, line 78: You almost certainly need to normalize !ot (or just the outer type of the entire expression) before testing the outer structure of the type, and then normalize the content again:

src/lowering/desugar.ml
```diff
@@ -68,6 +68,20 @@ and exp' at note = function
...
+    | T.Opt (T.Tup ts1) ->
+      I.PrimE (I.DeserializePrim ts1, [exp e])
+    | T.Opt t ->
+      I.PrimE (I.DeserializePrim [t], [exp e])
+    | _ -> assert false
+    end
```
crusso, line 84:
```suggestion
  | S.FromCandidE (ot, e) ->
    begin match T.normalize (!ot) with
    | T.Opt t1 -> 
      (match T.normalize t1 with
        | (T.Tup ts1) ->
          I.PrimE (I.DeserializePrim ts1, [exp e])
        | t1 -> I.PrimE (I.DeserializePrim [t1], [exp e]))
    | _ -> assert false
    end
```
(and similarly in ToCandidE unless you know that will already be a normalized T.Tup).

src/mo_frontend/typing.ml
```diff
@@ -972,6 +972,12 @@ and infer_exp'' env exp : T.typ =
       ot := t
     end;
     T.text
+  | ToCandidE (ot, exps) ->
+    let ts = List.map (infer_exp env) exps in
+    ot := T.Tup ts;
```
crusso, line 977: You need to also check the type is `shared`.

src/mo_frontend/typing.ml
```diff
@@ -972,6 +972,12 @@ and infer_exp'' env exp : T.typ =
...
+    T.Prim T.Blob
+  | FromCandidE (ot, exp1) ->
+    error env exp.at "MXXXX" "from_candid requires but is missing a known type (from context)"
```
crusso, line 980: You need to also check the type is `shared`

src/mo_frontend/typing.ml
```diff
@@ -1379,6 +1385,24 @@ and check_exp' env0 t exp : T.typ =
...
+    begin match t with
+    | T.Prim T.Blob -> begin
+        let ts = List.map (infer_exp env) exps in
+        ot := T.Tup ts;
+        T.Prim T.Blob
+      end
+    | _ ->
+      error env exp.at "MXXXX" "to_candid produces Blob, not type%a"
+        display_typ_expand t
+    end
```
crusso, line 1398: Instead of matching on the structure of `t` you should just check that `Blob` is subtype of `t` (which could be Any).
Also, you need to check the ts (or ot) all `shared`.

src/mo_frontend/typing.ml
```diff
@@ -1379,6 +1385,24 @@ and check_exp' env0 t exp : T.typ =
...
+  | FromCandidE (ot, exp1), t when T.shared t && T.is_opt t ->
```
crusso, line 1399:
```suggestion
  | FromCandidE (ot, exp1), t when T.shared t && T.is_opt t ->
```
Can you assume `t` is normalized here? If not, you need to `normalize` (or perhaps even `promote`) before checking `T.is_opt...` Maybe look if you can find some similar code elsewhere

crusso: Actually, I think you can assume it's normalized here. Nvm,

src/mo_frontend/typing.ml
crusso: And only check infer the types of subexpressions when `not env.pre` (see ShowE above for an example).

crusso:
@matthewhammer we should get together and nail this next week - I was distracted by my own PR and reviewing some code for @ggreif. Sorry for the lack of attention.

matthewhammer:
@crusso @ggreif I'm back from PTO as of today.  Looking at this PR again now.  Happy to meet soon.

src/lowering/desugar.ml
```diff
@@ -68,6 +68,22 @@ and exp' at note = function
...
+      (optE (primE (I.SerializePrim ts1) (exps es))).it
+    | _ -> assert false
+    end
+  | S.FromCandidE (ot, e) ->
+    (* begin match T.normalize note.Note.typ with *) (* not a T.Tup _ *)
+    begin match T.normalize (!ot) with
+    | T.Opt t ->
+      begin match T.normalize t with
+        | T.Tup ts1 -> I.PrimE (I.DeserializePrim ts1, [exp e])
+        | t (* XXX check me. *) -> I.PrimE (I.DeserializePrim [t], [exp e])
```
matthewhammer, line 83: @crusso I needed line just above `(XXX check me)` to get the test `idl.mo` to pass without triggering an assertion failure here.

src/lowering/desugar.ml
```diff
@@ -68,6 +68,22 @@ and exp' at note = function
...
+      end
```
crusso, line 83:
```suggestion
       I.PrimE (I.DeserializePrim (T.as_seq t), [exp e])
```
Is a more concise way to do the same (T.as_seq normalizes first - I checked).

doc/modules/language-guide/examples/grammar.txt
```diff
@@ -169,6 +169,8 @@
     'actor' <exp_plain>
     'not' <exp_un>
     'debug_show' <exp_un>
+    TO_C'and'<id> '(' <list(<exp>, ',')> ')'
+    FROM_C'and'<id> <exp_un>
```
crusso, line 173: To fix this, you'll need to adjust the grammar massaging sed script in motoko/src/grammar.sed to replace TO_CANDID and FROM_CANDID by lc versions, probably before the rewrite of AND to avoid interference.

src/ir_def/construct.ml
```diff
@@ -104,6 +104,7 @@ let primE prim es =
     | SerializePrim _ -> T.blob
     | SystemCyclesAvailablePrim
     | SystemCyclesAcceptPrim -> T.nat
+    | DeserializePrim ts -> T.Tup ts
```
crusso, line 107:
```suggestion
    | DeserializePrim ts -> T.seq ts
```

crusso: T.seq avoids creating the one-tuple.

src/lowering/desugar.ml
```diff
@@ -68,6 +68,22 @@ and exp' at note = function
...
+    | T.Tup ts -> (primE (I.SerializePrim [T.Tup ts]) [arg]).it
+    | _ -> assert false
+    end
+  | S.FromCandidE e ->
+    begin match T.normalize note.Note.typ with
+    | T.Opt t ->
+      begin match T.normalize t with
+        | T.Tup ts1 -> (optE (primE (I.DeserializePrim ts1) [exp e])).it
+        | t -> (optE (primE (I.DeserializePrim [t]) [exp e])).it
+      end
```
crusso, line 84:
```suggestion
      optE (primE (I.DeserializePrim (T.as_seq t)) [exp e])).it
```
Should suffice.

src/mo_def/syntax.ml
```diff
@@ -253,6 +256,7 @@ let (@?) it at = Source.({it; at; note = empty_typ_note})
 let (@!) it at = Source.({it; at; note = Type.Pre})
 let (@=) it at = Source.({it; at; note = None})
 
+let is_tuple_lit e = match e.Source.it with TupE _ -> true | _ -> false
```
crusso, line 259: Is this still necessary? If not, delete.

src/mo_frontend/typing.ml
```diff
@@ -1379,6 +1389,23 @@ and check_exp' env0 t exp : T.typ =
       warn env exp.at "M0155" "operator may trap for inferred type%a"
         display_typ_expand t;
     t
+  | ToCandidE exps, _ ->
+    if not env.pre then begin
+      let ts = List.map (infer_exp env) exps in
+      if not (T.sub (T.Prim T.Blob) t) then
+        error env exp.at "MXXXX" "to_candid produces a Blob that is not a subtype of %a"
+          display_typ_expand t;
+      if not (T.shared (T.Tup ts)) then
```
crusso, line 1398:
```suggestion
      if not (T.shared (T.seq ts)) then
```

src/mo_frontend/typing.ml
```diff
@@ -1379,6 +1389,23 @@ and check_exp' env0 t exp : T.typ =
...
+        error env exp.at "MXXXX" "to_candid argument types be shared, but instead found %a"
+          display_typ_expand (T.Tup ts);
```
crusso, line 1400:
```suggestion
          display_typ_expand (T.seq ts);
```

src/ir_def/check_ir.ml
```diff
@@ -528,6 +528,8 @@ let rec check_exp env (exp:Ir.exp) : unit =
       check (Show.can_show ot) "show is not defined for operand type";
       typ exp1 <: ot;
       T.text <: t
+    | SerializePrim ots, [] -> (* unit argument (no arguments). *)
+      T.blob <: t
```
crusso, line 532: Is this necessary? Or left-over from a previous experiment

src/mo_frontend/typing.ml
```diff
@@ -972,6 +972,16 @@ and infer_exp'' env exp : T.typ =
       ot := t
     end;
     T.text
+  | ToCandidE exps ->
+    if not env.pre then begin
+        let ts = List.map (infer_exp env) exps in
+        if not (T.shared (T.Tup ts)) then
```
crusso, line 978:
```suggestion
        if not (T.shared (T.seq ts)) then
```

src/mo_frontend/typing.ml
```diff
@@ -972,6 +972,16 @@ and infer_exp'' env exp : T.typ =
...
+          error env exp.at "MXXXX" "to_candid argument types be shared, but instead found %a"
+            display_typ_expand (T.Tup ts);
```
crusso, line 980:
```suggestion
            display_typ_expand (T.seq ts);
```

src/mo_frontend/typing.ml
```diff
@@ -972,6 +972,16 @@ and infer_exp'' env exp : T.typ =
       ot := t
     end;
     T.text
+  | ToCandidE exps ->
+    if not env.pre then begin
+        let ts = List.map (infer_exp env) exps in
+        if not (T.shared (T.Tup ts)) then
+          error env exp.at "MXXXX" "to_candid argument types be shared, but instead found %a"
```
crusso, line 979:
```suggestion
          error env exp.at "MXXXX" "to_candid argument must have shared type, but instead found non-shared type %a"
```

crusso (commented):
Found the bug and some minor nits.

src/ir_def/construct.ml
crusso: I think that this was the bug you asked about on slack.

matthewhammer: This overcomes the type mismatch in the IR, yes.

However, now I get another error, since morally, we *are* encoding a 1-tuple when we invoke `to_candid "Foo"`.

Namely, `IDL error: unexpected IDL type when parsing Text`, which I suspect could be because the candid for `Text` and `(Text,)` are not the same?

Do I need to also special-case the `to_candid` desugaring to avoid 1-tuples there?

I had a version like that when I was experimenting, but it seemed too complex to case analyze for that.  Now I wonder if that's what you had imagined?

matthewhammer: Or maybe the other suggestion compensates:
https://github.com/dfinity/motoko/pull/3155/files#r860161312

Let me try.

matthewhammer: Nevermind.  I tried some other variations of the test program.

The deserialization issue that I see now is not specific to 1-tuples.  It also happens for the `(Nat, Nat, Nat)` case:

```
+IDL error: too few arguments NNN
+Error: failed to run main module `_out/idl.wasm`
+
+Caused by:
+    1: wasm trap: unreachable
+       wasm backtrace:
+         0: @deserialize<NNN>
+         1: deserNats
+         2: init
+         3: _start
```

Will push my code now.

src/mo_frontend/typing.ml
matthewhammer: I get that we are uniformly making this change (and I will here), but this chance is stylistic only, right?

In other words, the two program versions are equivalent under all inputs, right?

test/run/idl.mo
```diff
@@ -1,34 +1,37 @@
...
 Prim.debugPrint(debug_show (serUnit ()));
 Prim.debugPrint(debug_show (serNats (1,2,3)));
 Prim.debugPrint(debug_show (serText "Hello World!"));
 Prim.debugPrint(debug_show (serBool true));
 Prim.debugPrint(debug_show (serBool false));
 
+// unit and triples
+
+assert (?() == deserUnit (serUnit ()));
+assert(?(1,2,3) == (deserNats (serNats (1,2,3)) : ?(Nat,Nat,Nat)));
```
matthewhammer, line 26: @crusso This line fails

```
+IDL error: too few arguments NNN
+Error: failed to run main module `_out/idl.wasm`
+
+Caused by:
+    1: wasm trap: unreachable
+       wasm backtrace:
+         0: @deserialize<NNN>
+         1: deserNats
+         2: init
+         3: _start
```

And I suspect that the issue is that we are desugaring `to_candid` so that it always introduces a single-component message, in this case a 3-tuple holding three `Nat`s.

This logic is mis-matched:

```
 | S.ToCandidE es ->
    (* desugaring to_candid introduces explicit IR tuple: *)
    let arg = tupE (exps es) in
    begin match T.normalize arg.note.Note.typ with
    | T.Tup ts -> (primE (I.SerializePrim [T.Tup ts]) [arg]).it
    | _ -> assert false
    end
```

Since we are not doing the same thing when deserializing now.

test/run/idl.mo
```diff
@@ -1,34 +1,37 @@
...
-func serBool(x: Bool) : Blob = (prim "serialize" : Bool -> Blob) x;
-func deserBool(x: Blob) : Bool = (prim "deserialize" : Blob -> Bool) x;
+func serBool(x: Bool) : Blob = to_candid (x);
+func deserBool(x: Blob) : ?(Bool) = from_candid x;
 
+func serText(x: Text) : Blob = to_candid (x);
+func deserText(x: Blob) : ?(Text) = from_candid x;
 
 Prim.debugPrint(debug_show (serUnit ()));
 Prim.debugPrint(debug_show (serNats (1,2,3)));
```
matthewhammer, line 18: @crusso This line prints 

`"\44\49\44\4C\00\03\7D\7D\7D\01\02\03"`

So, a message vector of three components, each is a `Nat`.

test/run/idl.mo
matthewhammer: There's something extremely confusing about the games that DeserializePrim is playing with its argument "list".

It is written as if it could work in an n-ary way, but it cannot, obviously.  All of those code paths (for typing the IR, and for generating code) are missing!  I think you warned me about this, but I didn't appreciate it earlier.

So we are *always* using this tuple-encoding trick, it seems.  But there's something strange happening that's preventing the trick from working here.

test/run/idl.mo
matthewhammer: (**Not** a single-component message, with 3-tuple of `Nat`s)

src/lowering/desugar.ml
```diff
@@ -68,6 +68,18 @@ and exp' at note = function
     I.PrimE (I.RelPrim (!ot, o), [exp e1; exp e2])
   | S.ShowE (ot, e) ->
     I.PrimE (I.ShowPrim !ot, [exp e])
+  | S.ToCandidE es ->
+    (* desugaring to_candid introduces explicit IR tuple: *)
+    let arg = tupE (exps es) in
+    begin match T.normalize arg.note.Note.typ with
+    | T.Tup ts -> (primE (I.SerializePrim [T.Tup ts]) [arg]).it
+    | _ -> assert false
+    end
```
crusso, line 77:
```suggestion
    (* desugaring to_candid introduces explicit IR tuple: *)
    let arg = tupE (exps es) in
    begin match T.normalize arg.note.Note.typ with
    | T.Tup ts -> (primE (I.SerializePrim [T.Tup ts]) [arg]).it
    | _ -> assert false
    end
```

crusso: If you look at the typing of SerializePrim in check_ir.ml, and the way it is used in codegen/compile.ml,
I think you'll come to same conclusion.

crusso: It took me a few attemps to get this right because I used `tupE args` instead of `seqE args` - the latter matches at the term leve of the T.seq functions used at the typ-level, ()avoiding 1-tuples)

crusso: https://github.com/dfinity/motoko/pull/3214 is a branch with the necessary change (also suppressing the -v flag in the test)

test/run/idl.mo
crusso: > I actually do not see another way to get this to work

There is, I think, see below.

src/lowering/desugar.ml
matthewhammer: Thanks for digging into this and finding a more direct solution.  I can see how this would work, but the subtle balance of Candid tuples versus "meta-level" Candid value sequences throughout the abstract syntax feels complex and unresolved for me, probably because I'm still not familiar enough, though it's making sense as we go, and this helps a lot.  (Again, Thanks!)

I'll stare at your solution today and try to appreciate more what still feels unresolved for me.

doc/modules/language-guide/examples/grammar.txt
matthewhammer: Finally got to this.

I followed your advice, but also needed to hoist above 'ID' as well as 'AND'.

src/mo_frontend/typing.ml
```diff
@@ -1379,6 +1389,23 @@ and check_exp' env0 t exp : T.typ =
...
         display_typ_expand t;
     t
+  | ToCandidE exps, _ ->
+    if not env.pre then begin
+      let ts = List.map (infer_exp env) exps in
+      if not (T.sub (T.Prim T.Blob) t) then
+        error env exp.at "MXXXX" "to_candid produces a Blob that is not a subtype of %a"
+          display_typ_expand t;
+      if not (T.shared (T.seq ts)) then
+        error env exp.at "MXXXX" "to_candid argument types be shared, but instead found %a"
```
crusso, line 1399:
```suggestion
          error env exp.at "MXXXX" "to_candid argument must have shared type, but instead found non-shared type %a"
```
(to match above)

src/mo_frontend/typing.ml
```diff
@@ -972,6 +972,16 @@ and infer_exp'' env exp : T.typ =
       ot := t
     end;
     T.text
+  | ToCandidE exps ->
+    if not env.pre then begin
+        let ts = List.map (infer_exp env) exps in
+        if not (T.shared (T.seq ts)) then
+          error env exp.at "MXXXX" "to_candid argument must have shared type, but instead found non-shared type %a"
```
crusso, line 979:
```suggestion
          error env exp.at "MXXXX" "to_candid argument must have shared type, but instead has non-shared type %a"
```

ByronBecker:
@crusso @matthewhammer  I know the team has a million different priorities, but I just wanted to follow up regarding a conservative estimate of this feature so that I can plan my own project roadmap around its expect delivery.

I'm dynamically spinning up canisters, and would love to have this feature so that I can use a record for a constructor some of my auto-scaling canister actors. Here's what one of those canisters looks like.

```
shared ({ caller = owner }) actor class UserCanister({ 
  pk: Text;
  owners: ?[Principal];
}) {
```

The issue is that I would like to dynamically upgrade my canisters through my indexing canister meaning that I need the record -> `Blob` or [Nat8]` conversion for the `arg` parameter of the `install_code` method.
```
install_code : shared {
      mode : { #install; #reinstall; #upgrade};
      canister_id : Principal;
      wasm_module : [Nat8];
      arg : [Nat8];
    } -> async ();
```

Since there does not yet exist a record -> `Blob` or [Nat8]` conversion to do so, I am either forced to do my canister upgrades purely through the frontend with JavaScript targeting the individual canisters directly, or to simplify my existing canister constructors to take just a single argument which can be converted to Blob.

crusso:
> @crusso @matthewhammer I know the team has a million different priorities, but I just wanted to follow up regarding a conservative estimate of this feature so that I can plan my own project roadmap around its expect delivery.

We are hoping to release this within the next or maybe subsequent release (so weeks, not months). Maybe by the end of the week if you care comfortable installing the compiler yourself (rather than waiting for dfx to include it in a release).

> I'm dynamically spinning up canisters, and would love to have this feature so that I can use a record for a constructor some of my auto-scaling canister actors. Here's what one of those canisters looks like.
> 
> ```
> shared ({ caller = owner }) actor class UserCanister({ 
>   pk: Text;
>   owners: ?[Principal];
> }) {
> ```
> 
> The issue is that I would like to dynamically upgrade my canisters through my indexing canister meaning that I need the record -> `Blob`/`[Nat8]` conversion for the `arg` parameter of the `install_code` method.
> 
> ```
> install_code : shared {
>       mode : { #install; #reinstall; #upgrade};
>       canister_id : Principal;
>       wasm_module : [Nat8];
>       arg : [Nat8];
>     } -> async ();
> ```
> 
> Since there does not yet exist a record -> `Blob` or [Nat8]` conversion to do so, I am either forced to do my canister upgrades purely through the frontend with JavaScript targeting the individual canisters directly, or to simplify my existing canister constructors to take just a single argument which can be converted to Blob.


You don't actually need to wait for this surface syntax to land and could access some of it already by jumping through some undocumented hoops (or work around it more safely by abusing `ExperimentalInternetComputer` at some expense). I'd be happy to share instructions privately by email (or you can actually figure it out from the revised tests in this PR).

Note that there is a restriction that you need to know the (non-generic) static type of the thing that you want to serialize, but it sounds like that may be the case in your application. Is the record type fixed at compile time?

BTW, I'd use Blob. not [Nat8] for the wasm blob and argument blob. A Blob is much more compact (every Nat8 takes 4 bytes in our current representation).

> @crusso @matthewhammer I know the team has a million different priorities, but I just wanted to follow up regarding a conservative estimate of this feature so that I can plan my own project roadmap around its expect delivery.
> 
> I'm dynamically spinning up canisters, and would love to have this feature so that I can use a record for a constructor some of my auto-scaling canister actors. Here's what one of those canisters looks like.
> 
> ```
> shared ({ caller = owner }) actor class UserCanister({ 
>   pk: Text;
>   owners: ?[Principal];
> }) {
> ```
> 
> The issue is that I would like to dynamically upgrade my canisters through my indexing canister meaning that I need the record -> `Blob`/`[Nat8]` conversion for the `arg` parameter of the `install_code` method.
> 
> ```
> install_code : shared {
>       mode : { #install; #reinstall; #upgrade};
>       canister_id : Principal;
>       wasm_module : [Nat8];
>       arg : [Nat8];
>     } -> async ();
> ```
> 
> Since there does not yet exist a record -> `Blob` or [Nat8]` conversion to do so, I am either forced to do my canister upgrades purely through the frontend with JavaScript targeting the individual canisters directly, or to simplify my existing canister constructors to take just a single argument which can be converted to Blob.

ByronBecker:
@crusso Thanks for offering the experimental featureset & "undocumented" approach, but I'm building a developer tool so I'm looking to expose this functionality to end users. Therefore, it's probably best that I wait until this feature is released.

> We are hoping to release this within the next or maybe subsequent release (so weeks, not months).

Ideally I'd like to push this out during the hackathon (early June), but the clarity is nice so I can plan between building a demo and additional features myself vs. actually releasing a pre-alpha version for developers to play with during the hackathon. Regardless, happy that this is on the order of a month away from being released 👍 

> Note that there is a restriction that you need to know the (non-generic) static type of the thing that you want to serialize, but it sounds like that may be the case in your application. Is the record type fixed at compile time?

Yes, the use case is for the developer to define a static record type as the parameters passed to the constructor of a canister actor. This type could be modified through upgrades over time, but is fixed at compile time.

> A Blob is much more compact (every Nat8 takes 4 bytes in our current representation).
Thanks for the tip - will make these changes!

crusso:
@matthewhammer feel free to merge #3232 into this branch if it helps you make more progress.

src/mo_frontend/typing.ml
```diff
@@ -982,6 +982,16 @@ and infer_exp'' env exp : T.typ =
       ot := t
     end;
     T.text
+  | ToCandidE exps ->
+    if not env.pre then begin
+        let ts = List.map (infer_exp env) exps in
+        if not (T.shared (T.seq ts)) then
+          error env exp.at "MXXXX" "to_candid argument must have shared type, but instead has non-shared type %a"
```
crusso, line 989: You'll need to choose fresh error codes for the MXXXX placeholders, and either make sure some test flags them or add them to a list of ignored error codes by following instructions. Writing a test is easier...

Maybe just grep for an existing code to see how...

matthewhammer: Ah, I forgot that this was remaining.  Will do.

matthewhammer: Done.

I also recall a CI process that checks for failure test coverage.  Next I'll look at how to do negative failing tests to cover each new error number.

crusso: test/fail is the natural home

test/run/idl.mo
```diff
@@ -1,33 +1,35 @@
 //MOC-ENV MOC_UNLOCK_PRIM=yesplease
 import Prim "mo:⛔";
 
-func serUnit() : Blob = (prim "serialize" : () -> Blob) ();
-func deserUnit(x : Blob) : () = (prim "deserialize" : Blob -> ()) x;
+func serUnit() : Blob = to_candid ();
```
crusso, line 4: Hmm, its a bit weird that this file is testing both the new syntax and the old prims. Maybe revert this one to what it was, and write an additional one, derived from it, that uses the new syntax throughout (even for the array examples at the end that you kept in the old format).

crusso (commented):
See comment but, itherwise, I think this is good to go.

crusso:
@Matthew, I addressed my own comment by restoring idl.mo and adding idl-ops.mo. Let's merge.

crusso (approved):

test/run/idl.mo
matthewhammer: Okay.

It's not weird if you think (like I did until now) that the new surface syntax was replacing these undocumented primitives.  Why even expose them now?  Is there some benefit to having both approaches?

In any case, if we have both, it makes sense to test them both, and to separate the tests.  Thanks.

PR #3159, opened 2022-03-16
Inline prim-wrapping functions

nomeata:
The prelude/prim.mo is full of functions simply wrapping a prim, e.g.

    func int64ToNat64(n : Int64) : Nat64 = (prim "num_wrap_Int64_Nat64" : Int64 -> Nat64) n;

generating a Wasm function for them and calling them is absurdly expensive
when the prim is just a simple Wasm instruction. Also, it requires boxing
and unboxing arguments and results.

So we recognize such functions when creating the `const` summary, and
use the prim directly when calling such function.


nomeata:
Some sneak previews:

```diff
-    i32.const 0
     local.get $n1_wrapped
-    call $intToInt64
-    call $unbox_i64
+    call $Int->Int64
…
-  (func $intToInt64 (type 14) (param $clos i32) (param $n i32) (result i32)
-    local.get $n
-    call $Int->Int64
-    call $box_i64)
```


Todo:

* [ ] Maybe extract the refactoring (`comple_prim_invocation`) in a separate PR, to keep diff and conflicts small
* [ ] Previously, the backend assumed that prims are always in functions, and does not care if they are large. Now large prims should be wrapped in functions in the backend (using share_code).
* [ ] why does it not kick in in `run/float-ops.mo`

nomeata:
This will make @ggreif happy I assume :-)

nomeata:
I guess it would have been cleaner to have this as a Ir-to-Ir pass, instead of complicating the backend even further. But it fits there nicely right now.

nomeata:
The next big win for numeric heavy code is to have non-recursive `let` in the IR, because that makes it easier to keep things unboxed even across lets. But that’s a slightly larger undertaking.

src/codegen/compile.ml
```diff
@@ -7526,8 +7546,39 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
+           code1 ^^
+           compile_exp_as env ae (StackRep.of_arity 0) e2 ^^
+           code2
+         | 1, _ ->
+           compile_prim_invocation env ae prim [e2] at
+         | n, PrimE (TupPrim, es) ->
+           assert (List.length es = n);
+           compile_prim_invocation env ae prim es at
+         | _, _ ->
+           (* ugly case; let's just call this as a function for now *)
```
crusso, line 7578: What would you do here. Evaluate the arguments once and project its components?

nomeata: Yes, something like that. Needs a bunch of locals. Overall, very rare I'd expect.

crusso (approved):
LGTM and great to have. @ggreif should review too.

crusso (changes requested):
Approved too soon - I guess  you still need to make large prims shared...

nomeata:
Good that it wasn’t merged yet, I noticed that some prims (in particular `RetPrim`) are not control-flow-side-effect-free in a way that inlining is safe. fixed.

Otherwise there aren’t that many prim implementations where I would be too worried about code duplication (compared to what we do else in the compiler, not necessarily to the completely thought through global optimum). Many complex prims are calls into the Rust RTS anyways, so they are fine.

I think I’ll put the cycle stuff into separate fuctions and move them out of the huge `compile_exp` block into the `IC` module.

nomeata:
Sharing the code for the cycle prims is now in https://github.com/dfinity/motoko/pull/3162.

Any other ones we should definitely share?

ggreif:
> Sharing the code for the cycle prims is now in #3162.
> 
> Any other ones we should definitely share?

So both "Todo"s above can be checked?

nomeata:
Perf numbers look good, I’m in favor of merging now, and sharing more primitive functions when and if we have evidence that they are better shared.

src/codegen/compile.ml
```diff
@@ -8931,17 +8982,35 @@ and compile_decs env ae decs captured_in_body : VarEnv.t * scope_wrap =
...
-          then fatal "internal error: const \"%s\": captures \"%s\", not found in static environment\n" name v
-        ) (Freevars.M.keys (Freevars.exp e));
-        compile_exp_as env ae (StackRep.of_arity (List.length return_tys)) e in
-      FuncDec.closed env sort control name args mk_body return_tys exp.at
+    let fun_rhs =
+
+      (* a few prims cannot be safely inlined *)
+      let inlineable_prim = function
+      | RetPrim -> false
+      | BreakPrim _ -> false
```
crusso, line 8990: What about similar ThrowPrim and perhaps even AssertPrim - are there others?

crusso: Actually, any prim whose implementation/rhs contains such prims should not be naively inlined, unless we, eg,  add a block and an compile RetPrim as Break etc. But perhaps we don't have such beasts.

crusso: And it looks like ThrowPrim no longer occurs in the final IR - I guess I get rid of it in the async/await translation...

src/codegen/compile.ml
crusso: > Note that the code falls though to a normal function call; it doesn't crash.

Do you mean a normal indirect or direct function call. I think because the next pattern matches on Const.Complicated (not `_`), it will actually fall through to the indirect call case. Is that what you intended?

src/codegen/compile.ml
nomeata: > Actually, any prim whose implementation/rhs contains such prims should not be naively inlined

Which is precisely the point of the `inlineable_prim` function…

`ThrowPrim` _should_ be away at this point, but better assert that explicitly…

```suggestion
      | BreakPrim _ -> false
      | ThrowPrim _ -> assert false
```

src/codegen/compile.ml
crusso: Ah, ok, I think its fine. (Just tested it myself).

crusso: Actually, not it's not. In rare cases I think this will generate indirect calls that were previously direct, because of the bad fall-thru. See #3196 for a repro and suggested fix.

crusso (approved):

nomeata:
Hmm, the instruction count improvement was less than I hoped for. Maybe function calls are just too cheap :-)

With the sha256 test case in #3206 included I measure
```
In terms of gas, 3 tests improved and the mean change is -0.2%.
In terms of size, 2 tests regressed, 1 tests improved and the mean change is +0.1%.
```
which is also quite small. Maybe together with a possible non-recursive let (or another mechanism to get unboxed variables) this will be better.

PR #3161, opened 2022-03-16
Make module `Serialization` parametrised in the output streaming direction

ggreif:
This is a prelude to #3149. It doesn't change the generated code, just introduces a few holes in the serialiser that can be plugged with code fragments from a `Stream`. Later there will be different `Stream` implementations, but for now we only have the legacy `BumpStream` one. (Actually, there is a minor change in the generated code, which reflects how the stream now `reserve`s a small scratch area for writing, whereas previously it would write right away, and bump the pointer thereafter.)

**Big caveat**: The generated functions coming from `Serialization` (by `Func.share_code`) are still named the same. To make the naming sane, the `Stream` needs to bring its own name plumber. Since we only instantiate once in this PR, this is currently not a problem.

TODO: We definitely need two more interfaces to make this work with generic streams:
- `setup`: create the stream token
- `destroy`: flush the stream and terminate the token.

These are not mandatory in this PR, but when the first other stream appears.

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -4174,16 +4174,22 @@ module type Stream = sig
   val write_unsigned : E.t -> G.t -> G.t -> G.t
   val write_signed : E.t -> G.t -> G.t -> G.t
 
-  (* opportunity to flush or update, stream token is on  stack *)
+  (* opportunity to flush or update, stream token is on stack *)
   val checkpoint : E.t -> G.t -> G.t
+
+  (* reserve a small fixed number of bytes in the stream and return an address to it *)
+  val reserve : E.t -> G.t -> int32 -> G.t
 end
 
 
 module BumpStream : Stream = struct
   let advance_data_buf get_data_buf =
-    get_data_buf ^^ G.i (Binary (Wasm.Values.I32 I32Op.Add)) ^^ (G.setter_for get_data_buf)
+    get_data_buf ^^ G.i (Binary (Wasm.Values.I32 I32Op.Add)) ^^ G.setter_for get_data_buf
 
   let checkpoint _env get_data_buf = G.setter_for get_data_buf
+
+  let reserve _env get_data_buf bytes =
+    get_data_buf ^^ get_data_buf ^^ compile_add_const bytes ^^ G.setter_for get_data_buf
 
   let write_word_leb env get_data_buf code =
     let set_word, get_word = new_local env "word" in
@@ -4666,13 +4672,13 @@ module MakeSerialization (Strm : Stream) = struct
         get_x ^^ Float.unbox env ^^
         G.i (Store {ty = F64Type; align = 0; offset = 0l; sz = None}) ^^
         (*FIXME: reserve*)compile_unboxed_const 8l ^^
-        get_data_buf ^^ G.i (Binary (Wasm.Values.I32 I32Op.Add)) ^^ (G.setter_for get_data_buf)
+        get_data_buf ^^ G.i (Binary (Wasm.Values.I32 I32Op.Add)) ^^ G.setter_for get_data_buf
       | Prim (Int64|Nat64) ->
         get_data_buf ^^
         get_x ^^ BoxedWord64.unbox env ^^
         G.i (Store {ty = I64Type; align = 0; offset = 0l; sz = None}) ^^
         (*FIXME: reserve*)compile_unboxed_const 8l ^^
-        get_data_buf ^^ G.i (Binary (Wasm.Values.I32 I32Op.Add)) ^^ (G.setter_for get_data_buf)
+        get_data_buf ^^ G.i (Binary (Wasm.Values.I32 I32Op.Add)) ^^ G.setter_for get_data_buf
       | Prim (Int32|Nat32) ->
         write_word32 (get_x ^^ BoxedSmallWord.unbox env)
       | Prim Char ->
@@ -4682,7 +4688,7 @@ module MakeSerialization (Strm : Stream) = struct
         get_x ^^ TaggedSmallWord.lsb_adjust Nat16 ^^
         G.i (Store {ty = I32Type; align = 0; offset = 0l; sz = Some Wasm.Types.Pack16}) ^^
         (*FIXME: reserve*)compile_unboxed_const 2l ^^
-        get_data_buf ^^ G.i (Binary (Wasm.Values.I32 I32Op.Add)) ^^ (G.setter_for get_data_buf)
+        get_data_buf ^^ G.i (Binary (Wasm.Values.I32 I32Op.Add)) ^^ G.setter_for get_data_buf
       | Prim (Int8|Nat8) ->
         write_byte (get_x ^^ TaggedSmallWord.lsb_adjust Nat8)
       | Prim Bool ->
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -4606,16 +4606,14 @@ module MakeSerialization (Strm : Stream) = struct
       let set_ref_buf = G.setter_for get_ref_buf in
 
       (* Some combinators for writing values *)
-      let    [write_word;     write_word32;  write_byte; write_blob; write_text; write_unsigned;   write_signed] =
-        Strm.[write_word_leb; write_word_32; write_byte; write_blob; write_text; write_bignum_leb; write_bignum_sleb]
-        |> List.map (fun f -> f env get_data_buf) in
+      let open Strm in
 
       let write env t =
         get_data_buf ^^
         get_ref_buf ^^
         serialize_go env t ^^
         set_ref_buf ^^
-        Strm.checkpoint env get_data_buf
+        checkpoint env get_data_buf
       in
 
       let write_alias write_thing =
@@ -4627,18 +4625,18 @@ module MakeSerialization (Strm : Stream) = struct
         G.if0
         begin
           (* This is the real data *)
-          write_byte (compile_unboxed_const 0l) ^^
+          write_byte env get_data_buf (compile_unboxed_const 0l) ^^
           (* Remember the current offset in the tag word *)
           get_x ^^ get_data_buf ^^ Heap.store_field Tagged.tag_field ^^
           (* Leave space in the output buffer for the decoder's bookkeeping *)
-          write_word32 (compile_unboxed_const 0l) ^^
-          write_word32 (compile_unboxed_const 0l) ^^
+          write_word_32 env get_data_buf (compile_unboxed_const 0l) ^^
+          write_word_32 env get_data_buf (compile_unboxed_const 0l) ^^
           (* Now the data, following the object field mutbox indirection *)
           write_thing ()
         end
         begin
           (* This is a reference *)
-          write_byte (compile_unboxed_const 1l) ^^
+          write_byte env get_data_buf (compile_unboxed_const 1l) ^^
           (* Sanity Checks *)
           get_tag ^^ compile_eq_const Tagged.(int_of_tag MutBox) ^^
           E.then_trap_with env "unvisited mutable data in serialize_go (MutBox)" ^^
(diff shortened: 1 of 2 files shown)


src/codegen/compile.ml
```diff
@@ -4529,39 +4592,18 @@ module Serialization = struct
...
-      in
+      let write_word, write_word32, write_byte, write_blob, write_text, write_unsigned, write_signed = Strm.(write_word_leb env get_data_buf, write_word_32 env get_data_buf, write_byte env get_data_buf, write_blob env get_data_buf, write_text env get_data_buf, write_unsigned env get_data_buf, write_signed env get_data_buf) in
 
       let write env t =
         get_data_buf ^^
         get_ref_buf ^^
         serialize_go env t ^^
         set_ref_buf ^^
-        set_data_buf
+        G.setter_for get_data_buf
```
ggreif, line 4606: this breaks the `Stream` abstraction

src/codegen/compile.ml
```diff
@@ -4529,39 +4603,18 @@ module Serialization = struct
...
-        G.i (Store {ty = I32Type; align = 0; offset = 0l; sz = None}) ^^
-        compile_unboxed_const Heap.word_size ^^ advance_data_buf
-      in
-
-      let write_byte code =
-        get_data_buf ^^ code ^^
-        G.i (Store {ty = I32Type; align = 0; offset = 0l; sz = Some Wasm.Types.Pack8}) ^^
-        compile_unboxed_const 1l ^^ advance_data_buf
-      in
+      let write_word, write_word32, write_byte, write_blob, write_text, write_unsigned, write_signed = Strm.(write_word_leb env get_data_buf, write_word_32 env get_data_buf, write_byte env get_data_buf, write_blob env get_data_buf, write_text env get_data_buf, write_unsigned env get_data_buf, write_signed env get_data_buf) in
```
ggreif, line 4610: Use `List.map`

src/codegen/compile.ml
```diff
@@ -4164,7 +4164,81 @@ module RTS_Exports = struct
 
 end (* RTS_Exports *)
 
-module Serialization = struct
+module type Stream = sig
+  (*               env    ref    code *)
+  val write_byte : E.t -> G.t -> G.t -> G.t
```
nomeata, line 4169: It would be good to note which arguments are guaranteed to be used once (the `code`), and which ones these function like to duplicate and should ideally be a simple `get_foo` (the `ref`).

At some point we might refactor this file to let the type system distinguish the two. Maybe.

src/codegen/compile.ml
```diff
@@ -4164,7 +4164,81 @@ module RTS_Exports = struct
...
+  val write_word_leb : E.t -> G.t -> G.t -> G.t
+  val write_word_32 : E.t -> G.t -> G.t -> G.t
+  val write_blob : E.t -> G.t -> G.t -> G.t
+  val write_text : E.t -> G.t -> G.t -> G.t
+  val write_unsigned : E.t -> G.t -> G.t -> G.t
```
nomeata, line 4174:
```suggestion
  val write_bignum_unsigned : E.t -> G.t -> G.t -> G.t
```
maybe?

src/codegen/compile.ml
nomeata: Cant we use the module system to make this prettier, @crusso? (map won't work if the result types are different)

src/codegen/compile.ml
```diff
@@ -4164,7 +4164,81 @@ module RTS_Exports = struct
...
+  val write_signed : E.t -> G.t -> G.t -> G.t
+
+  (* opportunity to flush or update, stream token is on stack *)
+  val checkpoint : E.t -> G.t -> G.t
```
nomeata, line 4178: I don't understand this. Do we need it?

src/codegen/compile.ml
```diff
@@ -4164,7 +4164,81 @@ module RTS_Exports = struct
...
+  (* reserve a small fixed number of bytes in the stream and return an address to it *)
+  val reserve : E.t -> G.t -> int32 -> G.t
+end
+
+
+module BumpStream : Stream = struct
+  let advance_data_buf get_data_buf =
+    get_data_buf ^^ G.i (Binary (Wasm.Values.I32 I32Op.Add)) ^^ G.setter_for get_data_buf
+
+  let checkpoint _env get_data_buf = G.setter_for get_data_buf
```
nomeata, line 4189: Haven't seen `setter_for` yet … hmm, feels a bit hacky to me. Maybe really worth introducing a type to abstract over things that can be set and get'ed.

src/codegen/compile.ml
ggreif: But here the types are the same, so a homogeneous `List` will do. This looks like a _very_ specific use case for a module system.

Also I did it this way to minimise the changes relative to the original code. Sounded better to me to apply boring arguments here then changing a lot of call sites.

src/codegen/compile.ml
ggreif: Yes, this needs some care in formulation. `ref` is the moral equivalent of `this` in C++. I could have
``` camp
module type Stream = sig
  type t
  val write_byte : E.t -> G.t -> t -> G.t
  ...
```
Looks more idiomatic, and in the concretisation we would always have `type t = G.t`. What do you think?

src/codegen/compile.ml
ggreif: `write_bignum_leb` and `write_bignum_sleb`?

src/codegen/compile.ml
ggreif: It is only used in one place (`write`). It calls `serialize_go` in turn, which is a function, so it cannot communicate via locals, only stack. Hence for a `BumpStream` the pointer in the caller needs to be updated. For a `Rust` stream (where `ref` is a handle) no update is needed and `checkpoint` can be `nop`.

src/codegen/compile.ml
ggreif: I believe you reviewed my "hack" back then :-)

I don't see how extracting the local index from the getter is a bad thing. A Van Laarhoven lens-like approach would be ideal, but will OCaml's type system carry the weight? I still want `get_data_buf : G.t` in all contexts!

src/codegen/compile.ml
nomeata: Hmm, that might be hiding too much implementation - the caller still needs to be aware that the `this` is a piece of wasm code that only makes sense within a certain function, right?

src/codegen/compile.ml
nomeata: I'd redactor this to `get data_buf : G.t` where `data_buf` is of a data type that's either a local or global. But not in this PR

nomeata: And I believe I tried that before, so this may hit some snag somewhere else

src/codegen/compile.ml
nomeata: Ah, in other words, this is to take the steam, put it's (local) data on the stack and pass to a functions or return from one.

If the pointer would be in memory we wouldn't need this, right?

nomeata: Ok, I understand now.

I wonder if an API consisting of `get_local_state` and `set_local_state` would be easier to understand. Although it's kinda silly if the first one always just returns the token. Ah: but if that token was or were abstract, it _would_ make sense.

src/codegen/compile.ml
crusso: Yeah, I would actually use a local functor for this, parameterizing Stream by env and get_data_buf, and instantiating once - at least that way there is less pausibility of mismatching the entries. You could use open to dump the bindings but I think projecting from the module would actually be clearer...

src/codegen/compile.ml
ggreif: Done: 5c4819fec

src/codegen/compile.ml
```diff
@@ -4529,39 +4603,18 @@ module Serialization = struct
     let name = "@serialize_go<" ^ typ_hash t ^ ">" in
```
crusso, line 4603: I think the name of this serialization function will need to depend on the Strm somehow, otherwise you will share code from different Streams, which will be disastrous.

ggreif: See the caveat in the description :-)

src/codegen/compile.ml
crusso: I kinda agree with Joachim. Also, get_data_buf below is not a local but a function parameter. I know wasm function parameters are index as locals, but are they actually updateable? the implementation of share_code makes them so by declaring extra locals, but that's a bit gross.

crusso (commented):
So the approach here will lead us down the path where we generate one set of serialization functions per stream type rather than passing a real object with functions for each write operation, and sharing a single set of serialization function. I guess that'll be as fast as now but I wonder how much code duplication we'll get. I guess we can try this first and back off if code size is an issue.

src/codegen/compile.ml
crusso: Oops.

src/codegen/compile.ml
```diff
@@ -4164,7 +4164,81 @@ module RTS_Exports = struct
...
+  let checkpoint _env get_data_buf = G.setter_for get_data_buf
+
+  let reserve _env get_data_buf bytes =
+    get_data_buf ^^ get_data_buf ^^ compile_add_const bytes ^^ G.setter_for get_data_buf
+
+  let write_word_leb env get_data_buf code =
+    let set_word, get_word = new_local env "word" in
+    code ^^ set_word ^^
+    I32Leb.compile_store_to_data_buf_unsigned env get_word get_data_buf ^^
+        advance_data_buf get_data_buf
```
crusso, line 4198:
```suggestion
    advance_data_buf get_data_buf
```

src/codegen/compile.ml
nomeata: But the databuf value isnt constant, is it? in different functions they are different

src/codegen/compile.ml
nomeata: Yes, they are updateable, we use this in other places as well

nomeata:
How many streams do we expect? Only a fixed number of built in ones: 

 * replies
 * call args
 * Stable memory
 * possibly: blobs

That's not too bad. And probably we can unify some of them: have _one_ buffering stream implementation, which is run-time parametrized by how to write out chunks.

ggreif:
Right, I don't expect too many. The stream I am sketching out in https://github.com/dfinity/motoko/pull/3149/files#diff-a80f0a6d585986554d840e4e2454b2caefd74c9e4273664deef9bf1f195c2e04 can split (when dimensioned big enough to hold all the bytes) and then one has a bona-fide `Blob`. There is a tension between speed and machine-code. I expect that there will be 2-3 max streams. And they can be configurable by the command line, i.e. small dApps don't need anything fancy, since there is enough left over memory in heap.

src/codegen/compile.ml
ggreif: All I want is
``` ocaml
      let    [write_word;     write_word32;  write_byte; write_blob; write_text; write_unsigned;   write_signed] =
        Strm.[write_word_leb; write_word_32; write_byte; write_blob; write_text; write_bignum_leb; write_bignum_sleb]
        |> List.map (fun f -> f env get_data_buf) in
```
But I get a `Warning 8 [partial-match]: this pattern-matching is not exhaustive.` Damn! `map` is structure-preserving...

ggreif: Yep, it isn't worth the complication. 1d32004dc just cuts the Gordian knot.

crusso (approved):
LGTM.

Sorry, I was mistaken about Wasm function parameters - they are mutable. Fire away!

src/codegen/compile.ml
ggreif: We can always change the stream API as we see fit. The next PR will implement a Rust buffered stream and that may give us hints how this could be improved.

src/codegen/compile.ml
ggreif: Okay, I don't see an obvious way forward, so I am being lazy and just document the facts: see 13cc22c.

PR #3162, opened 2022-03-16
compile.ml: Wrap cycle prim impls in `share_code`

nomeata:
in preparation for #3159

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -699,14 +699,16 @@ module FakeMultiVal = struct
 
   let store env tys =
     if !Flags.multi_value || List.length tys <= 1 then G.nop else
-    G.concat_mapi (fun i _ ->
+    G.concat_mapi (fun i ty ->
+      assert(ty = I32Type);
       G.i (GlobalSet (nr (global env i)))
     ) tys
 
   let load env tys =
     if !Flags.multi_value || List.length tys <= 1 then G.nop else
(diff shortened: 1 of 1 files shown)


src/codegen/compile.ml
```diff
@@ -3832,7 +3834,91 @@ module IC = struct
 
 end (* IC *)
 
+module Cycles = struct
```
ggreif, line 3837: This module just moved, right?

src/codegen/compile.ml
```diff
@@ -8302,62 +8348,18 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
-      get_dst ^^
-      Cycles.load_cycles env
-    )
+    SR.Vanilla, Cycles.refunded env
+
   | SetCertifiedData, [e1] ->
-    SR.unit,
-    compile_exp_vanilla env ae e1 ^^
-    IC.set_certified_data env
+    SR.unit, compile_exp_vanilla env ae e1 ^^ IC.set_certified_data env
```
ggreif, line 8362: was this intentional?

nomeata: Yes, a bit more consistent with the lines above. Irrelevant cosmetics, though

ggreif (approved):
LGTM, but this is a rather sensitive area, so another pair of eyeballs would be preferable. @crusso?

crusso (approved):
LGTM

PR #3165, opened 2022-03-17
`perf-delta.nix`: use `--omit-metadata motoko:compiler` to filter out revision fuzz

ggreif:
also
-  include the `moc` version into the log
- and fix the ok/fail messages (was inverted)

We still have a lot of ignored compilations, because we probably have to mangle the inputs like `run.sh` does.


crusso (approved):
rubber stamping

PR #3166, opened 2022-03-17
experiment: failed attempt to extend opt subtyping to candid sequences.

crusso:


rossberg:
Yeah, if I parse the blobs correctly, then these are all cases where the new "untyped" semantics now allows success.

PR #3168, opened 2022-03-18
nix: get releaseVersion from Changelog.md via nix

### LINKED ISSUE #3167: Use version from `Changelog.md` when not available otherwise
Oh good point! Can you open an issue (on the phone right now)? I might give it a shot at some point

_Originally posted by @nomeata in https://github.com/dfinity/motoko/issues/3163#issuecomment-1072480897_

nomeata:
previously, this only happened when built via the github release action,
but that’s round-about, when we can read and parse `Changelog.md`
directly in nix. Now even `nix-build -A moc` produces a file with a
proper version number.

But it would be a lie to just say `moc-0.6.15` when built that way, as
there are likely changes on top of 0.6.15. So append `+` _unless_
this build is done via the github action and thus likely an official
clean release (the github action even compares git tag with changelog
file).

Drive-by-contributions: use `--omit-metadata` in `test/compare-wat.sh`.

fixes #3167


crusso (commented):
Looks like an improvement to me, but I'll leave @ggreif to judge...

nomeata:
I’ll need to update the test output or adjust the canonicalization patterns.

nomeata:
@ggreif ping?

ggreif (approved):
Looks fine to me. But I don't really get the full picture how the version string is constructed when the `git` repo is available, etc.

_EDIT_: the PR description is good enough, now I can join the pieces.

PR #3170, opened 2022-03-23
feat: optimized candid subtype check

crusso:
Builds on #3151 to create a global type table and (eventually) memoize results to subtype calls.


ggreif (approved):
LGTM!

PR #3115, opened 2022-02-18
feat:  implementation of rts candid subtyping check

crusso:
oh woe is me.

- [x] memoization
- [x] function annotations
- [x] opt rules
- [x] records
- [x] variants

basically the hard stuff.

PR #3171, opened 2022-03-24
feat: candid subtype check

crusso:
Builds on (in reverse order)

* #3170 feat: optimized candid subtype check (adding a global type table of types involved in candid subtype checks)
* #3151 feat: candid subtype check, during deserialization (first integration of subtype checks with deserialization)
* #3115 feat: implementation of rts candid subtyping check  (implementaton of (unextended) type table base candid subtype checks 

Since this PR includes more test and fixes to previous PR, its probably best to just review this one as a diff against master:

Over #3170, this PR:

- Introduces a stack allocated memo table allocated on entry to deserialize and shared amongst all recursive calls. The `extended` boolean flag is replaced by a possibly null memo table (`bit_rel_opt`) - the table is `null` for extended deserialization of stable values (for which subtype checks on references are just ommitted since unnecessary).

- Generalises the subtype check and bitrel cache to support sharing across multiple calls and caching of both positive and *negative* subtype test results:
- Adds 1 bit per pair of types in subtype `cache` to record true/false outcome separately from plain addition to cache. 
- Adjust each exit point that returns false to invalidate the positive subtyping assumption already in the cache (thus recording the negative result in the cache).

Unfortunately caching negative results doubles the space required to store the cache (we now need 2 bits, not 1, for every possible pair of type indices).
This subtyping between type tables with 128 entries each will consume 8K = (2 * 2 * 128 * 128/ 8) of stack space. Not great, but not terrible either.

The first, `visited` bit (bit 0 per (t1,t2) pair), is stored in unnegated form. The second, `related` bit (bit 1 per (t1,t2) type pair), is stored in negated form so that assuming the relation holds when first visited is actually a no-op (since the matrix is initialized with zero bits).

Also fixes a lurking bug where I failed to skip the value when the subtype check fails (meaning previous PR were buggy).

- [x] create the cache once on entry to deserialize and share it between calls to idl_sub
- [x] avoid explicitly assuming true by defaulting to full relations on entry (perhaps by inverting true and false bits)
- [ ] ~only produce global type table entries in unextended deserialization.~ Not worth the effort.
- [x] add basic test for recursive types
- [x] overflow check on Stack.dynamic_alloc
- [ ] idl_sub on future types.
- [x] fix broken re-initialization of bitrel on every sub type check
- [x] integrate with changes due to to_candid/from_candid #3155


chenyan-dfinity:
A yes/no cache means the type pair is visited, so I think we only need the cache, and don't need to store the visited set?

crusso:
> A yes/no cache means the type pair is visited, so I think we only need the cache, and don't need to store the visited set?

In a higher-level implementation I would use a mutable, partial map : (Ty*Ty)->bool and use the domain of the map to track visit/not visit, as I think you are suggesting.

The representation I'm using is very naive and uses 1 bit to say whether the pair is in the map or not and another to store the result.

src/codegen/compile.ml
```diff
@@ -8541,7 +8760,7 @@ The compilation of declarations (and patterns!) needs to handle mutual recursion
 This requires conceptually three passes:
  1. First we need to collect all names bound in a block,
     and find locations for then (which extends the environment).
-    The environment is extended monotonously: The type-checker ensures that
+    The environment is extended monotonicly: The type-checker ensures that
```
crusso, line 8763:
```suggestion
    The environment is extended monotonically: The type-checker ensures that
```

rts/motoko-rts/src/idl.rs
```diff
@@ -486,3 +512,335 @@ unsafe extern "C" fn skip_fields(tb: *mut Buf, buf: *mut Buf, typtbl: *mut *mut
...
+            (IDL_CON_service, IDL_CON_service) => {
+                let mut n1 = leb128_decode(&mut tb1);
+                let n2 = leb128_decode(&mut tb2);
+                for _ in 0..n2 {
+                    if n1 == 0 {
+                        break 'return_false;
+                    };
+                    let len2 = leb128_decode(&mut tb2);
+                    let p2 = tb2.ptr;
+                    Buf::advance(&mut tb2, len2);
```
nomeata, line 765: Worth abstracting over this, with a function taking `tb` and returning `(len, ptr)`?

rts/motoko-rts/src/idl.rs
```diff
@@ -486,3 +512,335 @@ unsafe extern "C" fn skip_fields(tb: *mut Buf, buf: *mut Buf, typtbl: *mut *mut
...
+            // default
+            (_, _) => {
+                break 'return_false;
+            }
+        }
+    }
+    // remember negative result ...
+    if t1 >= 0 && t2 >= 0 {
+        rel.disprove(p, t1 as u32, t2 as u32);
+    }
```
nomeata, line 801: Remembering negative results is “just” for performance; it would terminate and be correct without this, right?

src/codegen/compile.ml
```diff
@@ -789,6 +812,9 @@ module RTS = struct
     E.add_func_import env "rts" "memcmp" [I32Type; I32Type; I32Type] [I32Type];
     E.add_func_import env "rts" "version" [] [I32Type];
     E.add_func_import env "rts" "parse_idl_header" [I32Type; I32Type; I32Type; I32Type; I32Type] [];
+    E.add_func_import env "rts" "idl_sub_buf_words" [I32Type; I32Type] [I32Type];
+    E.add_func_import env "rts" "idl_sub"
+      [I32Type; I32Type; I32Type; I32Type; I32Type; I32Type; I32Type; I32Type; I32Type] [I32Type];
```
nomeata, line 817:
```suggestion
      (i32s 9) [I32Type];
```

src/codegen/compile.ml
```diff
@@ -1030,12 +1057,42 @@ module Stack = struct
     G.i (Binary (Wasm.Values.I32 I32Op.Add)) ^^
     set_stack_ptr env
 
+  (* TODO: why not just remember and reset the stack pointer, instead of calling free_words? Also below *)
```
nomeata, line 1060: Don’t remember, maybe no reason at all.

nomeata (commented):
First skim; didn’t look at the `compile.ml` stuff closely yet.

rts/motoko-rts/src/idl.rs
crusso: Yes but a vec might involved many repeated negative tests without failing so I think it's worth optimizing.

rts/motoko-rts/src/idl.rs
```diff
@@ -486,3 +512,335 @@ unsafe extern "C" fn skip_fields(tb: *mut Buf, buf: *mut Buf, typtbl: *mut *mut
...
+                }
+            }
+            (IDL_CON_func, IDL_CON_func) => {
+                // contra in domain
+                let in1 = leb128_decode(&mut tb1);
+                let mut in2 = leb128_decode(&mut tb2);
+                for _ in 0..in1 {
+                    let t11 = sleb128_decode(&mut tb1);
+                    if in2 == 0 {
+                        if !is_null_opt_reserved(typtbl1, end1, t11) {
```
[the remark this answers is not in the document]
crusso: Oh, are you sure null isn't allowed? Then some compile.ml code may need changeing too. @chenyan-dfinity what is your take on this? I don't remember if I cribbed this from the Rust code or Candid  spec change... but don't think I improvised it either....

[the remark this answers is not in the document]
chenyan-dfinity: @nomeata is correct. I think it's related to this: https://github.com/dfinity/candid/issues/319

rts/motoko-rts/src/idl.rs
nomeata: Yes, I agree. Worth leaving comments many to highlight the different role of caching positive and negative results here, for our future selfs.

src/codegen/compile.ml
nomeata: If we do this, worth checking that the stackpointer is as expected afterwards? At least in debug mode?

src/codegen/compile.ml
```diff
@@ -4395,6 +4452,23 @@ module MakeSerialization (Strm : Stream) = struct
       by the next GC.
   *)
 
+  let register_delayed_globals env =
+    (E.add_global32_delayed env "__typtbl" Immutable,
+     E.add_global32_delayed env "__typtbl_end" Immutable,
+     E.add_global32_delayed env "__typtbl_size" Immutable,
+     E.add_global32_delayed env "__typtbl_idltyps" Immutable)
```
nomeata, line 4459: Hmm, it would be good to have a `Note` somewhere, referenced from here and other places, that explains the big picture: What is the meaning of these globals, how are they constructed, etc.

src/codegen/compile.ml
```diff
@@ -4579,10 +4661,52 @@ module MakeSerialization (Strm : Stream) = struct
...
+       in
+       let tbl =
+         let buf = Buffer.create (List.length offsets * 4) in
+         begin
+           List.iter (fun offset ->
+             Buffer.add_int32_le buf
+               Int32.(add static_typedesc (of_int(offset))))
+           offsets;
+           Buffer.contents buf
+         end
```
nomeata, line 4689: Can we use a helper function here? Or can we use the `StaticBytes` DSL? I think this would work:

```
let tbl = StaticBytes.(as_Bytes (List.map (fun offset -> I32 (Int32.(add static_typedesc (of_int(offset)))))))
```

similar elsewhere where buffers are assembled.

src/codegen/compile.ml
```diff
@@ -4579,10 +4661,52 @@ module MakeSerialization (Strm : Stream) = struct
...
+         begin
+           List.iter (fun idltyp ->
+             Buffer.add_int32_le buf idltyp)
+           idltyps;
+           Buffer.contents buf
+         end
+       in
+       let static_idltyps =
+         Int32.add (Blob.vanilla_lit env static_idltyps) Blob.unskewed_payload_offset
+       in
```
nomeata, line 4705: Why are you adding this as `Blob`s? Why not add this static data to static memory directly, without a header  (with `add_static`, which conveniently takes `StaticData.t`, so you don’t have to use `StaticData.as_bytes` above)

src/codegen/compile.ml
```diff
@@ -4903,20 +5027,61 @@ module MakeSerialization (Strm : Stream) = struct
...
+    if extended then
+      f (compile_unboxed_const 0l)
+    else
+      get_typtbl_size1 ^^ get_typtbl_size env ^^
+      E.call_import env "rts" "idl_sub_buf_words" ^^
+      Stack.dynamic_with_words env "rel_buf" f
+
+
+  let idl_sub env t2 =
+    let idx = List.length (!(env.E.typtbl_typs)) in
```
nomeata, line 5041:
```suggestion
    let idx = List.length !(env.E.typtbl_typs) in
```

src/codegen/compile.ml
```diff
@@ -4903,20 +5027,61 @@ module MakeSerialization (Strm : Stream) = struct
...
+    env.E.typtbl_typs := !(env.E.typtbl_typs) @ [t2];
```
nomeata, line 5042: I tend to keep that “return index and append” logic within `module Env`. See functions like `add_global`, and the helper function `reg`.

src/codegen/compile.ml
```diff
@@ -4903,20 +5027,61 @@ module MakeSerialization (Strm : Stream) = struct
...
+    get_typtbl_idltyps env ^^
+    compile_add_const (Int32.of_int (idx * 4)) ^^
+    G.i (Load {ty = I32Type; align = 0; offset = 0l; sz = None}) ^^
```
nomeata, line 5045: One could put the offset into `offset` here.

src/codegen/compile.ml
```diff
@@ -5057,7 +5224,7 @@ module MakeSerialization (Strm : Stream) = struct
 
       let read_actor_data () =
         read_byte_tagged
-          [ E.trap_with env "IDL error: unexpected actor reference"
+          [ E.trap_with env ("IDL error: unexpected actor reference")
```
nomeata, line 5227:
```suggestion
          [ E.trap_with env "IDL error: unexpected actor reference"
```

src/codegen/compile.ml
```diff
@@ -5595,45 +5794,64 @@ module MakeSerialization (Strm : Stream) = struct
...
-          deserialize_go env t ^^ set_val ^^
-          get_val ^^ compile_eq_const (coercion_error_value env) ^^
-          E.then_trap_with env ("IDL error: coercion failure encountered") ^^
-          get_val
+          let can_recover, default_or_trap = Type.(
+            match normalize t with
+            | Prim Null | Opt _ | Any ->
+              (compile_unboxed_one, fun msg -> Opt.null_lit env)
+            | _ ->
+              (compile_unboxed_zero, fun msg -> E.trap_with env msg))
```
nomeata, line 5812: I’d keep the boolean in the ocaml level for longer
```suggestion
              (true, fun msg -> Opt.null_lit env)
            | _ ->
              (false, fun msg -> E.trap_with env msg))
```

src/codegen/compile.ml
```diff
@@ -5595,45 +5794,64 @@ module MakeSerialization (Strm : Stream) = struct
...
+           (default_or_trap ("IDL error: too few arguments " ^ ts_name))
+           (begin
+             get_rel_buf_opt ^^
+             get_data_buf ^^ get_ref_buf ^^
+             get_typtbl_ptr ^^ load_unskewed_ptr ^^
+             get_maintyps_ptr ^^ load_unskewed_ptr ^^ (* typtbl_end *)
+             get_typtbl_size_ptr ^^ load_unskewed_ptr ^^
+             ReadBuf.read_sleb128 env get_main_typs_buf ^^
+             compile_unboxed_const 0l ^^ (* initial depth *)
+             can_recover ^^
```
nomeata, line 5826:
```suggestion
             Bool.lit can_recover ^^
```

src/codegen/compile.ml
```diff
@@ -5595,45 +5794,64 @@ module MakeSerialization (Strm : Stream) = struct
...
-          get_val
+          let can_recover, default_or_trap = Type.(
+            match normalize t with
+            | Prim Null | Opt _ | Any ->
+              (compile_unboxed_one, fun msg -> Opt.null_lit env)
+            | _ ->
+              (compile_unboxed_zero, fun msg -> E.trap_with env msg))
+          in
+          get_arg_count ^^
+          compile_rel_const I32Op.Eq 0l ^^
```
nomeata, line 5815:
```suggestion
          compile_eq_const 0l ^^
```

nomeata (commented):
look plausible so far

rts/motoko-rts/src/idl.rs
crusso: You coming back?

src/codegen/compile.ml
crusso: Ignorance, really.

rts/motoko-rts/src/idl.rs
nomeata: I worringly feel like I have never left

crusso: That'll change...

src/codegen/compile.ml
```diff
@@ -5472,16 +5639,44 @@ module MakeSerialization (Strm : Stream) = struct
...
+            get_typtbl_end ^^
+            get_typtbl_size ^^
+            get_idltyp ^^
+            idl_sub env t
+            end)
+          (compile_unboxed_const 1l) ^^
+        G.if1 I32Type
+          (with_composite_typ idl_func (fun _get_typ_buf ->
+            read_byte_tagged
+              [ E.trap_with env ("IDL error: unexpected function reference")
```
crusso, line 5656:
```suggestion
              [ E.trap_with env "IDL error: unexpected function reference"
```

test/run-drun/idl-sub.mo
```diff
@@ -0,0 +1,183 @@
...
+      await this.send_f15(f15);
+
+   };
+
+
+}
+//SKIP run
+//SKIP run-ir
+//SKIP run-low
+//CALL ingress go "DIDL\x00\x00"
```
crusso, line 183:
```suggestion
//CALL ingress go "DIDL\x00\x00"

```

test/run-drun/idl-sub.mo
```diff
@@ -0,0 +1,183 @@
...
+      await this.send_f7(f7);
+      await this.send_f8(f8);
+      await this.send_f9(f9);
+      await this.send_f10(f10);
+      await this.send_f11(f11);
+      await this.send_f12(f12);
+      await this.send_f13(f13);
+      await this.send_f14(f14);
+      await this.send_f15(f15);
+
```
crusso, line 175:
```suggestion
```

test/run-drun/idl-sub-rec.mo
```diff
@@ -0,0 +1,233 @@
...
+      };
+    };
+
+
+  };
+}
+//SKIP run
+//SKIP run-ir
+//SKIP run-low
+//CALL ingress go "DIDL\x00\x00"
```
crusso, line 233:
```suggestion
//CALL ingress go "DIDL\x00\x00"

```

test/run-drun/idl-sub-ho.mo
```diff
@@ -0,0 +1,241 @@
...
+        catch e { Prim.debugPrint "wrong_6_0"; }
+      };
+
+   };
+
+}
+//SKIP run
+//SKIP run-ir
+//SKIP run-low
+//CALL ingress go "DIDL\x00\x00"
```
crusso, line 241:
```suggestion
//CALL ingress go "DIDL\x00\x00"

```

crusso:
Hmmn, since I stack allocate a chunk and use Rust recursively, would it be wise to reserve a second page for the C/Rust stack? Or is that really just the C stack, not used by Rust at all or only sparingly? @nomeata

nomeata:
Why are you worried about using the same stack? Because it could be too small?

crusso:
> Why are you worried about using the same stack? Because it could be too small?

Yep. For two 128 type type tables, I would allocate 8K of stack. That's a reasonable chunk of the 64K we allow

nomeata:
> Yep. For two 128 type type tables, I would allocate 8K of stack. That's a reasonable chunk of the 64K we allow

You could allocate the type tables on the heap instead

src/codegen/compile.ml
nomeata: Oh, sorry. How about
```suggestion
      (Lib.List.make 9 I32Type);
```

crusso:
> > Yep. For two 128 type type tables, I would allocate 8K of stack. That's a reasonable chunk of the 64K we allow
> 
> You could allocate the type tables on the heap instead

Seems a shame to, since I know the lifetime (and I know our GC).

src/codegen/compile.ml
crusso: I guess there is too much code in the world already...

nomeata:
Yeah, but it's what we do elsewhere, e.g. when allocating a buffer for serialization. So it'd be consistent, and can be improved together with the other instances when we have an allocator. Or leave it as it is until someone complains. Or bump the stack size?

rts/motoko-rts/src/idl.rs
```diff
@@ -486,3 +512,335 @@ unsafe extern "C" fn skip_fields(tb: *mut Buf, buf: *mut Buf, typtbl: *mut *mut
...
+                let mut t11 = 0;
+                let mut advance = true;
+                for _ in 0..n2 {
+                    let tag2 = leb128_decode(&mut tb2);
+                    let t21 = sleb128_decode(&mut tb2);
+                    if n1 == 0 {
+                        // check all remaining fields optional
+                        if !is_null_opt_reserved(typtbl2, end2, t21) {
+                            break 'return_false;
+                        }
```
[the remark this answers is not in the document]
crusso: Done

src/codegen/compile.ml
crusso: Done

src/codegen/compile.ml
crusso: Done

src/codegen/compile.ml

src/codegen/compile.ml

src/codegen/compile.ml
crusso: Done

rts/motoko-rts/src/idl.rs
crusso: Done

crusso:
@nomeata PTAL. I think this is ready to go in.

crusso:
@chenyan-dfinity @nomeata @rossberg should we merge this now or wait for the Rust cdk/candid repo to catchup first?

nomeata:
I'd prefer to have all implementations ready, and then merge spec change and impl changes at roughly the same time.

Or at least merge spec change now and give it a proper version before merging this.

crusso:
> I'd prefer to have all implementations ready, and then merge spec change and impl changes at roughly the same time.
> 
> Or at least merge spec change now and give it a proper version before merging this.

Sounds reasonable to me.

src/codegen/compile.ml
```diff
@@ -4579,10 +4677,39 @@ module MakeSerialization (Strm : Stream) = struct
...
+    (Buffer.contents buf,
+     offsets,
+     List.map idx ts)
+
+  (* See Note [Candid subtype checks] *)
+  let set_delayed_globals (env : E.t) (set_typtbl, set_typtbl_end, set_typtbl_size, set_typtbl_idltyps) =
+    let typdesc, offsets, idltyps = type_desc env (E.get_typtbl_typs env) in
+    let static_typedesc =
+      E.add_static_unskewed env [StaticBytes.Bytes typdesc]
+    in
```
nomeata, line 4697:
```suggestion
    let static_typedesc = E.add_static_unskewed env [StaticBytes.Bytes typdesc] in
```

src/codegen/compile.ml
```diff
@@ -4579,10 +4677,39 @@ module MakeSerialization (Strm : Stream) = struct
...
+      let bytes = StaticBytes.i32s
+        (List.map (fun offset ->
+          Int32.(add static_typedesc (of_int(offset))))
+        offsets)
+      in
+      E.add_static_unskewed env [bytes]
+    in
+    let static_idltyps =
+      E.add_static_unskewed env [StaticBytes.i32s idltyps]
+    in
```
nomeata, line 4708:
```suggestion
    let static_idltyps = E.add_static_unskewed env [StaticBytes.i32s idltyps] in
```
(just to prove that I read the code)

src/codegen/compile.ml
```diff
@@ -4890,8 +5017,49 @@ module MakeSerialization (Strm : Stream) = struct
...
 
+  (* See Note [Candid subtype checks] *)
+  let with_rel_buf_opt env extended get_typtbl_size1 f =
+    if extended then
+      f (compile_unboxed_const 0l)
+    else
+      get_typtbl_size1 ^^ get_typtbl_size env ^^
+      E.call_import env "rts" "idl_sub_buf_words" ^^
+      Stack.dynamic_with_words env "rel_buf"
+      (fun get_ptr ->
```
nomeata, line 5028:
```suggestion
      Stack.dynamic_with_words env "rel_buf" (fun get_ptr ->
```

src/codegen/compile.ml
```diff
@@ -5472,16 +5644,46 @@ module MakeSerialization (Strm : Stream) = struct
...
+        get_rel_buf_opt ^^
+        G.if1 I32Type
+          (begin
+            get_rel_buf_opt ^^
+            get_typtbl ^^
+            get_typtbl_end ^^
+            get_typtbl_size ^^
+            get_idltyp ^^
+            idl_sub env t
+            end)
```
nomeata, line 5657: Begin-end _and_ parenthesis is redundant
```suggestion
          begin
            get_rel_buf_opt ^^
            get_typtbl ^^
            get_typtbl_end ^^
            get_typtbl_size ^^
            get_idltyp ^^
            idl_sub env t
          end
```

src/codegen/compile.ml
```diff
@@ -5472,16 +5644,46 @@ module MakeSerialization (Strm : Stream) = struct
...
+          (compile_unboxed_const 1l) ^^
```
nomeata, line 5658:
```suggestion
          (Bool.lit true) ^^ (* if we don't have a subtype memo table, assume the types are ok *)
```

src/codegen/compile.ml
```diff
@@ -5472,16 +5644,46 @@ module MakeSerialization (Strm : Stream) = struct
...
```
nomeata, line 5681:
```suggestion
          begin
            get_rel_buf_opt ^^
            get_typtbl ^^
            get_typtbl_end ^^
            get_typtbl_size ^^
            get_idltyp ^^
            idl_sub env t
          end
          (Bool.lit true) ^^
```

src/codegen/compile.ml
```diff
@@ -5595,45 +5802,64 @@ module MakeSerialization (Strm : Stream) = struct
...
+        compile_while env
+         (get_arg_count ^^ compile_rel_const I32Op.GtU 0l)
+         (begin
+            get_data_buf ^^
+            get_typtbl_ptr ^^ load_unskewed_ptr ^^
+            ReadBuf.read_sleb128 env get_main_typs_buf ^^
+            compile_unboxed_const 0l ^^
+            E.call_import env "rts" "skip_any" ^^
+            get_arg_count ^^ compile_sub_const 1l ^^ set_arg_count
+           end) ^^
```
nomeata, line 5854:
```suggestion
         begin
           get_data_buf ^^
           get_typtbl_ptr ^^ load_unskewed_ptr ^^
           ReadBuf.read_sleb128 env get_main_typs_buf ^^
           compile_unboxed_const 0l ^^
           E.call_import env "rts" "skip_any" ^^
           get_arg_count ^^ compile_sub_const 1l ^^ set_arg_count
         end ^^
```

src/codegen/compile.ml
```diff
@@ -5738,6 +5964,53 @@ To detect and preserve aliasing, these steps are taken:
    after checking the type hash field to make sure we are aliasing at the same
    type.
 
+ *)
+
+(*
+Note [Candid subtype checks]
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Deserializing Candid values now requires a Candid subtype check when
```
nomeata, line 5973:
```suggestion
Deserializing Candid values requires a Candid subtype check when
```

src/codegen/compile.ml
```diff
@@ -5738,6 +5964,53 @@ To detect and preserve aliasing, these steps are taken:
...
+each payload to decode.
+
+The known Motoko types are accumulated in a global list as required
+and then, in a final compilation step, encoded to global type table
+and sequence of type indices. The encoding is stored as static
+data referenced by dedicated wasm globals so that we can generate
+code that reference the globals before their final definitions are
+known.
+
+Deserializing a vanilla (not extended) Candid value stack allocates a
```
nomeata, line 5991:
```suggestion
Deserializing a proper (not extended) Candid value stack allocates a
```
(To avoid confusion with vanilla stack representation)

src/codegen/compile.ml
```diff
@@ -5738,6 +5964,53 @@ To detect and preserve aliasing, these steps are taken:
...
+Rust memo table (see bitrel.rs) memoizing the result of sub and
+super type tests performed during deserialization of a given Candid
+value sequence.  The memo table is initialized once, using `idl_sub_buf_init`,
+then shared between recursive calls to deserialize, by threading the (possibly
+null) wasm address of the word buffer as an optional argument.  The
+word buffer is stack allocated in generated code, not Rust, because
+it's size is dynamic and Rust doesn't seem to support dynamically-sized
+stack allocation.
+
+Currently, we only perform Candid subtype checks when decoding vanilla
```
nomeata, line 6003:
```suggestion
Currently, we only perform Candid subtype checks when decoding proper
```

src/codegen/compile.ml
```diff
@@ -540,6 +547,14 @@ module E = struct
     in
     let gc_fn = if !Flags.force_gc then gc_fn else "schedule_" ^ gc_fn in
     call_import env "rts" (gc_fn ^ "_gc")
+
+  (* See Note [Candid subtype checks] *)
+  let add_typtbl_typ (env : t) ty : Int32.t =
+    reg env.typtbl_typs ty
```
nomeata, line 553: It seems this may be called multiple times for the same `ty`. Should it look if the same type has been registered before?

nomeata (commented):
Thanks for the note! I’m expect someone will be happy about it in the future.

src/codegen/compile.ml
nomeata: It should probably do the same thing as 
```
  let func_type (env : t) ty =
    let rec go i = function
      | [] -> env.func_types := !(env.func_types) @ [ ty ]; Int32.of_int i
      | ty'::tys when ty = ty' -> Int32.of_int i
      | _ :: tys -> go (i+1) tys
       in
    go 0 !(env.func_types)
```
Maybe this can even be extracted into a `lookup_or_reg` helper (like `reg`). Not sure if that works well with the different equalities on ocaml types.

nomeata: Or if you don't like the linear scan, add a second field to the env with a `i32 type_map` (or however that's called) to find existing entries.

crusso: Yeah, I considered this but didn't think it was that important since the size of the type table is not affected and this was just more straigthforward (one entry per static test). But sure, I'll jump.

crusso: It's kind of ugly to do this and I know realize it's actually unecessary. `idl_sub` and `add_typtblt_typ` will be called at most "once" for a given type because of the way we share the serialization code. 

Though "once" at the moment actually means twice because we generate different code for extended and unextended serialization, but I want to fix that in future (#3185)

nomeata: > since the size of the type table is not affected

But the size of the array with the table indices, right?

> it's actually unnecessary

Ok, if it’s shared anyways, then fine. Maybe leave a comment near this function to explain that.

src/codegen/compile.ml
```diff
@@ -540,6 +547,14 @@ module E = struct
     in
     let gc_fn = if !Flags.force_gc then gc_fn else "schedule_" ^ gc_fn in
     call_import env "rts" (gc_fn ^ "_gc")
+
+  (* See Note [Candid subtype checks] *)
+  let add_typtbl_typ (env : t) ty : Int32.t =
```
crusso, line 552:
```suggestion
  (* NB: we don't bother detecting duplicate registrations here because the sharing machinery should ensure
     that `add_typtbl_typ t` is called at most once for any `t`  with a distinct type hash *)
  let add_typtbl_typ (env : t) ty : Int32.t =
```

src/codegen/compile.ml
crusso: Done

nomeata (approved):
I guess with all comments from last time resolved, I can approve now. Let me know if there is something you'd like me to have a closer look at.

src/codegen/compile.ml
```diff
@@ -9918,6 +10194,9 @@ let compile mode rts (prog : Ir.prog) : Wasm_exts.CustomModule.extended_module =
   GC.register_globals env;
   StableMem.register_globals env;
 
+  (* See Note [Candid subtype checks] *)
+  let set_serialization_globals =  Serialization.register_delayed_globals env in
```
crusso, line 10198:
```suggestion
  let set_serialization_globals = Serialization.register_delayed_globals env in
```

test/run-drun/idl-sub-opt-any-record.mo
```diff
@@ -0,0 +1,71 @@
...
+        }
+      };
+
+   };
+
+}
+//SKIP run
+//SKIP run-ir
+//SKIP run-low
+//CALL ingress go "DIDL\x00\x00"
```
crusso, line 71:
```suggestion
//CALL ingress go "DIDL\x00\x00"

```

rts/motoko-rts/src/bitrel.rs
```diff
@@ -0,0 +1,107 @@
...
+        if (self.end as usize) < (self.ptr as usize) {
+            idl_trap_with("BitRel invalid fields");
+        };
+
+        let bytes = ((self.end as usize) - (self.ptr as usize)) as u32;
+        if (2 * self.size1 * self.size2 * BITS) > bytes * 8 {
+            idl_trap_with("BitRel not enough bytes");
+        };
+        memzero(self.ptr as usize, Words(bytes / WORD_SIZE));
+    }
```
luc-blaeser, line 36: Not sure whether some bits may remain uninitialized, if bytes is not a multiple of word size.

rts/motoko-rts/src/bitrel.rs
```diff
@@ -0,0 +1,107 @@
+//! This module implements a simple subtype cache used by the compiler (in generated code)
```
luc-blaeser, line 1: Just a detail, not important: For `make test`, there are some compiler warnings that this is unused. Maybe,`#[allow(dead_code)]` or a simple unit test would resolve this.

rts/motoko-rts/src/bitrel.rs
```diff
@@ -0,0 +1,107 @@
...
+
+        let bytes = ((self.end as usize) - (self.ptr as usize)) as u32;
+        if (2 * self.size1 * self.size2 * BITS) > bytes * 8 {
+            idl_trap_with("BitRel not enough bytes");
+        };
+        memzero(self.ptr as usize, Words(bytes / WORD_SIZE));
+    }
+
+    unsafe fn locate_ptr_bit(
+        self: &Self,
```
luc-blaeser, line 39:
```suggestion
        &self
```

rts/motoko-rts/src/bitrel.rs
```diff
@@ -0,0 +1,107 @@
...
+        let (ptr, bit) = self.locate_ptr_bit(p, i_j, j_i, bit);
+        let mask = 1 << bit;
+        return *ptr & mask == mask;
+    }
+
+    pub(crate) unsafe fn visited(&self, p: bool, i_j: u32, j_i: u32) -> bool {
+        self.get(p, i_j, j_i, 0)
+    }
+
+    pub(crate) unsafe fn visit(&self, p: bool, i_j: u32, j_i: u32) -> () {
```
luc-blaeser, line 90:
```suggestion
    pub(crate) unsafe fn visit(&self, p: bool, i_j: u32, j_i: u32) {
```

rts/motoko-rts/src/bitrel.rs
```diff
@@ -0,0 +1,107 @@
...
+        self.set(p, i_j, j_i, 0, true)
+    }
+
+    // NB: we store related bits in negated form to avoid setting on assumption
+    // This code is a nop in production code.
+    pub(crate) unsafe fn assume(&self, p: bool, i_j: u32, j_i: u32) -> () {
```
luc-blaeser, line 96:
```suggestion
    pub(crate) unsafe fn assume(&self, p: bool, i_j: u32, j_i: u32) {
```

rts/motoko-rts/src/bitrel.rs
```diff
@@ -0,0 +1,107 @@
...
+        debug_assert!(!self.get(p, i_j, j_i, 1));
+    }
+
+    pub(crate) unsafe fn related(&self, p: bool, i_j: u32, j_i: u32) -> bool {
+        !self.get(p, i_j, j_i, 1)
+    }
+
+    pub(crate) unsafe fn disprove(&self, p: bool, i_j: u32, j_i: u32) -> () {
```
luc-blaeser, line 104:
```suggestion
    pub(crate) unsafe fn disprove(&self, p: bool, i_j: u32, j_i: u32) {
```

rts/motoko-rts/src/idl.rs
```diff
@@ -488,3 +516,344 @@ unsafe extern "C" fn skip_fields(tb: *mut Buf, buf: *mut Buf, typtbl: *mut *mut
...
+    if is_primitive_type(t) {
+        return t == IDL_PRIM_reserved;
+    }
+
+    // unfold t
+    let mut t = t;
+
+    let mut tb = Buf {
+        ptr: *typtbl.add(t as usize),
+        end: end,
```
luc-blaeser, line 530:
```suggestion
        end,
```

rts/motoko-rts/src/idl.rs
```diff
@@ -488,3 +516,344 @@ unsafe extern "C" fn skip_fields(tb: *mut Buf, buf: *mut Buf, typtbl: *mut *mut
...
+unsafe extern "C" fn idl_sub_buf_words(typtbl_size1: u32, typtbl_size2: u32) -> u32 {
+    return BitRel::words(typtbl_size1, typtbl_size2);
+}
+
+#[no_mangle]
+unsafe extern "C" fn idl_sub_buf_init(
+    rel_buf: *mut u32,
+    typtbl_size1: u32,
+    typtbl_size2: u32,
+) -> () {
```
luc-blaeser, line 816:
```suggestion
) {
```

rts/motoko-rts/src/bitrel.rs
luc-blaeser: Just added some code style suggestions from `cargo clippy` checker tool.

luc-blaeser (commented):
Looks very good. Huge work! Maybe just the question about the `BitRel.init` could be interesting from my comments...

rts/motoko-rts/src/bitrel.rs
crusso: Good catch - will need to check what this was doing - wrote it in April....

crusso: Note to self: I think this is ok because 
```
(self.end as usize) - (self.ptr as usize) = words(self.size1,self.size2)*WORD_SIZE
```

by the way this is initialized in `compile.ml` (so bytes is a mutlple of WORD_SIZE) but I should add that as an assert or check to make the code clearer.

test/run-drun/idl-sub-ho-neg.mo
```diff
@@ -0,0 +1,422 @@
...
+        }
+      };
+
+   };
+
+}
+//SKIP run
+//SKIP run-ir
+//SKIP run-low
+//CALL ingress go "DIDL\x00\x00"
```
crusso, line 422:
```suggestion
//CALL ingress go "DIDL\x00\x00"

```

rts/motoko-rts/src/idl.rs
crusso: Hmm, I'm not sure I like that one since it confused field names with identifiers. And doesn't match the style in the rest of the file

src/codegen/compile.ml
```diff
@@ -6095,6 +6319,53 @@ To detect and preserve aliasing, these steps are taken:
...
    type.
 
+ *)
+
+(*
+Note [Candid subtype checks]
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Deserializing Candid values requires a Candid subtype check when
+deserializing value of reference types (actors and functions).
```
crusso, line 6329:
```suggestion
deserializing values of reference types (actors and functions).
```

rts/motoko-rts/src/bitrel.rs
crusso: I first suppressed the warning but then decided adding a unit test would be wiser and, indeed, it revealed a bug that would have been very hard to diagnose later! 

See #3529

rts/motoko-rts/src/bitrel.rs
crusso: I changed the code to make it clear that bytes muse be a multiple of WORD_SIZE (by adjusting the previous check)

rts/motoko-rts/src/idl.rs
```diff
@@ -46,10 +46,35 @@ const IDL_CON_alias: i32 = 1;
...
+    let len = min(len1, len2);
+    let cmp = libc::memcmp(
+        p1 as *mut libc::c_void,
+        p2 as *mut libc::c_void,
+        len as usize,
+    );
+    if cmp == -1 {
+        return -1;
+    } else if cmp == 1 {
+        return 1;
```
ggreif, line 68:
```suggestion
    if cmp != 0 {
        return cmp;
```

rts/motoko-rts/src/idl.rs
```diff
@@ -488,3 +516,344 @@ unsafe extern "C" fn skip_fields(tb: *mut Buf, buf: *mut Buf, typtbl: *mut *mut
...
+
+    // NB we use a trivial labelled loop so we can factor out the common failure continuation.
+    // exit either via 'return true' or 'break 'return_false' to memoize the negative result
+    'return_false: loop {
+        match (u1, u2) {
+            (_, IDL_CON_alias) | (IDL_CON_alias, _) => idl_trap_with("sub: unexpected alias"),
+            (_, IDL_PRIM_reserved) => return true,
+            (IDL_PRIM_empty, _) => return true,
+            (IDL_PRIM_nat, IDL_PRIM_int) => return true,
+            (_, IDL_CON_opt) => return true, // apparently, this is admissable
```
ggreif, line 607: these all return `true`, why not have an or-pattern?

rts/motoko-rts/src/idl.rs
```diff
@@ -488,3 +516,344 @@ unsafe extern "C" fn skip_fields(tb: *mut Buf, buf: *mut Buf, typtbl: *mut *mut
...
+                        Buf::advance(&mut tb1, len1);
+                        t11 = sleb128_decode(&mut tb1);
+                        n1 -= 1;
+                        cmp = utf8_cmp(len1, p1, len2, p2);
+                        if cmp < 0 && n1 > 0 {
+                            continue;
+                        };
+                        break;
+                    }
+                    if !(cmp == 0) {
```
ggreif, line 783:
```suggestion
                    if cmp != 0 {
```

src/codegen/compile.ml
```diff
@@ -573,6 +581,16 @@ module E = struct
     in
     let gc_fn = if !Flags.force_gc then gc_fn else "schedule_" ^ gc_fn in
     call_import env "rts" (gc_fn ^ "_gc")
+
+  (* See Note [Candid subtype checks] *)
+  (* NB: we don't bother detecting duplicate registrations here because the code sharing machinery
+     ensures that `add_typtbl_typ t` is called at most once for any `t`  with a distinct type hash *)
```
ggreif, line 587:
```suggestion
     ensures that `add_typtbl_typ t` is called at most once for any `t` with a distinct type hash *)
```

nix/sources.json
```diff
@@ -1,15 +1,15 @@
 {
     "candid": {
-        "branch": "master",
+        "branch": "claudio/candid-subtype-check",
```
ggreif, line 3: is `candid-subtype-check` already merged to `master`?

ggreif (approved):
Looks good, but it is a lot.

nix/sources.json
crusso: Nope, it's still waiting in PR https://github.com/dfinity/candid/pull/323

rts/motoko-rts/src/idl.rs
crusso: Will do

nix/sources.json
crusso: Done

Changelog.md
```diff
@@ -2,6 +2,19 @@
 
 * motoko (`moc`)
 
+  * BREAKING CHANGE
+
+    Motoko Candid de-serialization now checks that the actual type of a deserialized actor or function reference is a subtype of the expected type (#3171),
```
chenyan-dfinity, line 7: It was checking subtype before, now we *only* check reference types. But this whole thing probably doesn't make sense for external developers. Maybe just say we implement Candid spec 1.4?

crusso: Actually, no. Motoko was never checking subtyping at all, even at option types. We skipped that version of Candid.

crusso: But yeah, maybe this is too much information.

chenyan-dfinity: Ah okay, then this changelog makes sense, but probably too detailed, even I didn't know about this :)

chenyan-dfinity (approved):

crusso:
@chenyan-dfinity PTAL at the changelog.

chenyan-dfinity (approved):
LGTM

PR #3173, opened 2022-03-29
Cons less when decoding bignums

### LINKED ISSUE #3098: bug: deserialization of Candid `nat` always allocates, even when the `nat` is small
(The examples below are written so they can run as drun tests without trashing the file system - allocation is transient with during the upgrade, to prevent large replica state hitting disk)

This runs correctly, ending in a "deliberate trap"

deser-many-bool.mo
```motoko
import P "mo:⛔";

actor {
  type T = Bool;
  let value = true;
  stable var a : [T] = [];
  let GB = 1024*1024*1024;
  let bytes = 4;

  system func preupgrade() {
    P.debugPrint("(pre");

    a := P.Array_tabulate<T>(GB/4/bytes, func _ { value } ); // 0.25 GB array

    P.debugPrint("pre)");
   };

  system func postupgrade() {
    P.debugPrint("(post)");
    P.trap("deliberate trap");
  }
}

//SKIP run
//SKIP run-low
//SKIP run-ir
// too slow on ic-ref-run:
//SKIP comp-ref
// too resource heavy on GH:
////SKIP comp

//CALL upgrade ""
```

While the very similar, `Nat`-based version crashes with OOM in postupgrade (because deserializing all those small nats allocs like crazy, but needn't)

deser-many-nat.mo:

```motoko
import P "mo:⛔";

actor {
  type T = Nat;
  let value = 0;
  stable var a : [T] = [];
  let GB = 1024*1024*1024;
  let bytes = 4;

  system func preupgrade() {
    P.debugPrint("(pre");

    a := P.Array_tabulate<T>(GB/4/bytes, func _ { value } ); // 0.25 GB array

    P.debugPrint("pre)");
   };

  system func postupgrade() {
    P.debugPrint("(post)");
    P.trap("deliberate trap");
  }
}

//SKIP run
//SKIP run-low
//SKIP run-ir
// too slow on ic-ref-run:
//SKIP comp-ref
// too resource heavy on GH:
////SKIP comp

//CALL upgrade ""
```

ggreif:
Fixes #3098.

## Analysis and solution

The problem was that decoding bignums always happened via TomMathLib, generating a ton of intermediate calculations. This sometimes filled up the heap. Besides being costly in terms of cycles, it could lead to heap overflow. Now we have optimal behaviour, in that we never generate garbage if the bignum ends up as a compact one.

The final implemented method is the 🚀 one :-). We use a loop and sliding masks instead of sequential code and constant masks.

## Plans of attack

Below we sketch a few ways how the buffer's first few bytes worth of prefix can be inspected and decoded, given a (S)LEB EOM bit can be seen.

### Pedestrian 🚶🏻‍♂️

- pull a byte, speculate for EOM, prepare result if so
- pull second byte, speculating, combine two bytes to result if EOM
- etc. if no EOM seen, fall back

I have that for 2 bytes in f5af7ef and it works, but seems like a maintenance nightmare and it bloats Wasm.

### Car 🚘

- have a loop inline with max 5 byte pulls from the buffer
- ⚠️ caveat: needs to move buffer pointer back, and invoke another routine if 5 bytes are not enough

### Rocket 🚀

- pull 8 bytes from buffer if possible
- mask for EOMs and surplus bits (SLEB needs a repeated-sign test), if not found, fall back
- move 7 bit groups to result

## Stats

I have manually disabled the speculation part by
```  diff
diff --git a/src/codegen/compile.ml b/src/codegen/compile.ml
index 42a266485..93a9c5920 100644
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -1980,10 +1980,7 @@ module ReadBuf = struct
     check_page_end env get_buf (compile_add_const 8l) ^^
     G.if1 I64Type
       (compile_const_64 (-1L))
-      begin
-        get_ptr get_buf ^^
-        G.i (Load {ty = I64Type; align = 0; offset = 0l; sz = None})
-      end
+      (compile_const_64 (-1L))
 
   let read_word64 env get_buf =
     check_space env get_buf (compile_unboxed_const 8l) ^^
```
and compared the memory use coming from the array (de-)serialisation (see c66bc8e73). It came out as
8_936 (vs. 71_836 previously). I declare this as success.


src/codegen/compile.ml
```diff
@@ -1734,7 +1738,7 @@ module TaggedSmallWord = struct
   let ctz_kernel ty =
     compile_word_padding ty ^^
     compile_rotr_const (shift_of_type ty) ^^
-    G.i (Unary (Wasm.Values.I32 I32Op.Ctz)) ^^
+    G.i (Unary (Wasm.Values.I64 I32Op.Ctz)) ^^
```
ggreif, line 1741:
```suggestion
    G.i (Unary (Wasm.Values.I32 I32Op.Ctz)) ^^
```

ggreif:
What the fresh hell!!!
```
test-qc>     intra-actor:                    FAIL (8.63s)
[6725](https://github.com/dfinity/motoko/runs/5759216498?check_suite_focus=true#step:7:6725)
test-qc>       *** Failed! Assertion failed (after 29 tests and 2 shrinks):
[6726](https://github.com/dfinity/motoko/runs/5759216498?check_suite_focus=true#step:7:6726)
test-qc>       Matching (Pair (Ctz (Lit 248)) (Lit (-2)),(NatN 3,IntN (-2)))
[6727](https://github.com/dfinity/motoko/runs/5759216498?check_suite_focus=true#step:7:6727)
test-qc>       STDERR:
[6728](https://github.com/dfinity/motoko/runs/5759216498?check_suite_focus=true#step:7:6728)
test-qc>       Error: failed to run main module `matchLocally.wasm`
[6729](https://github.com/dfinity/motoko/runs/5759216498?check_suite_focus=true#step:7:6729)
test-qc>       Caused by:
[6730](https://github.com/dfinity/motoko/runs/5759216498?check_suite_focus=true#step:7:6730)
test-qc>           0: WebAssembly failed to compile
[6731](https://github.com/dfinity/motoko/runs/5759216498?check_suite_focus=true#step:7:6731)
test-qc>           1: WebAssembly translation error
[6732](https://github.com/dfinity/motoko/runs/5759216498?check_suite_focus=true#step:7:6732)
test-qc>           2: Invalid input WebAssembly code at offset 1133: type mismatch: expected i64, found i32
[6733](https://github.com/dfinity/motoko/runs/5759216498?check_suite_focus=true#step:7:6733)
test-qc>       Use --quickcheck-replay=361254 to reproduce.
[6734](https://github.com/dfinity/motoko/runs/5759216498?check_suite_focus=true#step:7:6734)
test-qc>       Use -p '/intra-actor/' to rerun this test only.
[6735](https://github.com/dfinity/motoko/runs/5759216498?check_suite_focus=true#step:7:6735)
test-qc>   Encoding
[6736](https://github.com/dfinity/motoko/runs/5759216498?check_suite_focus=true#step:7:6736)
```

src/codegen/compile.ml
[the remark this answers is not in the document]
ggreif: Happened in c158921c654a1e97687fe97a7141d8086ca360f1, no idea how...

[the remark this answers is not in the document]
ggreif: Reverted in d56fcd71e4868c88d62de28916d92996648ef73f.

rts/motoko-rts/src/bigint.rs
```diff
@@ -523,3 +564,48 @@ pub unsafe extern "C" fn bigint_sleb128_decode(buf: *mut Buf) -> Value {
...
+#[no_mangle]
+pub unsafe extern "C" fn bigint_sleb128_decode_word64(
+    mut leb: i64,
+    bits: u64,
+    buf: *mut Buf,
+) -> Value {
+    let mut acc = leb & 0b111_1111;
+    if bits < 8 {
+        buf.advance(1);
+        return Value::from_signed_scalar((acc as i32) << 25 >> 25);
```
ggreif, line 578: These are ugly. one probably could just copy into the upper 7 bits all the time and then just shift once.

ggreif: 9f06cca12 builds up the result from the top-bits perspective. It has a simpler sign-extension regime, but still has the same number of shifts.

ggreif:
Looking at the wasm, I suspect having loops would reduce the byte code size considerably. Here are the size stats as of 9f06cca1252b370b748e6388514be48df98ee4c4:
```
$ llvm-nm rts/motoko-rts/target/wasm32-unknown-emscripten/release/libmotoko_rts.a | grep decode_word64 | grep ' T '
00000fbe T bigint_leb128_decode_word64
00001246 T bigint_sleb128_decode_word64
```

rts/motoko-rts/src/bigint.rs
```diff
@@ -494,6 +494,40 @@ pub unsafe extern "C" fn bigint_leb128_decode(buf: *mut Buf) -> Value {
...
+                break;
+            }
+            return Value::from_signed_scalar(acc as i32);
+        }
+        bits -= 8;
+        mask <<= 7;
+        leb >>= 1;
+    }
+
+    let tentative = (acc as i32) << 1 >> 1;
```
ggreif, line 523: Having code her like
``` rust
    let tentative = acc as i32;
    if tentative as u64 == acc && (tentative as u32) < (1 << 30) {
        // roundtrip is valid
        return Value::from_signed_scalar(tentative);
    }
```
leads to brachia code and bigger Wasm.

src/codegen/compile.ml
```diff
@@ -5642,6 +5643,35 @@ module MakeSerialization (Strm : Stream) = struct
...
+(*
+Note [speculating for short (S)LEB encoded bignums]
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+#3098 highlighted that a lot of heap garbage can be generated while reading in
+(S)LEB-encoded bignums. To make heap consumption optimal for every compactly
+representable `Int`, we resort to speculatively reading a 64-byte chunk from
+the `ReadBuf`. We call it speculative, because it may read past the end of the
+buffer (and thus end up containing junk bytes) or even fail because reading
+across Wasm page boundaries could cause segfaulting. (Consider the buffer ending
```
nomeata, line 5655: I don't think wasm can segfault. Do you mean trap?

Also I didn't think reading across page boundaries is a problem. Do you mean reading across the end of the memory?

ggreif: It can  trap when we read past the last _allocated_ page's boundary. Since we don't want to get the number of allocated pages here, we just exclude the last few bytes of _each_  page from consideration.

crusso: But would that trap be bad? Shouldn't we trap eventually anyway if the leb encoding runs off end off memory?

ggreif: Fixed the wording. Also the `(Consider the buffer ending  ...` remark does describe the problem at hand.

src/codegen/compile.ml
```diff
@@ -5642,6 +5643,35 @@ module MakeSerialization (Strm : Stream) = struct
...
+3 bytes before the page boundary and issuing a speculative 64-bit read for the
```
ggreif, line 5656:
```suggestion
3 bytes before the last-memory-page boundary and issuing a speculative 64-bit read for the
```

src/codegen/compile.ml
ggreif: @crusso The leb-encoded buffer could be just 1-2 bytes, and we want to read 8. The read could clearly run into nowhere-land, but the buffer still ends cleanly in an allocated page.

src/codegen/compile.ml
nomeata: Would it be worth to allocate, in general in the RTS always 8 bytes extra, so that reading a 64bit value can never fail? Would simplify this code (and maybe others), at a very small cost.

ggreif: Your idea (IIRC) is being explored in another PR, where it would come very handy too:  #3090. But I prefer to keep this for now and do the trap suppression there.

src/codegen/compile.ml
crusso: Ah OK.

But actually, why did you discard the car solutiin above? Just gave two loops, the first reading up to five bytes one by one into a word64, returning if needed, followed by a second loop reading the remaining bytes into a bignum initialized  from the word64?

ggreif: I never implemented 🚘. Going via Rust would give an inferior cycle count and inlining the loop would bloat the Wasm each time  `Int/Nat` appears in a message (or stable `var`).

ggreif: @crusso exploring this speculative method will come in handy when speeding up `bigint_(s)leb128_decode` which does a `read_byte` in a loop now.

rts/motoko-rts/src/bigint.rs
```diff
@@ -523,3 +561,43 @@ pub unsafe extern "C" fn bigint_sleb128_decode(buf: *mut Buf) -> Value {
...
+            }
+            let sext = 25 - 7 * continuations; // this many top bits will get a copy of the sign
+            return Value::from_signed_scalar((acc as i32) << sext >> sext);
+        }
+        bits -= 8;
+        mask <<= 7;
+        sleb >>= 1;
+    }
+
+    let signed = (acc as i64) << 29 >> 29; // sign extends
```
crusso, line 595:
```suggestion
    let signed = (acc as i64) << 29 >> 29; // sign extend
```

crusso:
I'm still uncomfortable with the complexity of this code and wonder if the extra perf is worth it compared to the simpler solution suggested above. I'll approved but do worry this might bite us later.

I wonder what @nomeata thinks?

crusso (approved):
But see my comments

src/codegen/compile.ml
crusso: > I never implemented 🚘. Going via Rust would give an inferior cycle count and inlining the loop would bloat the Wasm each time `Int/Nat` appears in a message (or stable `var`).

Don't we always ultimately go via rust anyway? I think isolating the changes to the original Rust routines, adding a < 5 iteration loop, would have been simpler and easier to grok, even if slightly less efficient with cycles.

ggreif:
> complexity of this code

Some complexity will go away when reading buffer past page boundaries can't trap any more (hopefully with #3090).

crusso:
https://github.com/dfinity/motoko/pull/3178 adds some more rudimentary tests if you want them.

PR #3182, opened 2022-04-06
Preparations and  refactoring for #3149

ggreif:
This is mostly mechanical. The intention is to take out some of the patch pressure (and mental load) from #3149.

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -5551,7 +5551,7 @@ module MakeSerialization (Strm : Stream) = struct
       E.else_trap_with env "data buffer not filled" ^^
 
       get_refs_size ^^
-      G.i (Test (Wasm.Values.I32 I32Op.Eqz)) ^^
+      compile_unboxed_const 0l ^^
       E.else_trap_with env "cannot send references on IC System API" ^^
 
       get_data_start ^^
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -4630,7 +4630,7 @@ module MakeSerialization (Strm : Stream) = struct
         G.i (Binary (Wasm.Values.I32 I32Op.Or)) ^^
         get_tag ^^ compile_eq_const Tagged.(int_of_tag Array) ^^
         G.i (Binary (Wasm.Values.I32 I32Op.Or)) ^^
-        E.else_trap_with env "object_size/Mut: Unexpected tag " ^^
+        E.else_trap_with env "object_size/Mut: Unexpected tag" ^^
         (* Check if we have seen this before *)
         get_tag ^^ compile_eq_const Tagged.(int_of_tag StableSeen) ^^
         G.if0 begin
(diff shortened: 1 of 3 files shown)


src/codegen/compile.ml
```diff
@@ -5756,6 +5755,10 @@ module Serialization = MakeSerialization(BumpStream)
 
 module Stabilization = struct
 
+  let extend64 code = code ^^ G.i (Convert (Wasm.Values.I64 I64Op.ExtendUI32))
```
nomeata, line 5758: My style so far was to pass `code` only when the argument can’t simply be the top element of the stack. Higher order stuff is alwasy harder to understand when called. So I’d advise against it.

You maybe could add `compile_extend64` to our list of `compile_…` functions. (And maybe they need a shorter naming convention).

nomeata (approved):
Minor matter-of-opinion comment, but approving to unblock you

src/codegen/compile.ml
```diff
@@ -5544,15 +5544,14 @@ module MakeSerialization (Strm : Stream) = struct
...
 
       get_data_start ^^ get_data_size ^^ G.i (Binary (Wasm.Values.I32 I32Op.Add)) ^^
       G.i (Compare (Wasm.Values.I32 I32Op.Eq)) ^^
-      E.else_trap_with env "data buffer not filled " ^^
+      E.else_trap_with env "data buffer not filled" ^^
 
       get_refs_size ^^
-      compile_unboxed_const 0l ^^
-      G.i (Compare (Wasm.Values.I32 I32Op.Eq)) ^^
+      G.i (Test (Wasm.Values.I32 I32Op.Eqz)) ^^
```
nomeata, line 5554: Using
```
compile_eq_const 0l
```
and optimizing there or in the peep-hole-optimizier might be a bit more declarative and less line-noisy.

ggreif: Yes, #3149 already does the emissions part. I will prepare a peephole patch when I have time.

ggreif, line 5554:
```suggestion
      compile_unboxed_const 0l ^^
```

src/codegen/compile.ml
ggreif: My hope is that we'll be able to get rid of these conversions when gradually more variables/data trust to 64 bits. This `extend64` is well-confined here and   will be hopefully eliminated soon in one way or the other.

src/codegen/compile.ml
```diff
@@ -5551,7 +5551,7 @@ module MakeSerialization (Strm : Stream) = struct
       E.else_trap_with env "data buffer not filled" ^^
 
       get_refs_size ^^
-      G.i (Test (Wasm.Values.I32 I32Op.Eqz)) ^^
+      compile_unboxed_const 0l ^^
```
nomeata, line 5554: `compile_eq_const`, not `compile_unboxed_const`

src/codegen/compile.ml
```diff
@@ -5544,15 +5544,14 @@ module MakeSerialization (Strm : Stream) = struct
...
-      E.else_trap_with env "reference buffer not filled " ^^
+      E.else_trap_with env "reference buffer not filled" ^^
 
       get_data_start ^^ get_data_size ^^ G.i (Binary (Wasm.Values.I32 I32Op.Add)) ^^
       G.i (Compare (Wasm.Values.I32 I32Op.Eq)) ^^
-      E.else_trap_with env "data buffer not filled " ^^
+      E.else_trap_with env "data buffer not filled" ^^
 
       get_refs_size ^^
       compile_unboxed_const 0l ^^
```
ggreif, line 5554:
```suggestion
      compile_eq_const 0l ^^
```

src/codegen/compile.ml
ggreif: Yeah, silly me. Copied from the wrong place. Corrected.

src/codegen/compile.ml
ggreif: I left https://github.com/dfinity/motoko/issues/3183 for the peephole.

PR #3187, opened 2022-04-09
experiments to get extended candid-tests passing

crusso:
Mostly good except (only 2 tests fail). Lessons:

- Motoko can't handle empty candid method names
- Future types break moc (in idl subtyping?)
- Moc pretty printing of types of higher order async function types may produce unparseable types, breaking suite.
- mo_idl lib needs a hack to avoid printing type stamps in idl function values and references.

Do not merge as type pretty printing needs (better) fixing (best in separate PR I guess)

 Failind candid tests:
 ```
construct:209 skipping minimal future type ... not ok (unwanted trap)!
IDL error: deserializing value of type None
Error: failed to run main module `tmp.wasm`

Caused by:
    0: failed to invoke command default
    1: wasm trap: unreachable
       wasm backtrace:
construct:210 skipping future type with data ... not ok (unwanted trap)!
IDL error: deserializing value of type None
Error: failed to run main module `tmp.wasm`

Caused by:
    0: failed to invoke command default
    1: wasm trap: unreachable
       wasm backtrace:
```

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -5687,7 +5687,8 @@ module MakeSerialization (Strm : Stream) = struct
           Heap.store_field MutBox.field
         )
       | Non ->
-        E.trap_with env "IDL error: deserializing value of type None"
+        skip get_idltyp ^^
+        coercion_failed "IDL error: deserializing value of type None"
       | _ -> todo_trap env "deserialize" (Arrange_ir.typ t)
       end ^^
       (* Parsed value on the stack, return that, unless the failure flag is set *)
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -6,10 +6,10 @@
         "homepage": "",
         "owner": "dfinity",
         "repo": "candid",
-        "rev": "734b796a7c2ffc8f4bb317c99ee3281d54bb36d5",
-        "sha256": "0nx0bw0r3l9zpvczrh0w97sng2900hh987yx7gakzahff4albl5v",
+        "rev": "0889e288c8c44ce36682ffa67df010c83e912a33",
+        "sha256": "122v440prlcr0z6zj1a7j0rjpk1a2xwpvj8vxrdbldl6ss1krbff",
         "type": "tarball",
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by nomeata** [comment]
Maybe I should add the future type tests separately, if they uncover separate bugs or existing omissions in Motoko

**Comment by crusso** [comment]
> Maybe I should add the future type tests separately, if they uncover separate bugs or existing omissions in Motoko

@nomeata They all pass locally now, after fixing deserialization at null to skip n fail not tral

But I still need to implement subtyping of future types, which is probably untested by your suite and perhaps not testable in motoko.

**Comment by crusso** [comment]
Not really ready for review. Finger slipped

File: src/codegen/compile.ml [hunk 1]
```diff
@@ -5687,7 +5687,8 @@ module MakeSerialization (Strm : Stream) = struct
           Heap.store_field MutBox.field
         )
       | Non ->
-        E.trap_with env "IDL error: deserializing value of type None"
+        skip get_idltyp ^^
+        coercion_failed "IDL error: deserializing value of type None"
```
nomeata, line 5691: This looks wrong; Non is a different type than future types. Maybe the type table parser returns Non for future types and should return Future

crusso: So the reason I added this is that your new future type tests
```
// Future types
// This uses 0x67 for the “future type”; bump (well, decrement) once that
// becomes a concrete future type
assert blob "DIDL\01\67\00\02\00\7e\00\00\01" == "(null, true)" : (opt empty,bool) "skipping minimal future type";
assert blob "DIDL\01\67\03ABC\02\00\7e\05\00hello\01" == "(null,true)" : (opt empty,bool) "skipping future type with data";
```
need to be able to deserialise at `opt empty` (? None) and fail when deserialization of the None value fails. Are you sure that's wrong then - seems right to me.

nomeata: Ah, yes, you are right: This not actually about our handling of future types, but just a bug in our subtyping of `Non`.

PR #3196, opened 2022-04-19
bug: prim inliner produces unexpected indirect calls when call_as_prim is false.

crusso:
Add a bug repro for #3159
```
//MOC-ENV MOC_UNLOCK_PRIM=yesplease
import Prim "mo:⛔";

actor {
  func inline(t1: Text, t2 : Text) : Text = t1 # t2;
  func makecalls() : () {
    let t1 = inline("a","b"); // should be inlined
    var p = ("c","d");
    let t2 = inline p; // should not be inlined, but a still direct call
  };
  public func go() : async () {
    makecalls();
  }
}

//SKIP run
//SKIP run-ir
//SKIP run-low
//SKIP comp-ref
//CALL ingress go "DIDL\x00\x00"
```
generates 
```
(func $makecalls (type 4) (param $clos i32)
    (local $t1 i32) (local $p i32) (local $t2 i32) (local $clos.1 i32)
    i32.const 66431
    i32.const 66419
    call $text_concat
    local.set $t1
    i32.const 66403
    local.set $p
    i32.const 66367
    local.tee $clos.1
    local.get $p
    call $from_2_tuple
    global.get 2
    global.get 3
    local.get $clos.1
    i32.load offset=5
    call_indirect (type 11)
    local.set $t2)
```

but the last `call_indirect` could be a `call`, I think.

Added https://github.com/dfinity/motoko/pull/3196/commits/81e215e737818f247daaca53a266b73351ae8cf4 as a fix, after which we produce:
```
  (func $makecalls (type 4) (param $clos i32)
    (local $t1 i32) (local $p i32) (local $t2 i32)
    i32.const 66419
    i32.const 66407
    call $text_concat
    local.set $t1
    i32.const 66391
    local.set $p
    i32.const 0
    local.get $p
    call $from_2_tuple
    global.get 2
    global.get 3
    call $inline
    local.set $t2)
```

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -7728,7 +7728,7 @@ and compile_prim_invocation (env : E.t) ae p es at =
            (* ugly case; let's just call this as a function for now *)
            raise (Invalid_argument "call_as_prim was true?")
          end
-      | SR.Const (_, Const.Fun (mk_fi, Const.Complicated)), _ ->
+      | SR.Const (_, Const.Fun (mk_fi, _)), _ ->
          assert (sort = Type.Local);
          StackRep.of_arity return_arity,
 
--- a/test/run-drun/inline.mo
+++ b/test/run-drun/inline.mo
@@ -2,11 +2,14 @@
 import Prim "mo:⛔";
 
 actor {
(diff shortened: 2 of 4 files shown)


nomeata (approved):
Good catch

PR #3198, opened 2022-04-20
`Stream`: start field

ggreif:


src/codegen/compile.ml
```diff
@@ -5994,18 +5994,16 @@ module Stabilization = struct
 
     let finalize_buffer _ = G.nop (* everything is outputted already *)
 
-    (* Returns a 32-bit number that is reasonably close to the number of bytes
-       that would bave been written to stable memory if flushed. The difference
+    (* Returns a 32-bit number that is the number of bytes that would
+       have been written to stable memory if flushed. The difference
        of two such number will always be an exact byte distance. *)
```
crusso, line 5999:
```suggestion
       of two such numbers will always be an exact byte distance. *)
```

src/codegen/compile.ml
```diff
@@ -5994,18 +5994,16 @@ module Stabilization = struct
 
     let finalize_buffer _ = G.nop (* everything is outputted already *)
 
-    (* Returns a 32-bit number that is reasonably close to the number of bytes
-       that would bave been written to stable memory if flushed. The difference
+    (* Returns a 32-bit number that is the number of bytes that would
```
crusso, line 5997:
```suggestion
    (* Returns a 32-bit unsigned int that is the number of bytes that would
```

crusso (approved):
LGMT

src/codegen/compile.ml
```diff
@@ -5874,7 +5874,7 @@ module BlobStream : Stream = struct
   let name_for fn_name ts = "@Bl_" ^ fn_name ^ "<" ^ Typ_hash.typ_seq_hash ts ^ ">"
 
   let absolute_offset env get_token =
```
crusso, line 5876:
```suggestion
  let absolute_offset env get_token =
```
Maybe rename this to avoid confusion with rebound `absolute_offset` below?

crusso: (or rename the other one)

ggreif: It is part of the module `sig`nature. I cannot rename just one.

PR #3205, opened 2022-04-22
rts: make `Blob` immutable by default

ggreif:
In most contexts `Blob` is being read from, so make `const` the default.
`as_blob_mut` is provided to allow write access.

### DIFF
--- a/rts/motoko-rts/src/text.rs
+++ b/rts/motoko-rts/src/text.rs
@@ -51,7 +51,7 @@ unsafe fn alloc_text_blob<M: Memory>(mem: &mut M, size: Bytes<u32>) -> Value {
 #[ic_mem_fn]
 pub unsafe fn text_of_ptr_size<M: Memory>(mem: &mut M, buf: *const u8, n: Bytes<u32>) -> Value {
     let blob = alloc_text_blob(mem, n);
-    let payload_addr = blob.as_blob().payload_addr();
+    let payload_addr = blob.as_blob_mut().payload_addr();
     memcpy_bytes(payload_addr as usize, buf as usize, n);
     blob
 }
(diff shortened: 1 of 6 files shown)


rts/motoko-rts/src/text.rs
```diff
@@ -83,11 +83,15 @@ pub unsafe fn text_concat<M: Memory>(mem: &mut M, s1: Value, s2: Value) -> Value
         let blob2 = s2.as_blob();
 
         let r = alloc_text_blob(mem, new_len);
-        let r_payload: *const u8 = r.as_blob().payload_addr();
```
ggreif, line 86: I think this should have been a `*mut` from the get-go.

rts/motoko-rts/src/types.rs
```diff
@@ -460,12 +465,16 @@ impl Blob {
         self.add(1) as *mut u8 // skip closure header
     }
 
-    pub unsafe fn len(self: *mut Self) -> Bytes<u32> {
+    pub unsafe fn payload_const(self: *const Self) -> *const u8 {
```
ggreif, line 468: Rust has no overloading, so I had to choose another name. What do you think of renaming `payload_addr` to `payload_mut`?

rts/motoko-rts/src/types.rs
```diff
@@ -213,7 +213,7 @@ impl Value {
     /// `Value::get_raw` and `unskew`) in our cost model where every Wasm instruction costs 1
     /// cycle.
     pub fn get(&self) -> PtrOrScalar {
-        if self.0 & 0b1 == 0b1 {
+        if self.0 & 0b1 != 0 {
```
ggreif, line 216: The `rustc` seems to generate identical code, but this form rings to me more like "the bit is set".

rts/motoko-rts/src/types.rs
crusso: No object, but maybe check if _addr is already used consistently elsewhere. Perhaps we should mandate hungarian notation in the backend ;->

rts/motoko-rts/src/text.rs
crusso: Because of the memcpy below? Seems reasonable.

crusso (approved):

rts/motoko-rts/src/types.rs
ggreif: Hungarian notation is a tool to enforce context-freeness in naming. As a Hungarian, I disapprove of it generally :-)

I think `payload_addr` exists for arrays too, so that would need a change also. This is not at all urgent, so let it soak and decide next time this comes up.

rts/motoko-rts/src/text.rs
ggreif: Yeah, it goes into the `dest` arg of memcpy.

PR #3206, opened 2022-04-23
Perf test suite: Add Enzo’s SHA256 module

nomeata:
this module is used in practice, I belive, and a good example for
numeric-heavy code involving fixed width nats, so maybe a good addition
to our test suite.


ggreif (approved):
Awesome idea!

PR #3207, opened 2022-04-23
Backend: Unboxed local variables

nomeata:
Right now, this code

    func go() {
      let x : Nat32 = 1 +% 1;
      assert (x *% x == 4);
    };

stores the the 32 bit nat in `x` in vanilla representation; doing
complicated shifts and possibly even allocation. To improve this, we use
the type in the AST node to determine if we can use a better stack
representation than `Vanilla`; in this case `UnboxedWord32`, and try to
use that both when filling patterns and when reading variables. This has
the following nice effect:

```diff
   (func $go (type 4) (param $clos i32)
     (local $x i32)
     i32.const 1
     i32.const 1
     i32.add
-    call $box_i32
     local.tee $x
-    call $unbox_i32
     local.get $x
-    call $unbox_i32
     i32.mul
     i32.const 4
     i32.eq
     if  ;; label = @1
     else
       i32.const 65652
       i32.const 51
       call 27
       unreachable
     end)
```

Applies to `{Nat,Int}{32,64}` and `Float`.


nomeata:
I thought I could do this optimization only easily for non-recursive lets (which we don't have yet), but on second thought it seems not too hard to do it in general and now.

TODO:
 
 * [ ] Make it work for mutable variables, and add test
 * [ ] Double-check that it doesn’t break capturing (non-mutable) values.
 * [ ] Possibly: Extend to 64 bits.
 
Interestingly, this doesn’t help with code that uses lots of `Nat8`, because there `Vanilla` is already the best we can do.

nomeata:
For
```
func go() {
 var x : Nat32 = 1 +% 1;
 x := 1 +% 1;
 assert (x *% x == 4);
};
```
we get
```diff
   (func $go (type 4) (param $clos i32)
     (local $x i32)
     i32.const 1
     i32.const 1
     i32.add
-    call $box_i32
     local.set $x
     i32.const 1
     i32.const 1
     i32.add
-    call $box_i32
     local.tee $x
-    call $unbox_i32
     local.get $x
-    call $unbox_i32
     i32.mul
     i32.const 4
     i32.eq
     if  ;; label = @1
     else
       i32.const 65652
       i32.const 51
       call 27
       unreachable
     end)

```

nomeata:
I’m a bit torn. On the one hand, the abstractions in the backend seem to hold up, and this feature is actually simpler than I expected it to. OTOH, it only really improves `Nat32` and `Int32`, so I am not sure if it worth the extra bit of complexity. OTOH, I think it’s what our users would expect – no heap allocations when working with these types.

nomeata:
Now also for `Nat64`, `Int64` and `Float`:

```diff
   (func $goFloat (type 4) (param $clos i32)
-    (local $x i32) (local $y i32)
+    (local $x f64) (local $y f64)
     f64.const 0x1p+0 (;=1;)
     f64.const 0x1p+0 (;=1;)
     f64.add
-    call $box_f64
     local.set $x
-    i32.const 66447
+    f64.const 0x1p+0 (;=1;)
     local.tee $y
-    f64.load offset=5 align=4
     f64.const 0x1p+1 (;=2;)
     f64.mul
-    call $box_f64
     local.set $y
     local.get $x
-    f64.load offset=5 align=4
     local.get $y
-    f64.load offset=5 align=4
     f64.mul
     f64.const 0x1p+2 (;=4;)
     f64.eq
     if  ;; label = @1
     else
-      i32.const 66392
+      i32.const 66376
       i32.const 53
       call 27
       unreachable
     end)
```

I think we should do this; it fits our aims of exposing Wasm closely if people can at least write simple numerical code without any overhead over the primitive types.

crusso:
Sounds great, will need to review carefully later...

src/codegen/compile.ml
```diff
@@ -7212,11 +7245,11 @@ module AllocHow = struct
...
   let how_of_ae ae : allocHow = M.map (function
-    | VarEnv.Const _ -> (Const : how)
-    | VarEnv.HeapStatic _ -> StoreStatic
-    | VarEnv.HeapInd _ -> StoreHeap
-    | VarEnv.Local _ -> LocalMut (* conservatively assume immutable *)
-    | VarEnv.PublicMethod _ -> LocalMut
+    | VarEnv.Const _        -> (Const : how)
+    | VarEnv.HeapStatic _   -> StoreStatic
+    | VarEnv.HeapInd _      -> StoreHeap
+    | VarEnv.Local (sr, _)  -> LocalMut sr (* conservatively assume immutable *)
```
crusso, line 7251:
```suggestion
    | VarEnv.Local (sr, _)  -> LocalMut sr (* conservatively assume mutable *)
```
?

src/codegen/compile.ml
```diff
@@ -9126,16 +9161,16 @@ and compile_pat_local env ae pat : VarEnv.t * patternCode =
   let fill_code = fill_pat env ae1 pat in
   (ae1, fill_code)
 
-(* Used for let patterns: If the patterns is an n-ary tuple pattern,
-   we want to compile the expression accordingly, to avoid the reboxing.
+(* Used for let patterns: If the pattern can consume its scrutinee in a better form
+   than vanilla (e.g. unboxed tuple, unboxed 32), lets do that.
```
crusso, line 9165:
```suggestion
   than vanilla (e.g. unboxed tuple, unboxed 32/64), lets do that.
```

src/codegen/compile.ml
```diff
@@ -9145,13 +9180,18 @@ and compile_n_ary_pat env ae how pat =
     match pat.it with
     (* Nothing to match: Do not even put something on the stack *)
     | WildP -> None, G.nop
```
crusso, line 9182:
```suggestion
    | WildP -> None, G.nop
```
How is `None` different from `Some (SR.UnboxedTuple 0`)? Just curious...

src/codegen/compile.ml
```diff
@@ -9179,21 +9219,21 @@ and compile_dec env pre_ae how v2en dec : VarEnv.t * G.t * (VarEnv.t -> scope_wr
...
+                     | Some (LocalMut _ | StoreHeap | StoreStatic) -> true
                      | _ -> false);
       let pre_ae1, alloc_code = AllocHow.add_local env pre_ae how name in
 
       ( pre_ae1,
         alloc_code,
-        (fun ae -> compile_exp_vanilla env ae e ^^ Var.set_val env ae name),
+        (fun ae -> let sr, code = Var.set_val env ae name in compile_exp_as env ae sr e ^^ code),
         unmodified
       )
```
crusso, line 9238: out-dent?

crusso (commented):
Good stuff! This would actually be quite useful for StableMemory address calculations, I expect.

Mostly looks good but makes me a little nervous - I'll wait for @ggreif review too, if you don't mind.

For binops, what happens if the two args are in mixed stackrep?

nomeata:
> For binops, what happens if the two args are in mixed stackrep?

The backend effectively does a bidirectional inference for the representation. Binops expect unboxed, so if any argument isn't unboxed, the `SR adjust` function will convert as needed - nothing new on this front, and the new stuff about locals fits just right in

src/codegen/compile.ml
nomeata: None means please drop, and applies to values of any type.

UnboxedTuple 0 is an invalid stack rep for all types but `()`.

I think I tried to do as you say initially back then, but somehow that didn't work nicely.

src/codegen/compile.ml
```diff
@@ -6413,10 +6423,10 @@ module VarEnv = struct
 
   (* A type to record where Motoko names are stored. *)
   type varloc =
-    (* A Wasm Local of the current function, directly containing the value
-       (note that most values are pointers, but not all)
+    (* A Wasm Local of the current function, directly containing the value,
+       in the given stackrep (Vanilla or UnboxedWord32) so far
```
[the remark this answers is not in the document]
nomeata: Or float…

```suggestion
       in the given stackrep (Vanilla, UnboxedWord32, …) so far
```

src/codegen/compile.ml
```diff
@@ -6640,7 +6660,10 @@ module FuncDec = struct
     let rec go i ae = function
     | [] -> ae
     | a::args ->
-      let ae' = VarEnv.add_local_local env ae a.it (Int32.of_int i) in
+      (* Function arguments are always vanialla, due to subtyping and uniform representation.
```
ggreif, line 6663:
```suggestion
      (* Function arguments are always vanilla, due to subtyping and uniform representation.
```

src/codegen/compile.ml
```diff
@@ -7156,12 +7179,12 @@ module AllocHow = struct
...
   let is_local_mut _ = function
-    | LocalMut -> true
+    | LocalMut _ -> true
     | _ -> false
 
   let is_local _ = function
-    | LocalImmut -> true
-    | LocalMut -> true
+    | LocalImmut _ -> true
+    | LocalMut _ -> true
```
ggreif, line 7187:
```suggestion
    | LocalImmut _ | LocalMut _ -> true
```

ggreif (approved):
Looks good, but please go through the suggestions and close them.

PR #3210, opened 2022-04-26
feat: Implement system method `inspect` to satisfy popular demand

crusso:
Implement new system method "inspect" to selectively accept or reject ingress update calls (even of queries called in update mode).

`inspect` takes a record of arguments indicating the `caller`, argument `blob`, and a variant, `msg`, of strongly typed decoding functions. The variant tag indicates the name of the underlying message; its payload is a strongly typed decoder of the (eagerly captured but lazily decoded) blob. Unlike other system methods, that have fixed types, the expected type of `inspect_message` is derived from the type of the actor's fields - in particular, their method arguments.

The argument is a record so we can extend it in future. An example might be providing a new field: the message name as text so we can do content based filtering (think regexp).

Example:

```motoko
actor {

   var c = 0;

   public func inc() : async () { c += 1 };
   public func set(n : Nat) : async () { c := n };
   public query func read() : async Nat { c };
   public func reset() : () { c := 0 }; // oneway

   system func inspect(
     { caller : Principal;
       arg : Blob;
       msg : {
         #inc : Any;
         #set : () -> Nat;
         #read : Any;
         #reset : () -> ();
      }
    }
    ) : Bool  {
      if (arg.size() > 512) return false;
      switch (msg) {
         case (#inc _) { true };
         case (#read _) { true };
         case (#set n)  { (n()) != 666 };
         case (#reset f) { false };
      }
     }
};
```

Full inspection:

[test/run-drun/inspect_message.mo](https://github.com/dfinity/motoko/blob/a844b15b567cc0ff1ac9d661d6ae4d7b9ea1cbf9/test/run-drun/inspect_message.mo)

Selective inspection:
[test/run-drun/inspect_message_blob.mo](https://github.com/dfinity/motoko/blob/a844b15b567cc0ff1ac9d661d6ae4d7b9ea1cbf9/test/run-drun/inspect_message_blob.mo)

Future enhancements: 

It's annoying that the user has to specify the type of the variant if they want to use access it. They can promote some arguments to `Any`, but do have to list the complete superset of expected variants (including hidden `__*` underscore methods. It would be nice if we could just allow the record pattern to be unannotated by using the expected type, in analysis mode. 
But let's see how many user's actually use this optional feature before we invest more effort into it.


chenyan-dfinity:
> It's annoying that the user has to specify the type of the variant if they want to use access it. They can promote some arguments to Any, but do have to list the complete superset of expected variants (including hidden __* underscore methods. It would be nice if we could just allow the record pattern to be unannotated by using the expected type, in analysis mode.

Can the users omit method names they don't want to inspect? Besides the internal ones, if the actor has 100 methods, it's tedious to list them all. Also requiring users to know method names they didn't write seems odd.

We can even provide a default rejection for `__motoko_async_helper`, because who would want to enable that?

crusso:
> > It's annoying that the user has to specify the type of the variant if they want to use access it. They can promote some arguments to Any, but do have to list the complete superset of expected variants (including hidden __* underscore methods. It would be nice if we could just allow the record pattern to be unannotated by using the expected type, in analysis mode.
> 
> Can the users omit method names they don't want to inspect? Besides the internal ones, if the actor has 100 methods, it's tedious to list them all. Also requiring users to know method names they didn't write seems odd.
>
In the pattern yes, they could use a catch all, but in the type they currently need to name all the variants (but can type argument as 'Any' if no need to decode). If I used pattern checking against inferred type, the variants could be omitted from the type. 
> Also the [spec](https://smartcontracts.org/docs/current/references/ic-interface-spec#ingress-message-inspection) says: "The canister_inspect_message is not invoked for query calls, inter-canister calls or calls to the management canister". So we can omit all the query methods?



No, I don't think so because static queries can be invoked as update methods with consensus one result but no persisted state change) and still might need filtering in that mode (I think). 

> We can even provide a default rejection for `__motoko_async_helper`, because who would want to enable that?
Yeah we could, but the others are marginally less clear cut, though I guess we could just reject calling them in upgrade mode and users could hack around that if necessary using a cast and update wrapper. Not opposed to that really.

crusso:
@chenyan-dfinity I've removed the internal method names from the variant type, rejecting them by default. I was worried that this might break `_get_interface_tmp_hack` but since it only applies to queries executed in update mode, not vanilla queries, we should be ok.

I've experimented with allowing the user to omitting the argument and return types on `inspect_message` but it requires rather gross hacks to the type checker which I'd rather avoid. Even then, I wasn't satisfied with the result.

I suggest we ship as is (once I add some doc) and revisit only if users complain. Listing all the methods is only needed if the user cares about which individual method is called.
I expect that in practice, they will be more likely to reject based on caller id, in which case the variant argument can be omitted entirely. If the user-supplied type is wrong, they should be able to use the type in the error message to cut 'n paste the correct type (or define it locally). Not great, but not terrible either and easy to maintain once they rip of the band-aid the first time.

If users really complain, we can of course revisit and adjust the type checker more radically.

chenyan-dfinity:
As a side note, I think I might have a use case to use `inspect_message` for Motoko Playground to enable actor class: the playground backend provides its own version of `create_canister` and `install_code`, and we redirect the Wasm calls to `ic0` with the playground methods.

For safety, we want to disable ingress messages for these methods, but even for inter-canister calls, I need to check if the caller comes from the canister pool. Hmm, now that I think about it, the inspect_message part is not that important, as we do ACL in the function anyway...

chenyan-dfinity (approved):

src/mo_types/type.ml
```diff
@@ -1075,14 +1075,51 @@ let glb t1 t2 = let glbs = ref M.empty in combine glbs (ref M.empty) glbs t1 t2
...
+let motoko_stable_var_size_fld =
+  { lab = "__motoko_stable_var_size";
+    typ = Func(Shared Query, Promises, [scope_bind], [], [nat64]);
+    depr = None;
+  }
+
+let get_candid_interface_fld =
+  { lab = "__get_candid_interface_tmp_hack";
+    typ = Func(Shared Query, Promises, [scope_bind], [], [text]);
+    depr = None;
```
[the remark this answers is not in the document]
crusso: Don't we just remove it?

crusso:
> Sounds good to me. I assume `inspect_message` won't trigger with ingress query method calls.

It shouldn't unless those calls are promoted to replicated update semantics - let's hope no-one is using that for getting interfaces more reliably.

src/mo_types/type.ml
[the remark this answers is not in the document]
crusso: Or maybe we keep the entrypoint, but have it reject with a deprecation message.

doc/modules/language-guide/pages/language-manual.adoc
```diff
@@ -1276,22 +1276,45 @@ The declaration `<dec>` of a `system` field must be a manifest `func` declaratio
...
+
+[#inspect-message]
+===== `inspect_message`
+
+Given a record of message attributes, this function produces a `Bool` that indicates whether to accept or decline the message by returning `true` or `false`.
+The function is invoked (by the system) on each ingress message (excluding non-replicated queries).
+Similar to a query, any side-effects of an invocation are discarded and transient.
+A call that traps due to some fault has the same result as returning `false` (message denial).
+
+The argument type `inspect_message` depends on the interface of the enclosing actor.
```
crusso, line 1307:
```suggestion
The argument type of `inspect_message` depends on the interface of the enclosing actor.
```

doc/modules/language-guide/pages/language-manual.adoc
```diff
@@ -1276,22 +1276,45 @@ The declaration `<dec>` of a `system` field must be a manifest `func` declaratio
...
+In particular, the formal argument of `inspect_message` is a record of fields of the following types:
+
+* `caller : Principal`: the principal, possibly anonymous, of the caller of the message;
+* `arg : Blob` : the raw, binary content of the message argument;
+* `msg : <variant>` where `<variant> == {...; #<id>: () -> T; ...}` is a variant of _decoding_ functions,
+   one per shared function of the actor called `<id>`. Each variant contains a function that, when applied, returns the argument of function `<id>` as a value of type `T`.
```
crusso, line 1313:
```suggestion
   one per shared function, '<id>', of the actor. The  `msg` arguments contains a function that, when applied, returns the (decoded) argument of function `<id>` as a value of type `T`.
```

doc/modules/language-guide/pages/language-manual.adoc
```diff
@@ -1276,22 +1276,45 @@ The declaration `<dec>` of a `system` field must be a manifest `func` declaratio
...
+
+Using a variant, tagged with `#<id>`, allows the return type, `T`, of the decoding function to vary with the identifier, `<id>`, of the shared function.
+
+WARNING: An actor that fails to declare system field `inspect_message` will simply accept all ingress messages.
```
crusso, line 1317:
```suggestion

The variant's arguments are functions so that one can avoid the expense of message decoding (when appropriate).

WARNING: An actor that fails to declare system field `inspect_message` will simply accept all ingress messages.
```

doc/modules/language-guide/pages/message-inspection.adoc
```diff
@@ -0,0 +1,121 @@
+= Message Inspection
+:proglang: Motoko
+:company-id: DFINITY
+
+
+On the {IC}, a canister can selectively
+https://smartcontracts.org/docs/current/references/ic-interface-spec/#ingress-message-inspection[inspect and decline] ingress messages submitted through the https interface:
```
crusso, line 7:
```suggestion
_inspect_ and _accept_ or _decline_ ingress messages submitted through the HTTP interface:
```

doc/modules/language-guide/pages/message-inspection.adoc
```diff
@@ -0,0 +1,121 @@
...
+> If the canister is empty (i.e. does not have a Wasm module), then the ingress message will be rejected.
+> If the canister is not empty and does not implement `canister_inspect_message`, then the ingress message will be accepted.
+>
+> In `canister_inspect_message`, the canister can accept the message by invoking
+`ic0.accept_message : () -> ()`.
+> This function traps if invoked twice.
+> If the canister traps in `canister_inspect_message` or does not call `ic0.accept_message`, then the access is denied.
+>
+> NOTE: The `canister_inspect_message` is _not_ invoked for query calls, inter-canister calls or calls to the management canister.
+> -- IC Interface specification
```
crusso, line 20:
```suggestion
> -- https://smartcontracts.org/docs/current/references/ic-interface-spec/#ingress-message-inspection[IC Interface Specification]
```

doc/modules/language-guide/pages/message-inspection.adoc
```diff
@@ -0,0 +1,121 @@
...
+In particular, the formal argument of `inspect_message` is a record of fields of the following types:
+
+* `caller : Principal`: the principal, possibly anonymous, of the caller of the message;
+* `arg : Blob` : the raw, binary content of the message argument;
+* `msg : <variant>` where `<variant> == {...; #<id>: () -> T; ...}` is a variant of _decoding_ functions,
+   one per shared function of the actor called `<id>`. Each variant contains a function that, when applied, returns the argument of function `<id>` as a value of type `T`.
+
+Using a variant, tagged with `#<id>`, allows the return type, `T`, of the decoding function to vary with the identifier, `<id>`, of the shared function.
+
+Exploiting {proglang}'s subtyping, the formal argument can omit record fields it does not require, or selectively ignore the arguments of particular shared functions, for example,
```
crusso, line 37:
```suggestion
Exploiting subtyping, the formal argument can omit record fields it does not require, or selectively ignore the arguments of particular shared functions, for example,
```

doc/modules/language-guide/examples/InspectTrick.mo
```diff
@@ -0,0 +1,18 @@
...
+   public query func read() : async Nat { c };
+   public func reset() : () { c := 0 }; // oneway
+/*
+// tag::inspect-trick[]
+   system func inspect_message() : Bool {
+     false
+   }
+// end::inspect-trick[]
+*/
+};
```
crusso, line 18:
```suggestion
};

```

doc/modules/language-guide/pages/message-inspection.adoc
```diff
@@ -0,0 +1,121 @@
...
+> A canister can inspect ingress messages before executing them. When the IC receives an update call from a user, the IC will use the canister method `canister_inspect_message` to determine whether the message shall be accepted.
+> If the canister is empty (i.e. does not have a Wasm module), then the ingress message will be rejected.
+> If the canister is not empty and does not implement `canister_inspect_message`, then the ingress message will be accepted.
+>
+> In `canister_inspect_message`, the canister can accept the message by invoking
+`ic0.accept_message : () -> ()`.
+> This function traps if invoked twice.
+> If the canister traps in `canister_inspect_message` or does not call `ic0.accept_message`, then the access is denied.
+>
+> NOTE: The `canister_inspect_message` is _not_ invoked for query calls, inter-canister calls or calls to the management canister.
```
crusso, line 19:
```suggestion
> NOTE: The `canister_inspect_message` is _not_ invoked for HTTP query calls, inter-canister calls or calls to the management canister. 
> NOTE:  Confusingly, a `shared query` function _can_ be called using a regular HTTP call to obtain a certified response: this is why the variant type also includes `shared query` functions.
```

test/fail/inspect_message_wrong.mo
```diff
@@ -0,0 +1,12 @@
...
+   var c = 0;
+
+   public func inc() : async () { c += 1 };
+   public func set(n : Nat) : async () { c := n };
+   public query func read() : async Nat { c };
+   public func reset() : () { c := 0 }; // oneway
+
+   system func inspect_message()  {}
+
+};
```
crusso, line 12:
```suggestion
};

```

doc/modules/language-guide/pages/language-manual.adoc
```diff
@@ -1276,22 +1276,49 @@ The declaration `<dec>` of a `system` field must be a manifest `func` declaratio
 |  name | type | description
 
 | `heartbeat`  | `+() -> async ()+` | heartbeat action
+| `inspect_message` | `+{ caller : Principal; msg : <Variant>; arg : Blob } -> Bool+` | message predicate
```
ggreif, line 1279: `inspect` sounds better to me.

doc/modules/language-guide/pages/language-manual.adoc
```diff
@@ -1276,22 +1276,49 @@ The declaration `<dec>` of a `system` field must be a manifest `func` declaratio
...
 | `postupgrade` | `+() -> ()+` | post upgrade action
+
 |===
 
 * `heartbeat`, when declared, is called on every Internet Computer subnet *heartbeat*,
 scheduling an asynchronous call to the `heartbeat` function. Due to its `async` return type, a heartbeat function may send messages and await results. The result of a heartbeat call, including
 any trap or thrown error, is ignored. The implicit context switch means that
 the time the heartbeat body is executed may be later than the time the
 heartbeat was issued by the subnet.
+* `inspect_message`, when declared, is called as a predicate on every Internet Computer ingress message (with the exception of queries). The return value, a `Bool`, indicates whether to accept or decline the given message. The argument type depends on the interface of the enclosing actor (see <<inspect-message>>).
```
ggreif, line 1290: You told that queries going through concensus will also pass through `inspect_message`. Also we might explain that canister-to-canister sends are not filtered.

doc/modules/language-guide/pages/language-manual.adoc
```diff
@@ -1276,22 +1276,49 @@ The declaration `<dec>` of a `system` field must be a manifest `func` declaratio
...
 
-During an upgrade, a trap occuring in the implicit call to `preupgrade()` or `postupgrade()` causes the entire upgrade to trap, preserving the pre-upgrade actor.
+During an upgrade, a trap occurring in the implicit call to `preupgrade()` or `postupgrade()` causes the entire upgrade to trap, preserving the pre-upgrade actor.
+
+[#inspect-message]
+===== `inspect_message`
+
+Given a record of message attributes, this function produces a `Bool` that indicates whether to accept or decline the message by returning `true` or `false`.
+The function is invoked (by the system) on each ingress message (excluding non-replicated queries).
+Similar to a query, any side-effects of an invocation are discarded and transient.
```
ggreif, line 1304:
```suggestion
Similar to a query, any side-effects of an invocation are transient and discarded eventually.
```

doc/modules/language-guide/pages/message-inspection.adoc
```diff
@@ -0,0 +1,132 @@
+= Message Inspection
+:proglang: Motoko
+:company-id: DFINITY
+
+
+On the {IC}, a canister can selectively
+_inspect_ and _accept_ or _decline_ ingress messages submitted through the HTTP interface:
+
+
+> A canister can inspect ingress messages before executing them. When the IC receives an update call from a user, the IC will use the canister method `canister_inspect_message` to determine whether the message shall be accepted.
```
ggreif, line 10: `canister_inspect_message` ,- yer mother term?

ggreif: Ah, this is the Wasm-module level interface? Can Motoko actors add such a method? Probably not, since all callables must have a space char in the name.

doc/modules/language-guide/pages/message-inspection.adoc
```diff
@@ -0,0 +1,132 @@
...
+> This function traps if invoked twice.
+> If the canister traps in `canister_inspect_message` or does not call `ic0.accept_message`, then the access is denied.
+>
+> The `canister_inspect_message` is _not_ invoked for HTTP query calls, inter-canister calls or calls to the management canister.
+> -- https://smartcontracts.org/docs/current/references/ic-interface-spec/#ingress-message-inspection[IC Interface Specification]
+
+Message inspection mitigates some denial of service attacks,
+designed to drain canisters of cycles by placing unsolicited free calls.
+
+REMARK: You can think of `inspect_message` as providing the "Collect call from <name>. Do you accept charges?" prologue of an old-fashioned, operator-assisted, collect phone call.
```
ggreif, line 25: `inspect_message` is Motoko lingo.

doc/modules/language-guide/pages/language-manual.adoc
ggreif, line 1290:
```suggestion
* `inspect_message`, when declared, is called as a predicate on every {IC} ingress message (with the exception of queries). The return value, a `Bool`, indicates whether to accept or decline the given message. The argument type depends on the interface of the enclosing actor (see <<inspect-message>>).
```

doc/modules/language-guide/pages/language-manual.adoc
```diff
@@ -1276,22 +1276,49 @@ The declaration `<dec>` of a `system` field must be a manifest `func` declaratio
 |  name | type | description
 
 | `heartbeat`  | `+() -> async ()+` | heartbeat action
+| `inspect_message` | `+{ caller : Principal; msg : <Variant>; arg : Blob } -> Bool+` | message predicate
 | `preupgrade`  | `+() -> ()+` | pre upgrade action
 | `postupgrade` | `+() -> ()+` | post upgrade action
+
 |===
 
 * `heartbeat`, when declared, is called on every Internet Computer subnet *heartbeat*,
```
ggreif, line 1285:
```suggestion
* `heartbeat`, when declared, is called on every {IC} subnet *heartbeat*,
```

doc/modules/language-guide/pages/language-manual.adoc
ggreif:
```suggestion
* `inspect_message`, when declared, is called as a predicate on every {IC} ingress message (with the exception of queries). The return value, a `Bool`, indicates whether to accept or decline the given message. The argument type depends on the interface of the enclosing actor (see <<inspect-message>>).
```

doc/modules/language-guide/pages/message-inspection.adoc
```diff
@@ -0,0 +1,132 @@
...
+to obtain the required argument type.
+
+For example, in the actor from the previous section, incorrectly declaring:
+
+[source.no-repl,motoko]
+----
+include::../examples/InspectTrick.mo[tag=inspect-trick]
+----
+
+forces the compiler to cough up the expected type below:
```
ggreif, line 112: reveal? :-)

test/run-drun/inspect_message_blob.mo
```diff
@@ -0,0 +1,15 @@
+// NB: inspect_message cannot be tested in drun.
```
ggreif, line 1: then maybe `drun` should be enhanced?

ggreif (commented):
Looks good , but `system func inspect_message` looks very clumsy.

doc/modules/language-guide/pages/message-inspection.adoc
crusso:
```suggestion
forces the compiler to report the expected type below:
```

test/run-drun/inspect_message_blob.mo
crusso: Maybe we, or we add a new class of test that uses dfx (though that inverts the dependency relationship). I'm not sure we have the resources to extend drun right now.

doc/modules/language-guide/pages/language-manual.adoc
crusso: I could live with that. Let's see what the others say.

doc/modules/language-guide/pages/language-manual.adoc
crusso:
```suggestion
Similar to a query, any side-effects of an invocation are transient and discarded.
```

doc/modules/language-guide/pages/message-inspection.adoc
crusso:
```suggestion
REMARK: You can think of method inspection as providing the "Collect call from <name>. Do you accept charges?" prologue of an old-fashioned, operator-assisted, collect phone call.
```

doc/modules/language-guide/pages/message-inspection.adoc
crusso: Yes this is literally quoting from the spec so talking about canisters not actors.

doc/modules/language-guide/pages/message-inspection.adoc
```diff
@@ -0,0 +1,132 @@
...
+
+The variant's argument is a function so that one can avoid the expense of message decoding (when appropriate).
+
+Exploiting subtyping, the formal argument can omit record fields it does not require, or selectively ignore the arguments of particular shared functions, for example,
+in order to simply dispatch on the name of a function without inspecting its actual argument.
+
+NOTE:  Confusingly, a `shared query` function _can_ be called using a regular HTTP call to obtain a certified response: this is why the variant type also includes `shared query` functions.
+
+WARNING: An actor that fails to declare system field `inspect_message` will simply accept all ingress messages.
+
```
crusso, line 52:
```suggestion

WARNING: `inspect_message` should **not** be used for definitive access control and any reliable access control checks should be performed within the `shared` functions guarded by `inspect_message`. This is because `inspect_message` is executed by a single replica, without consensus. and its result could be spoofed by a malicious boundary node.
```

doc/modules/language-guide/pages/language-manual.adoc
```diff
@@ -1276,22 +1276,49 @@ The declaration `<dec>` of a `system` field must be a manifest `func` declaratio
...
+
 |===
 
-* `heartbeat`, when declared, is called on every Internet Computer subnet *heartbeat*,
+* `heartbeat`, when declared, is called on every {IC} subnet *heartbeat*,
 scheduling an asynchronous call to the `heartbeat` function. Due to its `async` return type, a heartbeat function may send messages and await results. The result of a heartbeat call, including
 any trap or thrown error, is ignored. The implicit context switch means that
 the time the heartbeat body is executed may be later than the time the
 heartbeat was issued by the subnet.
+* `inspect_message`, when declared, is called as a predicate on every {IC} ingress message (with the exception of queries). The return value, a `Bool`, indicates whether to accept or decline the given message. The argument type depends on the interface of the enclosing actor (see <<inspect-message>>).
```
crusso, line 1290:
```suggestion
* `inspect_message`, when declared, is called as a predicate on every {IC} ingress message (with the exception of HTTP query calls). The return value, a `Bool`, indicates whether to accept or decline the given message. The argument type depends on the interface of the enclosing actor (see <<inspect-message>>).
```

ggreif (approved):
Looks good to me. I wonder what happens when the `inspect` system method tries to send a message. The future will be built, but will anything be inserted in the ingest queue? I then the state will be rolled back. What about other side effects like `print`?

crusso:
> Looks good to me. I wonder what happens when the `inspect` system method tries to send a message. The future will be built, but will anything be inserted in the ingest queue? I then the state will be rolled back. What about other side effects like `print`?

It's not an async function, so Motoko won't let you send. Not sure about Debug.print - at worst it'll trap I imagine.

PR #3211, opened 2022-04-27
Doing `Stream`-based leb128 encoding

ggreif:
Implement routines for encoding bignums directly into the stream. This path is slightly more expensive than encoding into a buffer, so we move the tight-loop bottleneck function into `stream.rs`, such that we get access to the (otherwise opaque) `Stream` members.

This fixes a real problem, where arbitrary buffer reservations (extending beyond the legal end of the cache) could interfere with heap traffic because of bignum arithmetic (i.e. 7-bit shifts) happening while encoding.

Resolves the residual problem from #3149.

---------
TODO:
- [x] Why is `stable-exp.mo` failing suddenly? — probably 495c91f
- [ ] add `QuickCheck` (de)serialisation tests for `Nat/Int`
- [x] explore having a `stream::write_(s)leb128`


crusso:
I took a quick peek, but I wonder if you can actually simplify this by adding the bigint (sleb_word) encoders as methods on stream and repeatedly writing as much as possible to the cache directly instead of calling a Stream.cache() per byte on the slow path.

I think there would also be no need to count the bytes needed for the (now) redundant fast path.

This way we might be able to completely remove the troublesome `reserve` and new `try_reserve` function. 

Happy to discuss tomorrow sometime.

rts/motoko-rts/src/bigint.rs
```diff
@@ -140,11 +142,11 @@ unsafe fn mp_get_u64(p: *const mp_int) -> u64 {
     mp_get_i64(p) as u64
 }
 
-unsafe fn mp_isneg(p: *const mp_int) -> bool {
+pub(crate) unsafe fn mp_isneg(p: *const mp_int) -> bool {
     (*p).sign == 1
```
ggreif, line 146: this might better be `(*p).sign != 0`.

ggreif:
@crusso I have now `stream::write_leb128` now that implements the inner loop for serialisation. PTAL.

src/codegen/compile.ml
```diff
@@ -2530,6 +2538,50 @@ module MakeCompact (Num : BigNumType) : BigNumType = struct
...
+    try_unbox I32Type
+      (fun env ->
+        BitTagged.untag_i32 ^^ set_x ^^
+        (* get size & reserve & encode *)
+        I32Leb.compile_store_to_data_buf_unsigned env get_x
+          begin
+            get_stream ^^
+            I32Leb.compile_leb128_size get_x ^^
+            E.call_import env "rts" "stream_reserve"
+          end)
```
crusso, line 2555:
```suggestion
        let set_dst, get_dst = new_local env "dst" in
        get_stream ^^
        I32Leb.compile_leb128_size get_x ^^
        E.call_import env "rts" "stream_reserve" ^^
        set_dst  ^^
        I32Leb.compile_store_to_data_buf_unsigned env get_x get_dst)
```
Might be easier to follow (assuming equivalent). Ditto below.

crusso (approved):
LGTM.

I was looking at our existing functions and it seems a pity that the (big and small) leb encoders just return void, not the number of bytes written. Then we wouldn't always need to recompute the size after encoding. But not worth fixing now (if that's even a good idea).

src/codegen/compile.ml
ggreif: It looks equivalent (functionality-wise) but it adds two extra instructions, so I prefer to leave as-is.

ggreif:
> not the number of bytes written

They should return the current data pointer, as it is cheap to have, and then the caller doesn't need to add to it.

src/codegen/compile.ml
ggreif: I am using an OCaml `let` instead.

PR #3213, opened 2022-04-28
Do not merge.

matthewhammer:


src/lowering/desugar.ml
```diff
@@ -68,6 +68,23 @@ and exp' at note = function
...
+  | S.ToCandidE es ->
+    (* desugaring to_candid introduces explicit IR tuple: *)
+    let arg = tupE (exps es) in
+    begin match T.normalize arg.note.Note.typ with
+    | T.Tup ts (* our original idea: *)
+        when false -> (primE (I.SerializePrim [T.Tup ts]) [arg]).it
+
+    | T.Tup ts -> (* my proposal 2022-04-28: *)
+      (primE (I.SerializePrim ts) (exps es)).it
+
```
matthewhammer, line 80: @crusso I keep wanting to deviate from our original plan, and try something like above, where we use the argument-list structure of the primitive, instead of eschewing it with an extra tuple.

I've actually convinced myself that we need to do this if we really want to encode and decode messages.  Otherwise, we can never encode a multi-component message with `to_candid`, which is actually what `from_candid` expects, hence the [deserialization mismatch in our original PR](https://github.com/dfinity/motoko/pull/3155/files#r861110699).

matthewhammer: WDYT?

matthewhammer:
This PR has served its purpose (communication).  Closing.

PR #3220, opened 2022-05-03
fix: use symbolic mangling in rts to avoid platform differences

crusso:
We currently produces different rts binaries on darwin and linux, due to the legacy Rust mangling scheme.

This PR instructs rustc to use the new(ish) symbolic mangling scheme in the hope that it resolves platform differences.

### DIFF
--- a/rts/motoko-rts/.cargo/config
+++ b/rts/motoko-rts/.cargo/config
@@ -1,4 +1,4 @@
 # NB. codegen-units=1 is not necessary, but it generates less .o files for core,
 # std, and compiler_builtins and makes it easier to find symbols.
 [build]
-rustflags = ["-Crelocation-model=pic", "-Ccodegen-units=1"]
+rustflags = ["-Crelocation-model=pic", "-Ccodegen-units=1", "-Csymbol-mangling-version=v0"]
--- a/rts/motoko-rts/.cargo/config
+++ b/rts/motoko-rts/.cargo/config
@@ -1,4 +1,4 @@
 # NB. codegen-units=1 is not necessary, but it generates less .o files for core,
 # std, and compiler_builtins and makes it easier to find symbols.
 [build]
-rustflags = ["-Crelocation-model=pic", "-Ccodegen-units=1", "-Csymbol-mangling-version=v0"]
+rustflags = ["-Crelocation-model=pic", "-Ccodegen-units=1", "-Csymbol-mangling-version=v0", "-Zremap-cwd-prefix=."]
--- a/rts/motoko-rts/src/lib.rs
+++ b/rts/motoko-rts/src/lib.rs
@@ -93,8 +93,9 @@ pub(crate) unsafe fn rts_trap_with(msg: &str) -> ! {
 
 #[cfg(feature = "ic")]
 #[panic_handler]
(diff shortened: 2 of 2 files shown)


ggreif:
Is this `symbol-mangling-version=v0` a legacy option? I am a bit concerned about it being a dead-end...

nomeata:
We could also, at least for the official releases made by CI, use the linux-built RTS for both platforms. Easy due to nix.

crusso:
> We could also, at least for the official releases made by CI, use the linux-built RTS for both platforms. Easy due to nix.

Given that the hack in this PR doesn't actually work, I'd be up for that. Not sure my Nix foo is up to it though. @ggreif you seem more confident about hacking our build system....

nomeata:
Why does it matter that the two compilers produce different rtss?

crusso:
> Why does it matter that the two compilers produce different rtss?

Roman is trying to use bazel to build the replica and wants the darwin and linux mocs to produce identical binaries.

nomeata:
Hmm, I’d punt for that now. We could make it work, but it would complicate CI (we’d have separate build-rts and build steps, hope Github runs them in the right order, and uses the cache properly to pass the artifact between them).

Or at least I’ll drag my feet a bit to see if the request persists or vanishes :-)

How does Rust fare on this front?

crusso:
Our definition of `panic` in lib.rs is causing mangled names to be exported (dunno why, but something to do with the fmt crate and panic's attempt to give useful info with a crash). If I simplify it, the mangled exports disappear, GOT relocations seem unnecessary (the dylink section is empty) and I think it might just work (once I strip the name section, or drop it from the motoko linker). All very fragile... Anyone know a good tool for stripping the name section only? wasm-strip is too caustic and takes out the `dylink` section too.

crusso:
Ok, after stripping out the name section I get the binaries to differ in just one constant in some panic related code: a TypeId generated by the Rust compiler in some host-platform specific way. At this point, I declare defeat and Rust rather unsuited to bit compatible cross-compilation from different platforms. Who'd a thunk. 

I think our best bet is to do as Joachim suggests and build the runtime on one platform (linux) and embed the same wasm binary in both unix and darwin moc, and hope that ocaml doesn't have a similar problem producing the rest of our binaries.

Sigh.

crusso:
FTR, this is the one difference (before stripping name section in case anyone has a lead:

Darwin:
```wasm
  (func $_RNvXNtCs840lMrrG1oY_4core3anyNtNvMNtNtB4_5panic10panic_infoNtBw_9PanicInfo20internal_constructor9NoPayloadNtB2_3Any7type_idB4_a (type 10) (param i32) (result i64)
    i64.const 1863502032208819429)
```

Linux
```wasm
  (func $_RNvXNtCslSBBKmmhGCS_4core3anyNtNvMNtNtB4_5panic10panic_infoNtBw_9PanicInfo20internal_constructor9NoPayloadNtB2_3Any7type_idB4_ (type 10) (param i32) (result i64)
    i64.const -3347047285503844273)
```

chenyan-dfinity:
Ah, if it's the `TypeId` from Rust, `rustc` doesn't guarantee the same id on different platforms. I wonder if the same problem happens in Rust CDK.

luc-blaeser:

PR #3224, opened 2022-05-05
rts: micro-optimise `mp_isneg`

ggreif:
Non-negatives have `sign == 0`, so using inequality. There are only two possibilities:
https://github.com/libtom/libtommath/blob/develop/mtest/mpi.h#L35


crusso:
Are these micro-optimization actually worth doing?

ggreif:
> Are these micro-optimization actually worth doing?

That always depends on the assumed metric. But here we get a real improvement at a very small cost (code change) and can even simplify the code by not using the `1` magic.

Issue #2903, opened 2021-11-17
GC: Avoid pushing large array fields when marking arrays


osa1:
Currently in mark-compact GC (soon again the default collector), when marking an array, we push all elements of the array to the mark stack.

For large arrays this can be problematic as mark stack can grow quite large.

Instead, if we push something like `(array offset, array end)` to the mark stack, the unmarked range of the array will be represented as one entry in the mark stack.

The main question is how to represent this pair `(array offset, array end)` without making every other entry in the mark stack one word larger.

Note that when pushing a `(array offset, array end)` pair for an array for the first time, we should mark every element in the range (similar to how we mark objects when pushing to the mark stack). This is to avoid pushing these objects to the mark stack in the case where every object in the array is also referenced by other objects in the heap. In that case if we push `(array offset, array end)` without marking the elements in the range, it is possible that we will visit other objects that refer to the objects referenced by the array and end up pushing all of them to the stack, basically ending up with where we started.

matthewhammer:
The proposal makes sense, even without me fully appreciating the implementation (yet).

I'm curious about what we can say about the best case behavior for this proposal?   Mainly, I want to confirm that I see the benefit clearly.

Suppose that the heap is mostly a huge array-based hash table, and there is no cycle from which we can travel through the array more than once (entering initially from some root pointer, for instance).  Suppose there are `N` entries in the table.

In that case, for a system with `N` records, we avoid pushing `O(N)` on the mark stack, but will "eagerly" visit each of those locations rather than use the mark stack to remember to visit them (the current behavior, IIUC).

So, if the mark stack is supposed to be something like at most `O(log N)`, then this improvement works, by letting us operate within that bound, and avoid growing it, and potentially running out of resources.

But this leads me to wonder what is the most pathological data structure for the new mark stack.  Suppose that I fill memory with a graph of small records, and no arrays?  That's outside of the scope of this issue, but is there another one that tracks it?

osa1:
> In that case, for a system with N records, we avoid pushing O(N) on the mark stack, but will "eagerly" visit each of those locations rather than use the mark stack to remember to visit them (the current behavior, IIUC).

[...]

PR #3231, opened 2022-05-11
feat: introduce array-slice processing for compacting GC

ggreif:
This PR introduces a modified marking (and visiting) technique for array fields, where `Array`s are GC-ed by considering their suffixes as slices, pushing these on the mark stack as bulk collections, and treating the fields in the prefix as individual objects being pushed onto the mark stack.

Following key changes are made:
- when visiting an `Array` in the dynamic heap we consult a new callback that can opt-out of the individual processing of some suffix of the array (by returning an index smaller than the array length)
- fields before this index will be processed as before by the first callback
- for the suffix the new callback is in charge to perform the bulk action
- in case of array marking, the new callback checks if the current slice is longer than some cutoff, and if so, it will push the array and the start of the suffix to the mark stack. (We abuse the tag entries as the start index for the suffix slice, so the cutoff must be numerically bigger than the biggest heap tag; this invariant is enforced)
- when pulling entries `(obj, tag)` from the mark stack a `tag >= TAG_ARRAY_SLICE_LOW_LIMIT` indicates that we have to deal with a suffix slice of an `Array`. The visitor will then again check the slice's length and do the subdivision again, as necessary.

This way of treating suffix-slices as pseudo-heapobjects originates from the suggestion of @ulan in a SGM (2022-05-30). It sidesteps the issue with layering violations, and restricts the handling of pseudo-tags to a very restricted portion of the collector source code, viz. the pushing of the range, and the visitor's case analysis on the tag, which now needs to  know about the suffix slices. Fortunately the visitor already accepts the heap tag as an argument (and doesn't extract it from the object; which would be unreliable due to threading), so the appearance of pseudo-tags is not a big deal.

The cutoff is currently chosen as 127, so for any `Array` at most 128 new entries are being placed on the mark stack.

The charm of this solution is that the same code is doing the marking pushing/popping and visiting as before, only that now it happens in an interleaved manner. The mark is already present on the array, so no reëntrancy can't occur, and eventually all its fields get processed.

-----

Below is an account of how field-by-field marking, pushing and visiting was done formerly (still in use for prefix portion of arrays), and an aborted attempt to a scheme as suggested by @osa1. I leave it for reference.

----------
See #2903. This reduces the marking stack traffic for GC-ed arrays substantially. Not applicable to copying GC!

## Status quo

Here is what happens to an array object's (single) field while `mark_fields`:
 - considered only when it is a skewed pointer and points into the dynamic heap (otherwise it is a _static object_ or a _magic_, no need to GC)
 - visitor callback is invoked:
   - `mark_object` will place a bit and push to the mark stack when not marked yet
     (when being pushed onto the mark stack the (unskewed) pointer and the heap tag are remembered)
   - the field address gets `thread()`-ed when the contained object is physically located at a less or equal address than the array itself
     (threading will put the field address into the object tag saving the tag into the word where the field address points to; the old object pointer is overwritten, but may live on in the mark stack)

## How bulk marking could work (ABANDONED)

When discovering a sequence of value fields that are laid out successively in memory and a count field in front, we can do a bunch of nice shortcuts. Below items sketch the how this could work. Invariant: there must be a tag field immediately before the length field.

- the bulk visitor callback gets called with a pointer to the length field
- it decides that it is worth doing a ranged marking and divides the bulk into a (small) prefix and a non-empty suffix `(pointer_to_length, start_index)`
- set all marks corresponding to (dynamic heap) objects in the suffix to avoid re-visiting (see [open issue](#Revisit-avoidance-is-more-complicated))
- push `(pointer_to_length, start_index)` onto the mark stack, using low bits in the pointer to differentiate from `(obj, tag)` pairs
- bulk visitor callback returns the prefix length, the fields in this area will be visited individually

When it comes to popping the mark stack and there is a field range on its top we want to be as transparent as possible, but also have to perform steps that had to be skipped (relative to the proceeding on individual fields)

 - identify that the ToS is a range `(pointer_to_length, start_index)`
 - obtain a pointer to the field indexed by `start_index`
 - increment `start_index` (still remains on the stack)
 - if  `start_index` is equal to the length, pop the entry
 - check that the field is a GC-able pointer
 - if so, extract the object and the heap tag corresponding to the field
 - otherwise pop again and return that
 - apply the threading to the field address (use the invariant to get the home object and apply the threading criterion)
 - return the object and the heap tag

## Open issues

There are some ugly wrinkles in this design, that need to be addressed somehow.

### Revisit avoidance is more complicated

An object is either unmarked, in the mark stack or is being (has been) visited. When pushing entire ranges, we cannot process individual objects and as such marking and threading may happen in unexpected ways.

#### Saving the marked bit

When eagerly marking the whole range of fields, we must not drop the info whether the individual mark is already set. This is because a mark means that either an object is in the mark stack or being/have-been visited. We have two possibilities:
- mark range eagerly on push, but remember previous marks (e.g. in the unused bit of the `Value` in the fields), or
- instead mark lazily on popping an object from the range, performing the threading anyway, but never returning already marked objects.

The latter option is in conflict with what #2903 suggests, and could potentially invalidate the idea of pushing ranges at all.

#### Intervening `thread()` call after range marking

After a range has been pushed and all dynamic objects marked, an individual field with an object also referenced by the range might become threaded. In this case the tag field of that object gets overwritten, but can be distinguished from regular tags. Such objects got individually pushed and already popped from the stack, and thus have been visited. So the field in the range must be threaded, but the pointed object must be skipped (not visited again).

### Layering violation

The fact that the `pop_mark_stack` needs to know about object layout and then how to do the threading is disgusting.
Maybe there should be a callback parameter that is invoked then a non-pointer (i.e. range) entry is encountered in ToS position.

### `pointer_to_dynamic_heap` needs the heap base

If `pop_mark_stack` is to check pointers for pointing into the dynamic heap, it needs to receive the heap base, but it currently doesn't. It is open if the caller has this piece of information. (It has, what a relief!)

### Special format of `usize` passed to `push_mark_stack` and `push_range_mark_stack`

Distinguishing by the lowest bit is hacky.

## Further improvement opportunities

I spotted below optimisation tricks while reading the code.

 - `pointer_to_dynamic_heap` unskews a lot, why not compare with a `heap_base` that is 1 less?
 - `mark_object` doesn't need to get the previous mark (`bool`), but compare the (64-bit) word the mark is in whether it changed
 - `mark_object` unskews. Would it be possible to do `set_bit` using the `raw_ptr`?
 - `field_value.get_ptr() <= obj` could be `field_value.raw_ptr() + 1 <= obj` and further `field_value.raw_ptr() < obj` saving us an addition
 - can the double storing of the heap tag (by `thread` and `push_mark_stack`) be consolidated?
   (we have 2 cases for whether `thread` happens and 2 cases whether the `push_mark_stack` happens)
 - can the dynamic heap pointer check be reduced to a shift and a comparison with `BITMAP_FORBIDDEN_PTR` (at least while marking, as it is not available in copying GC)?


crusso:
This sounds great. Go for it. I would leave the optimisations for a separate PR.

crusso:
> * set all marks corresponding to (dynamic heap) objects in the suffix to avoid re-visiting.
This smells quadratic to me - if you are doing it on every iteration over the same array - is it really necessary or am I misunderstanding?

The special object placed on the stack is essentially a defunctionalized continuation, i think, recording where to continue when popped. I don't see the case for the "continuing" bit. When you try to pop and there is more work to do for the same array, won't you need to push the next chunk of objects?

crusso:
For the actual mark stack entries, can we push two words for an array continuation, using a bit in the topmost entry to distinguish from skewed pointer and one skewed pointer for all other entries.

crusso:
@ggreif, @ulan mentioned that this optimization is actually fairly standard. If you get stuck, perhaps he can point you to something in the literature on this. I guess integrating this with the optimized two pass threading might be an additional challenge.

ggreif:
I am chipping away the failures, and currently have
```
Some tests failed:
stable-exp.mo stable-mem-grow.mo stable-size-overflow.mo stable-size-overflow.mo
```

rts/motoko-rts-tests/src/mark_stack.rs
```diff
@@ -57,12 +57,12 @@ fn test_<M: Memory>(mem: &mut M, n_objs: u32) -> TestCaseResult {
         alloc_mark_stack(mem);
 
         for obj in &objs {
-            push_mark_stack(mem, *obj as usize, TAGS[(*obj as usize) % TAGS.len()]);
+            push_mark_stack(mem, *obj as usize * 2, TAGS[(*obj as usize) % TAGS.len()]);
```
ggreif, line 60: this is not needed any more, probably

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -218,17 +242,24 @@ unsafe fn update_refs<SetHp: Fn(u32)>(set_hp: SetHp, heap_base: u32) {
...
+                thread(field_addr)
+            }
+        },
+        |_, _, arr| arr.len(),
+    );
 }
 
 /// Thread a pointer field
-unsafe fn thread(field: *mut Value) {
+pub(crate) unsafe fn thread(field: *mut Value) {
```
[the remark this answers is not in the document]
ggreif: Nope!

ggreif:
@ulan I have redone the concept, spurred by your suggestions in Monday's SGM. PTAL!

rts/motoko-rts/src/visitor.rs
```diff
@@ -3,13 +3,16 @@ use crate::types::*;
...
 /// callback
-pub unsafe fn visit_pointer_fields<F>(
+pub unsafe fn visit_pointer_fields<C, F, G>(
+    ctx: &mut C,
     obj: *mut Obj,
     tag: Tag,
     heap_base: usize,
-    mut visit_ptr_field: F,
+    visit_ptr_field: F,
+    visit_field_range: G,
```
ulan, line 12: Could you please describe the contract of the new callback in the comments?

rts/motoko-rts/src/visitor.rs
```diff
@@ -18,18 +21,25 @@ pub unsafe fn visit_pointer_fields<F>(
...
                 let field_addr = obj_payload.add(i as usize);
                 if pointer_to_dynamic_heap(field_addr, heap_base) {
-                    visit_ptr_field(obj_payload.add(i as usize));
+                    visit_ptr_field(ctx, obj_payload.add(i as usize));
                 }
             }
         }
 
-        TAG_ARRAY => {
+        TAG_ARRAY | TAG_ARRAY_SLICE_LOW_LIMIT.. => {
```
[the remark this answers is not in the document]
ggreif: If we add a new tag, we can also increase `TAG_ARRAY_SLICE_LOW_LIMIT`. I don't think this is a big issue.

rts/motoko-rts/src/visitor.rs
ggreif: See d10949a956c08f3bf11d5e0eb383ac80c48ce278.

rts/motoko-rts/src/types.rs
```diff
@@ -351,6 +351,9 @@ pub const TAG_CONCAT: Tag = 25;
 pub const TAG_NULL: Tag = 27;
 pub const TAG_ONE_WORD_FILLER: Tag = 29;
 pub const TAG_FREE_SPACE: Tag = 31;
+// special value to visit only a range
```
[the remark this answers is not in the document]
ggreif: See c9b2839f42cfed28520f6ceb644e8619d85130c1.

rts/motoko-rts/src/visitor.rs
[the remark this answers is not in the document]
ggreif: See c9b2839f42cfed28520f6ceb644e8619d85130c1.

crusso:
Looks very nice from a distance. Not sure I can review carefully enough tonight, but hopefully tomorrow.

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -220,11 +238,18 @@ unsafe fn update_refs<SetHp: Fn(u32)>(set_hp: SetHp, heap_base: u32) {
 
 /// Thread forward pointers in object
 unsafe fn thread_fwd_pointers(obj: *mut Obj, heap_base: u32) {
-    visit_pointer_fields(obj, obj.tag(), heap_base as usize, |field_addr| {
-        if (*field_addr).get_ptr() > obj as usize {
-            thread(field_addr)
-        }
-    });
+    visit_pointer_fields(
+        &mut (),
```
crusso, line 242: What does this do? Is that a dummy pointer?

ggreif: Yes. The receiving side wants a `*mut`, so we give it one.

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -158,15 +158,33 @@ unsafe fn mark_stack<M: Memory>(mem: &mut M, heap_base: u32) {
...
-        let field_value = *field_addr;
-        mark_object(mem, field_value);
-
-        // Thread if backwards or self pointer
-        if field_value.get_ptr() <= obj as usize {
-            thread(field_addr);
-        }
-    });
+    visit_pointer_fields(
+        mem,
```
crusso, line 162: Is the mem argument still required? Looks like the old code manage to capture mem in the first callback without it.

rts/motoko-rts/src/types.rs
```diff
@@ -352,6 +352,14 @@ pub const TAG_NULL: Tag = 27;
 pub const TAG_ONE_WORD_FILLER: Tag = 29;
 pub const TAG_FREE_SPACE: Tag = 31;
 
+// Special value to visit only a range of array fields.
+// This and all values above it are reserved and mean
+// a slice of an array object (i.e. start index) for
+// purposes of `visit_pointer_fields`.
+// Invariant: the value of this (pseudo-)tag must be
+//            higher than all other tags defined above
+pub const TAG_ARRAY_SLICE_LOW_LIMIT: Tag = 32;
```
crusso, line 361:
```suggestion
pub const TAG_ARRAY_SLICE_MIN: Tag = 32;
```

rts/motoko-rts/src/visitor.rs
```diff
@@ -3,13 +3,33 @@ use crate::types::*;
 
 /// A visitor that passes field addresses of fields with pointers to dynamic heap to the given
 /// callback
-pub unsafe fn visit_pointer_fields<F>(
+///
+/// Arguments:
+///
+/// * `ctx`: any context passed to the `visit_*` callbacks
```
crusso, line 9: Drop `ctx`?

rts/motoko-rts/src/visitor.rs
```diff
@@ -3,13 +3,33 @@ use crate::types::*;
...
+///   Arguments:
+///   * `&mut C`: passed context
+///   * `u32`: start index of array suffix slice being visited
+///   * `*mut Array`: home object of the slice (its heap tag may be invalid)
+///   Returns:
+///   * `u32`: start of the suffix slice of fields not to be passed to `visit_ptr_field`;
+///            it is the callback's responsibility to deal with the spanned slice
+
+pub unsafe fn visit_pointer_fields<C, F, G>(
+    ctx: &mut C,
```
crusso, line 24: Drop `ctx`?

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -158,15 +158,33 @@ unsafe fn mark_stack<M: Memory>(mem: &mut M, heap_base: u32) {
...
+            let field_value = *field_addr;
+            mark_object(mem, field_value);
+
+            // Thread if backwards or self pointer
+            if field_value.get_ptr() <= obj as usize {
+                thread(field_addr);
+            }
+        },
+        |mem, slice_start, arr| {
+            const SLICE_INCREMENT: u32 = 127;
```
crusso, line 176: Move this magic constant to more prominent top of file?

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -158,15 +158,33 @@ unsafe fn mark_stack<M: Memory>(mem: &mut M, heap_base: u32) {
...
+            debug_assert!(SLICE_INCREMENT >= TAG_ARRAY_SLICE_LOW_LIMIT);
```
crusso, line 177: Could that be a static assert at top of file?

crusso (approved):
Would be nice if there was some way to test this. Any thoughts?

ggreif:
> Would be nice if there was some way to test this. Any thoughts?

The random heap tests in `rts/motoko-rts-tests` is testing this, I planted asserts into the new code and it always hit.

rts/motoko-rts/src/gc/mark_compact.rs
ggreif: I could be, but we currently don't have static asserts in the RTS. I don't think it is super urgent.

rts/motoko-rts/src/gc/mark_compact.rs
ggreif: I tried to beat the borrow check, but it was relentless on me. Only a unified callback would do the trick, and that has other problems.

rts/motoko-rts/src/gc/mark_compact.rs
ggreif: Nobody else needs to know, I prefer it to live in a narrow scope.

rts/motoko-rts/src/types.rs
ggreif: 9970b8a70.

rts/motoko-rts/src/gc/mark_compact.rs
ggreif: This happens when I close over `mem` from two callbacks: 4db6013188985c37ba8546ef7970e01af9e4bc83.

PR #3232, opened 2022-05-11
feat: implement partial candid deserialization (DeserializeOptPrim) (to be merged with #3155)

crusso:
Implements partial deserialization (DeserializeOptPrim) of valid candid (returning null when valid candid not of expected type but trapping on invalid candid).

Completes @matthewhammer's #3155.

I notice the parser doesn't accept unparenthesised arguments. Is that deliberate or an omission. Perhaps it's actually what we want to avoid the ambiguity with 1-tuples.

The tests could use expanding, e.g. for record, arrays, variants, including some subtype failures that map to `null`. @matthewhammer I'd be grateful if you could extend the file candid.mo along those lines.

@nomeata, @ggreif  I wouldn't mind a quick check of my code in case I'm doing something odd.

NB: from_candid, as implemented, assumes the blob is valid candid and will trap, not return null, otherwise, i.e. null is returned if candid subtyping/coercion fails. That's all.


chenyan-dfinity:
> I notice the parser doesn't accept unparenthesised arguments. Is that deliberate or an omission. Perhaps it's actually what we want to avoid the ambiguity with 1-tuples.

You mean the Candid parser? That's intentional, because arguments are tuples: `<arg> ::= ( <annval>,* )`

crusso:
> > I notice the parser doesn't accept unparenthesised arguments. Is that deliberate or an omission. Perhaps it's actually what we want to avoid the ambiguity with 1-tuples.
> 
> You mean the Candid parser? That's intentional, because arguments are tuples: `<arg> ::= ( <annval>,* )`

No, I meant the Motoko parser. to_candid 1 is illegal, but to_candid(1) is ok, but I think that's actually sensible to avoid the ambiguity of to_candid () (is that the empty sequence, or singleton sequence with a unit tuple as first entry)?

matthewhammer:
> but I think that's actually sensible to avoid the ambiguity of to_candid () (is that the empty sequence, or singleton sequence with a unit tuple as first entry)?

Exactly.

matthewhammer (approved):

matthewhammer:
> The tests could use expanding, e.g. for record, arrays, variants, including some subtype failures that map to `null`. @matthewhammer I'd be grateful if you could extend the file candid.mo along those lines.

Doing this now in the main PR (#3155).  Thanks for the help!

PR #3239, opened 2022-05-16
nix: Build drun with rust 1.60

nomeata:
to match https://github.com/dfinity/ic/blob/master/rust-toolchain.toml


ggreif (approved):
Thanks!

PR #3262, opened 2022-05-23
Record typing test

matthewhammer:
Reproduces a bug in the lowering or other front-end phases of the compiler, where the initial type-check passes, but post lowering, the IR does not carry enough type information to check anymore (my best guess).

Thanks to @ninegua for reporting to the team Slack today, based on another dev's comment.

I'm not committing the new test outcome, because it's erroneous.

This message appears:
```
(unknown location): IR type error [M0000], subtype violation:
  {var myList : List<None>}
  {var myList : List<Nat>}
```

### DIFF
--- a/test/run/records.mo
+++ b/test/run/records.mo
@@ -23,3 +23,7 @@ assert(rec3.y == y);
 assert(rec3.z == 13);
 assert(rec3.a == 4);
 assert(rec3.b == 5);
+
+type List<X> = ?(X, List<X>);
+func nil<X>() : List<X> = null;
+let rec4 : { var myList : List<Nat> } = { var myList = nil() };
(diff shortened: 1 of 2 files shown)


test/run/records.mo
```diff
@@ -23,3 +23,7 @@ assert(rec3.y == y);
 assert(rec3.z == 13);
 assert(rec3.a == 4);
 assert(rec3.b == 5);
+
+type List<X> = ?(X, List<X>);
+func nil<X>() : List<X> = null;
+let rec4 : { var myList : List<Nat> } = { var myList = nil() };
```
nomeata, line 29: Probably better putting that regression tests into its own file (`issue3265.mo` maybe).

crusso:
Investigating...

test/run/records.mo
crusso: Done as part of #3268

crusso:
(Closing - fixed in #3268)

Issue #3078, opened 2022-01-28
`languageServer` support for explicit symbol `import`s


ggreif:
#3076 added explicit symbol `import`s to Motoko, but the `languageServer` still ignores them.

Blocks #2354.

kritzcreek:
Unfortunately I don't have an environment with which I can build the compiler at the moment, but I can point you at a couple of places that would need changing and what level of "sophistication" are possible. Also happy to jump on a call some time :)

1. This function needs updating: https://github.com/dfinity/motoko/blob/ecadc8dc6f9318799bbb0dd41591272095778cce/src/languageServer/source_file.ml#L78-L108. The way it's written is from before we had the _shudders_ `parse_module_header` function https://github.com/dfinity/motoko/blob/ecadc8dc6f9318799bbb0dd41591272095778cce/src/mo_frontend/parser.mly#L300
2. For go-to-definition or type-on-hover you can decide if you just want to resolve all unqualified names, or be smarter and try to respect scoping. But you can't rely on the full module being parseable! For example (with `|` representing the users cursor): 
    ```motoko
    import { def } "amodule";

    module {
      let def = 10;
      let x = de|f;
    }
    ```

    The "dumb but robust" thing would be to just resolve `def` according to the import list and have go-to-definition take the user to `amodule` even if that's wrong here. I'd say it's rare enough that users shadow imports that this is valuable even if it produces wrong results occasionally. Otherwise you need an error-recovering parser or for `go-to-definition` to only work on syntactically valid source files.

3. For completions you'll probably have to respect the new more complicated return from `parse_module_header` and handle the unqualified prefix cases accordingly. Right now unqualified names are _always_ completed with respect to the current module, but now you'll also have to consider the explicit imports.
4. If you want to get fancy: For completions we create import edits for the user. You could decide if you want the user to be able to complete the name `filter` even if their import list only says `import { map } from "mo:base/Array";` so far and add `filter` to the import list as an edit on completion.

ggreif:
@kritzcreek Thanks a lot for the detailed answer! Right now this is low-prio, but will want to tackle it as soon as the first user begins to complain! 😉

PR #3263, opened 2022-05-23
Improve support for navigating to definitions in external modules

rvanasa:
I rewrote most of `definition_handler` to account for edge cases related to finding definitions in external files. 

- It's now possible to jump to the definition of an import identifier (rather than having to click on a qualified member name). For example, `import M|ap "HashMap"` (where `|` is the cursor) jumps to the top of `HashMap.mo` instead of doing nothing (fixes dfinity/vscode-motoko#17). 

- The new implementation also has a more permissive default behavior for missing declarations. In cases where the module path is known but the qualified symbol is unknown, the language server attempts to direct the user to the relevant module. This is mostly intended as a fallback for situations where the index is missing an expected declaration, such as when the other file has a syntax error. 

Possible future improvements:

- Using the start/end position of the `module` or equivalent keyword (when possible) rather than the top of the file
- Providing the same functionality when the cursor is on the import string literal (and perhaps even the `import` keyword)
- Supporting object pattern-style imports as mentioned in #3078 (currently experimenting with this)

Passes all tests on my WSL setup and works as expected in VSCode with the [official Motoko extension](https://github.com/dfinity/vscode-motoko). 

(@matthewhammer)

### DIFF
--- a/src/languageServer/definition.ml
+++ b/src/languageServer/definition.ml
@@ -36,17 +36,42 @@ let definition_handler index position file_contents project_root file_path =
       Source_file.identifier_at_pos project_root file_path file_contents
         position
     in
-    let* path, region =
+    (* Find the relevant path for the declaration *)
+    let* path =
       match ident with
-      | Alias _ -> None
+      | Alias (_, path) -> Some path
       | Unresolved _ -> None
-      | Resolved resolved ->
-          let* module_ = DI.lookup_module project_root resolved.path index in
-          find_named resolved.ident module_
-      | Ident ident ->
-          let* uri = Lib.FilePath.relative_to project_root file_path in
-          let* module_ = DI.lookup_module project_root uri index in
-          find_named ident module_
+      | Resolved { path; _ } -> Some path
+      | Ident _ -> Lib.FilePath.relative_to project_root file_path
+    in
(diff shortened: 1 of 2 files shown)


matthewhammer:
Thanks @rvanasa!  I'm adding @crusso and @ggreif as reviewers.

matthewhammer:
cc @kritzcreek

kritzcreek (commented):
Haven't looked all that closely but the changes seem reasonable. Either way I'd expect a change like this to also include a test.

crusso:
Hi Ryan, thanks for the PR(s)!

 I'm in the middle of a family emergency so won't get a chance to review today. Hopefully someone else can oblige or I'll get to it in the next few days

crusso:
(merging with master to kick off another build)

crusso:
I'm not sure why this hasn't merged. @rvanasa are you not allowed to merge, even with approval, or were you holding back for #3282 instead?

ggreif:
@rvanasa if you cannot merge, try adding the `automerge-squash` label. If that also fails then we are happy to push the button for you!

rvanasa:
@ggreif it looks like I would need write permissions to add a label to the PR. Thanks for merging when you get the chance!

crusso:
Fab! Thanks again!

PR #3172, opened 2022-03-28
experiment: markdown docs

crusso:
- add manual makefile target to produce markdown from adoc, generated into folder docs/md.

Just an experiment to find the right incantation of asciidoctor and fixes to adoc to make it work.

PR #3269, opened 2022-05-27
Bump nixpkgs to 22.05 release branch

nomeata:


nomeata:
First issue: https://github.com/WebAssembly/spec/issues/1480

nomeata:
Also affected by https://github.com/NixOS/nixpkgs/issues/155113

nomeata:
We are moving away from asciidoc and antora anyways, right? So I can just dump the antora-related CI jobs, instead of investing time into getting them to work?

nomeata:
We are moving away from asciidoc and antora anyways, right, @crusso? So I can just dump the antora-related CI jobs, instead of investing time into getting them to work?

nomeata:
Whoohoo, that upgrade seems to go through fine :-)

crusso:
> We are moving away from asciidoc and antora anyways, right, @crusso? So I can just dump the antora-related CI jobs, instead of investing time into getting them to work?

Eventually, but I think I'll still want asciidoctor (not asciidoc) working for a while (thought asciidoctor isn't even used by CI I expect)

nomeata:
Wouldn't we just have .md files once #3172 merged?

crusso:
> Wouldn't we just have .md files once #3172 merged?

Well, no. I like the fact that adoc lets you include files (and regions of) files that we can check separately. Until we have a workaround, I'd prefer to generate the markdown off the adoc.

Next question: Is there a workaround? I guess we could write a script to extract the examples somehow, but yuck.

crusso:
https://github.com/ocaml/omd might let us add a .md mode to moc for checking code embedded in markdown files, but it doesn't handle gfm tables... sigh

nomeata:
I see, so we'll keep the adocs, and generate md for the doc site?

But even then we don't need antora, just any adoc processor, in our CI, right?

crusso:
> I see, so we'll keep the adocs, and generate md for the doc site?
> 
> But even then we don't need antora, just any adoc processor, in our CI, right?

I think that's correct. Indeed, for the conversion I need to use asciidoctor, not asciidoc. Sorry for the delay.

nomeata:
Depends on #3350

default.nix
```diff
@@ -313,7 +313,7 @@ rec {
     };
 
     testDerivationDeps =
-      (with nixpkgs; [ wabt bash perl getconf moreutils nodejs-16_x sources.esm ]) ++
+      (with nixpkgs; [ wabt bash perl getconf moreutils nodejs-16_x ]) ++ # TODO: sources.esm
```
nomeata, line 316: Unclear why we needed this, but it seems we don’t need it any more.
```suggestion
      (with nixpkgs; [ wabt bash perl getconf moreutils nodejs-16_x ]) ++
```

nomeata:
Looks good, can go in, I think

ggreif (approved):
Great!

ggreif:
Looks like `nix/155128.patch` got added, but never applied by `nix`. @nomeata was  this intentional?

nomeata:
Hmm, maybe it wasn't needed after all. Feel free to remove.

PR #3268, opened 2022-05-27
bugfix: fix issue #3265 (typing of  annotated, recursive records)

crusso:
Fixes #3265 (adding repro from #3262 as regression test)

PR #3277, opened 2022-05-31
bugfix: fix bug introduced in #3268 (lowering of recursive records)

crusso:
Fix bug that snuck in with over-eager refactoring in #3268


nomeata (approved):
Rubber stamp

PR #3278, opened 2022-06-01
feat: refactor `stableVarInfo` as prim returning an obviously pure (and more efficient) query

crusso:
The current implementation of `stableVarInfo` is an asynchronous function that:
* has (first-order) type `(actor {}) -> async { size : Nat64 }` 
* is a local, async *update* function
* takes a redundant actor reference,
* takes two hops through the scheduler.
* doesn't reflect that it's essentially pure.

This refactoring, instead, provides a *higher-order* primitive that, when applied, returns an obviously pure *query*:
* has (higher-order) type `() -> (shared query () -> { size : Nat64 })` 
* is a local synchronous function, returning a *query*.
* avoiding the extraneous scheduling,
* simplifying the implementation,
* avoiding the need to pass the actor reference.
 
(The prim is higher-order only because we don't allow primitive values, just local functions. Otherwise I would just provide access to a well-known query value `self.__motoko_stable_var_info`.)

I think we can still do this refactoring since we haven't exposed this facility in motoko-base.

@ggreif whaddya think? @nomeata is this just ugly?

(scratches a perhaps somewhat anal itch from https://github.com/dfinity/motoko-base/pull/365)

### DIFF
--- a/src/prelude/prim.mo
+++ b/src/prelude/prim.mo
@@ -262,7 +262,7 @@ func principalOfActor(act : actor {}) : Principal = (prim "cast" : (actor {}) ->
 // Untyped dynamic actor creation from blobs
 let createActor : (wasm : Blob, argument : Blob) -> async Principal = @create_actor_helper;
 
-// An async function for querying stable variable statistics
+// Returns a query that computes the current actor's stable variable statistics (for now, the current size, in bytes, of serialized stable variable data).
 func stableVarInfo() : shared query () -> async {size : Nat64} =
   (prim "stableVarInfo" : () -> (shared query () -> async {size : Nat64})) () ;
 
--- a/src/prelude/prim.mo
+++ b/src/prelude/prim.mo
@@ -363,3 +363,5 @@ func stableMemoryStoreBlob(offset : Nat64, val :  Blob) : () =
   (prim "stableMemoryStoreBlob" : (Nat64, Blob) -> ()) (offset, val);
 
 let call_raw = @call_raw;
+
+let stableVarInfo = @stableVarInfo;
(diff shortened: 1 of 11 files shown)


src/prelude/prim.mo
```diff
@@ -263,7 +263,8 @@ func principalOfActor(act : actor {}) : Principal = (prim "cast" : (actor {}) ->
 let createActor : (wasm : Blob, argument : Blob) -> async Principal = @create_actor_helper;
 
 // An async function for querying stable variable statistics
```
crusso, line 265:
```suggestion
// Returns a query that computes the current actor's stable variable statistics (for now, the current size, in bytes, of serialized stable variable data).
```

src/prelude/prim.mo
```diff
@@ -262,8 +262,9 @@ func principalOfActor(act : actor {}) : Principal = (prim "cast" : (actor {}) ->
 // Untyped dynamic actor creation from blobs
 let createActor : (wasm : Blob, argument : Blob) -> async Principal = @create_actor_helper;
 
-// An async function for querying stable variable statistics
-let stableVarInfo = @stable_var_info;
+// Returns a query that computes the current actor's stable variable statistics (for now, the current size, in bytes, of serialized stable variable data).
+func stableVarQuery() : shared query () -> async {size : Nat64} =
+  (prim "stableVarQuery" : () -> (shared query () -> async {size : Nat64})) () ;
```
ggreif, line 267: Oh, I see. The `prim` returns a query function. Which is a pair of a `(principal,〈method-name〉)`.

ggreif (approved):
Looks reasonable, the size increase is marginal, and can be explained by the slightly higher administrative effort to construct the `{ size = ... }` record.

src/prelude/prim.mo
crusso: Yep. I may actually be able to get rid of the clunky `() ->...` delay if I shove this into internals and reference it. I don't think we do a definedness check there ;->

ggreif:
> One slight worry I have is that this allocated in the helper, after we've mutated the heap, but I think it's probably ok since we never collect. Seems to work ;->

But `Blob.lit` will allocate into the static heap, so where is the problem?

PR #3282, opened 2022-06-02
Add basic language server support for explicit symbol imports

### LINKED ISSUE #3078: `languageServer` support for explicit symbol `import`s

rvanasa:
This PR adds straightforward "go to definition" support for Motoko's pattern-style import syntax. 

For example, `import { v|als } "mo:base/Array"` and `v|als(x)` (where `|` represents the cursor) will correctly navigate to the corresponding `Array.vals` declaration. 

As with existing language server functionality, these changes do not yet account for variable shadowing. However, the updated `parse_module_header` function already parses import patterns such as (`import {alias = field} "..."`) based on the proposed generalized import syntax (#3076). 

Please let me know if additional explanation would be helpful for any of these changes. I also included four test cases to cover the newly expected behavior of `definition_handler` and `parse_module_header`. 

Tested in VSCode on WSL using the official Motoko extension.

Fixes #3078
Supersedes #3263

CC @kritzcreek, @crusso, @matthewhammer

### DIFF
--- a/test/lsp-int-test-project/definitions.mo
+++ b/test/lsp-int-test-project/definitions.mo
@@ -1,11 +1,11 @@
 import MyDependency "mo:mydep/lib";
 import List "lib/list";
-import { last } "lib/list";
+import { nil; last = lastElement } "lib/list";
 
 module {
     public func myFunc() {
         let myClass = MyDependency.MyClass();
         let myNil = List.nil<Nat>();
-        assert last(myNil) == null;
+        assert lastElement(myNil) == null;
     }
 }
--- a/test/lsp-int-test-project/definitions.mo
+++ b/test/lsp-int-test-project/definitions.mo
@@ -1,9 +1,11 @@
 import MyDependency "mo:mydep/lib";
 import List "lib/list";
+import { last } "lib/list";
 
 module {
     public func myFunc() {
         let myClass = MyDependency.MyClass();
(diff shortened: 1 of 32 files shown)


crusso:
Thanks @rvanasa these contributions are much appreciated. I won't get  a chance to review today but perhaps someone else will.

test/lsp-int-test-project/definitions.mo
```diff
@@ -1,9 +1,11 @@
 import MyDependency "mo:mydep/lib";
 import List "lib/list";
+import { last } "lib/list";
```
ggreif, line 3: I suggest to also add a _renaming on import_ declaration.

ggreif (approved):
LGTM. Thanks!

test/lsp-int-test-project/definitions.mo
rvanasa: Thanks for taking a look. I added an LSP test based on your suggestion (slightly modified since there seems to be a compiler bug resulting in `type error [M0121], duplicate field last in object pattern`). I created an issue #3290 and will see if there's an elegant way to fix this in a separate PR.

ggreif: Nice find!

PR #3283, opened 2022-06-02
feat: refactor `stableVarQuery` as prim returning an obviously pure (and more efficient) query

crusso:
(cherry-picked commits from #3278 before things went south)

The current implementation of `stableVarInfo` is an asynchronous function that:
* has (first-order) type `(actor {}) -> async { size : Nat64 }` 
* is a local, async *update* function
* takes a redundant actor reference,
* takes two hops through the scheduler.
* doesn't reflect that it's essentially pure.

This refactoring, instead, provides a *higher-order* primitive that, when applied, returns an obviously pure *query*:
* has (higher-order) type `() -> (shared query () -> { size : Nat64 })` 
* is a local synchronous function, returning a *query*.
* avoiding the extraneous scheduling,
* simplifying the implementation,
* avoiding the need to pass the actor reference.
 
(The prim is higher-order only because we don't allow primitive values, just local functions. Otherwise I would just provide access to a well-known query value `self.__motoko_stable_var_info`.)

I think we can still do this refactoring since we haven't exposed this facility in motoko-base.

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -8797,7 +8797,7 @@ and compile_prim_invocation (env : E.t) ae p es at =
     compile_exp_as env ae SR.UnboxedWord64 e ^^
     StableMem.logical_grow env
 
-  | OtherPrim ("stableVarInfo"), [] ->
+  | OtherPrim ("stableVarQuery"), [] ->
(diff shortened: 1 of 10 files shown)


ggreif (approved):
LGTM!

PR #3286, opened 2022-06-03
Update doc for interpreter support for markdown

chenyan-dfinity:
Tested in https://github.com/wackyleo459/test-motoko-interpreter/pull/1

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -26,65 +26,15 @@ CI pushes these docs for latest master to
 
 # Support interpreter in documentation
 
-## Page attributes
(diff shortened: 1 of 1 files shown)


crusso (approved):
This is great. Can we keep the old doc around too, until we've done the migration?

I need to figure out how to get adoc to translate or preserve the full language attribute. I think they might not survive translation at the moment.

chenyan-dfinity:
Yeah, not easy to translate. We could manually add back the attributes. There are not many of them.

PR #3289, opened 2022-06-03
Claudio/performance counter bump drun

crusso:


nix/drun.nix
```diff
@@ -15,7 +15,8 @@ pkgs:
       # installed. You will normally not be bothered to perform
       # the command therein manually.
 
-      cargoSha256 = "sha256-Oi7hKhif0CpnovGNX+/hpKzfMDuSpoWwtmH0WbnrLqg=";
+      cargoSha256 = "sha256-EbYM5DxJILstvNjQS4Jr9RvCxsT71XfcZ4lWzBiJaTU=";
+      # "sha256-Oi7hKhif0CpnovGNX+/hpKzfMDuSpoWwtmH0WbnrLqg=";
```
crusso, line 19:
```suggestion
```

src/codegen/compile.ml
```diff
@@ -3671,6 +3672,14 @@ module IC = struct
         edesc = nr (FuncExport (nr (E.built_in env "print_ptr")))
       })
 
+
+  let performance_counter env =
+    match E.mode env with
+    | Flags.ICMode | Flags.RefMode ->
```
ggreif, line 3678:
```suggestion
    | Flags.(ICMode | RefMode) ->
```

ggreif (approved):
Cool!

src/codegen/compile.ml
crusso: We use this style elsewhere...

ggreif: My conception is that it makes the two relevant pieces  of information appear nearer and is thus better digestible. But I am not radical enough about it to change all occurrences.

PR #3298, opened 2022-06-07
feat: three-way `Text` comparison

### LINKED ISSUE #3150: Perf: Motoko Base Text.compare needs an efficient RTS implementation
Text.compare currently requires two relational comparisons, but there is a more efficient implementation based on RTS.text_compare that we should expose as a Prim and use.

ggreif:
This adds a primitive `textCompare` of type
``` Motoko
(Text, Text) -> Int8
```
representing the ordering relation on `Text`. -1 means the first text is smaller, 0 means they are equal, and 1 otherwise.
This is basically a backend wrapper around Rust `text_compare`.

Partly fixes #3150.

### DIFF
--- a/src/mo_values/prim.ml
+++ b/src/mo_values/prim.ml
@@ -218,7 +218,9 @@ let prim =
     k (Int (Nat.of_int (List.length (Wasm.Utf8.decode (Value.as_text v)))))
   | "text_compare" -> fun _ v k ->
     (match Value.as_tup v with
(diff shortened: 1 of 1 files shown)


crusso:
Good man - I guess all you need to do is convert to Order.t in base/Text.mo, right?

nomeata (approved):
LGTM

crusso:
Me too.

crusso (approved):
Good stuff.

PR #3301, opened 2022-06-08
Abolish `UnboxedBool` stack representation

### LINKED ISSUE #2906: Can we also get rid of shifted (when `Vanilla`) `Bool`?
Since the first page in Wasm heap is reserved as stack/scratch area for RTS use, the number `1 : U32` won't be traced by the GC _ever_. Can we move back and say that `true` is `1 : U32` in `SR.Vanilla` too?

(Which similarly would allow us to also represent `Nat8` unshifted, but let's not open that can of worms just yet.)

ggreif:
Abolish `UnboxedBool` in favour of `Vanilla`. We now always use {`0`, `1`} for `Bool` values, since they never get dereferenced, and GC avoids visiting everything that is not in the dynamic heap (e.g. small constants, pointers into the first page etc.)

Resolves #2906.

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -1141,15 +1141,14 @@ module BitTagged = struct
      signed numbers as well.
 
      Boolean false is a non-pointer by construction.
-     Boolean true (1) needs to be shifted to be a non-pointer.
-     No unshifting necessary before a branch.
+     Boolean true (1) needs not be shifted as GC will not consider it.
 
      Summary:
 
        0b…11: A pointer
        0b…x0: A shifted scalar
        0b000: `false`
-       0b010: `true`
+       0b001: `true`
 
      Note that {Nat,Int}{8,16} do not need to be explicitly bit-tagged:
      The bytes are stored in the _most_ significant byte(s) of the `i32`,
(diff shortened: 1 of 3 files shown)


nomeata (commented):
This looks reasonable if we have no spot where we assume that any non-zero (i.e. not just 1) value is a true `UnboxedBool` value. Not sure if we took that short cut anywhere.

Probably not because it wouldn't be correct to shift left then.

ggreif:
@nomeata I am confident that we didn't use non-0/1 values in `UnboxedBool`, since the only places these originate from are the `lit` and primitive predicates. We don't cast e.g. `Nat16` to `Bool` ever.

src/codegen/compile.ml
```diff
@@ -172,7 +171,7 @@ module SR = struct
 
   let unit = UnboxedTuple 0
 
-  let bool = UnboxedBool
+  let bool = Vanilla
```
ggreif, line 174: Can we zap this binding now?

nomeata: Yes, I'd say so

nomeata (approved):
Nice cleanup

src/codegen/compile.ml
ggreif: I'll leave it in, as it is helpful documentation that this is the {`0`, `1`} subset of `Vanilla`.

test/run/ok/array.tc.ok
```diff
@@ -0,0 +1,3 @@
+array.mo:89.27-89.64: warning [M0074], this array has type
+  [Any]
+because elements have inconsistent types
```
matthewhammer, line 3: Interesting that the compiler still warns about it using this type `[Any]` even though the program has an explicit annotation saying that "it knows" that already.

ggreif: Yeah, that's why I filed #3302!

PR #3305, opened 2022-06-10
Releasing 0.6.29

ggreif:


ggreif:
@nomeata I think we need some guard to disallow pushes to the `release` brach that is not also on `master`. Currently we have a  divergence, and as such I cannot cut a release. (I asked for admin help.) But maybe a commit hook can be crafted that would do the job?

PR #3317, opened 2022-06-15
WIP: skew the heap base while `mark_compact`

ggreif:


luc-blaeser:
Probably obsolete, as we focus on incremental GC and EOP in future.

PR #3334, opened 2022-06-27
always inline worker into wrapper when compiling for IC

ggreif:


ggreif:
It is not clear why this happens: `+0.0%`.

ggreif:
This is  a pessimisation.

PR #3335, opened 2022-06-28
Quicker test for `i32` taggability and more tagging opportunities

ggreif:
see https://github.com/dfinity/motoko/pull/3330#discussion_r907673478

`"can_tag_i32"` is now one instruction shorter
`"box_i32"` is now keeping more values tagged and less go to the heap

TODO:
- write a hashing-heavy benchmark

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -1205,14 +1205,14 @@ module BitTagged = struct
     compile_shrS_const 1l ^^
     G.i (Convert (Wasm.Values.I64 I64Op.ExtendSI32))
 
-  (* 32 bit numbers, dynamic *)
+  (* 32 bit numbers, dynamic, w.r.t `Int` *)
 
   let if_can_tag_i32 env retty is1 is2 =
     Func.share_code1 env "can_tag_i32" ("x", I32Type) [I32Type] (fun env get_x ->
-      (* checks that all but the low 30 bits are either all 0 or all 1 *)
-      get_x ^^ compile_shl_const 1l ^^
-      get_x ^^ G.i (Binary (Wasm.Values.I32 I32Op.Xor)) ^^
-      compile_shrU_const 31l
+      (* checks that all but the low 30 bits are both either 0 or 1 *)
+      get_x ^^ compile_shrU_const 30l ^^
+      G.i (Unary (Wasm.Values.I32 I32Op.Popcnt)) ^^
+      G.i (Unary (Wasm.Values.I32 I32Op.Ctz))
     ) ^^
     E.if_ env retty is2 is1 (* NB: swapped branches *)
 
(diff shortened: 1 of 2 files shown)


test/perf/heap-32-bench.mo
```diff
@@ -0,0 +1,48 @@
+// allocate 2 big arrays with a 10^5 entries each, and populate them
+// with Int32s and Nat32s (somewhat randomly)
+import Prim "mo:⛔";
```
ggreif, line 3: I'll break out this benchmark to another PR, so it gets  merged ahead of this one. Then we'll see the diffs for more.

ggreif: Still, comments are warmly welcomed!

src/codegen/compile.ml
```diff
@@ -1205,16 +1205,16 @@ module BitTagged = struct
...
-      get_x ^^ compile_shl_const 1l ^^
-      get_x ^^ G.i (Binary (Wasm.Values.I32 I32Op.Xor)) ^^
-      compile_shrU_const 31l
+      (* checks that all but the low 30 bits are both either 0 or 1 *)
+      get_x ^^ compile_shrU_const 30l ^^
+      G.i (Unary (Wasm.Values.I32 I32Op.Popcnt)) ^^
+      G.i (Unary (Wasm.Values.I32 I32Op.Ctz))
     ) ^^
-    E.if_ env retty is2 is1 (* NB: swapped branches *)
+    E.if_ env retty is1 is2
```
nomeata, line 1217: Rename to cannot_tag then?

ggreif: Nope, the new algorithm just swaps the cases that `if` scrutinises. The semantics remains.

test/perf/heap-32-bench.mo
ggreif: I see, I didn't explicitly register that. Would you suggest a new category (`test/benchmarks`) or scrapping this test altogether?

nomeata: Separate microbenchmarks would be fine. What I want to avoid is that we think some random change had a significant real world perf impact, one way or another, when it “just” triggered a microbenchmark.

As for more reporting: the data is there, the report can be extended to contain more data. Maybe hidden in `<details>`?

src/codegen/compile.ml
nomeata: I mean the `share_code` name, not the Ocaml function name

test/perf/heap-32-bench.mo
ggreif: I would also spare the user from seeing a comment at all coming from benchmarks, unless there is an actual cycle/size change in one of them.

nomeata (approved):

ggreif:
Above commit documents some of the tagging success, from 100_000 `Int32`s previously 75_000 went to  heap, now only 50_000. The difference is 25_000. Each heap object occupies 8 bytes, so the win is about 200_000, as shown by the diff.

PR #3338, opened 2022-06-30
Benchmarks (beginnings of the infrastructure)

ggreif:
See https://github.com/dfinity/motoko/pull/3335#discussion_r910747576 and following.

This adds a `tests.bench` derivation and runs the microbenchmarks from `test/bench` directory.
For now there is only one benchmark present.

TODO:
- mine `stats.csv` for deltas (will be done in a follow-up PR as the derivation needs to be on `master` so that `from` will contain it)


test/Makefile
```diff
@@ -1,5 +1,5 @@
 QUICK_DIRS = fail idl ld mo-idl repl run
-OTHER_DIRS = perf run-deser run-drun trap
+OTHER_DIRS = perf bench run-deser run-drun trap
```
crusso, line 2: maintain in alphabetical order?

crusso (approved):
Apart from nit. I did not review the test in detail. Did you want me to?

nomeata (approved):

PR #3350, opened 2022-07-13
doc: delete adoc/antora sources; adapt CI

crusso:
* delete adoc/antora files
* adjust CI


nomeata (approved):
Looks good. Once on `master` we can check if https://dfinity.github.io/motoko/ still has it. I think the prose in CI.md isn’t affected by this.

PR #3353, opened 2022-07-14
doc: add basic docusaurus project to enable (live) preview of doc

crusso:
* adds a basic docusaurus project cobbled together from the default docusaurus projects and bits and pieces of the portal need to support the live interpreter and remark-code-imports.
* in nix-shell, under doc, do `make preview` to build the js moc interpreter and doc site, opening it in the browser.
* preview is live, so doc changes cause the browser to refresh.

We are now pretty much at parity with the pre-markdown age.

I still get some vulnerability warning about the npm packages I'm using. @chenyan-dfinity would you take a look and try to fix them? My js/node is crap.

Would also be good if someone tried this on darwin to see it works there.


nomeata:
Is this pushed to https://dfinity.github.io/motoko/ already?

crusso:
> Is this pushed to https://dfinity.github.io/motoko/ already?

No, I'm just using it for local preview. That's an obvious next step, but I may need improve the navigation a bit. Or we just live with it. Also, I'm getting lots of warnings about vulnerabilities from npm so maybe we should fix those first.

crusso:
> Is this pushed to https://dfinity.github.io/motoko/ already?

Looks like I screwed that up in my previous PR. I think this #3355 fixes it.

chenyan-dfinity:
npm vulnerabilities look fine to me, since it's only a local preview. But some other questions:
* What's the diff between our own docusaurus and the official one? Do we want to keep them in sync?
* We seems to check in moc-interpreter.js manually. Is that needed?

chenyan-dfinity (approved):
Looks great! I changed the makefile to use the generated interpreter instead of checking in the code.

crusso:
> npm vulnerabilities look fine to me, since it's only a local preview. But some other questions:
> 

> * We seem to check in moc-interpreter.js manually. Is that needed?

Weird, I didn't want to do that and thought I had symlinked it, but copying `cp -f` in Makefile works too. (I added the `-f` so a second `make preview` doesn't fail with permission denied)

crusso:
@chenyan-dfinity can you take a quick look at ryans commit 

https://github.com/dfinity/motoko/pull/3353/commits/09e5ab1f2d3c5e8333f47b95a5a97e78a07461ad
https://github.com/dfinity/motoko/pull/3353/commits/9d811eed9a67ca2d006d7a8a543036bb542d2400

If good, you may want to patch the portal code too.

chenyan-dfinity:
Symlink doesn't work for docusaurus, we can only `cp` there.

> If appropriate, you may want to patch the portal code too.

Yep, already done: https://github.com/dfinity/portal/pull/320

> I guess it would be better if this code used the base library referenced in nix/sources.json

Should be doable. `sources.json` is pinned at a commit not a tag, so we need another function to load from github commits instead of CDN.

PR #3364, opened 2022-07-19
Adapt to Wasm 0.2 library

### LINKED ISSUE #3363: use `ocamlPackages.wasm` v2.0.0
https://github.com/NixOS/nixpkgs/commit/713b7b04045268f155dc27b3a5e42ec3fef09104

I think this will be available once we bump `nixpkgs`. May need OCaml-4.13, though.

ggreif:
DON'T LOOK YET, THIS IS A MESS!
----------

Fixes #3363.


src/wasm-exts/utf8.ml
```diff
@@ -0,0 +1,45 @@
+type codepoint = int
```
ggreif, line 1: maybe we can remove the other Utf8 decoder and use this only? Ditto the other `.mli` file.

crusso: IIRC correctly, this version isn't tail-recursive and can blow up - is the other one written to be tail-recursive perhaps?

ggreif: Turns out, this file is not needed, `encoder.ml` references it as `Wasm.utf8`. But it uses it for (short?) identifiers only.

src/wasm-exts/free.ml
```diff
@@ -0,0 +1,147 @@
...
+
+let start (s : start) = funcs (var s.it.sfunc)
+
+let module_ (m : module_) =
+  list type_ m.it.types ++
+  list global m.it.globals ++
+  list table m.it.tables ++
+  list memory m.it.memories ++
+  list func m.it.funcs ++
+    (*opt start m.it.start ++*)
```
ggreif, line 143: This gives me a type error. Needs fix.

nomeata:
Can we maybe first bump to the new wasm version, and then ocaml and nixpkgs separately? Or is that too tricky?

luc-blaeser:

PR #3367, opened 2022-07-20
alt syntax for record extension

crusso:
This PR switches the order of fields with bases and adds the new `with` keyword to separate them:
``` Motoko
// before
{ f1 = v1; ... in base1 and base2 and ... }
// after
{ base1 and base2 and ... with f1 = v1; ...; }
```
---------

variant of #3084 with alternative syntax

note this only changes the syntax, not evaluation order, which would need to be adjusted in both the interpreter and desugaring to evaluate bases before fields. The order of the arguments to ObjE should also be swapped.

@ggreif whaddya think?

PS. I may not have updated all tests...

### DIFF
--- a/doc/md/examples/grammar.txt
+++ b/doc/md/examples/grammar.txt
@@ -141,7 +141,7 @@
 <exp_obj> ::= 
     '{' <list(<exp_field>, ';')> '}'
     '{' <exp_post> 'and' <exp_post> ('and' <exp_post>)* '}'
-    '{' <exp_field> (';' <exp_field>)* 'in' <exp_post> ('and' <exp_post>)* '}'
+    '{' <exp_post> ('and' <exp_post>)* '|' <list(<exp_field>, ';')> '}'
 
 <exp_plain> ::= 
     <lit>
--- a/doc/md/examples/grammar.txt
+++ b/doc/md/examples/grammar.txt
@@ -141,7 +141,7 @@
 <exp_obj> ::= 
     '{' <list(<exp_field>, ';')> '}'
     '{' <exp_post> 'and' <exp_post> ('and' <exp_post>)* '}'
-    '{' <exp_post> ('and' <exp_post>)* '|' <list(<exp_field>, ';')> '}'
+    '{' <exp_post> ('and' <exp_post>)* WITH <list1(<exp_field>, ';')> '}'
 
 <exp_plain> ::= 
     <lit>
--- a/src/mo_frontend/printers.ml
+++ b/src/mo_frontend/printers.ml
@@ -189,7 +189,6 @@ let string_of_symbol = function
   | X (N N_seplist_dec_field_semicolon_) -> "seplist(<dec_field>,<semicolon>)"
   | X (N N_seplist_exp_ob__COMMA_) -> "seplist(<exp(ob)>,,)"
   | X (N N_seplist_exp_field_semicolon_) -> "seplist(<exp_field>,<semicolon>)"
-  | X (N N_separated_nonempty_list_semicolon_exp_field_) -> "seplist+(<exp_field>,<semicolon>)"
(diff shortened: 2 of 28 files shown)


test/repl/object-ext.sh
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 # Tests that fields completely override base fields
 moc -i <<__END__
-{ x = -25 in { x = 42 : Nat } };
+{ { x = 42 : Nat } | x = -25 };
 
 module X {public let a = 7};
-{ y = -25 in { y = 3; z = "H" } and X };
+{ { y = 3; z = "H" } | y = -25 };
```
crusso, line 7:
```suggestion
{ { y = 3; z = "H" } and X | y = -25 };
```

ggreif (commented):
The evaluation order of "bases before overwritten fields" is already given, to wit:
https://github.com/dfinity/motoko/pull/3084/files#diff-d77fc442c1bac05b23e2365d94307c741c01b841e870e6edadfd224bcfa36f60R14

doc/md/examples/grammar.txt
```diff
@@ -141,7 +141,7 @@
 <exp_obj> ::= 
     '{' <list(<exp_field>, ';')> '}'
     '{' <exp_post> 'and' <exp_post> ('and' <exp_post>)* '}'
-    '{' <exp_field> (';' <exp_field>)* 'in' <exp_post> ('and' <exp_post>)* '}'
+    '{' <exp_post> ('and' <exp_post>)* '|' <list(<exp_field>, ';')> '}'
```
matthewhammer, line 144: Compared to the keyword `in`, using the vertical bar appears more natural to me, especially if I read it as "such that", or "with".

But then  I wonder, would the word `with` be even better?

Currently, the role of `|` in Motoko is limited, but it still does convey something about "or" to me when I read it as a word, which is not appropriate here.

src/mo_frontend/printers.ml
```diff
@@ -189,7 +189,6 @@ let string_of_symbol = function
   | X (N N_seplist_dec_field_semicolon_) -> "seplist(<dec_field>,<semicolon>)"
   | X (N N_seplist_exp_ob__COMMA_) -> "seplist(<exp(ob)>,,)"
   | X (N N_seplist_exp_field_semicolon_) -> "seplist(<exp_field>,<semicolon>)"
-  | X (N N_separated_nonempty_list_semicolon_exp_field_) -> "seplist+(<exp_field>,<semicolon>)"
```
ggreif, line 192: I think this will need a corresponding new clause. It only gets triggered when the parser wants to emit a syntax hint.

doc/md/examples/grammar.txt
rossberg: Agreed with @matthewhammer. :)

ggreif (approved):
This can go into the big one

src/mo_frontend/printers.ml
crusso: I didn't understand this comment. Is it resolved?

ggreif: Nope, you can crash the interpreter with a syntax error, when it tries to report it.

ggreif: I'll try to fix now.

ggreif: Looks like there is already a clause (you have reused one), so no problem:
``` Motoko
> { a and a with true };
stdin:1.16-1.20: syntax error [M0001], unexpected token 'true', expected one of token or <phrase> sequence:
  seplist1(<exp_field>,<semicolon>) }
```

ggreif: You added it in this PR, actually. Cool, resolved!

PR #3378, opened 2022-07-30
Added Copyrights year and updated with MD format

Kellsonphilips:
Added link and applied the standard markdown formatting


crusso:
The information in this PR is actually held in a separate file

https://github.com/dfinity/motoko/blob/master/NOTICE

I'm not sure if we should change the original or not.  @dfx-json do you know?

PR #3381, opened 2022-08-02
feat: record instruction costs of mutator and collector

crusso:
Instrument calls to the collector to record IC instruction count just before and after collection, reporting the first number as instructions due to mutator (computation) and the difference as the instructions due to GC (may be low when scheduled selectively).

Exposes two new prims:
* rts_mutator_instructions: () -> Nat
* rts_collector_instructions : () -> Nat

Both return the stats for the *previous* messages.

(NB. We can't obtain these numbers using the user-level performance_counter since that always excludes GC times).

Intended for internal benchmarking of GC.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,7 +7,7 @@
       Prim.rts_mutator_instructions : () -> Nat
       Prim.rts_collector_instructions : () -> Nat
     ```
-    to report appoximate IC instruction costs of last message
+    to report approximate IC instruction costs of last message
     due to mutation (computation) and collection (GC), respectively (#3381).
 
 ## 0.6.29 (2022-06-10)
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,15 @@
 # Motoko compiler changelog
 
+* motoko (`moc`)
+
+  * add primitives
+    ```motoko
+      Prim.rts_mutator_instructions : () -> Nat
+      Prim.rts_collector_instructions : () -> Nat
+    ```
+    to report appoximate IC instruction costs of last message
+    due to mutation (computation) and collection (GC), respectively (#3381).
+
 ## 0.6.29 (2022-06-10)
 
 * motoko (`moc`)
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -937,6 +937,50 @@ module RTS = struct
 
 end (* RTS *)
 
+module GC = struct
+  (* Record mutator/gc instructions counts *)
+
+  let instruction_counter env =
+    compile_unboxed_zero ^^
+    E.call_import env "ic0" "performance_counter"
+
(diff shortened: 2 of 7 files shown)


test/run-drun/rts_gc_instructions.mo
```diff
@@ -0,0 +1,38 @@
...
+   }
+};
+
+//CALL ingress go "DIDL\x00\x00"
+
+// no point running these in the interpreter
+//SKIP run
+//SKIP run-low
+//SKIP run-ir
+//SKIP ic-ref-run
```
crusso, line 38:
```suggestion
//SKIP ic-ref-run

```

nomeata:
So these numbers are only per message execution, not accumulated over multiple message executions?

src/codegen/compile.ml
```diff
@@ -937,6 +937,50 @@ module RTS = struct
...
+  let record_mutator_instructions env =
+    match E.mode env with
+    | Flags.ICMode | Flags.RefMode  ->
+      instruction_counter env ^^
+      set_mutator_instructions env
+    | _ -> G.nop
+
+  let record_collector_instructions env =
+    match E.mode env with
+    | (Flags.ICMode | Flags.RefMode)  ->
```
ggreif, line 970:
```suggestion
    | Flags.(ICMode | RefMode)  ->
```
ditto line 963

ggreif (approved):
Neat! I wonder, though, if we can further parametrise `Flags` and create this extra overhead only when the user asks for it.

ggreif:
> So these numbers are only per message execution, not accumulated over multiple message executions?

No, the performance counters begin with 0 at each message.

crusso:
> So these numbers are only per message execution, not accumulated over multiple message executions?

Hmm, I guess we could also track the cumulative numbers so we can extract a ratio of time spent in mutation/GC.
Would an overflowing U64 be enough or should I use a BigInt?

I also noticed that some of the rts_ functions have dummy implementations in the interpreter (that return 0), others are unimplemented and will just crash. Should I give them all dummy implementations so examples can be interpreted without  crashing?

src/codegen/compile.ml

crusso:
I went ahead and adjust the interpreter(s). PTAL @ggreif

Changelog.md
```diff
@@ -1,5 +1,15 @@
 # Motoko compiler changelog
 
+* motoko (`moc`)
+
+  * add primitives
+    ```motoko
+      Prim.rts_mutator_instructions : () -> Nat
+      Prim.rts_collector_instructions : () -> Nat
+    ```
+    to report appoximate IC instruction costs of last message
```
crusso, line 10:
```suggestion
    to report approximate IC instruction costs of last message
```

ggreif (approved):

PR #3383, opened 2022-08-04
feat: actor class management (low level install, reinstall and upgrade)

crusso:
Give some actor class, 'Class', provide a typed interface to canister management as an additional curried function, `installClass`:

```
module Lib  {
   type Class = ...;
   Class : (T1,...TN) -> async Class;
   installClass: {mode : {#install, #reinstall; #upgrade} ; principal : Principal } -> 
                         (T1,..TN) -> 
                           async <ClassName>
}
```

This is a thin veneer around a call to the lower-level `ic00.install_code` method, with the benefit of not exposing the actor class wasm and returning a strongly type actor, not just side-effecting a principal.

Note that to #install an actor from fresh using installClass, you'd need to first call `ic00.create_canister(<canister_settings>)` , unless we choose to hide that and perhaps pass the canister settings, not an existing principal,  as an argument `#install`. TBD.

TODO:
* [ ] clean up code (add some helpers for constructing records etc).
* [x] fix non-deterministic test output causing CI to fail
* [x] extend test to evolve interfaces
* [ ] hoist out the duplicated wasm blob instead of relying on backend to CSE it
* [ ] consider a more refined method that takes a single variant with (fresh) principal and (reused) `actor {}` payloads (not a record).
```
installClass:  {#new CanisterSetting; #install : Principal, #reinstall : actor {} ; #upgrade : actor {}} ->
                       (T1,..,TN) -> 
                            async Class
```
The idea is that this reflects the fact that you can only install code to a fresh principal, and upgrade or reinstall a canister with already installed code (i.e. an `actor {}`). The #new variant is like the default constructor (calling create_canister), but let's you specify the canister settings.

Notes:

Gabor's suggestion: Instead of literally adding a user accessible second constructor with a derived name (`install`...) add dedicated syntax for this system functionality.

Ie. call installCall, @Class (inaccessible to users) and add syntactic sugar

system e.id-> e.@id

So our call to 
```
Lib.InstallNode {mode = #upgrade; principal = p} args becomes:

system Lib.Node {mode = #upgrade, principal = p} args 
```
Caveats:
1. Without more static checks, let's you access all inaccessible @-prefixed members if you know the name. Suffix instead?
2. Won't survive rebinding

```
type N = Lib.Node
system N ...
```
won't work. (But that's true for ordinary class constructors too).


src/prelude/internals.mo
```diff
@@ -370,7 +370,7 @@ func @new_async<T <: Any>() : (@Async<T>, @Cont<T>, @Cont<Error>) {
 module @ManagementCanister = {
   public type wasm_module = Blob;
   public type canister_settings = {
-    controller : ?Principal;
+    controllers : ?[Principal];
```
ggreif, line 373: Why not model `null` with `[]`?

ggreif: Oh, this is an interface given by the IC?

crusso: Yep, and I expect candid would default to null this way if the field is omitted. Our code was actually out-of-date in a benign way.

test/run-drun/map-upgrades/map0.mo
```diff
@@ -0,0 +1,82 @@
...
+  type Value = Text;
+
+  // Number of Nodes
+  let n = 8;
+
+  type Node = Lib.Node;
+
+  stable let savedNodes : [var ?(actor{})] = Prim.Array_init(n, null);
+
+  let nodes : [var ?Node] = Prim.Array_init(n, null);
```
chenyan-dfinity, line 22: Why not making `nodes` stable directly?

test/run-drun/map-upgrades/map0.mo
```diff
@@ -0,0 +1,82 @@
...
+    await node.insert(k, v);
+  };
+
+  system func preupgrade () {
+     for (i in nodes.keys()) {
+       savedNodes[i] := nodes[i];
+     }
+  };
+
+  public func upgradeNodes() {
```
chenyan-dfinity, line 52: Is it better to put this in `postupgrade`?

ggreif:
So in today's SGM I suggested (basically) that
`actor class Name` desugars to two bindings
- `@Name : InstallMode -> ClassArgs -> async actor { ... }`
- `Name : ClassArgs -> async actor { ... }` where `Name` is just `@Name(#new {})`

Then the extended call syntax can be used with
``` Motoko
await system Name (#reinstall actr) (args...)
```
and the conventional call
``` Motoko
await Name (args...)
```
would create a fresh actor as previously.

test/run-drun/map-upgrades/map0.mo
crusso: You can't because `postupgrade` can't send messages - you need to design a user-level protocol to do upgrades of instances, unfortunately.

crusso:
> So in today's SGM I suggested (basically) that `actor class Name` desugars to two bindings
> 
> * `@Name : InstallMode -> ClassArgs -> async actor { ... }`
> * `Name : ClassArgs -> async actor { ... }` where `Name` is just `@Name(#new {})` applied partially
> 
> Then the extended call syntax can be used with
> 
> ```mo
> await system Name (#reinstall actr) (args...)
> ```
> 
> and the conventional call
> 
> ```mo
> await Name (args...)
> ```
> 
> would create a fresh actor as previously.

The only problem is these two caveats (see description)

> 1. Without more static checks, let's you access all inaccessible @-prefixed members if you know the name. Suffix instead?
> 2. Won't survive rebinding

src/prelude/internals.mo
nomeata: The empty set of controllers is a very different setting than `null`, which says “use the default ”, which is the caller.

test/run-drun/map-upgrades/map0.mo
crusso: It's mutable, so invariant, and I expect you wouldn't be able to upgrade to a richer Node type. Using the less informative savedNodes works across evolution of the Node type.

But I could be wrong...

ggreif:
> 1. let's you access all inaccessible @-prefixed members

That is just an example mangling. Just switch so a longer prefix like `@#!Name` or such to avoid unintentional capture. Or to be totally anal one could make names an ADT and have an extra branch for this beast.

crusso:
revised version (WIP) here #3386

crusso:
superseded by #3386

PR #3385, opened 2022-08-05
Update moc.js interface for `motoko` NPM package

rvanasa:
This branch is currently being used to generate the `moc.js` file for the [motoko](https://www.npmjs.com/package/motoko) NPM package, and by extension [embed.smartcontracts.org](https://embed.smartcontracts.org). 

**Changes:**

- Switches the target environment from "isomorphic" to "browser" for consistency between `Sys` and `Sys_js` (everything now works consistently between Node.js and the browser)
- Fixes `readDir(path)`, which previously returned an OCaml array rather than JS array
- Adds `readFile(path)`, `parseMotoko(string)` and `parseCandid(string)` functions
- Increases the interpreter stdout/stderr buffers from 100 to 1000 characters

**Future goals:**
- Fix a corner case where `clearPackages()` doesn't remove packages after running the interpreter
- Pass line/column and type information in an function similar to `parseMotoko()`
- Reset type name counter (`T/1`, `T/2`, etc.) after running the interpreter

**Questions:**
- Is there currently a way to use `compileWasm(path, 'wasi')` without crashing moc.js? I am trying to find a working example to experiment with WASI bindings for Motoko Playground.


chenyan-dfinity:
> Is there currently a way to use compileWasm(path, 'wasi') without crashing moc.js? I am trying to find a working example to experiment with WASI bindings for Motoko Playground.

The code in WASI mode cannot contain `actor`. You can try it in my [first version of playground](https://isdig-jiaaa-aaaab-qabla-cai.raw.ic0.app/). `fac.mo` can be compiled in WASI. `test.mo` also works if you click `New package` and import the default value.

rvanasa:
(Updating the test cases to account for virtual file system error messages)

rvanasa:
Good to know; that makes sense. I'll try it out right now.

chenyan-dfinity:
> This branch is currently being used to generate the moc.js file for the [motoko](https://www.npmjs.com/package/motoko) NPM package, and by extension [embed.smartcontracts.org](https://embed.smartcontracts.org/).

For embed, I think it will use `moc_interpreter.js` instead of `moc.js`? The JS size is much smaller for interpreter.js

> Future goals

Another nice to have future goal: The type name id (e.g. `T/1`, `T/2`) keeps increasing each time we call `moc` in browser. It would be nice to find a way to reset the number for each compiler invocation.

rvanasa:
It seems like moc.js and moc_interpreter.js are both about 20 MB prior to minification and 5 MB afterwards. Does create-react-app / Webpack perform some additional tree-shaking on moc_interpreter.js for production builds?

chenyan-dfinity:
The default `Makefile` is only a debug build. You need to pass `--profile=release` to make the code smaller, but it takes quite a long time to compile. [Here](https://github.com/dfinity/motoko/blob/master/default.nix#L524) is our release build setup.

chenyan-dfinity (approved):
LGTM

rvanasa:
Good to know about the debug vs. release builds. Thanks; I'll update the `motoko` package to offer a choice between using moc.js and moc_interpreter.js for optimization.

crusso:
Hmm, that test that seems to fail (rts_gc_instructions.mo) is one I added recently and uses the performance counter and two new prims to measure instructions costs. It might be producing output that depends on the gc flavour and might need changing. Perhaps disable it for now to see if you pass CI otherwise.

rvanasa:
@chenyan-dfinity, I managed to space-optimize the release builds by an additional 20% using [uglify-js](https://www.npmjs.com/package/uglify-js) ([benchmark](https://github.com/privatenumber/minification-benchmarks#-results)). Would you mind if I added this directly to the Nix build process? I'm thinking of using `npx -y uglify-js@3.16.3 ...` to remove the need to introduce any new build dependencies.

chenyan-dfinity:
Can you confirm the optimized code still works? I tried a few minimizer in the past, they either make the code crash or make the code bigger...

rvanasa:
Found a surprisingly clean way to add minification to the build process. Everything still works as expected (the same minifier is being used on https://embed.smartcontracts.org), and there's a ~20% file size reduction for both moc.js and moc_interpreter.js. To ensure backwards compatibility, the minified files are named with `.min.js` suffixes in the build output directory (alongside the original `.js` files).

chenyan-dfinity:
Sounds good, as long as the CI is happy :)

PR #3393, opened 2022-08-09
Reduce number of serialized `Nat` values in `ser-many-int` test

rvanasa:
Because this test often times out during CI, I increased the testing interval from the 1.000th prime number (7919) to the 10.000th prime number (104729).


crusso:
@ggreif, ok by you? Trying to reduce pressure on CI...

ggreif (approved):
LGTM!

PR #3386, opened 2022-08-05
feat: actor class management secondary class constructor and sugar

crusso:
revision of #3383 with better API 

Given some actor class, `Class`, provide a typed interface to canister management as an additional curried function, `Class` hidden in a user-inaccessible `system` module:

``` motoko
actor class Class (a1:T1, ... an:TN)  {
   ....
}
```
is imported as:

``` motoko
module Lib  {
   type Class = ...;
   Class : (T1, ... TN) -> async Class;
   system : module {
      Class : 
        { #new CanisterSettings; 
          #install : Principal;
          #reinstall : actor {} ;
          #upgrade : actor {} }  ->
            (T1, ... TN) -> 
                async Class
  }
}
```
Abusing keyword system as in inaccessible field name and
where `CanisterSettings` is taken directly from the Interface Spec:

``` motoko
type CanisterSettings = {
     settings : ?{
        controllers : ?[Principal];
        compute_allocation : ?Nat;
        memory_allocation : ?Nat;
        freezing_threshold : ?Nat;
     }
}
```

The variant argument reflects the fact that you can only install code to a fresh principal, and upgrade or reinstall a canister with already installed code (i.e. an `actor {}`). The #new variant is like the default constructor (calling `create_canister`), but lets you specify the canister settings.

Since the module name `system` is an inaccessible keyword, also add sugar

`( system <exp_post> DOT <id>)` 

 desugaring to

`<exp_post>.system.<id>`

to enable explicit access to the secondary constructor.

PR #3394, opened 2022-08-10
feat: actor class management syntactic sugar

crusso:
Builds on #3386 
Add sugar 
```
<exp_post> ::=
   ...
  '(' 'system' ')' <exp_nullary> '.' <id>
```
for actor class management.

For example:

```
 await (system) Lib.Node (#upgrade n) (i); // upgrade!
````
is just

```
await Lib.installNode (#upgrade n) (i); // upgrade!
```
(accessing the secondary, mangled, power-user constructor).

This is just sugar so the error messages will be cryptic if the expression isn't an actor class lib or the name is wrong. To improve that, we need a new AST node, typing rules and slightly richer type environment (tracking classes vs functions).

### DIFF
--- a/doc/md/examples/grammar.txt
+++ b/doc/md/examples/grammar.txt
@@ -158,6 +158,7 @@
     <exp_post> '.' <id>
     <exp_post> ('<' <list(<typ>, ',')> '>')? <exp_nullary>
     <exp_post> BANG
+    '(' 'system' ')' <id> '.' <id>
 
 <exp_un> ::= 
     <exp_post>
--- a/doc/md/examples/grammar.txt
+++ b/doc/md/examples/grammar.txt
@@ -158,7 +158,7 @@
     <exp_post> '.' <id>
     <exp_post> ('<' <list(<typ>, ',')> '>')? <exp_nullary>
     <exp_post> BANG
-    '(' 'system' ')' <id> '.' <id>
+    '(' 'system' ')' <exp_nullary> '.' <id>
 
 <exp_un> ::= 
     <exp_post>
--- a/src/mo_frontend/parser.mly
+++ b/src/mo_frontend/parser.mly
@@ -582,9 +582,8 @@ exp_post(B) :
     { CallE(e1, inst, e2) @? at $sloc }
   | e1=exp_post(B) BANG
     { BangE(e1) @? at $sloc }
-  | LPAR SYSTEM RPAR lib = id DOT c=id
-    { DotE((VarE(lib.it @@ lib.at)) @? lib.at,
-       { c with it = "install" ^ c.it}) @? at $sloc }
+  | LPAR SYSTEM RPAR e1=exp_nullary(B) DOT c=id
+    { DotE(e1, { c with it = "install" ^ c.it}) @? at $sloc }
 
 exp_un(B) :
   | e=exp_post(B)
--- a/src/mo_frontend/parser.mly
+++ b/src/mo_frontend/parser.mly
@@ -556,6 +556,7 @@ exp_plain :
   | LPAR es=seplist(exp(ob), COMMA) RPAR
     { match es with [e] -> e | _ -> TupE(es) @? at $sloc }
 
+
 exp_nullary(B) :
   | e=B
     { e }
@@ -581,6 +582,9 @@ exp_post(B) :
     { CallE(e1, inst, e2) @? at $sloc }
   | e1=exp_post(B) BANG
     { BangE(e1) @? at $sloc }
+  | LPAR SYSTEM RPAR lib = id DOT c=id
+    { DotE((VarE(lib.it @@ lib.at)) @? lib.at,
+       { c with it = "install" ^ c.it}) @? at $sloc }
 
 exp_un(B) :
   | e=exp_post(B)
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by crusso** [comment]
Variations on a theme:

```
<exp_post> ::=
   ...
  '(' 'system'  <exp_nullary> '.' <id> ')'

(system Lib.Node) (#upgrade n) (i);
```
or (not sure this would work)

```
<exp_post> ::=
   ...
  '(' 'system' <exp_nullary> <exp_nullary> '.' <id> <params> ')'
(system (#uprade n) Lib.Node (i));
````

**Comment by crusso** [comment]
Another option is to put the secondary constructor in a submodule called `system` and extend dot notation to allow `exp.system` (projection on the keyword `system`).

That should prevent users from redefining `system` fields, but allow them to access them, and might be used for other `system` functions.

```
eg. Lib.system.C(#upgrade a)(i);
```

**Comment by ggreif** [comment]
> Another option is to put the secondary constructor in a submodule called `system` and extend dot notation to allow `exp.system` (projection on the keyword `system`).
> 
> That should prevent users from redefining `system` fields, but allow them to access them, and might be used for other `system` functions.
> 
> ```
> eg. Lib.system.Node(#upgrade a)(i);
> ```
> 
> @ggreif @nomeata @rossberg @matthewhammer @rvanasa @kentosugama: yuck?

Have fun parsing that path with a reserved identifier as one of the components! Yuck.

**Comment by matthewhammer** [comment]
If we can settle on "Lib.system.NodeFactory", or some variation of that, then we have a deal!

All joking aside, using `system` as a keyword seems nice and appropriate to me, given how Motoko is trying to expose "the system," albeit in a somewhat managed way.  The double dots seems intimidating to me, like there's going to be a lot of stuff associated with `Lib` other than `system`-related things.

**Comment by crusso** [comment]
> > Another option is to put the secondary constructor in a submodule called `system` and extend dot notation to allow `exp.system` (projection on the keyword `system`).
> > That should prevent users from redefining `system` fields, but allow them to access them, and might be used for other `system` functions.
> > ```
> > eg. Lib.system.Node(#upgrade a)(i);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @ggreif @nomeata @rossberg @matthewhammer @rvanasa @kentosugama: yuck?
> 
> Have fun parsing that path with a reserved identifier as one of the components! Yuck.

It's actually trivial to parse #3403. Given than, is it worth doing though?

**Comment by crusso** [comment]
> If we can settle on "Lib.system.NodeFactory", or some variation of that, then we have a deal!
> 
> All joking aside, using `system` as a keyword seems nice and appropriate to me, given how Motoko is trying to expose "the system," albeit in a somewhat managed way. The double dots seems intimidating to me, like there's going to be a lot of stuff associated with `Lib` other than `system`-related things.

There is already stuff in `Lib`, namely the type `Lib.Node`, and the value `Lib.Node`.

At the moment, an actor class Node is imported as a module
```
{ type Node;
  Node : int -> Node;
}
```
This PR generalises that to 
```
{ type Node;
  Node : int -> Node;
  installNode: {#new ...} -> int -> Node
}
```
with sugar to call the secondary `install`-prefixed system constructor.

The idea with the `system` field is to, instead, import a nested module carrying the (unmangled) secondary constructor:

```
{ type Node;
  Node : int -> Node;
  system : module { 
      Node: {#new ...} -> int -> Node
   }
}
```
Using module name spacing instead of name mangling plus sugar.

kentosugama:
Agree with @matthewhammer, I think having the `system` keyword be standalone and visible like in its other uses is really nice and clear.

My favorite is this one from higher up in the thread:
`system (#upgrade n) Lib.Node (i)`

because it goes from the most general scope to the most specific. `system` indicates you're invoking some IC feature, #upgrade indicates which feature you're invoking, Lib.Node indicates which entity you're invoking it on.

Maybe `#upgrade` is more specific in scope than `Lib.Node`, because the "upgrade" is specific to `Lib.Node`. In which case, I like this one:
`system Lib.Node (#upgrade n) (i)`

But the first argument makes more sense to me.

And,
`Lib.system.Node`... tentatively, yuck, Unless there's some reason this would be better for developer experience than the other options.

crusso:
> Agree with @matthewhammer, I think having the `system` keyword be standalone and visible like in its other uses is really nice and clear.
> 
> My favorite is this one from higher up in the thread: `system (#upgrade n) Lib.Node (i)`
> 
> because it goes from the most general scope to the most specific. `system` indicates you're invoking some IC feature, #upgrade indicates which feature you're invoking, Lib.Node indicates which entity you're invoking it on.
> 
> Maybe `#upgrade` is more specific in scope than `Lib.Node`, because the "upgrade" is specific to `Lib.Node`. In which case, I like this one: `system Lib.Node (#upgrade n) (i)`
> 
> But the first argument makes more sense to me.
> 
> And, `Lib.system.Node`... tentatively, yuck, Unless there's some reason this would be better for developer experience than the other options.

The only advantages is that it's re-using module namespacing, could be re-(ab)used for other, unrelated system features,  and perhaps most DX related: would support dot notation driven auto-completion in an IDE.

crusso:
superseded by #3386

PR #3403, opened 2022-08-16
feat: actor class management: syntactic sugar using 'system' sub module

crusso:
This PR puts the secondary constructor in a submodule called `system` and extends the dot notation to allow `exp.system` (projecting on the *keyword* `system`).

That should prevent users from redefining `system` fields, but allow them to access them, and might be used for other `system` functions.

```
eg. Lib.system.Node(#upgrade a)(i);
```

Implements the alternative sugar described here:

https://github.com/dfinity/motoko/pull/3394#issuecomment-1215603163


crusso:
superseded by #3386

Issue #3402, opened 2022-08-16
Optimise Motoko Assignment


ggreif:
I have just encountered such compiled code:
``` wasm
    …
    call $compute
    local.set $new_val
    i32.const 65727      // destination mutcell
    local.get $new_val
    i32.store offset=5
```
Should be (when destination is not side-effecting!)
``` wasm
    i32.const 65727      // destination mutcell
    …
    call $compute
    i32.store offset=5
```

Source is
``` Motoko
object a {
    var v = 42;
    public func incr() = v += 1
};
```

I think mutable object access (`DotLE`) is (almost) never side-effecting in Motoko, as folks seldom write stuff like
``` Motoko
var foo = 0;
{ var a = do {foo := 7; 9} }.a += 1;
```

nomeata:
Even when the destination is sideeffecting it would be reasonable to calculate the destination first, and then the value, and then invoke `store`.

nomeata:
Ah, simple, just make a local change here

https://github.com/dfinity/motoko/blob/f325002a4d10a172d96a317d8e00bd7a9cb41478/src/codegen/compile.ml#L6768-L6772

nomeata:
This comes from

https://github.com/dfinity/motoko/blob/f325002a4d10a172d96a317d8e00bd7a9cb41478/src/codegen/compile.ml#L6768-L6772

and `set_val` expects the value on the stack.

Could be fixed if `set_val` returns some `pre_code` and some `post_code` instead.

nomeata:
https://github.com/dfinity/motoko/blob/f325002a4d10a172d96a317d8e00bd7a9cb41478/src/codegen/compile.ml#L8143-L8145

`compile_lexp` is already set up to use that pattern, so I don’t see problems pushing that pattern through to `set_val`.

PR #3406, opened 2022-08-17
Backend: Avoid temporary variable in `set_val`

nomeata:
as suggested by @ggreif. Fixes: #3402


nomeata:
Something is still wrong:
```
--- actor-class-arg.drun (expected)
+++ actor-class-arg.drun (actual)
@@ -20,8 +20,7 @@
 debug.print: Hello
 debug.print: World
-debug.print: Hello
-ingress Completed: Reply: 0x4449444c0000
+ingress Err: IC0502: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped: stack overflow
 Ok: Reply: 0x4449444c0000
 debug.print: Hello
 debug.print: World
```

nomeata:
What is the difference between 
```
actor class (t : Text, t1 : Text) {
```
and
```
shared actor class (t : Text, t1 : Text) {
```
?

src/codegen/compile.ml
```diff
@@ -6770,11 +6770,24 @@ module Var = struct
     | Some (PublicMethod _) -> fatal "set_val: %s is PublicMethod" var
     | None   -> fatal "set_val: %s missing" var
 
-  (* Stores the payload (which is found on the stack, in Vanilla stackrep) *)
+  (* Stores the payload (using preparation code and consumptiong code) *)
```
ggreif, line 6773: consuming? downstream?

nomeata:
```suggestion
  (* Stores the payload. Returns stack preparation code, and code that consumes the values from the stack *)
```

nomeata:
So much for premature optimization:

>  In terms of size, 4 tests improved and the mean change is -0.1%.

crusso:
> What is the difference between
> 
> ```
> actor class (t : Text, t1 : Text) {
> ```
> 
> and
> 
> ```
> shared actor class (t : Text, t1 : Text) {
> ```
> 
> ?
Six characters? 

The latter gives you the option of accessing the caller...
```
 shared {caller} actor class (t : Text, t1 : Text) {
```

crusso:
> > What is the difference between
> > ```
> > actor class (t : Text, t1 : Text) {
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and
> > ```
> > shared actor class (t : Text, t1 : Text) {
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> 
> Six characters?
> 
> The latter gives you the option of accessing the caller...
> 
> ```
>  shared {caller} actor class (t : Text, t1 : Text) {
> ```

Hmm, maybe we should have attached it to `actor` and `func` (as an optional prefix) instead, so you could get the owner of a plain old actor without having to declare a class....

nomeata:
I was just surprised it changed behaviour somehow – with `shared` it triggered an earelier bug on this PR, without it didn’t

ggreif (commented):
Is there any change in the evaluation order?

nomeata:
No, it's only moving the address pushes earlier, no side effects there

crusso:
> No, it's only moving the address pushes earlier, no side effects there 

It might have to move the context pat into the scope of the arguments,  requiring some renaming and extra code. Maybe the compiler doesn't spot that the pat is still trivial in this case.

nomeata:
I don't follow. I merely reorder some wasm instructions. Nothing big source-levely happening.

ggreif (approved):
LGTM! Thanks for tackling this!

PR #3407, opened 2022-08-17
doc: describe typing and evaluation of record combination/extensions

crusso:
Document of object combination/extension describing typing and evaluation.

### DIFF
--- a/doc/md/language-manual.md
+++ b/doc/md/language-manual.md
@@ -467,6 +467,7 @@ The syntax of an *expression* is as follows:
   <exp> . <nat>                                  tuple projection
   ? <exp>                                        option injection
   { <exp-field>;* }                              object
+  { <exp> (and <exp>)* (with <exp-field>;+)? }   object combination/extension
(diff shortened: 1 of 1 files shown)


doc/md/language-manual.md
```diff
@@ -467,6 +467,7 @@ The syntax of an *expression* is as follows:
   <exp> . <nat>                                  tuple projection
   ? <exp>                                        option injection
   { <exp-field>;* }                              object
+  { <exp> (and <exp>)* (with <exp-field>;+)? }   object combination/extension
```
ggreif, line 470: Can we say `<exp_post>` here, i.e. `M.Class()` is allowed, but `a <some operator> b` is not.

crusso: Right, but the manual isn't really talking about the grammar but the ast. This would be the only place it does that place.

ggreif: That's fine, then.

ggreif (approved):
LGTM!

PR #3408, opened 2022-08-18
fix: type field related corner cases to #3084

crusso:
fixes more corner cases of #3084, adding some tests. In particular,

* value and type fields reside in different namespaces and should not be inadvertently masked by explicit field overrides.
* ambiguous type fields can be accommodate when they have equivalent definitions.


test/fail/ok/bad-obj-base-types.tc.ok
```diff
@@ -0,0 +1,12 @@
+bad-obj-base-types.mo:8.18-8.20: type error [M0177], ambiguous type field in base
+  t
+bad-obj-base-types.mo:8.11-8.13: info, field also present in base, here (consider overwriting)
```
ggreif, line 3: I'd drop the `(consider overwriting)` part for now in the case of type fields.

src/mo_frontend/typing.ml
```diff
@@ -1070,23 +1070,35 @@ and infer_exp'' env exp : T.typ =
       if s = T.Actor then
         error env base.at "M0178"
           "actors cannot serve as bases in record extensions";
-      T.(Obj (Object, filter (fun f -> not (mem f.lab fls)) tfs)) in
+      T.(Obj (Object, filter (fun ft -> not (List.exists (eponymous_fields ft) fts)) base_fts))
+    in
     let stripped_bases = map strip bases in
 
+    let ambiguous_fields ft1 ft2 =
```
ggreif, line 1077: This is great!

test/fail/ok/bad-obj-base-types.tc.ok
ggreif: I have a patch for this!

crusso: Yeah, I agree.

ggreif: Okay to `git push`?

src/mo_frontend/typing.ml
```diff
@@ -1070,23 +1070,35 @@ and infer_exp'' env exp : T.typ =
...
+      eponymous_fields ft1 ft2 &&
+      (* allow equivalent type fields *)
+      T.(match (ft1.typ, ft2.typ) with
```
ggreif, line 1080:
```suggestion
      T.(match ft1.typ, ft2.typ with
```

test/fail/ok/bad-obj-base-types.tc.ok
crusso: Sure, you might need to pull first.

ggreif: Yeah `git pull --rebase` to the rescue! Pushed.

test/run/object-extend-type.mo
```diff
@@ -20,3 +20,14 @@ let bm = { b with M };
 
 let aM = { a and M };
 let bM = { b and M };
+
+// poor man's type definition to declare object type with type fields
```
ggreif, line 24: This is clever, I didn't imagine this is possible at all with type fields! Now I can implement the commented test above :-)

ggreif (approved):
Thanks, this looks perfect!

test/run/object-extend-type.mo
crusso: We should just add type fields proberly (to record types and values), but for now the hack works.

PR #3409, opened 2022-08-19
Bump nixpkgs, without upgrading the Ocaml wasm library

nomeata:
Adjusting our code to work with wasm-2.0.0 happens in #3364.


ggreif (approved):
Good idea to keep `wasm 1.1.1`! LGTM!

PR #3410, opened 2022-08-19
feat: yet another variant of actor class management syntax

crusso:
Builds on the Lib.system variant (#3403) (to avoid mangling), but also adds sugar. We can drop the Lib.system support, if desired (though its actually more discoverable in an IDE).
```
'(' 'system' <exp_nullary> <exp_nullary> '.' <id> <exp_nullary> ')'
```

eg.
```
 (system (#upgrade n) Lib.Node(i))
```
(but `Lib.system.Node(#upgrade)(i)` also works)


crusso:
superseded by #3386

PR #3411, opened 2022-08-20
Make overwriting of `var` fields from bases obligatory

ggreif:
Make mutable field aliasing (when extending/merging records) an experimental opt-in feature, and hide it behind a feature flag. When the flag is not supplied, an error is emitted that instructs the user to overwrite all `var` fields from bases explicitly. When `--experimental-field-aliasing` is given no such error gets emitted, and the mutable cells of bases are aliased in the resulting record.

See discussion in #3084 why this is a necessity.

Also contains a few minor tweaks.

### DIFF
--- a/src/exes/moc.ml
+++ b/src/exes/moc.ml
@@ -154,9 +154,9 @@ let argspec = [
   Arg.Set_int Flags.max_stable_pages,
   "<n>  set maximum number of pages available for library `ExperimentalStableMemory.mo` (default " ^ (Int.to_string Flags.max_stable_pages_default) ^ ")";
 
-  "--experimental-references",
-  Arg.Unit (fun () -> Flags.experimental_references := true),
-  " enable (experimental, limited) support for var references"
+  "--experimental-field-aliasing",
+  Arg.Unit (fun () -> Flags.experimental_field_aliasing := true),
+  " enable experimental support for aliasing of var fields"
   ]
 
   @  Args.inclusion_args
--- a/src/exes/moc.ml
+++ b/src/exes/moc.ml
@@ -152,7 +152,7 @@ let argspec = [
 
   "--max-stable-pages",
   Arg.Set_int Flags.max_stable_pages,
-  "<n>  set maximum number of pages available for library `ExperimentStableMemory.mo` (default " ^ (Int.to_string Flags.max_stable_pages_default) ^ ")";
+  "<n>  set maximum number of pages available for library `ExperimentalStableMemory.mo` (default " ^ (Int.to_string Flags.max_stable_pages_default) ^ ")";
 
   "--experimental-references",
   Arg.Unit (fun () -> Flags.experimental_references := true),
--- a/src/exes/moc.ml
+++ b/src/exes/moc.ml
@@ -152,9 +152,12 @@ let argspec = [
 
   "--max-stable-pages",
   Arg.Set_int Flags.max_stable_pages,
-  "<n>  set maximum number of pages available for library `ExperimentStableMemory.mo` (default " ^ (Int.to_string Flags.max_stable_pages_default) ^ ")"
-  ]
+  "<n>  set maximum number of pages available for library `ExperimentStableMemory.mo` (default " ^ (Int.to_string Flags.max_stable_pages_default) ^ ")";
 
+  "--experimental-references",
+  Arg.Unit (fun () -> Flags.experimental_references := true),
+  " enable (experimental, limited) support for var references"
+  ]
 
   @  Args.inclusion_args
 
--- a/src/mo_config/flags.ml
+++ b/src/mo_config/flags.ml
@@ -40,4 +40,4 @@ let error_detail = ref 2
 let sanity = ref false
 let gc_strategy = ref Copying
 let force_gc = ref false
-let experimental_references = ref false
+let experimental_field_aliasing = ref false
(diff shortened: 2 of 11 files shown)


test/fail/ok/object-ext-var-base.tc.ok
```diff
@@ -0,0 +1,6 @@
+object-ext-var-base.mo:9.11-9.18: type error [M0179], base has var field 
+  v
+object-ext-var-base.mo:9.11-9.18: info, overwrite field to resolve error
+object-ext-var-base.mo:9.11-9.18: type error [M0179], base has var field
```
ggreif, line 4: Buuuh, why do we get 2 error messages?

ggreif: probably because of synthesis context, with
``` diff
index 9a14cd5c8..ccf97b925 100644
--- a/test/fail/object-ext-var-base.mo
+++ b/test/fail/object-ext-var-base.mo
@@ -6,4 +6,4 @@ object no_var {
     public let w = 25
 };
 
-let _ = { has_var and no_var }
+let _ = { has_var and no_var } : {var v : Nat; w : Nat }
```
I only get a single error.

src/exes/moc.ml
```diff
@@ -152,9 +152,12 @@ let argspec = [
 
   "--max-stable-pages",
   Arg.Set_int Flags.max_stable_pages,
-  "<n>  set maximum number of pages available for library `ExperimentStableMemory.mo` (default " ^ (Int.to_string Flags.max_stable_pages_default) ^ ")"
-  ]
+  "<n>  set maximum number of pages available for library `ExperimentStableMemory.mo` (default " ^ (Int.to_string Flags.max_stable_pages_default) ^ ")";
```
crusso, line 155:
```suggestion
  "<n>  set maximum number of pages available for library `ExperimentalStableMemory.mo` (default " ^ (Int.to_string Flags.max_stable_pages_default) ^ ")";
```

src/exes/moc.ml
```diff
@@ -152,9 +152,12 @@ let argspec = [
...
 
+  "--experimental-references",
```
crusso, line 157:
```suggestion
  "--experimental-field-aliasing",
```

src/exes/moc.ml
```diff
@@ -152,9 +152,12 @@ let argspec = [
...
+  Arg.Unit (fun () -> Flags.experimental_references := true),
+  " enable (experimental, limited) support for var references"
```
crusso, line 159:
```suggestion
  " enable experimental support for aliasing of var fields"
```

src/mo_frontend/typing.ml
```diff
@@ -1106,6 +1106,20 @@ and infer_exp'' env exp : T.typ =
...
     disjoint (map2 (fun b_t b -> b_t, b) stripped_bases exp_bases);
 
+    (* do not allow var fields for now (to avoid aliasing) *)
+    if not !Flags.experimental_references then
+      T.(let immutable_base b_typ b_exp =
+           let constant_field (ft : field) =
+             if (is_mut ft.typ) then
+               begin
+                 local_error env b_exp.at "M0179"
+                   "base has var field %a"
```
crusso, line 1116:
```suggestion
                   "base has var field %a that cannot be aliased"
```

src/mo_config/flags.ml
```diff
@@ -40,3 +40,4 @@ let error_detail = ref 2
 let sanity = ref false
 let gc_strategy = ref Copying
 let force_gc = ref false
+let experimental_references = ref false
```
crusso, line 43:
```suggestion
let experimental_field_aliasing = ref false
```

src/mo_frontend/typing.ml
```diff
@@ -1106,6 +1106,20 @@ and infer_exp'' env exp : T.typ =
         disjoint t in
     disjoint (map2 (fun b_t b -> b_t, b) stripped_bases exp_bases);
 
+    (* do not allow var fields for now (to avoid aliasing) *)
+    if not !Flags.experimental_references then
```
```suggestion
    if not !Flags.experimental_field_aliasing then
```

src/exes/moc.ml
```diff
@@ -152,9 +152,12 @@ let argspec = [
 
   "--max-stable-pages",
   Arg.Set_int Flags.max_stable_pages,
-  "<n>  set maximum number of pages available for library `ExperimentStableMemory.mo` (default " ^ (Int.to_string Flags.max_stable_pages_default) ^ ")"
-  ]
+  "<n>  set maximum number of pages available for library `ExperimentStableMemory.mo` (default " ^ (Int.to_string Flags.max_stable_pages_default) ^ ")";
 
+  "--experimental-references",
+  Arg.Unit (fun () -> Flags.experimental_references := true),
```
crusso, line 158:
```suggestion
  Arg.Unit (fun () -> Flags.experimental_field_aliasing := true),
```

crusso (approved):

test/fail/ok/object-ext-var-base.tc.ok
ggreif: resolved by 1a688a28369d809d8ca7af32254b117c5791d41f, thanks @crusso!

src/mo_frontend/typing.ml
ggreif: 9bd42f9d9

PR #3415, opened 2022-08-22
feat: yet another variant of actor class management syntax (combining features of two previous version)

crusso:
Combines #3403 (placing secondary constructor in `system` submodule)`, but uses dedicated syntax `(system) e` to access `e.system`.

```
<exp_post> ++=
  '(' 'system' ')' <exp_nullary>
```
E.g.

```
await (system) Lib.Node(#upgrade n)(i);
```

`(system) <exp>` is just notation for (otherwise inexpressible)  `<exp>.system` and nests nicely with the dot notation and application.

(Missed the argument re-ordering of #3410 "(system (#upgrade) Lib.Node(i))" desired by @kentosugama but the desugaring here is much simpler and could be used for other functionality hidden in `system` modules (unlike #3410)).

### DIFF
--- a/doc/md/examples/grammar.txt
+++ b/doc/md/examples/grammar.txt
@@ -155,6 +155,7 @@
     '[' 'var'? <list(<exp_nonvar>, ',')> ']'
     <exp_post> '[' <exp> ']'
     <exp_post> '.'<nat>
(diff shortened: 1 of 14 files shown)


crusso:
superseded by #3386

PR #3433, opened 2022-08-26
Add status badges to readme

rvanasa:
Adds a few status badges to the readme (from [Shields.io](https://shields.io/), which is essentially the standard for these on GitHub).

For context, here is [a similar example from the `embed-motoko` repository](https://github.com/dfinity/embed-motoko#readme).

Since so many people seem to love these badges, I'll also look into submitting the DFINITY logo to [the icon pack used by Shields.io](https://simpleicons.org/) so that developers can include an "IC mainnet badge" for their dapps.


ggreif (approved):
Eye candy? I am fine with that! Can you explain the `npm` bit?

nomeata (approved):

rvanasa:
The `npm` badge lets people know that we have a way of using `moc` through JavaScript, which is an appeal to developer grant recipients and open-source contributors (who are most likely to care about the badges).

PR #3450, opened 2022-09-16
Add `bat` to the shell toolset

ggreif:
a small colored `cat` utility, highlighting code


nomeata (dismissed):
Ok, (in general I don't think we should all useful random unrelated tools.)

ggreif:
> Ok, (in general I don't think we should all useful random unrelated tools.)

Heh, it's like 2 tools per year, maybe with a decreasing cadence :-)

ggreif:
@rvanasa currently `bat` won't highlight Motoko (and neither Viper), do you have an idea what it would take to add support for them?

luc-blaeser:

Issue #3449, opened 2022-09-16
Bug: type components in stable types lead to compilation failure.


crusso:
See https://forum.dfinity.org/t/stable-types-build-error-moving-to-0-9-3-and-later-versions-of-the-sdk/15225.

Possible fixes (in order of complexity).

- deem type components as non-shared & non-stable (might break old programs, but can always ingore upgrade warning)
- erase type components before producing stable variable sig (?)
- extend syntactic types to support type components #3457

crusso:
Fixed in #3457

PR #3453, opened 2022-09-17
experiment: make type fields non-shared and non-stable

crusso:
Possible fix for #3449.


src/pipeline/pipeline.ml
```diff
@@ -264,6 +264,7 @@ let stable_compatible pre post : unit Diag.result =
 let validate_stab_sig s : unit Diag.result =
   let open Diag.Syntax in
   let name = "stable-types" in
+  Printf.eprintf "%s" s;
```
crusso, line 267: Delete me

PR #3455, opened 2022-09-19
Add source locations and exported types to `moc.js` parser output 

rvanasa:
The main addition in this PR is `Mo_def.Arrange_source`, a more verbose alternative to `Mo_def.Arrange` which includes source locations in the generated s-expression tree. This will be used in the new `dfx`-agnostic VS Code language server as well as in a future upgrade to `mo-doc`. 

Follow-up questions:
- Is there a previously existing way to intersperse type information throughout an AST? 
  - Anything similar to transforming `1 + 2` to `((1: Nat) + (2: Nat)): Nat` would work for the anticipated use cases. 
- Does anything currently rely on the `Mo_def.Arrange` s-expressions, aside from debug output? Since there's a lot of duplicated code between `Arrange` and `Arrange_source`, it probably makes sense to either combine the modules (and pass along a config value), or simply deprecate `Arrange` and include source information everywhere. Curious to hear any opinions about this.


crusso:
> The main addition in this PR is `Mo_def.Arrange_source`, a more verbose alternative to `Mo_def.Arrange` which includes source locations in the generated s-expression tree. This will be used in the new `dfx`-agnostic VS Code language server as well as in a future upgrade to `mo-doc`.
> 
> Follow-up questions:
> 
> * Is there a previously existing way to intersperse type information throughout an AST?

No. I think the -v -t -dt option will dump the tree after parsing with some type info printed out (but not using sexpresssions but pretty printed types for types).

  
>   * Anything loosely equivalent to transforming `1 + 2` to `((1: Nat) + (2: Nat)): Nat` would work for the anticipated use cases.

The problem with this is that the type information will probably get very large.

> * Does anything currently rely on the `Mo_def.Arrange` s-expressions, aside from debug output? Since there's a lot of duplicated code between `Arrange` and `Arrange_source`, it probably makes sense to either combine the modules (and pass along a config value), or simply deprecate `Arrange` and include source information everywhere. Curious to hear any opinions about this.

No, it's just for debug output. Feel free to combine if that's sensible.

rvanasa:
This should be ready to merge (since it's already being used in both `node-motoko` and `vscode-motoko`). 

A few additional improvements and optimizations will be included in #3478.

rvanasa:
@crusso, would you mind if I merged this (into `master` or at least `viper` for now)? This contains most of the recent VS Code extension improvements, which could be handy for the Viper demo.

chenyan-dfinity (approved):
LGTM. Do we know the JS size diff after adding this?

rvanasa:
The difference is 17 KB for `moc.js` and 11 KB for `moc_interpreter.js`, so less than a 1% increase in file size.

PR #3463, opened 2022-09-27
perf: optimize various hot functions on large compiles

crusso:
- Avoid use of polymorphic comparison on types, use bespoke comparison instead.
- Optimize hashing of field/variant names.

Cuts compilation times roughly in half (but largely because we type check all intermediate representations).


nomeata (commented):
Did you use profiling to identify the hot paths? Or just gut feeling?

crusso:
> Did you use profiling to identify the hot paths? Or just gut feeling?

Profiling. On large examples we spend 25 percent of time in polymorphic comparison.

I'm currently shaving 12 seconds off a 50 second compile.

crusso:
But it makes very little difference with -no-check-ir (but I'd expect that).

crusso:
@nomeata optimizing Hash.hash to avoid String.explode shaves off another 10s from the (original 50s) compile, now down to 28s.

Not too shabby.

nomeata:
Pretty good. Make sure to brag a bit in the release notes :-)

src/mo_frontend/variance.ml
```diff
@@ -34,7 +34,11 @@ let flip p =
 module PS = Set.Make
  (struct
     type pol = t (* rename to avoid capture below *)
-    type t = pol * typ let compare = compare
+    type t = pol * typ
+    let compare (p1,t1) (p2, t2) =
+      match compare p1 p2 with
+      | 0 -> Ord.compare t1 t2
+      | o -> o
```
crusso, line 41:
```suggestion
      | ord -> ord
```

src/mo_types/cons.ml
```diff
@@ -33,5 +33,9 @@ let name c = c.name
...
   if not show_stamps || c.stamp = 0 then c.name else Printf.sprintf "%s%s%i" c.name sep c.stamp
 
-let eq c1 c2 = (c1.name, c1.stamp) = (c2.name, c2.stamp)
-let compare c1 c2 = compare (c1.name, c1.stamp) (c2.name, c2.stamp)
+let eq c1 c2 = c1.stamp = c2.stamp && c1.name = c2.name
+
+let compare c1 c2 =
+  match Int.compare c1.stamp c2.stamp with
+  | 0 -> String.compare c1.name c2.name
+  | o -> o
```
crusso, line 41:
```suggestion
  | ord -> ord
```

crusso:
WTF? Why am I getting a stack overflow?

src/mo_frontend/variance.ml
```diff
@@ -34,7 +34,11 @@ let flip p =
 module PS = Set.Make
  (struct
     type pol = t (* rename to avoid capture below *)
-    type t = pol * typ let compare = compare
+    type t = pol * typ
+    let compare (p1,t1) (p2, t2) =
```
crusso, line 38:
```suggestion
    let compare (p1, t1) (p2, t2) =
```

src/mo_types/hash.ml
```diff
@@ -5,11 +5,8 @@ which results in values [0..2^31-1]
...
-
-  logand (shift_right_logical minus_one 1) (
-    List.fold_left
-      (fun s c -> add (mul s (of_int 223)) (of_int (Char.code c)))
-      zero
-      (Lib.String.explode s)
-  )
-
+  let sum = ref zero in
+  for i = 0 to String.length s -1 do
```
crusso, line 9:
```suggestion
  for i = 0 to String.length s - 1 do
```

src/mo_frontend/typing.ml
```diff
@@ -633,7 +633,7 @@ and check_con_env env at ce =
   if not (T.ConSet.is_empty cs) then
     error env at "M0157" "block contains non-productive definition%s %s"
       (if T.ConSet.cardinal cs = 1 then "" else "s")
-      (String.concat ", " (List.map Cons.name (T.ConSet.elements cs)));
+      (String.concat ", " (List.sort compare (List.map Cons.name (T.ConSet.elements cs))));
```
ggreif, line 636:
```suggestion
      (String.concat ", " List.(sort compare (map Cons.name (T.ConSet.elements cs))));
```

src/mo_types/type.ml
```diff
@@ -67,6 +67,191 @@ and kind =
   | Def of bind list * typ
   | Abs of bind list * typ
 
+(* Efficient comparison *)
+let tag_prim p =
+  match p with
+  | Null -> 0
```
ggreif, line 73: Isn't there an OCaml magic builtin to get the raw tag of a sum type?

src/mo_types/type.ml
```diff
@@ -67,6 +67,191 @@ and kind =
...
+  | Obj _ -> 8
+  | Variant _ -> 9
+  | Mut _ -> 10
+  | Any -> 11
+  | Non -> 12
+  | Pre -> 13
+  | Typ _ -> 14
+
+let compare_prim p1 p2 =
+  let d = (tag_prim p1) - (tag_prim p2) in
```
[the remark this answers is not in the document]
ggreif: also redundant parens :-)

src/mo_types/type.ml
crusso: Yeah, but I don't want to use magic.

src/mo_types/type.ml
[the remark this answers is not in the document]
crusso: Morally yes, but Int.compare is actually defined as Stdlib.compare. I thought I'd measured this to be faster but I think it's actually lost in the noise anyway.

src/mo_frontend/typing.ml
crusso: Tried that, but picks up wrong compare.

src/mo_types/type.ml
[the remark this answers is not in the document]
crusso: However, cumulatively, I've literally halved compilation times this way.

src/mo_types/type.ml
ggreif: Otherwise this could be interesting
https://www.reddit.com/r/ocaml/comments/mzlsb4/comment/gw1dsqv

src/mo_types/type.ml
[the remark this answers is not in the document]
crusso: Ok, I tried just inlining calls to Int.compare (tag_xxx x) (tag_xxx  x) and it seems just as fast, so I'll do that to save some local functions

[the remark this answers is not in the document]
ggreif: Probably because you are inlining :-)

src/mo_types/type.ml
crusso: Yeah, I'm aware there's a ppx for deriving fast comparison functions but I didn't want to depend on that (there's not that much code here). I guess we could replace by that in future, but there's also the possibility that we can do better with hand-rolled code (by choosing the order of comparison for multiple fields, for example).

src/mo_types/type.ml
```diff
@@ -67,6 +67,205 @@ and kind =
...
+  | Actor -> 2
+  | Memory -> 3
+
+let tag_control = function
+  | Returns -> 0
+  | Promises -> 1
+  | Replies -> 2
+
+let tag t =
+  match t with
```
ggreif, line 109:
```suggestion
let tag = function
```

ggreif: Or is that too slow?

src/mo_types/type.ml
```diff
@@ -67,6 +67,205 @@ and kind =
...
+let compare_rel (t1, t2) (u1, u2) =
+  match compare_typ (t1 : typ) (u1 : typ) with
+  | 0 -> compare_typ (t2 : typ) (u2 : typ)
+  | ord -> ord
+
+(* Syntactic orderings *)
+
+module Ord = struct
+  type t = typ
+  let compare t1 t2 = compare_typ t1 t2
```
ggreif, line 261: would eta-reducing speed this up?

src/mo_types/type.ml
```diff
@@ -67,6 +67,205 @@ and kind =
...
+end
+
+module OrdPair = struct
+  type t = typ * typ
+  let compare p1 p2 = compare_rel p1 p2
```
ggreif, line 266: ditto

test/run/ok/3057.tc.ok
```diff
@@ -1,6 +0,0 @@
-OOPS! You've triggered a compiler bug.
```
ggreif, line 1: OOOH! Nice side-effect of making things performant!

src/mo_types/type.ml
```diff
@@ -67,6 +67,205 @@ and kind =
...
+  | Scope, Scope -> 0
+  | Type, Scope -> -1
+  | Scope, Type -> 1
+
+let compare_depr d1 d2 =
+  match (d1, d2) with
+  | None, None -> 0
+  | Some s1, Some s2 -> String.compare s1 s2
+  | None, Some _ -> -1
+  | Some _, None -> 1
```
ggreif, line 154:
```suggestion
  | _ -> 1
```

src/mo_types/type.ml
```diff
@@ -67,6 +67,205 @@ and kind =
...
+let compare_func_sort s1 s2 =
+  let d = tag_func_sort s1 - tag_func_sort s2 in
+  if d > 0 then 1 else if d < 0 then -1 else 0
+
+let compare_bind_sort s1 s2 =
+  match s1, s2 with
+  | Type, Type
+  | Scope, Scope -> 0
+  | Type, Scope -> -1
+  | Scope, Type -> 1
```
ggreif, line 147:
```suggestion
  | _ -> 1
```

ggreif (approved):
LGTM! Modulo my suggestions.

src/mo_types/type.ml
crusso: I doubt it makes much difference since it isn't a recursive entry point, but I can try.

src/mo_types/type.ml
crusso: I rather not add a default here in case we extend the type.

src/mo_types/type.ml
```diff
@@ -67,6 +67,205 @@ and kind =
...
+    (match compare_typ tb1.bound tb2.bound with
+     | 0 -> compare_bind_sort tb1.sort tb2.sort
+     | ord -> ord)
+  | ord ->  ord
+
+and compare_tbs tbs1 tbs2 =
+  match (tbs1, tbs2) with
+  | [], [] -> 0
+  | [], (_::_) -> -1
+  | (_::_, []) -> 1
```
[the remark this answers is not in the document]
crusso: Will do.

src/mo_types/type.ml
crusso: Couldn't really tell a difference but shorter code, so better.

src/mo_types/type.ml
```diff
@@ -67,6 +67,204 @@ and kind =
...
+  | Obj (s1, fs1), Obj (s2, fs2) ->
+    (match compare_obj_sort s1 s2 with
+     | 0 -> compare_flds fs1 fs2
+     | ord -> ord)
+  | Variant fs1, Variant fs2 ->
+    compare_flds fs1 fs2
+  | Mut t1, Mut t2 ->
+    compare_typ t1 t2
+  | Any, Any -> 0
+  | Non, Non -> 0
```
ggreif, line 199:
```suggestion
  | Any, Any
  | Non, Non
```

Issue #2348, opened 2021-02-14
Type representation design


nomeata:
In this issue I’d like to design our initial type representation (typerep) design.

## What’s this?

The (run-time) type representation is a value at run-time that indicates a type.

## What for, and why now?

We need (or can use this) for:

 * Currently, we do type-driven code generation in the IR-to-IR passes (`show.ml`) and the code generator (for serialization/deseriaizaion). This could be replaced with type-driven generation of a typrep value, and then interpreting that value in (static) rust code.
   The benefits are: 
     - Simpler code generation
     - We can write more code in Rust (simpler, easier to maintain, easier to do complex stuff)
     - (Likely) smaller code this way.

 * In particular, I don’t see how we can extend our existing approach to deserialization code generation to the new requirement of dynamic subtyping checks (https://github.com/dfinity/candid/pull/168); this really seems to require a typerep value (with identity).

 * A bit down the road, this is a requisite for Shared Generics (https://github.com/dfinity/motoko/issues/2096), and even if we don’t do that right away, it would be good to keep that use-case in mind in the design.

Because this seems to be a blocker for correct Candid deserialization, now seems to be the right time to tackele this. Nevertheless, it might be advisable to first migrate `debug_show`,  as it is simpler.

## Requirements

 * The typrep values should contain enough information to reconstruct the full type, including field names (if only for error messages), not just field hashes.
 * The typrep values need to be statically decomposable (e.g. the typrep for `?t` should contain the typrep for `t`). Statically meaning the typrep for `t` exists in memory already. This seems to be necessary for all kind of recursive algorithms (`show`), and it would be odd if they have to allocate. This likely rules out a format like our “type hash” (which otherwise is, in a way, a very compact representation of the full type).
 * The typrep values must be able to represent coinductive values, i.e. be cyclic. This is a fundamental difference to Motoko value values!
 * The typrep values must support some form of identity, or at least a conservative to way to identify them (conservative in the sense that different typreps of the same type may look different, as long as one doesn’t run into cycles). The pointer to the static memory location of the typrep might work here. This is needed for the caching that one has to do when implementing the Candid  subtyping check.
 * (soft goal) The typreps should be compact.
 * (soft goal) The strings the typrep (field names) should be deduplicated.
 * (soft goal) The typreps should be easy to parse/traverse
 * (soft goal) The typreps memory structure should be the representation of a suitable Rust enum/struct, for ease of writing the Rust code.

The following requirements come from the usecase of shared generics.
[...]

PR #3467, opened 2022-09-30
LANG-28: flesh out a concept for reflected type parameters

ggreif:
Researching a type reflection mechanism. JIRA: https://dfinity.atlassian.net/browse/LANG-28.

I expect that deconstruction will be rather easy, but construction will be very tricky (needs CPS?).


ggreif:
As of 14051b1a3b540314328819d723ebebb2d4cda206 the following abilities are present
- [ ] analysis side (deconstruct `A`)
  - [ ] primitive types
  - [ ] pairs (tuples?)
  - [ ] up-structuring (integrate `A` into something bigger)
  - [ ] variants
- [ ] synthesis side

ggreif:
see also https://github.com/dfinity/motoko/issues/2348

ggreif:
In today's _ICP.Lab_ Motoko office hours the following simplified context came up:
``` Motoko
func ws_send<to_candid A>(data : A)
```
would allow using `to_candid(data)` to obtain a `Blob`, where the caller would pass in a conversion (serialisation) function implicitly.
Similarly
``` Motoko
func ws_receive<from_candid A>(blob : Blob) : A { let ?(a : A) = from_candid(blob) else trap(); a }
```
would work for deserialising.
Stacking would make sense too `<switch from_candid to_candid A>` etc. (not pretty though...)

luc-blaeser:

ggreif:
Here are a few new ideas I had recently:
Declare parametric functions like this:
``` Motoko
type Const<T> = T; // just to make it more complicated
func foo<T>(arg : Const<T> with type) = switch arg with type {
  case 5 with type Nat: "It's a natural";
  case (?"Joker!") with type ?Text: "text-option"
};
```
Maybe have this shorthand (valid syntax already) expanding to `with type` in the above context:
```
  case (?"Joker!" : ?Text): "text-option"
```
For now I'd restrict type passing to `to_candid`-capable values, and would pass the Candid type representation behind the scenes suitably interpreting it in the `with type`-style `switch`.

rvanasa:
This would be highly useful for the base library, since we could simplify type conversions, equality, comparison, etc. Maybe we could find a way to extend this for record / object types with unknown fields for use cases such as (de)serialization.

PR #3478, opened 2022-10-08
Reset type name counter for subsequent `moc.js` function calls

rvanasa:
This PR makes progress towards fixing https://github.com/dfinity/node-motoko/issues/25.

Changes specific to this branch (which will be more evident after merging #3455):
- Empty the `Mo_types.Cons.stamps` counter before compiling via JS ([commit](https://github.com/dfinity/motoko/pull/3478/commits/ee3a40ca00564c4e9f62db67f778801cfd3e25be); fixes some Wasm differences but introduces others)
- Replace with backing up and restoring the type stamps to account for any global types ([commit](https://github.com/dfinity/motoko/pull/3478/commits/761200692d92b8c964fffaa196e1ecf748d1e980); same result)

@chenyan-dfinity, are there any subtleties that might be causing this to work differently than expected?


chenyan-dfinity (commented):
LGTM. I will defer to @crusso for a second look.

My only concern is that is it possible to get different type con names in `js_check` and `js_compile`? My guess is no, but I'm not 100% sure.

src/mo_types/cons.ml
```diff
@@ -17,6 +17,12 @@ module Stamps = Env.Make(String)
 
 let stamps : int Stamps.t ref = ref Stamps.empty
 
+let session f =
+  let original = !stamps in
+  let result = f () in
+  stamps := original;
+  result
```
crusso, line 24:
```suggestion
let session f =
  let original = !stamps in
  try let result = f () in
       stamps := original;
       result
  match e -> begin
     stamps := original;
     raise e
  end
```
Looks good but probably worth protecting the exception exit too, with something like the above.

crusso (commented):
Sorry, I never noticed the mention - best to ping me on slack.

rvanasa:
This should be ready to merge as a step towards Wasm reproducibility.

crusso (approved):
I think this is good for now. The other thing which might be affecting reproducibility is our IR variable name generator. To fix that, we'd probably want to introduce a more general Session module and have  stateful components register themselves (using save/restore callbacks) with that.

rvanasa:
For sure; introducing a Session module seems like a good general solution.

PR #3486, opened 2022-10-14
Support generation of __init__ methods based on actor private fields

aterga:
This PR extends the Motoko to Viper translation with initialization statements for each actor private field. These statements form the body of a special `__init__` method (that establishes the canister's invariant for the first time). 

Main assumptions:
* Initializing expressions can be translated as expressions; no imperative code (a.k.a. Viper statements) needs to be used
* The name `__init__` is not used elsewhere in the program
* The method `__init__` should be supplied with permissions to write into each actor private field (alternatively, the translation would need to be extended with `inhale` statements corresponding to each field)


src/viper/trans.ml
```diff
@@ -41,15 +63,16 @@ let rec unit (u : Mo_def.Syntax.comp_unit) : prog =
 and dec_fields (ctxt : ctxt) (ds : M.dec_field list) =
   match ds with
   | [] ->
-    (ctxt, [])
+    (ctxt, [], [])
   | d :: ds ->
-    let ctxt, mk_i = dec_field ctxt d in
-    let ctxt, mk_is = dec_fields ctxt ds in
-    (ctxt, mk_i::mk_is)
+    let ctxt, init, mk_i = dec_field ctxt d in
```
ggreif, line 68: `ctxt` gets shadowed below. Use `_`?

src/viper/trans.ml
```diff
@@ -41,15 +63,16 @@ let rec unit (u : Mo_def.Syntax.comp_unit) : prog =
...
+    let ctxt, inits, mk_is = dec_fields ctxt ds in
+    (ctxt, (match init with Some i -> i::inits | _ -> inits), mk_i::mk_is)
 
 and dec_field ctxt d =
-  let (ctxt, mk_i) = dec_field' ctxt d.it in
+  let (ctxt, init, mk_i) = dec_field' ctxt d.it in
```
ggreif, line 73:
```suggestion
  let ctxt, init, mk_i = dec_field' ctxt d.it in
```

src/viper/trans.ml
```diff
@@ -30,8 +30,30 @@ let rec unit (u : Mo_def.Syntax.comp_unit) : prog =
   match body.it with
   | M.ActorU(id_opt, decs) ->
      let ctxt = { self = None; ids = Env.empty } in
-     let (ctxt', mk_is) = dec_fields ctxt decs in
-     let is = List.map (fun mk_i -> mk_i ctxt') mk_is in
+     let ctxt', inits, mk_is = dec_fields ctxt decs in
+     let is' = List.map (fun mk_i -> mk_i ctxt') mk_is in
+
+     let init_id = {it = "__init__"; at = Source.no_region; note = NoInfo } in
+     let self_id = {it = "$Self"; at = Source.no_region; note = NoInfo } in
```
ggreif, line 37: isn't `id` the helper for this? And then inline the two values?

src/viper/trans.ml
```diff
@@ -30,8 +30,30 @@ let rec unit (u : Mo_def.Syntax.comp_unit) : prog =
...
+     let ctxt'' = { ctxt' with self = Some self_id.it } in
+     let init_list = (List.map (fun (id, init) -> {
+      at = init.at;
+      it = FieldAssignS((self ctxt'' init.at, id), exp ctxt'' init);
+      note = NoInfo
+    }) inits)
```
ggreif, line 43:
```suggestion
     let init_list = List.map (fun (id, init) -> {
      at = init.at;
      it = FieldAssignS((self ctxt'' init.at, id), exp ctxt'' init;
      note = NoInfo
    }) inits)
```

src/viper/trans.ml
```diff
@@ -30,8 +30,30 @@ let rec unit (u : Mo_def.Syntax.comp_unit) : prog =
...
+    in
+     let init_body = {
+      at = body.at;
+      it = [], init_list;
+      note = NoInfo
+     }
+     in
```
ggreif, line 50: join with line 49

src/viper/trans.ml
ggreif: Okay, scratch that, let's intro an `id_at` for this, so that we can supply a region, use from line 98 too.

src/viper/trans.ml
```diff
@@ -30,8 +30,30 @@ let rec unit (u : Mo_def.Syntax.comp_unit) : prog =
...
+     let m = {
+      it = MethodI(init_id, [self_id, {it = RefT; at = Source.no_region; note = NoInfo}], [], [], [], Some init_body);
```
ggreif, line 52:
```suggestion
      it = MethodI(init_id, [self_id, {it = RefT; at = self_id.at; note = NoInfo}], [], [], [], Some init_body);
```

src/viper/trans.ml
```diff
@@ -30,8 +30,30 @@ let rec unit (u : Mo_def.Syntax.comp_unit) : prog =
...
-     let (ctxt', mk_is) = dec_fields ctxt decs in
-     let is = List.map (fun mk_i -> mk_i ctxt') mk_is in
+     let ctxt', inits, mk_is = dec_fields ctxt decs in
+     let is' = List.map (fun mk_i -> mk_i ctxt') mk_is in
+
+     let init_id = {it = "__init__"; at = Source.no_region; note = NoInfo } in
+     let self_id = {it = "$Self"; at = Source.no_region; note = NoInfo } in
+     let ctxt'' = { ctxt' with self = Some self_id.it } in
+     let init_list = (List.map (fun (id, init) -> {
+      at = init.at;
```
ggreif, line 40: Maybe join to an initialiser and have `{ left = id.at.left; right = init.at.right }`?

ggreif (commented):
Looks fine so far. Please check `ctxt` usage.

src/viper/trans.ml
aterga: I think @crusso used meant to use shadowing in this case, right?
As in, we first transform `ctxt` via `let ctxt, init, mk_i = dec_field ctxt d` and then via `let ctxt, inits, mk_is = dec_fields ctxt ds`.

src/viper/trans.ml
aterga: Could you clarify what you mean, please?

src/viper/trans.ml
ggreif: I am fine with it when it is intentional :-)

src/viper/trans.ml
ggreif: To me it looks like the `FieldAssignS` inherits the source range from the `init` (value). I am suggesting to give it the source range from the left of the identifier (field name) until the right of the `init` value.

ggreif:
```suggestion
      at = { left = id.at.left; right = init.at.right };
```

aterga: Thanks for the clarification. I didn't know we already supported `left` and `right` in source positions. This makes sense!

aterga: Yep, it compiles. Thanks!

ggreif (commented):
The merge conflict is easy to resolve. Happy to do it...

ggreif (approved):
LGTM!

ggreif:
> * The method `__init__` should be supplied with permissions to write into each actor private field (alternatively, the translation would need to be extended with `inhale` statements corresponding to each field)

Some of this will be done with https://github.com/dfinity/motoko/pull/3491.

PR #3487, opened 2022-10-14
viper: testing

ggreif:
Set up testing infrastructure for Viper
- moved tests to `test/viper`
- added `Makefile`
- augmented `run.sh`
- accepted initial output
- the `.vpr` files are also checked with `silicon` backend

Run these by `make -C test/viper`
but this is not (yet) added to the outer `Makefile`.


default.nix
```diff
@@ -780,6 +780,7 @@ rec {
           nixpkgs.nix-update
           nixpkgs.rlwrap # for `rlwrap moc`
           nixpkgs.difftastic
+          nixpkgs.openjdk nixpkgs.z3 nixpkgs.jq # for viper dev
```
ggreif, line 783: back this out for now?

aterga: FYI, these dependencies are needed not only for "viper dev" but also for using viper (e.g., as part of the IDE extension). 

There's probably no ready-to-use Nix package for ViperServer itself. What would it take to create one? (Not urgent)

ggreif: See line 753 :-)

PR #3488, opened 2022-10-14
viper: invariants

ggreif:
This implements actor-global invariants. They get (currently) declared with a actor-level `assert` and translated to a Viper macro, that gets both `require`d and `ensure`d from each method by calls to those macros.

TODOs:
- [ ] find better syntax
- [x] use invariant(s) for pre/postconditions on each actor method
- [x] `self` access in `define`
- [x] `__init__` should only have a postcondition (depends on #3486)
- [x] allow more than one invariant


mraszyk:
Looks good to me so far, but

> use invariant(s) for pre/postconditions on each actor method

this seems to be missing.

ggreif:
> this seems to be missing.

Yes, that is the TODO list :-)
I am far from done

ggreif:
So regarding global invariants, we need a transformation pass that adorns the corresponding `requires` and `ensures` onto each method, right?

PR #3491, opened 2022-10-16
viper: adorn invariants with permissions

ggreif:
EXPERIMENTAL, but hey, it's a start!

TODO:
- [ ] `pp_perm`
- [ ] remove duplicates (optimise conjunction expressions)
- [x] harvest `perm`s from statement lists (`write` and _read_)
  - [x] apply to methods
  - [x] apply to `__init__`
- [ ] deduplicate perms and eliminate the trailing `true`.

### DIFF
--- a/src/viper/syntax.ml
+++ b/src/viper/syntax.ml
@@ -49,7 +49,7 @@ and perm = (perm', info) Source.annotated_phrase
 
 and perm' =
   | WildcardP
-  | FullP
+  | FullP of fldacc
   | NoP
   | EpsilonP
 (* | FractionalP of exp * exp | ...*)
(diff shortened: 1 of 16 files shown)


src/viper/syntax.ml
```diff
@@ -49,7 +49,7 @@ and perm = (perm', info) Source.annotated_phrase
 
 and perm' =
   | WildcardP
-  | FullP
+  | FullP of fldacc
```
ggreif, line 52: What is `FullP`? :-)

aterga: So, none of the named values need to carry any additional parameters, whereas we do need to add an arbitrary permission node that would carry two parameters, e.g., the `p` and `q` in `acc(x.f, p/q)`.

aterga:
I created another PR that implement this PR's functionality (without permission inference) https://github.com/dfinity/motoko/pull/3496

aterga:
This PR might be used later for extracting fine-grained permissions from the code.

PR #3495, opened 2022-10-18
Generational GC

luc-blaeser:
# Generational GC

Generational garbage collector, based on two generations:

* Young generation: Objects that have been created since the last GC run.
* Old generation: Objects that exist since more than one GC run.

## Design

Generational garbage collections optimizes towards fast reclamation of short-lived objects (generational hypothesis). 

New objects allocated in the young generation are more frequently collected than the older objects having already survived a GC run. 

* **Young collection (fast, frequent)**: 
  Only the young generation is marked and compacted.
  For this purpose, an additional root set has to be considered, namely pointers installed in the older generation that lead to the young generation.
  A write barrier tracks such extra root pointers and records them in a remembered set.
  Alive objects are compacted and appended to the end of the old generation.
  Eventually, a new empty young generation is initiated.

* **Full collection (slow, seldom)**:
  Collecting the old generation always includes the young generation.
  It is equivalent to a full mark-and-compact GC run.

The generations are adjacent logical partitions in the heap, separated by a pointer (`LAST_HP`, the free pointer of the last GC run). 

***Limitation***:

The GC is still non-incremental ("stop the world"), blocking the mutator during the short young generation collection and during the long full heap collection.
Therefore, programs managed by the generational GC may still exceed the message instruction limit for large heaps, in particular during the full heap collection.

## Components

* ***Write barrier***: Based on [this previous PR](https://github.com/dfinity/motoko/pull/2702), however, adjusted to post-update barriers to examine the new pointer locations. The barrier is only instrumented by the compiler if the generational GC is activated.
* ***Remembered set***: A hash table recording pointer changes of the mutator. Collisions are handled by linked lists. The hash table doubles when occupation exceeds 65%. 
* ***Threaded compaction***: Based on the existing compacting GC, adjusted for also allowing selective compaction of the young generation.
* ***Mark bitmap***: Same as in the existing compacting GC.

## Scheduling

1. Young collection is always triggered when it is larger than 8 MB.
2. Full collection is only triggered when the heap is nearly full or the old generation has grown by a factor of 2 or more.

The scheduling heuristics can be easily adjusted in `decide_strategy()` and `update_strategy()` in `generational.rs`.

## Performance

According to the GC benchmark with `dfx` 0.12.0, the generational GC seems to be generally more efficient than the existing compacting and copying GCs.

The release build and the benchmark results below include additional assertion checks that will be removed later in a future release step (see "Release Plan" section below).

**Note**: In future, performance will further improve by **3.6%** with regard to the total instructions and also to other runtime metrics (mutator utilization, longest GC pause) when the additional assertion checks are removed.

### Total Instructions

| **Generational**  | **2.61e10**             | 
| Compacting        | 3.44e10                 | 
| Copying           | 3.11e10                 | 

***16% faster than the copying GC and 24% faster than the compacting GC on average.***

The worst result per benchmark case is a 25% degrade for `linked-list` compared to the copying GC.

### Mutator Utilization

| **Generational**  | **74.0%**                |
| Compacting        | 61.9%                    |
| Copying           | 64.3%                    |

9.7% higher than the copying GC and 12% higher than the compacting GC.

Always better, except for some negligible degrades for `cancan`, `qr-code`, and `sha256`.

### Write Barrier Overhead

Measured as the total number of mutator instructions, average across all benchmark cases:

| GC                | Avg. Mutator Instructions |
| **Generational**  | **1.89e10**               |
| None              | 1.75e10                   |

***Moderate 8% average overhead caused by the write barrier.***

Depending on the benchmark case, the overhead varies between 0% and 52% (for `graph`). Higher barrier costs do generally not yet imply a total runtime slowdown, e.g. `graph` with generational GC is still faster than with the compacting GC.

### Longest GC Pause

Longest GC pause, benchmark average:

| GC                | Longest GC Pause on Avg.  |
| **Generational**  | **9.65e8**                |
| Compacting        | 11.4e8                    |
| Copying           | 9.61e8                    |

Similar like the copying GC, 15% shorter than the compacting GC on average.

### Memory Size

Allocated wasm memory space, benchmark average:

| **Generational**  | **234 MB**              |
| Compacting        | 235 MB                  |
| Copying           | 305 MB                  |

Similar like the compacting GC, 23% less than the copying GC.

### Final Heap Size

Heap size at the end of each benchmark case, average of all cases:

| GC                | Avg. Final Heap Size    |
| **Generational**  | **151 MB**              |
| Compacting        | 152 MB                  |
| Copying           | 152 MB                  |

Nearly equal.

### Allocation Limit

Average amount of allocations for the benchmark limit cases, until reaching a limit (instruction limit or heap limit).

| **Generational**  | **16e6**                |
| Compacting        | 14e6                    |
| Copying           | 15e6                    |

Nearly equal.

Average size of allocatable heap space for the benchmark limit cases.

| **Generational**  | **949 MB**              |
| Compacting        | 715 MB                  |
| Copying           | 866 MB                  |

Similar. Slightly higher with the generational GC.

**Note**: As the GC is not incremental, the limit is still exceeded for a large number of objects.

## Testing

1. RTS unit tests

    In Motoko repo folder `rts`:
    ```
    make test
    ```

2. Motoko test cases

    ```
    export EXTRA_MOC_ARGS="--sanity-checks --generational-gc --force-gc"
    make
    ```

3. GC Benchmark cases

    In `gcbench` repo: 
    ```
    ./measure-all.sh
    ```

4. Tests with sanity checks

    The debug RTS build contains the following extra checks:
    * Heap sanity check, traversing all pointers in the static root set and in heap objects, analyzing whether all pointers refer to a plausible target (valid tag).
    * Barrier coverage check, based on [this previous PR](https://github.com/dfinity/motoko/pull/2702), detecting pointer modifications through memory snapshot comparisons and verifying that all pointer changes are also contained in the remembered structure.
    * **Extra barrier coverage check**: On full collection, checking that all locations with all old-to-young generation pointers are contained in the remembered set that is recorded by the write barrier.

    Debug checks are enabled by the `--sanity-checks` compile flag.

    Please note that several benchmark cases cannot complete with debug mode as they reach the message instruction limit, which is an expected behavior.
    Nevertheless, testing with this mode as much as possible is useful to verify that the debug assertions and sanity checks hold.

## Possible Future Optimizations

* ***Static barrier elimination***: Based on a conservative static analysis, the compiler could eliminate a write barrier if the assigned value is certainly not a pointer. 

    This has already been implemented in the [write barrier for the future incremental GC](https://github.com/dfinity/motoko/pull/3502).
    
* ***Three generations***: Introducing a third medium-age generation (between young and old) might further improve the average GC performance.

## Explored Design Alternative

* ***Remembered log***: Instead of a hash set, the write barrier stores relevant pointer locations in a remembered log, a simple list of log tables. While this performs slightly better (1%) than remembered set in the GC benchmark, it may run into performance problems for cases with lots of duplicate pointer changes. This is why remembered set is preferred.

  Please see the [specific remembered log branch](https://github.com/dfinity/motoko/tree/luc/measure-log-generational-gc).

  ***Two-phase GC***: The generational GC can combine marking and backwards threading in one case. However, the GC then has to perform unthreading even if there is no or only very few garbage. The latter optimization (skipping GC after marking when there is few garbage) performs better by around 3%.

  Please see the [specific two-phases branch](https://github.com/dfinity/motoko/tree/luc/generational-gc-two-phases).

## Release Plan

As a safety measure, assertions are initially retained in production (release mode). This benchmark above already includes this overhead of around 3.6%. These checks will be removed in future when GC stability has been proven in practice.

1. **Beta Test Phase**: Explicit option for using the generational GC. Assertions checks are turned on even in release mode. Ask community users to beta test this GC in practice.

    ```
    {
      "canisters": {
        "mycanister": {
          "main": "myprogram.mo",
          "type": "motoko",
          "args" : "--generational-gc"
        }
      },
      "version": 1
    }
    ```

2. **Release with Checks**: Make the generational GC the default GC. Keep the assertion checks on in the productive mode.

3. **Final Release**: Change to debug assertions, which will **further improve the performance of the generational GC by 3.6%** on average.


crusso:
Good stuff!

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,501 @@
...
+    } else {
+        None
+    }
+}
+
+#[cfg(feature = "ic")]
+unsafe fn update_strategy(strategy: Strategy, limits: &Limits) {
+    const GROWTH_RATE: f64 = 1.5;
+    if strategy == Strategy::Full {
+        OLD_GENERATION_THRESHOLD = (limits.free as f64 * GROWTH_RATE) as usize;
```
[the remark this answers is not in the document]
luc-blaeser: Interesting question! I suppose the question is about the conversion of f64 back to usize that this does not wrap? 

With Rust, following IEEE 754 floating point numbers, it should take the nearest unsigned integer, no wrapping. Therefore, even if the float (f64) has become greater than usize::MAX by the multiplication, it should be usize::MAX. 

(On future 64-bit addresses, there may occur imprecision with the fraction being smaller than 64 bits. However, contrary to wrapping, the imprecision should be no problem. The threshold is then just rounded to a nearest representable value.)

rts/motoko-rts/src/gc/generational/sanity_checks.rs
```diff
@@ -0,0 +1,183 @@
+//! Extensive sanity checks for geberational GC features.
```
crusso, line 1:
```suggestion
//! Extensive sanity checks for generational GC features.
```

rts/motoko-rts/src/gc/generational/sanity_checks.rs
```diff
@@ -0,0 +1,183 @@
...
+use super::{Heap, Limits, Roots};
+use crate::mem_utils::memcpy_bytes;
+use crate::memory::{alloc_blob, Memory};
+use crate::types::*;
+use crate::visitor::{pointer_to_dynamic_heap, visit_pointer_fields};
+
+static mut SNAPSHOT: *mut Blob = null_mut();
+
+/// Take a memory snapshot. To be initiated after GC run.
+pub unsafe fn take_snapshot<M: Memory>(heap: &mut Heap<M>) {
```
crusso, line 18: Not sure this is worth it, but I guess you could consider snapshoting to stable memory which might work with larger heaps without exhausting main memory. The verification would be much slower though, unless you do some caching of stable memory pages during verification.

crusso: Maybe not worth the complexity...

rts/motoko-rts/src/gc/generational/sanity_checks.rs
```diff
@@ -0,0 +1,183 @@
...
+            if object.tag() != TAG_ONE_WORD_FILLER {
+                self.check_object(object);
+            }
+            pointer += object_size(pointer as usize).to_bytes().as_usize();
+        }
+    }
+
+    unsafe fn is_ptr(value: Value) -> bool {
+        const TRUE_VALUE: u32 = 1;
+        value.is_ptr() && value.get_raw() != TRUE_VALUE
```
crusso, line 181: Should value.is_ptr() not filter out TRUE_VALUE itself?

crusso (commented):
Some interim comments - will resume soon(ish). Lovely code.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,490 @@
...
+            self,
+            object,
+            object_tag,
+            self.heap.limits.base,
+            |gc, field_address| {
+                let field_value = *field_address;
+                gc.mark_object(field_value);
+            },
+            |gc, slice_start, array| {
+                const SLICE_INCREMENT: u32 = 127;
```
crusso, line 294: Maybe we should define SLICE_INCREMENT more prominently somewhere

rts/motoko-rts/src/gc/generational/sanity_checks.rs
luc-blaeser: This is an interesting idea. As it is only for sanity checks, I may try this when there is need for for memory. The snapshot check is anyway very expensive in terms of runtime, and therefore probably only used for very specific tests in the longer term. Probably it will be worse with stable memory.

rts/motoko-rts/src/gc/generational/sanity_checks.rs
luc-blaeser: I moved this logic. First, I was unsure what impact it has on the remaining code (also performance-wise). I checked this and it seems to be good.

rts/motoko-rts/src/gc/generational.rs
luc-blaeser: The constant is not really a relevant one, but only for scalability tuning inside the GC, such that the breadth-first search does not copy too many element pointers in the mark stack at once. I copied that logic from the existing compacting GC. The constant could be locally changed without any impact on the remaining logic or other GCs.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,490 @@
...
+
+    pub unsafe fn run(&mut self) {
+        assert_eq!(self.heap.limits.base % 32, 0);
+        let mem_size = Bytes(self.heap.limits.free as u32 - self.heap.limits.base as u32);
+        alloc_bitmap(
+            self.heap.mem,
+            mem_size,
+            self.heap.limits.base as u32 / WORD_SIZE,
+        );
+        alloc_mark_stack(self.heap.mem);
```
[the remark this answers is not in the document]
luc-blaeser: This would save free reserve for young collection. It will not improve runtime performance, as there is already the `advance()` function in bitmap that is used by the generational GC. And for full collection, it needs this space anyway. I will try this out.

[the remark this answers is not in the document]
crusso: No need to try it out if you don't think it's useful. I was just curious. I'll defer to your judgement on all things GC, tbh.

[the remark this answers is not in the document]
crusso: Doesn't the bitmap need to be zeroed before use though? That's 256K= 8MiB /4/8 minor heap vs 128 MiB = 4GB/4/8 for a worst case full collection.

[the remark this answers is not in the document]
crusso: Maybe best to try out as a stacked PR.... easier to review in isolation. But up to you.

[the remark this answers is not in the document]
crusso: Given the current bitmap implementation with the forbidden zone, it might actually be trivial to try out (IIUC): just consider the major heap as part of the static heap when you set up the bitmap.

rts/motoko-rts/src/gc/generational.rs
crusso: Ok

rts/motoko-rts/src/gc/generational.rs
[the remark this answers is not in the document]
luc-blaeser: I implemented this. It was indeed a trivial change. Unfortunately, there is no noticeable performance gain. (I used the mark bitmap for an extra sanity check on threading in the old generation to detect multiple threading attempts, but removed this logic now as with remembered sets there should be no duplicates anyway.)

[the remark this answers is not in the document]
crusso: That's a shame, sorry for the extra work. I thought this would have showed up as a reduced instruction count for minor gcs.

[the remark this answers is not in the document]
luc-blaeser: No worries, the change is quite simple and interesting. Probably, it could be beneficial once we have more test cases with larger heaps in the benchmark (with deterministic time slicing).

default.nix
```diff
@@ -478,9 +483,10 @@ rec {
       run-dbg    = snty_subdir "run"        [ moc ] ;
       ic-ref-run = test_subdir "run-drun"   [ moc ic-ref-run ];
       ic-ref-run-compacting-gc = compacting_gc_subdir "run-drun" [ moc ic-ref-run ] ;
+      ic-ref-run-generational-gc = generational_gc_subdir "run-drun" [ moc ic-ref-run ] ;
       drun       = test_subdir "run-drun"   [ moc nixpkgs.drun ];
       drun-dbg   = snty_subdir "run-drun"   [ moc nixpkgs.drun ];
-      drun-compacting-gc = compacting_gc_subdir "run-drun" [ moc nixpkgs.drun ] ;
+      drun-generational-gc = generational_gc_subdir "run-drun" [ moc nixpkgs.drun ] ;
```
crusso, line 489:
```suggestion
      drun-compacting-gc = compacting_gc_subdir "run-drun" [ moc nixpkgs.drun ] ;
      drun-generational-gc = generational_gc_subdir "run-drun" [ moc nixpkgs.drun ] ;
```
It think we still want to rest the compacting_gc, right?

luc-blaeser: Yes, thanks a lot, I will add this

rts/motoko-rts/src/gc/generational/sanity_checks.rs
luc-blaeser: I optimized this in [6fc0079](https://github.com/dfinity/motoko/pull/3495/commits/6fc0079c5696b287e538026cbd2f399c4d379ce9). A similar optimization was done in compacting GC for comparing the field addresses. So, performance is as good as it was before this change.

rts/motoko-rts/src/gc/generational/remembered_set.rs
```diff
@@ -0,0 +1,282 @@
...
+            count: 0,
+            fast_cache: null_ptr_value(),
+        }
+    }
+
+    pub unsafe fn insert<M: Memory>(&mut self, mem: &mut M, value: Value) {
+        if is_null_ptr_value(value) || value.get_raw() == self.fast_cache.get_raw() {
+            return;
+        }
+        let index = self.hash_index(value);
```
crusso, line 85: Shouldn't you update fast_cache here? Or is it set elsewhere?

crusso: It looks to me like fast_cache is only ever set to null_ptr_value() (at line 77)

crusso (commented):
Add interim comments (notably fast_cache looks broken)

rts/motoko-rts/src/gc/generational/remembered_set.rs
luc-blaeser: Thanks a lot again. Indeed, never updated. I will fix this asap. Maybe, performance improves. 
(Btw: This bug probably happened while changing from remembered log to remembered set.)

rts/motoko-rts/src/gc/generational/remembered_set.rs
```diff
@@ -82,6 +82,8 @@ impl RememberedSet {
         if is_null_ptr_value(value) || value.get_raw() == self.fast_cache.get_raw() {
             return;
         }
+        self.fast_cache = Value::from_raw(value.get_raw());
```
crusso, line 85:
```suggestion
        self.fast_cache = value;
```
Or are the conversions doing something?

luc-blaeser: I removed the cache as there is no relevant performance gain. It was useful for the remembered log (to eliminate a lot of duplicates)

ulan (commented):
Very clean implementation, @luc-blaeser!

I need a couple of days for detailed review. Please feel free to ack and ignore the comments marked as "nits" because I don't have strong opinion there.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,499 @@
...
+    } else {
+        None
+    }
+}
+
+#[cfg(feature = "ic")]
+unsafe fn update_strategy(strategy: Strategy, limits: &Limits) {
+    const GROWTH_RATE: f64 = 1.5;
+    if strategy == Strategy::Full {
+        OLD_GENERATION_THRESHOLD = (limits.free as f64 * GROWTH_RATE) as usize;
```
crusso, line 140:
```suggestion
        OLD_GENERATION_THRESHOLD = (limits.free - limits.base) as f64 * GROWTH_RATE) as usize;
```
It looks like this is a threshold on  ``old_generation_size`, given the comparison in `decide_strategy` on line 127. So I guess you should be subtracting limits.base (or use a common helper)

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,499 @@
...
+            return;
+        }
+        set_bit(obj_idx);
+
+        push_mark_stack(self.heap.mem, pointer as usize, object.tag());
+        self.marked_space += object_size(pointer as usize).to_bytes().as_usize();
+    }
+
+    unsafe fn mark_all_reachable(&mut self) {
+        while let Some((obj, tag)) = pop_mark_stack() {
```
crusso, line 263: I expect its still safe to recover the tag from the object itself here. I guess a simple assert could verify that claim.

crusso (commented):
Add some interim comments, resuming review straight after.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,499 @@
...
+        }
+        debug_assert_eq!(pointer % WORD_SIZE, 0);
+
+        let obj_idx = pointer / WORD_SIZE;
+        if get_bit(obj_idx) {
+            return;
+        }
+        set_bit(obj_idx);
+
+        push_mark_stack(self.heap.mem, pointer as usize, object.tag());
```
crusso, line 258: Strange. I though I commented here but it seems to have been lost.

For the generational gc , I wonder if you can use a simpler mark stack that doesn't store the original object tags. I think Omer only pushed the tags on the stack because threading during marking would trash them, but you now do that after marking, right? I think it's very small variation of markstack.rs.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,499 @@
...
+                #[cfg(debug_assertions)]
+                gc.barrier_coverage_check(field_address);
+            },
+            |gc, slice_start, array| {
+                const SLICE_INCREMENT: u32 = 127;
+                debug_assert!(SLICE_INCREMENT >= TAG_ARRAY_SLICE_MIN);
+                if array.len() - slice_start > SLICE_INCREMENT {
+                    let new_start = slice_start + SLICE_INCREMENT;
+                    // push an entire (suffix) array slice
+                    push_mark_stack(gc.heap.mem, array as usize, new_start);
```
crusso, line 287:
```
                    push_mark_stack(gc.heap.mem, array as usize, new_start);
```
Ah, maybe this prevents once from not pushing the tags onto the markstack..

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,499 @@
...
+            self.mark_fields(obj as *mut Obj, tag)
+        }
+    }
+
+    unsafe fn mark_fields(&mut self, object: *mut Obj, object_tag: Tag) {
+        visit_pointer_fields(
+            self,
+            object,
+            object_tag,
+            self.heap.limits.base,
```
crusso, line 273:
```suggestion
            self.heap.limits.base,
```
Couldn't this only be self.heap.limits.last_free for a young GC? Or am i confused. I think that cause the visitor to skip values older than this argument.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,499 @@
...
+        }
+    }
+
+    unsafe fn thread_backward_pointer_fields(&mut self, object: *mut Obj) {
+        debug_assert!(object.tag() < TAG_ARRAY_SLICE_MIN);
+        visit_pointer_fields(
+            &mut (),
+            object,
+            object.tag(),
+            self.heap.limits.base,
```
crusso, line 401: Again, for a young colletion, can we restrict this to `heap.limits.last_free`?
Similar to previous comment.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,499 @@
...
+        self.heap.limits.free = free;
+    }
+
+    /// Thread forward pointers in object
+    unsafe fn thread_forward_pointers(&mut self, object: *mut Obj) {
+        visit_pointer_fields(
+            &mut (),
+            object,
+            object.tag(),
+            self.heap.limits.base,
```
crusso, line 469: Again, for a young collection, can we restrict this to `heap.limits.last_free`?
Similar to previous comment.

rts/motoko-rts/src/gc/generational.rs
crusso: (modulo the special treatment of arrray)

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -157,6 +158,21 @@ pub unsafe fn iter_bits() -> BitmapIter {
 pub const BITMAP_ITER_END: u32 = u32::MAX;
 
 impl BitmapIter {
+    pub unsafe fn advance(&mut self, address: u32) {
```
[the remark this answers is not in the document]
luc-blaeser: Thank you very much. Very good point. I removed the bitmap as it is actually no longer really since I reduced the bitmap size for the young generation collection (as suggested by Claudio).

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -157,6 +158,21 @@ pub unsafe fn iter_bits() -> BitmapIter {
...
+        let base = get_bitmap_forbidden_size() as u32 * 8;
+        debug_assert!(base % 8 == 0);
+        debug_assert!(self.current_bit_idx >= base);
+        let word_idx =
+            get_bitmap_forbidden_size() + (self.current_bit_idx - base) as usize / 64 * 8;
+        self.current_word = *(BITMAP_FORBIDDEN_PTR.add(word_idx) as *const u64);
+        self.leading_zeros = self.current_word.leading_zeros();
+        let bit_offset = (self.current_bit_idx - base) % 64;
+        self.current_word >>= bit_offset;
+        self.leading_zeros = core::cmp::min(self.leading_zeros, 64 - bit_offset);
```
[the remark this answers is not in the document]
luc-blaeser: This logic is indeed very complicated. It is good that we can remove this logic,as it is no longer really relevant. 
(Just for interest, and if I am not mistaken: The leading zeros count how much the iterator can advance to complete the word, if there are no further set bits in that word. The shift by `bit_offset` means that we have visited that many bits in the words, such that the remainder of bits to visit is `64-bit_offset`. The minimum handles the case that there would be more set bits in the remainder.)

rts/motoko-rts/src/gc/generational.rs
luc-blaeser: Thank you for this great suggestion. I simplified this. This is possible due to the separated mark and thread phases.

rts/motoko-rts/src/gc/generational.rs
luc-blaeser: The change was really simple. [7dd1787](https://github.com/dfinity/motoko/pull/3495/commits/7dd178754f73f21d364c56173feaa15ab557b54b)

rts/motoko-rts/src/gc/generational.rs
luc-blaeser: Indeed. I fixed this. Thank you.

rts/motoko-rts/src/gc/generational.rs
luc-blaeser: I changed it and now temporarily store the slice offset inside the object tag while traversing the array with multiple mark steps.

crusso: Excellent hack!

crusso: Although I guess that touches lots of pages in a full gc. But I guess we'll do that anyway.

rts/motoko-rts/src/gc/generational.rs
luc-blaeser: Yes, indeed. This simplifies things. I improved it in [fcbb1a4](https://github.com/dfinity/motoko/pull/3495/commits/fcbb1a434dac240daaac1e059804a3cfaa3f1459).

rts/motoko-rts/src/gc/generational.rs
luc-blaeser: Improved in [fcbb1a4](https://github.com/dfinity/motoko/pull/3495/commits/fcbb1a434dac240daaac1e059804a3cfaa3f1459).

rts/motoko-rts/src/gc/generational.rs
luc-blaeser: Also improved in [fcbb1a4](https://github.com/dfinity/motoko/pull/3495/commits/fcbb1a434dac240daaac1e059804a3cfaa3f1459).

rts/motoko-rts/src/gc/mark_compact/bitmap.rs
```diff
@@ -157,6 +158,21 @@ pub unsafe fn iter_bits() -> BitmapIter {
...
 
 impl BitmapIter {
+    pub unsafe fn advance(&mut self, address: u32) {
+        debug_assert!(address % WORD_SIZE == 0);
+        self.current_bit_idx = address / WORD_SIZE;
+        let base = get_bitmap_forbidden_size() as u32 * 8;
+        debug_assert!(base % 8 == 0);
+        debug_assert!(self.current_bit_idx >= base);
+        let word_idx =
+            get_bitmap_forbidden_size() + (self.current_bit_idx - base) as usize / 64 * 8;
```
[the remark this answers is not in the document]
luc-blaeser: I removed the function, thanks to your valuable suggestion.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,499 @@
...
+    fn generation_size(&self) -> usize {
+        let base = match self.strategy {
+            Strategy::Young => self.heap.limits.last_free,
+            Strategy::Full => self.heap.limits.base,
+        };
+        self.heap.limits.free - base
+    }
+
+    fn survival_rate(&self) -> f64 {
+        self.marked_space as f64 / self.generation_size() as f64
```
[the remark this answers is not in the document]
luc-blaeser: I think it should be safe because it is a floating point division, resulting in `f64::INF`. The function `is_compaction_beneficial()` would then evaluate to `false`. I will add a comment on that.

rts/motoko-rts/src/gc/generational/remembered_set.rs
```diff
@@ -0,0 +1,280 @@
...
+//! ----------------------------------
+//!
+//! Per collision a new linked list node is appended:
+//!
+//!                               Collision node
+//!                        ------------------------------
+//! prev_collision_ptr --> | entry | next_collision_ptr |
+//!                        ------------------------------
+//!
+//! Amortized hash-table growth when exceeding a defined threshold.
```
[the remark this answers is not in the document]
luc-blaeser: Good idea, I also thought of that. Fallback to full collection in the case of too many remembered set entries. I currently tested the case but because GC is anyway often running out of message limit, it was not (yet) a problem. Probably more a topic for later optimization once DTS is available and as a precaution for a probably relative pathological case. Another remedy would be to schedule young GC when it grows to big. But this would require call stack root set analysis. I keep it on my todo list for the moment.

rts/motoko-rts/src/continuation_table.rs
```diff
@@ -92,8 +93,13 @@ pub unsafe fn remember_continuation<M: Memory>(mem: &mut M, ptr: Value) -> u32 {
 
     let idx = FREE_SLOT;
 
-    FREE_SLOT = TABLE.as_array().get(idx).get_scalar();
-    TABLE.as_array().set(idx, ptr);
+    let table = TABLE.as_array();
+
+    FREE_SLOT = table.get(idx).get_scalar();
+
+    table.set(idx, ptr);
```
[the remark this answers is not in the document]
luc-blaeser: Excellent idea. I implemented this. This also simplifies code. Only, an array has a set function in the runtime system; object fields are only directly written by compiler-generated code.

rts/motoko-rts/src/gc/generational.rs
luc-blaeser: True. Especially threading changes a lot, and if we perform full heap compaction, probably, the whole heap is anyway modified (if there is garbage at the beginning of the heap).

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,499 @@
...
+                    (&self).thread(field_address)
+                }
+            },
+            |_, _, array| array.len(),
+        );
+    }
+
+    unsafe fn thread(&self, field: *mut Value) {
+        let pointed = (*field).get_ptr() as *mut Obj;
+        if self.should_be_threaded(pointed) {
```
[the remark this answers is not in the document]
luc-blaeser: Good idea. Also suggested by Claudio. I improved this.

ulan (approved):
Sorry for the delay. The implementation of the GC, write barrier, remembered set, and marking stack lgtm.

The compiler bits also lgtm, but I am deferring the thorough review of that to Claudio.

rts/motoko-rts/src/blob_iter.rs
```diff
@@ -16,8 +19,8 @@ unsafe fn blob_iter<M: crate::memory::Memory>(mem: &mut M, blob: Value) -> Value
     (*iter_array).header.tag = TAG_ARRAY;
     (*iter_array).len = 2;
 
-    iter_array.set(ITER_BLOB_IDX, blob);
-    iter_array.set(ITER_POS_IDX, Value::from_scalar(0));
+    iter_array.set(ITER_BLOB_IDX, blob, true, mem);
```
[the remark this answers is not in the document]
luc-blaeser: Thank you very much. This makes it more reliable, also for future changes. A little bit different to the enum, I now introduced two separate functions, one for pointer value writes that always implies the barrier (`set_pointer()`), and another for scalar value writes without a barrier (`set_scalar()`). Assertions additionally check that the right version is actually used.

rts/motoko-rts/src/continuation_table.rs
```diff
@@ -47,7 +47,7 @@ unsafe fn create_continuation_table<M: Memory>(mem: &mut M) {
 
     let table = TABLE.as_array();
     for i in 0..INITIAL_SIZE {
-        table.set(i, Value::from_scalar(i + 1));
+        table.set(i, Value::from_scalar(i + 1), false, mem);
```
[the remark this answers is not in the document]
luc-blaeser: Thank you. I actually implemented it this way: Instead of the enum, I introduced two functions `set_scalar()` and `set_pointer()`.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,503 @@
...
+    unsafe fn mark_root_mutbox_fields(&mut self, mutbox: *mut MutBox) {
+        let field_address = &mut (*mutbox).field;
+        if pointer_to_dynamic_heap(field_address, self.generation_base()) {
+            self.mark_object(*field_address);
+        }
+    }
+
+    unsafe fn should_be_threaded(&self, object: *mut Obj) -> bool {
+        let address = object as usize;
+        match self.strategy {
```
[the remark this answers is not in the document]
luc-blaeser: Thank you for the suggestion. This simplifies and reduces the code.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,503 @@
...
+
+    unsafe fn thread_root_mutbox_fields(&self, mutbox: *mut MutBox) {
+        let field_addr = &mut (*mutbox).field;
+        if pointer_to_dynamic_heap(field_addr, self.generation_base()) {
+            self.thread(field_addr);
+        }
+    }
+
+    unsafe fn thread_all_backward_pointers(&mut self) {
+        let mut bitmap_iter = iter_bits();
```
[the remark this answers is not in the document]
luc-blaeser: The bitmap is reused from the existing mark & compact GC. Although I agree with the better naming, I kept it in this PR to minimize changes to the other GC.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,503 @@
...
+    pub unsafe fn run(&mut self) {
+        assert_eq!(self.heap.limits.base % 32, 0);
+        self.alloc_mark_structures();
+        self.mark_phase();
+        self.compact_phase();
+        self.free_mark_structures();
+    }
+
+    unsafe fn alloc_mark_structures(&mut self) {
+        const BITMAP_ALGNMENT: usize = 8 * WORD_SIZE as usize;
```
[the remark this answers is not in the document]
luc-blaeser: Thank you.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,503 @@
...
+        let mut bitmap_iter = iter_bits();
+        if self.strategy == Strategy::Young {
+            free = self.heap.limits.last_free;
+        }
+        let mut bit = bitmap_iter.next();
+        while bit != BITMAP_ITER_END {
+            let old_pointer = (bit * WORD_SIZE) as *mut Obj;
+            let new_pointer = free;
+
+            // Update backwards references to the object's new location and restore object header
```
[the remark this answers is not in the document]
luc-blaeser: The threading is performed in two phases for the pointers inside heap objects: First, it only threads and unthreads backward pointers. Second, it focusses on forward pointers. Here, it only recovers the backward pointers. If both directions would be combined in one threading phase, it would be difficult to decide whether later visited pointers are already threaded or not.

rts/motoko-rts/src/gc/generational/remembered_set.rs
```diff
@@ -0,0 +1,280 @@
...
+    debug_assert!(
+        entry as u32 + size_of::<HashEntry>() as u32
+            <= table as u32 + object_size(table as usize).to_bytes().as_u32()
+    );
+    entry
+}
+
+unsafe fn table_set(table: *mut Blob, index: u32, value: Value) {
+    debug_assert!(table != null_mut());
+    let entry =
```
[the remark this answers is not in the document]
luc-blaeser: Thank you for this suggestion. This indeed removes code duplication.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,503 @@
...
+
+    write_barrier::init_write_barrier(gc.heap.mem);
+}
+
+#[cfg(feature = "ic")]
+unsafe fn get_limits() -> Limits {
+    use crate::memory::ic;
+    Limits {
+        base: ic::get_aligned_heap_base() as usize,
+        last_free: core::cmp::max(ic::LAST_HP as usize, ic::get_aligned_heap_base() as usize), // max because of aligned heap base
```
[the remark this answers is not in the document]
luc-blaeser: Yes, indeed. The maximum function is unnecessary. I removed it and added a debug assertion instead.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,503 @@
...
+    }
+}
+
+pub struct Heap<'a, M: Memory> {
+    pub mem: &'a mut M,
+    pub limits: Limits,
+    pub roots: Roots,
+}
+
+pub struct Roots {
```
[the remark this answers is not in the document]
luc-blaeser: Thank you for this suggestion. I will add a hint for potential future additional roots.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,503 @@
...
+    pub static_roots: Value,
+    pub continuation_table_ptr_loc: *mut Value,
+}
+
+pub struct Limits {
+    pub base: usize,
+    pub last_free: usize,
```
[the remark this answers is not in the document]
luc-blaeser: Yes, this is helpful comment. I add this.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,503 @@
...
+
+    // Thread forward pointers in old generation leading to young generation
+    unsafe fn thread_old_generation_pointers(&mut self) {
+        let mut iterator = REMEMBERED_SET.as_ref().unwrap().iterate();
+        while iterator.has_next() {
+            let location = iterator.current().get_raw() as *mut Value;
+            let value = *location;
+            // value in the location may have changed since recording by the write barrer
+            if value.points_to_or_beyond(self.heap.limits.last_free) {
+                debug_assert!((location as usize) >= self.heap.limits.base);
```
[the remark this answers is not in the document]
luc-blaeser: Thank you very much for this suggestion. This indeed makes it cleaner and the assertion stronger.

rts/motoko-rts/src/gc/generational/mark_stack.rs
```diff
@@ -0,0 +1,73 @@
...
+
+/// Pop a unskewed object pointer if existend to be visited next
+pub unsafe fn pop_mark_stack() -> Option<usize> {
+    if STACK_PTR == STACK_BASE {
+        return None;
+    }
+    STACK_PTR = STACK_PTR.sub(1);
+    let object = *STACK_PTR;
+    return Some(object);
+}
```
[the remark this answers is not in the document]
luc-blaeser: Thank you. I added this.

rts/motoko-rts/src/gc/generational/sanity_checks.rs
```diff
@@ -0,0 +1,167 @@
...
+
+/// Take a memory snapshot. To be initiated after GC run.
+pub unsafe fn take_snapshot<M: Memory>(heap: &mut Heap<M>) {
+    let length = Bytes(heap.limits.free as u32);
+    let blob = alloc_blob(heap.mem, length).get_ptr() as *mut Blob;
+    memcpy_bytes(blob.payload_addr() as usize, 0, length);
+    SNAPSHOT = blob;
+}
+
+/// Verify write barrier coverag by comparing the memory against the previous snapshot.
```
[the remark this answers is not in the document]
luc-blaeser: Thank you. I fixed it.

rts/motoko-rts/src/gc/generational/write_barrier.rs
```diff
@@ -0,0 +1,42 @@
...
+    LAST_HP = ic::LAST_HP;
+}
+
+/// Write barrier to be called AFTER the pointer store, used for generational GC.
+/// `location`: location of modified pointer (address of object field or array element).
+///
+/// As the barrier is called after the write, `*location` refers to the NEW value.
+/// No effect is the write barrier is deactivated.
+#[ic_mem_fn]
+pub unsafe fn write_barrier<M: Memory>(mem: &mut M, location: u32) {
```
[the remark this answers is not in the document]
luc-blaeser: This has indeed high optimization potential. I will measure this and update accordingly.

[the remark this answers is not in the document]
luc-blaeser: I have updated the function with the corresponding optimization. The check `location < LAST_HP` filters most of the barrier calls, then the pointer check and/or `value.points_to_or_beyond(LAST_HP)`.

src/codegen/compile.ml
```diff
@@ -8187,18 +8211,41 @@ let rec compile_lexp (env : E.t) ae lexp =
...
+     then
+      get_field ^^
+      compile_add_const ptr_unskew ^^
+      E.call_import env "rts" "write_barrier"
+     else G.nop
+    )
+   )
   | DotLE (e, n) ->
+    let (set_field, get_field) = new_local env "field" in
+   (
```
crusso, line 8234:
```suggestion
```

src/codegen/compile.ml
```diff
@@ -8187,18 +8211,41 @@ let rec compile_lexp (env : E.t) ae lexp =
...
-
+     store_ptr ^^
+     (if !Flags.gc_strategy = Mo_config.Flags.Generational
+      then  
+        get_field ^^
+        compile_add_const ptr_unskew ^^
+        E.call_import env "rts" "write_barrier"
+      else G.nop
+     )
+   )
```
crusso, line 8248:
```suggestion
```

src/codegen/compile.ml
```diff
@@ -9859,8 +9906,14 @@ and conclude_module env start_fi_o =
...
   let rts_start_fi = E.add_fun env "rts_start" (Func.of_body env [] [] (fun env1 ->
-    Bool.lit (!Flags.gc_strategy = Mo_config.Flags.MarkCompact) ^^
+    Bool.lit (!Flags.gc_strategy = Mo_config.Flags.MarkCompact || !Flags.gc_strategy = Mo_config.Flags.Generational) ^^
     E.call_import env "rts" "init" ^^
+    (if !Flags.gc_strategy = Mo_config.Flags.Generational
+     then
+      E.call_import env "rts" "init_write_barrier"
+     else 
+      G.nop 
+    ) ^^
```
crusso, line 9916:
```suggestion
      G.nop) ^^
```

src/codegen/compile.ml
```diff
@@ -8187,18 +8211,41 @@ let rec compile_lexp (env : E.t) ae lexp =
...
   | VarLE var -> Var.set_val env ae var
   | IdxLE (e1, e2) ->
-     compile_exp_vanilla env ae e1 ^^ (* offset to array *)
-     compile_exp_vanilla env ae e2 ^^ (* idx *)
-     Arr.idx_bigint env,
-     SR.Vanilla,
-     store_ptr
+    let (set_field, get_field) = new_local env "field" in
+    
+   (
```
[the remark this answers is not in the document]
luc-blaeser: Thank you. I improved this.

src/codegen/compile.ml
luc-blaeser: Thank you. I improved this.

src/codegen/compile.ml
```diff
@@ -8187,18 +8211,41 @@ let rec compile_lexp (env : E.t) ae lexp =
...
+    SR.Vanilla,
+    store_ptr ^^
+    (if !Flags.gc_strategy = Mo_config.Flags.Generational
+     then
+      get_field ^^
+      compile_add_const ptr_unskew ^^
+      E.call_import env "rts" "write_barrier"
+     else G.nop
+    )
+   )
```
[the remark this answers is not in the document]
luc-blaeser: Thank you. I improved this.

src/codegen/compile.ml
luc-blaeser: Thank you. I improved this.

Changelog.md
```diff
@@ -1,5 +1,27 @@
 # Motoko compiler changelog
```
crusso, line 2: Delete the ## 0.7.4 (tbd) header - we apply it just before making a release.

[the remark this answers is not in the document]
luc-blaeser: Thanks, I removed it.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,507 @@
...
+        static_roots: ic::get_static_roots(),
+        continuation_table_ptr_loc: crate::continuation_table::continuation_table_loc(),
+    };
+    let heap = Heap {
+        mem,
+        limits: get_limits(),
+        roots,
+    };
+    let strategy = decide_strategy(&heap.limits);
+    #[cfg(debug_assertions)]
```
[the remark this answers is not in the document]
luc-blaeser: It applies to the following statement `let forced_gc = strategy.is_none();` This is only used in debug mode. But probably, I should separate the code fragment with some blank lines, to make it clearer that only that code piece is debug mode logic.

The same applies to e.g.:
```
#[cfg(debug_assertions)]
sanity_checks::verify_snapshot(&gc.heap, false);
```

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,507 @@
...
+        self.alloc_mark_structures();
+        self.mark_phase();
+        self.compact_phase();
+        self.free_mark_structures();
+    }
+
+    unsafe fn alloc_mark_structures(&mut self) {
+        const BITMAP_ALIGNMENT: usize = 8 * WORD_SIZE as usize;
+        let mark_offset = match self.strategy {
+            Strategy::Young => self.heap.limits.last_free / BITMAP_ALIGNMENT * BITMAP_ALIGNMENT,
```
crusso, line 197: Curious - do we not have the invariant that last_free is word aligned?

luc-blaeser: Good question: It should be guaranteed because `Memory.allow_words()` is the primitive operation for heap allocations, and it uses words as the unit to allocate.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,507 @@
...
+            Strategy::Full => self.heap.limits.base,
+        };
+        let mem_size = Bytes((self.heap.limits.free - mark_offset) as u32);
```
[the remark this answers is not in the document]
luc-blaeser: Yes, indeed. This makes it more clearer. I will adjust this.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,509 @@
...
+                debug_assert!(SLICE_INCREMENT >= TAG_ARRAY_SLICE_MIN);
+                if array.len() - slice_start > SLICE_INCREMENT {
+                    let new_start = slice_start + SLICE_INCREMENT;
+                    // Remember to visit the array suffix later, store the next visit offset in the tag.
+                    (*array).header.tag = new_start;
+                    push_mark_stack(gc.heap.mem, array as usize);
+                    new_start
+                } else {
+                    // No further visits of this array. Restore the tag.
+                    (*array).header.tag = TAG_ARRAY;
```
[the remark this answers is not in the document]
luc-blaeser: This function only determines the array slice length.
The array elements are eventually marked in the preceding lambda function, with the same logic as for object fields:
```
|gc, field_address| {
    let field_value = *field_address;
    gc.mark_object(field_value);

     // Should become a debug assertion in future.
     gc.barrier_coverage_check(field_address);
},
```

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -0,0 +1,509 @@
...
+        if self.strategy == Strategy::Young {
+            self.mark_additional_young_root_set();
+        }
+    }
+
+    unsafe fn mark_static_roots(&mut self) {
+        let root_array = self.heap.roots.static_roots.as_array();
+        for i in 0..root_array.len() {
+            let object = root_array.get(i).as_obj();
+            assert_eq!(object.tag(), TAG_MUTBOX);
```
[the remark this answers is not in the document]
luc-blaeser: I additionally checked the example by a heap dump. The array is referenced by a mutbox that is again referenced by the static root set (which only refers to mutboxes).

rts/motoko-rts/src/gc/generational.rs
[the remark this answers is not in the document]
crusso: Thanks!

rts/motoko-rts/src/gc/generational.rs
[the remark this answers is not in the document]
crusso: yep!

rts/motoko-rts/src/gc/generational.rs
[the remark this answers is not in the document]
crusso: I think we're safe. Whew!

rts/motoko-rts/src/gc/generational/remembered_set.rs
```diff
@@ -0,0 +1,274 @@
...
+        RememberedSet {
+            hash_table,
+            count: 0,
+        }
+    }
+
+    pub unsafe fn insert<M: Memory>(&mut self, mem: &mut M, value: Value) {
+        if is_null_ptr_value(value) {
+            return;
+        }
```
crusso, line 82: Why would you ever insert the sentinel value? I wonder if this test is unnessary.

src/codegen/compile.ml
```diff
@@ -8187,18 +8209,34 @@ let rec compile_lexp (env : E.t) ae lexp =
...
+    set_field ^^
+    get_field,
+    SR.Vanilla,
+    store_ptr ^^
+    (if !Flags.gc_strategy = Mo_config.Flags.Generational
+    then  
+      get_field ^^
+      compile_add_const ptr_unskew ^^
+      E.call_import env "rts" "write_barrier"
+    else G.nop)
```
crusso, line 8239: I guess in all these cases the local "field" is only necessary if we need to insert the write barrier. 
It might be better to avoid declaring and storing the field when the write barriers are not needed, perhaps
by just having two version of this function and switching appropriately (or using Ocaml patttern match guards on the cases, switching on the compiler flag).

crusso: BTW, what happened to the type based optimization (no write barriers for always scalar types)? Perhaps best in a follow on PR, but I thought you'd already done it!)

crusso (approved):
Just some minor things. Generally approved. Let's ship this puppy! (But not as the default yet).

rts/motoko-rts/src/gc/generational/remembered_set.rs
luc-blaeser: Indeed, not necessary. I will remove this.

src/codegen/compile.ml
crusso: Also, wasm has tee instruction that sets a local without popping the value (which might useful) to save on get_field

luc-blaeser: I will improve this with separate versions.

luc-blaeser: I have implemented the static optimization as preparation for the incremental GC and thought to include it later as a separate PR (to reduce complexity and risk).

src/codegen/compile.ml
```diff
@@ -8209,34 +8209,50 @@ let rec compile_lexp (env : E.t) ae lexp =
...
-    get_field,
-    SR.Vanilla,
-    store_ptr ^^
-    (if !Flags.gc_strategy = Mo_config.Flags.Generational
-     then
-      get_field ^^
-      compile_add_const ptr_unskew ^^
-      E.call_import env "rts" "write_barrier"
-     else G.nop)
+      let store_no_barrier =
```
crusso, line 8212: Hmm, this is kinda worse since it generates both version of the code only to throw one away. Want me to have a go?

src/codegen/compile.ml
```diff
@@ -8209,34 +8209,50 @@ let rec compile_lexp (env : E.t) ae lexp =
   match lexp.it with
   | VarLE var -> Var.set_val env ae var
```
crusso, line 8210: Doesn't the VarLE case need a write barrier too? Or is that in set_val?

src/codegen/compile.ml
luc-blaeser: Yes, thanks a lot.

src/codegen/compile.ml
luc-blaeser: I added your PR. Thanks again.

PR #3499, opened 2022-10-19
viper: assertions

ggreif:
This adds the various kinds of static assertions to the Motoko grammar.

TODO:
- [ ] tokens
  - [x] `invariant`
  - [ ] `static`
- [x] `> assert:;` -> extend `Error_reporting.terminal2token`
- [x] `AssertE` _how_ argument
- [ ] purity declaration (`function` instead of `method` in Viper)
- [x] pre/post-conditions (`assert:func`, `assert:return`)
- [ ] loop invariants (`assert:loop`, `assert:invariant`, `assert:break` for enter, stay and exit)
- [x] `implies` (translates to Viper `==>`)


crusso:
Hm, I'm not sure this is the way to go though it might work for a quick hack.

In general, invariants belong to just loops, pre and post to just methods headers etc. Allowing them all over the place would allow  eg. preconditions to incorrectly refer to local variables etc.

src/mo_frontend/assertions.mly
```diff
@@ -0,0 +1,11 @@
+%%
+
+%public exp_nondec(B) :
+  | ASSERT COLON STABLE e=exp_nest
```
ggreif, line 4: should be `INVARIANT`

ggreif (commented):
Yes, this is just for a quick hack. `assert:invariant` may occur in `actor` at top-level.
`assert:func` and `assert:return` are ignored (print as comments).

ggreif:
> preconditions to incorrectly refer to local variables

Right. This will ultimately cause hard errors on the Viper side.

src/mo_frontend/assertions.mly
```diff
@@ -0,0 +1,23 @@
+%token INVARIANT
+(* I get
+> Error: 54 states have shift/reduce conflicts.
```
ggreif, line 3: To me this looks like a `menhir` bug.

src/viper/syntax.ml
```diff
@@ -72,6 +72,8 @@ and stmt' =
   | IfS of exp * seqn * seqn
   | WhileS of exp * invariants * seqn
   | LabelS of id * invariants
+  | PreconditionS of exp
```
ggreif, line 75: These are currently get translated as `stmt` lists, but before constructing the `MethodI` they get converted to the pre/post entries of `MethodI`, and stripped from the body. 63189ba6b5b0d32fa95b3edfad89ff5eaad66881

test/viper/invariant.mo
```diff
@@ -4,10 +4,13 @@ actor {
 
   var count = 0 : Int;
 
-  assert claimed and not (-1 == -1) and (-42 == -42) or true;
-  assert count > 0;
+  assert:invariant claimed and not (-1 == -1) and (-42 == -42) or true;
```
ggreif, line 7: Of course we can change this to just be `invariant`.

test/viper/invariant.mo
```diff
@@ -4,10 +4,13 @@ actor {
...
+  assert:invariant count > 0;
+  assert:invariant not claimed implies count == 0;
 
   public shared func claim() : async () {
+      assert:func count >= 0;
+      assert:return count >= 0;
```
aterga, line 13: What do these two lines mean?

aterga: Is this how preconditions and postconditions being rendered?

ggreif: We don't have dedicated Motoko syntax yet, so they are just Motoko statements.

aterga: Sounds good. Thanks!

test/viper/invariant.mo
aterga: This looks much better, thanks!

aterga (approved):

PR #3500, opened 2022-10-19
experiment: viper await async ghost state

crusso:
- adds await async await ghost variables and internal assertions 
- experiment with lighter weight helper for constructing internal nodes.


src/viper/trans.ml
```diff
@@ -204,14 +226,35 @@ and stmt ctxt (s : M.exp) : seqn =
...
+     in
+
+     ctxt.ghost_inits := mk_s :: !(ctxt.ghost_inits);
+     { !!([],
+          [ !!(FieldAssignS(
+              (self ctxt Source.no_region, id),
+              (!!(AddE(!!(FldAcc (self ctxt Source.no_region, id)),
+                       !!(IntLitE (Mo_values.Numerics.Int.of_int 1)))))));
+            !!(ExhaleS (!!(BoolLitE true)));
+            !!(InhaleS (!!(BoolLitE true)));
```
crusso, line 250: Complete me...

src/viper/trans.ml
```diff
@@ -5,6 +5,23 @@ open Syntax
 module T = Mo_types.Type
 module M = Mo_def.Syntax
 
+module Stamps = Env.Make(String)
+
+let stamps : int Stamps.t ref = ref Stamps.empty
+
+let (!!) it = { it; at = Source.no_region; note = NoInfo}
```
ggreif, line 12: thank you :-)

src/viper/trans.ml
ggreif: Both should be `MacroCall`

src/viper/trans.ml
aterga:
```suggestion
let (!!) it = { it; at = Source.no_region; note = NoInfo }
```

;-)

aterga: Btw, what's your intuition for calling this symbol `!!`?

aterga:
This is great, thank you @crusso ! I spotted only one thing that's missing: after the first havoc, we must inhale `self._message_async > 0`, meaning that the code block after the havoc cannot be executed unless this block is actually entered. Does this make sense? See Martin's snippet here: https://dfinity.slack.com/archives/C03AWN58T32/p1666097506313909

src/viper/trans.ml
crusso: None.

crusso:
> 

Good catch. Will fix then merge

PR #3501, opened 2022-10-20
viper: error source mappings in `moc.js`

rvanasa:
Passes the `lookup` function through `moc.js` to facilitate source mapping in VS Code.


rvanasa:
There is currently a bug hiding somewhere in the way that `moc.js` initializes or calls the `lookup` function from `Viper.Pretty.prog_mapped`. 

I expect this to be a quick fix tomorrow, but I would appreciate a second pair of eyes in case anyone gets a chance to skim this PR. This is most likely the final step to a working demo for source locations. 

Here's a command to reproduce the issue:

```bash
make -C src moc.js && node -e "const {Motoko} = require('./src/moc.js'); Motoko.saveFile('File.mo', 'actor { public func abc(): async () { } };'); console.log('Result:', Motoko.viper(['File.mo']).code.lookup('', [0,0,0,0]))"
```

Replacing `...lookup('', [0,0,0,0]))` with any other values `(filename, [leftLine, leftCol, rightLine, rightCol])` always seems to return `null` (indicating `None` is returned from the `lookup` function). This probably has something to do with this PR's changes in `Pipeline.viper_files'`, since I checked that the `js_of_ocaml` conversions work as expected. 

Does anything stand out as the issue? Otherwise I'll continue narrowing this down tomorrow morning.

src/pipeline/pipeline.ml
```diff
@@ -494,14 +494,12 @@ let viper_files' parsefn files : viper_result =
   let prog = CompUnit.combine_progs progs in
   let u = CompUnit.comp_unit_of_prog false prog in
   let v = Viper.Trans.unit u in
-  let s = Viper.Pretty.prog v in
+  let s = Viper.Pretty.prog_mapped "" v in
```
ggreif, line 497: The filename passed I must be the one that the Viper-side diagnostic points to.

ggreif:
> Does anything stand out as the issue?

My hunch is that the `let inside` function in `Pretty.prog_mapped` takes filenames in account, so that may fail. We could beef up `inside` to skip the name comparison when the filename passed to `Pretty.prog_mapped` is empty.

src/exes/moc.ml
```diff
@@ -191,7 +191,7 @@ let process_files files : unit =
   | Check ->
     Diag.run (Pipeline.check_files files)
   | Viper ->
-    let s = Diag.run (Pipeline.viper_files files) in
+    let s = Diag.run (fst (Pipeline.viper_files files)) in
```
crusso, line 194:
```suggestion
    let (s, _) = Diag.run (Pipeline.viper_files files) in
```
Seems to be required otherwise I get a type error build moc

ggreif:
@rvanasa can this go in?

src/exes/moc.ml
rvanasa: Thanks! Looks much clearer this way as well.

ggreif (approved):
LGTM

rvanasa:
Merging. If anything needs a quick fix, I'll just commit directly to the `viper` branch.

PR #3502, opened 2022-10-21
Write Barrier (Preparation for Incremental GC)

luc-blaeser:
**Note**: Not for merging. This PR is only a small sub-component for the future incremental GC.

# Write Barrier (Preparation for Incremental GC)

Empty write barrier to be used for snapshot-at-the-beginning marking by the future incremental GC.

Based on [this previous PR](https://github.com/dfinity/motoko/pull/2702), optimized with static barrier elimination.

## Properties

* ***Pre-update***: The barrier is called before a pointer write, as needed for snapshot-at-the-beginning marking.
* ***Field-level***: The barrier receives the address of the field or array element address that will be modified. 
* ***Static optimization***: Compile-time barrier elimination based on a simple conservative analysis of the type of the modified field/array element. Specifically, the barrier is skipped if the type of the written location does not allow pointers (`Bool`, `?Bool`, `Char`, `?Char`, `Nat8`, `?Nat8`, `Nat16`, `?Nat16`, `Int8`, `?Int8`, `Int16`, `?Int16`, `()`, `?()`). (Multi-level optional types can not be ommited as `??null`, `???null` etc. refer to heap objects. Scalar types with >=32 bits can be indirected due to pointer tagging.)

**Note**: Currently, the barrier only contains an empty inner logic (in release mode) to measure the base overhead of the barrier instrumentation.

## Compile Flag

The write barrier can be activated by the compile flag `--write-barrier`.

## Performance

Measuring the performance of the GC benchmark in release mode (no sanity checks), by running it first without barriers, and then again with activated barriers.

Comparing the number of mutator instructions, average across all benchmark cases:

Configuration   | Avg. Mutator Instructions
----------------|--------------------------
No barrier      | 1.51e10
Write barrier   | 1.52e10

**Overhead of <1% on average for empty barrier.**

Costs vary between 0 and 2% depending on the GC benchmark case (highest cost for `buffer` and `linked-list`).

## Testing

Two types of sanity checks are enabled in debug mode:

* ***Snapshot comparison***: Based on [this previous PR](https://github.com/dfinity/motoko/pull/2702), memory snapshots are taken and compared at regular GC intervals to verify that for each pointer modification, the corresponding pointer location has been recorded by the write barrier.
* ***Location checks***: Checking that the argument passed to the write barrier is a valid memory location. If it contains a pointer, check the referenced object tag.
* ***Elimination checks***: Special branch `luc/write-barrier-optimization` with additional checks verifying that all statically eliminated barriers indeed refer to scalar values.

1. Mototoko `test/run` and `test/run-drun` test cases:

    ```
    export EXTRA_MOC_ARGS="--write-barrier --sanity-checks --force-gc"
    make
    ```

    and 

    ```
    export EXTRA_MOC_ARGS="--write-barrier --sanity-checks --force-gc --compacting-gc"
    make
    ```

2. Special branch `luc/write-barrier-optimization`:

    Re-run all tests on that branch to verify the static optimization.

## Other Branch

* **Object-level barrier**: A version of object-level barrier is implemented in the branch `luc/write-barrier-object-level`.


src/codegen/compile.ml
```diff
@@ -3411,6 +3417,9 @@ module Arr = struct
       idx env
   )
 
+  let element_type env typ = match typ with
+    | Type.Array (element_type) -> element_type
+    | _ -> assert false
```
crusso, line 3422: This worries me slightly if you are trying to access the type of the element T in e.g. [var T] or [T].
I wonder if it should be
```suggestion
  let element_type env typ = 
   Type.(match normalize typ with
    | Array (Mut element_type)  
    | Array element_typ -> element_type
    | _ -> assert false)
```

The other thing I'm not sure of is whether T could be a typ variable, in which case it might be necessary to promote (to its bound) instead of normalize.  


An example to test whether this works with code that requires promotion might be:

```
func sub<T, A <: [Var T]>(a : A) : T { a[0]; }
```

luc-blaeser: Thank you very much.
There was indeed an issue with generic array type definitions that I now fixed by type promotion.
Handling mutable and non-mutable types is covered by `potential_pointer_type` and can be omitted in this function.

luc-blaeser:
Integrated in the incremental GC: https://github.com/dfinity/motoko/pull/3642

PR #3503, opened 2022-10-21
viper: map translation exceptions to diagnostic errors

crusso:


test/viper/unsupported.mo
```diff
@@ -0,0 +1,5 @@
+actor {
+
+  let x = ""; // strings aren't supported
+
+}
```
crusso, line 5:
```suggestion
}

```

PR #3504, opened 2022-10-21
viper: parse and emit concurrency assertions

ggreif:
This adds assertions of the form `assert:1:async <cond>;` to the grammar, and each `async { ... }` block should specify one.
The conditions are collected and inserted into the summary actor invariant as implication tails with a head condition that the corresponding `async` block is in-flight. This PR builds on the foundation of #3500.

Notes:
- the `1` between the colons signifies that _at most one_ instantiation of the `async` block is permitted (unicity). The business logic is required to ensure this side condition
- the boolean `<cond>` is evaluated in actor context (no local access allowed) and states a condition that an in-flight concurrent action must respect (upon invocation)
- the `assert:1:async` should be the first expression in the `async` block.

This PR does not touch the subject of _message callback_ s, which are a similar concept.

TODOs:
- [x] harvest `ConcurrencyS`
- [x] build condition
- [x] include into `$Inv`
- [x] remove hacks
- [ ] can we improve the context-dependent templating?
- [ ] cleanups
- [ ] can we get rid of the imperative flag numbers and ghost state?


src/viper/trans.ml
```diff
@@ -205,6 +226,9 @@ and dec_field' ctxt d =
         let ctxt'' = { ctxt' with self = Some self_id.it }
         in (* TODO: add args (and rets?) *)
         let stmts = stmt ctxt'' e in
+        let conc, stmts = extract_concurrency stmts in
+        (*assert List.(length conc = 3);
```
ggreif, line 235: remove

src/viper/pretty.ml
```diff
@@ -14,6 +14,11 @@ let semi ppf () = fprintf ppf ";@ "
 let pp_info ppf NoInfo = ()
 
 let rec pp_prog ppf p =
+  marks := p.at :: !marks;
```
ggreif, line 17: maybe back this out?
otherwise it will underline the whole program on error...
can we limit to the `actor { ... }`?

src/viper/trans.ml
```diff
@@ -80,6 +85,27 @@ let rec extract_invariants : item list -> (par -> invariants -> invariants) = fu
...
+
+  let extr (concs, stmts) s : stmt' list * stmt list =
+    match s.it with
+    | ConcurrencyS _ -> s.it :: concs, stmts
+    | SeqnS seq ->
+      let concs', seq = extract_concurrency seq in
+      List.(append (rev concs') concs), { s with it = SeqnS seq } :: stmts
+    | WhileS (e, inv, seq) ->
+      let concs', seq = extract_concurrency seq in
+      List.(append (rev concs') concs), { s with it = WhileS (e, inv, seq) } :: stmts
```
ggreif, line 97: use `rev_append`

ggreif: Done 671d53fe3.

src/viper/trans.ml
```diff
@@ -312,6 +337,22 @@ and stmt ctxt (s : M.exp) : seqn =
...
+     let mk_c = match conc with
+       | [] ->
+         fun _ x -> x
+       | ConcurrencyS ("1", _, cond) :: _ ->
+         let (!!) p = !!! (cond.at) p in
+         let zero, one = intLitE Source.no_region 0, intLitE Source.no_region 1 in
+         fun ctxt x ->
+           let ghost_fld = !!(FldAcc (self ctxt Source.no_region, id)) in
+           let between = !!(AndE (!!(LeCmpE (zero, ghost_fld)), !!(LeCmpE (ghost_fld, one)))) in
+           let is_one = !!(EqCmpE (ghost_fld, one)) in
```
crusso, line 352:
```suggestion
           let ghost_fld () = !!(FldAcc (self ctxt Source.no_region, id)) in
           let between = !!(AndE (!!(LeCmpE (zero, ghost_fld ())), !!(LeCmpE (ghost_fld (), one)))) in
           let is_one = !!(EqCmpE (ghost_fld (), one)) in
```
To avoid sharing mutable notes (which might be problem in the future, but not now)

src/viper/syntax.ml
```diff
@@ -58,6 +58,9 @@ and invariants = exp list
 
 and stmt = (stmt', info) Source.annotated_phrase
 
+and tmpl = (tmpl', info) Source.annotated_phrase
+and tmpl' = (Mo_def.Syntax.exp -> exp) -> exp
```
crusso, line 62: Mixing the two syntaxes (Motoko and Viper) is kinda gross, but we can let it slide for now.

I wonder if this would be simpler if you just collected the async invariants from the motoko term during the translation and translated eager to vanilla viper, rather than introducing the intermediate, artificial ConcurrencyS statement. Ditto for PreconditionS and PostconditionS. None of these are mean to appear in the final target term, right?

src/viper/trans.ml
ggreif: Me getting lazy because of Haskell exposure... Yep, makes sense!

src/viper/syntax.ml
```diff
@@ -74,6 +77,7 @@ and stmt' =
   | LabelS of id * invariants
   | PreconditionS of exp
   | PostconditionS of exp
+  | ConcurrencyS of string * exp * tmpl
```
crusso, line 80:
```suggestion
  (* TODO: these are temporary helper terms  that should not appear in the final translation 
       we should avoid introducing them in the first place if possible, so they can be removed *)
  | PreconditionS of exp
  | PostconditionS of exp
  | ConcurrencyS of string * exp * tmpl
```

src/mo_frontend/assertions.mly
```diff
@@ -19,5 +19,7 @@ when I write this here (instead of in parser.mly)
     { AssertE(Precondition, e) @? at $sloc }
   | ASSERT COLON RETURN e=exp_nest
     { AssertE(Postcondition, e) @? at $sloc }
+  | ASSERT COLON s=NAT COLON ASYNC e=exp_nest
```
crusso, line 22:
```suggestion
  | ASSERT COLON s=NAT COLON ASYNC e=exp_nest
```
why are we adding the nat again if we always require it to be 1 anyway?

crusso (approved):
Bit hackier than I'd like but in the interested of getting it done, feel free to merge (but see suggestions)

src/viper/syntax.ml
ggreif: Yep, they get all filtered out, eventually. The problem is that I cannot pretty-print the Viper term as the `DefinitionI` base context is not available there; it gets created later on. Also `trans.ml` should not depend on `pretty.ml`, right?

src/mo_frontend/assertions.mly
ggreif: Good q. Would that involve defining a new token? Too much hassle for very transient syntax.

src/viper/pretty.ml
ggreif: Done in f4d82dc7e. We can bring it back any time if it appears to be useful.

src/mo_frontend/assertions.mly
crusso: I just mean simplify to:
` 
| ASSERT COLON ASYNC e=exp_nest
`

src/viper/syntax.ml
crusso: Actually, I don't mind `InvartiantI` since it closely corresponds to a MacroDefinition in Viper (source) syntax. So we could keep that but rename to `DefineI`. But we can sort this out later...

src/mo_frontend/assertions.mly
ggreif: Aaah, but we may figure out what to do for `assert:2:async` (etc.) on Monday...

PR #3507, opened 2022-10-22
viper: add the concurrency condition, observe verification fail

ggreif:


default.nix
```diff
@@ -786,7 +786,7 @@ rec {
           nixpkgs.nix-update
           nixpkgs.rlwrap # for `rlwrap moc`
           nixpkgs.difftastic
-          nixpkgs.openjdk nixpkgs.z3 nixpkgs.jq # for viper dev
+          nixpkgs.openjdk nixpkgs.z3 nixpkgs.jq nixpkgs.vscodium # for viper dev
```
ggreif, line 789: ehh, this is a stray one

aterga:
I think the test that this PR adds makes sense. To make it even more realistic, I would make the spec exactly the same as in the verified `claim.mo` example, namely, by adding the desired canister invariant:

```motoko
  assert:invariant not claimed implies count == 0;
  assert:invariant count == 0 or count == 1;
```

aterga (changes requested):
See comment

crusso (approved):

default.nix
ggreif:
```suggestion
          nixpkgs.openjdk nixpkgs.z3 nixpkgs.jq # for viper dev
```

PR #3510, opened 2022-10-23
chore: try fixing the NO-PULL problem in `update-hash`

ggreif:
See the failure in #3509.

Looks like there is a new field:
hash-updater


nomeata (approved):
Rubber stamp

PR #3520, opened 2022-10-25
Support static assertions as `assert:system` in Motoko to Viper translation

aterga:
Why this is needed:
- Peace of mind assertion that Viper can derive what you think should hold.
- Smoke testing soundness of the verification technique or proving that code is unreachable via assert:system false.
- One can use static assertions to introspect a program that fails to verify. 
- There’s also the possibility to replace dynamic assertion with static ones (e.g., to have stronger guarantees or simply save some cycles)
- In case the proof search is undecidable, user-defined static assertions can guide the quantifier instantiation heuristics.


ggreif (approved):
Perfect! Thanks for adding this!

ggreif (commented):
Only needs a test case!

ggreif (approved):
LGTM!

Issue #3519, opened 2022-10-25
Type checker doesn't reject parametrised shared functions


borovan:
with 0.7.1

![image](https://user-images.githubusercontent.com/84897664/197779413-70949eb8-ccc3-4be1-b386-3831876852c3.png)

doing silly generic tests and 

my god our life would be easier if this could work :)
```

Fatal error: exception "Assert_failure mo_types/typ_hash.ml:123:27"
Raised at Mo_types__Typ_hash.go.(fun) in file "mo_types/typ_hash.ml", line 123, characters 27-53
Called from Stdlib__list.iter in file "list.ml", line 110, characters 12-15
Called from Mo_types__Typ_hash.go in file "mo_types/typ_hash.ml", line 123, characters 4-58
Called from ExpGraph.unfold.go in file "lang_utils/expGraph.ml", line 92, characters 22-28
Called from Stdlib__list.map in file "list.ml", line 92, characters 20-23
Called from Stdlib__list.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__list.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__list.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__list.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__list.map in file "list.ml", line 92, characters 32-39
Called from ExpGraph.unfold.go in file "lang_utils/expGraph.ml", line 93, characters 18-34
Called from ExpGraph.unfold in file "lang_utils/expGraph.ml", line 97, characters 10-17
Called from Mo_types__Typ_hash.typ_hash in file "mo_types/typ_hash.ml", line 164, characters 11-25
Called from Stdlib__list.map in file "list.ml", line 92, characters 20-23
Called from Mo_types__Typ_hash.typ_seq_hash in file "mo_types/typ_hash.ml" (inlined), line 167, characters 19-41
Called from Codegen__Compile.BumpStream.name_for in file "codegen/compile.ml", line 4648, characters 50-74
Called from Codegen__Compile.MakeSerialization.serialize in file "codegen/compile.ml", line 5838, characters 15-43
Called from Codegen__Compile.compile_prim_invocation in file "codegen/compile.ml", line 9023, characters 6-36
Called from Codegen__Compile.compile_exp in file "codegen/compile.ml", line 9120, characters 4-54
Called from Codegen__Compile.compile_exp_as in file "codegen/compile.ml", line 9253, characters 24-44
[...]

PR #3522, opened 2022-10-26
fix: reject generic shared functions and function types (fixes issue #3519)

crusso:
#3519 causes a compiler crash because we:

* don't reject generic shared functions
* can't hash types involving generic functions.

The actual crash was probably caused when compiling the async function for instantiating the imported actor class, since the type of the actor (with the slipped in polymorphic method) could not be hashed when constructing the candid type table.


test/run-drun/issue-3519.mo
```diff
@@ -0,0 +1,5 @@
+import C "issue-3519/class";
+
+actor {
+
+}
```
crusso, line 5:
```suggestion
}

```

test/run-drun/issue-3519/class.mo
```diff
@@ -0,0 +1,14 @@
...
+  };
+
+  // test
+  public shared ({ caller }) func asplode<T>() : async GenericType<T> {
+    return {
+      thing = 1;
+    };
+  };
+
+};
```
crusso, line 14:
```suggestion
};

```

nomeata (approved):

PR #3523, opened 2022-10-27
viper: fix source locations for repeated code translations

rvanasa:
Fixes a bug which causes the source map to remain the same in subsequent runs of `Pretty.prog_mapped` via `moc.js`.


ggreif (approved):
Good find!

PR #3524, opened 2022-10-27
viper: associate the first `exhale` with the `async` expression

ggreif:
This changes the srcloc of certain Viper subexpression to better track the original Motoko.


src/viper/trans.ml
```diff
@@ -323,37 +324,36 @@ and stmt ctxt (s : M.exp) : seqn =
...
           [ !!(FieldAssignS(
               (self ctxt Source.no_region, !!id),
               (!!(AddE(!!(FldAcc (self ctxt (s.at), !!id)),
                        intLitE Source.no_region 1)))));
-            !!(ExhaleS (!!(AndE(!!(MacroCall("$Perm", self ctxt at)),
-                                !!(MacroCall("$Inv", self ctxt at))))));
-            !!(SeqnS (
-                let (!!) p = !!! at p in
-                !!([],
+            !@(ExhaleS (!@(AndE(!@(MacroCall("$Perm", self ctxt at)),
```
ggreif: This is thee relevant change, sorry for the other noise...

src/viper/trans.ml
```diff
@@ -323,40 +324,39 @@ and stmt ctxt (s : M.exp) : seqn =
...
-                                         !!(MacroCall("$Inv", self ctxt at)))))) ])));
-            !!(InhaleS (!!(AndE(!!(MacroCall("$Perm", self ctxt at)),
-                                !!(MacroCall("$Inv", self ctxt at))))));
-          ])
+     !!([],
+        [ !!(FieldAssignS(
+            (self ctxt Source.no_region, !!id),
+            (!!(AddE(!!(FldAcc (self ctxt (s.at), !!id)),
+                     intLitE Source.no_region 1)))));
+          !@(ExhaleS (!@(AndE(!@(MacroCall("$Perm", self ctxt at)),
```
ggreif, line 341: This is the relevant change, sorry for the other noise...

ggreif: `!!` still marks up the `await`, but `!@` points at the `async`.

crusso: Right but doesn't the exhale belong to the await, not the async? This is when you need to establish the invariant before suspending. That was my thinking at least. But I don't know why it doesn't highlight the await rather than default to the unknow region.

crusso (approved):
I'm not sure this is correct, but I'll approve for the sake of the demo.

src/viper/trans.ml
```diff
@@ -323,40 +324,39 @@ and stmt ctxt (s : M.exp) : seqn =
...
+                 [
+                   !@(InhaleS (!@(AndE(!@(MacroCall("$Perm", self ctxt at)),
+                                  !@(AndE(!@(MacroCall("$Inv", self ctxt at)),
+                                          !@(GtCmpE(!@(FldAcc (self ctxt at, !@id)),
+                                               intLitE Source.no_region 0))))))));
+                   !@(FieldAssignS(
+                          (self ctxt at, !@id),
+                          (!@(SubE(!@(FldAcc (self ctxt at, !@id)),
+                                   intLitE at 1)))));
+                   !!! (e.at) (SeqnS stmts);
```
crusso, line 354: Is this maybe confusing things and should really be !@

ggreif:
> sake of the demo

Not needed for the demo and I won't merge right now. This can wait.

src/viper/trans.ml
ggreif: Not so sure.

src/viper/trans.ml
ggreif: I think the `exhale` is due to the `async` being filed into the continuation table. But you are right, that only the `await` opens up the context switch so that a _havoc_ can occur.

aterga (approved):

PR #3526, opened 2022-10-27
Performance Improvement: Array.init and Array.tabulate

luc-blaeser:
# Faster Array Init and Tabulate

* Eliminating unnecessary repeated array index bounds check while initializing an array.
* Faster iteration through new array on initialization by advancing an element pointer.

Applies to `Array.init` and `Array.tabulate`.

## Performance Improvement

**More than 2x faster array initialization**

### Micro-Benchmark

Measuring the number of mutator instructions with an array of 10_000_000 Nat elements. Tabulate uses `x -> x`.

Operation   | Unoptimized   | Optimized | Speedup
------------|---------------|-----------|-----------
`init`      | 260e6         | 120e6     | 2.17     
`tabulate`  | 380e6         | 280e6     | 1.36

### GC Benchmark

Measuring the number of mutator instructions in GC benchmark. 

Benchmark Case      | Unoptimized   | Optimized | Speedup
--------------------|---------------|-----------|-----------
`blobs`             | 4.56e7        | 4.44e7    | 1.03
`buffer`            | 2.06e10       | 2.04e10   | 1.01
`graph`             | 1.09e10       | 1.02e10   | 1.07
`random-maze`       | 2.51e8        | 2.47e8    | 1.02
`sha256`            | 3.31e10       | 2.97e10   | 1.14

No relevant improvement for the other benchmark cases.

On GC benchmark, the optimization improves performance by **3.4%** on average.

### Testing

1. Mototoko `test/run` and `test/run-drun` test cases with sanity checks.

    * Extended the test case `test/run/array-gen.mo` with larger arrays and edge cases (size 0, 1).

3. Special branch `luc/array-init-checks` with extensive sanity checks:

    * Full memory sanity check after every `Array.init` and `Array.tabulate` call.
    * Checking the bounds of every element access during `Array.init` and `Array.tabulate`.

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -3411,7 +3411,6 @@ module Arr = struct
       idx env
   )
 
-
   let vanilla_lit env ptrs =
     E.add_static env StaticBytes.[
       I32 Tagged.(int_of_tag Array);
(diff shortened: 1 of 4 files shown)


crusso:
Nice!

crusso:
Now tempting to inline lambda argument of tabulate when obvious...

src/codegen/compile.ml
```diff
@@ -3428,50 +3427,76 @@ module Arr = struct
...
+
+      (* Next element pointer, skewed *)
+      get_pointer ^^
+      compile_add_const element_size ^^
+      set_pointer
+    )
+
   (* The primitive operations *)
   (* No need to wrap them in RTS functions: They occur only once, in the prelude. *)
   let init env =
```
crusso, line 3464: Once Gabor finish his PR to update our wasm library, we could also use a memory fill instruction to replicate the values. 
In the meantime, we could improve `init` by doing mostly Word64 based writes.

luc-blaeser: True, this could be improved much further. If it would be a zero-initialization, it may even use the new WASM bulk memory operation.

nomeata (approved):

crusso:
@luc-blaeser if you label the PR as auto-merge squash, it'll use your nice description as the commit message without further work.

crusso:
You could also add a simple one-line description to the changelog (Changelog.md) referencing the PR number.

src/codegen/compile.ml
```diff
@@ -3428,50 +3427,76 @@ module Arr = struct
...
-    from_0_to_n env (fun get_i ->
-      (* Where to store *)
-      get_r ^^ get_i ^^ idx env ^^
+    (* Initial index *)
+    compile_unboxed_const 0l ^^
+    set_i ^^
+
+    (* Write elements *)
+    iterate env get_r (fun get_pointer -> 
+      get_pointer ^^
```
crusso, line 3499:
```suggestion
    iterate env get_r (fun get_pointer -> 
      get_pointer ^^
      ...
      get_i
      BigNum.from_word32 env
```
Since we are optimizing... is it possible to avoid the size test implicit in BigNum.from_word32 if we know that all array sizes are less than 30 bits (and do we know that?)

crusso: I'm thinking all you need to do is add (something like) 4(?) to the tagged scalar representiation of `i` directly at the end of the loop and avoid the BigNum.from_word32 (and implicit conditional) on each iteration.

crusso: @ggreif didn't you do something like this for array iteration?

ggreif: Let's not cram too much inn one PR. That said, the bignum-heap-redirect is indicated by the LSBit. So (assuming there are no heap-stored _small_ `Int` numbers) one can rotate right by one and compare against a reasonable element count upper bound. (like 500M elements?). Then leave the allocator do its magic. If there is a heap redirect, it will cause huge element counts after the `rot`.

src/codegen/compile.ml
```diff
@@ -3428,50 +3427,76 @@ module Arr = struct
   (* Does not initialize the fields! *)
   let alloc env = E.call_import env "rts" "alloc_array"
 
+  let iterate env get_array body = 
+    let (set_boundary, get_boundary) = new_local env "boundary" in
+    let (set_pointer, get_pointer) = new_local env "pointer" in
+    
+    (* Initial element pointer, skewed *)
+    compile_unboxed_const header_size ^^
+    compile_mul_const element_size ^^
```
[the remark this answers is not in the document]
luc-blaeser: A helper would be great. There are various such places.

PR #3529, opened 2022-10-28
test: unit test `bitrel.rs`, fixing bug

crusso:
Fixes bug in #3171  

 * add unit test for BitRel.rs
 * fix insidious bug revealed by test - conversion to bit position was broken, failing to account for all bits and leading to confusion of BitRel entries.
 * optimize indexing arithmetic and sanity checks

Why wasn't this caught by the moc tests? Probably because the bitrels used by candid subtype checking are actually sparse so the interference was never observed.

### DIFF
--- a/rts/motoko-rts/src/bitrel.rs
+++ b/rts/motoko-rts/src/bitrel.rs
@@ -38,26 +38,16 @@ impl BitRel {
     unsafe fn locate_ptr_bit(&self, p: bool, i_j: u32, j_i: u32, bit: u32) -> (*mut u32, u32) {
         let size1 = self.size1;
         let size2 = self.size2;
-        let (base, i, j) = if p {
-            (0, i_j, j_i)
-        } else {
-            (size1 * size2 * BITS, j_i, i_j)
-        };
-        if i >= size1 {
-            idl_trap_with("BitRel i out of bounds");
-        };
-        if j >= size2 {
-            idl_trap_with("BitRel j out of bounds");
-        };
-        if bit >= BITS {
-            idl_trap_with("BitRel bit out of bounds");
-        };
-        let k = base + (i * size2 + j) * BITS + bit;
+        let (base, i, j) = if p { (0, i_j, j_i) } else { (size1, j_i, i_j) };
+        debug_assert!(i < size1);
+        debug_assert!(j < size2);
+        debug_assert!(bit < BITS);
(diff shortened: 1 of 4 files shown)


rts/motoko-rts/src/bitrel.rs
```diff
@@ -39,30 +35,19 @@ impl BitRel {
...
+        debug_assert!(j < size2);
+        debug_assert!(bit < BITS);
+        let k = ((base + i) * size2 + j) * BITS + bit;
         let word = (k / usize::BITS) as usize;
         let bit = (k % usize::BITS) as u32;
         let ptr = self.ptr.add(word);
         if ptr > self.end {
-            idl_trap_with("BitRel ptr out of bounds");
+            idl_trap_with("BitRel indices out of bounds");
         };
```
ggreif, line 51: this should be checked for debug only, right?

rts/motoko-rts/src/bitrel.rs
```diff
@@ -39,30 +35,19 @@ impl BitRel {
...
-            idl_trap_with("BitRel bit out of bounds");
-        };
-        let k = base + i * size2 * BITS + j + bit;
+        let (base, i, j) = if p { (0, i_j, j_i) } else { (size1, j_i, i_j) };
+        debug_assert!(i < size1);
+        debug_assert!(j < size2);
+        debug_assert!(bit < BITS);
+        let k = ((base + i) * size2 + j) * BITS + bit;
         let word = (k / usize::BITS) as usize;
         let bit = (k % usize::BITS) as u32;
```
ggreif, line 47: I winder if this could be some `divMod`.

ggreif (approved):
LGTM!

rts/motoko-rts/src/bitrel.rs
crusso: It could be, but I'd rather keep the check to avoid memory corruption even in production builds.

rts/motoko-rts/src/bitrel.rs
crusso: I couldn't find any such operation in the Rust libs.

ggreif: Yeah, the std crates doesn't seem to have it. But I don't think it is necessary to pull in https://docs.rs/num/latest/num/integer/fn.div_rem.html.

PR #3533, opened 2022-11-01
Extra Write Barrier Coverage Test

luc-blaeser:
Not for merging - only for testing.

# Extra Write Barrier Coverage Test

Builds on [Generational GC PR](https://github.com/dfinity/motoko/pull/3495)

## Properties

Extra checks inserted in the compacting GC and the copying GC to detect write barrier gaps:

* Write barriers are enabled in all GCs and record relevant pointer location in the remembered set.
* The mark phase in the copying GC and the compacting GC ckecks whether all relevant pointers are contained in the remembered set. 
* A pointer is relevant if it has been stored in an old object (`heap_base` <= pointer location < `LAST_HP`) and leads to a young object (pointer value >= `LAST_HP`).

## Performance Overhead

Total number of instructions in the GC benchmark, average across all benchmark cases:

GC          | Without Checks  | Enabled Checks  | Overhead
------------|-----------------|-----------------|-----------
Copying     | 2.88e10         | 3.23e10         | 12%
Compacting  | 3.20e10         | 3.41e10         | 6.6%

The barrier overhead itself (as can be measured by the mutator instructions) is around 7%.


crusso:
Since this build on #3495, can you make the parent branch above be that branch? Then the diff should be smaller and  easier to understand.

src/codegen/compile.ml
```diff
@@ -24,6 +24,8 @@ let (@@) = Stdlib.(@@)
 module G = InstrList
 let (^^) = G.(^^) (* is this how we import a single operator from a module that we otherwise use qualified? *)
 
+let enable_write_barrier = true
```
crusso, line 27: I wonder if it would make sense to just make this enabled only when sanity checking for the compacting or copying_gc and always for the generational gc. Then you could actually check it in (not just have it as an experiment).

I support there would still be the overhead of testing whether the check needs to done during collection, so perhaps unacceptable for the non-generational gcs...

PR #3541, opened 2022-11-02
experiment: implement basic trampoline for moc ast interpreter

crusso:
Although we support browser based execution of the moc interpreter, it currently only runs on tiny examples before exhausting the JS stack. 

This is because the interpreter is written in CPS and ocaml_js does not honour tail calls to unknown functions, i.e. every call of a (dynamic) continuation. 

This PR abuses our existing event loop to implement a simple trampoline for the cps based interpreter, exiting the current interpretation to clear the stack, but resuming it from the front of the event queue.

This allows the js compiled interpreter to run in the browser without running out of stack.

TODO: 
- [x] only do this JS code, not OCaml native code.
- [x] optimize to avoid allocation
- [ ] use a stack local counter, or hide it in the environment?
- [ ] use a dequeue instead of abusing the queue (for more efficient queue jumping)?
- [x] interpretation of prim expressions uses try match and isn't tail rec. FIX FIRST, probably by merging https://github.com/dfinity/motoko/pull/2895

Now, instead of sorting a buffer of 4 elements failing with stack overflow, we can sort hundreds of elements!

Screenshot: 

Before:
image

After:

image


crusso:
Ok, this is not a cure. For example

src/mo_interpreter/interpret.ml
```diff
@@ -119,6 +120,27 @@ struct
...
+  let bounce work =
+    (* add work to *front* of queue *)
+    Queue.transfer q tmp;
+    Queue.add work q;
+    Queue.transfer tmp q
+
+  let interval = 128
+  let count = ref interval
+  let trampoline3 f x y z =
+    if !Flags.ocaml_js then
```
chenyan-dfinity, line 134: Curious, what's the down side of trampolining the interpreter in native OS?

chenyan-dfinity (approved):
LGTM

src/mo_interpreter/interpret.ml
crusso: Dunno, but I image its more expensive than a tailcall.

crusso:
@chenyan-dfinity PTAL - I've merged in @osa1's PR that removes another source of failure (even in the ocaml intpreter).

The while loop above remains problematic, however.

chenyan-dfinity (approved):

PR #3542, opened 2022-11-02
feat: timers

ggreif:
This PR adds the necessary primitives for the base library to rely on.
``` Motoko
setTimer : (delayNanos : Nat64, recurring : Bool, job : () -> async ()) -> (id : Nat)
cancelTimer : (id : Nat) -> ()
```
It also adds a mechanism `system func timer`, so that the user can choose to provide a custom implementation.
A new compiler flag `-no-timer` is also provided to disable all timer-related functionality (e.g. for critical applications).

-----------
# How it works

By default the timer mechanism is enabled, and the `setTimer`/`cancelTimer` utilities can be used. If the `system func timer` is present, the user is responsible for the timer implementation, i.e. she can supply a very simple or very elaborate mechanism. When `moc` is invoked with `-no-timer` then the whole mechanism is disabled. This is useful for minimising the code footprint.

## Canister-global timer discipline

The IC's underlying timer mechanism is one timer per canister. The canister infrastructure is responsible for providing the abstraction of several (and recurring) timers. Timer expiration is best-effort, i.e. the timer's action will never be executed before the desired expiration time, but there is no guarantee of execution _at_ expiration time. It will usually be later.

## The default mechanism

When the user doesn't choose to implement `system func timer` (and doesn't suppress it by `-no-timer`), a default mechanism gets generated. The default mechanism is basically a call to the builtin `@timer_helper`, which will be provided by the compiler.

In the default mechanism the user calls the `setTimer` call from (Motoko-base). This will register an expiration relative to the current time in form of a timer node that `@timer_helper` will visit. Correspondingly a `cancelTimer` call will expunge a timer node. Canister-global timer expirations will also do this, as well as pruning the node tree (a priority-queue).

### The `Node` structure

`Node`s will track the expiration time, the timer Id, and optionally a duration for recurring timers. For administrative reasons they also keep pointers to the earlier and later trees. Each timer node further has a thunk (a function of type `() -> async ()`) which (when called) returns a _future_ (concurrent action).

The most important invariant for nodes is that the expiration time divides the earlier and later trees temporally:
- every expiration on a `Node` in the earlier-tree is `<=` the expiration of the parent node, and analogously
- every expiration on a `Node` in the later-tree is `>=` the expiration of the parent node.

There is a twist to the expiration of the `Nodes` though, as they are mutable, and are set to 0 when the timer expires or gets cancelled. So the above invariant is strictly only valid for nodes (with their child trees) where no node therein is expunged.

This allows for some neat optimisations. E.g. expiration is caused by time marching from early to later, so if a node's expiration is 0 then we can assume that the earlier-tree is completely expired. _Pruning_ (the cleaning of a timer tree from expired nodes) becomes a later-tree-only operation in this case.

The node expired _implies_ earlier-tree expired is a _corollary of the fundamental invariant_ with expungable nodes.

### Insertion

Timer Ids are given out as `1..` and never reused (but begins after upgrade from 1 again).
The _initial_ insertion of a timer (`setTimer`) respects the invariant that a higher Ids will execute after lower Ids when they have the same expiration time. For _recurrent_ expirations this is also the case if the delays are identical.

### Cancellation

Each timer node carries the Id, and this allows expunging the node when `cancelTimer` is called. To maintain the above corollary, each `cancelTimer` must be followed by a compacting operation (basically pruning, but considering the earlier-trees too). Alternatively a grafting operation can be used, appending the later-tree onto the latest slot of the earlier-tree.

It is important to note that while code is running that expunges nodes as a result of canister global timer expiration, there must not be calls to `cancelTimer` (or similar) since that would likely interfere with the corollary. This is only potentially a problem for self-implemented `system func timer`, as the default mechanism will use `@timer_helper`, which is not doing such things.

### Recurrence

When a timer expires that was set up being recurrent, the timer node needs to be re-added to the timer tree but with an expiration shifted by the timer's delay. This will happen in the loop that expunges timers that are already expired. Thus it may happen that the loop will detect the same timer and expunge it again. For a delay of 0 this might cause an infinite loop, so recurrent timers with 0 delay will be treated as non-recurring. 5b9b62faa implements this by avoiding re-adding.

### Tree walking and reconstruction

Several operations walk (traverse) the timer tree. `setTimer` in particular prunes the expired nodes while adding the new node and similarly the reinsertion of recurrent timers.
Since expiration times are kept in mutable ~~fields~~ array elements, node reinsertion due to recurrence will not cause duplicated expirations when tree nodes get copied. This is due to mutable ~~cell~~ array aliasing in the tree walk code. ~~If we decide to get rid of this feature we have to come up with a different scheme to suppress duplicated firing of timers.~~

### Setting the canister-global timer

There are three triggers to call the system API `global_timer_set`:
- connected to `setTimer` (with the expiration time of the earliest active timer)
- when the timer expiration callback is being invoked
  - and there is another active timer
  - no active timer (`global_timer_set(0)`)
- `cancelTimer` is called, and there is no active timer remaining.

### Trap avoidance

A timed action trapping must not impede the execution of other actions, and thus we use the type `() -> async ()` for registering such actions with `setTimer`. This ensures that
- each action has its own future (even for each recurrence)
- each future gets run in its own `async` context and as such traps and `throw`s are isolated.

### Reclaiming the continuation slots

_Note_: originally the futures were `awaited`, but this seems redundant and has been removed in 204d87a39. Also, an upgrade test is now run to check that executed timer actions don't cause stuck canisters.

~~Upon timer expiration the callback is responsible for installing the workers into the continuation table (this is the `async` part), but it is also important to `await` those futures, so they don't fill up the continuation table. I expect that this can be done by a trailing `await` of an `async forall <futures> try await <it> catch {}`, since the `canister_global_timer` endpoint is called in an `async` context.~~

### The `@timer_helper` builtin

When no user-written `system func timer` is present, `moc` will insert a call to an internal function `@timer_helper`. This
works by creating a worklist of expired timers relative to the current time, then expunging the corresponding nodes and re-inserting recurring ones with respective delays added.

The worklist is then transformed into an array of futures, each in its trap-proof compartment. Finally all the pending futures are `await`ed.

### The upgrade story
<a name="upgrades"></a>

Easy. The global timer gets jettisoned on upgrade, and the timers need to be set up in the post-upgrade hook. Stable variables can be used to remember the timers if they don't have a rigid structure.

## Opting out

To opt-out of the default mechanism the user just declares `system func timer`. By using the passed in function argument to set the global timer, full control is given to either implement a very basic wakeup scheme or an even more elaborate mechanism than the one presented above.

### The initial expiration

After the canister starts running, `system func timer` gets invoked automatically. This also happens after an upgrade.

### An example of a periodic wakeup

The simplest way to achieve a periodic wakeup of a canister with fixed intervals can be coded as
``` Motoko
import Time "mo:base/Time";
import { fromIntWrap } = "mo:base/Nat64";

actor {

  system func timer(set : Nat64 -> ()) : async () {
    set(fromIntWrap(Time.now()) + 60_000_000_000); // 60 seconds from now
    doSomething();
  };

  func doSomething() {
    // whatever
  }

}
```

### An example of a exponential approximation

Imagine you are at a hackathon, and there is a deadline to submit the hash of your solution. In the last hour you want to be reminded 60, 30, 15, 7.5, etc. minutes before the deadline. Following implementation could serve the purpose:
``` Motoko
import { now } = "mo:base/Time";
import { fromIntWrap } = "mo:base/Nat64";
import { print } = "mo:base/Debug";

actor {
  let hour = 60 * 60_000_000_000;
  let deadline = now() + hour; // nanos

  system func timer(set : Nat64 -> ()) : async () {
    let time = now();
    let toGo = deadline - time;
    if (toGo < 0) { set 0; return };

    debug { print("Still to go: " # debug_show toGo) };

    set(fromIntWrap(time + toGo / 2));
  }
}
```

# Open question(s)

- ~~Should the recurring timer be re-inserted with expiration relative to old expiration time or current time (of callback)?~~ — to avoid time skew, using the planned expiration time as the base
- Can we use `async*` to avoid the context switch when calling into `@timer_helper`? — probably yes, but in another PR

------------
TODO:
- [x] add `-no-timer` option to `test/perf` sources — dbfaf545b
- [x] release notes
- [x] check that `setTimer` documented as taking nanoseconds
- [ ] debug deactivated tests (compacting GC in `ic-ref` are a possible indication of GC bug! Or maybe starvation.)
- [x] stabilisation (needed? — CWV: no)
- [ ] tests
  - [ ] after upgrade, still on? ("the timers get canceled on upgrade")
  - [x] when defining `system func timer`, after upgrade it also gets called — 8df4d6f79d67877b43e386b52cdc9f28b9ecad82
  - [x] can upgrade after a timer has run
  - [x] test recurrent delay 0 — See 5b9b62faa and c440b3e8d.
  - [ ] does `actor class` behaves correctly?
- [x] docs: `doc/md/heartbeats.md`
  - [ ] what more is needed?
- [x] run with `ic-ref` too (currently no support, but see https://github.com/dfinity/interface-spec/pull/111)
- [x] `var` aliasing is being used to good effect (can we remove it?) — replaced in #3617
- [ ] check the invariant: no more expiration <=> global timer is 0
- [x] add `test/fail` for `-no-timer`
- [x] cancelling of a timer must be followed by compaction (to not break the corollary)
- [x] duration 0 with recurring timers should be excluded
- [x] traps in jobs, `catch`
- [x] add the low-level API `setGlobalTimer` (but see also: #3614)
- [x] try it on the Playground — _works!_


src/codegen/compile.ml
```diff
@@ -3892,6 +3893,18 @@ module IC = struct
       edesc = nr (FuncExport (nr fi))
     })
 
+  let export_timer env =
+    assert (E.mode env = Flags.ICMode || E.mode env = Flags.RefMode);
+    let fi = E.add_fun env "canister_global_timer"
+      (Func.of_body env [] [] (fun env ->
+        G.i (Call (nr (E.built_in env "timer_exp"))) ^^
+        GC.collect_garbage env))
```
ggreif, line 3901: really?

nix/drun.nix
```diff
@@ -17,6 +17,8 @@ pkgs:
 
       cargoSha256 = "sha256-dhDXhVNAAHzLRHdA6MpIGuoY76UhiF4ObeLO4gG/wo4=";
 
+      patches = [./now.patch];
```
ggreif, line 20: Andriy Berestovskyy's heroic patch!

This can be removed when we import the fixed `drun` from GH...

mraszyk:
> * run with `ic-ref` too (currently no support, but see [Introduce canister global timer interface-spec#111](https://github.com/dfinity/interface-spec/pull/111))

I'm adding support for the timer in ic-ref. I'll open an MR soon.

test/run-drun/timer.mo
```diff
@@ -70,26 +70,32 @@ actor {
...
-        func hunt(n : ?Node) = switch n {
-          case null ();
+        func hunt(n : ?Node) : ?Node = switch n {
+          case null n;
           case (?n) { if (n.id == id) {
-                          n.expire := 0
-                      };
-                      hunt(n.ante);
-                      hunt(n.dopo)
+                          n.expire := 0;
```
ggreif, line 84: No need to do this. Since we don't iterate over this a.t.m., we can just drop `n`. Also, pattern match on `n`.

ggreif: done: 4a7cdef12

test/run-drun/timer.mo
```diff
@@ -0,0 +1,201 @@
...
+  let raw_rand = (actor "aaaaa-aa" : actor { raw_rand : () -> async Blob }).raw_rand;
+
+  var count = 0;
+  var max = 10;
+
+  system func timer() : async () {
+    let now = time();
+
+    type CNode = ?{var expire : Nat64; id : TimerId; delay : ?Nat64; ante : CNode; dopo : CNode };
+    func clean(n : ?Node) : CNode =
```
ggreif, line 111: this is debugging code, remove!

test/run-drun/timer.mo
```diff
@@ -0,0 +1,201 @@
...
+          case null null;
+          case (?n) ?{n with ante = clean(n.ante); dopo = clean(n.dopo) }
+      };
+    //debugPrint(debug_show {now; timers = clean timers});
+
+    let exp = nextExpiration timers;
+    let prev = (prim "global_timer_set" : Nat64 -> Nat64) exp;
+
+    if (exp == 0) {
+        return
```
ggreif, line 122: `timers := null;` ?

test/run-drun/timer.mo
```diff
@@ -0,0 +1,201 @@
...
+    };
+
+    var gathered = 0;
+    let thunks : [var ?(() -> async ())] = Array_init(10/*FIXME*/, null);
```
ggreif, line 126: Either expand the array when filled or stop putting elements in it when full (and leave the expiration of surplus timers to the next firing of the global timer).

test/run-drun/timer.mo
```diff
@@ -0,0 +1,201 @@
...
+         throw error("he's dead Jim");
+     };
+     debugPrint(debug_show {count});
+  };
+};
+
+//SKIP run
+//SKIP run-low
+//SKIP run-ir
+//SKIP ic-ref-run
```
ggreif, line 199: Remove (when ready).

test/fail/ok/no-timer.tc.ok
```diff
@@ -0,0 +1 @@
+no-timer.mo:5.3-6.4: type error [M0222], system function timer is present but -no-timer flag is specified
```
ggreif, line 1: `M0222` needs to be corrected

mraszyk:
The timers are now available on master in ic-hs.

src/codegen/compile.ml
```diff
@@ -3744,6 +3744,7 @@ module IC = struct
       E.add_func_import env "ic0" "stable64_size" [] [I64Type];
       E.add_func_import env "ic0" "stable64_grow" [I64Type] [I64Type];
       E.add_func_import env "ic0" "time" [] [I64Type];
+      E.add_func_import env "ic0" "global_timer_set" [I64Type] [I64Type];
```
ggreif, line 3747: Only when we need this, conditionalise!

test/run-drun/timer-default.mo
```diff
@@ -0,0 +1,39 @@
...
+  // relying on the compiler filling in a default implementation
+
+  var count = 0;
+  var max = 10;
+  let raw_rand = (actor "aaaaa-aa" : actor { raw_rand : () -> async Blob }).raw_rand;
+
+  public shared func go() : async () {
+     var attempts = 0;
+
+     let id1 = setTimer(1, false, func () : async () { count += 1; debugPrint "YEP!" });
```
ggreif, line 15: this can be _much_ shorter

src/codegen/compile.ml
ggreif: We probably need it. It could be made cheaper if we'd compare the HP before with after and only call `GC.collect_garbage` when there is a change. Similarly for `heartbeat`.

test/run-drun/timer-default.mo
```diff
@@ -0,0 +1,39 @@
...
+         throw error("he's dead Jim");
+     };
+     debugPrint(debug_show {count});
+  };
+};
+
+//SKIP run
+//SKIP run-low
+//SKIP run-ir
+//SKIP ic-ref-run
```
ggreif, line 37: enable this when ready

src/lowering/desugar.ml
```diff
@@ -497,6 +499,14 @@ and build_actor at ts self_id es obj_typ =
          (match call_system_func_opt "heartbeat" es obj_typ with
           | Some call -> call
           | None -> tupE []);
+       timer =
+         (match call_system_func_opt "timer" es obj_typ with
+          | Some call -> call
+          | None when !Mo_config.Flags.global_timer ->
+            blockE
+              [ expD T.(callE (varE (var "@run_timers" Mo_frontend.Typing.heartbeat_type)) [unit] (unitE())) ]
```
ggreif, line 507: will this call cause a context switch (`await` gap)? @crusso

test/run-drun/timer.mo
```diff
@@ -0,0 +1,42 @@
...
+  };
+
+  var count = 0;
+  var max = 10;
+  let raw_rand = (actor "aaaaa-aa" : actor { raw_rand : () -> async Blob }).raw_rand;
+
+  public shared func go() : async () {
+     var attempts = 0;
+
+     let id1 = setTimer(1, false, func () : async () { count += 1; debugPrint "YEP!" });
```
ggreif, line 18: FileCheck that the `global_timer` calls are there?

Conversely have a `no-timer.mo` test that is `-no-timer` flagged, and FileChecks that there is no such thing.

ggreif:
> I'm adding support for the timer in ic-ref. I'll open an MR soon.

@mraszyk I don't see callbacks into my canister from `ic-ref` as a response to timer expirations. Something is not working yet :-(

test/run-drun/timer.mo
ggreif: Done in 3788550.

src/prelude/timers.mo
```diff
@@ -0,0 +1,150 @@
+// default timer mechanism implementation
+// fundamental node invariant: max_exp ante <= expire <= min_exp dopo
+// corollary: if expire == 0 then the ante is completely expired
+//
+type @TimerId = Nat;
```
ggreif, line 5: pointless

ggreif: Cleaned up in fc234fa9c.

test/run-drun/ok/timer-trap.drun-run.ok
```diff
@@ -0,0 +1,8 @@
+ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
+ingress Completed: Reply: 0x4449444c0000
+debug.print: YEP!
+debug.print: EEK!
+debug.print: BEAM!
+debug.print: YEP!
+debug.print: {count = 3}
```
ggreif, line 7: `EEK!` got rolled back, so to count until 3 another `YEP!` had to happen.

test/run-drun/generic-tail-rec.mo
```diff
@@ -51,3 +51,5 @@ func Loop<T,U>(n:Int){
 
 Loop<Int,Bool>(0);
 };
+
+//SKIP run
```
ggreif, line 55: the interpreter crashes on M1 with the depth. `bound = 10000` works

test/run-drun/timer-default.mo
```diff
@@ -0,0 +1,31 @@
...
+
+  public shared func go() : async () {
+     var attempts = 0;
+
+     ignore setTimer(1, false, func () : async () { count += 1; debugPrint "YEP!" });
+
+     while (count < max) {
+       ignore await raw_rand(); // yield to scheduler
+       attempts += 1;
+       if (attempts >= 800 and count == 0)
```
ggreif, line 20:
```suggestion
       if (attempts >= 8000 and count == 0)
```
Ridiculously large, so that ic-ref-compacting-GC also has a chance.

nix/sources.json
```diff
@@ -34,16 +34,16 @@
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     },
     "ic-hs": {
-        "branch": "master",
+        "branch": "mraszyk/periodic-system-tasks",
```
ggreif, line 37: This needs to be merged first!

mraszyk: I've merged the MR.

test/run-drun/timer.mo
mraszyk: I'm consistently getting this output:
```
=> update provisional_create_canister_with_cycles(record {settings = null; amount = null})
<= replied: (record {hymijyo = principal "rwlgt-iiaaa-aaaaa-aaaaa-cai"})
=> update install_code(record {arg = blob ""; kca_xin = blob "\00asm\01\00\00\00\0...
<= replied: ()
=> update go()
debug.print: YEP!
debug.print: DIM!
debug.print: ROOK!
debug.print: DIM!
debug.print: BATT!
debug.print: DIM!
debug.print: BATT!
debug.print: BATT!
debug.print: BATT!
debug.print: BATT!
debug.print: {count = 10}
<= replied: ()
```

nix/sources.json
ggreif: thanks! 80a8b03665e405a2bd653c988970ba7e47ed12ab

src/prelude/timers.mo
```diff
@@ -0,0 +1,149 @@
...
+            }
+        };
+        exp
+    }
+};
+
+// Function called by backend to run eligible timed actions.
+// DO NOT RENAME without modifying compilation.
+func @timer_helper() : async () {
+    func Array_init<T>(len : Nat,  x : T) : [var T] {
```
ggreif, line 35: use from prelude?

src/prelude/timers.mo
```diff
@@ -0,0 +1,149 @@
...
+    let now = (prim "time" : () -> Nat64)();
+    let exp = @nextExpiration @timers;
+    let prev = (prim "global_timer_set" : Nat64 -> Nat64) exp;
+
+    if (exp == 0) {
+        return
+    };
+
+    var gathered = 0;
+    let thunks : [var ?(() -> async ())] = Array_init(10, null); // we want max 10
```
ggreif, line 47: This is at most 20, due to IC self-send slots.

src/prelude/timers.mo
```diff
@@ -0,0 +1,149 @@
...
+                else ({ n with dopo = ?insert(n.dopo) })
+            }
+    };
+    @timers := ?insert(@prune(@timers));
+
+    let exp = @nextExpiration @timers;
+    if (exp == 0) @timers := null;
+    let prev = (prim "global_timer_set" : Nat64 -> Nat64) exp;
+    /*FIXME: this is expensive*/
+    if (prev != 0 and prev != 0 and exp > prev) {
```
ggreif, line 116: shouldn't happen

test/run-drun/timer-default.mo
```diff
@@ -0,0 +1,31 @@
...
+
+  public shared func go() : async () {
+     var attempts = 0;
+
+     ignore setTimer(1, false, func () : async () { count += 1; debugPrint "YEP!" });
+
+     while (count < max) {
+       ignore await raw_rand(); // yield to scheduler
+       attempts += 1;
+       if (attempts >= 8000 and count == 0)
```
ggreif, line 20: 100

src/lowering/desugar.ml
```diff
@@ -499,6 +501,14 @@ and build_actor at ts self_id es obj_typ =
          (match call_system_func_opt "heartbeat" es obj_typ with
           | Some call -> call
           | None -> tupE []);
+       timer =
+         (match call_system_func_opt "timer" es obj_typ with
+          | Some call -> call
+          | None when !Mo_config.Flags.global_timer ->
+            blockE
+              [ expD T.(callE (varE (var "@timer_helper" Mo_frontend.Typing.heartbeat_type)) [unit] (unitE())) ]
```
ggreif, line 509: later `callAwaitStarE`

src/prelude/timers.mo
```diff
@@ -0,0 +1,149 @@
...
+    };
+
+    for (f in futures.vals()) {
+        switch f { case (?f) try await f catch _ { }; case _ () }
+    };
+};
+
+var @lastId = 0;
+
+func @setTimer(delaySecs : Nat64, recurring : Bool, job : () -> async ()) : (id : Nat) {
```
dfinity-berestovskyy, line 93: We have `set_timer`() and `set_timer_interval()` implemented in [Rust CDK](https://github.com/dfinity/cdk-rs/pull/342/files#diff-b199dfe6f7df9ad54346b557c68334c9b3a75f22784b195d6696f80586469bffR172).

Do you think it makes sense to provide the same API for Motoko, i.e. `setTimer()` and `setTimerInterval()` for recurring tasks?

doc/md/language-manual.md
```diff
@@ -1209,12 +1209,15 @@ The declaration `<dec>` of a `system` field must be a manifest `func` declaratio
 | name          | type                                                          | description         |
 |---------------|---------------------------------------------------------------|---------------------|
 | `heartbeat`   | `() -> async ()`                                              | heartbeat action    |
+| `timer`       | `() -> async ()`                                              | timer action        |
```
ggreif, line 1212: feedback by @crusso: pass in the `setGlobalTimer` binding here

nix/sources.json
```diff
@@ -34,16 +34,16 @@
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     },
     "ic-hs": {
-        "branch": "master",
+        "branch": "joachim/ic-ref-run-heartbeat",
```
ggreif, line 37: testing only. move to `master` ASAP

src/exes/moc.ml
```diff
@@ -101,6 +101,7 @@ let argspec = [
   "-no-async", Arg.Clear Flags.async_lowering, " no async-lowering (with -iR)";
 
   "-no-link", Arg.Clear link, " do not statically link-in runtime";
+  "-no-timer", Arg.Clear Flags.global_timer, " do not create a global timer endpoint";
```
ggreif, line 104:
```suggestion
  "-no-timer", Arg.Clear Flags.global_timer, " do not create a global timer expiration endpoint";
```

src/prelude/prelude.mli
```diff
@@ -1,3 +1,3 @@
 val prelude : string
-val internals : string
-val prim_module : string
+val internals : bool -> string
```
ggreif, line 2: revert this

src/prelude/timers.mo
ggreif: unfortunately it is defined in `prim.mo` which is using this file

src/prelude/timers.mo
ggreif: I don't see why one would want two setting routines. I think the `setTimer`/`cancelTimer` pair is more economical.

doc/md/language-manual.md
ggreif: This is now a PR: #3614

crusso (commented):
I'm not too keen on the selective prelude hacks and the need for --no-timer  and wonder if this couldn't be simplified to a tiny system method and an optional base library.

Would it not be possible to add base/Timers.mo

``` motoko

module {
  public class Timer() {
       
       // adapted contents of prelude/timers.mo...
       public func setTimer(delaySecs : Nat64, recurring : Bool, job : () -> async ()) : (id : Nat) {
         ...
       };
      
       public func cancelTimer(id : Nat) : () {
         ...
       };

       public func onGlobalTimer(global_timer_set: Nat64 -> Nat64) {
         ...
       };
   }
}
```

And then have the actor elect to use this, or another implementation, by, invoking the Timer.onGlobalTimer callback in a generalized system method;


``` motoko
import Timers = "mo:base/Timers";

actor {

     let t = Timers.Timer();

     var ticks = 0; 
     let id = t.setTimer(1, true, func () : async () { ticks +=1 });
     t.cancelTimer(id);

     system func timer (global_timer_set: Nat64 -> Nat64) { // should this be Nat64 or Int64?
        t.onGlobalTimer(global_timer_set); // wire up the timer.
     };

}
```

That would decouple the timer code from the prelude, and localise the ability to call global_set_time to the system method, which passes it to the callback as a scoped capability.
(Of course, there's nothing to prevent the library from leaking that capability further, but at least it's not ambient by default).

One difficuly I can see is that your current setTimer/cancelTimer implementations call global_timer_set, but do they actually need to? One call is marked as unnecessary in your review comment, perhaps the other is too.

The other difficulty is that the timer is only scheduled if the canister calls ic0.global_timer_set(i64) with a non-zero value.
Perhaps the init method could do this if you declare the system method, or we give up on making global_time_set a scoped capability and add a `start` method to the Timer class...

Playground code sketch here:
https://m7sm4-2iaaa-aaaab-qabra-cai.raw.ic0.app/?tag=3314411626

src/codegen/compile.ml
crusso: I wouldn't special case anything. Just let the gc do it's thing.

src/pipeline/pipeline.ml
```diff
@@ -230,7 +230,12 @@ let check_builtin what src senv0 : Syntax.prog * stat_env =
 let prelude, initial_stat_env0 =
   check_builtin "prelude" Prelude.prelude Typing.initial_scope
 let internals, initial_stat_env =
-  check_builtin "internals" Prelude.internals initial_stat_env0
+  let aliasing = !Flags.experimental_field_aliasing in
```
crusso, line 233: I don't like the use of the aliasing feature in prod. Let's try to remove that, if only to keep pipeline as simple as possible

src/prelude/timers.mo
```diff
@@ -0,0 +1,149 @@
...
+    let expire = now + delayNanos;
+    let delay = if recurring ?delayNanos else null;
+    // only works on pruned nodes
+    func insert(n : ?@Node) : @Node =
+        switch n {
+            case null ({ var expire; id; delay; job; ante = null; dopo = null });
+            case (?n) {
+                assert n.expire != 0;
+                if (expire < n.expire) ({ n with ante = ?insert(n.ante) })
+                else ({ n with dopo = ?insert(n.dopo) })
```
crusso, line 107:
```suggestion
                if (expire < n.expire) {
                   n.before := ?insert(n.before) 
                }
                else { 
                  n.after := ?insert(n.after) 
                }
```
With the imperative (renamed) version, you could just insert in place, without re-allocating (I think).

src/prelude/timers.mo
```diff
@@ -0,0 +1,149 @@
...
+                    case (?delay) {
+                        // re-add the node
+                        let expire = n.expire + delay;
+                        // N.B. insert only works on pruned nodes
+                        func reinsert(m : ?@Node) : @Node = switch m {
+                            case null ({ n with var expire; ante = null; dopo = null });
+                            case (?m) {
+                                assert m.expire != 0;
+                                if (expire < m.expire) ({ m with ante = ?reinsert(m.ante) })
+                                else ({ m with dopo = ?reinsert(m.dopo) })
```
crusso, line 65:
```suggestion
                                if (expire < m.expire) {
                                  m.before :=  ?reinsert(m.before) 
                                }
                                else { 
                                  m.after := ?reinsert(m.after)
                                }
```
With the imperative (renamed) version, you could just insert in place, without re-allocating (I think).

test/run-drun/heartbeat.mo
```diff
@@ -22,7 +22,7 @@ actor {
      while (count < max) {
        ignore await raw_rand(); // yield to scheduler
        attempts += 1;
-       if (attempts >= 50 and count == 0)
+       if (attempts >= 200 and count == 0)
```
crusso, line 25: Spurious change? Revert?

test/run-drun/timer-trap.mo
```diff
@@ -0,0 +1,36 @@
...
+
+     ignore setTimer(1, true,
+                     func () : async () { count += 1; debugPrint "YEP!" });
+     ignore setTimer(1, false,
+                     func () : async () { count += 1; debugPrint "EEK!"; assert false });
+     ignore setTimer(1, false,
+                     func () : async () { count += 1; debugPrint "BEAM!"; throw error("beam me up Scotty!") });
+
+     while (count < max) {
+       ignore await raw_rand(); // yield to scheduler
```
crusso, line 23: What happens if you just do await async {} - do the timers never get a chance to run?

src/prelude/prelude.ml
```diff
@@ -1,3 +1,13 @@
 let prelude = [%blob "prelude.mo"]
-let internals = [%blob "internals.mo"]
-let prim_module = [%blob "prim.mo"]
+let internals' = [%blob "internals.mo"]
+let timers = [%blob "timers.mo"]
+let internals want_timers =
+  if want_timers
+  then internals' ^ timers
```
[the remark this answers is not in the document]
crusso: Even better, abstract out the timer stuff into a base library if possible. That might also make it easier to test, though I'm not sure about that since you also depend on Time.now as well as ic0.global_timer_set

crusso (changes requested):

crusso (changes requested):
Sorry about the delay.

src/prelude/timers.mo
ggreif: I removed this code.

src/prelude/prelude.ml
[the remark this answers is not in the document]
ggreif: Happy to do it if I figure out how to reliably call a function that resides in some user-defined location from `desugar.ml` or `compile.ml`. Currently I can't see a way...

src/prelude/prelude.mli
ggreif: done.

test/run-drun/timer-trap.mo
ggreif: That should work too. This code was copied from `test/run-drun/heartbeat.mo` and I didn't optimise this part.

src/codegen/compile.ml
```diff
@@ -3897,6 +3899,19 @@ module IC = struct
       edesc = nr (FuncExport (nr fi))
     })
 
+  let export_timer env =
+    assert !Flags.global_timer;
+    assert (E.mode env = Flags.ICMode || E.mode env = Flags.RefMode);
+    let fi = E.add_fun env "canister_global_timer"
+      (Func.of_body env [] [] (fun env ->
+        G.i (Call (nr (E.built_in env "timer_exp"))) ^^
+        GC.collect_garbage env))
```
crusso, line 3908: I guess we'll want to suppress GC here to. Lemme check with Ulan.

dfinity-berestovskyy: Right, just like with the heartbeats, timers do not support DTS.

src/prelude/timers.mo
dfinity-berestovskyy: The thread is marked as outdated, but I guess the comment still stands: we better to align with Rust CDK and either implement in both languages `setTImer` + `setTimerInterval` or `setTimer(..., recurring)`...

src/prelude/internals.mo
```diff
@@ -428,3 +428,153 @@ func @create_actor_helper(wasm_module_ : Blob, arg_ : Blob) : async Principal =
...
+    };
+
+    for (f in futures.vals()) {
+        switch f { case (?f) try await f catch _ { }; case _ () }
+    };
+};
+
+var @lastId = 0;
+
+func @setTimer(delaySecs : Nat64, recurring : Bool, job : () -> async ()) : (id : Nat) {
```
dfinity-berestovskyy, line 529: My previous comment was marked as outdated, but I guess the comment still stands: we better to align with Rust CDK and either implement in both languages `setTImer()` + `setTimerInterval()` or `setTimer(..., recurring)`...

ggreif: So this PR is not the last word about user-facing interfaces. This just adds the _primitives_ to a (basically) hidden internal module for the compiler. As such less is more. What we choose to export from the base library is a different issue and we may go with the Rust-like naming. But I suspect that this is a non-issue as I don't expect that both Rust and Motoko code will surface in the same project soon :-)

src/codegen/compile.ml
ByronBecker: @crusso @ggreif 

For the current project that I'm working on, the canister timer will kick off a series of inter-canister calls to retrieve a considerable amount of data and update the state of the canister. This heartbeat initiated "cron job" is the major workhorse of the canister.

@ulan Will heartbeat/timers ever support DTS?

ggreif: @ByronBecker the `job` of an expired timer will always be executed in a context that can do DTS.

src/prelude/prelude.mli
```diff
@@ -1,3 +1,3 @@
 val prelude : string
 val internals : string
-val prim_module : string
+val prim_module : bool -> string
```
ggreif, line 3: make this a keyword input argument?

test/run-drun/generic-tail-rec.mo
```diff
@@ -1,7 +1,7 @@
 import Prim "mo:⛔";
 /* test generic self-tail-recursion */
 
-let bound:Int = 100000;
+let bound:Int = 10000;
```
ggreif, line 4: undo this? (only relevant for M1)

test/run-drun/generic-tail-rec.mo
ggreif: undo!

test/run-drun/heartbeat.mo
ggreif: yes, I have disabled the `ic-ref` test path

src/mo_frontend/typing.mli
```diff
@@ -10,3 +10,6 @@ val infer_prog : scope -> Syntax.prog -> (typ * scope) Diag.result
 val check_lib : scope -> Syntax.lib -> scope Diag.result
 val check_actors : scope -> Syntax.prog list -> unit Diag.result
 val check_stab_sig : scope -> Syntax.stab_sig -> (field list) Diag.result
+
+val heartbeat_type : typ
+val timer_type : typ
```
ggreif, line 15: no need to export

ggreif, line 15:
```suggestion
val heartbeat_type : typ
```

src/prelude/timers.mo
```diff
@@ -0,0 +1,149 @@
+// default timer mechanism implementation
+// fundamental node invariant: max_exp ante <= expire <= min_exp dopo
+// corollary: if expire == 0 then the ante is completely expired
+//
+type @Node = { var expire : Nat64; id : Nat; delay : ?Nat64; job : () -> async (); ante : ?@Node; dopo : ?@Node };
```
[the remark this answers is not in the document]
ggreif: All these pieces should be resolved by now.

src/prelude/timers.mo
```diff
@@ -0,0 +1,149 @@
...
+    let prev = (prim "global_timer_set" : Nat64 -> Nat64) exp;
+
+    if (exp == 0) {
+        return
+    };
+
+    var gathered = 0;
+    let thunks = Array_init<?(() -> async ())>(10, null); // we want max 10
+
+    func gatherExpired(n : ?@Node) = switch n {
```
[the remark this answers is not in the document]
ggreif: I'll leave that to an improvement PR. So the static scope checks don't apply to `async*`, right?

src/prelude/timers.mo
```diff
@@ -0,0 +1,149 @@
...
+    let now = (prim "time" : () -> Nat64)();
+    let exp = @nextExpiration @timers;
+    let prev = (prim "global_timer_set" : Nat64 -> Nat64) exp;
+
+    if (exp == 0) {
+        return
+    };
+
+    var gathered = 0;
+    let thunks = Array_init<?(() -> async ())>(10, null); // we want max 10
```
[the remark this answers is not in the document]
ggreif: For example queue limits. But I had to set some limit as I wanted to use an array and not a linked list. What doesn't fit into 10 elements will be handles at the next (ASAP) expiration.

src/prelude/internals.mo
```diff
@@ -435,3 +435,151 @@ func @create_actor_helper(wasm_module_ : Blob, arg_ : Blob) : async Principal =
...
+        gathered += 1;
+      };
+      gatherExpired(n.post);
+    }
+  };
+
+  gatherExpired(@timers);
+
+  for (k in thunks.keys()) {
+    ignore switch (thunks[k]) { case (?thunk) ?thunk(); case _ null };
```
ggreif, line 525: @crusso So what happens if this `throw`s? maybe we can just run the `jobs on the fly while gathering and `catch` when the queue is full. Otherwise expunge.

src/codegen/compile.ml
```diff
@@ -3913,7 +3920,11 @@ module IC = struct
     let fi = E.add_fun env "canister_global_timer"
       (Func.of_body env [] [] (fun env ->
         G.i (Call (nr (E.built_in env "timer_exp"))) ^^
-        GC.collect_garbage env))
+        (* TODO(3622)
+           Until DTS is implemented for heartbeats, don't collect garbage here,
```
crusso, line 3924:
```suggestion
           Until DTS is implemented for timers, don't collect garbage here,
```

src/codegen/compile.ml
```diff
@@ -3913,7 +3920,11 @@ module IC = struct
...
+           just record mutator_instructions and leave GC scheduling to the
+           already sheduled async message running `system` function `heartbeat` *)
```
crusso, line 3926:
```suggestion
           already sheduled async message running `system` function `timer` *)
```

src/mo_frontend/typing.ml
```diff
@@ -207,12 +207,16 @@ let infer_mut mut : T.typ -> T.typ =
 
 (* System method types *)
 
-let heartbeat_type = T.(Func (Local, Returns, [scope_bind], [], [Async (Var (default_scope_var, 0), unit)]))
+let heartbeat_type = T.(Func (Local, Returns, [scope_bind], [], [Async (Fut, Var (default_scope_var, 0), unit)]))
+
+let timer_type = T.(Func (Local, Returns, [scope_bind],
+                          [Func (Local, Returns, [], [Prim Nat64], [])],
+                          [Async (Fut, Var (default_scope_var, 0), unit)]))
```
crusso, line 214:
```suggestion
let heartbeat_type = 
  T.(Func (Local, Returns, [scope_bind], [], [Async (Fut, Var (default_scope_var, 0), unit)]))

let timer_type = 
  T.(Func (Local, Returns, [scope_bind],
    [Func (Local, Returns, [], [Prim Nat64], [])],
    [Async (Fut, Var (default_scope_var, 0), unit)]))
```

src/prelude/internals.mo
crusso: Yes, I think you could do that - no need for the array and second traversal then.

src/prelude/internals.mo
```diff
@@ -448,140 +448,138 @@ func @call_raw(p : Principal, m : Text, a : Blob) : async Blob {
...
-        futures[k] := switch (thunks[k]) { case (?thunk) ?thunk(); case _ null };
-    };
+  gatherExpired(@timers);
 
-    for (f in futures.vals()) {
-        switch f { case (?f) try await f catch _ { }; case _ () }
-    };
+  for (k in thunks.keys()) {
+    ignore switch (thunks[k]) { case (?thunk) ?thunk(); case _ null };
+  }
```
crusso, line 526:
```suggestion
  for (o in thunks.keys()) {
    switch (o) { case (?thunk) { ignore ?thunk()} ; case _ {} };
  }
```
(if you keep the array, this might be sweeter).

src/prelude/prelude.ml
```diff
@@ -2,7 +2,7 @@ let prelude = [%blob "prelude.mo"]
 let internals = [%blob "internals.mo"]
 let timers_api = [%blob "timers-api.mo"]
 let prim_module' = [%blob "prim.mo"]
-let prim_module want_timers =
-  if want_timers
+let prim_module ~timers:want =
+  if want
```
crusso, line 6:
```suggestion
let prim_module ~timers:required =
  if required
```

src/prelude/internals.mo
```diff
@@ -448,140 +448,138 @@ func @call_raw(p : Principal, m : Text, a : Blob) : async Blob {
...
+        exp := @nextExpiration(n.post)
+      }
+    };
+    exp
+  }
 };
 
 // Function called by backend to run eligible timed actions.
 // DO NOT RENAME without modifying compilation.
 func @timer_helper() : async () {
```
crusso, line 477:
```suggestion
func @timer_helper() : async () {
```
It would be nice if the helper took the global_timer_set capability as an argument and used it, just as we are expecting clients to do, but I guess that can happen in a separate PR if necessary

src/prelude/internals.mo
crusso: I wonder what will happen with upgrades if the timers haven't been stopped. All of these async calls will extend the continuation table until completed, if we don't await them.

Ideally, motoko would offer a way to do oneway scheduling of async blocks, or local oneway functions to avoid setting up the continuations, but we don't have that facility yet.

src/prelude/prelude.mli
ggreif: done

src/mo_frontend/typing.mli
ggreif: done

src/prelude/internals.mo
ggreif: IIRC, the spec says that the global timer gets reset to 0 on upgrades.

crusso (approved):
Just a few comments.
I guess one remaining comment is how does this API compare to the Rust one, e.g. in terms of timer resolution (delay in nanoseconds or seconds, like here). I guess we should aim to keep them as similar as possible for ease of writing dual language examples.

src/prelude/internals.mo
```diff
@@ -435,3 +435,151 @@ func @create_actor_helper(wasm_module_ : Blob, arg_ : Blob) : async Principal =
...
+      gatherExpired(n.pre);
+      if (n.expire[0] > 0 and n.expire[0] <= now and gathered < thunks.size()) {
+        thunks[gathered] := ?(n.job);
+        switch (n.delay) {
+          case (?delay) if (delay != 0) {
+            // re-add the node
+            let expire = n.expire[0] + delay;
+            // N.B. insert only works on pruned nodes
+            func reinsert(m : ?@Node) : @Node = switch m {
+              case null ({ n with expire = [var expire]; pre = null; post = null });
```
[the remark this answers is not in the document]
ggreif: This was _not_ that line, just search for the other `with`s!

ggreif:
> I guess one remaining comment is how does this API compare to the Rust one, e.g. in terms of timer resolution (delay in nanoseconds or seconds, like here). I guess we should aim to keep them as similar as possible for ease of writing dual language examples.

Cf. https://github.com/dfinity/examples/pull/419/files for the Rust example.

test/run-drun/upgrades/upgrade4.mo
```diff
@@ -0,0 +1,15 @@
...
+  stable let c : Text = do { assert false; loop {}};
+  stable var i : Nat = do { assert false; loop {}};
+  public func inc() { i += 1; };
+  public query func check(n : Int) : async () {
+    assert (c.size() == 3);
+    assert (i == n);
+  };
+  system func timer(set : Nat64 -> ()) : async () {
+      Prim.debugPrint ("timer endpoint in 4");
+      set 0
```
ggreif, line 13: needed for (outdated?) `drun` only:
``` diff
diff --git a/test/run-drun/ok/upgrades.drun.ok b/test/run-drun/ok/upgrades.drun.ok
index e0de5c306..a20a83383 100644
--- a/test/run-drun/ok/upgrades.drun.ok
+++ b/test/run-drun/ok/upgrades.drun.ok
@@ -32,5 +32,25 @@ debug.print: init'ed 4
 debug.print: timer endpoint in 4
 Ok: Reply: 0x4449444c0000
+debug.print: timer endpoint in 4
+debug.print: timer endpoint in 4
+debug.print: timer endpoint in 4
+debug.print: timer endpoint in 4
+debug.print: timer endpoint in 4
+debug.print: timer endpoint in 4
+debug.print: timer endpoint in 4
+debug.print: timer endpoint in 4
+debug.print: timer endpoint in 4
+debug.print: timer endpoint in 4
+debug.print: timer endpoint in 4
 Ok: Reply: 0x4449444c0000
+debug.print: timer endpoint in 4
+debug.print: timer endpoint in 4
+debug.print: timer endpoint in 4
+debug.print: timer endpoint in 4
+debug.print: timer endpoint in 4
+debug.print: timer endpoint in 4
+debug.print: timer endpoint in 4
+debug.print: timer endpoint in 4
+debug.print: timer endpoint in 4
```

test/run-drun/timer-default.mo
ggreif: ab19f6f55

doc/md/language-manual.md
```diff
@@ -1220,12 +1220,15 @@ The declaration `<dec>` of a `system` field must be a manifest `func` declaratio
...
 | `preupgrade`  | `() -> ()`                                                    | pre upgrade action  |
 | `postupgrade` | `() -> ()`                                                    | post upgrade action |
 
 -   `heartbeat`, when declared, is called on every Internet Computer subnet **heartbeat**, scheduling an asynchronous call to the `heartbeat` function. Due to its `async` return type, a heartbeat function may send messages and await results. The result of a heartbeat call, including any trap or thrown error, is ignored. The implicit context switch means that the time the heartbeat body is executed may be later than the time the heartbeat was issued by the subnet.
 
+-   `timer`, when declared, is called as a response of the canister global timer's expiration. The canister's global timer can be manipulated with the passed-in function argument of type `Nat64 -> ()` upon which libraries can build their own abstractions. When not declared (and in absence of the `-no-timer` flag), this system action is provided with default implementation by the compiler (additionally `setTimer` and `cancelTimer` are provided as primitives).
```
crusso, line 1230: So is the argument here in nanoseconds (like the system API) or seconds, like your prims?

src/prelude/internals.mo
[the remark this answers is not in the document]
crusso: Ok

crusso:
> > I guess one remaining comment is how does this API compare to the Rust one, e.g. in terms of timer resolution (delay in nanoseconds or seconds, like here). I guess we should aim to keep them as similar as possible for ease of writing dual language examples.
> 
> Cf. https://github.com/dfinity/examples/pull/419/files for the Rust example.

Right, they are constructing the Duration from seconds, but I think you can also construct from nanoseconds - plus it would be best if the unit was consistent with whatever the capability is expecting, I think. I assume Time.now is in nanoseconds?

doc/md/language-manual.md
ggreif: This is in nanos. (Not a convenience interface.)

crusso:
> > I assume `Time.now` is in nanoseconds?
> 
> Yes. I am a bit torn whether to make the `setTimer` primitive nanoseconds-based. I am pretty sure we will never see (significantly) sub-second expiration granularity on the IC, so seconds looks like the correct unit of measure. OTOH the system call is defined in nanos, so we could easily support that. But I have seen lots of confusion where people were passing 10 for a nanosecond argument and wondering that it expired immediately. In base we should have a `setTimer(duration : { #seconds : Nat64; #nanoseconds : Nat64)`

I would keep the low-level API as close to the System API as possible (perhaps even returning the Nat64 in case it ever gets used for something useful), but I like the idea of using the variant in Timer.setTimer. That does require nanoseconds for the prims though, right, unless you want to round.

dfinity-berestovskyy:
IMO the `ic0.global_timer_set()` argument type should be exactly the same what `ic0.time()` returns, i.e. an expression like `ic0.global_timer_set(ic0.time() + 1)` should just work (in Motoko primitives).

We can't set an absolute timer without knowing the current time. So it would be counter-intuitive if we get the current time, and then we need to convert it to feed into the timer...

crusso:
slack > What will happen if a user specifies -no-timer but uses mo:base/Timer.mo that relies on prims? a compiler crash?

If so, why not just always define the prims, but compile them to trap on use  when --no-timer is specified. That would simplify other logic elsewhere, i.e. the prelude construction.

src/codegen/compile.ml
ulan: > Will heartbeat/timers ever support DTS?

Sorry, I missed this question. +1 to Gabor's answer. I am also implement DTS support for the initial heartbeat/global_timer endpoint.

doc/md/language-manual.md
ggreif, line 1230:
```suggestion
-   `timer`, when declared, is called as a response of the canister global timer's expiration. The canister's global timer can be manipulated with the passed-in function argument of type `Nat64 -> ()` (taking an absolute time in nanoseconds) upon which libraries can build their own abstractions. When not declared (and in absence of the `-no-timer` flag), this system action is provided with default implementation by the compiler (additionally `setTimer` and `cancelTimer` are available as primitives).
```

ggreif:
> low-level API as close to the System API as possible

Done that. PTAL.

src/prelude/internals.mo
ggreif: internal are under out control, I don't see why we shouldn't use the most performant way to implement this.

src/prelude/internals.mo
ggreif: I agree but we are not there yet. I leave tis to a follow-up PR.

test/run-drun/generic-tail-rec.mo
ggreif: b42ac5abf

test/perf/Makefile
```diff
@@ -1,4 +1,5 @@
 RUNFLAGS = -p
+export EXTRA_MOC_ARGS = -no-timer
```
ggreif, line 2: this avoids the 3% size increase

test/run-drun/generic-tail-rec.mo
ggreif: e5b44745b

test/run-drun/heartbeat.mo
ggreif: 200 looks like a good compromise

test/run-drun/ok/timer-trap.drun-run.ok
ggreif: this is working as expected

crusso (approved):
LGTM

PR #3547, opened 2022-11-03
chore: remove use of `set-output` from CI

ggreif:
Removes warnings as noted in
https://github.com/dfinity/motoko/pull/3483#issue-1406029328

resolved according to
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands

_N.B._: we do not use `save-state`, so nothing to do on that front.

Also changes some convoluted shell substitutions to `perl`.


nomeata (approved):

ggreif (commented):
Added another tweak...

PR #3551, opened 2022-11-04
Backend: Correctly join Const and UnboxedTuple reps

nomeata:
previously the code assumed that we never join a constant value and a unboxed tuple. If it tried, it would print a warning, and simply use one of the two.

@crusso reports that this was observed in the wild, and indeed such joins are easy to trigger.

I don’t recall why I thought that this combination could not occur. In fact, it seems that the best stackrep when a const value meets a non-const value is to use that other value's stack representation. So lets put that in.


crusso:
Can you explain a little why this is sound? I expect it has something to do with unifying the stack representation at control flow join points, so would expect the stacks to be required to have the same shape.

crusso:
I can confirm that this removes all the warnings from the real-world scenario I had.... nice! (Though I can't actually run that code to see if it works...)

nomeata:
This is part of the code that tries to figure out a “good” stack representation for a given value. If both branches of an if have the same stack representation, we can obviously just use that. If not, we have to make a choice – but mostly no choice is “wrong”, as the backend will insert the right coercions (via `StackRep.adjust`). And indeed, if you run my test case with the old code, `adjust` would complain about `Unknown stack_rep conversion UnboxedTuple 2 -> Const`.

A good way to think about this whole stackrep business is that it is a bidirectional type inference algorithm that in a type system with coercions.

Because of these coercions, both branches can have different stack representation, and the join (or “join”, maybe a bad name) can be still different.

PR #3554, opened 2022-11-05
Minor cleanup: Partially revert because of performance degrade

luc-blaeser:
One optimization suggested by clippy of collapsing nested `if` degrades performance by 1%. So reverting this change and disabling the clippy warning with a corresponding comment.


luc-blaeser:
Next week, I will also analyze why the Rust compiler makes a difference between && and nested ifs. Seems like a Rust compiler issue...

test/bench/ok/heap-32.drun-run.ok
```diff
@@ -1,5 +1,5 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
-debug.print: (50_227, +91_377_860, 1_453_205_851)
-debug.print: (50_070, +102_586_000, 1_526_830_363)
+debug.print: (50_227, +91_377_860, 1_446_975_760)
```
nomeata, line 3: Is that test outputting instruction counts? Having to update expected test output that often is bad ergonomics, isn't it?

nomeata (approved):
Pretty steep cost for such a small change. I wonder how much similar low hanging fruit is out there

test/bench/ok/heap-32.drun-run.ok
luc-blaeser: Yes, you are right.

luc-blaeser:
Probably, there are more. But it seems more like a Rust compiler performance issue (because short-circuit logic should be equivalent to nested if-statements). I will analyze this.

PR #3560, opened 2022-11-09
experimentally try using the GH-generated token

ggreif:
The `NIV_UPDATER_TOKEN` is not up to the task pushing to the repo. Let's see if this gets better.

See: https://dev.to/github/the-githubtoken-in-github-actions-how-it-works-change-permissions-customizations-3cgp


nomeata (approved):

ggreif:
In retrospective this was not the reason, see #3549.

PR #3561, opened 2022-11-09
chore: enable building `moc` on the M1

ggreif:
Testing mostly works too.

_Note_: this only fixes compilation of the `aarch64-darwin` target, which is otherwise not supported by the replica team. X86-64 compilation is unaffected. The compilation failure is well-documented on the `rocksdb` GH page.

We should regularly examine whether the buggy package has been bumped on the replica side, and if so, we can probably remove this kludge.


ggreif:
We have 2 fails in the Haskell property tests:
```
    expected failures:              FAIL (10.23s)
      *** Failed! Assertion failed (after 16 tests and 5 shrinks):
      let _ = (++++((Prim.nat32ToNat ((((65534) : Nat32) % ((0) : Nat32)) * (((0) : Nat32) % ((4294967295) : Nat32))))));
      Use --quickcheck-replay=30960 to reproduce.
      Use -p '/expected failures/' to rerun this test only.
obj-ext-await: [tc] [run] [run-ir] [run-low] [comp] [comp-ref] [valid] [valid-ref] [drun-run]
oom: [tc] [run] [run-ir] [run-low] [comp] [comp-ref] [valid] [valid-ref] [drun-run]
rts-stats2: [tc] [comp] [comp-ref] [valid] [valid-ref] [drun-run]
optimise-for-array-eff: [tc] [run] [run-ir] [run-low] [comp] [comp-ref] [valid] [valid-ref] [drun-run]
```
and
```
    expected successes:             FAIL (23.00s)
      *** Failed! Assertion failed (after 42 tests and 8 shrinks):
      assert ((++++((Prim.nat32ToNat (((4294967295) : Nat32) / ((4294967295) : Nat32))))) == 1);
      
      STDERR:
      thread '<unnamed>' panicked at 'assertion failed: first_mov_emitted', cranelift/codegen/src/isa/aarch64/lower/isle.rs:221:9
      note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
      
      Use --quickcheck-replay=908584 to reproduce.
      Use -p '/expected successes/' to rerun this test only.
```

nix/drun.nix
```diff
@@ -17,6 +17,26 @@ pkgs:
...
 
+      patchPhase = ''
+      cd ../drun-vendor.tar.gz
+      ls -l librocksdb-sys/build.rs
+      patch librocksdb-sys/build.rs << EOF
+@@ -118,6 +118,8 @@
+         config.define("OS_MACOSX", Some("1"));
+         config.define("ROCKSDB_PLATFORM_POSIX", Some("1"));
+         config.define("ROCKSDB_LIB_IO_POSIX", Some("1"));
++        config.define("isSSE42()", Some("0"));
```
ggreif, line 28: do this for `aarch64` only

ggreif: fixed in 54fd3f4d9d06ee08a900b77d5bd48eb34073940a

nomeata (approved):
Rubber stamp

nix/drun.nix
```diff
@@ -17,6 +17,28 @@ pkgs:
 
       cargoSha256 = "sha256-dhDXhVNAAHzLRHdA6MpIGuoY76UhiF4ObeLO4gG/wo4=";
 
+      patchPhase = ''
+      cd ../drun-vendor.tar.gz
+      ls -l librocksdb-sys/build.rs
```
ggreif, line 22: remove

PR #3563, opened 2022-11-10
chore: use the mighty `NPM_TOKEN`

ggreif:
this is already working for vulnerability mitigation, and has proven itself.


kentosugama (approved):

ggreif:
In retrospective this was not the reason, see #3549.

PR #1459, opened 2020-05-02
Experiment: Enable multi-value

nomeata:
drun now uses wasmtime-0.15, which supports multi-value.

 * [x] bump our wasmtime to 0.15 as well
 * [x] pass `--enable-multi-value` to it
 * [x] implement multi-value in `winter` (https://github.com/dfinity-side-projects/winter/commit/9f9b702c3f06840202771940c04ef19595d07ea3)
 * [x] bump `winter` and `ic-ref` (https://github.com/dfinity-lab/motoko/pull/1462)
 * [x] in `dfinity` enable multi-value in `parity_wasm` parser: https://github.com/dfinity-lab/dfinity/pull/5321
 * [ ] in `dfinity` enable multi-value in wasmtime: https://github.com/dfinity-lab/dfinity/pull/5324
 * [x] `parity_wasm` parser support it upstream: https://github.com/paritytech/parity-wasm/pull/288
 * [x] enable multi-value unconditionally in the backend
 * [x] bump to `wasm` v1.1.1 (https://github.com/dfinity-lab/motoko/pull/1467)
 * [x] recover multi-value support in the Wasm encoder (undo parts of acae13f1b1b660c6d1920951e607e7110a57908b, needs #1467)
 * [ ] remove fake multi-value code

We can test all that on branches. Likely, though, allowing multi-value needs to
be part of a Public Spec release, so this can’t be the default for a while.

PR #3566, opened 2022-11-10
Multi-value emulation: Use globals, not allocated tuples

nomeata:
So far, the IC does not support wasm using multi-value returns from functions and blocks (see #1459). The backend was originally written with that feature in mind, and later we added a work-around (module `FakeMultiVal`).

For functions we work around it by storing multiple values in fixed globals, and reading them after the call. No allocations needed.

For blocks with multiple values, however, we allocate a tuple (array) on the heap, write the values there, and read them later. This adds unnecessary heap churn and fills up memory faster.

So this PR uses the mechanism from the functions also for blocks.

Goes well with #3556

Here is a (particularly bad) example:

In
```motoko
func multi_value_blocks(n : Nat) : (Nat, Nat) {
  return (
    if (n == 0) {
      returns_tuple()
    } else {
      if (n == 1) {
        returns_tuple()
      } else {
        returns_tuple()
      }
   }
 );
};
```
Now no allocation happens (and all the reads and writes to the global variables are eliminated due to #3556):

```diff
   (func $multi_value_blocks (type 1) (param $clos i32) (param $n i32)
     local.get $n
     i32.const 0
     call $B_eq
-    if (result i32)  ;; label = @1
+    if  ;; label = @1
       i32.const 0
       call $returns_tuple
-      global.get 4
-      global.get 5
-      call $to_2_tuple
     else
       local.get $n
       i32.const 2
       call $B_eq
-      if (result i32)  ;; label = @2
+      if  ;; label = @2
         i32.const 0
         call $returns_tuple
-        global.get 4
-        global.get 5
-        call $to_2_tuple
       else
         i32.const 0
         call $returns_tuple
-        global.get 4
-        global.get 5
-        call $to_2_tuple
       end
     end
-    call $from_2_tuple
     return)
```

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -749,6 +749,11 @@ module FakeMultiVal = struct
       G.i (GlobalGet (nr (global env (n - i))))
     ) tys
 
+  (* A drop-in replacement for E.if_ *)
+  let if_ env bt thn els =
+    E.if_ env (ty bt) (thn ^^ store env bt) (els ^^ store env bt) ^^
+    load env bt
+
 end (* FakeMultiVal *)
 
 module Func = struct
@@ -9148,13 +9153,12 @@ and compile_exp (env : E.t) ae exp =
     let sr1, code1 = compile_exp env ae e1 in
     let sr2, code2 = compile_exp env ae e2 in
     let sr = StackRep.join sr1 sr2 in
-    let bt = StackRep.to_block_type env sr in
     sr,
     code_scrut ^^
-    E.if_ env (FakeMultiVal.ty bt)
-      (code1 ^^ StackRep.adjust env sr1 sr ^^ FakeMultiVal.store env bt)
-      (code2 ^^ StackRep.adjust env sr2 sr ^^ FakeMultiVal.store env bt) ^^
-   FakeMultiVal.load env bt
+    FakeMultiVal.if_ env
+      (StackRep.to_block_type env sr)
+      (code1 ^^ StackRep.adjust env sr1 sr)
+      (code2 ^^ StackRep.adjust env sr2 sr)
   | BlockE (decs, exp) ->
     let captured = Freevars.captured_vars (Freevars.exp exp) in
     let ae', codeW1 = compile_decs env ae decs captured in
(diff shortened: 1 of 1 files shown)


crusso (commented):
LGTM, but @ggreif should take a look too.

Will this automatically kick in for switches too, assuming we reduce those to `if_`?

ggreif:
> Will this automatically kick in for switches too, assuming we reduce those to `if_`?

Last time I looked the `switch` result was communicated back via Wasm locals (set by assignment).

nomeata:
> Last time I looked the switch result was communicated back via Wasm locals (set by assignment).

Correct:
```ocaml
  | SwitchE (e, cs) ->
    SR.Vanilla,
    let (set_i, get_i) = new_local env "switch_in" in
    let (set_j, get_j) = new_local env "switch_out" in

    let rec go env cs = match cs with
      | [] -> CanFail (fun k -> k)
      | {it={pat; exp=e}; _}::cs ->
          let (ae1, code) = compile_pat_local env ae pat in
          orElse ( CannotFail get_i ^^^ code ^^^
                   CannotFail (compile_exp_vanilla env ae1 e) ^^^ CannotFail set_j)
                 (go env cs)
          in
      let code2 = go env cs in
      code1 ^^ set_i ^^ orTrap env code2 ^^ get_j
```

I don’t remember why it was hard to return the result on the stack. Maybe the pattern matching code DSL (`CannotFail` etc.) simply wasn’t set up for it. No need to change that now.

crusso:
Is there any chance of this code going wrong if you are already in the midst of producing a flattened tuple on the stack?

nomeata:
You mean the Switch code or the IF code?

But I don’t think so; whatever is below the stack is untouched by the generated code.

crusso:
> You mean the Switch code or the IF code?
> 
> But I don’t think so; whatever is below the stack is untouched by the generated code.

I meant this PR's code, but I guess these are just extra registers and never live long after a call but always loaded onto the stack

crusso:
> You mean the Switch code or the IF code?
> 
> But I don’t think so; whatever is below the stack is untouched by the generated code.

I meant this PR's code, but I guess these are just extra registers and never live long after a call but always loaded onto the stack

nomeata:
Right, they are always short-lived, from the end of a block or function to directly after the block (or after the call).

crusso (approved):
Nice one!

PR #3567, opened 2022-11-11
chore: automatise `docusaurus/package-lock.json`

ggreif:
as suggested in https://github.com/dfinity/motoko/pull/3559#issuecomment-1308730139

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -62,8 +62,10 @@ pull_request_rules:
     conditions:
       - files=doc/docusaurus/package-lock.json
       - "#files=1"
-      - label=automerge-squash
     actions:
       review:
         type: APPROVE
         message: A change to `doc/docusaurus/package-lock.json`? I can handle that!
+      label:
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -58,3 +58,12 @@ pull_request_rules:
       review:
         type: APPROVE
         message: A change to `Changelog.md`? I can handle that!
+  - name: Auto-approve dependabot PR
+    conditions:
+      - files=doc/docusaurus/package-lock.json
+      - "#files=1"
+      - label=automerge-squash
```
nomeata, line 65: Not sure if these PRs get that label. I'd assume not, but you can use mergify to label them.

ggreif: Oh, I thought that entry is putting a label on it. Looking further...

ggreif: b28f41a34

chenyan-dfinity (approved):

PR #3573, opened 2022-11-14
experiment: merge and revive do async block proposal

crusso:
Revives #2245 from 2 years ago, addressing #1482.

Note this is just an experiment, to see if we can integrate with the interim changes to yield on completed as well as incompleted futures. The answer seems yes, but we still make awaits have fuzzy commit points.

### Do-async block

The do-async block, `do async <block>,` returns a future containing the result of `<block>`.
Unlike an `async` expression, `do async <block>` evaluates `<block>` __synchronously__ at first,  switching to asynchronous computation _only_ at the first await of a _pending_ future within `<block>`.

The expression `do async <block>` has type `async T` provided

* `<block>` has type `T`;
* the expression occurs in an asynchronous context (i.e. in the body of a (non-`query`) `shared` function, asynchronous function, `async` expression or another `do async` block).

Do-async blocks can be nested, with the exit of a `return` or `throw` determined by lexical scoping.

It is a compile-time error for `<block>` to contain a `break` to a label declared outside the do-async block.

The expression `do async <block>` evaluates by entering the evaluation of `<block>`, executing synchronously  --  just like a `do <block>` --
until the first await of a pending future, if any.
Unless evaluation traps, it returns a future holding the actual or eventual result of `<block>`.

During the synchronous execution of `<block>`:

* a `return` expression exits the block with a completed future, storing the returned value;
* a `throw` expression resumes execution from the nearest enclosing `try ... catch ...` handler within `<block>` or, if there is none, exits the do-async block
  with a failed future storing the error;
* a trap causes `do async <block>` to trap;
* an `await` expression of
** a successfully completed future continues execution with the contents of that future, or
** a failed, completed future propagates the error contents of the future as if thrown (see above), or
** a pending future exits the do-async block with a new, pending future whose completion with the result of `<block>` is suspended, waiting on the result of the awaited future.

:::note

Evaluation of a do-async block is synchronous while possible, switching to asynchronous when necessary.
By contrast, evaluation of the similar `async <block-or-exp>` (without `do`) immediately returns a pending future and always computes the result of `<block-or-exp>` _asynchronously_,
never blocking evaluation of the enclosing expression.

:::

:::note

Do-async blocks are useful for abstracting asynchronous code into re-useable functions, without introducing redundant and expensive `async` expressions just to satisfy the type checker.

:::

# A better way (perhaps)

A better solution,avoing fuzzy commits and allowing better abstraction, might be to introduce a separate type:

```
type ++= await t (no restrictions on t)
```

and operations

```
<exp> ++=
| do await <exp>    (intro) 
| await in <exp>    (elim)
```
Typing:

```
e : t
-------------------------
do await e : await t

e : await t
-------------
await in e : t
```

Hence `do await e` just introduces a thunk and `await in e` forces the thunk, making explicit that it may await.


luc-blaeser:

PR #3574, opened 2022-11-14
viper: conditionalise lexer/grammar

ggreif:
Enable verification-specific keywords (and grammar productions) under two circumstances:
- when `--viper` option is given to `moc`
- when the environment variable `MOC_UNLOCK_VERIFICATION` is set

The latter way is needed for running also the backend tests (and the env-var is set by by the `Makefile` centrally).

The grammar productions for the ad-hoc `assert:*` productions are specifically disabled when not in verification mode.

Tests added to check the absence of verification-mode constructs in normal mode.

### DIFF
--- a/src/mo_frontend/lexer_lib.ml
+++ b/src/mo_frontend/lexer_lib.ml
@@ -5,13 +5,16 @@
 *)
 type mode = {
   privileged : bool;
+  verification : bool;
 }
 
 let mode : mode = {
   privileged = Option.is_some (Sys.getenv_opt "MOC_UNLOCK_PRIM");
+  verification = false;
 }
 
 let mode_priv : mode = { mode with privileged = true }
+let mode_verification : mode = { mode with verification = true }
 
 
 exception Error of Source.region * string
--- a/src/mo_frontend/lexer_lib.ml
+++ b/src/mo_frontend/lexer_lib.ml
@@ -10,7 +10,7 @@ type mode = {
 
 let mode : mode = {
   privileged = Option.is_some (Sys.getenv_opt "MOC_UNLOCK_PRIM");
-  verification = false;
+  verification = Option.is_some (Sys.getenv_opt "MOC_UNLOCK_VERIFICATION");
 }
 
 let mode_priv : mode = { mode with privileged = true }
--- a/src/pipeline/pipeline.ml
+++ b/src/pipeline/pipeline.ml
@@ -237,7 +237,7 @@ let internals, initial_stat_env =
 
 let parse_stab_sig s name  =
   let open Diag.Syntax in
-  let mode = Lexer.{mode with privileged = false} in
+  let mode = Lexer.{privileged = false; verification = false} in
   let lexer = Lexing.from_string s in
   let parse = Parser.Incremental.parse_stab_sig in
   let* sig_ = generic_parse_with mode lexer parse name in
@@ -247,7 +247,7 @@ let parse_stab_sig_from_file filename : Syntax.stab_sig Diag.result =
   let ic = Stdlib.open_in filename in
   Diag.finally (fun () -> close_in ic) (
     let open Diag.Syntax in
-    let mode = Lexer.{mode with privileged = false} in
+    let mode = Lexer.{privileged = false; verification = false} in
     let lexer = Lexing.from_channel ic in
     let parse = Parser.Incremental.parse_stab_sig in
     let* sig_ = generic_parse_with mode lexer parse filename in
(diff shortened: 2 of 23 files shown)


src/pipeline/pipeline.ml
```diff
@@ -236,7 +237,7 @@ let internals, initial_stat_env =
 
 let parse_stab_sig s name  =
   let open Diag.Syntax in
-  let mode = {Lexer.privileged = false} in
+  let mode = Lexer.{mode with privileged = false} in
```
crusso, line 240:
```suggestion
  let mode = Lexer.{privileged = false; verification = false} in
```

src/pipeline/pipeline.ml
```diff
@@ -246,7 +247,7 @@ let parse_stab_sig_from_file filename : Syntax.stab_sig Diag.result =
   let ic = Stdlib.open_in filename in
   Diag.finally (fun () -> close_in ic) (
     let open Diag.Syntax in
-    let mode = {Lexer.privileged = false} in
+    let mode = Lexer.{mode with privileged = false} in
```
crusso, line 250:
```suggestion
    let mode = Lexer.{privileged = false; verification = false} in
```

src/mo_frontend/lexer_lib.ml
```diff
@@ -5,13 +5,16 @@
 *)
 type mode = {
   privileged : bool;
+  verification : bool;
 }
 
 let mode : mode = {
   privileged = Option.is_some (Sys.getenv_opt "MOC_UNLOCK_PRIM");
+  verification = Option.is_some (Sys.getenv_opt "MOC_UNLOCK_VERIFICATION");
```
crusso, line 13: Do we need the environment variable at all? Why not just use flags.viper (or add one if not present), implicitly set by --viper?

test/viper/claim-broken.mo
```diff
@@ -1,4 +1,5 @@
 // @verify
+//MOC-ENV MOC_UNLOCK_VERIFICATION=viper
```
crusso, line 2: ditto

test/viper/claim.mo
```diff
@@ -1,4 +1,5 @@
 // @verify
+//MOC-ENV MOC_UNLOCK_VERIFICATION=viper
```
crusso, line 2: ditto

test/viper/invariant.mo
```diff
@@ -1,4 +1,5 @@
 // @verify
+//MOC-ENV MOC_UNLOCK_VERIFICATION=viper
```
crusso, line 2: ditto

src/mo_frontend/lexer_lib.ml
ggreif: The problem with that is that it won't compile the `make -C test/viper` mode.

test/viper/assertions.mo
```diff
@@ -1,4 +1,5 @@
 // @verify
+//MOC-ENV MOC_UNLOCK_VERIFICATION=viper
```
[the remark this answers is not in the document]
ggreif: See above. Maybe the makefile can. I have to try.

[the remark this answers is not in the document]
ggreif: `--viper` already sets the `verification` flag.

src/mo_frontend/lexer_lib.ml
ggreif: See 571b576ca

crusso:
> The latter way is needed for running also the backend tests (and the env-var is set by by the Makefile centrally).
Is that still true?

It would be good if we could make the assertions fail too. I think you can do that by raising a syntax error in the action of production, conditional on the viper flag.

test/fail/ok/verification-asserts.tc.ok
```diff
@@ -0,0 +1,10 @@
+verification-asserts.mo:1.1-1.19: syntax error [M0181], verification assertions not permitted in normal mode
+verification-asserts.mo:2.1-2.20: syntax error [M0181], verification assertions not permitted in normal mode
+verification-asserts.mo:3.1-3.17: syntax error [M0181], verification assertions not permitted in normal mode
+verification-asserts.mo:4.1-4.19: syntax error [M0181], verification assertions not permitted in normal mode
+verification-asserts.mo:7.8-7.17: syntax error [M0001], unexpected token 'invariant', expected one of token or <phrase> sequence:
+  system <exp_nest>
+  return <exp_nest>
+  invariant <exp_nest>
```
ggreif, line 8: this is a bit yucky, but harder to resolve nicely...

crusso (approved):
LGTM

PR #3586, opened 2022-11-16
bugfix:  consider records with mutable fields non-static (fixes #3585)

### LINKED ISSUE #3585: bug: inconsistent definition of static expressions in imported modules
@timohanke reasonably asks:
Why is

Field.mo
```
module {
  var n = 0; // rejected, non-static
}
```
not allowed but

Record.mo
```
module {
  let x = { var n = 0 }; // allowed, static?
}
```
is allowed when imported as a library:

```
import F = "FIeld";
import R = "Record";
import O = "Object";
actor {
}

We should either rule out or allow both, I'd say, but which? For purity, I'd say rule them both out, but our own libraries like ExperimentalStableMemory are actually stateful (at least for now).

Curiously:

Object.mo
```
module {
  let x = object { var n = 0 }; // rejected, non-static
}
```
is not allowed!

https://m7sm4-2iaaa-aaaab-qabra-cai.raw.ic0.app/?tag=1172251413

### PR DESCRIPTION
**Pull request by crusso**
Fixes #3585 by treating records literals with var fields as non-static (to be consistent with modules and objects).

### CODE REVIEW

File: test/fail/issue-3585/Object.mo [hunk 1]
```diff
@@ -0,0 +1,3 @@
+module  {
+  public let x = object { var v = 1 }; // rejected as non-static
+}
```
crusso, line 3:
```suggestion
}

```

test/fail/issue-3585/Record.mo
```diff
@@ -0,0 +1,3 @@
+module  {
+  public let x =  { var v = 1 }; // was allowed, though impure, now reject as non-static!
+}
```
crusso, line 3:
```suggestion
}

```

test/fail/issue-3585a.mo
```diff
@@ -0,0 +1,4 @@
+import R = "issue-3585/Record";
+
+actor {
+}
```
crusso, line 4:
```suggestion
}

```

test/fail/issue-3585b.mo
```diff
@@ -0,0 +1,4 @@
+import O = "issue-3585/Object";
+
+actor {
+}
```
crusso, line 4:
```suggestion
}

```

test/fail/issue-3585c.mo
```diff
@@ -0,0 +1,4 @@
+import F = "issue-3585/Field";
+
+actor {
+}
```
crusso, line 4:
```suggestion
}

```

nomeata (approved):

PR #3588, opened 2022-11-16
feat: `let` ... `else`

ggreif:
This was 100% written by @kentosugama.

---------
New `let` binding that allows a failure block in case of a pattern-match failure. The main motivation for this feature is to avoid deeply nested switch statements that lead to less-readable code.

See forum discussion [here](https://forum.dfinity.org/t/match-and-take-in-motoko-do-for-variants-was-when/13427/6?u=kentosugama).

Simple example:
``` Motoko
let ?x = foo() else { Debug.print "foo failed"; return };
```

The failure block must always have type `None` to ensure that execution does not continue beyond the failed binding. Failure blocks may thus trap, return out of the function, or break / continue if in the context of a loop.

Failure blocks may also perform asynchronous effects, if the binding itself is in an asynchronous context.

``` Motoko
let (#ok x) = foo() else { await logError(); return; };
```

More complex example:
Without this binding, you might have to either do the following
``` Motoko
switch (foo()) {
  case (?x) {
    switch (bar()) {
      case (?y) {
        switch (baz()) {
          case (?z) {
            return #ok x + y + z;
          }
          case null {
            return #err "baz failed";
          }
        }
      };
      case null {
        return #err "bar failed";
      }
    }
  };
  case null {
    return #err "foo failed with";
  }
}
```
or
``` Motoko
switch (foo(), bar(), baz()) {
  case (?x, ?y, ?z) {
    return #ok x + y + z;
  }
  case (null, _, _) {
    return #err "foo failed";
  }
  case (_, null, _) {
    return #err "bar failed";
  }
  case (_, _, null) {
    return #err "baz failed";
  }
}
```
however this second version does not execute examples sequentially (i.e. even if `foo()` fails, `bar()` and `baz()` will execute).

With `let-else` bindings:
``` Motoko
let ?x = foo() else { return #err "foo failed" };
let ?y = bar() else { return #err "bar failed" };
let ?z = baz() else { return #err "baz failed" };
   
return #ok x + y + x
```

Inspired by the same feature in [rust](https://rust-lang.github.io/rfcs/3137-let-else.html)

------------------------------------------------------

Items left to do:
- [x] Change parser to parse the body of `else` as a block instead of a value
- [x] Have `return` at the end of `else` block return out of the function
- [x] Finish the compilation with desugaring
- [x] Come up with lots of positive (`run`) and negative (`fail`) tests
  - [x] `return`, `continue`, `break` in failure part (also with 0, 1 and more return values)
  - [x] `return` in monadic context (`async`)
  - [x] `await` in failure part
- [x] Assign error code and explanation (`M0500` is provisional) — _not needed_
- [x] Suppress coverage warning when fail-expression is present
- [x] Let-bound terminal expressions (in `compUnit.ml`)?
- [x] Double check all the `None` pattern matches
- [x] Add documentation
- [x] remove trailing whitespace from PR
- [x] add trailing newlines to files that miss it
- [x] Fix repl tests
- [x] Fix async effects
- [x] Fix null pattern matches
- [x] Find more complex example
- [x] Have the type checker reject let-else bindings to actors

Other PR:
- `type continuation = V.(value cont);`
- May want to warn dead code when fail present for irrefutable match

Potential further optimisations:
- `let (b_1, b_2, ..., b_n) = switch { case <pat> (b_1, b_2, ..., b_n); ... };`


src/mo_frontend/parser.mly
```diff
@@ -830,7 +830,10 @@ dec_var :
     { VarD(x, annot_exp e t) @? at $sloc }
 
 dec_nonvar :
-  | LET p=pat EQ e=exp(ob)
+  | LET p=pat EQ e=exp(ob) %prec LET_NO_ELSE
+    { let p', e' = normalize_let p e in
+      LetD (p', e') @? at $sloc }
+  | LET p=pat EQ e=exp(ob) ELSE fail=exp(bl)
```
ggreif, line 837: `IF` uses `ELSE e2=exp_nest`, maybe we should follow that example?

kentosugama: So I think `exp_nest` can either be a block or an expression, which sounds like what we want

kentosugama:
So the type checker accepts `return` at the end of the else block without any extra logic...

src/mo_interpreter/interpret.ml
```diff
@@ -705,43 +705,43 @@ and declare_pat_fields pfs ve : val_env =
 and define_id env id v =
   Lib.Promise.fulfill (find id.it env.vals) v
 
-and define_pat env pat v =
-  let err () = trap pat.at "value %s does not match pattern" (string_of_val env v) in
-  define_pat' env err pat v
-
-and define_pat' env err pat v =
+and define_pat env (k : V.value V.cont) err pat v =
```
ggreif, line 708: maybe we could have a `type continuation = V.(value cont);` in a cleanup PR sometime.

src/mo_interpreter/interpret.ml
```diff
@@ -705,43 +705,43 @@ and declare_pat_fields pfs ve : val_env =
...
+  | TupP pats ->
+    define_pats env k err pats (V.as_tup v)
+  | ObjP pfs -> 
+    define_pat_fields env k err pfs (V.as_obj v)
   | OptP pat1 ->
     (match v with
-    | V.Opt v1 -> define_pat' env err pat1 v1
+    | V.Opt v1 -> define_pat env k err pat1 v1
     | V.Null -> err ()
     | _ -> assert false
```
ggreif, line 726: I wonder what this covers...

src/mo_interpreter/interpret.ml
```diff
@@ -885,16 +885,14 @@ and interpret_dec env dec (k : V.value V.cont) =
...
-      define_pat' env err pat v;
-      k v
+  | LetD (pat, exp, fail) -> 
+    interpret_exp env exp (fun v -> 
+      let err () = 
+        match fail with
+        | Some fail -> interpret_exp env fail (fun _ -> assert false)
+        | None -> trap pat.at "value %s does not match pattern" (string_of_val env v)
+      in
+      define_pat env k err pat v
```
ggreif, line 895: This looks like a particularly clean solution to the `return` problem. Great you figured this out!

src/mo_interpreter/interpret.ml
```diff
@@ -881,10 +885,14 @@ and interpret_dec env dec (k : V.value V.cont) =
...
   | ExpD exp ->
     interpret_exp env exp k
-  | LetD (pat, exp) ->
-    interpret_exp env exp (fun v ->
-      define_pat env pat v;
-      k v
+  | LetD (pat, exp, fail) -> 
+    interpret_exp env exp (fun v -> 
+      let err () = 
+        match fail with
```
ggreif, line 891:
```suggestion
      let err () = match fail with
```

src/mo_frontend/typing.ml
```diff
@@ -2300,8 +2302,14 @@ and infer_block_exps env decs : T.typ =
...
   let t =
   match dec.it with
-  | ExpD exp
-  | LetD (_, exp) ->
+  | ExpD exp 
+  | LetD (_, exp, None) -> infer_exp env exp
+  | LetD (_, exp, Some fail) -> 
+    let fail_type = infer_exp env fail in
+    if not T.(sub fail_type Non) then 
+      local_error env fail.at "M0500"
```
ggreif, line 2310: I never know if this should be `error` or `local_error`...

ggreif: Also, pick a fresh code from `src/lang_utils/error_codes.ml`.

src/mo_frontend/typing.ml
```diff
@@ -2300,8 +2302,14 @@ and infer_block_exps env decs : T.typ =
 and infer_dec env dec : T.typ =
   let t =
   match dec.it with
-  | ExpD exp
-  | LetD (_, exp) ->
+  | ExpD exp 
+  | LetD (_, exp, None) -> infer_exp env exp
+  | LetD (_, exp, Some fail) -> 
+    let fail_type = infer_exp env fail in
+    if not T.(sub fail_type Non) then
```
ggreif, line 2309: maybe there is a nicer way to put this? @crusso

ggreif: `infer_exp` or `infer_exp_strong`?

src/mo_frontend/typing.ml
```diff
@@ -2380,6 +2388,8 @@ and check_dec env t dec =
   | ExpD exp ->
     check_exp env t exp;
     dec.note <- exp.note
+  | LetD (pat, exp, Some fail) -> 
+    check_exp env T.Non fail (* TODO FIX THIS not being triggered *)
```
ggreif, line 2392: One can trigger it by `let x : Int = (let ... else);`

src/mo_frontend/typing.ml
```diff
@@ -2426,7 +2436,8 @@ and gather_dec env scope dec : Scope.t =
   | LetD (
       {it = VarP id; _},
       ({it = ObjBlockE (obj_sort, dec_fields); at; _} |
-       {it = AwaitE { it = AsyncE (_, {it = ObjBlockE ({ it = Type.Actor; _} as obj_sort, dec_fields); at; _}) ; _  }; _ })
+       {it = AwaitE { it = AsyncE (_, {it = ObjBlockE ({ it = Type.Actor; _} as obj_sort, dec_fields); at; _}) ; _  }; _ }),
+       None
```
ggreif, line 2440: Need to consider `Some _` here too.

src/mo_interpreter/interpret.ml
kentosugama: Fixed!

src/mo_interpreter/interpret.ml
kentosugama: Yay

src/mo_frontend/typing.ml
kentosugama: Is there a scheme to these numbers? Picking 159 if not

src/mo_frontend/typing.ml
```diff
@@ -2300,9 +2302,19 @@ and infer_block_exps env decs : T.typ =
...
-    infer_exp env exp
+  | ExpD exp 
+  | LetD (_, exp, None) -> infer_exp env exp
+  | LetD (_, exp, Some fail) -> 
+    let fail_type = infer_exp env fail in
+    if not T.(sub fail_type Non) then 
+      local_error env fail.at "M0159"
+        "pattern binding failure handler of type%a\ndoes not match expected type None"
+        display_typ_expand fail_type;
+    (* check_exp env T.Non fail; *)
```
ggreif, line 2313: remove

src/mo_frontend/typing.ml
ggreif: no, see the end of `src/lang_utils/error_codes.ml`

src/lowering/desugar.ml
```diff
@@ -666,15 +666,28 @@ and block force_unit ds =
...
+    let x = fresh_var "x" (e.note.S.note_typ) in
+    (decs prefix @ [letD x (exp e); letP (pat p) (varE x)], varE x)
+  | false, S.LetD (p, e, Some f) ->
+    let x = fresh_var "x" (e.note.S.note_typ) in
+    let p' = pat p in
+    let e' = exp e in
+    let f' = exp f in
+    let switchE = 
+      I.{ 
+        it = I.SwitchE(
```
ggreif: Why do you prefer `I.SwitchE`?

src/lowering/desugar.ml
```diff
@@ -666,15 +666,28 @@ and block force_unit ds =
...
+          exp e,
+          [
+            { it = { pat = pat p; exp = exp e }; at = (exp e).at; note = () };
+            { it = { pat = { it = WildP; at = (exp f).at; note = (pat p).note }; exp = exp f }; at = (exp f).at ; note = () }
+          ]);
+        at = (exp e).at;
+        note = (exp e).note
```
ggreif: why not use `e'`?

src/lowering/desugar.ml
```diff
@@ -657,6 +657,21 @@ and text_dotE proj e =
     | "chars" -> call "@text_chars" [] [T.iter_obj T.char]
     |  _ -> assert false
 
+and let_else_switch p e f = 
+  let e' = exp e in
+  let p' = pat p in
+  let f' = exp f in
```
ggreif, line 663:
```suggestion
  let e', p', f' = exp e, pat p, exp f in
```

src/lowering/desugar.ml
```diff
@@ -657,6 +657,21 @@ and text_dotE proj e =
...
+  I.{ 
+    it = I.SwitchE(
```
ggreif, line 665: `{ e' with it = ... }`

src/lowering/desugar.ml
```diff
@@ -657,6 +657,21 @@ and text_dotE proj e =
...
+      exp e,
+      [
+        { it = { pat = p'; exp = e' }; at = e'.at; note = () };
+        { it = { pat = { it = WildP; at = f'.at; note = p'.note }; exp = f' }; at = f'.at ; note = () }
+      ]);
+    at = e'.at;
+    note = e'.note
```
ggreif, line 672: not needed

src/lowering/desugar.ml
ggreif: `I.SwitchE` -> `SwitchE`

src/lowering/desugar.ml
```diff
@@ -657,6 +657,21 @@ and text_dotE proj e =
...
+and let_else_switch p e f = 
+  let e' = exp e in
+  let p' = pat p in
+  let f' = exp f in
+  I.{ 
+    it = I.SwitchE(
+      exp e,
+      [
+        { it = { pat = p'; exp = e' }; at = e'.at; note = () };
+        { it = { pat = { it = WildP; at = f'.at; note = p'.note }; exp = f' }; at = f'.at ; note = () }
```
ggreif, line 669: normally you could use `cases` here if the input is on the `S` (`Source`) side...

src/mo_frontend/typing.ml
kentosugama: Ah okay. Will fix

src/mo_frontend/typing.ml
kentosugama: Seems like check_exp again None type is working

src/lowering/desugar.ml
kentosugama: I was maximally expanding everything to avoid the dag error at the time

src/lowering/desugar.ml
kentosugama: I was keeping in line with the other cases in `block` and `dec'` lowering directly to IR. Seemed less hacky / convoluted to go straight to IR instead of rewriting to another source expression (especially since there wasn't a lot of complexity to do it).

src/lowering/desugar.ml
ggreif: IIRC inly the `it` part is DAG-checked. That's why I am advocating for Building a `S.SwitchE` because that automatically creates a non-DAG intermediate representation.

src/lowering/desugar.ml
```diff
@@ -658,16 +658,10 @@ and text_dotE proj e =
...
-    it = S.SwitchE(
-      e,
-      S.[
-        { it = { pat = p; exp = e }; at = e.at; note = () };
-        { it = { pat = { it = WildP; at = f.at; note = p.note }; exp = f }; at = f.at ; note = () }
-      ]
-    )
-  }
+  exp { e with it = S.SwitchE(e,
+        S.[ { it = { pat = p; exp = e }; at = e.at; note = () };
```
ggreif, line 662: This has a small problem. If there are side effects in `e`, then those will happen potentially twice, since `e` is not used linearly. So theoretically we have to bind e to a fresh binding before doing the `SwitchE` and use that twice.

ggreif: Maybe do a `debugPrint`-test first to see the problem and then we'll have to fix this sometime.

ggreif: Btw. `and` patterns would solve this neatly:
``` Motoko
let <pat> = switch <exp> {
    case (<pat> and binding) binding;
    case _ fail
}
```
Alas, we don't have them (yet).

kentosugama: Ah! Yes, side effects. Didn't think of that. Will take a look

kentosugama: So currently:
``` 
     func () { 
       var y = 0; 
       let x = (y := 2) else { return }; 
     };
```
compiles to
```(VarD y Nat (LitE (NatLit 0)))
        (LetD
          (VarP x)
          (SwitchE
            (AssignE (VarLE y) (LitE (NatLit 2)))
            (case (VarP x) (AssignE (VarLE y) (LitE (NatLit 2))))
            (case WildP (PrimE RetPrim (PrimE TupPrim)))
          )
        )```

which is going to evaluate the assignment to `y` twice

Comment (kentosugama) [reply src/lowering/desugar.ml:662]: Uhh bad idempotent example, but you get the point haha

Comment (kentosugama) [reply src/lowering/desugar.ml]: Okay this should be fixed now

File: doc/md/language-manual.md [hunk 18]
```diff
@@ -1367,6 +1367,15 @@ The declaration `let <pat> = <exp>` evaluates `<exp>` to a result `r`. If `r` is
 
 All bindings declared by a `let` (if any) are *immutable*.
 
+#### Handling pattern match failures
+
+In the presence of of refutable patterns a `let` declaration may fail to bind the expression. The default consequence of such a failure is trapping. The compiler will additionally emit a warning if a pattern match failure is a possibility.
+
+There are cases, however, when the user wants to explicitly handle such pattern match failures. For such cases
+the let declaration `let <pat> = <exp> else <fail-expr>` is provided, and has identical static and dynamic semantics with
```
crusso, line 1375:
```suggestion
the let declaration `let <pat> = <exp> else <fail-expr>` is provided, and has identical static and dynamic semantics with ....
```
with what?

crusso (commented):
Driveby: what happens to legacy 
```
let x = if e0 e1 else e2;
```
Does that still have the same meaning? Even if it does, I think we might need a more distinctive syntax to avoid any confusion.

kentosugama:
At the moment that does have the same meaning (the `else` associates with the `if`).

Apparently the rust people ran into this same problem: [link](https://rust-lang.github.io/rfcs/3137-let-else.html#let-pattern--if--else--else-)

They're approach was to enforce parenthesis around the `if-else` for disambiguation. I believe they also always require brackets around the failing `else` block, which we don't atm.

test/run/let-else-await.mo
```diff
@@ -0,0 +1,8 @@
+func bar() : async Nat { 4 };
+
+func foo() : async Nat {
+    let 2 = 3 else { return await bar() };
+    5
+};
+
+assert ((await foo()) == 4);
```
ggreif, line 8:
```suggestion
assert (await foo()) == 4;
```

src/lowering/desugar.ml
ggreif: btw., easy non-idempotents are `v += 1` etc.

doc/md/language-manual.md
ggreif: I could swear I did complete this sentence. Let me fix.

ggreif: Maybe you reviewed before I finished it.

test/run-drun/let-else-await.mo
```diff
@@ -1,6 +1,6 @@
 actor {
     public func go() : async Nat {
-        let 2 = 3 else { return 4 };
+        let 2 = 3 else { return await async 4 };
```
ggreif, line 3: now that we have `async*` too maybe add a `go_star` method as well...

crusso:
https://forum.dfinity.org/t/match-and-take-in-motoko-do-for-variants-was-when/13427?u=claudio was (one) of the forum threads asking for something like this.

test/run/let-else-if.mo
```diff
@@ -0,0 +1,8 @@
+let x = if true 4 else 5 else { assert false; loop() };
+
+func foo() {
+    let 2 = if true 3 else 3 else { return };
+    assert false;
+};
+
+foo();
```
crusso, line 8:
```suggestion
foo();

```

test/run/let-else-return.mo
```diff
@@ -0,0 +1,19 @@
...
+};
+
+assert (2 == single());
+
+func multiple() : (Nat, Nat) {
+    let 2 = 3 else { return (2, 3) };
+    (0, 0)
+};
+
+assert ((2, 3) == multiple());
```
crusso, line 19:
```suggestion
assert ((2, 3) == multiple());

```

test/run/let-else.mo
```diff
@@ -0,0 +1,3 @@
+let x = 3 else { assert false; loop() };
+
+let (y, z) = (4, 5) else { assert false; loop() };
```
crusso, line 3:
```suggestion
let (y, z) = (4, 5) else { assert false; loop() };

```

test/fail/let-else-not-none.mo
```diff
@@ -0,0 +1 @@
+let x = 4 else { 5 };
```
crusso, line 1:
```suggestion
let x = 4 else { 5 };

```

crusso:
Should this (in the description) 
```
switch (foo(), bar(), baz()) {
  let ?x = foo() else { return #err "foo failed };
  let ?y = bar() else { return #err "bar failed };
  let ?z = baz() else { return #err "baz failed };
   
  return #ok x + y + x
}
```
be:

```
  let ?x = foo() else { return #err "foo failed };
  let ?y = bar() else { return #err "bar failed };
  let ?z = baz() else { return #err "baz failed };
  return #ok x + y + x
}
```
I like it!

kentosugama:
Yup, updated. Thanks.

crusso:
What happens with
```
....
let ?x = ... else { return x; }
...
```
Hopefully, the definedness analysis will reject it.

src/lowering/desugar.ml
```diff
@@ -684,11 +704,14 @@ and block force_unit ds =
...
     (decs ds, varE (var x.it e.note.S.note_typ))
-  | false, S.LetD (p', e') ->
-    let x = fresh_var "x" (e'.note.S.note_typ) in
-    (decs prefix @ [letD x (exp e'); letP (pat p') (varE x)], varE x)
+  | false, S.LetD (p, e, None) ->
+    let x = fresh_var "x" (e.note.S.note_typ) in
+    (decs prefix @ [letD x (exp e); letP (pat p) (varE x)], varE x)
+  | false, S.LetD (p, e, Some f) ->
+    let x = fresh_var "x" (e.note.S.note_typ) in
+    (decs prefix @ [letD x (let_else_switch p e f last.note); letP (pat p) (varE x)], varE x)
```
kentosugama, line 714: just return `let_else_switch`. Make sure a test is hitting this line

kentosugama:
Added a definedness test to check for the case

ggreif (commented):
In a small group walkthrough we just has a look at the code changes and this looks very reasonable to me.

src/mo_frontend/parser.mly
```diff
@@ -162,7 +162,8 @@ let share_exp e =
 
 let share_dec d =
   match d.it with
-  | LetD (p, e) -> LetD (p, share_exp e) @? d.at
+  | LetD (p, e, None) -> LetD (p, share_exp e, None) @? d.at
```
kentosugama, line 165: Fix this:
`actor { public let x = shared func f() {} else { loop() }; };`

ggreif: But here the `else ...` part is dead, right?

src/mo_frontend/typing.ml
```diff
@@ -2462,7 +2466,8 @@ and gather_dec env scope dec : Scope.t =
   | LetD (
       {it = VarP id; _},
       ({it = ObjBlockE (obj_sort, dec_fields); at; _} |
-       {it = AwaitE (_,{ it = AsyncE (_, _, {it = ObjBlockE ({ it = Type.Actor; _} as obj_sort, dec_fields); at; _}) ; _  }); _ })
+       {it = AwaitE (_,{ it = AsyncE (_, _, {it = ObjBlockE ({ it = Type.Actor; _} as obj_sort, dec_fields); at; _}) ; _  }); _ }),
+       None (* FIXME ?? *)
```
kentosugama, line 2470: Fix this

src/mo_frontend/typing.ml
```diff
@@ -2637,9 +2644,13 @@ and infer_dec_valdecs env dec : Scope.t =
...
     let _ve = check_pat env obj_typ pat in
     Scope.{empty with val_env = T.Env.singleton id.it obj_typ}
-  | LetD (pat, exp) ->
+  | LetD (pat, exp, fail) ->
     let t = infer_exp {env with pre = true} exp in
-    let ve' = check_pat_exhaustive (if is_import dec then local_error else warn) env t pat in
+    let consequence = match fail with
+      | None when is_import dec -> local_error
+      | None -> warn
+      | Some _ -> fun _ _ _ -> Format.kasprintf ignore in
```
kentosugama, line 2652: Check this

src/mo_frontend/typing.ml
```diff
@@ -2699,7 +2710,7 @@ let infer_prog scope prog : (T.typ * Scope.t) Diag.result =
 let is_actor_dec d =
   match d.it with
   | ExpD e
-  | LetD (_, e) -> CompUnit.is_actor_def e
+  | LetD (_, e, None) -> CompUnit.is_actor_def e
```
kentosugama, line 2713: Check this. Similar to actor method case.

ggreif:
In the description
``` Motoko
let ?x = foo() else { return #err "foo failed" };
let ?y = bar() else { return #err "bar failed" };
let ?z = baz() else { return #err "baz failed" };
```
probably could be simplified to
``` Motoko
let ?x = foo() else return #err "foo failed";
let ?y = bar() else return #err "bar failed";
let ?z = baz() else return #err "baz failed";
```

PR #3598, opened 2022-11-21
experiment: using `await t`, `do await {}` and `await in` syntax for await abstraction

crusso:


ggreif:
My suggestion is `await? <some-async>`. Here the `?` would mean _optionally_ or _if-needed_.

crusso:
variant implemented in #3609

PR #3599, opened 2022-11-21
fix: thread the mode into the parser

ggreif:
Turns out I was doing a bunch of things wrong. Now we make the relevant `Lexer_lib.mode` accessible from the parser productions (previously it just accessed a constant thing).

This makes `--viper` to select the `verification` mode in the parser too (lexer was already fine).

### DIFF
--- a/src/languageServer/imports.ml
+++ b/src/languageServer/imports.ml
@@ -9,7 +9,10 @@ type import = string * string
 
 let parse_with mode lexbuf parser =
   let tokenizer, _ = Lexer.tokenizer mode lexbuf in
-  Ok (Parsing.parse Lexer_lib.mode 0 (parser lexbuf.Lexing.lex_curr_p) tokenizer lexbuf)
+  Ok
+    (Parsing.parse Lexer_lib.mode 0
+       (parser lexbuf.Lexing.lex_curr_p)
+       tokenizer lexbuf)
 
 let parse_string s =
   try
--- a/src/languageServer/imports.ml
+++ b/src/languageServer/imports.ml
@@ -9,7 +9,7 @@ type import = string * string
 
 let parse_with mode lexbuf parser =
   let tokenizer, _ = Lexer.tokenizer mode lexbuf in
-  Ok (Parsing.parse 0 (parser lexbuf.Lexing.lex_curr_p) tokenizer lexbuf)
+  Ok (Parsing.parse Lexer_lib.mode 0 (parser lexbuf.Lexing.lex_curr_p) tokenizer lexbuf)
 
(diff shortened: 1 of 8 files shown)


src/mo_frontend/parser.mly
```diff
@@ -205,7 +205,10 @@ let (&&&) cond action =
   if not cond then syntax_error action.at "M0181" "verification assertions not permitted in normal mode";
   action
 
-let is_verification = Lexer_lib.(mode.verification)
+let is_verification () =
```
ggreif, line 208: can we migrate these two to `assertions.mly`?

src/mo_frontend/assertions.mly
```diff
@@ -1,4 +1,26 @@
-(* Viper only tokens and productions *)
+%{
```
ggreif, line 1: moved here from `parser.mly`

aterga:
I've tested that PR enables the `--viper` flag, but without the flag I get, e.g.

```bash
moc test/viper/invariant.mo
test/viper/invariant.mo:9.10-9.19: syntax error [M0001], unexpected token 'invariant', expected one of token or <phrase> sequence:
  system <exp_nest>
  return <exp_nest>
  invariant <exp_nest>
  func <exp_nest>
  <nat> : async <exp_nest>
```

whereas beforehand we had 
```bash
[nix-shell:~/workdir/motoko]$ moc test/viper/invariant.mo
test/viper/invariant.mo:9.3-9.71: syntax error [M0181], verification assertions not permitted in normal mode
test/viper/invariant.mo:10.3-10.29: syntax error [M0181], verification assertions not permitted in normal mode
test/viper/invariant.mo:11.3-11.50: syntax error [M0181], verification assertions not permitted in normal mode
test/viper/invariant.mo:14.7-14.29: syntax error [M0181], verification assertions not permitted in normal mode
test/viper/invariant.mo:16.7-16.31: syntax error [M0181], verification assertions not permitted in normal mode
```

which was more verbose. This is just FYI; I don't consider the new behavior problematic.

src/languageServer/imports.ml
```diff
@@ -9,7 +9,10 @@ type import = string * string
 
 let parse_with mode lexbuf parser =
   let tokenizer, _ = Lexer.tokenizer mode lexbuf in
-  Ok (Parsing.parse 0 (parser lexbuf.Lexing.lex_curr_p) tokenizer lexbuf)
+  Ok
+    (Parsing.parse Lexer_lib.mode 0
```
ggreif, line 13: passing an additional argument

crusso (approved):
LGTM

PR #3606, opened 2022-11-24
refactor: omit all overhead of write-barriers in non-generational gc

crusso:


luc-blaeser (commented):
Really nice solution. Thank you very much.

PR #3609, opened 2022-11-25
feat: implement `async*`/`await*` for efficient async abstraction

### LINKED ISSUE #1482: Support direct abstraction of code that awaits into functions, without requiring an unnecessary async

crusso:
Add opt-in support for efficient abstraction of asynchronous code using `async*`/`await*` to delineate possible commits/context switches. Fixes #1482 and addresses https://forum.dfinity.org/t/canister-output-message-queue-limits-and-ic-management-canister-throttling-limits/15972/22.

Syntax:

```
<type ::= ...
  async* <typ>                                  delayed, asynchronous computation

<exp> ::= ...
  async* <block-or-exp>                          delay an asynchronous computation
  await* <block-or-exp>                          await a delayed computation (only in async)
```

This is another take on the design, avoid some of the pitfalls (but also advantages) of #3573 which proposed an eager `do async <block_or_exp>` but no new types.

## Async* types

`async* <typ>` specifies a delayed, asynchronous computation producing a value of type `<typ>`.

Computation types typically appear as the result type of a `local` function that produces an `await*`-able value.

(They cannot be used as the return types of `shared` functions.)

## Async*

The async expression `async* <block-or-exp>` has type `async* T` provided:

-   `<block-or-exp>` has type `T`;

-   `T` is shared.

Any control-flow label in scope for `async* <block-or-exp>` is not in scope for `<block-or-exp>`. However, `<block-or-exp>` may declare and use its own, local, labels.

The implicit return type in `<block-or-exp>` is `T`. That is, the return expression, `<exp0>`, (implicit or explicit) to any enclosed `return <exp0>?` expression, must have type `T`.

Evaluation of `async* <block-or-exp>` produces a delayed computation to evaluate `<block-or-exp>`. It immediately returns a value of type `async* T`.
The delayed computation can be executed using `await*`, producing one evaluation
of the computation `<block-or-exp>`.

### Danger

Note that `async <block-or-exp>` has the effect of scheduling a single asynchronous computation of `<exp>`, regardless of whether its result, a future, is consumed with an `await`.
Moreover, each additional consumption by an `await` just returns the previous result, without repeating the computation.

In comparison, `async* <block-or_exp>`,  has *no effect* until its value is consumed by an `await*`.
Moreover, each additional consumption by an `await*` will trigger a new evaluation of `<block-or-exp>`.

Be careful of this distinction, and other differences, when refactoring code.

### Note:

The `async*` and corresponding `await*` constructs are useful for efficiently abstracting asynchronous code into re-useable functions.
In comparison, calling a local function that returns a proper `async` type requires committing state and suspending execution with each `await` of its result, which can be undesirable.

## Await*

The `await*` expression `await* <exp>` has type `T` provided:

-   `<exp>` has type `async* T`,

-   `T` is shared,

-   the `await*` is explicitly enclosed by an `async`-expression or appears in the body of a `shared` function.
Expression `await <exp>` evaluates `<exp>` to a result `r`. If `r` is `trap`, evaluation returns `trap`. Otherwise `r` is a delayed computation `<block-or-exp>`. The evaluation of `await* <exp>` proceeds
with the evaluation of `<block-or-exp>`, executing the delayed computation.

### Danger

During the evaluation of `<block-or-exp>`, the state of the enclosing actor may change due to concurrent processing of other incoming actor messages. It is the programmer’s responsibility to guard against non-synchronized state changes.

### Note

Unlike `await`, which, regardless of the dynamic status of the future, ensures that all tentative state changes and message sends prior to the `await` are committed and irrevocable, `await*` does not, in itself, commit any state changes, nor does it suspend computation.
Instead, evaluation proceeds immediately according to `<block-or-exp>` (the value of `<exp>`), committing state and suspending execution whenever `<block-or-exp>` does (but not otherwise).

### Note

Evaluation of a delayed `async*` block is synchronous while possible, switching to asynchronous when necessary due to a proper `await`.

Using `await*` signals that the computation *may* commit state and suspend execution during the evaluation of `<block-or-exp>`, that is, that evaluation of `<block-or-exp>` may perform zero or more proper `await`s and may be interleaved with the execution of other, concurrent messages.

## TODO:

Future:
- [ ] generics and flattening - rule out generics for now?
- [ ] relax shared content type for `async*` types?

One annoying thing is that we cannot make actor class instantiation yet more efficient without returning  an `async*`, breaking code. Although I guess users could opt in to that if wanted (by giving an `async*` return type). 
That's a drawback compared to the previous `do async/await` approach. But them's the breaks.


src/ir_passes/await.ml
```diff
@@ -99,26 +102,51 @@ and t_exp' context exp' =
...
+        let context' = LabelEnv.add Return Label LabelEnv.empty in
+        t_exp context' exp1
+      | T.Await ->
+        (* define k_ret as identity continuation and enter cps *)
+        let k_ret = fresh_cont (typ exp1) (typ exp1) in
+        let v = fresh_var "v" (typ exp1) in
+        let context' = LabelEnv.add Return (Cont (ContVar k_ret)) LabelEnv.empty in
+        blockE [funcD k_ret v (varE v)]
+          (c_exp context' exp1 (ContVar k_ret)))
   | ActorE (ds, ids, { meta; preupgrade; postupgrade; heartbeat; inspect}, t) ->
```
crusso, line 150: Why has this changed? Inherited from previous PR.

src/mo_def/syntax.ml
```diff
@@ -47,7 +49,7 @@ and typ' =
   | VariantT of typ_tag list                       (* variant *)
   | TupT of typ_item list                          (* tuple *)
   | FuncT of func_sort * typ_bind list * typ * typ (* function *)
-  | AsyncT of scope * typ                          (* future *)
+  | AsyncT of async_sort * scope * typ                    (* future / computation *)
```
crusso, line 52: indentation

src/mo_frontend/typing.ml
```diff
@@ -1568,10 +1575,16 @@ and check_exp' env0 t exp : T.typ =
         display_typ_expand (T.Array t');
     List.iter (check_exp env (T.as_immut t')) exps;
     t
-  | AsyncE (tb, exp1), T.Async (t1', t') ->
+  | AsyncE (s1, tb, exp1), T.Async (s2, t1', t') ->
     error_in [Flags.WASIMode; Flags.WasmMode] env exp1.at "M0086"
       "async expressions are not supported";
     let t1, next_cap = check_AsyncCap env "async expression" exp.at in
+    if s1 <> s2 then begin
+      local_error env exp.at "M0XXX" (* check number! *)
```
crusso, line 1583: new error code!

doc/md/examples/grammar.txt
```diff
@@ -190,7 +191,9 @@
     <exp_bin> <binassign> <exp>
     'return' <exp>?
     'async' <exp_nest>
+    'async'STAR <block>
     'await' <exp_nest>
+    'await'STAR <exp_nest>
```
crusso, line 196: Fix sed script

ggreif: FTFY: c79306147, please `git pull --rebase`.

src/mo_def/syntax.ml
ggreif: fixed: 19c852fb5c24cdd4347a490fff5bd8d7703b591d

src/mo_frontend/typing.ml
```diff
@@ -1429,15 +1414,21 @@ and infer_exp'' env exp : T.typ =
...
-  | AwaitE exp1 ->
+    T.Async (s, t1, t')
+  | AwaitE (s, exp1) ->
     let t0 = check_AwaitCap env "await" exp.at in
     let t1 = infer_exp_promote env exp1 in
     (try
-       let (t2, t3) = T.as_async_sub t0 t1 in
+       let (t2, t3) = T.as_async_sub s t0 t1 in
+       (* if s1 <> s2 then begin
+           local_error env exp.at "M0XXX" (* check number! *)
```
ggreif, line 1427: This will be `M0183`.

src/mo_frontend/typing.ml
ggreif: `M0184` if distinct from above.

src/mo_frontend/typing.ml
crusso: Thanks!

src/mo_frontend/typing.ml
crusso: Done

doc/md/language-manual.md
```diff
@@ -777,6 +779,14 @@ The optional `<shared>` qualifier specifies whether the function value is shared
 
 Future types typically appear as the result type of a `shared` function that produces an `await`-able value.
 
+### Async* types
+
+`async* <typ>` specifies a delayed asynchronous computation producing a value of type `<typ>`.
+
+Computation types typically appear as the result type of a `local` function that produces an `await*`-able value.
+
+(They cannot be used as the return types of shared functions.)
```
crusso, line 788:
```suggestion
(They cannot be used as the return types of `shared` functions.)
```

doc/md/language-manual.md
```diff
@@ -552,6 +553,7 @@ Type expressions are used to specify the types of arguments, constraints (a.k.a
   ? <typ>                                       option
   <shared>? <typ-params>? <typ> -> <typ>        function
   async <typ>                                   future
+  async* <typ>                                  delayed asynchronous computation
```
crusso, line 556:
```suggestion
  async* <typ>                                  delayed, asynchronous computation
```

doc/md/language-manual.md
```diff
@@ -2231,6 +2197,77 @@ Note: suspending computation on `await`, regardless of the dynamic status of the
...
+
+The `await*` expression `await* <exp>` has type `T` provided:
+
+-   `<exp>` has type `async* T`,
+
+-   `T` is shared,
+
+-   the `await*` is explicitly enclosed by an `async`-expression or appears in the body of a `shared` function.
+Expression `await <exp>` evaluates `<exp>` to a result `r`. If `r` is `trap`, evaluation returns `trap`. Otherwise `r` is a delayed computation `<block-or-exp>`. The evaluation of `await* <exp>` proceeds
+with the evaluation of `<block-or-exp>`, forcing the delayed computation.
```
crusso, line 2250:
```suggestion
with the evaluation of `<block-or-exp>`, executing the delayed computation.
```

doc/md/language-manual.md
```diff
@@ -2231,6 +2197,77 @@ Note: suspending computation on `await`, regardless of the dynamic status of the
...
+-   `<block-or-exp>` has type `T`;
+
+-   `T` is shared.
+
+Any control-flow label in scope for `async* <block-or-exp>` is not in scope for `<block-or-exp>`. However, `<block-or-exp>` may declare and use its own, local, labels.
+
+The implicit return type in `<block-or-exp>` is `T`. That is, the return expression, `<exp0>`, (implicit or explicit) to any enclosed `return <exp0>?` expression, must have type `T`.
+
+Evaluation of `async* <block-or-exp>` produces a delayed computation to evaluate `<block-or-exp>`. It immediately returns a value of type `async* T`.
+The delayed computation can be performed using `await*`, producing one evaluation
```
crusso, line 2217:
```suggestion
The delayed computation can be executed using `await*`, producing one evaluation
```

src/ir_def/check_ir.ml
```diff
@@ -541,17 +541,17 @@ let rec check_exp env (exp:Ir.exp) : unit =
...
+        try T.as_async_sub s T.Non (T.normalize (typ a))
         with _ -> error env exp.at "CPSAwait expect async arg, found %s" (T.string_of_typ (typ a))
       in
       (match cont_typ with
        | T.Func(T.Local, T.Returns, [], ts1, ts2) ->
          begin
            (match ts2 with
             | [] -> ()
-            | [T.Async (_, _)] -> ()
+            (*            | [T.Async (_, _)] -> () *)
```
crusso, line 554:
```suggestion
```

design/asynccps.md
```diff
@@ -32,9 +34,9 @@ The aim of the game is to remove async and await and try catch/throw  by a sourc
 
 Terms have effect `T` (trivial) or `A` (await) with `T` < `A`. A term has effect `A` if any subterm not enclosed by `async` is `await`.
```
crusso, line 35:
```suggestion
Terms have effect `T` (trivial) or `A` (await) with `T` < `A`. A term has effect `A` if any subterm not enclosed by `async` or `async*` is `await` or `await*`.
```

nomeata:
>  Moreover, each additional consumption by an await* will trigger a new evaluation of <block-or-exp>.

That sounds like confusing behavior. Making them normal futures is too hard?

crusso:
> > Moreover, each additional consumption by an await* will trigger a new evaluation of .
> 
> That sounds like confusing behavior. Making them normal futures is too hard?

You mean evaluate on first `await*` but cache the result and re-use on next `await*`? Once could do that, but then you'd still have no effect if you don't `await*`, so there's still a chance on confusion.

Another option is to make the `async*` type second class, a bit like `var T`, so it can only be introduced under an `await*`, forcing unique consumption. 

Or we reconsider https://github.com/dfinity/motoko/pull/3573 (the `do async <block-or-exp>`) option, which avoids the issue introducing only mild peril.

nomeata:
Hmm, I am not sure. It looks a bit like a future, even if it it’s one that’s triggered upon `await*`. Maybe fine.

luc-blaeser:
I am not sure whether I fully understand the danger of concurrent state changes on `await*`: Is it not similar to `await`? Other messages or async message parts may interleave at any `async` or `await`, or between the `async*` and its subsequent `await*`. Or is there a new risk of race conditions (compared to `await`)?

luc-blaeser:
Very nice implementation. I also tested several cases locally.

luc-blaeser (approved):
Very nice solution, also containing a lot of code refactoring. 
A remaining question is whether repeated `await*` of the same `async*` should be allowed, or should be cached, or cause repeated execution...

design/asynccps.md
```diff
@@ -30,11 +32,11 @@ e = x
 
 The aim of the game is to remove async and await and try catch/throw  by a source-to-source translation, leaving as much code as possible in direct-style.
 
-Terms have effect `T` (trivial) or `A` (await) with `T` < `A`. A term has effect `A` if any subterm not enclosed by `async` is `await`.
+Terms have effect `T` (trivial) or `A` (await) with `T` < `A`. A term has effect `A` if any subterm not enclosed by `async` or `async*` is `await` or `await*`.
 
-The only terms that introduce effect `A` is `await`, `try` or `throw` - the effect is masked by its innermost enclosing `async` (if any).
+The only terms that introduce effect `A` is `await`, `try` or `throw`  and `do async` - the effect is masked by its innermost enclosing `async` (if any).
```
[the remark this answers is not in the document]
crusso: Yep. Feel free to just make suggestions - then I can just accept by clicking a button! I'll fix this one manually.

crusso:
> I am not sure whether I fully understand the danger of concurrent state changes on `await*` as mentioned in the description above: Is it not similar to `await`? Other messages or async message parts may interleave at any `async` or `await`, or between the `async*` and its subsequent `await*`. Or is there a new risk of race conditions (compared to `await`)?

It is very similar. I was just repeating the warning in case readers read one but not the other. I just copied the PR description from the doc (lazy me).

You actually only get interleaving at await and await*, the async construct just enqueues, and async* is inert. 
The real difference is that await* may not yield at all, and won't, in itself, commit.

design/asynccps.md
crusso, line 37:
```suggestion
The only terms that introduce effect `A` is `await`, `await*`, `try` or `throw` - the effect is masked by its innermost enclosing `async` or `async*` (if any)
```

src/ir_def/check_ir.ml
```diff
@@ -222,12 +223,12 @@ let rec check_typ env typ : unit =
...
   | T.Opt typ ->
     check_typ env typ
-  | T.Async (typ1, typ2) ->
+  | T.Async (s, typ1, typ2) ->
     check_typ env typ1;
     check_typ env typ2;
     check env no_region env.flavor.Ir.has_async_typ "async in non-async flavor";
     let t' = T.promote typ2 in
-    check_shared env no_region t'
+    if s = T.Fut then check_shared env no_region t'
```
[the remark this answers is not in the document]
crusso: Reverted.

design/asynccps.md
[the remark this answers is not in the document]
crusso: Adopted

crusso:
> I know I am a little biased but believe this could be a viable approach as well. Using `C#`-style semantics for `async*/await*`. 

That would mean eagerly executing the `async*` which Andreas would object to  (even more) since the potential `yield&commit` point is not apparent from a function call that returns an async*. That's really the point of this PR: to delay the effect until an explicit `await*`.

>Also the current approach seems useful: Maybe one could forbid repeated `await*` on the same future (by a runtime exception)?

Yeah, I'm still not sure what to do about this, if anything. Making it future like fixes only half the problem, because it still has no effect unless awaited, unlike async, which sends a message regardless of whether awaited or not.

My other thought was to warn about async* values that don't appear under `await*` or in contexts expecting an `async*` (exploiting bi-directional type checking to suppress warnings):

Something like:
```
func f() : async* Int {
};
func g(a : async* Nat) : async * Nat {
   await* a;
};
func h(_ : Any) { };

let _ = f(); // warn, thunk discarded
let _ : async* Int = f() // no warning, person might know what they are doing.
await* f(); // no warning
g(f()) // no warning, since g explicitly takes an async*
func h(f()) // warn, since h takes an `Any`, not an async*.
```
but that all seems ad hoc.

However, I think most people use the pattern of `await* f()` and not treat the thunk in a first-class way. So perhaps I'm worrying about nothing.

design/asynccps.md
```diff
@@ -19,6 +19,8 @@ e = x
   | if e then e else e
   | async e
   | await e                // may throw
+  | async* e
+  | await* e
```
ggreif, line 23: this may throw too, right?

doc/md/language-manual.md
```diff
@@ -2186,6 +2197,77 @@ Note: suspending computation on `await`, regardless of the dynamic status of the
...
+The delayed computation can be executed using `await*`, producing one evaluation
+of the computation `<block-or-exp>`.
+
+:::danger
+
+Note that `async <block-or-exp>` has the effect of scheduling a single asynchronous computation of `<exp>`, regardless of whether its result, a future, is consumed with an `await`.
+Moreover, each additional consumption by an `await` just returns the previous result, without repeating the computation.
+
+In comparison, `async* <block-or_exp>`,  has *no effect* until its value is consumed by an `await*`.
+Moreover, each additional consumption by an `await*` will trigger a new evaluation of `<block-or-exp>`.
```
ggreif, line 2226:
```suggestion
Moreover, each additional consumption by an `await*` will trigger a new evaluation of `<block-or-exp>`, including repeated effects.
```

src/ir_def/construct.ml
```diff
@@ -132,28 +132,29 @@ let assertE e =
...
   }
 
-let awaitE e =
-  { it = PrimE (AwaitPrim, [e]);
+let awaitE s e =
+  let (s, _ , typ) = T.as_async (T.normalize (typ e)) in
+  { it = PrimE (AwaitPrim s, [e]);
     at = no_region;
-    note = Note.{ def with typ = snd (T.as_async (T.normalize (typ e))) ; eff = T.Await }
+    note = Note.{ def with typ = typ; eff = T.Await }
```
ggreif, line 145:
```suggestion
    note = Note.{ def with typ; eff = T.Await }
```

src/ir_def/construct.ml
```diff
@@ -570,15 +576,18 @@ let funcD ((id, typ) as f) x exp =
 let nary_funcD ((id, typ) as f) xs exp =
   letD f (nary_funcE id typ xs exp)
 
-(* Continuation types *)
+(* Continuation types with explicit answer typ *)
 
-let answerT = T.unit
+let contT typ ans_typ = T.Func (T.Local, T.Returns, [], T.as_seq typ, T.as_seq ans_typ)
```
ggreif, line 581:
```suggestion
let contT typ ans_typ = T.(Func (Local, Returns, [], as_seq typ, as_seq ans_typ))
```

src/ir_def/construct.ml
```diff
@@ -570,15 +576,18 @@ let funcD ((id, typ) as f) x exp =
...
 
-let contT typ = T.Func (T.Local, T.Returns, [], T.as_seq typ, [])
+let err_contT ans_typ =  T.Func (T.Local, T.Returns, [], [T.catch], T.as_seq ans_typ)
```
ggreif, line 583:
```suggestion
let err_contT ans_typ =  T.(Func (Local, Returns, [], [catch], as_seq ans_typ))
```

src/ir_def/construct.ml
```diff
@@ -570,15 +576,18 @@ let funcD ((id, typ) as f) x exp =
...
 
-let err_contT =  T.Func (T.Local, T.Returns, [], [T.catch], [])
+let answerT typ : T.typ =
+  match typ with
+  | T.Func (T.Local, T.Returns, [], ts1, ts2) -> T.seq ts2
+  | _ -> assert false
 
-let cpsT typ = T.Func (T.Local, T.Returns, [], [contT typ; err_contT], [])
+let cpsT typ ans_typ = T.Func (T.Local, T.Returns, [], [contT typ ans_typ; err_contT ans_typ], T.as_seq ans_typ)
```
ggreif, line 590:
```suggestion
let cpsT typ ans_typ = T.(Func (Local, Returns, [], [contT typ ans_typ; err_contT ans_typ], as_seq ans_typ))
```

src/mo_interpreter/interpret.ml
```diff
@@ -115,6 +115,7 @@ struct
 
   let queue work = Queue.add work q
   let yield () =
+    (*    Printf.printf "!"; (* TODO: delete me *) *)
```
ggreif, line 118: delete!

src/mo_frontend/typing.ml
```diff
@@ -1420,15 +1421,15 @@ and infer_exp'' env exp : T.typ =
         scopes = T.ConEnv.add c exp.at env.scopes } in
     let t = infer_exp env' exp1 in
     let t' = T.open_ [t1] (T.close [c] t)  in
-    if not (T.shared t') then
+    if s = Type.Fut && not (T.shared t') then
```
luc-blaeser, line 1424: Not sure whether `s = Type.Fut &&` is needed if `async*` always has a shared type.

design/asynccps.md
crusso, line 23:
```suggestion
  | await e                // may throw
  | async* e
  | await* e               // may throw
```

src/mo_interpreter/interpret.ml
crusso, line 118:
```suggestion
```

src/mo_frontend/typing.ml
crusso: I would like to remove the restriction for both types, but should do that in a separate PR. Thanks for finding this.

crusso, line 1424:
```suggestion
    if not (T.shared t') then
```

crusso: Reverted

crusso:
> I don't know if I understand your concern already. I think it is possible to disable future references to the async* f() function. To disable the use of.
> 
> ```
> let a = f(); 
> await* a;
> ```
> 
> Must use.
> 
> ```
> let a = await* f();
> ```

crusso:
@luc-blaeser @ggreif should I implement a type based warning or restriction, or just merge this is as is?

ggreif:
Leave as-is for now. But we should not advertise this feature as a beginner's _solves-all-problems_ tool.

Changelog.md
```diff
@@ -2,11 +2,24 @@
 
 * motoko (`moc`)
 
+  * Add new keywords `async*` and `await*` (note the `*`) for efficient abstraction of asynchronous code.
```
ggreif, line 5:
```suggestion
  * Add new keywords `async*` and `await*` (note the `*`) for efficient abstraction of asynchronous code (#3609).
```

crusso:
>  But we should not advertise this feature as a beginner's _solves-all-problems_ tool.

Should I change the Changelog entry?

ggreif:
> Should I change the Changelog entry?

`This feature is experimental and may evolve in future.` is enough I think.

PR #3612, opened 2022-11-29
Fix bugs in memory sanity check of generational GC

luc-blaeser:
Fix two bugs (both only present in debug mode):
* The memory sanity check in the generational GC has to support null pointers that can occur in `text_iter`.
* Fix a broken debug assert in remembered set `contains()` function that is used in barrier coverage check.


test/run-drun/text-iter-gc.mo
```diff
@@ -0,0 +1,18 @@
...
+   let iterator = text.chars();
+   await async {}; // Trigger GC.
+   ignore iterator.next();
+  }
+};
+
+a.go(); //OR-CALL ingress go "DIDL\x00\x00"
+
+//SKIP run-ir
+//SKIP run-low
```
crusso, line 18:
```suggestion
//SKIP run-low

```

crusso (approved):
LGTM (Looks good to me)

rts/motoko-rts/src/gc/generational/remembered_set.rs
```diff
@@ -105,9 +105,9 @@ impl RememberedSet {
         }
     }
 
-    // Will only be used for debug assertions in future (barrier coverage check).
+    // Only used for debug assertions (barrier coverage check).
     pub unsafe fn contains(&self, value: Value) -> bool {
-        debug_assert!(is_null_ptr_value(value));
+        debug_assert!(!is_null_ptr_value(value));
```
crusso, line 110: Oh dear!

crusso (approved):

PR #3614, opened 2022-11-30
make `setGlobalTimer` a scoped capability

ggreif:
In the review walkthrough (SGM 2022-11-29) @crusso suggested that `setGlobalTimer` should be a capability in the scope of `system func timer` only.

This shields libraries from moving the global timer's expiration around without central coordination.

There is an initial call to the `system func timer` just after the canister is started, to allow setting up the first actual expiration. This initial priming of the global timer is only added when there is a _user-written_ `system func timer` present, and internally we detect this in the backend by observing a source location on the function body. (Defaulted and disabled timers will have no non-trivial srcloc on the body.)


doc/md/language-manual.md
```diff
@@ -1209,14 +1209,14 @@ The declaration `<dec>` of a `system` field must be a manifest `func` declaratio
 | name          | type                                                          | description         |
 |---------------|---------------------------------------------------------------|---------------------|
 | `heartbeat`   | `() -> async ()`                                              | heartbeat action    |
-| `timer`       | `() -> async ()`                                              | timer action        |
+| `timer`       | `(Nat64 -> ()) -> async ()`                                   | timer action        |
```
crusso, line 1212: Shouldn't this be Nat64 -> Nat64 (or even Int64 to Int64 - does the API guarantee this is unsigned?)

crusso (commented):
I'm no wondering whether scoping it to just the timer system method is actually good enough, since the canister has to call `ic0.global_time_set(n>0)` to start the timer in the first place.

Annoying.

doc/md/language-manual.md
ggreif: It is unsigned. But as Andriy pointed out, inside of the callback the timer is reset to 0, so there is no point in returning anything. This also makes it a pure setter, which is nicer.

ggreif:
> I'm now wondering whether scoping it to just the timer system method is actually good enough, since the canister has to call `ic0.global_time_set(n>0)` to start the timer in the first place (unless the init method does it for you with argument 1, say)
> 
> Annoying.

Damn, that's a very good point. This happens when I try to outsmart myself: I feel outsmarted... Good that I made this a disposable PR.

ggreif:
> (unless the init method does it for you with argument 1, say)

I gave this some thought, and it doesn't appear as appalling to me as before. Let's see how I can fit it in.

doc/md/language-manual.md
ggreif: So I am testing this "being reset to 0" claim, but cannot confirm (at least on `drun`). `ic-ref` isn't tested yet. See https://github.com/dfinity/motoko/pull/3614.

src/codegen/compile.ml
```diff
@@ -9985,7 +9985,12 @@ let compile mode rts (prog : Ir.prog) : Wasm_exts.CustomModule.extended_module =
   compile_init_func env prog;
   let start_fi_o = match E.mode env with
     | Flags.ICMode | Flags.RefMode ->
-      IC.export_init env;
+      let start_timer =
+        !Flags.global_timer &&
```
ggreif, line 9989: this is redundant

doc/md/language-manual.md
```diff
@@ -1209,14 +1209,14 @@ The declaration `<dec>` of a `system` field must be a manifest `func` declaratio
...
 
--   `timer`, when declared, is called as a response of the canister global timer's expiration. When not declared (and in absence of the `-no-timer` flag) this system action is defaulted by the compiler. The canister's global timer can be manipulated with the system API `global_timer_set` upon which libraries can build their own abstractions. The corresponding primitive is `setGlobalTimer`. When using the default mechanism, `setGlobalTimer` shouldn't be used, instead `setTimer` and `cancelTimer` are provided. 
+-   `timer`, when declared, is called as a response of the canister global timer's expiration. The canister's global timer can be manipulated with the passed-in function argument of type `Nat64 -> ()` upon which libraries can build their own abstractions. When not declared (and in absence of the `-no-timer` flag), this system action is defaulted by the compiler (additionally `setTimer` and `cancelTimer` are provided as primitives).
```
crusso, line 1219:
```suggestion
-   `timer`, when declared, is called as a response of the canister global timer's expiration. The canister's global timer can be manipulated with the passed-in function argument of type `Nat64 -> ()` upon which libraries can build their own abstractions. When not declared (and in absence of the `-no-timer` flag), this system action is provided with default implementation by the compiler (additionally `setTimer` and `cancelTimer` are provided as primitives).
```
Where are the primitives provided? In prim or base/Timers.mo? I'd maybe just mention Timers.mo since prim is an implementation detail.

src/lowering/desugar.ml
```diff
@@ -320,13 +320,19 @@ and call_system_func_opt name es obj_typ =
...
       when id.it = name ->
       Some (
         match name with
         | "timer" when not !Mo_config.Flags.global_timer -> assert false;
-        | "timer"
+        | "timer" ->
+           let timer =
+             blockE
+               [ expD T.(callE (varE (var id.it note)) [Any]
+                           (varE (var "@set_global_timer" (Func (Local, Returns, [], [Prim Nat64], []))))) ]
```
crusso, line 333:
```suggestion
               [ expD T.(callE (varE (var id.it note)) [Any]
                   (varE (var "@set_global_timer" (Func (Local, Returns, [], [Prim Nat64], []))))) ]
```

src/codegen/compile.ml
ggreif: removed

doc/md/language-manual.md
ggreif: Looks like, indeed, the most recent versions of `ic-ref` and `drun` reset the global timer to 0 before entering the expiration callback.

crusso (approved):
Nice one!

doc/md/language-manual.md
ggreif: The primitives come from `prim.mo` (and co.)

nix/sources.json
```diff
@@ -27,23 +27,23 @@
         "homepage": "",
         "owner": "dfinity",
         "repo": "ic",
-        "rev": "b5fa78c590188b96f9635f4e82ed1e958884e19c",
+        "rev": "b5fa78c5",
```
ggreif, line 30: this should be a long hash

PR #3617, opened 2022-12-02
resort to one-element mutable arrays as a workaround for aliased `var` field

ggreif:
This is a response to the review comment about better not use experimental features in production code.

Personally I am not sure this is an issue :-)

### DIFF
--- a/src/prelude/internals.mo
+++ b/src/prelude/internals.mo
@@ -433,12 +433,17 @@ func @call_raw(p : Principal, m : Text, a : Blob) : async Blob {
 // fundamental node invariant: max_exp ante <= expire <= min_exp dopo
 // corollary: if expire == 0 then the ante is completely expired
 //
-type @Node = { var expire : Nat64; id : Nat; delay : ?Nat64; job : () -> async (); ante : ?@Node; dopo : ?@Node };
+// Note: Below the `expire` field is an encoding of an aliased mutable field with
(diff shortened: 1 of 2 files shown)


crusso (approved):
Why not use an object? But I'll let it go.... thanks for making change.

src/prelude/internals.mo
```diff
@@ -481,26 +486,26 @@ func @timer_helper() : async () {
...
+            if (n.expire[0] > 0 and n.expire[0] <= now and gathered < thunks.size()) {
                 thunks[gathered] := ?(n.job);
                 switch (n.delay) {
                     case (?delay) {
                         // re-add the node
-                        let expire = n.expire + delay;
+                        let expire = n.expire[0] + delay;
                         // N.B. insert only works on pruned nodes
                         func reinsert(m : ?@Node) : @Node = switch m {
-                            case null ({ n with var expire; ante = null; dopo = null });
```
kentosugama, line 492: So is this the operation that was problematic? What is particular about aliased `var` fields that is a problem?

ggreif:
> Why not use an object?

Objects would be even bigger in the heap (array: 12 bytes, object: 16 bytes and static heap vocabulary) and add even more syntax. Contrast this with mutable cells: 8 bytes.

nomeata:
And objects are more expensive to index into.

PR #3618, opened 2022-12-03
feat: add configurable interpreter step limit for `moc.js`

rvanasa:
Adds an optional deterministic step limit to the `moc.js` interpreter. 

This feature will protect users from accidentally crashing the browser with a code snippet such as `while true {}` now that we are using a trampoline (#3541), so I suppose this is a nice problem to have.

### DIFF
--- a/src/js/common.ml
+++ b/src/js/common.ml
@@ -59,6 +59,9 @@ let js_check source =
   Mo_types.Cons.session (fun _ -> 
     js_result (Pipeline.check_files [Js.to_string source]) (fun _ -> Js.null))
 
+let js_set_run_step_limit limit =
+  Mo_interpreter.Interpret.step_limit := limit
+
 let js_run list source =
   Mo_types.Cons.session (fun _ -> 
     let list = Js.to_array list |> Array.to_list |> List.map Js.to_string in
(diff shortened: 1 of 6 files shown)


crusso (commented):
If needs must.

Does the broswer hand indefinitely or offer a cancel script prompt eventually?

If the former, I wonder why a whiile (true) diverges, but the similar bounded loop on the trampoline PR stack-overflowed, decrement counter for 163840 steps exits with stack overflow....

crusso (approved):

rvanasa:
> Does the browser hang indefinitely or offer a cancel script prompt eventually?

Browser hangs indefinitely and requires reopening the tab. 

> I wonder why a `while (true) { }` diverges, but the similar bounded loop on the trampoline PR stack-overflowed

Yeah, same. I'll try to reverse-engineer the stack trace in case there's a way we can mitigate whatever is going on here.

crusso:
Yeah, I actually didn't think to look a the developer tool pane to figure out why the overflow was happening. It could be that we are missing some extra trampolines in other interpreter functions, but I experimented with adding some and didn't get anywhere (or maybe didn't refresh the browser hard enough).

Issue #3622, opened 2022-12-06
todo: restore GC in heartbeat once IC heartbeats support DTS


crusso:
There are two instances of this problem:
- heartbeat
- timer

luc-blaeser:
Done

PR #3623, opened 2022-12-06
hotfix: suppress GC in heartbeat

crusso:
Mainnet doesn't yet support DTS for `canister_heartbeat` yet, so a long GC may fail to complete when it would in an ordinary message supporting DTS.

This PR suppresses the GC call, leaving it to the already scheduled async message executing Motoko `system` function `heartbeat`, which will run with DTS enabled, avoiding the issue (and a potential space leak).

#3622 tracks the need to revert this hack once IC heartbeats support DTS.

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -3896,8 +3896,8 @@ module IC = struct
         G.i (Call (nr (E.built_in env "heartbeat_exp"))) ^^
         (* TODO(3622)
            Until DTS is implemented for heartbeats, don't collect garbage here,
-           just record mutator_instructiond and leave GC scheduling to the
-           already sheduled async message running `system` function `hearbeat` *)
+           just record mutator_instructions and leave GC scheduling to the
+           already sheduled async message running `system` function `heartbeat` *)
         GC.record_mutator_instructions env (* future: GC.collect_garbage env *)))
     in
     E.add_export env (nr {
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -3894,7 +3894,11 @@ module IC = struct
     let fi = E.add_fun env "canister_heartbeat"
       (Func.of_body env [] [] (fun env ->
         G.i (Call (nr (E.built_in env "heartbeat_exp"))) ^^
-        GC.collect_garbage env))
+        (* TODO(3622)
+           Until DTS is implemented for heartbeats, don't collect garbage here,
+           just record mutator_instructiond and leave GC scheduling to the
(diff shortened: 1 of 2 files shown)


src/codegen/compile.ml
```diff
@@ -3894,7 +3894,11 @@ module IC = struct
     let fi = E.add_fun env "canister_heartbeat"
       (Func.of_body env [] [] (fun env ->
         G.i (Call (nr (E.built_in env "heartbeat_exp"))) ^^
-        GC.collect_garbage env))
+        (* TODO(3622)
+           Until DTS is implemented for heartbeats, don't collect garbage here,
+           just record mutator_instructiond and leave GC scheduling to the
```
ggreif, line 3899:
```suggestion
           just record mutator_instructions and leave GC scheduling to the
```

src/codegen/compile.ml
```diff
@@ -3894,7 +3894,11 @@ module IC = struct
...
+           already sheduled async message running `system` function `hearbeat` *)
```
ggreif, line 3900:
```suggestion
           already sheduled async message running `system` function `heartbeat` *)
```

src/codegen/compile.ml
```diff
@@ -3894,7 +3894,11 @@ module IC = struct
     let fi = E.add_fun env "canister_heartbeat"
       (Func.of_body env [] [] (fun env ->
         G.i (Call (nr (E.built_in env "heartbeat_exp"))) ^^
-        GC.collect_garbage env))
+        (* TODO(3622)
```
ggreif, line 3897: I saw this while implementing `system func timer`, and I remember considering this a bit expensive/dangerous. But with DTS not covering this, yuck!

crusso: Yeah, I remember you suggesting it but thought better no to mess with the GC. This just a temp hack. Let's see what Luc thinks.

luc-blaeser:
Looks good. I also checked it with an example (that creates garbage in the heartbeat). The question is whether this extra GC schedule call is needed at all in the future even if DTS is available for the heartbeat call (since the async function contains GC scheduling anyway).

crusso:
> Looks good. I also checked it with an example (that creates garbage in the heartbeat). The question is whether this extra GC schedule call is needed at all in the future even if DTS is available for the heartbeat call (since the async function contains GC scheduling anyway).

My initial intention was to hopefully remove the extra async hop in future, so the heartbeat code runs during the heartbeat method. The fact that it runs with a delay is an accident to the async/await implementation (that async*/await* will optimize). If we do that, we'll need to do the GC I think.

PR #3630, opened 2022-12-07
feat: throw on send failure

crusso:
On the IC, the act of making a call to a canister function can fail, so that the call cannot (and will not be) performed.
 This can happen due to a lack of canister resources, typically because the local message queue for the destination canister is full,  or because performing the call would reduce the current cycle balance of the calling canister to a level below its freezing threshold.
Such call failures, instead of trapping, are now reported by throwing an `Error` with new `ErrorCode` `#call_error { err_code = n }`, where `n` is the non-zero `err_code` value returned by the IC.
Like other errors, call errors can be caught and handled using `try ... catch ...` expressions, if desired.

 The constructs that now may throw call errors, instead of trapping as with previous versions of Motoko, are:
    * calls to `shared` functions (including oneway functions that return `()`).
      These involve sending a message to another canister, and can fail when the queue for the destination canister is full.
    * calls to local functions with return type `async`. These involve sending a message to self, and can fail when the local queue for sends to self is full.
    * `async` expressions. These involve sending a message to self, and can fail when the local queue for sends to self is full.
    * `await` expressions. These can fail on awaiting an already completed future, which requires sending a message to self to suspend and commit state.

(On the other hand, `async*` (being delayed) cannot throw, and evaluating `await*` will at most propagate an error from its argument but not, in itself, throw.)

# Implementation

Add explicit checks for message send failures, reified as catchable errors (exceptions).

* saves status of `ic0.call_perform` to new (internal) global
* test global after each message send, of which there are several flavours (async expressions, message sends and dynamic calls (a.k.a. `call_raw`), awaits on already completed promises)

Most of the transformation happen in ir_passes/await.ml where we have access to the failure/throw continuation, if any.

# Tasks

- [x] message sends (request and oneways)
- [x] async expressions 
- [x] raw calls (dynamic calls)
- [x] implicit yield on completed future (yuck).  Note this means not every await will commit, since, in (not so) rare cases, the self send queue will be full when trying to artificially suspend on a completed future. Not sure what to do in that case, trap?
- [x] test local asyncs (might be broken)
- [x] decide on error message and/or introduction of a new error code (would be a breaking change).
- [x] figure out how to fix the effect system (async can now throw synchronously, so need to have effect 'Await' to always be in cps context where throws can be handled)
- [x] update documentation for Error.mo in motoko-base.
- [x] split monolithic test into separate tests for async, send, await etc.

# Open Issues (Feedback wanted)

One minor worry with  this PR is that it is a breaking change. Previously a block of code that sends multiple messages, one of which fails to send, would trap, rolling back all the sends to the previous commit point. Now, the successful sends will remain enqueued, and the error/exception will flow to the nearest handler. If there is none, the method will exit with a canister_reject error, which will still send the other enqueued messages and commit state changes.

This is a breaking change, so we should release with 0.8.0.

I've added an opt-out compiler flag `--throw-on-call-error` to restore the old behaviour for users who don't care to audit their code.


src/ir_passes/await.ml
```diff
@@ -295,8 +338,8 @@ and c_exp' context exp k =
     letcont f (fun f ->
     letcont k (fun k ->
     match eff exp1 with
-    | T.Triv ->
-      varE k -*- (t_exp context exp1)
+    | T.Triv (* ->  FIXME
+      varE k -*- (t_exp context exp1) *)
```
crusso, line 342: revert if possible.

crusso: Turns outs this optimization no longer applies since the handler will always be needed in case we do an async expression with effect Triv (since we don't yet track the potential synchronous failure of async expressions in the effect system)

test/run-drun/self-send.mo
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+};
+
+//SKIP run
+//SKIP run-ir
+//SKIP run-low
+//SKIP ic-ref-run
+
+//await a.go(); //OR-CALL ingress go "DIDL\x00\x00"
```
crusso, line 288:
```suggestion
//await a.go(); //OR-CALL ingress go "DIDL\x00\x00"

```

test/run-drun/self-send-async.mo
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+};
+
+//SKIP run
+//SKIP run-ir
+//SKIP run-low
+//SKIP ic-ref-run
+
+//await a.go(); //OR-CALL ingress go "DIDL\x00\x00"
```
crusso, line 63:
```suggestion
//await a.go(); //OR-CALL ingress go "DIDL\x00\x00"

```

src/ir_passes/await.ml
```diff
@@ -536,17 +619,35 @@ and t_comp_unit context = function
...
+  | Ir.PrimE (Ir.TupPrim, []) ->
+     exp
+  | _ ->
+     let throw = fresh_err_cont T.unit in
+     let context' = LabelEnv.add Throw (Cont (ContVar throw)) context in
+     let e = fresh_var "e" T.catch in
+     { (blockE [
+          funcD throw e (tupE[]);
+        ]
+       (c_exp context' exp (meta (T.unit) (fun v1 -> tupE []))))
```
crusso, line 644:
```suggestion
        (c_exp context' exp (meta (T.unit) (fun v1 -> tupE []))))
```

src/ir_passes/await.ml
```diff
@@ -61,15 +84,35 @@ type label_sort = Cont of kont | Label
...
-and t_exp' context exp' =
-  match exp' with
+(* Trivial translation of pure terms (eff = T.Triv) *)
+and t_exp context exp =
+  if (eff exp != T.Triv) then
+  begin
+      Printf.printf "in IR:\n%s"
+        (Wasm.Sexpr.to_string 80 (Arrange_ir.exp exp));
+      assert false;
+  end;
```
crusso, line 110:
```suggestion
  assert (eff exp = T.Triv);
```

src/codegen/compile.ml
```diff
@@ -7259,7 +7281,7 @@ module FuncDec = struct
       (fun get_cb_index ->
         get_cb_index ^^
         BoxedSmallWord.box env ^^
-        Serialization.serialize env Type.[Prim Nat32])
+          Serialization.serialize env Type.[Prim Nat32])
```
crusso, line 7284:
```suggestion
        Serialization.serialize env Type.[Prim Nat32])
```

src/codegen/compile.ml
```diff
@@ -4067,7 +4077,7 @@ module IC = struct
         (fun env -> system_call env "msg_reject_msg_size")
         (fun env -> system_call env "msg_reject_msg_copy")
         (fun env -> compile_unboxed_const 0l)
-    )
+      )
```
crusso, line 4080:
```suggestion
    )
```

src/ir_def/check_ir.ml
```diff
@@ -363,7 +363,14 @@ let store_typ t  =
...
   (* helpers *)
-  let check p = check env exp.at p in
+  let check p s =
+    try
+      check env exp.at p s
+    with e ->
+      Printf.printf "in IR:\n%s"
+       (Wasm.Sexpr.to_string 80 (Arrange_ir.exp exp));
+      raise e
+  in
```
crusso, line 373:
```suggestion
  let check p = check env exp.at p in
```

src/ir_def/check_ir.ml
```diff
@@ -390,6 +397,7 @@ let rec check_exp env (exp:Ir.exp) : unit =
   | PrimE (p, es) ->
     List.iter (check_exp env) es;
     begin match p, es with
+    | CallPrim insts, [exp1; exp2; _] (*HACK*)
```
crusso, line 400:
```suggestion
    | CallPrim insts, [exp1; exp2]
```

src/ir_def/ir.ml
```diff
@@ -78,10 +78,11 @@ and exp' =
 
 and system = {
   meta : meta;
+  (* TODO: use option expressions for (some or all of) these *)
   preupgrade : exp;
   postupgrade : exp;
   heartbeat : exp;
-  timer : exp;
+  timer : exp; (* TODO: use an option type *)
```
crusso, line 85:
```suggestion
  timer : exp; (* TODO: use an option type : (Default of exp | UserDefined of exp) option *)
```

test/run-drun/ok/self-send.drun-run.ok
```diff
@@ -0,0 +1,28 @@
+ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
+ingress Completed: Reply: 0x4449444c0000
+debug.print: test1:
+debug.print: test1: (#canister_reject, "ic0.call_perform failed")
+debug.print: test2:
+debug.print: caught (#call_error({err_code = 2}), "ic0.call_perform failed")
```
crusso, line 6: Is the code really 2 or am I doing the conversion wrong?

src/mo_frontend/effect.ml
```diff
@@ -25,13 +25,32 @@ let typ phrase = phrase.note.note_typ
...
 let is_triv phrase  =
-    eff phrase = T.Triv
+  eff phrase = T.Triv
+
+let is_shared_func exp =
+  T.(match typ exp with
+    | Func (Shared _, _, _, _, _) -> true
+    | _ -> false)
+
+let is_local_async_func exp =
```
luc-blaeser, line 35: Same helper functions seem to occur in `effect.ml`, `ir_effect.ml`, and `await.ml`. Could this be shared in a common place?

luc-blaeser (approved):
The implementation looks very good. 

I am not sure what the impact of the changing behavior (no rollback of previous send/side effects on a send failure) is in practice. A possibility (as discussed) could be to offer this feature only as a special compiler option.

src/mo_frontend/effect.ml
crusso: Good catch!

Changelog.md
```diff
@@ -1,5 +1,25 @@
...
+
+    The constructs that now throw call errors, instead of trapping as with previous version of Motoko are:
+    * calls to `shared` functions (including oneway functions that return `()`).
+      These involve sending a message to another canister, and can fail when the queue for the destination canister is full.
+    * calls to local functions with return type `async`. These involve sending a message to self, and can fail when the local queue for sends to self is full.
+    * `async` expressions. These involve sending a message to self, and can fail when the local queue for sends to self is full.
+    * `await` expressions. These can fail on awaiting an already completed future, which requires sending a message to self to suspend and commit state.
+
+    (On the other hand, `async*` (being delayed) cannot throw, and evaluating `await*` will at most propagate an error from its argument but not, in itself, throw.)
+
```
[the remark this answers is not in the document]
crusso: Yeah, I'm still undecided how to proceed, but will definitely add an example showing the difference.

src/mo_frontend/effect.ml
crusso: Done in #3697

crusso:
@luc-blaeser @ggreif @nomeata PTAL.

I've basically added an opt out compiler flag, some examples in the Changelog and better textual message on errors to match the once we previously use for traps.

test/run-drun/empty-actor.mo
```diff
@@ -8,13 +8,13 @@ actor {};
 // CHECK-NEXT:    call $init
 // CHECK-NEXT:    i32.const 0
 // CHECK-NEXT:    call 30
-// CHECK-NEXT:    global.set 1
+// CHECK-NEXT:    global.set 2
```
crusso, line 11:
```suggestion
// CHECK-NEXT:    global.set 3
```

test/run-drun/empty-actor.mo
```diff
@@ -8,13 +8,13 @@ actor {};
...
 // CHECK-NEXT:    call ${{copying_gc|compacting_gc|generational_gc}}
 // CHECK-NEXT:    i32.const 0
 // CHECK-NEXT:    call 30
-// CHECK-NEXT:    global.get 1
+// CHECK-NEXT:    global.get 2
```
crusso, line 15:
```suggestion
// CHECK-NEXT:    global.get 3
```

test/run-drun/empty-actor.mo
```diff
@@ -8,13 +8,13 @@ actor {};
...
 // CHECK-NEXT:    i64.sub
-// CHECK-NEXT:    global.set 2
+// CHECK-NEXT:    global.set 3
```
crusso, line 17:
```suggestion
// CHECK-NEXT:    global.set 4
```

crusso (commented):
Fix test

Changelog.md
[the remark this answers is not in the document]
crusso: Done in Changelog

test/run-drun/self-send-trap.mo
```diff
@@ -0,0 +1,294 @@
...
+  public func request() : async () {
+  };
+
+  public func oneway() : () {
+  };
+
+  public func test1() : async () {
+    var n = 0;
+    while (n < DOUBLE_CAPACITY) {
+//      Prim.debugPrint(debug_show n);
```
ggreif, line 24: should these stay in this file (there are a bunch of them)

crusso: Done

src/codegen/compile.ml
```diff
@@ -7221,9 +7243,25 @@ module FuncDec = struct
...
       IC.system_call env "call_perform" ^^
-      (* Check error code *)
+      IC.set_call_perform_status env ^^
+      Blob.lit env message ^^
+      IC.set_call_perform_message env ^^
+      IC.get_call_perform_status env ^^
+      (* save error code, cleanup on error *)
       G.i (Test (Wasm.Values.I32 I32Op.Eqz)) ^^
-      E.else_trap_with env (Printf.sprintf "could not perform %s" purpose)
+      G.if0
```
ggreif, line 7252: I am a bit confused by this. `status` = 0 means no error. `Test` will turn that around to `1`. Then `if0` will proceed to the _else_ branch, cleaning up...

My feeling is that we should be able to eliminate the `Test` instruction.

src/ir_def/construct.ml
```diff
@@ -135,7 +137,9 @@ let assertE e =
 let asyncE s typ_bind e typ1 =
   { it = AsyncE (s, typ_bind, e, typ1);
     at = no_region;
-    note = Note.{ def with typ = T.Async (s, typ1, typ e); eff = T.Triv }
+    note =
+      Note.{ def with typ = T.Async (s, typ1, typ e);
+                      eff = if s = T.Fut then T.Await else T.Triv }
```
ggreif, line 142: suggestion `T.(if s = Fut ...)`

src/ir_def/construct.ml
```diff
@@ -245,6 +249,12 @@ let blockE decs exp =
       note = Note.{ def with typ; eff }
     }
 
+let nat32E n =
+  { it = LitE (Nat32Lit n);
+    at = no_region;
+    note = Note.{ def with typ = T.Prim (T.Nat32) }
```
ggreif, line 255: `T.(Prim Nat32)`

src/ir_def/construct.ml
```diff
@@ -299,14 +309,17 @@ let callE exp1 typs exp2 =
     | T.Non -> T.Non
     | _ -> raise (Invalid_argument "callE expect a function")
   in
+  let p = CallPrim typs in
+  let es = [exp1; exp2] in
   { it = PrimE (CallPrim typs, [exp1; exp2]);
```
ggreif, line 314: `{ it = PrimE (p, es);`

src/ir_def/construct.ml
```diff
@@ -727,3 +740,16 @@ let objE sort typ_flds flds =
...
+  ifE
+    (callE
+      (varE (var "@call_succeeded"
+        T.(Func (Local, Returns, [], [], [bool]))))
+      [] (unitE ()))
+    success
+    (mk_failure
+      (callE
+        (varE (var "@call_error"
+          T.(Func (Local, Returns, [], [], [T.error]))))
```
ggreif, line 754: `T.(Func (Local, Returns, [], [], [error]))))`

src/codegen/compile.ml
crusso: Agree, fixing in 3727

src/ir_def/construct.ml
crusso: Agreed, fixing in #3727

src/ir_def/construct.ml
crusso: Agreed, fixing in #3727

src/ir_def/construct.ml
crusso: Agreed, fixing in #3Agreed, fixing in #3727727

src/ir_def/construct.ml
crusso: Agreed, fixing in #3727

src/ir_def/ir_effect.ml
```diff
@@ -20,12 +20,29 @@ let is_triv phrase = eff phrase = T.Triv
 
 let effect_exp (exp: exp) : T.eff = eff exp
 
+let is_async_call p exps =
+  match (p, exps) with
+  | CallPrim _, [exp1; _] ->
+    T.is_shared_func (typ exp1) ||
+    T.is_local_async_func (typ exp1)
+  | OtherPrim "call_raw", _ ->
+    true
```
ggreif, line 29: good catch!

src/ir_def/ir_effect.ml
```diff
@@ -20,12 +20,29 @@ let is_triv phrase = eff phrase = T.Triv
 
 let effect_exp (exp: exp) : T.eff = eff exp
 
+let is_async_call p exps =
+  match (p, exps) with
+  | CallPrim _, [exp1; _] ->
+    T.is_shared_func (typ exp1) ||
+    T.is_local_async_func (typ exp1)
```
ggreif, line 27: I assume this also includes fire&forget sends?

src/ir_def/ir_effect.ml
```diff
@@ -49,7 +65,9 @@ let rec infer_effect_prim = function
     let e1 = effect_exp exp1 in
     let e2 = effect_cases cases in
     max_eff e1 e2
-  | AsyncE _ ->
+  | AsyncE (T.Fut, _, _, _) ->
+    T.Await
```
ggreif, line 69: This I interpret as the need for having a {`Triv`, `Throw`, `Await`} effect lattice, right?

src/ir_passes/async.ml
```diff
@@ -29,13 +29,10 @@ let selfcallE ts e1 e2 e3 =
     at = no_region;
     note = Note.{ def with typ = T.unit } }
 
-let error_ty = T.(Tup [ Variant [
-    {lab = "error"; typ = unit; depr = None};
-    {lab = "system"; typ = unit; depr = None}
-  ]; text])
+let error_rep_ty = T.(Tup [ Variant T.catchErrorCodes; text])
```
ggreif, line 32: `[ Variant`

src/ir_passes/async.ml
```diff
@@ -252,7 +249,11 @@ let transform prog =
                   unitE()); (* suspend *)
                 ("schedule", varP schedule, (* resume later *)
                   (* try await async (); schedule() catch e -> r(e) *)
-                  (selfcallE [] (ic_replyE [] (unitE())) (varE schedule) (projE (varE vkr) 1))) ]
+                 (let v = fresh_var "call" T.unit in
+                  letE v
+                    (selfcallE [] (ic_replyE [] (unitE())) (varE schedule) (projE (varE vkr) 1))
+                    (check_call_perform_status (varE v) (fun e -> ((projE (varE vkr) 1) -*- e)))))
```
ggreif, line 255: `((projE (varE vkr) 1) -*- e)` -> `projE (varE vkr) 1 -*- e`

ggreif: I have trouble understanding the line 255.

src/mo_types/type.mli
```diff
@@ -78,6 +78,7 @@ end
 
 val is_shared_sort : 'a shared -> bool
 
+
```
ggreif, line 81: was this intentional?

test/fail/bad-call-raw.mo
```diff
@@ -0,0 +1,5 @@
+import P "mo:⛔";
+
+actor self {
+  let a = P.call_raw("","foo",""); // reject, send capability required
```
ggreif, line 4: So this was a bug formerly?

src/mo_frontend/effect.ml
```diff
@@ -92,7 +111,9 @@ let rec infer_effect_exp (exp:Syntax.exp) : T.eff =
     let e1 = effect_exp exp1 in
     let e2 = effect_cases cases in
     max_eff e1 e2
-  | AsyncE _ ->
+  | AsyncE (T.Fut, _, _) ->
+    T.Await
```
ggreif, line 115: Here too this would ideally be `T.Throw`, right?

src/mo_frontend/effect.ml
```diff
@@ -25,13 +25,32 @@ let typ phrase = phrase.note.note_typ
...
+   is_shared_func exp1 ||
+   is_local_async_func exp1
 
 let effect_exp (exp:Syntax.exp) : T.eff = eff exp
 
 (* infer the effect of an expression, assuming all sub-expressions are correctly effect-annotated es*)
 let rec infer_effect_exp (exp:Syntax.exp) : T.eff =
   match exp.it with
+  | CallE (exp1, inst, exp2) when is_async_call exp1 inst exp2 ->
+    T.Await
```
ggreif, line 53: So calls to private `async` functions can throw now even if both `exp*` are `Triv`.

src/mo_frontend/effect.ml
```diff
@@ -25,13 +25,32 @@ let typ phrase = phrase.note.note_typ
...
+
+let is_local_async_func exp =
+ T.(match typ exp with
+   | Func (Local, Returns,
+      { sort = Scope; _ }::_,
+       _,
+      [Async (Fut, Var (_ ,0), _)]) -> true
+   | _ -> false)
+
+let is_async_call exp1 inst exp2 =
```
ggreif, line 43: `inst` and  `exp2` are irrelevant for the result

test/fail/bad-call-raw.mo
crusso: No, I just added a test.

src/ir_def/ir_effect.ml
crusso: Yes, that could be a way to go, especially if we supported throw outside of async blocks (where we could maybe compile try/catch/throw without actually going to full CPS.)

src/ir_passes/async.ml
crusso: It checks the global flag, returns unit if successfull and otherwise passes the error e to the failure continuation that throws.

src/mo_types/type.mli
crusso: The space? Probably not.

src/mo_frontend/effect.ml
crusso: Yes. As desired, because the application itself can throw, even if the function and arg are trivial. This is the usual unleashing of effects at application.

src/ir_def/ir_effect.ml
crusso: Yes, it should. We only have remote fire-and-forget and both remote request and remote oneways are shared funcs.

src/mo_frontend/effect.ml
crusso: Yes, but its a predicate on calls, so I've include them.

src/mo_frontend/effect.ml
crusso: Yes.

src/ir_passes/async.ml
crusso: Fixed in #3723

src/mo_types/type.mli
crusso: Actually it was - Andreas left two space before sections. Adding some more ;->  in #3723

src/ir_def/ir_effect.ml
ggreif: Reminder in #3742.

src/ir_passes/async.ml
ggreif: Understood!

PR #3634, opened 2022-12-08
Disable extensive memory sanity checks in CI

luc-blaeser:
Introducing a switch for extensive memory sanity checks as used in the generational GC. 

Disable these checks in CI, as they otherwise take too long with `ic-ref-run`.

To enable the checks during engineering or testing, manually set the features in `Cargo.toml` as follows:
```
[features]
default = ["ic", "memory_check"]
```

### DIFF
--- a/rts/motoko-rts/Cargo.toml
+++ b/rts/motoko-rts/Cargo.toml
@@ -13,10 +13,15 @@ crate-type = ["staticlib"]
 # we enable the "ic" feature. `native/Cargo.toml` doesn't have this feature and
 # is used in RTS tests.
 default = ["ic"]
+# default = ["ic", "memory_check"]
+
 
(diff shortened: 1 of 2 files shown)


crusso:
Did this reduce the CI times at all?

luc-blaeser:
> Did this reduce the CI times at all?

I will compare and verify this with a second CI build without this option. It still took 2h55m: https://github.com/dfinity/motoko/actions/runs/3648374585/jobs/6161720339

crusso:
I wonder if we should also just disable all sanity checks on ic-ref, in the moc compiler itself, perhaps.

crusso (approved):

luc-blaeser:
Resolved with https://github.com/dfinity/motoko/pull/3652

PR #3645, opened 2022-12-13
bugfix: fix spurious discarded field errors in --stable-compatible check

crusso:
Fix bug in stable-compatiblity check reported here:

https://forum.dfinity.org/t/weird-warning-message-when-deploying-existing-canister/17276/5.

The problem is that the verbose check (that emits multiple diagnostic errors) would, on finding a discarded field from the lhs,  recurse only on the tail of the rhs, not the rhs itself, causing all remaining fields of the lhs to be reported as discarded, even when present in the (original) rhs.

Added a small repro and manually verified it fixes the issue on the full `*.most` files shared by the user in a private forum message.

The bug itself is benign though obviously confusing to the user.


chenyan-dfinity (approved):

test/cmp/old-new.cmp
```diff
@@ -0,0 +1 @@
+files/old.most files/new.most
```
crusso, line 1:
```suggestion
files/old.most files/new.most

```

test/cmp/files/new.most
```diff
@@ -0,0 +1,141 @@
...
+stable var data_saleSoldTracker : [(Nat, AccountIdentifier__2, Nat64)];
+stable var data_saleTokensForSale : [TokenIndex__1];
+stable var data_saleTransactions : [SaleTransaction];
+stable var data_soldIcp : Nat64;
+stable var data_storedChunkSize : Nat;
+stable var data_supply : Nat;
+stable var data_tokenListingTableState : [(TokenIndex__1, Listing)];
+stable var data_tokenMetadataTableState : [(TokenIndex__1, Metadata__1)];
+stable var data_transactions : [Transaction];
+};
```
crusso, line 141:
```suggestion
};

```

test/cmp/files/old.most
```diff
@@ -0,0 +1,141 @@
...
+stable var data_saleTransactions : [SaleTransaction];
+stable var data_soldIcp : Nat64;
+stable var data_storedChunkSize : Nat;
+stable var data_supply : Nat;
+stable var data_tokenListingTableState : [(TokenIndex__1, Listing)];
+stable var data_tokenMetadataTableState : [(TokenIndex__1, Metadata__1)];
+stable var data_transactions : [Transaction];
+stable var config_royalty_address : Text;
+stable var config_royalty_fee : Nat64;
+};
```
crusso, line 141:
```suggestion
};

```

PR #3649, opened 2022-12-14
Feature/trie examples

letmejustputthishere:


crusso:
How come the md/base files changed? Did you do a make base or were these manual edits (in the wrong repo)? Just want to make sure you don't lose work next time we generate docs from motoko-base on release..

letmejustputthishere:
for `Array` they changed because I merged `master` into this branch hoping it would solve the inconsistencies I had between the motoko playground and the local preview running. i then realised the issue was solved by bumping the `moc` version instead. the other changes in `Trie` result from the `make base` command I automatically run on file changes in the `motoko-base` repo locally.

luc-blaeser:

PR #3671, opened 2023-01-03
`niv-updater-action` bump (experiment)

ggreif:
This is a test balloon for gaining experience with the `main` revision of the GH action. As soon this hits our `master` branch I can test it by triggering the action via nudging.

If it doesn't work, no biggie, the action will need another update.

_Context_: https://github.com/knl/niv-updater-action/issues/56


rvanasa (approved):

ggreif:
And it works! 😎

PR #3675, opened 2023-01-03
Support private functions and `old` expressions in Motoko-to-Viper translation

aterga:
This PR adds support for translating private Motoko functions into Viper. The translation is mostly analogous to that for public `async` functions. However, a private function does not have to be called from a state that satisfies the canister invariant (i.e., it may be called even in intermediate states), and a private function does not have to preserve the canister invariant (i.e., it may result in an intermediate state).

Additionally, we add `old` expressions to enable useful specs in private actor functions, e.g.:

```motoko
private func reward() : () {
  count += 1;
  assert:return count == old count + 1;
};
```


crusso:
(I'm off tomorrow, @ggreif can you take a look if on duty?)

src/mo_frontend/parser.mly
```diff
@@ -238,6 +238,7 @@ and objblock s dec_fields =
 
 %nonassoc IMPLIES (* see assertions.mly *)
 
+
```
ggreif, line 241: remove added line?

test/viper/claim-reward-naive.mo
```diff
@@ -0,0 +1,21 @@
...
+    assert:return count == (old count) + 1;
+  };
+
+  public shared func claim() : async () {
+    if (not claimed) {
+      reward();
+      claimed := true;
+    };
+  };
+}
```
ggreif, line 21: please add a newline

ggreif (approved):
LGTM! A few nits noted. Also test/viper/ok/private.silicon.ok is an empty file, `make -C test/viper accept` _should_ get rid of it.

Maybe also update the `src/viper/README.md` regarding the new features?

src/mo_interpreter/interpret.ml
```diff
@@ -580,6 +580,8 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
         k V.(Bool (as_bool v1 <= as_bool v2))
       )
     )
+  | OldE exp1 ->
+    interpret_exp env exp1 (fun v1 -> k v1)
```
ggreif, line 584: making `old` an identity looks like the best option if we want to preserve interpretability. Can `old` appear in all expressions or only in assertions/*conditions?

aterga: Conceptually, `old` can appear in any expression, referring to the values of actor fields in the initial state of a function invocation. For example:

```motoko
var x = 42;
private func inc() : () {
    x -= 1;
    assert x > old(x);
};
```

Should be equivalent to

```motoko
var x = 42;
private func inc() : () {
    let old_x = x;
    x -= 1;
    assert x > old_x;
};
```

While translating this code via the semantics of the identity function would not be correct:

```motoko
var x = 42;
private func inc() : () {
    x -= 1;
    assert x > x;
};
```

Since implementing the proper semantics can be tricky (engineering wise), I would suggest that we ignore `old`-expressions in the IR for now.

Wdyt?

aterga: To answer your question, for now I would allow `old` only in static assertions and conditions (not even in regular `assert`).

src/mo_frontend/parser.mly
aterga:
```suggestion
```

test/viper/claim-reward-naive.mo
aterga:
```suggestion
}

```

src/mo_interpreter/interpret.ml
ggreif: Can `old` wrap arbitrary expressions or only bindings? I.e. is `old (x + 1)` allowed? If not, then the parser or type checker should deal with the error generation. But this can wait. Let's keep it a Viper translation-only thing for now.

src/mo_frontend/definedness.ml
```diff
@@ -114,6 +114,7 @@ let rec exp msgs e : f = match e.it with
   | AndE (e1, e2)       -> exps msgs [e1; e2]
   | OrE (e1, e2)        -> exps msgs [e1; e2]
   | ImpliesE (e1, e2)   -> exps msgs [e1; e2]
+  | OldE (e)            -> exp msgs e
```
ggreif, line 117:
```suggestion
  | OldE e              -> exp msgs e
```

src/mo_frontend/effect.ml
```diff
@@ -49,6 +49,7 @@ let rec infer_effect_exp (exp:Syntax.exp) : T.eff =
   | TagE (_, exp1)
   | DotE (exp1, _)
   | NotE exp1
+  | OldE (exp1)
```
ggreif, line 52:
```suggestion
  | OldE exp1
```

src/viper/syntax.ml
```diff
@@ -41,6 +45,7 @@ and exp' =
   | AndE of exp * exp
   | OrE of exp * exp
   | Implies of exp * exp
+  | Old of exp
```
ggreif, line 48: Conceptually, is `old` just a flag on `VarE`? (I am not suggesting to change, just want to understand.)

aterga: In simple cases yes, but one old expression can wrap multiple variables (it distributes over everything, more or less). 

For example `old(x + y)` is the same value as `old(x) + old(y)` (assuming that `x` and `y` are actor fields).

aterga: In general, we could also have `old(foo())` with, e.g.: 

```motoko
private func foo() : Int {
    x
}
```

where `x` is a private field.

src/mo_interpreter/interpret.ml
aterga: `old(x + 1)` should be the same as `(old(x)) + 1`

ggreif: Gotcha!

src/lowering/desugar.ml
```diff
@@ -202,6 +202,7 @@ and exp' at note = function
   | S.AndE (e1, e2) -> (andE (exp e1) (exp e2)).it
   | S.OrE (e1, e2) -> (orE (exp e1) (exp e2)).it
   | S.ImpliesE (e1, e2) -> (impliesE (exp e1) (exp e2)).it
+  | S.OldE (e) -> (oldE (exp e)).it
```
ggreif, line 205:
```suggestion
  | S.OldE e -> (oldE (exp e)).it
```

src/mo_def/arrange.ml
```diff
@@ -93,6 +93,7 @@ module Make (Cfg : Config) = struct
     | AndE (e1, e2)       -> "AndE"    $$ [exp e1; exp e2]
     | OrE (e1, e2)        -> "OrE"     $$ [exp e1; exp e2]
     | ImpliesE (e1, e2)   -> "ImpliesE"$$ [exp e1; exp e2]
+    | OldE (e)            -> "OldE"    $$ [exp e]
```
ggreif, line 96:
```suggestion
    | OldE e              -> "OldE"    $$ [exp e]
```

src/mo_frontend/assertions.mly
```diff
@@ -45,5 +46,7 @@ let is_verification () =
     { is_verification () &&& AssertE(Postcondition, e) @? at $sloc }
   | ASSERT COLON s=NAT COLON ASYNC e=exp_nest
     { is_verification () &&& AssertE(Concurrency s, e) @? at $sloc }
+  | OLD e=exp_nest
+    { is_verification () &&& OldE(e) @? at $sloc }
```
ggreif, line 50:
```suggestion
    { is_verification () &&& OldE e @? at $sloc }
```

ggreif:
As you see I am not a fan of redundant parentheses :->
feel free to ignore those nits.

ggreif (approved):

Issue #3646, opened 2022-12-13
Incorrect handling of negative zero in floating points


luc-blaeser:
It seems that **negation of zero and `nan`** is not correctly represented in the compiler according to IEEE 754 standard. 
This, for example, leads to the following inconsistencies:

```
let wrongNegativeZero = -0.0;
let correctNegativeZero = Prim.floatCopySign(0.0, -1.0);

Prim.debugPrint(debug_show(wrongNegativeZero) # " vs. " # debug_show(correctNegativeZero));
Prim.debugPrint(debug_show(1.0 / wrongNegativeZero) # " vs. " # debug_show(1.0 / correctNegativeZero));
Prim.debugPrint(debug_show(Prim.floatCopySign(1.0, wrongNegativeZero)) # " vs. " # debug_show(Prim.floatCopySign(1.0, correctNegativeZero)));
Prim.debugPrint(debug_show(Prim.arctan2(0.0, wrongNegativeZero)) # " vs. " # debug_show(Prim.arctan2(0.0, correctNegativeZero)));

```
Output:
```
0.000000 vs. -0.000000
inf vs. -inf
1.000000 vs. -1.000000
0.000000 vs. 3.141593
```

(Please note that `-0.0 == 0.0` according to the specification. However, as the examples above show, `-0.0` needs to be represented with a negative sign because of other operations that may rely on the sign.)

Moreover, negation of `nan` is not correctly handled (on `drun`):

```
let NaN = 0.0/0.0;

let wrongNegativeNaN = -NaN;
let correctNegativeNaN = Prim.floatCopySign(NaN, -1);

Prim.debugPrint(debug_show(wrongNegativeNaN) # " vs. " # debug_show(correctNegativeNaN));
Prim.debugPrint(debug_show(Prim.floatCopySign(1.0, wrongNegativeNaN)) # " vs. " # debug_show(Prim.floatCopySign(1.0, correctNegativeNaN)));
```

Output:
```
nan vs. -nan
1.000000 vs. -1.000000
```

luc-blaeser:
Updated the description, since it is more general, negation of zero and `nan`.

luc-blaeser:
The other issue `https://github.com/dfinity/motoko/issues/3647` is related but different. While the other issue only affects `wasmtime`, this issue is general, also relevant for IC compilation.

PR #3676, opened 2023-01-05
Fix zero negation for floating point numbers

luc-blaeser:
Change the floating point negation to WASM `f64.neg` to correctly support the sign of zero (negative zero, negation of zero).  
Fix for https://github.com/dfinity/motoko/issues/3646.


ggreif (approved):
Good catch! LGTM!

PR #3687, opened 2023-01-11
fix(interpreter): create placeholder functions for actor expressions

### LINKED ISSUE #3683: Interpreter returns an empty Blob for `ActorUrlE`
[Example](https://embed.smartcontracts.org/motoko/g/tRGPBbvb9rfpzMpD93yxK5DBPALTEuhuyTHbT1EajoN7sUPAMxK5ofnA64MbwKrUSC7Xzod52fmNo3mm65muo3xLLvnC3TifwG6tsHhpS?lines=6):
```motoko
let ic = actor "aaaaa-aa" : actor {
  raw_rand : () -> async Blob
}; // => "" : actor {raw_rand : shared () -> async Blob}

ic.raw_rand // Error; resolves to ("" : Blob).raw_rand
```

This currently prevents loading the `mo:base/Random` module in the interpreter. 

Would there be a relatively simple way to fix this? Otherwise, we could temporarily move the `raw_rand` declaration into the `Random.blob()` function to fix the initialization error (https://github.com/dfinity/motoko-base/pull/501). 

I briefly started working on a solution (using the actor type to create an object with placeholder functions) but wanted to check in about this before proceeding further.

rvanasa:
Fixes #3683 (`mo:base/Random` can now be imported in the interpreter). 

Let me know if it would make sense to adjust any of these new error messages to follower a broader convention.

### DIFF
--- a/src/mo_interpreter/interpret.ml
+++ b/src/mo_interpreter/interpret.ml
@@ -430,12 +430,14 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
           List.iter (fun f ->
             match f.T.typ with
             | T.Func (s, c, b, a, r) ->
-              env' := V.Env.add f.T.lab (V.Func (Call_conv.{
-              sort = s;
-              control = c;
-              n_args = List.length a;
-              n_res = List.length r;
-            }, (fun _ _ _ -> trap exp.at "unable to call method %s in actor \"%s\"" f.T.lab (V.as_text v1)))) !env'
+              let conv = Call_conv.{
+                sort = s;
+                control = c;
+                n_args = List.length a;
+                n_res = List.length r;
+              } in
+              let func = fun _ _ _ -> trap exp.at "unable to call method %s in actor \"%s\"" f.T.lab (V.as_text v1) in
+              env' := V.Env.add f.T.lab (V.Func (conv, func)) !env'
             | _ -> trap exp.at "unexpected field type %s for external actor method" (T.string_of_typ f.T.typ)
           ) fs;
           k (V.Obj !env')
--- a/src/mo_interpreter/interpret.ml
+++ b/src/mo_interpreter/interpret.ml
@@ -433,7 +433,7 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
               let func = fun _ _ k ->
                 match (url_text, f.T.lab) with
                 | ("aaaaa-aa", "raw_rand") ->
-                  k (V.Text (String.init 32 (fun _ -> Char.chr (Random.int 256))))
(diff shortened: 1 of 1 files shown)


src/mo_interpreter/interpret.ml
```diff
@@ -497,21 +516,21 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
           | "put" -> array_put
           | "keys" -> array_keys
           | "vals" -> array_vals
-          | _ -> assert false
+          | s -> invalid_arg s
```
crusso, line 519: This and below were assert false because they should just be unreachable code in type correct programs. But the change is probably fine.

crusso (changes requested):
Interestingly, the ir_interpreter actually models actors id properly https://github.com/dfinity/motoko/pull/1116 using a separate environment so that you can cast between principal and actor value as an object.

Eventually, we could do that and just install "aaaaa-a" in the initial environment

I actually did this many moons ago for both interpreters, anticating casts, but Andreas made me take it out at the time. 

Maybe I should revisit that. Note to self: it was part of this PR https://github.com/dfinity/motoko/pull/1027,
caba0bda6e620739e1ed0e5f7e5494c5e2e5e835 was the commit that reverted.

src/mo_interpreter/interpret.ml
```diff
@@ -422,7 +422,26 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
   | ActorUrlE url ->
     interpret_exp env url (fun v1 ->
       match Ic.Url.decode_principal (V.as_text v1) with
-      | Ok bytes -> k (V.Blob bytes)
+      (* create placeholder functions (see #3683) *)
+      | Ok bytes ->
+        (match exp.note.note_typ with
+        | T.Obj (_, fs) ->
+          let env' = ref V.Env.empty in
+          List.iter (fun f ->
```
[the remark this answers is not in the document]
rvanasa: Right, thanks; been writing too much JS lately. Simplified the code and added a fallback for `raw_rand` (building and checking in moc.js right now).

src/mo_interpreter/interpret.ml
rvanasa: This change was useful for finding the original cause of the issue. On the other hand, I suppose it's no longer necessary.

rvanasa:
Including some sort of rand prim seems like the ideal solution. I'll push what I've got so far and defer to what you're working on.

crusso:
work in progress here #3691

crusso (approved):
Pre-approving in the hope that #3693 passes CI and can be merged with this.

crusso (approved):
Really approving

rvanasa:
Superseded by #3693

PR #3691, opened 2023-01-11
feat: implement IC.raw_rand as a prim that we can run in interpreters as well on the IC

crusso:
Adds a new primitive "rand" so we can both support rand in the interpreter(s) and compile to a well-known management canister method for the IC (without adding more interpreter support for canister ids and casting).

Unlike other prims, this is actually a constant, not operator, that is compiled as such in the backend (i.e. statically allocated).

Won't work for WASI backend though, since it's still an async function and those remain verboten.

I'm not too sure about the const stuff, so if @nomeata could take a quick look, I'd feel more reassured.

Similar hacks should let us define infinities for floats as constants, not functions, as @luc-blaeser was hoping for.

NB. Our Motoko AST supports (unapplied) primitive expressions, `S.PrimE s` , but the IR only knows about applications of primitives `IR.PrimE(p, es)`.


src/mo_interpreter/interpret.ml
```diff
@@ -400,6 +400,14 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
   last_env := env;
   Profiler.bump_region exp.at ;
   match exp.it with
+  | PrimE "rand" ->
+    k (Value.async_func T.Write 0 1
+      (fun c v k' ->
+         async env
+           exp.at
+           (fun k'' r ->
+             k'' (V.Blob ""))
```
rvanasa, line 409: I used this code snippet in the other PR:
```suggestion
             k'' (V.Blob (String.init 32 (fun _ -> Char.chr (Random.int 256)))))
```

rvanasa:
Looking good so far! This also seems appealing as a way to hide the implementation details of randomness in Motoko.

crusso:
Interestingly, the assert that two random blobs are different fails on the replica. Investigating if this also fails in playground. That would be pretty terrible....

rvanasa:
> Interestingly, the assert that two random blobs are different fails on the replica.

Whoa, yeah, that could be a problem (especially if it happens intermittently). I tried on both MP and the local replica, and it seems to be working correctly on my end. However, any instance of this happening could be a major issue for generating cryptographic keys, etc. 

> One drawback is that we will need to change Random.mo to use the prim, so you'll need to wait a release for this to work, if that's ok.

This should be alright. If necessary, we could cherry-pick this PR into `moc_interpreter.js`.

crusso:
> > Interestingly, the assert that two random blobs are different fails on the replica.
> 
> Whoa, yeah, that could be a problem (especially if it happens inconsistently). I tried on both MP and the local replica, and it seems to be working correctly on my end. However, any instance of this happening could be a major issue for generating cryptographic keys, etc.

Yeah, I could repro on the playground. Good to know it doesn't happen in dfx. I think it's just a drun issue:

https://dfinity.slack.com/archives/CGZJ7G1J6/p1673461080359549

Thanks for checking.

src/ir_def/check_ir.ml
```diff
@@ -859,6 +859,8 @@ let rec check_exp env (exp:Ir.exp) : unit =
       check e1.note.Note.const "constant DotPrim on non-constant subexpression"
     | PrimE (ProjPrim _, [e1]) ->
       check e1.note.Note.const "constant ProjPrim on non-constant subexpression"
+    | PrimE (OtherPrim "rand", []) ->
+      ()
```
crusso, line 863:
```suggestion
    | PrimE (OtherPrim "rand", []) -> ()
```

nomeata (commented):
I didn't quite get why this can't just be the interpreter knowing about `aaaaa-aa` and synthesizing responses to it?

crusso:
The ir interpreter can do that but the source one can't because Andreas didn't let me add the actor_env mapping ids to implementations. So you'd need to catch the actorlit expression and fake up an IC with random implementation. Would also work, but I wanted to do the experiment of adding a prim constants (need for Float.infinities and possible elsewhere).

https://github.com/dfinity/motoko/pull/3687 (Ryan) started with your approach, but I thought the other was cleaner until I discovered all the const hacks I had to do to keep the same perf ;->

Looking at it now, perhaps #3687 is cleaner (though the ir_interpreter also needs fixing).

nomeata:
I find Ryan's cleaner as well, but no reason not to be pragmatic if needs be

crusso:
closing in favour of #3693

PR #3693, opened 2023-01-12
feat: support ManagementCanister.raw_rand in interpreters

crusso:
* adds `actorenv` to AST interpreter, as per ir_interpreter, modelling actor values as blob that need to be indirected through the heap.
* registers ManagementCanister at blob "" with raw_rand method (only).


src/mo_interpreter/interpret.ml
```diff
@@ -431,10 +431,15 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
             match f.T.typ with
             | T.Func (s, c, b, a, r) ->
               let func = fun _ _ k ->
-                match (url_text, f.T.lab) with
-                | ("aaaaa-aa", "raw_rand") ->
-                  k (V.Blob (String.init 32 (fun _ -> Char.chr (Random.int 256))))
-                | _ ->  trap exp.at "unsupported method %s in actor \"%s\"" f.T.lab (V.as_text v1)
+                match (bytes, f.T.lab) with
+                | ("", "raw_rand") ->
```
rvanasa, line 435: Is it possible for the IC canister to use a different URL from `aaaaa-aa`? Otherwise, we could maybe do something like this:

```suggestion
                match (bytes, url_text, f.T.lab) with
                | ("", "aaaaa-aa", "raw_rand") ->
```

src/mo_interpreter/interpret.ml
```diff
@@ -520,21 +525,21 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
           | "put" -> array_put
           | "keys" -> array_keys
           | "vals" -> array_vals
-          | s -> invalid_arg s
+          | s -> assert false
```
rvanasa, line 528: 👍

rvanasa (approved):

src/mo_interpreter/interpret.ml
crusso: I believe "aaaaa-aa" is hardcoded into the interface spec of the IC.

crusso: https://internetcomputer.org/docs/current/references/ic-interface-spec#ic-management-canister

rvanasa: Okay, good to know. Although it's unlikely for anyone to run into this, it seems reasonable to restrict the `raw_rand` behavior to the management canister's principal (in case someone tries to use an unrelated `raw_rand` implementation from a different canister).

crusso:
Bugger. I'll try reverting the text to blob changes tomorrow in the ir_interpreter....tomorrow

src/mo_interpreter/interpret.ml
```diff
@@ -421,8 +421,36 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
     k (interpret_lit env lit)
   | ActorUrlE url ->
     interpret_exp env url (fun v1 ->
-      match Ic.Url.decode_principal (V.as_text v1) with
-      | Ok bytes -> k (V.Blob bytes)
```
crusso, line 425: OK, no wonder lots of test are failing. This used to return a blob, but is now returning an object, so the tests will fail in different ways.

Fundamentally, the problem is that the frontend interpreter needs to model actor id and represent actors values using actor ids, not objects.

src/mo_interpreter/interpret.ml
```diff
@@ -875,10 +892,19 @@ and match_shared_pat env shared_pat c =
 (* Objects *)
 
 and interpret_obj env obj_sort dec_fields (k : V.value V.cont) =
-  let self = if obj_sort = T.Actor then V.fresh_id() else env.self in
-  let ve_ex, ve_in = declare_dec_fields dec_fields V.Env.empty V.Env.empty in
-  let env' = adjoin_vals { env with self = self } ve_in in
-  interpret_dec_fields env' dec_fields ve_ex k
+  match obj_sort with
+  | T.Actor ->
+     let self = if obj_sort = T.Actor then V.fresh_id() else env.self in
```
nomeata, line 897: Checking `obj_sort` twice?

src/mo_interpreter/interpret.ml
```diff
@@ -995,12 +1021,33 @@ and interpret_func env name shared_pat pat f c v (k : V.value V.cont) =
 
 (* Programs *)
 
+let ensure_management_canister env =
+  if V.Env.mem "" (!(env.actor_env))
+  then ()
+  else
+    env.actor_env :=
+      V.Env.add
+        (* ManagementCanister with raw_rand (only) *)
```
nomeata, line 1030: Can't you just initialize the actor env with the management canister preinstalled, instead adding it later? (If no, maybe leave a comment somewhere nearby?)

Also, maybe move the `V.Obj` literal into its own top-level definition?

crusso: Yeah, I tried, but its awkward without losing the logging and access to other async helpers that Andreas provided.

For the ir interpreter, the initial env actually depends on the flavour so has to be done late.

This was of a rabbit hole than I had intended, though I guess it fixes other infelicities in the interpreter.

src/mo_interpreter/interpret.ml
crusso: Dooh

crusso, line 897:
```suggestion
     let self = V.fresh_id() in
```

crusso:
Ok, this works but was bit uglier than expected.

One slight change in observable behaviour is that the repl will now display actor values as blobs, not records of opaque function values.  Can we live with that?

Also, we've added more top-level state to the compiler (the actor environment) which might need saving and restoring in future.

All said, I'm kinda sorry I didn't stick with the add a new prim approach after all.

src/mo_interpreter/interpret.ml
```diff
@@ -17,6 +17,9 @@ type lib_env = V.value V.Env.t
 type lab_env = V.value V.cont V.Env.t
 type ret_env = V.value V.cont option
 type throw_env = V.value V.cont option
+type actor_env = V.value V.Env.t ref (* indexed by actor ids *)
+
+let state = ref V.Env.empty
```
crusso, line 22:
```suggestion
(* The actor heap. 
    NB: A cut-down ManagementCanister with id "" is added later, to enjoy access to logging facilities. *)
let state = ref V.Env.empty
```

nomeata:
> Also, we've added more top-level state to the compiler (the actor environment) which might need saving and restoring in future.

But why is it global? Isn’t there an `env` that is initialized once and passed around, where the actorenv can live?

> All said, I'm kinda sorry I didn't stick with the add a new prim approach after all.

It might pay off with the next feature of `aaaaa-aa` that we want to support in the interpreter…

crusso:
> > Also, we've added more top-level state to the compiler (the actor environment) which might need saving and restoring in future.
> 
> But why is it global? Isn’t there an `env` that is initialized once and passed around, where the actorenv can live?
> 
> > All said, I'm kinda sorry I didn't stick with the add a new prim approach after all.
> 
> It might pay off with the next feature of `aaaaa-aa` that we want to support in the interpreter…

It's passed in and accessed via the env, but  not the 'scope' yet, so needs to bet kept in a global to get the repl to work (unlike the  interpreters which aren't used in the repl or don't work with it IIRC.

I can fix it later but am about to go off for a week's holiday.

Weird CI failed on Ubuntu but succeeded on Linux. Kicked it off again.

PR #3695, opened 2023-01-12
fix: timer tree invariant was broken

ggreif:
The expired timer node must be expunged before calling `prune`, otherwise the same Id ends up twice in the tree. This would break `@cancelTimer`'s invariant.

### DIFF
--- a/src/prelude/internals.mo
+++ b/src/prelude/internals.mo
@@ -503,7 +503,8 @@ func @timer_helper() : async () {
           case (?delay) if (delay != 0) {
             // re-add the node
             let expire = n.expire[0] + delay;
-            // N.B. insert only works on pruned nodes
+            n.expire[0] := 0;
+            // N.B. reinsert only works on pruned nodes
             func reinsert(m : ?@Node) : @Node = switch m {
               case null ({ n with expire = [var expire]; pre = null; post = null });
               case (?m) {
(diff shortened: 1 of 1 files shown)


ggreif:
@crusso re-introducing the bug would reverse the output according https://github.com/dfinity/motoko/pull/3695/commits/2aa30f3e5cf459936d1feb6eeb574c387eaf7776

src/prelude/internals.mo
```diff
@@ -503,7 +503,8 @@ func @timer_helper() : async () {
           case (?delay) if (delay != 0) {
             // re-add the node
             let expire = n.expire[0] + delay;
-            // N.B. insert only works on pruned nodes
+            n.expire[0] := 0;
```
crusso, line 506:
```suggestion
            n.expire[0] := 0;
```
Is that so it gets pruned?

crusso (approved):
LGTM

src/prelude/internals.mo
ggreif: Yes. But why the suggestion? There seems to be no change...

luc-blaeser (approved):
Also looks good to me.

src/prelude/internals.mo
crusso: Oh sorry, I was just asking a question. No change intended.

PR #3697, opened 2023-01-13
feat: throw on send failure, compatibility flag and examples

crusso:
Resolve the outstanding issue with #3630 by just adding a flag `--trap-on-call-error` to enable backwards compatibility, relying on users to read the release notes.
Assume the release including this will be 0.8.0.

I'll eventually add an example to the changelog, adapted from one of the tests.

TIL that we didn't even trap on oneway failure. Really fire and forget, I guess.


doc/md/compiler-ref.md
```diff
@@ -49,7 +49,11 @@ You can use the following options with the `moc` command.
 | `--release`                               | Ignores debug expressions in the source.                                                                                                              |
 | `--stable-types`                          | Compile binary and emit signature of stable types to `.most` file.                                                                                    |
 | `--stable-compatible <pre> <post>`        | Test upgrade compatibility between stable-type signatures `<pre>` and `<post>`.                                                                       |
-| `-t`                                      | Activates tracing in interpreter.                                                                                                                     |
-| `-v`                                      | Generates verbose output.                                                                                                                             |
+| `--trap-on-call-error                     | Trap, don't throw an `Error`, when an IC call fails due to destination queue full or freezing threshold is crossed.
```
crusso, line 52:
```suggestion
| `--trap-on-call-error`                     | Trap, don't throw an `Error`, when an IC call fails due to destination queue full or freezing threshold is crossed.
```

doc/md/compiler-ref.md
```diff
@@ -49,7 +49,11 @@ You can use the following options with the `moc` command.
...
 | `--stable-types`                          | Compile binary and emit signature of stable types to `.most` file.                                                                                    |
 | `--stable-compatible <pre> <post>`        | Test upgrade compatibility between stable-type signatures `<pre>` and `<post>`.                                                                       |
-| `-t`                                      | Activates tracing in interpreter.                                                                                                                     |
-| `-v`                                      | Generates verbose output.                                                                                                                             |
+| `--trap-on-call-error`                     | Trap, don't throw an `Error`, when an IC call fails due to destination queue full or freezing threshold is crossed.
+|                                           | Emulates behaviour of moc versions < 0.8.0.
```
crusso, line 53:
```suggestion
|                                           | Emulates behaviour of moc versions < 0.8.0.
|
```

test/run-drun/send-failure-example-trap.mo
```diff
@@ -0,0 +1,38 @@
...
+    Prim.debugPrint(debug_show {count});
+    return count;
+  }
+
+}
+//SKIP ic-ref-run
+//CALL ingress repeat 0x4449444C0000
+//CALL ingress get 0x4449444C0000
+//CALL ingress repeatUntil 0x4449444C0000
+//CALL ingress get 0x4449444C0000
```
crusso, line 38:
```suggestion
//CALL ingress get 0x4449444C0000

```

test/run-drun/send-failure-example.mo
```diff
@@ -0,0 +1,37 @@
...
+    Prim.debugPrint(debug_show {count});
+    return count;
+  }
+
+}
+//SKIP ic-ref-run
+//CALL ingress repeat 0x4449444C0000
+//CALL ingress get 0x4449444C0000
+//CALL ingress repeatUntil 0x4449444C0000
+//CALL ingress get 0x4449444C0000
```
crusso, line 37:
```suggestion
//CALL ingress get 0x4449444C0000

```

src/codegen/compile.ml
```diff
@@ -7241,6 +7241,9 @@ module FuncDec = struct
         G.nop
       end
       begin
+        if !Flags.trap_on_call_error then
+          E.trap_with env (Printf.sprintf "could not perform %s" purpose)
```
crusso, line 7245:
```suggestion
          E.trap_with env (Printf.sprintf "could not perform %s" purpose)
```
Hmm, maybe it would be nice if we could preserve the message for throw errors too....

PR #3717, opened 2023-01-22
chore: bump to `dune_3`

ggreif:


nomeata (approved):

default.nix
```diff
@@ -775,7 +775,7 @@ rec {
     sha256 = "sha256-debC8ZpbIjgpEeISCISU0EVySJvf+WsUkUaLuJ526wA=";
   };
 
-  shell = stdenv.mkDerivation {
+  shell = nixpkgs.mkShell {
```
ggreif, line 778: @nomeata this was in my worktree (I've adopted it from a reddit post) and snuck in with `-a`. I guess I'll leave it in as it conveys the intent better.

default.nix
```diff
@@ -806,7 +806,7 @@ rec {
           nixpkgs.rlwrap # for `rlwrap moc`
           nixpkgs.difftastic
           nixpkgs.openjdk nixpkgs.z3 nixpkgs.jq # for viper dev
-        ]
+        ] ++ nixpkgs.lib.optional stdenv.isDarwin nixpkgs.darwin.apple_sdk.frameworks.Security
```
ggreif, line 809: This was in my checkout too. It helps linking `ic-repl` on the Mac, so it is actually useful. But I can back it out if you think it doesn't belong here.

PR #3719, opened 2023-01-23
fix: bump required `dune` and introduce `dune-workspace`

ggreif:
This factors out the `js_of_ocaml` flags into a central location and enables better control
(also fixing a problem when linking under `dune-3.2`).

@crusso has seen the problem on `master` when doing
```
make -C src moc_interpreter.js
```
that compilation flags were passed to the linker of jsoo:
```
js_of_ocaml: unknown option `--no-extern-fs'.
             unknown option `--no-source-map'.
             unknown option `--target-env'.
Usage: js_of_ocaml link [OPTION]... [JS_FILES]...
```


ggreif:
@rvanasa with some luck this may even reduce the `.js` sizes further, as now more files are compiled with frugal flags.

crusso (approved):
LGTM - why do the files includes need to change?

nomeata (approved):
Rubber stamp, looks plausible

src/exes/candid_tests.ml
```diff
@@ -42,14 +42,6 @@ let argspec = Arg.align
 
 module Pretty = Type.MakePretty(Type.ElideStamps)
 
-let load_file f =
```
ggreif, line 45: this binding was unused

ggreif:
> LGTM - why do the files includes need to change?

Probably because the filenames are now workspace-relative. I dunno exactly. Maybe before each directory was its own (implicit) workspace.

rvanasa (approved):
Looks good; thanks for the solution.

crusso:
Just checked it out locally and can verify that make moc_interpreter.js (and make preview in doc) seem to work and create working  js.

PR #3722, opened 2023-01-24
fix(mo-doc): add constructors to Markdown output

### LINKED ISSUE #3721: bug: [mo-doc] the md output doesn't document a class constructor as a function, just the class itself.
For an example, see Random.mo

image

There should probably be an entry for the constructor *function* Finite(b : Blob) -> Finite.

rvanasa:
Resolves #3721 by adjusting the output of class members to use the same format as functions, types, and values. 

Since there's a lot of noise in this PR from other base library changes, here's an example of the new output: [`Random.Finite()`](https://github.com/dfinity/motoko/pull/3722/files#diff-0e47a97d3a5d33f1c8aa4ec08b3d79551388f614fc34d0ca958bf19eee0bc5ccR42-R46)

### DIFF
--- a/doc/md/base/Array.md
+++ b/doc/md/base/Array.md
@@ -480,3 +480,23 @@ sum
 Runtime: O(1)
 
 Space: O(1)
+
+## Function `size`
+``` motoko no-repl
+func size<X>(array : [X]) : Nat
+```
+
+Returns the size of `array`.
+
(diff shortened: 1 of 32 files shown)


crusso (approved):
Looks great! Thanks for fixing this so quickly!

PR #3727, opened 2023-01-25
chore: minor revisions to #3630 (throw on send failure)

crusso:
Follow ups to #3630


crusso:
@ggreif I'm seeing non-det behaviour from cancel-timer.mo. Maybe just suppress the output?

ggreif:
> @ggreif I'm seeing non-det behaviour from cancel-timer.mo. Maybe just suppress the output?

Yeah, `cancel-timer.mo` needs another bump to the initial timer. I have seen the test failing too. But now it is rather rare.

ggreif (approved):
Thanks, LGTM!

PR #3734, opened 2023-01-26
chore: make timer tests more deterministic

crusso:
This removes the textual outputs and relies on checking the globals instead.


ggreif (approved):
Thanks!

crusso:
> Thanks!

I must admit, I was a little perplexed by what this is doing... can you recall?
```
public shared func go() : async () {
     var attempts = 0;
     while (counter < max) {
       ignore await get();
       ignore await raw_rand(); // yield to scheduler
       if (counter == 3) attempts += 1;
       if (attempts >= 200)
         counter += 1;
     };
     debugPrint(debug_show {counter});
  };

```

ggreif:
> I must admit, I was a little perplexed by what this is doing... can you recall? @ggreif

It sets up a single-expiration timer, which in its job spawns a recurrent time that on its 3rd expiration cancels itself.

Then the `go` routine waits a few rounds to be sure that it is dead for good and exits.

PR #3741, opened 2023-01-26
bugfix:  for #3740

### LINKED ISSUE #3740: bug: async.ml transformation produces bad code.
https://forum.dfinity.org/t/type-error-m0139-inner-actor-classes-are-not-supported-yet/18135/12

Fixed by #3741.

crusso:
Fixes #3740.

Motoko uses the syntactic tuple arity of a message return type to determine the message return arity.
```
actor {
  type Unit = ();

  public func f() : async Unit {
  };

  public func go() : async () {
    await f();
  };

}
```

This means that `f` and `go` have different signatures:
* `f` returns 1 result that is a 0-tuple.
* `g` returns 0 results.

Getting this to work correctly requires some additional impedance matching in the async lowering phase to deal with functions like `f`, which used to crash `moc` (but now don't).


src/ir_passes/async.ml
```diff
@@ -93,10 +101,10 @@ let new_nary_async_reply ts =
     let vs, seq_of_vs =
       match ts with
       | [t1] ->
-        let v = fresh_var "rep" t1 in
+        let v = fresh_var "rep1" t1 in
```
crusso, line 104:
```suggestion
        let v = fresh_var "rep" t1 in
```

src/ir_passes/async.ml
```diff
@@ -96,7 +104,7 @@ let new_nary_async_reply ts =
         let v = fresh_var "rep" t1 in
         [v], varE v
       | ts1 ->
-        let vs = fresh_vars "rep" ts1 in
+        let vs = fresh_vars "rep2" ts1 in
```
crusso, line 107:
```suggestion
        let vs = fresh_vars "rep" ts1 in
```

test/run-drun/unit-messages.mo
```diff
@@ -0,0 +1,42 @@
...
+
+    let () = await async (() : ());
+    let () = await async (() : Unit);
+    let () : Unit = await async (() : ());
+    let () : Unit = await async (() : Unit);
+
+};
+
+};
+await a.go(); //OR-CALL ingress go "DIDL\x00\x00"
```
crusso, line 42:
```suggestion
await a.go(); //OR-CALL ingress go "DIDL\x00\x00"

```

src/ir_def/check_ir.ml
```diff
@@ -364,7 +364,15 @@ let store_typ t  =
 let rec check_exp env (exp:Ir.exp) : unit =
   (* helpers *)
   let check p = check env exp.at p in
-  let (<:) t1 t2 = check_sub env exp.at t1 t2 in
+  let (<:) t1 t2 =
+(*  try *)
```
ggreif, line 368: remove debug code?

ggreif (commented):
I don't grok the fix yet. For me this is a problem of being a syntactic tuple in the return type or not. In the latter case it is a unary return, in the former the return arity is given by the syntactic arity of the tuple.

src/ir_def/check_ir.ml
crusso: Well, it a pain to reinvent this everytime I need to track down a bug so I'd prefer to just comment it out...

crusso:
> I don't grok the fix yet. For me this is a problem of being a syntactic tuple in the return type or not. In the latter case it is a unary return, in the former the return arity is given by the syntactic arity of the tuple.

Yeah, I find it hard to get my head around this too and am not 100% confident about the fix though it seems to work.

The problem is that we lose the syntactic distinction  once we are in the IR and need to do some additional impedance matching.
I think fundamentally the problem is that `T.seq (T.as_seq (T.seq [()])) == [] <> [()]' and that is biting us here.

nomeata (commented):
I'd have expected a fix to be touching the desugaring, but not the IR - at the level of the IR I'd have hoped that no more syntactic disambiguaties remain.

nomeata:
Would the problem only happen with first class unary tuple `Unit`, or other first class tuples as well?

crusso:
> Would the problem only happen with first class unary tuple `Unit`, or other first class tuples as well?

Oh crap, good point. I'll test tomorrow.

crusso:
> Would the problem only happen with first class unary tuple `Unit`, or other first class tuples as well?



> I'd have expected a fix to be touching the desugaring, but not the IR - at the level of the IR I'd have hoped that no more syntactic disambiguaties remain.

Yeah, I did look there but wasn't able to spot the problem. I'm not super happy about this fix, but it was the only one I could make stick. I'll have another go tomorrow, after checking the tuple thing.

crusso:
Ok, I've extended the fix to pairs and other tuples, including the tests. Although this might not be the right fix, it might be worth getting this in until I figure out what it is. Suggestions welcome ;->

PTAL @nomeata @ggreif

nomeata:
Is that something we can check in the IR checkers?

crusso:
> Is that something we can check in the IR checkers?

I'm not sure what to check, but the IR checker caught the bug in the first place (good thing we didn't turn it off by default)
I wonder if the ultimate problem is that async types don't have an n-ary payload at the moment, so we are losing too much info in the desugaring (or some transformation). On the other hand, I don't want to make them n-ary so we can later support some abstraction over them. Will mull it over...

crusso:
@ggreif Just got another non-det failure in timer.mo... (ic-ref)

```
test-ic-ref-run> Some tests failed:
test-ic-ref-run> timer.mo
test-ic-ref-run> make: *** [Makefile:4: all] Error 1
test-ic-ref-run> make: Leaving directory '/private/tmp/nix-build-test-ic-ref-run.drv-0/test-run-drun-src/run-drun'
error: builder for '/nix/store/8q4rr1s0wh75v34gv9msppdizswzhn8v-test-ic-ref-run.drv' failed with exit code 2;
       last 10 log lines:
       > tailpositions-actor: [actor1.ic-ref-run.comp] [actor2.ic-ref-run.comp] [ic-ref-run]skipped (no drun)
       >
       > upgrade-hooks: [upgrade-hook.ic-ref-run.comp] [upgrade-hook0.ic-ref-run.comp] [upgrade-hook1.ic-ref-run.comp] [upgrade-hook2.ic-ref-run.comp] [ic-ref-run]skipped (no drun)
       >
       > upgrades: [upgrade.ic-ref-run.comp] [upgrade0.ic-ref-run.comp] [upgrade1.ic-ref-run.comp] [upgrade2.ic-ref-run.comp] [upgrade3.ic-ref-run.comp] [upgrade4.ic-ref-run.comp] [ic-ref-run]skipped (no drun)
       >
       > Some tests failed:
       > timer.mo
       > make: *** [Makefile:4: all] Error 1
       > make: Leaving directory '/private/tmp/nix-build-test-ic-ref-run.drv-0/test-run-drun-src/run-drun'
       For full logs, run 'nix log /nix/store/8q4rr1s0wh75v34gv9msppdizswzhn8v-test-ic-ref-run.drv'.
error: 
nix-build-uncached: nix build failed: exit status 1
```

https://github.com/dfinity/motoko/actions/runs/4053633247/jobs/6974477511

nomeata:
> Just got another non-det failure in timer.mo... (ic-ref)

Looks like making an originally intentionally deterministic tool non-deterministic because of time-based concurrency bites…

src/ir_passes/async.ml
```diff
@@ -80,7 +80,21 @@ let new_nary_async_reply ts =
...
+      | T.Tup ts1 ->
+        let k' = fresh_var "k" (contT t1 T.unit) in
+        let r' = fresh_var "r" (err_contT T.unit) in
+        [k';r'] -->* (
+          varE unary_async -*-
+            (tupE [
+               ([v'] -->* (varE k' -*- varE v'));
+               varE r'
+             ])
+        )
```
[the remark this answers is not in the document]
ggreif: Maybe you have uncommitted changes in your repo?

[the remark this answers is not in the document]
crusso: Ah, sorry, I was looking at lines 99-103 of the original, not current.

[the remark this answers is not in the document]
crusso: PTAL

crusso:
> LGTM! will you write a follow-up issue?

I'll keep the original one open.

PR #3744, opened 2023-01-27
chore: regenerate `motoko-base` docs

ggreif:


crusso (approved):
But I think we can just do this prior to release...

PR #3745, opened 2023-01-27
Releasing 0.8.0

ggreif:


crusso (approved):
Your probably want to add the ## 0.8.0 (2023-01-27 header too

PR #3748, opened 2023-01-27
Exclude updating nix hash for dependabot

kentosugama:
Restrict _dependabot_ from running on actions that need access token. The reason is that PRs initiated by dependabot don't have access to the `NIV_UPDATER_TOKEN` and thus cannot modify the repository, causing every build to fail twice:
- in the `update-hash` workflow
- in the performance diffing step of the `build` workflow when changing a PR comment.

This PR suppresses those mutations and as such the automatisms can bring the PR to completion.

For reference, a [recent failed CI run's log](https://github.com/dfinity/motoko/actions/runs/4025099722/jobs/6918617978).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,7 +39,7 @@ jobs:
       run: nix-build-uncached --max-jobs 4 -A all-systems-go -build-flags -L
 
     - name: Calculate performance delta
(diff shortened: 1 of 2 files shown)


nomeata (approved):
Rubber stamp

ggreif (approved):
Nice work, thanks!

PR #3753, opened 2023-01-30
chore: adjust test rand-JIRA-EXC-1318 to fixed drun

crusso:


ggreif (commented):
Should `ic-ref-run` be enabled now?

PR #3754, opened 2023-01-30
chore: fix make base target

crusso:
I notice that doc/make base tried to produce docs for the motoko-base test dir too, and complained.

Fixed the base target to only produce doc for motoko-base src dir, not the root dir.

I wonder how this ever worked in the past, especially since it used by our release instructions...


crusso:
Probably a red-herring. I redefined MOTOKO_BASE incorrectly.

PR #3758, opened 2023-02-01
bugfix: fix for quadratic import of actor classes

### LINKED ISSUE #3769: test: add a regression test for the bug fixed in #3758 (code explosion due to unused, imported actor classes)
It would be nice to a have a FileCheck like regression test to verify the fix #3758, by perhaps counting the number of embedded wasms  matches expectations in a project involving actor classes ( counting `candid:service` custom sections might suffice).

crusso:
The introduction of system class constructors had the unfortunate side-effect of making actor class libraries impure (due to a partial application at top-level of the 'system' class constructor to define the default class constructor). This meant that the compiler was no longer dead-coding them leading to quadratic behaviour in code generation: each actor class would include the code of all preceding actor classes, even if not actually used.

This PR make the translation of actor classes pure again, by defining the default constructor as a delayed application of the system constructor, allowing unused actor classes to be dead-code eliminated by the back-end.

In particular, reduces the code size of the large actor on the forum with six imported actor classes from 11MB to 3MB, 2.2MB after wasm shrink and 600K after gzip.

Fixes the bug reported here:
https://forum.dfinity.org/t/motoko-wasm-repeated-candid-declarations/18306

The original regression was introduced here:
https://github.com/dfinity/motoko/pull/3386

 * [x] the delay introduces an additional async - lets remove that using async*/await* and a readable helper in `internals.mo`
 * [x] the interpreted code still needs adjusting (see FIXME).

Closes #3769


src/lowering/desugar.ml
```diff
@@ -911,19 +911,12 @@ and to_args typ po p : Ir.arg list * (Ir.exp -> Ir.exp) * T.control * T.typ list
 
 type import_declaration = Ir.dec list
 
-let actor_class_mod_exp id class_typ install =
-  let install_typ = install.note.Note.typ in
+let actor_class_mod_exp id class_typ default system =
   let class_con = Cons.fresh id (T.Def([], class_typ)) in
-  let install_var = fresh_var id install_typ in
-  let install_new =
-    (varE install_var) -*- (tagE "new" (recordE ["settings", nullE()]))
```
crusso, line 919: This was the impure top level application preventing the imported actor class from being dead-coded, even if never instantiated.

src/prelude/internals.mo
```diff
@@ -397,38 +397,62 @@ let @ic00 = actor "aaaaa-aa" :
...
-};
-
-func @ic00_install_code() : shared {
-    mode : { #install; #reinstall; #upgrade };
-    canister_id : Principal;
-    wasm_module : @ManagementCanister.wasm_module;
-    arg : Blob;
-  } -> async () {
-  @ic00.install_code
+func @install_actor_helper(
```
crusso, line 400: The core of the installation code, expressed as a more readable helper rather than hand-constructed IR (now that we can use async*/await* for decent abstraction)

src/lowering/desugar.ml
```diff
@@ -946,105 +939,41 @@ let import_compiled_class (lib : S.comp_unit) wasm : import_declaration =
...
   let c', _ = T.as_con (List.hd cs') in
-  let available = fresh_var "available" T.nat in
-  let accepted = fresh_var "accepted" T.nat in
-  let cycles = var "@cycles" (T.Mut (T.nat)) in
-  let install_arg =
-    fresh_var "install_arg" T.install_arg_typ
+  let install_actor_helper = var "@install_actor_helper"
+    (T.Func (T.Local, T.Returns, [T.scope_bind],
+      [T.install_arg_typ; T.blob; T.blob],
+      [T.Async(T.Cmp, T.Var (T.default_scope_var, 0), T.principal)]))
```
ggreif, line 945:
```suggestion
    T.(Func (Local, Returns, [scope_bind],
      [install_arg_typ; blob; blob],
      [Async(Cmp, Var (default_scope_var, 0), principal)]))
```

ggreif (approved):
Smaller, faster, more pure and readable. I like it. Just a nit in the comments.
Also, would it be feasible to add (enable) a `FileCheck` (regression) test to ensure that there is no duplicate embedded actor class inside? (Can be another PR).

crusso:
> Smaller, faster, more pure and readable. I like it. Just a nit in the comments. Also, would it be feasible to add (enable) a `FileCheck` (regression) test to ensure that there is no duplicate embedded actor class inside? (Can be another PR).

Yeah, was I wondering about how to do that. Maybe just counting the occurrences of `candid:servicetype` would do it, if FileCheck  supports that. I'll open an issue for the regression - let's merge this now since people are waiting for a fix.

crusso:
@ggreif PTAL - I added a filecheck to count the expected __motoko_async_helper exports (and no more) using filecheck, abusing an existing test with multiple imports. I also manually verified it fails on master (the exe also shrinks by 2 MB, btw).

ggreif (approved):

PR #3765, opened 2023-02-02
Performance Improvement: Faster Allocation

luc-blaeser:
# Faster allocation

Removing the `ALLOCATIONS` statistics counter, as it can be derived from `get_heap_size()` and `RECLAIMED`. 

## Reasoning
- `get_heap_size()`: Current amount of dynamically allocated heap space.
- `RECLAIMED`: Total amount of memory reclaimed by the GC (originates from the dynamic heap).
- `ALLOCATIONS`: Total amount of dynamically allocated heap space.

Therefore: `ALLOCATIONS == get_heap_size() + RECLAIMED`

## Performance Improvement

**19% faster allocations**

1.6% faster on GC benchmark (copying GC).

### Micro-Benchmark

Instantiation of a class object with a Nat field, instructions:

| Unoptimized   | Optimized | Speedup      |
|---------------|-----------|------------- |
| 112           | 90        | 1.24         |     

### GC Benchmark

GC benchmark, total instructions, average across all benchmark cases:

| GC           | Unoptimized  | Optimized   | Speedup      |
|------------- | ------------ | ----------- | ------------ |
| Generational | 2.10e10      | 2.07e10     | 1.0145       |
| Copying      | 2.45e10      | 2.41e10     | 1.0166       |
| Compacting   | 2.66e10      | 2.63e10     | 1.0114       |

(The copying GC profits more because it performs more allocations.)


test/bench/ok/heap-32.drun-run.ok
```diff
@@ -1,5 +1,5 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
-debug.print: (50_227, +91_377_860, 1_446_975_760)
-debug.print: (50_070, +102_586_000, 1_520_068_314)
+debug.print: (50_227, +91_377_860, 1_437_185_617)
+debug.print: (50_070, +102_586_000, 1_509_442_237)
```
nomeata, line 4: Neat idea! But it would feel safer if the numbers wouldn’t change :-). Should they change?

ggreif: They should. Less cycles are used for allocation (closures).

ggreif (approved):
Cool!

test/bench/ok/heap-32.drun-run.ok
nomeata: Oh, that’s cycles! Then all is well.

crusso:
Nice!

PR #3767, opened 2023-02-02
doc: fix filename reference

### LINKED ISSUE #3766: timers
In https://github.com/dfinity/motoko/blob/master/doc/md/timers.md the document refers incorrectly to timers.mo ... it should be timer.mo

ggreif:
Fixes #3766.


ggreif:
@crusso should we link https://github.com/dfinity/motoko-base/blob/master/src/Timer.mo?

doc/md/timers.md
```diff
@@ -12,7 +12,7 @@ The underlying mechanism is a [canister global timer](https://internetcomputer.o
 The timer mechanism can be disabled completely by passing the `-no-timer` flag to `moc`.
 
 When lower-level access to the canister global timer is desired, an actor can elect to receive timer expiry messages by declaring a `system` function, named `timer`. The function takes one argument (to re-set the global timer), and returns a future of unit type (`async ()`).
-If the `timer` system method is declared, the `Timers.mo` base library module may not function correctly and should not be used.
+If the `timer` system method is declared, the `Timer.mo` base library module may not function correctly and should not be used.
```
ggreif, line 15:
```suggestion
If the `timer` system method is declared, the [`Timer.mo`](https://github.com/dfinity/motoko-base/blob/master/src/Timer.mo) base library module may not function correctly and should not be used.
```

crusso: Why not just reference the local md file, `base/Timer.md`?

crusso: I'll approve anyway but you can  decide which to point to.

crusso (approved):

doc/md/timers.md
ggreif: Happy to do that if it works...
```suggestion
If the `timer` system method is declared, the [`Timer.mo`](base/Timer.md) base library module may not function correctly and should not be used.
```
Like this?

ggreif: Sorry, but I am super ignorant about how the docusaurus stuff is cohering.

crusso: You can test it easily. In nix-shell do `cd doc; make preview`.  With luck that'll open a browser window with the doc.

crusso: I just tested it, works fine. There's another reference that could be a link at the top. 
Want me to commit my changes to this branch?

crusso: I just committed them

PR #3779, opened 2023-02-07
bugfix: locate heap out bounds error in destabilization (canister_postupgrade)

crusso:
Likely culprit behind https://forum.dfinity.org/t/heap-out-of-bounds-use-motoko/10497

TL/DR our deserialization code use the C stack for temporary allocations, but omits a bounds check when growing the stack.

When the stack overflows due to deep recursion, the code tries to write to the last page of wasm memory, causing a heap out of bounds error in the replica.

This PR:
[ ] instruments the code with debug_prints for debugging (TBD).
[ ] implements an inefficient bound check.
[ ] allocates more pages to C stack.

Repro test to come (#3781)

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -1152,7 +1152,8 @@ module Stack = struct
      All pointers here are unskewed.
   *)
 
-  let end_ = Int32.mul 2l page_size (* 128k of stack *)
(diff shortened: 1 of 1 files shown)


crusso:
closing in favour of fix #3782 and repro #3781

PR #3781, opened 2023-02-07
test: repro for heap-out-of-bounds-failure on destabilization 

crusso:
Adds a repro for the heap-out-of-bounds failure on upgrade.
Fix is in #3782


crusso:
incorporated in #3793

Issue #2883, opened 2021-11-04
Stack allocations in generated code does not check bounds


osa1:
In the generated code we currently allocate on C/Rust stack in a few places, mainly for serialization. Stack allocation routines are implemented [here](https://github.com/dfinity/motoko/blob/20dd16533170a0e94bb06f1682c439ab15599fbb/src/codegen/compile.ml#L962-L1004). Currently the allocation routine does not check for stack overflows when bumping the stack pointer. While in the call sites size arguments are small (at most 8 bytes), we don't statically guarantee that the generated code won't be pushing too many values and overflow the stack. For example: https://github.com/dfinity/motoko/blob/20dd16533170a0e94bb06f1682c439ab15599fbb/src/codegen/compile.ml#L3860-L3861

Here `store` argument can generate pushes to the stack.

Even if we can prove today that there will be at most a few pushes and it's not possible to overflow 64 KiB stack, without static or dynamic checks this invariant won't survive refactoring, or can start failing with new features. Stack size can change too.

It could be helpful to implement stack overflow checks, at least in debug mode (with `--sanity-checks`).

nomeata:
I think the reasoning was that if we overflow the stack, because the stack grows _down_, it’ll wrap around and we’ll be accessing data at the far end of the heap, which will trap. But maybe that’s too risky (maybe we did actually allocate all 4GB? Maybe the stack grows very far?)

Does the rust code have stack overflow checks?

rossberg:
> maybe we did actually allocate all 4GB?

We could of course avoid that and set the limit to 4GiB - 1 page, so that the wrap-around trap is guaranteed to occur. And if there ever was code that could grow the stack by more than 64 KiB at a time then it needs to check explicitly or probe in 64KiB steps.

nomeata:
Right, that would mean a check only happening for dynamically sized allocations, and we can continue to omit it when the size is known to be small. No strong opinion here.

PR #3782, opened 2023-02-08
bugfix: fix heap out bounds error during destabilization

crusso:
Builds on https://github.com/dfinity/motoko/pull/3781, which adds a repro for a compiler crash on destabilization of deeply nested data.

This PR fixes that bug which is the likely culprit behind https://forum.dfinity.org/t/heap-out-of-bounds-use-motoko/10497.

TL/DR our deserialization code use the C stack for temporary allocations, but omits a bounds check when growing the stack.

The C stack lives in the first few pages of Wasm memory and grows downwards, previously relying on out of (high) memory read or writes to (unreliably)* detect stack overflow (i.e. stack ptr underflow).

When the stack overflows due to deep recursion, the code tries to write to the last page of wasm memory, causing a "heap out of bounds error" in the replica - an attempt to read from an unallocated wasm page.

This PR:

- [x] implements the missing bound check.
- [x] allocates more pages to C stack, (was 2 pages (128KiB), now 32 pages (2MiB)

Note, this is really a safer mitigation, not a fix per se. Some stable variable data will still cause the stack to overflow, but the failure will be safe, not potentially corrupt memory*. Reserving more stack space should make the failures rarer, if not impossible.

*memory corruption can happen if the last wasm page is already allocated and in use before the unchecked underflow, causing heap corruption. Rare, but possible.

Related issue: https://github.com/dfinity/motoko/issues/2883 due to @osa1 
That issue also suggests a more performant stack guard, at the cost of never allocating the last page. Not sure its worth it though.


src/codegen/compile.ml
```diff
@@ -1164,8 +1168,14 @@ module Stack = struct
   let set_stack_ptr env =
     G.i (GlobalSet (nr (E.get_global env "__stack_pointer")))
 
-  (* TODO: check for overflow *)
   let alloc_words env n =
+    (* first, check for stack underflow *)
+    (* optimize me *)
```
ggreif, line 1173: Maybe put this behind the `--check` flag.

ggreif: #3090 also suggests a stack guard that never allocates the topmost page.

src/codegen/compile.ml
```diff
@@ -1150,9 +1150,13 @@ module Stack = struct
      We sometimes use the stack space if we need small amounts of scratch space.
 
      All pointers here are unskewed.
+
+     (We report logical stack overflow as "RTS Stack underflow" as the stack
+     grows downwards.)
   *)
 
-  let end_ = Int32.mul 2l page_size (* 128k of stack *)
+  let stack_pages = 32l (* 2MiB of stack *)
```
ggreif, line 1158: If we can afford 32 pages for the C stack, then we can surely afford to leave the topmost page unallocated and serve as a guard page.

src/codegen/compile.ml
ggreif: Also only do the check if `n` is so big that it would step over the topmost guard page.

src/codegen/compile.ml
crusso: Yes I agree.  The 2M|B is arbirtrary of course, but 128k was clearly to small for real world use. The other option would be to not use a fixed size stack but that would be big change.

src/codegen/compile.ml
crusso: > Maybe put this behind the `--check` flag.

If we have some other reliable guard in place yes, but not until then.

luc-blaeser (approved):
Great bug analysis. And good fix (safety first).
It is certainly interesting to see what the performance impact is (e.g. on the GC benchmark). If it would be too bad, we may need to go for an invalid reserved page on top of the memory (wrap on underflow) for faster (hardware-supported) bounds check (as @ggreif suggested).

src/codegen/compile.ml
ggreif: I'd also offer users an option to configure it at compile time (a.k.a. `--rts-stack-pages=64` or such.

ggreif (approved):
I agree that this is a correct fix, and it leaves room for improvement. I think this can go in for now and should be run by the user who had the crashing symptoms. The +0.9% size regression (which is probably inflated) is a sacrifice I gladly make for the added security.

src/codegen/compile.ml
crusso: Sure - all reasonable optimizations, but I think we should always have some form of reliable guard. Let's see if this actually adds to the cost much anyway.

src/codegen/compile.ml
crusso: > I'd also offer users an option to configure it at compile time (a.k.a. --rts-stack-pages=64 or such) for special needs.
Yes I was wondering that too. It might let people recover from an upgrade that serializes ok but fails to deserialize.

crusso: Interestingly, the max_wasm_stack size allowed by the replica appears to be 5 MiB!
https://gitlab.com/dfinity-lab/public/ic/-/blob/903c8b3a520c69a953b4cdf6468bf2612e86be49/rs/config/src/embedders.rs#L76

test/run-drun/rts-stack-pages.mo
```diff
@@ -0,0 +1,48 @@
...
+   assert (opt != null);
+   // if we got here, deserialization has succeeded
+   Prim.debugPrint "deserialized";
+   assert false;
+
+}
+//SKIP run
+//SKIP run-ir
+//SKIP run-low
+//SKIP ic-ref-run
```
crusso, line 48:
```suggestion
//SKIP ic-ref-run

```

crusso:
Odd. Don't understand why this is failing on a debug run...

ggreif (approved):
Awesome!

You'll need to re-accept:
``` diff
test-drun-compacting-gc> -ingress Err: IC0503: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped explicitly: RTS Stack underflow
[2504](https://github.com/dfinity/motoko/actions/runs/4129130685/jobs/7134403859#step:7:2505)
test-drun-compacting-gc> +ingress Err: IC0502: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped: heap out of bounds
[2505](https://github.com/dfinity/motoko/actions/runs/4129130685/jobs/7134403859#step:7:2506)
```

crusso:
> Awesome!
> 
> You'll need to re-accept:
> 
> ```diff
> test-drun-compacting-gc> -ingress Err: IC0503: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped explicitly: RTS Stack underflow
> [2504](https://github.com/dfinity/motoko/actions/runs/4129130685/jobs/7134403859#step:7:2505)
> test-drun-compacting-gc> +ingress Err: IC0502: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped: heap out of bounds
> [2505](https://github.com/dfinity/motoko/actions/runs/4129130685/jobs/7134403859#step:7:2506)
> ```

Yeah, its a bit more difficult than that we seem to get different behaviour in debug and production gc. Luc and I have been struggling with this ost of the day but one mitigation is to leave a bit of reserve for any leaf Rust functions.

crusso:
@luc-blaeser passed CI with reserve of 4K. Shall I merge?

crusso:
@grreif Luc and I added a small amount of reserve stack so any Rust leaf function that deserialization calls has a chance to run even when the stack is filled (but not overflowed) by the last Motoko stack allocation. It's a bit of hack to ensure we get the same output on both production and sanity checking builds of the rts (where sometimes we see RTS Stack underflow (from Motoko) and sometime heap-out-of-bounds from a Rust helper. All a bit unsatisfactory.

crusso:
possibly superseeded by #3789

crusso:
superseded by #3793

PR #3783, opened 2023-02-08
Test `rustc` upgrade

luc-blaeser:
Generational GC seems to schedule at somewhat different points in the tests and thus occasionally exceed the limit for certain tests (that were already close at the limit before). In debug mode, the generational GC performs extensive full-heap sanity checks.

### DIFF
--- a/doc/md/about-this-guide.md
+++ b/doc/md/about-this-guide.md
@@ -52,7 +52,7 @@ For transparency into the principles that guide the engineering effort, the engi
 
 The following guiding principles represent the core values of the engineering organization in prioritized order:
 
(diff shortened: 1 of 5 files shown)


ggreif:
This is green now. Do you want to tweak the numbers further?

Anyway, thanks for doing this!

ggreif:
Question: where does the -3.8% size change come from?

ggreif (approved):
Thanks for figuring this out!

luc-blaeser:
> Question: where does the -3.8% size change come from?

I am not sure: Maybe it is due to some optimizations in the new Rust compiler. But this would need to be analyzed.

ggreif:
Let's see if those 3.8% also happen on the parent branch :-)

PR #3788, opened 2023-02-09
experiment: reserve last page for stack overflow detection

crusso:
exploratory alternative to #3782 
* reserve last page of memory for Rust stack overflow detection
* instrument allocations that might not land on last page (but a previous page) to force an explicit access violation (by reading from last, unallocated word in full wasm memory.)
* omit bounds checks to accesses that will land within last page, trigger AV on subsequent access
* prevent rust page allocator from allocating last page of memory (implements reservation).

Advantage over #3782:
* most  stack guards for static allocations are left implicit (less instructions, less cost). 
* all stack overflows are reported as access violations (IC "heap out of bounds")
* no need for hacky stack reservation at top of stack for Rust/C functions.


rts/motoko-rts/src/memory/ic.rs
```diff
@@ -96,5 +96,9 @@ unsafe fn grow_memory(ptr: u64) {
         if wasm32::memory_grow(0, total_pages_needed - current_pages) == core::usize::MAX {
             rts_trap_with("Cannot grow memory");
         }
+        if wasm32::memory_size(0) == 65536 {
```
ggreif, line 99: please see #3090!

crusso: Yes, that looks promising!

src/codegen/compile.ml
```diff
@@ -1190,7 +1211,9 @@ module Stack = struct
     compile_divU_const Heap.word_size ^^
```
crusso, line 1211: Optimize. Don't divide, just compare against n * word_size needed later anyway

crusso:
Note: this version regresses bench/palindrome because every page allocation does a bit more work, while in the other PR, only stack allocations do more work (and these mostly occur during deserialization and I guess not at all in palindrome.

We  should maybe add  a simple serialization benchmark, just using to_candid/from_candid.

crusso:
incorporated in #3793

PR #3789, opened 2023-02-10
experiment: stack guard with fast-path allocation

crusso:
Alternative to https://github.com/dfinity/motoko/pull/3782

* reserve last page of memory for Rust stack overflow detection
* instrument stack allocations that might not land on last page (but a previous page) to force an explicit access violation (by reading from last, unallocated word in full wasm memory.)
* omit bounds checks to accesses that will land within last page, trigger AV on subsequent access

Advantage over https://github.com/dfinity/motoko/pull/3782:

* most stack guards for static allocations are left implicit (less instructions, less cost).
* all stack overflows are reported as access violations (IC "heap out of bounds"), not some as "RTS Stack underflow" and some as "heap out of bounds"
* no need for hacky stack reservation at (logical) top of stack for Rust/C functions.


crusso:
This looks promising, but Timer.mo got in the way of CI again. Re-runing the job. And we probably want to merge the final version of #3090.

src/codegen/compile.ml
```diff
@@ -1150,22 +1150,43 @@ module Stack = struct
...
+    Func.share_code0 env "stack_overflow" [] (fun env ->
+      (* read last byte of reserved page to force trap *)
+      compile_unboxed_const 0xFFFF_FFFCl ^^
+      G.i (Load {ty = I32Type; align = 2; offset = 0l; sz = None}) ^^
+      G.i Drop
+    )
+
   let alloc_words env n =
+    (* first, check for stack underflow, if necessary *)
+    (if (n >= page_size) then
```
crusso, line 1180:
```suggestion
    (if (n >= Int32.div(page_size, v4)) then
```

PR #3794, opened 2023-02-13
experiment: investigate candid stack limits

crusso:
instrument the compiler to report maximum rts stack size using new prim `rts_max_stack_size` to determine suitable default for new compiler flag  `--rts-stack-pages`.

Added a simple test that tries to serialize and, optionally, deserialize a linked list, exploring maximum rts stack depth required.

TL/DR:
Unfortunately, we can serialize a list of  up to 39K elements, but only deserialize a list of around 13K elements. Which is pretty shabby. 
Moreover, the limiting stack appears to be not the rts C stack, but the Wasm stack, with deserialization apparently requiring more frames than deserialization, perhaps because of the support for backtracking option type, or other differences.

The good news, if you call that good news, is that we actually don't seem to need 32 pages of rts stack and a smaller default of 4 or so (not 32) might suffice. The bad news is that the quick fix of bumping the stack limit probably won't help users avoid the heap-of-bounds error for large, recursive data-structures. They will just be swapping the heap out of bounds for an an imminent Wasm stack overflow (I expect).

UPDATE: using variants, not option based lists, just increases the stack requirements a bit, but doesn't let us de-serialize longer lists.

UPDATE: running on the playground, not drun, doesn't seem to change the results, so I expect drun has the same wasm stack limits as the real replicas.

 Another nail in the coffin of using Candid for stable variables, I'm afraid.

Looking at compile.ml, the recursive functions serialize_go vs. deserialize_go take 3 vs. 9 arguments, which may well account for the difference in stack consumption, let alone any difference in locals per stack frame.

For deserialize, almost all the parameters are simply copied down the stack, unchanged, and could probably just be stored once on the rust stack or, even simpler, a re-used global.


src/codegen/compile.ml
```diff
@@ -1192,6 +1215,7 @@ module Stack = struct
     compile_unboxed_const n_bytes ^^
     G.i (Binary (Wasm.Values.I32 I32Op.Sub)) ^^
     set_stack_ptr env ^^
+    update_stack_min env ^^
```
luc-blaeser, line 1218: The assumption is that the allocated stack space is certainly accessed before the next allocation, in order to trap on an overflow (with allocation sizes < page size). Otherwise, multiple smaller allocations (without intermediate accesses) could skip below the last invalid page. I guess this is always guaranteed?

luc-blaeser (approved):
Great work in combination with the related PRs!
Some tests in CI failing, seems to be resolved with the top-most PR https://github.com/dfinity/motoko/pull/3809

src/codegen/compile.ml
crusso: Yeah, that was bothering me a little. There is no hard guarantee beyond assuming the code generator won't do stack allocation without using the space shortly. Maybe it would actually be better to always check, even when small or add an explicit argument to suppress the check so it's more obvious at call sites.

crusso: What if we instead forced a read from the last word, only when the allocation is small, before returning?
That would be slower, but safer.
@luc-blaeser @ggreif?
I'll open a PR on master to measure the cost of this...

crusso: See #3823

PR #3798, opened 2023-02-14
experiment: compress deserialization stack frames using globals for invariant arguments

crusso:
Compress deserialization stack frames so that the deserialization can actually cope with the depth of previously serialized data.

This PR attempts that by just passing the large number of unchanging/invariant arguments in 
dedicated registers (globals) not on the stack.

More than doubles capacity, but still not quite a match for producer (serialization).

The follow on PR #3809 reaches parity in serialization/deserialization capacity (on these tests) by additionally passing all deserialization arguments on the Rust (not Wasm) stack, reducing pressure on the Wasm stack.

Builds on #3794


crusso:
superseded by #3809

luc-blaeser (approved):
Great work in combination with the related PRs!
(The`timer` test is failing in the latest CI build, but this is due to some time dependency...)

PR #3804, opened 2023-02-16
feat(js): basic forward-compatible JS result values for `run()`

rvanasa:
The `run()` function in `moc.js` previously only returned `stdout` and `stderr` strings, along with a `result` value that was always `0` (moc.js conversion of `()`) regardless of whether the evaluation succeeded. 

This PR replaces this previously unused `result` value with `{ error: null }` for a successful run and `{ error: {} }` otherwise, setting the groundwork for returning more complex JS results in the future (for example, `{ error: { message: "arithmetic overflow", location: { ... } } }` or `{ error: null, value: [123, "abc"] }`). 

Returning JS results from the interpreter is motivated by the new Motoko test runner logic used in https://github.com/dfinity/vscode-motoko/pull/170 as well as recent conversations about using the interpreter with testing frameworks such as [Jest](https://jestjs.io/).


chenyan-dfinity (commented):
Is the idea to eventually convert `Value.value` into a JS value and make comparison on the JS side? What do we plan to do if the test is run in WASI mode?

rvanasa:
> Is the idea to eventually convert Value.value into a JS value and make comparison on the JS side?

Right; this would be a general-purpose feature (rather than for a test runner, which only needs something equivalent to an exit code). The main use cases right now are @krpeacock's idea of using Jest for mixed Motoko/JS unit testing and an interactive output panel in Embed Motoko.

chenyan-dfinity (approved):
Okay, so there is no WASI equivalent way of doing this. Or WASI can import an `ArrayBuffer` from JS and emit JS values as well?

rvanasa:
> Okay, so there is no WASI equivalent way of doing this.

We could do something similar for stable values by running `to_candid` on the return value and then converting via `agent-js`.

Here is a proof-of-concept that could be used to preprocess the script (appears to work with Wasmer / Wasmtime):

```
import __Prim__ "mo:prim";
// inject test imports

__Prim__.debugPrint("__output__" # debug_show to_candid(do {
  // inject test body
}));
```

I'm sure we could find a much more elegant way to do this, but this does indeed seem viable for Candid-serializable values.

PR #3805, opened 2023-02-16
rts: selective GC features

### LINKED ISSUE #3795: Separate runtimes for different GC strategies
Right now we have a release and a debug build o the RTS:
```
rts/mo-rts-debug.wasm
rts/mo-rts.wasm
```
these can be selected individually by giving flags to the `moc` build.

I propose to have 4 release builds, which contain the relevant GC strategy (only):
- copying
- compacting
- generational
- incremental

I do not care about the debug build being too big.

ggreif:
This fixes #3795. The idea is that a selected GC strategy should only pull in RTS code for exactly that strategy, thus avoiding dead code.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -257,9 +257,9 @@ rec {
 
       installPhase = ''
         mkdir -p $out/rts
-        ln mo-rts.wasm $out/rts
-        ln mo-rts-generational.wasm $out/rts
-        ln mo-rts-debug.wasm $out/rts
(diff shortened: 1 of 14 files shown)


luc-blaeser:
Mostly done in current RTS, EOP, classical mode, incremental and non-incremental GC. We should simplify things in future and cut down to a single mode for better maintainability and engineering velocity.

PR #3807, opened 2023-02-16
feat: bindings in `or`-patterns

ggreif:
Currently `or` patterns cannot harbour bindings. In Rust (and OCaml) they can. This is a (slow-moving) experiment to add them.

### TODOs
- [x] `test/fail` cases
- [x] checking
- [x] inference
  - [x] three-way `or` (also with conflicting types and `None` in the middle)
  - [x] checking also
- [x] interpreter
- [x] IR-interpreter
- [x] compiler
- [ ] coverage tests
- [x] remove `test/fail/M0105.mo`
- [x] decide which alternative applies
- [x] joining bindings for downstream use
- [ ] warn duplicates?
- [ ] documentation
- [x] changelog
- [x] check off the white paper mention of `or`-patterns

See also https://dfinity.atlassian.net/browse/LANG-121


src/mo_interpreter/interpret.ml
```diff
@@ -774,10 +774,14 @@ and define_pat env pat v =
   let err () = trap pat.at "value %s does not match pattern" (string_of_val env v) in
   match pat.it with
   | WildP -> ()
-  | LitP _ | SignP _ | AltP _ ->
+  | LitP _ | SignP _ ->
     if match_pat pat v = None
     then err ()
     else ()
+  | AltP (pat1, pat2) ->
+     if match_pat pat1 v = None
```
ggreif, line 782:
```suggestion
    if match_pat pat1 v = None
```

test/run/or-pattern-binds.mo
```diff
@@ -0,0 +1,23 @@
...
+// (should work, fails currently)
+let #a a4 or #b a4 = #a 5;
+let #a a5 or #b (a5 : Int) = #b 7;
+
+let (#a a6 or #b a6, _) = (#a 5, "Ignore");
+
+debugPrint (debug_show {a2l; a2r; a4; a5; a6});
+
+// should reject
+//YESSS! func foo(#a (a7 : Nat) or #b (a7 : Int)) : Nat = a7;
```
ggreif, line 13: move to separate test

src/codegen/compile.ml
```diff
@@ -10384,7 +10384,7 @@ and const_exp_matches_pat env pat exp : bool option =
   assert exp.note.Note.const;
   match exp.it with
   | _ when Ir_utils.is_irrefutable pat -> Some true
-  | PrimE (TagPrim _, _) ->
+  | PrimE ((TupPrim | ArrayPrim (Const, _) | TagPrim _), _) ->
```
ggreif, line 10387: This is not enough, `compile_const_exp` is more general.

src/lang_utils/error_codes.ml
```diff
@@ -187,4 +187,6 @@ let error_codes : (string * string option) list =
     "M0181", None; (* Verification mode assertions not allowed *)
     "M0182", None; (* Timer defined by user when -no-timer flag is specified *)
     "M0183", None; (* Incompatible async sorts *)
+    "M0185", None; (* Different set of bindings in pattern alternatives *)
```
ggreif, line 190: M0184!

src/mo_frontend/typing.ml
```diff
@@ -1872,9 +1872,16 @@ and infer_pat' env pat : T.typ * Scope.val_env =
         "pattern branches have incompatible types,\nleft consumes%a\nright consumes%a"
         display_typ_expand t1
         display_typ_expand t2;
-    if ve1 <> T.Env.empty || ve2 <> T.Env.empty then
-      error env pat.at "M0105" "variables are not allowed in pattern alternatives";
-    t, T.Env.empty
+    if T.Env.keys ve1 <> T.Env.keys ve2 then
+      error env pat.at "M0184" "different set of bindings in pattern alternatives";
+    let check_same_bind_type bind t1 t2 = if not (T.eq t1 t2) then
+      local_error env pat.at "M0185" "types for alternative pattern variables %s differ%a%a"
```
ggreif, line 1878: this should be `error`, since it makes checking the function's body impossible. Also need to spit the `fail/` test up in parts.

src/ir_def/check_ir.ml
```diff
@@ -1019,9 +1018,9 @@ and check_pat env pat : val_env =
     let ve2 = check_pat env pat2 in
     t <: pat1.note;
     t <: pat2.note;
-    check env pat.at (T.Env.is_empty ve1 && T.Env.is_empty ve2)
-      "variables are not allowed in pattern alternatives";
-    T.Env.empty
+    let common i1 i2 =
+      { typ = T.lub i1.typ i2.typ; loc_known = false; const = i1.const && i2.const } in
+    T.Env.merge (fun _ -> Lib.Option.map2 common) ve1 ve2
```
ggreif, line 1023: I am super insecure about this implementation.

crusso: Especially the const bit? What if they are different values, but both const?

ggreif: Hmmm, `check_pat` will always have `const = false`, but `gather_pat` needs a closer look as it receives `const` as a passed-in flag. `AltP` there is not symmetrical currently, that needs to be revised too. But if a `LetD` has a `const` RHS, than either alternative will bind to a constant, so the `&&`-connective will give the right result.

ggreif: this should be fixed in 0a4d4ab49

ggreif: We could factor out the `T.Env.merge (fun _ -> Lib.Option.map2 common)` parts to a top-level `merge_val_envs`.

src/mo_frontend/typing.ml
ggreif: see 130c51429 (and parent)

doc/md/language-manual.md
```diff
@@ -1370,7 +1371,8 @@ The or pattern `<pat1> or <pat2>` is a disjunctive pattern.
 
 The result of matching `<pat1> or <pat2>` against a value is the result of matching `<pat1>`, if it succeeds, or the result of matching `<pat2>`, if the first match fails.
 
-(Note, statically, neither `<pat1>` nor `<pat2>` may contain identifier (`<id>`) patterns so a successful match always binds zero identifiers.)
+(Note, a disjunctive pattern may contain identifier (`<id>`) patterns with the restriction that both alternatives must bind the same identifiers and these must
```
crusso, line 1374:
```suggestion
(Note, a disjunctive pattern may contain identifier (`<id>`) patterns with the restriction that both alternatives must bind the same set of identifiers with equivalent types.
```

doc/md/language-manual.md
```diff
@@ -1370,7 +1371,8 @@ The or pattern `<pat1> or <pat2>` is a disjunctive pattern.
...
+obtain values of the same type.)
```
crusso, line 1375:
```suggestion
(Note, a disjunctive pattern may contain identifier (`<id>`) patterns with the restriction that both alternatives must bind the same set of identifiers. Each identifier's type is the least upper bound of its type in `<pat1>` and `<pat2>`.
```
Ahh, but I now see that the source typing rules require equality but the ir ones just use the lub....

src/mo_interpreter/interpret.ml
```diff
@@ -773,7 +773,8 @@ and define_id env id v =
 and define_pat env pat v =
   match pat.it with
   | WildP -> true
-  | LitP _ | SignP _ | AltP _ -> match_pat pat v <> None
+  | LitP _ | SignP _ -> match_pat pat v <> None
+  | AltP (pat1, pat2) -> define_pat env (if match_pat pat1 v <> None then pat1 else pat2) v
```
crusso, line 777:
```suggestion
  | AltP (pat1, pat2) -> define_pat env (if match_pat pat1 v <> None then pat1 else pat2) v
```
This recursive call to `match_pat` looks fishy to me. I think we maybe need to rethink why we do the separate `define_pat` pass in the first place. I wonder if it predates the definedness check that Joachim introduced and was used to define identifiers after initially black-holing them, to detect undefinedness dynamically. Maybe there is a better way.

crusso: Indeed, looking at the code I think we could elimate define_pat using match_pat and a simpler map over the resulting environment to resolve all the promised recursive bindings (maybe). 
This would be both in the ir and source interpreters. Best done as a separate PR first, I think.

doc/md/language-manual.md
ggreif: Yes, the type checker wants equality. The error message should make that clear. Further downstream I am using `lub` for that being the correct connective if we want to water up the type-checker at a later point in time. It does the right thing when the types are equal.

I chose strict equality, so that no information is being lost when folks use `or` patterns. Otherwise we could get into the situation where the `lub` becomes `Any` and we'd have to warn like for the array construction case. If the user wants to deal with different types, that is still totally possible by writing out both `case`s.

test/run/or-pattern-binds.mo
```diff
@@ -0,0 +1,20 @@
...
+// should accept
+//func quux(#a (a7 : Nat) or #b a7) : Nat = a7;
+
+let v8 : { #a : Nat; #c : Nat } = #c (5 + 3);
+let #a a8 or #b a8 or #c a8 = v8;
+
+debugPrint (debug_show {a8});
+
+// should accept
+//func three(#a (a9 : Nat) or #b (a9 : Nat) or #c (a9 : Nat)) : Nat = a9;
```
ggreif, line 20: This gives
``` diff
+Ill-typed intermediate code after Desugaring (use -v to see dumped IR):
+or-pattern-binds.mo:20.12-20.42: IR type error [M0000], subtype violation:
+  {#a : Nat; #b : Nat}
+  {#a : Nat}
```
in the IR type-checker. (Similarly when I force association to the right.)

doc/md/language-manual.md
ggreif: The other reason to use `lub` in the IR is that `typing.ml` assigns type `Non` to dead branch patten bindings.

test/run/or-pattern-binds.mo
ggreif: At least it is not a regression, as
``` Motoko
func three(#a () or #b () or #c ()) {};
```
also fails :-)

ggreif: Still a regression relative to `master` :-(

doc/md/language-manual.md
rossberg: In a language with subtyping not using a lub here would be fishy, because that immediately breaks subtype weakening. Changing the type of some variable to a subtype should not affect typeability of a program -- that would be an obvious "refactoring hazard". Of course, warning if the lub is lossy is still a good idea.

src/mo_interpreter/interpret.ml
crusso: https://github.com/dfinity/motoko/pull/4011 attempts to simplify this.

doc/md/language-manual.md
crusso: It looks to me like you are taking the lub now, after warning (in typing.ml). Looks good.

src/mo_interpreter/interpret.ml
crusso: I think once you mere 4011, there should be not change to the interpreter code.

ggreif: Yes, I noticed. The interpreters are now almost unaffected.

Changelog.md
```diff
@@ -1,5 +1,9 @@
 # Motoko compiler changelog
 
+* motoko (`moc`)
+
+  * Allow indentifiers in `or`-patterns (#3807).
```
crusso, line 5:
```suggestion
  * Allow identifiers in `or`-patterns (#3807).
```
Maybe add a simple example?
```
switch r {
   case #ok t or #err t -> t
};
```

src/ir_def/check_ir.ml
ggreif: Yes, I think I am happy.

test/run/or-pattern-binds.mo
```diff
@@ -0,0 +1,20 @@
...
+let (#a a2r) or (#b a2r) : { #a : Nat; #b : Nat } = #b 7;
+let #a a4 or #b a4 = #a 5;
+let #a a5 or #b (a5 : Int) = #b 7;
+
+let (#a a6 or #b a6, _) = (#a 5, "Ignore");
+
+debugPrint (debug_show {a2l; a2r; a4; a5; a6});
+
+// should accept
+//func quux(#a (a7 : Nat) or #b a7) : Nat = a7;
```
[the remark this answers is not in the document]
ggreif: Inference has recently been wholesale disabled.

PR #3817, opened 2023-02-17
feat: `let ... else` binding

kentosugama:
Original discussions in #3588.

New `let` binding that allows a failure block in case of a pattern-match failure. The main motivation for this feature is to avoid deeply nested switch statements that lead to less-readable code.

See forum discussion [here](https://forum.dfinity.org/t/match-and-take-in-motoko-do-for-variants-was-when/13427/6?u=kentosugama).

Simple example:
``` Motoko
let ?x = foo() else { Debug.print "foo failed"; return };
```

The failure block must always have type `None` to ensure that execution does not continue beyond the failed binding. Failure blocks may thus trap, return out of the function, or break / continue if in the context of a loop.

Failure blocks may also perform asynchronous effects, if the binding itself is in an asynchronous context.

``` Motoko
let (#ok x) = foo() else { await logError(); return; };
```

More complex example:
Without this binding, you might have to either do the following
``` Motoko
switch (foo()) {
  case (?x) {
    switch (bar()) {
      case (?y) {
        switch (baz()) {
          case (?z) {
            return #ok x + y + z;
          }
          case null {
            return #err "baz failed";
          }
        }
      };
      case null {
        return #err "bar failed";
      }
    }
  };
  case null {
    return #err "foo failed with";
  }
}
```
or
``` Motoko
switch (foo(), bar(), baz()) {
  case (?x, ?y, ?z) {
    return #ok x + y + z;
  }
  case (null, _, _) {
    return #err "foo failed";
  }
  case (_, null, _) {
    return #err "bar failed";
  }
  case (_, _, null) {
    return #err "baz failed";
  }
}
```
however this second version does not execute examples sequentially (i.e. even if `foo()` fails, `bar()` and `baz()` will execute).

With `let-else` bindings:
``` Motoko
let ?x = foo() else { return #err "foo failed" };
let ?y = bar() else { return #err "bar failed" };
let ?z = baz() else { return #err "baz failed" };
   
return #ok x + y + x
```

Inspired by the same feature in [rust](https://rust-lang.github.io/rfcs/3137-let-else.html)

------------------------------------------------------

Items left to do:
- [x] Change parser to parse the body of `else` as a block instead of a value
- [x] Have `return` at the end of `else` block return out of the function
- [x] Finish the compilation with desugaring
- [x] Come up with lots of positive (`run`) and negative (`fail`) tests
  - [x] `return`, `continue`, `break` in failure part (also with 0, 1 and more return values)
  - [x] `return` in monadic context (`async`)
  - [x] `await` in failure part
- [x] Assign error code and explanation (`M0500` is provisional) — _not needed_
- [x] Suppress coverage warning when fail-expression is present
- [x] Let-bound terminal expressions (in `compUnit.ml`)?
- [x] Double check all the `None` pattern matches
- [x] Add documentation
- [x] remove trailing whitespace from PR
- [x] add trailing newlines to files that miss it
- [x] Fix repl tests
- [x] Fix async effects
- [x] Fix null pattern matches
- [x] Find more complex example

Other PR:
- `type continuation = V.(value cont);`
- May want to warn dead code when fail present for irrefutable match

Potential further optimisations:
- `let (b_1, b_2, ..., b_n) = switch { case <pat> (b_1, b_2, ..., b_n); ... };`


crusso (commented):
Added some interim comments on doc. Now looking at code.

src/mo_interpreter/interpret.ml
```diff
@@ -939,7 +939,8 @@ and declare_dec dec : val_env =
   match dec.it with
   | ExpD _
   | TypD _ -> V.Env.empty
-  | LetD (pat, _) -> declare_pat pat
+  | LetD (pat, _, None) -> declare_pat pat
+  | LetD (pat, _, Some _) -> declare_pat pat
```
crusso, line 943:
```suggestion
  | LetD (pat, _, _) -> declare_pat pat
```

src/mo_interpreter/interpret.ml
```diff
@@ -771,35 +771,35 @@ and define_id env id v =
...
     if lab = i.it
     then define_pat env pat1 v1
-    else err ()
+    else false
   | AnnotP (pat1, _)
   | ParP pat1 -> define_pat env pat1 v
 
 and define_pats env pats vs =
-  List.iter2 (define_pat env) pats vs
+  List.fold_left (fun acc (pat, v) -> acc && define_pat env pat v) true (List.combine pats vs)
```
crusso, line 799:
```suggestion
  List.forall2 (fun (pat, v) -> define_pat env pat v) pats vs
```
Faster and allocates less.

src/mo_interpreter/interpret.ml
```diff
@@ -771,35 +771,35 @@ and define_id env id v =
...
 
 and define_pat_fields env pfs vs =
-  List.iter (define_pat_field env vs) pfs
+  List.fold_left (fun acc pf -> acc && define_pat_field env vs pf) true pfs
```
crusso, line 802:
```suggestion
  List.for_all (fun pf -> define_pat_field env vs pf) pfs
```
Simpler?

src/mo_frontend/typing.ml
```diff
@@ -2337,7 +2338,12 @@ and infer_dec env dec : T.typ =
   let t =
   match dec.it with
   | ExpD exp
-  | LetD (_, exp) ->
+  | LetD (_, exp, None) -> infer_exp env exp
+  | LetD (_, exp, Some fail) ->
+    if env.in_actor then
+      error env dec.at "M0184" "let-else binding forbidden for actor fields";
```
crusso, line 2344: NIT: I wonder if you need to do this check/add this restriction.

I don't think you can `return` or `throw` here anyway, but why shouldn't a user decide to trap with a better message:

```
actor a {
     ...
     let ?b = find(id) else Debug.trap("coudn't find" # id);
     ...
}
```

src/mo_frontend/typing.ml
```diff
@@ -2637,9 +2646,13 @@ and infer_dec_valdecs env dec : Scope.t =
...
     Scope.{empty with val_env = T.Env.singleton id.it obj_typ}
-  | LetD (pat, exp) ->
+  | LetD (pat, exp, fail) ->
     let t = infer_exp {env with pre = true} exp in
-    let ve' = check_pat_exhaustive (if is_import dec then local_error else warn) env t pat in
+    let consequence = match fail with
+      | None when is_import dec -> local_error
+      | None -> warn
+      | Some _ -> fun _ _ _ -> Format.kasprintf ignore in
+    let ve' = check_pat_exhaustive consequence env t pat in
```
crusso, line 2655: I would maybe split this into two cases, one for Let(pat, exp, None) and one for Let(pat, exp, Some fail).

In the second case, I'd call check_pat instread of check_pat_exhaustive with an error function that discard the error.

Something like

```
| LetD (pat, exp, None) ->
    let warnOrError =
      if is_import dec then
        local_error
      else warn
    in
    let ve' = check_pat_exhaustive warnOrError env t pat in
    Scope.{empty with val_env = ve'}
  | LetD (pat, exp, Some fail) ->
    let ve' = check_pat env t pat in (* need not be exaustive *)
    Scope.{empty with val_env = ve'}
```

Or maybe even just:

```
| LetD (pat, exp, fail) ->
    let ve' = match fail with
      | None ->
        check_pat_exhaustive
         (if is_import dec then local_error else warn)
         env t pat
      | Some _exp1 ->
         check_pat env t pat
    in
    Scope.{empty with val_env = ve'}

```

src/mo_frontend/static.ml
```diff
@@ -99,7 +99,7 @@ and exp_fields m efs = List.iter (fun (ef : exp_field) ->
 and dec m d = match d.it with
   | TypD _ | ClassD _ -> ()
   | ExpD e -> exp m e
-  | LetD (p, e) -> pat m p; exp m e
+  | LetD (p, e, fail) -> pat m p; exp m e; Option.iter (exp m) fail
```
crusso, line 102: Note to self: I guess this will error pretty soon after entering any fail clause since all control flow construct error.

src/mo_def/compUnit.ml
```diff
@@ -44,9 +44,9 @@ let comp_unit_of_prog as_lib (prog : prog) : comp_unit =
       assert (List.length tbs > 0);
       finish imports { it = ActorClassU (sp, tid, tbs, p, typ_ann, self_id, fields); note = d.note; at = d.at }
     (* let-bound terminal expressions *)
-    | [{it = LetD ({it = VarP i1; _}, ({it = ObjBlockE ({it = Type.Module; _}, fields); _} as e)); _}] when as_lib ->
+    | [{it = LetD ({it = VarP i1; _}, ({it = ObjBlockE ({it = Type.Module; _}, fields); _} as e), None); _}] when as_lib ->
       finish imports { it = ModuleU (Some i1, fields); note = e.note; at = e.at }
-    | [{it = LetD ({it = VarP i1; _}, e); _}] when is_actor_def e ->
+    | [{it = LetD ({it = VarP i1; _}, e, None); _}] when is_actor_def e ->
```
crusso, line 49: Both of these worry me a little.
What if a library was written in desugared form with a fail clause
List.mo
```
let List = module { } else { loop {}}
```
If that is legal (and I think it is), will importing List.mo just crash.

Similarly,if the main actor file is authored as:


main.mo
```
let a = actor {} else {loop {}};
```
Will `moc main.mo` crash?

src/lowering/desugar.ml
```diff
@@ -701,14 +703,16 @@ and dec d = { (phrase' dec' d) with note = () }
...
     let p' = pat p in
     let e' = exp e in
     (* HACK: remove this once backend supports recursive actors *)
-    begin match p'.it, e'.it with
-    | I.VarP i, I.ActorE (ds, fs, u, t) ->
+    begin match p'.it, e'.it, f with
+    | I.VarP i, I.ActorE (ds, fs, u, t), None ->
       I.LetD (p', {e' with it = I.ActorE (with_self i t ds, fs, u, t)})
-    | _ -> I.LetD (p', e')
+    | _, I.ActorE _, Some _ -> assert false
```
crusso, line 713: why can't this happen?

src/lowering/desugar.ml
```diff
@@ -321,7 +321,7 @@ and call_system_func_opt name es obj_typ =
   List.find_map (fun es ->
     match es.it with
     | { S.vis = { it = S.System; _ };
-        S.dec = { it = S.LetD( { it = S.VarP id; note; _ }, _); at; _ };
+        S.dec = { it = S.LetD( { it = S.VarP id; note; _ }, _, None); at; _ };
```
crusso, line 324: Are you sure once can't write

```
system let preupgrade = func () {} else { loop {}}
```
In which case this match won't apply as expected?

crusso: Or is this ruled out by type error M0184?

crusso (commented):
Left some comments. I'm still worried by some of those pattern matches on None.

crusso (changes requested):
Left some comments. Generally looks good but I'm still worried by some of those pattern matches on None. The problem is that alot of our actor/module/object sugar desugars to let expressions (to support recursion) but the code to spot these may now fail or behave strangely differently when that desugaring suddenly contains a fail clause (because the author didn't use the sugar but programmed directly in the target language. I guess the problem wouldn't exist if people could not both write, e.g.

```
module X = {}
```
and 

```
let X = module {}
```
(but now also

```
let X = module {} else {...}
```

The first two are treated the same by the compiler, but the last winds up being treated differently in some cases the programmer might not be expecting.

src/mo_interpreter/interpret.ml
kentosugama: Ah, of course that's a thing

doc/md/language-manual.md
```diff
@@ -1394,6 +1394,15 @@ The declaration `let <pat> = <exp>` evaluates `<exp>` to a result `r`. If `r` is
 
 All bindings declared by a `let` (if any) are *immutable*.
```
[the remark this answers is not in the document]
luc-blaeser: Detail: Maybe one could briefly mention the restrictions, namely that let-else cannot be used in field initializers (with a brief rationale).

src/mo_frontend/typing.ml
kentosugama: Just talked about this with Gabor. I think the reason to keep the restriction is that:

If you don't restrict this behavior, you can infinitely loop in the else block. And that would be both annoying and expensive to debug.

Also, if we add this restriction now and remove it later (because some user really wants to add custom trap messages in actor field initialization) that's not a breaking change, verses the reverse would be.

What do you think?

kentosugama: Hmmm, but this argument also applies to non-actors, like objects...

kentosugama: Okay I'll try and remove the restriction

src/lowering/desugar.ml
kentosugama: I removed the type error, but this is giving a syntax error (misplaced token "system")

crusso: Fixed

src/mo_def/compUnit.ml
crusso: blah

crusso (approved):
Nice job!

PR #3823, opened 2023-02-20
experiment: add stack probe on small allocations

crusso:
In https://github.com/dfinity/motoko/pull/3794#discussion_r1111655007 Luc notes that relying on the access violation for small allocations only works if we are guaranteed to access the stack slot before the next allocation. Otherwise one could sneak past the guard using many little allocations.

This PR introduces a redundant read from just below the stack pointer on small allocations, to force the overflow error. I thought it would be very expensive but it doesn't seem to affect our benchmarks. Perhaps because those don't actually do very much RTS stack allocation, I expect. I think it mostly happens during deserialization.

@luc-blaeser @ggreif 

The other option is to manually specify in Stack.with_words whether to add or omit the check, e.g. `Stack.with_words (Checked|Unchecked)`, and review all call-sites to see if we can omit the check due to an imminent stack access.

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -1196,7 +1196,6 @@ module Stack = struct
     else if n > 0l then begin
       (* force read of last word *)
       get_stack_ptr env ^^
-      get_stack_ptr env ^^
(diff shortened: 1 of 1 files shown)


luc-blaeser:

Issue #3819, opened 2023-02-18
Bug: bad ir


crusso:
https://forum.dfinity.org/t/type-error-m0139-inner-actor-classes-are-not-supported-yet/18135/19?u=claudio

TBC

crusso:
Has this been fixed?

ggreif:
> Has this been fixed?

I doubt it.

PR #3824, opened 2023-02-21
test: repro for issue-3819 (await on async of type None crashes compiler

crusso:
repro for #3819


luc-blaeser:

PR #3827, opened 2023-02-22
chore: update `Timer.md` with caveats

ggreif:
... and suggestions. Touch on security and the default mechanism.

Also linking @dfinity-berestovskyy's timer FAQ, which is https://internetcomputer.org/docs/current/developer-docs/backend/periodic-tasks#timers-library-limitations


doc/md/base/Timer.md
```diff
@@ -1,11 +1,25 @@
...
+Timers for one-off or periodic tasks. Applicable as part of the default mechanism.
 
 Note: If `moc` is invoked with `-no-timer`, the importing will fail.
+
 Note: The resolution of the timers is in the order of the block rate,
       so durations should be chosen well above that. For frequent
       canister wake-ups the heatbeat mechanism should be considered.
 
+Note: Below functionality is active only if the user doesn't explicitly
+      implement `system func timer`.
```
crusso, line 12:
```suggestion

Note: The functionality described below is enabled only when the actor does not override it by declaring an explicit `system func timer`.
```

doc/md/base/Timer.md
```diff
@@ -1,11 +1,25 @@
 # Timer
-Timers for one-off or periodic tasks.
+
+Timers for one-off or periodic tasks. Applicable as part of the default mechanism.
 
 Note: If `moc` is invoked with `-no-timer`, the importing will fail.
+
 Note: The resolution of the timers is in the order of the block rate,
       so durations should be chosen well above that. For frequent
       canister wake-ups the heatbeat mechanism should be considered.
```
crusso, line 9:
```suggestion
      canister wake-ups the heartbeat mechanism should be considered.
```

crusso (approved):
But see suggestions

doc/md/base/Timer.md
```diff
@@ -1,11 +1,25 @@
...
 
+Note: Below functionality is active only if the user doesn't explicitly
+      implement `system func timer`.
+
+Note: Timers are _not_ persisted across upgrades. One possible strategy
+      to re-establish timers after an upgrade is to walk stable variables
+      in the `post_upgrade` hook and distill necessary timer information
+      from there.
```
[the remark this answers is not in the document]
roelstorms: Is this correct?
> is to **walk** stable variables

doc/md/base/Timer.md
```diff
@@ -1,11 +1,25 @@
...
+
+Note: Basing security (e.g. access control) on timers is almost always
```
roelstorms, line 19: What's the underlying reason that security controls shouldn't use timers?

I can't come up with a good scenario in which one would want to. Just wondering if this statements mean that timers are not very reliable or that usually if you need timers in a security control you there is a design issue. Agreed that access control would likely not need timers.

ggreif: It is triggered by your slack comment. There are access control methods that use time, e.g. exponential lockout for mistyped passwords etc. These are problematic here, as there is no protection against reentrancy.

doc/md/base/Timer.md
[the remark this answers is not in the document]
ggreif: Yep, I think it is common usage to _walk_ (i.e. traverse) data structures.

doc/md/base/Timer.md
ggreif: I have added a blurb. PTAL.

luc-blaeser:

doc/md/base/Timer.md
roelstorms: I missed this comment back then. LGTM if you still want to merge it.

roelstorms (approved):
LGTM, thanks for improving the docs!

PR #3830, opened 2023-02-23
chore: bump `rustc` edition

ggreif:


nomeata (dismissed):

test/bench/ok/heap-32.drun-run.ok
```diff
@@ -1,5 +1,5 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
-debug.print: (50_227, +91_377_860, 1_434_447_971)
-debug.print: (50_070, +102_586_000, 1_506_480_226)
+debug.print: (50_227, +91_377_860, 1_443_530_187)
+debug.print: (50_070, +102_586_000, 1_517_527_808)
```
nomeata, line 4: Are you Sure that instruction counts in the expected output are a good idea?

ggreif: I am checking this in just to demonstrate that there is something bad is going on. (I also removed the green label.) Before merging such a PR I have to know what caused these cycle increases. Maybe it is a valid thing (e.g. the RTS is now more defensive) or it is some sort of bug in `rustc` that we trigger.

I don't see why cycle counts checked in golden tests are such a bad idea.

nomeata: Hmm, I'd expect it's too much noise, and too much hassle to manually update. But if it doesn't bother you, that's of course fine. (if you want to get the signal without the churn, maybe include it in the performance message on the PR - you'll see it, but don't have to worry about minor changes)

ggreif: So far it was a level of hassle that is acceptable. I have run
```
nix-build -A tests.bench
```
locally and then pasted the diff part into of the error output into
```
patch test/bench/ok/heap-32.drun-run.ok
```
If it gets worse (in terms of insight/hassle ratio) I won't hesitate to nuke these or find another metric.

luc-blaeser:
Obsolete

PR #3831, opened 2023-02-23
Incremental GC Barriers

luc-blaeser:
### Incremental GC PR Stack
The Incremental GC is structured in three PRs to ease review:
1. https://github.com/dfinity/motoko/pull/3837
2. https://github.com/dfinity/motoko/pull/3831 **<-- this PR**
3. https://github.com/dfinity/motoko/pull/3829

# Incremental GC Barriers

Preparation support for write and allocation barriers for the incremental moving (evacuating compacting) GC.

**Write barrier**: All potential pointer writes are passed to the write barrier which performs the write with additional steps to be implemented in the incremental GC:
    * Incremental mark phase: Catch the overwritten pointers to realize incremental snapshot-at-the-beginning marking. 
    * Incremental update phase: If written pointers refer to old evacuated object locations, adjust them to point to the corresponding new forwarded locations.

**Allocation barrier**: The allocation barrier catches all newly created objects that are completely initialized (except the content of blobs). This will serve for the following purposes in the incremental GC:
    * Incremental mark and evacuation phase: Mark the newly allocated object to retain them during the GC that performs snapshot-at-the-beginning marking.
    * Incremental update phase: Update all pointers in the new object to refer to the new forwarded locations.
    * Additional GC increment: To limit memory reclamation latency at a high allocation rate during garbage collection, the barrier performs an additional small GC increment.

***Static optimization***: Compile-time barrier elimination based on a simple conservative analysis of the type of the modified field/array element. Specifically, the barrier is skipped if the type of the written location does not allow pointers (`Bool`, `?Bool`, `Char`, `?Char`, `Nat8`, `?Nat8`, `Nat16`, `?Nat16`, `Int8`, `?Int8`, `Int16`, `?Int16`, `()`, `?()`). (Multi-level optional types can not be ommited as `??null`, `???null` etc. refer to heap objects. Scalar types with >=32 bits can be indirected due to pointer tagging.)

## Runtime Costs

GC benchmark measurements, comparing the number of mutator instructions, average across all benchmark cases, release mode (no sanity checks):

The incremental GC barrier contains the logic of full GC (https://github.com/dfinity/motoko/pull/3837) including object forwarding, however without the allocation GC increment.
The measurements without barriers includes object forwarding to determine the barrier overheads.

Configuration               | Avg. Mutator Instructions
----------------------------|--------------------------
Incremental GC barriers     | 2.06e10
No barriers with forwarding | 1.80e10

**14% runtime overhead on top of forwarding pointers.**

## Testing

Write barrier coverage has been extensively tested by the generational GC and in a separate barrier preparation PR for the incremental GC (https://github.com/dfinity/motoko/pull/3502).
The allocation barrier is tested part of the incremental GC PR (https://github.com/dfinity/motoko/pull/3837).


ulan (commented):
The RTS changes lgtm. I am still reviewing the compile.ml changes.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -67,13 +70,12 @@ unsafe fn generational_gc<M: Memory>(mem: &mut M) {
     update_statistics(&old_limits, new_limits);
     update_strategy(strategy, new_limits);
 
-    #[cfg(debug_assertions)]
-    if !forced_gc {
```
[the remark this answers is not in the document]
luc-blaeser: Thanks for pointing this out. I believe this change is indeed in the wrong PR, should be at top-most PR where I introduced a special test mode (without force GC) for very GC-intense test cases such as `timer.mo` that sporadically fail with a more expensive GC...
I will move it up - or the other change down...

rts/motoko-rts/src/idl.rs
```diff
@@ -94,6 +94,7 @@ unsafe fn parse_fields(buf: *mut Buf, n_types: u32) {
 }
 
 // NB. This function assumes the allocation does not need to survive GC
+// Therefore, no post allocation barrier is applied.
```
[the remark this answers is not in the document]
luc-blaeser: Yes, this invariant should hold. Only in debug mode, I overwrite the evacuated memory with zeros to better detect accesses to stale object memory.

test/run-drun/max-stack-variant.mo
```diff
@@ -2,26 +2,15 @@
 import { errorMessage; performanceCounter; rts_heap_size; rts_max_stack_size; debugPrint; } = "mo:⛔";
```
crusso, line 2: No biggie, but why did you change the test output? I don't think it would have changed output (but maybe it did in previous version).

src/codegen/compile.ml
```diff
@@ -1714,6 +1731,31 @@ module Tagged = struct
   let _branch_typed_with env ty retty branches =
     branch_with env retty (List.filter (fun (tag,c) -> can_have_tag ty tag) branches)
 
+  let allocation_barrier env get_object = 
+    (if !Flags.gc_strategy = Flags.Incremental then
+      (* performance gain by first checking the GC state *)
```
crusso, line 1736: Is that faster than just checking the state within rts_allocation_barrier? Maybe it is in our weird cost model.

src/codegen/compile.ml
```diff
@@ -1714,6 +1731,31 @@ module Tagged = struct
...
+  
+  let write_with_barrier env =
+    let (set_value, get_value) = new_local env "written_value" in
+    let (set_location, get_location) = new_local env "write_location" in
+    set_value ^^ set_location ^^
+    (* performance gain by first checking the GC state *)
+    E.call_import env "rts" "running_gc" ^^
+    G.if0 (
+      get_location ^^ get_value ^^
+      E.call_import env "rts" "write_with_barrier"
```
crusso, line 1753: I wonder if it would be faster to do the check and site in write_with_barrier, avoiding the locals altogether.

src/codegen/compile.ml
```diff
@@ -7511,6 +7596,17 @@ type scope_wrap = G.t -> G.t
...
 
+let potential_pointer typ : bool = 
+  (* must not eliminate nested optional types as they refer to a heap object for ??null, ???null etc. *)
+  let rec can_be_pointer typ nested_optional = 
+    let open Type in
+    match normalize typ with
+    | Mut t -> (can_be_pointer t nested_optional)
+    | Opt t -> (if nested_optional then true else (can_be_pointer t true))
+    | Prim (Null| Bool | Char | Nat8 | Nat16 | Int8 | Int16) | Non | Tup [] -> false
+    | _ -> true in
```
crusso, line 7607:
```suggestion
    
    Type.(match normalize typ with
    | Mut t -> (can_be_pointer t nested_optional)
    | Opt t -> (if nested_optional then true else (can_be_pointer t true))
    | Prim (Null| Bool | Char | Nat8 | Nat16 | Int8 | Int16) | Non | Tup [] -> false
    | _ -> true) in
```

crusso (commented):
Interim comments (not done, but don't want to lose review sofar)

test/run-drun/max-stack-variant.mo
luc-blaeser: I observed a lower limit for the stack with the incremental GC (length `31_744` instead of `35_840`). For the other GCs, it is still the same. Probably because of more stack space with the extra logic and temporary state for the write barrier. As the results differ between the GC versions, I reduced the output.

rts/motoko-rts/src/gc/generational/write_barrier.rs
```diff
@@ -9,22 +9,22 @@ pub static mut REMEMBERED_SET: Option<RememberedSet> = None;
...
     HEAP_BASE = ic::get_aligned_heap_base();
     LAST_HP = ic::LAST_HP;
 }
 
-/// Write barrier to be called AFTER the pointer store, used for generational GC.
+/// Write barrier to be called AFTER the pointer store, used for the generational GC.
 /// `location`: location of modified pointer (address of object field or array element).
 ///
 /// As the barrier is called after the write, `*location` refers to the NEW value.
 /// No effect is the write barrier is deactivated.
```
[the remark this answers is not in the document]
luc-blaeser: Thank you.

rts/motoko-rts/src/gc/incremental/barriers.rs
```diff
@@ -0,0 +1,38 @@
...
+
+/// Allocation barrier to be called after a new object allocation.
+/// The new object needs to be fully initialized, except for the payload of a blob.
+/// Used for the incremental GC.
+#[ic_mem_fn]
+pub fn allocation_barrier<M: Memory>(_mem: &mut M, _new_object: Value) {}
+
+/// Allocation barrier for RTS-internal use by specific places that do not have
+/// a `Memory` reference. Does not perform an allocation increment.
+pub fn slim_allocation_barrier(_new_object: Value) {}
```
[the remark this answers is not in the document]
luc-blaeser: These are only preparation hooks for the next higher PR. They are implemented there. In fact, the `slim_allocation_barrier` is only temporary (intended as a TODO for remembering that special location) and unified in the higher PR.

rts/motoko-rts/src/memory.rs
```diff
@@ -42,7 +44,8 @@ pub unsafe fn alloc_blob<M: Memory>(mem: &mut M, size: Bytes<u32>) -> Value {
     ptr
 }
 
-/// Helper for allocating arrays
+/// Allocate a new array.
+/// Note: After initialization, the post allocation barrier needs to be applied to all mutator objects.
 #[ic_mem_fn]
 pub unsafe fn alloc_array<M: Memory>(mem: &mut M, len: u32) -> Value {
     // Array payload should not be larger than half of the memory
```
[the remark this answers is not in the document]
luc-blaeser: I guess we could remove or relax this restriction. There is no such conceptual limit in the other GCs. I will keep it in mind for a refactoring of the RTS (especially if we split the RTS builds depending on the GCs).

src/codegen/compile.ml
```diff
@@ -1597,25 +1610,29 @@ module Tagged = struct
...
+      else G.nop)
     else G.nop)
 
   let check_forwarding_for_store env typ =
-    let (set_value, get_value, _) = new_local_ env typ "value" in
-    set_value ^^ check_forwarding env false ^^ get_value
+    (if !Flags.gc_strategy = Flags.Incremental then
+      let (set_value, get_value, _) = new_local_ env typ "value" in
+      set_value ^^ check_forwarding env false ^^ get_value
+    else G.nop)
```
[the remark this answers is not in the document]
luc-blaeser: Indeed, this is too complex code and would benefit from RTS-implementation. I am interested in the two-version specialized implementation. Thank you very much.

src/codegen/compile.ml
luc-blaeser: Yes, unfortunately, a relatively high difference (but need to remeasure to get the detailed number). I observed that the RTS-calls are much more expensive if the `Memory` trait is used (no extra cost if it is declared but not used). Also `not inlining` functions at the right places shows noticeable differences. I hope it is improving with our future cost model.

src/codegen/compile.ml
luc-blaeser: Unfortunately, also this is a weird optimization, that I believe gave a high performance improvement. I can remeasure the performance difference and hopefully, we could overcome such in future with a better cost model (or some linker redesign).

rts/motoko-rts/src/memory.rs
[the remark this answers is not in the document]
crusso: Perhaps @nomeata remembers something about this?

src/codegen/compile.ml
```diff
@@ -3940,13 +4012,18 @@ module Arr = struct
...
+      (* Recompute the array element address because the array may be moved
+         by a GC increment triggered by the tabulate function. *)
+      (* Write array element *)
+      get_r ^^
       get_i ^^
-      compile_add_const 1l ^^
-      set_i
+      idx env ^^ (* Resolves object forwarding *)
+      get_value ^^
+      store_ptr
```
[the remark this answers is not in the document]
luc-blaeser: Indeed, it is the price for the allocation increments (to keep up with high allocation rates). If we would have a generational collection combined with the incremental one (with even higher complexity), we could eliminate the allocation increment and use this optimized logic. Maybe there are other possibilities for keeping this optimization?

For this PR, I believe I should consider to even remove the tabulate logic from the `compile.ml` since it is no longer optimized and could just as well be implemented in the Motoko in the compiler prelude.

src/codegen/compile.ml
crusso: We have a new cost model coming?

luc-blaeser: I think so, the runtime system team is working on this. But not sure about the details and timeline...

src/codegen/compile.ml
```diff
@@ -8994,14 +9123,30 @@ let rec compile_lexp (env : E.t) ae lexp =
     store_ptr ^^
     get_field ^^
     compile_add_const ptr_unskew ^^
-    E.call_import env "rts" "write_barrier"
+    E.call_import env "rts" "post_write_barrier"
+  | IdxLE (e1, e2), Flags.Incremental when potential_pointer (Arr.element_type env e1.note.Note.typ) ->
+    let (set_array, get_array) = new_local env "array" in
```
crusso, line 9128: I'd add a todo to the new cases here to try to optimized them in future. I worry that the use of locals will hurt us and wonder if it might be better to push the value on the stack in the preamble and then call a function to do the postamble. But probably not worth investigating now - just leave a TODO, possibily linking a GH issue.

crusso (commented):
Some more comments.

src/codegen/compile.ml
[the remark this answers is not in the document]
crusso: > For this PR, I believe I should consider to even remove the tabulate logic from the `compile.ml` since it is no longer optimized and could just as well be implemented in Motoko in the compiler prelude.

No, I don't think that's possible without adding some unsafe array primitives. Tabulate is used to initialize an  *immutable* array - that's why its implemented in the RTS, not motoko. Base has an implementation for mutable arrays (in Motoko) but Timo was actually asking if we could provide a quicker prim. 

This one should at least still be avoiding the bounds check on every write, so at least still a small win.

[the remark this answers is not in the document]
luc-blaeser: True, I did not consider that aspect (and would have noticed during implementation). Thanks for pointing this out.

src/codegen/compile.ml
```diff
@@ -1573,6 +1582,7 @@ module Tagged = struct
   let tag_field = 0l
   let forwarding_pointer_field = 1l
 
+  (* Note: Post allocation barrier must be applied after initialization *)
   let alloc env size tag =
```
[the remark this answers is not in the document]
luc-blaeser: Good idea. I will try this out and measure the difference.

src/prelude/internals.mo
```diff
@@ -519,15 +519,8 @@ func @timer_helper() : async () {
   func Array_init<T>(len : Nat,  x : T) : [var T] {
     (prim "Array.init" : <T>(Nat, T) -> [var T])<T>(len, x)
   };
-  let now = (prim "time" : () -> Nat64)();
```
luc-blaeser, line 522: This is weird because this file has no difference to master and to the base branch `luc/forwaring_pointers`.
Maybe a GitHub compare issue?

src/codegen/compile.ml
luc-blaeser: Update: I removed the allocation increments and postponed the time to the scheduled GC increments. Thanks @ulan for this suggestion. This makes many things easier: We do not have to care about temporarily resolved pointers during assignments and other operations between the usual regularly scheduled GC calls. As a result, also the barriers are not cheap again, without needing local variables.

src/codegen/compile.ml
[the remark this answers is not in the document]
luc-blaeser: Update: I removed the allocation increments and postponed them to next regularly scheduled GC call. With this change, we can again use the optimized tabulate function. Thanks to @ulan for this suggestion.

test/run-drun/max-stack-variant.mo
luc-blaeser: Update: I analyzed it further: It is the **allocation barrier** call that consumes the extra stack space (the call to the RTS), and reduces the limit for the incremental GC.

src/codegen/compile.ml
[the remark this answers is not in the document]
luc-blaeser: Thank you very much for the advice. I implemented this.

src/codegen/compile.ml
luc-blaeser: I remeasured this optimization (to have specific results): Together with the write barrier, it is a **>10%** runtime performance gain (total instruction count, average across benchmark cases). For the allocation barrier, it gives a **4%** benefit, and for the write barrier >6%.

src/codegen/compile.ml
luc-blaeser: **Update**: I remeasured this optimization: The runtime improvement for the write barrier alone is **>6%**. Together with the allocation barrier, it is a **>10%** improvement.

rts/motoko-rts/src/text_iter.rs
```diff
@@ -60,13 +63,13 @@ pub unsafe fn text_iter<M: Memory>(mem: &mut M, text: Value) -> Value {
...
-    // Initialize blob field
-    array.set_pointer(
+    // Initialize blob field, no pre-update barrier, but post-update barrier.
+    array.initialize(
         ITER_BLOB_IDX,
         find_leaf(mem, text, todo_addr as *mut _),
         mem,
     );
-
+    allocation_barrier(mem, iter);
```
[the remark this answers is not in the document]
luc-blaeser: This is very good idea that I implemented. It saved the weird optimization for testing `gc_running` at allocation barrier call in the compiler (which originally gave a performance gain of 4%). Thanks for this suggestion.

src/codegen/compile.ml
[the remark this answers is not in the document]
luc-blaeser: I implemented this. It even reduced cycle count a little bit (around 1%), not yet sure why...

src/codegen/compile.ml
```diff
@@ -8310,13 +8420,15 @@ module AllocHow = struct
...
+  let how_of_ae ae : allocHow = 
+    let locations = M.map (fun (l, _) -> l) ae.VarEnv.vars in 
+    M.map (function
     | VarEnv.Const _        -> (Const : how)
     | VarEnv.HeapStatic _   -> StoreStatic
     | VarEnv.HeapInd _      -> StoreHeap
     | VarEnv.Local (sr, _)  -> LocalMut sr (* conservatively assume mutable *)
     | VarEnv.PublicMethod _ -> LocalMut SR.Vanilla
-    ) ae.VarEnv.vars
+    ) locations
```
crusso, line 8431:
```suggestion
    ) ae.VarEnv.vars
```
(continuing above suggestion)

src/codegen/compile.ml
```diff
@@ -8335,25 +8447,25 @@ module AllocHow = struct
...
+  let add_local env ae how name typ : VarEnv.t * G.t =
     match M.find name how with
     | (Const : how) -> (ae, G.nop)
     | LocalImmut sr | LocalMut sr ->
-      let (ae1, i) = VarEnv.add_direct_local env ae name sr in
+      let (ae1, i) = VarEnv.add_direct_local env ae name sr typ in
       (ae1, G.nop)
     | StoreHeap ->
-      let (ae1, i) = VarEnv.add_local_with_heap_ind env ae name in
+      let (ae1, i) = VarEnv.add_local_with_heap_ind env ae name typ in
```
crusso, line 8457:
```suggestion
      let ae1, i = VarEnv.add_local_with_heap_ind env ae name typ in
```

src/codegen/compile.ml
```diff
@@ -10106,8 +10230,8 @@ and compile_exp_with_hint (env : E.t) ae sr_hint exp =
        G.i Unreachable (* We should always exit using the branch_code *)
     )
   (* Async-wait lowering support features *)
-  | DeclareE (name, _, e) ->
-    let (ae1, i) = VarEnv.add_local_with_heap_ind env ae name in
+  | DeclareE (name, typ, e) ->
+    let (ae1, i) = VarEnv.add_local_with_heap_ind env ae name typ in
```
crusso, line 10234:
```suggestion
    let ae1, i = VarEnv.add_local_with_heap_ind env ae name typ in
```

crusso (commented):
Added some comments

src/codegen/compile.ml
```diff
@@ -7540,6 +7616,17 @@ type scope_wrap = G.t -> G.t
...
+let potential_pointer typ : bool = 
+  (* must not eliminate nested optional types as they refer to a heap object for ??null, ???null etc. *)
+  let rec can_be_pointer typ nested_optional = 
+    
+    Type.(match normalize typ with
+    | Mut t -> (can_be_pointer t nested_optional)
+    | Opt t -> (if nested_optional then true else (can_be_pointer t true))
+    | Prim (Null| Bool | Char | Nat8 | Nat16 | Int8 | Int16) | Non | Tup [] -> false
+    | _ -> true) in
+  can_be_pointer typ false
```
crusso, line 7628:
```suggestion
let rec can_be_pointer typ nested_optional =     
  Type.(match normalize typ with
  | Mut t -> (can_be_pointer t nested_optional)  
  | Opt t -> (if nested_optional then true else (can_be_pointer t true))
  | Prim (Null| Bool | Char | Nat8 | Nat16 | Int8 | Int16) | Non | Tup [] -> false
  | _ -> true) 
  
let potential_pointer typ : bool = 
  (* must not eliminate nested optional types as they refer to a heap object for ??null, ???null etc. *)
  can_be_pointer typ false
```
Might be a little faster, if less clear, depending on how OCaml optimizes local functions.

crusso (approved):
Looks good to me. Now for the main event....

src/codegen/compile.ml
```diff
@@ -10455,20 +10579,20 @@ and compile_dec env pre_ae how v2en dec : VarEnv.t * G.t * (VarEnv.t -> scope_wr
       unmodified
     )
 
-  | VarD (name, _, e) ->
+  | VarD (name, typ, e) ->
```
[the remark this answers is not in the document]
luc-blaeser: Good that you mention this. I indeed assumed otherwise - and fixed it. Although, luckily, it does not matter for barrier optimization since it only considers the type without distinguishing Mut or not (because it knows it is already a write at those locations). But it would be a hard-to-find bug and pitfall if the type would be used for other purposes in future.

src/codegen/compile.ml
```diff
@@ -10455,20 +10579,20 @@ and compile_dec env pre_ae how v2en dec : VarEnv.t * G.t * (VarEnv.t -> scope_wr
...
     ( pre_ae1,
       alloc_code,
       (fun ae -> let pre_code, sr, code = Var.set_val env ae name in
                  pre_code ^^ compile_exp_as env ae sr e ^^ code),
       unmodified
     )
 
-  | RefD (name, _, { it = DotLE (e, n); _ }) ->
-    let pre_ae1, alloc_code = AllocHow.add_local_for_alias env pre_ae how name in
+  | RefD (name, typ, { it = DotLE (e, n); _ }) ->
```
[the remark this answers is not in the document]
luc-blaeser: I also believe it is the target type of the aliased field. For the write barrier optimization, it would be okay. However, I wonder whether I should register a different type for `AllocHow`/`VarEnv`. Neither `Mut` nor the plain target type seem to be a very good choice.

src/codegen/compile.ml
```diff
@@ -9017,7 +9129,7 @@ let rec compile_lexp (env : E.t) ae lexp : G.t * SR.t * G.t =
   (fun (code, sr, fill_code) -> G.(with_region lexp.at code, sr, with_region lexp.at fill_code)) @@
   match lexp.it, !Flags.gc_strategy with
   | VarLE var, _ -> Var.set_val env ae var
-  | IdxLE (e1, e2), Flags.Generational ->
+  | IdxLE (e1, e2), Flags.Generational when potential_pointer (Arr.element_type env e1.note.Note.typ) ->
```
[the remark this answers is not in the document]
luc-blaeser: I think we could assume that the information is available, because it is backend code and has passed the type checker.

src/codegen/compile.ml
```diff
@@ -8310,13 +8420,15 @@ module AllocHow = struct
 
   (* find the allocHow for the variables currently in scope *)
   (* we assume things are mutable, as we do not know better here *)
-  let how_of_ae ae : allocHow = M.map (function
+  let how_of_ae ae : allocHow = 
+    let locations = M.map (fun (l, _) -> l) ae.VarEnv.vars in 
+    M.map (function
```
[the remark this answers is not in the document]
luc-blaeser: I improved this. Thanks for the suggestion.

src/codegen/compile.ml
luc-blaeser: I improved this. Thank you.

src/codegen/compile.ml
luc-blaeser: Thank you.

ulan (approved):
Lgtm! I reviewed Rust code and mostly skimmed the compiler.ml changes

crusso (approved):
Looks like mostly white space changes since last reviewed. LGTM.

PR #3836, opened 2023-02-24
chore: update docs and changelog in prep for 0.8.3

crusso:


Changelog.md
```diff
@@ -1,10 +1,35 @@
...
+* motoko (`moc`)
+
+  * new 'let-else' construct for handling pattern-match failure.
+
+    The let declaration `let <pat> = <exp> else <block-or-exp>` has type `T` and declares the bindings in `<pat>` provided:
+    -   `<exp>` has type `T`,
+    -   `<pat>` has type `T`, and
+    -   `<block-or-exp>` has type `None`.
+
+    The declaration `let <pat> = <exp> else <block-or-exp>` evaluates `<exp>` to a result `r`.
```
ggreif, line 13: Isn't this too detailed for a changelog entry? (It would be adequate for the manual.) Something like "New feature `let-else` allows to take a different path in control flow (instead of trapping) when a pattern match failure happens". (You decide, but think of the target audience.)

ggreif (approved):

PR #3837, opened 2023-02-24
Incremental GC

luc-blaeser:
### Incremental GC PR Stack
The Incremental GC is structured in three PRs to ease review:
1. https://github.com/dfinity/motoko/pull/3837 **<-- this PR**
2. https://github.com/dfinity/motoko/pull/3831
3. https://github.com/dfinity/motoko/pull/3829

# Incremental GC

Incremental evacuating-compacting garbage collector.

**Objective**: Scalable memory management that allows full heap usage.

**Properties**:
* All GC pauses have bounded short time.
* Full-heap snapshot-at-the-beginning marking.
* Focus on reclaiming high-garbage partitions.
* Compacting heap space with partition evacuations.
* Incremental copying enabled by forwarding pointers.
* Using **mark bitmaps** instead of a mark bit in the object headers.
* Limiting number of evacuations on memory shortage.

## Design

The incremental GC distributes its workload across multiple steps, called increments, that each pause the mutator (user's program) for only a limited amount of time. As a result, the GC appears to run concurrently (although not parallel) to the mutator and thus allows scalable heap usage, where the GC work fits within the instruction-limited IC messages.

Similar to the recent Java Shenandoah GC [1], the incremental GC organizes the heap in equally-sized partitions and selects high-garbage partitions for compaction by using incremental evacuation and the Brooks forwarding pointer technique [2].

The GC runs in three phases:
1. **Incremental Mark**: The GC performs full heap incremental tri-color-marking with snapshot-at-the-beginning consistency. For this purpose, write barriers intercept mutator pointer overwrites between GC mark increments. The target object of an overwritten pointer is thereby marked. Concurrent new object allocations are also conservatively marked. To remember the mark state per object, the GC uses partition-associated mark bitmaps that are temporarily allocated during a GC run. The phase additionally needs a mark stack that is a growable linked table list in the heap that can be recycled as garbage during the active GC run. Full heap marking has the advantage that it can also deal with arbitrarily large cyclic garbage, even if spread across multiple partitions. As a side activity, the mark phase also maintains the bookkeeping of the amount of live data per partition. Conservative snapshot-at-the-beginning marking and retaining new allocations is necessary because the WASM call stack cannot be inspected for the root set collection. Therefore, the mark phase must also only start on an empty call stack.

2. **Incremental Evacuation**: The GC prioritizes partitions with a larger amount of garbage for evacuation based on the available free space. It also requires a defined minimum amount of garbage for a partition to be evacuated. Subsequently, marked objects inside the selected partitions are evacuated to free partitions and thereby compacted. To allow incremental object moving and incremental updating of pointers, each object carries a redirection information in its header, which is a forwarding pointer, also called Brooks pointer. For non-moved objects, the forwarding pointer reflexively points back to the object itself, while for moved objects, the forwarding pointer refers to the new object location. Each object access and equality check has to be redirected via this forwarding pointer. During this phase, evacuated partitions are still retained and the original locations of evacuated objects are forwarded to their corresponding new object locations. Therefore, the mutator can continue to use old incoming pointers to evacuated objects.

3. **Incremental Updates**: All pointers to moved objects have to be updated before free space can be reclaimed. For this purpose, the GC performs a full-heap scan and updates all pointers in alive objects to their forwarded address. As mutator may perform concurrent pointer writes behind the update scan line, a write barrier catches such pointer writes and resolves them to the forwarded locations. The same applies to new object allocations that may have old pointer values in their initialized state (e.g. originating from the call stack). Once this phase is completed, all evacuated partitions are freed and can later be reused for new object allocations. At the same time, the GC also frees the mark bitmaps stored in temporary partitions. The update phase can only be completed when the call stack is empty, since the GC does not access the WASM stack. No remembered sets are maintained for tracking incoming pointers to partitions.

**Humongous objects**:
* Objects with a size larger than a partition require special handling: A sufficient amount of contiguous free partitions is searched and reserved for a large object. Large objects are not moved by the GC. Once they have become garbage (not marked by the GC), their hosting partitions are immediately freed. Both external and internal fragmentation can only occur for huge objects. Partitions storing large objects do not require a mark bitmap during the GC.

**Increment limit**:
* The GC maintains a synthetic deterministic clock by counting work steps, such as marking an object, copying a word, or updating a pointer. The clock serves for limiting the duration of a GC increment. The GC increment is stopped whenever the limit is reached, such that the GC later resumes its work in a new increment. To also keep the limit on large objects, large arrays are marked and updated in incremental slices. Moreover, huge objects are never moved. 
For simplicity, the GC increment is only triggered at the compiler-instrumented scheduling points when the call stack is empty. The increment limit is increased depending on the amount of concurrent allocations, to reduce the reclamation latency on a high allocation rate during garbage collection.

**Memory shortage**
* If memory is scarce during garbage collection, the GC limits the amount of evacuations to available free space of free partitions. This is to prevent the GC to run out of memory while copying alive objects to new partitions.

## Configuration

* **Partition size**: 32 MB.

* **Increment limit**: Regular increment bounded to 3,500,000 steps (approximately 600 million instructions). Each allocation during GC increases the next scheduled GC increment by 20 additional steps.

* **Survival threshold**: If 85% of a partition space is alive (marked), the partition is not evacuated.

* **GC start**: Scheduled when the growth (new allocations since the last GC run) account for more than 65% of the heap size. When passing the critical limit of 3.25GB (on the 4GB heap size), the GC is already started when the growth exceeds 1% of the heap size.

## Measurement

The following results have been measured on the GC benchmark with `dfx` 0.13.1. The `Copying`, `Compacting`, and `Generational` GC are based on the original runtime system ***without*** the forwarding pointer header extension. `No` denotes the disabled GC based on the runtime system ***with*** the forwarding pointer header extension. 

### Scalability

**Summary**: The incremental GC allows full 4GB heap usage without that it exceeds the message instruction limit. It therefore scales much higher than the existing stop-and-go GCs and naturally also higher than without GC.

Average amount of allocations for the benchmark limit cases, until reaching a limit (instruction limit, heap limit, `dfx` cycles limit). Rounded to two significant figures.

| **Incremental**   | **150e6**               |

3x higher than the other GCs and also than no GC.

Currently, the following limit benchmark cases do not reach the 4GB heap maximum due to GC-independent reasons:
* `buffer` applies exponential array list growth where the copying to the larger array exceeds the instruction limit.
* `rb-tree`, `trie-map`, and `btree-map` are such garbage-intense that they run out of `dfx` cycles or suffer from a sudden `dfx` network connection interruption.

### GC Pauses

Longest GC pause, maximum of all benchmark cases:

| GC                | Longest GC Pause          |
| **Incremental**   | **0.712e9**               |
| Generational      | 1.19e9                    |
| Compacting        | 8.41e9                    |
| Copying           | 5.90e9                    |

Shorter than all the other GCs.

### Performance

| **Incremental**   | **1.85e10**             | 
| Generational      | 1.91e10                 | 
| Compacting        | 2.20e10                 | 
| Copying           | 2.05e10                 | 

Faster than all the other GCs.

| **Incremental**   | **94.6%**                |
| Generational      | 85.4%                    |
| Compacting        | 75.8%                    |
| Copying           | 78.7%                    |

Higher than the other GCs.

### Memory Size

Occupied heap size at the end of each benchmark case, average across all cases:

| GC                | Avg. Final Heap Occupation |
| ----------------- | -------------------------- |
| **Incremental**   | **176 MB**                 |
| No                | 497 MB                     |
| Generational      | 156 MB                     |
| Compacting        | 144 MB                     |
| Copying           | 144 MB                     |

Up to 22% higher than the other GCs.

| **Incremental**   | **296 MB**              |
| No                | 499 MB                  |
| Generational      | 191 MB                  |
| Compacting        | 188 MB                  |
| Copying           | 271 MB                  |

9% higher than the copying GC. 57% higher (worse) than the generational and the compacting GC.

## Overheads

* **Write barrier**: 
    - During the mark and evacuation phase: Marking the target of overwritten pointers.
    - During the update phase: Resolving forwarding of written pointers.
* **Allocation barrier**:
    - During the mark and evacuation phase: Marking new allocated objects.
    - During the update phase: Resolve pointer forwarding in initialized objects.
* **Pointer forwarding**:
    - Indirect each object access and equality check via the forwarding pointer.

Runtime costs for the barrier are reported in https://github.com/dfinity/motoko/pull/3831.
Runtime costs for the forwarding pointers are reported in https://github.com/dfinity/motoko/pull/3829.

## Testing

1. RTS unit tests

    In Motoko repo folder `rts`:
    ```
    make test
    ```

2. Motoko test cases

    ```
    export EXTRA_MOC_ARGS="--sanity-checks --incremental-gc"
    make
    ```

3. GC Benchmark cases

    In `gcbench` repo: 
    ```
    ./measure-all.sh
    ```

4. Extensive memory sanity checks

    Adjust `Cargo.toml` in `rts/motoko-rts` folder:
    ```
    default = ["ic", "memory-check"]
    ```

    Run selected benchmark and test cases. Some of the tests will exceed the instruction limit due to the expensive checks.

## Extension to 64-Bit Heaps

The design partition information would need to be adjusted to store the partition information dynamically instead of a static allocation. For example, the information could be stored in a reserved space at the beginning of a partition (except if the partition has static data or serves as an extension for hosting a huge object). Apart from that, the GC should be portable and scalable without significant design changes on 64-bit memory.

## Design Alternatives

* **Free list**: See the prototype in https://github.com/dfinity/motoko/pull/3678. The free-list-based incremental GC shows higher reclamation latency, slower performance (free list selection), and potentially higher external fragmentation (no compaction, just free neighbor merging).
* **Mark bit in object header**: See implementation in https://github.com/dfinity/motoko/pull/3756. Storing the mark bit in the object header instead of using a mark bitmap saves memory space, but is more expensive for scanning sparsely marked partitions. Moreover, it increases the amount of dirty pages.
* **Remembered set**: Inter-partition pointers could be stored in remembered set to allow more selective and faster pointer updates. Moreover, the write barrier would become more expensive to detect and store relevant pointers in the remembered set. Also, the remembered set would occupy additional memory.
* **Allocation increments**: On high allocation rate, the GC could also perform a short GC increment during an allocation. This design is however more complicated as it forbids that the compiler can store low-level pointers on the stack while performing an allocation (e.g. during assignments or array tabulate). It is also slower than the current solution where allocation increments are postponed to next regularly scheduled GC increment, running when the call stack is empty.
* **Special incremental GC**: Analzyed in PR https://github.com/dfinity/motoko/pull/3894. An incremental GC based on a central object table that allows easy object movement and incremental compaction. Compared to this PR, the special GC has 35% worse runtime performance. 
* **Combining tag and forwarding pointer**: https://github.com/dfinity/motoko/pull/3904. This seems to be less efficient than the Brooks pointer technique with a runtime performance degrade of 27.5%, while only offering a small memory saving of around 2%.

## References

[1] C. H. Flood, R. Kennke, A. Dinn, A. Haley, and R. Westrelin. Shenandoah. An Open-Source Concurrent Compacting Garbage Collector for OpenJDK. Intl. Conference on Principles and Practices of Programming on the Java Platform: Virtual Machines, Languages, and Tools, PPPJ'16, Lugano, Switzerland, August 2016.

[2] R. A. Brooks. Trading Data Space for Reduced Time and Code Space in Real-Time Garbage Collection on Stock Hardware. ACM Symposium on LISP and Functional Programming, LFP'84, New York, NY, USA, 1984.


ulan (commented):
I am half-way through. Looking good so far. Publishing a snapshot of my comments.

rts/motoko-rts/src/gc/incremental.rs
```diff
@@ -1,18 +1,439 @@
...
+        debug_assert!(state.partitioned_heap.is_initialized());
+        IncrementalGC { mem, state, time }
+    }
+
+    /// Special GC increment invoked when the call stack is guaranteed to be empty.
+    /// As the GC cannot scan or use write barriers on the call stack, we need to ensure:
+    /// * The mark phase can only be started on an empty call stack.
+    /// * The update phase can only be completed on an empty call stack.
+    pub unsafe fn empty_call_stack_increment(&mut self, roots: Roots) {
+        if self.pausing() {
```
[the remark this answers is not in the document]
luc-blaeser: Yes, thanks, this would make it safer (especially in combination with the previous review finding).

rts/motoko-rts/src/gc/incremental.rs
```diff
@@ -1,18 +1,439 @@
...
+        }
+    }
+
+    unsafe fn pausing(&mut self) -> bool {
+        self.state.phase == Phase::Pause
+    }
+
+    unsafe fn increment(&mut self) {
+        match self.state.phase {
+            Phase::Pause | Phase::Stop => {}
```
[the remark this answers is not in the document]
luc-blaeser: These should be unreachable. I will improve the logic that it is clear. Thanks for noticing this.

rts/motoko-rts/src/gc/incremental/mark_bitmap.rs
```diff
@@ -0,0 +1,174 @@
...
+
+        // Outer loop iterates the 64-bit words.
+        loop {
+            // Examine the least significant bit(s) in the current word.
+            if self.current_word != 0 {
+                let shift = self.current_word.trailing_zeros() as usize;
+                self.current_word >>= shift;
+                self.current_word >>= 1;
+                let bit_index = self.next_bit_index + shift;
+                self.next_bit_index = bit_index + 1;
```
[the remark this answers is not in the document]
luc-blaeser: I actually adopted the code from the mark-and-compact GC. I will look more thoroughly into this (also the other aspect with the bug on the bitmap end). Also to check this with the other GC.

rts/motoko-rts/src/gc/incremental/mark_bitmap.rs
```diff
@@ -0,0 +1,174 @@
...
+    /// Start address of the mark bitmap.
+    bitmap_pointer: *mut u8,
+    /// Index of next bit to continue iteration in the bitmap.
+    next_bit_index: usize,
+    /// Current 64-bit word in the bitmap that we are iterating.
+    /// Optimization: Reading in 64-bit chunks to check as many bits as
+    /// possible with a single `word != 0`.
+    current_word: u64,
+    /// Number of leading bits that are initially zeroed in `current_word`.
+    leading_zeros: usize,
```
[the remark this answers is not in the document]
luc-blaeser: Thanks a lot for this feedback and potential bug. I will check this, and improve the comments. As I adopted the logic from the other mark-and-compact GC, I will analyze this for other GC too.

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -0,0 +1,682 @@
...
+    gc::incremental::mark_bitmap::BITMAP_ITERATION_END, memory::Memory, rts_trap_with, types::*,
+};
+
+use super::{
+    mark_bitmap::{BitmapIterator, MarkBitmap, BITMAP_SIZE, DEFAULT_MARK_BITMAP},
+    time::BoundedTime,
+};
+
+/// Size of each parition.
+pub const PARTITION_SIZE: usize = 32 * 1024 * 1024;
```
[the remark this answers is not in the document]
luc-blaeser: Yes, very good point. I will do this.

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -0,0 +1,682 @@
...
+
+/// Total number of partitions in the memory.
+/// For simplicity, the last partition is left unused, to avoid a numeric overflow when
+/// computing the end address of the last partition.
+const MAX_PARTITIONS: usize = usize::MAX / PARTITION_SIZE;
+
+/// Partitions are only evacuated if the space occupation of alive object in the partition
+/// is greater than this threshold.
+pub const SURVIVAL_RATE_THRESHOLD: f64 = 0.85;
```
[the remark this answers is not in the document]
luc-blaeser: I agree. Will provide a rationale.

rts/motoko-rts/src/gc/incremental.rs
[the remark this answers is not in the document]
luc-blaeser: I added the assertion.

rts/motoko-rts/src/types.rs
```diff
@@ -326,7 +329,7 @@ impl Value {
     /// Get the pointer as `Array` using forwarding. In debug mode panics if the value is not a pointer or the
     /// pointed object is not an `Array`.
     pub unsafe fn as_array(self) -> *mut Array {
-        debug_assert_eq!(self.tag(), TAG_ARRAY);
+        debug_assert!(self.tag() == TAG_ARRAY || self.tag() >= TAG_ARRAY_SLICE_MIN);
```
ulan, line 332: Since we are relaxing this condition, we need to be careful to adjust all places that assume that arrays have `TAG_ARRAY` tags. @crusso: please keep this in mind when reviewing code.

ulan (commented):
I reviewed most of the PR. Looking good so far -- only small nits and comments.

I need a bit more time to finish the review and think this through.

rts/motoko-rts/src/gc/incremental/mark_bitmap.rs
```diff
@@ -0,0 +1,180 @@
...
+
+/// Adopted and adjusted from `mark_compact/bitmap.rs`.
+/// The iterator separates advancing `next()` from inspection `current_marked_offset()`
+/// to better support the incremental evacuation and update GC increments.
+pub struct BitmapIterator {
+    /// Start address of the mark bitmap.
+    bitmap_pointer: *mut u8,
+    /// Index of next bit to continue iteration in the bitmap.
+    /// Invariant during (initialzed and unfinished):
+    /// `lsb(current_word) == (next_bit_index - 1) % size_of::<u64>()`.
```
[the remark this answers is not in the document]
luc-blaeser: Oh yes, indeed. Thank you for noticing this mistake, I will fix it.

rts/motoko-rts/src/memory.rs
```diff
@@ -26,7 +26,12 @@ use motoko_rts_macros::ic_mem_fn;
 ///
 /// This function does not take any `Memory` arguments can be used by the generated code.
 pub trait Memory {
+    // General allocator working for all GC variants.
     unsafe fn alloc_words(&mut self, n: Words<u32>) -> Value;
+    // Optimized version, only to be used by non-incremental GC runtimes.
+    unsafe fn linear_alloc_words(&mut self, n: Words<u32>) -> Value;
+    // Grow the allocated memory size to at least the address of `ptr`.
+    unsafe fn grow_memory(&mut self, ptr: u64);
```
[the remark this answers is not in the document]
luc-blaeser: The `PartitionedHeap` also uses the `grow_memory` in certain cases. As `PartitionedHeap` has no direct dependency on `memory::ic` (for unit testing purposes), I also abstracted in under `Memory`. I also thought about the alternative of introducing a separate trait for memory growth but this also looks unnatural. Also passing the function separately would add more complexity. I wonder whether there is a nicer solution?

rts/motoko-rts/src/gc/incremental/time.rs
```diff
@@ -0,0 +1,24 @@
...
+    pub fn new(limit: usize) -> BoundedTime {
+        BoundedTime { steps: 0, limit }
+    }
+
+    pub fn tick(&mut self) {
+        self.steps = usize::saturating_add(self.steps, 1);
+    }
+
+    pub fn advance(&mut self, amount: usize) {
+        self.steps = usize::saturating_add(self.steps, amount);
```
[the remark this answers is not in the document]
luc-blaeser: Good idea. I added this check.

rts/motoko-rts/src/gc/incremental/mark_stack.rs
```diff
@@ -0,0 +1,115 @@
...
+        (*table).previous = previous;
+        (*table).next = null_mut();
+        if previous != null_mut() {
+            (*previous).next = table;
+        }
+        table
+    }
+
+    #[cfg(debug_assertions)]
+    pub unsafe fn assert_is_garbage(&self) {
```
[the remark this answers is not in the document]
luc-blaeser: Thank you for noticing this. This function indeed lacked some check (introduced with the change to the mark bitmaps). I fixed this, renamed the function and added some explanation comment.

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -0,0 +1,692 @@
...
+                && index + count < MAX_PARTITIONS
+                && self.get_partition(index + count).is_completely_free()
+            {
+                count += 1;
+            }
+            if count == number_of_partitions {
+                return index;
+            }
+        }
+        rts_trap_with("Cannot grow memory");
```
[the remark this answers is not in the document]
luc-blaeser: I thought so as well. The only problem are the integration test cases which are run with different GC modes and have an expected error message. I kept this (unspecific) description to obtain the same error as in other GCs when overly large objects are created and/or the heap runs out of memory.

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -0,0 +1,692 @@
...
+            let partition = self.get_partition(partition_index);
+            if partition.has_dynamic_space() && !partition.has_large_content() {
+                let bitmap_address = self.allocate_bitmap(mem);
+                self.mutable_partition(partition_index)
+                    .bitmap
+                    .assign(bitmap_address);
+                time.advance(Bytes(BITMAP_SIZE as u32).to_words().as_usize());
+            }
+        }
+        if self.allocation_partition().dynamic_size > PARTITION_SIZE / 2 {
```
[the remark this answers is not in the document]
luc-blaeser: I remeasured and saw the difference is minor. Therefore, I now simplified this by removing this logic. 

Measurements, average in GC benchmark, threshold means the minimum dynamic size to switch allocation partition:
| Threshold | Memory Size | Heap Size | Total Instructions | Comment |
| ----------| --------------| ---------- | -----------------| -------------|
| 0 |	304 MB | 165 MB | 	1.95e10 | always start new allocation partition
| PARTITION_SIZE * 0.25 |	294 MB | 162 MB	| 1.94e10 | |
| PARTITION_SIZE * 0.5 |	294 MB | 164 MB	| 1.92e10 | |
| PARTITION_SIZE * 0.75 | 292 MB | 169 MB	| 1.90e10 | |
| **PARTITION_SIZE** | **296 MB** | **176 MB** |	**1.86e10** | always keep the allocation partition |

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -0,0 +1,692 @@
...
+}
+
+/// Partitioned heap used by the incremental GC.
+pub struct PartitionedHeap {
+    partitions: [Partition; MAX_PARTITIONS],
+    heap_base: usize,
+    allocation_index: usize, // Index of the partition currently used for allocations.
+    evacuating: bool,
+    reclaimed: u64,
+    bitmap_pointer: usize, // Allocation pointer for mark bitmaps.
```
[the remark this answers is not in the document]
luc-blaeser: I improved this. This was indeed confusion (especially since the same variable name was used in `BitmapIterator` with the other meaning).

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -0,0 +1,692 @@
...
+        &self.partitions[index]
+    }
+
+    fn mutable_partition(&mut self, index: usize) -> &mut Partition {
+        &mut self.partitions[index]
+    }
+
+    unsafe fn allocate_temporary_partition(&mut self) -> &mut Partition {
+        for partition in &mut self.partitions {
+            if partition.free && partition.is_completely_free() {
```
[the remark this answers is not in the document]
luc-blaeser: Thank you, I removed the redundant condition.

test/run-drun/destabilization-crash/destabilization-crash.mo
```diff
@@ -23,7 +23,8 @@ actor a {
        c -= 1;
      };
      if (Prim.rts_heap_size() < mb * 1024 * 1024) {
-      Prim.debugPrint(debug_show({heap_MB = Prim.rts_heap_size()/1024/1024}));
+      let partitionSizeMB = 32;
```
crusso, line 26: Why the change?

test/run-drun/throw-on-async-failure.mo
```diff
@@ -1,3 +1,4 @@
+//MOC-NO-FORCE-GC
```
crusso, line 1: What is this new flag necessary?

test/run.sh
```diff
@@ -252,7 +249,13 @@ do
     # extra flags (allow shell variables there)
     moc_extra_flags="$(eval echo $(grep '//MOC-FLAG' $base.mo | cut -c11- | paste -sd' '))"
     moc_extra_env="$(eval echo $(grep '//MOC-ENV' $base.mo | cut -c10- | paste -sd' '))"
-    moc_with_flags="env $moc_extra_env moc $moc_extra_flags $EXTRA_MOC_ARGS"
+    if ! grep -q "//MOC-NO-FORCE-GC" $base.mo
```
crusso, line 252: Would it be easier to just add a `--dont-force-fc flag` to moc for those few test?

test/run-drun/destabilization-crash/destabilization-crash.mo
luc-blaeser: Good point. This is actually a difference arising from the different object header lengths due to the forwarding pointer, and it is unrelated to partition size. I tried to improve these changes:
* Rename `partitionSizeMB` to `toleranceMB` and set a smaller value of 4MB for `destabilization-crash` and 2MB for similar code in `rts-stack-pages`
* I also moved the change down to the forwarding pointer PR https://github.com/dfinity/motoko/pull/3829. Since the incremental GC is only running on this PR, certain changes in the tests of the lower PR are only preparation work for the higher PR.

test/run-drun/throw-on-async-failure.mo
luc-blaeser: Several long-running tests failed with the incremental GC with sanity checks and/or `ic-ref-run`, with the typical error `Ingress message did not finish executing within 10000 batches, panicking`. 

The reason is that the `--force-gc` is always activated by `drun-wrapper` and the incremental GC has higher runtime complexity if always forced to run, especially in busy loops, e.g. in timer tests etc..

test/run.sh
luc-blaeser: I was considering so, but the problem was that `drun-wrapper.sh` always unconditionally set `--force-gc` and combining this with an additional `--dont-force-gc` would be contradicting `moc` argument. 

In `drun-wrapper.sh`:
```
EXTRA_MOC_ARGS="--force-gc $EXTRA_MOC_ARGS"
```

Maybe there is a nicer solution that we conditionally activate `force-gc` without having to adjust the other tests...

rts/motoko-rts/src/gc/incremental.rs
```diff
@@ -1,18 +1,424 @@
...
+                &mut self.state.partitioned_heap,
+                _roots,
+                sanity_checks::CheckerMode::UpdateCompletion,
+            );
+        }
+    }
+}
+
+/// Write barrier to be called BEFORE a potential overwrite of a pointer value.
+/// `overwritten_value` (skewed if a pointer) denotes the value that will be overwritten.
```
crusso, line 308: This is new value, right? Not the current contents of location, before the write.

luc-blaeser: No, this should be the old value (maybe I should rename that it is clearer). This is important because of the snapshot-at-the-beginning principle: (At least) all objects are marked that have been reachable at the beginning of the GC run. If a value is overwritten, we could miss that original reference and need to mark it to be sure we have a super-set of the initial snapshot marked.

crusso: Ok, I guess that's the snap-shot at the beginning thing. Makes sense.

test/run.sh
crusso: Right, I'd be fine with it cancelling the preceding --force-gc.

Or // SKIP force-gc? to opt out?

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -0,0 +1,703 @@
...
+
+    unsafe fn allocate_normal_object<M: Memory>(&mut self, mem: &mut M, size: usize) -> Value {
+        debug_assert!(size <= PARTITION_SIZE);
+        let mut allocation_partition = self.allocation_partition();
+        debug_assert!(!allocation_partition.free);
+        let heap_pointer = allocation_partition.dynamic_space_end();
+        debug_assert!(size <= allocation_partition.end_address());
+        if heap_pointer <= allocation_partition.end_address() - size {
+            (*allocation_partition).dynamic_size += size;
+            Value::from_ptr(heap_pointer)
```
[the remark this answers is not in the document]
crusso: GOT HERE (pausing review)

crusso (commented):
partial review. TBC.

rts/motoko-rts/src/gc/incremental/mark_bitmap.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    /// Advance the iterator to the next marked offset.
+    pub fn next(&mut self) {
+        debug_assert!(self.next_bit_index <= BIT_INDEX_END);
+        // Outer loop iterates the 64-bit words.
+        while self.next_bit_index < BIT_INDEX_END {
+            // Examine the least significant bit(s) in the current word.
+            if self.current_word != 0 {
+                let shift = self.current_word.trailing_zeros() as usize;
+                // Two shifts to avoid an overflow in the case of `shift == 63`.
```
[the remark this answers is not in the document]
crusso: Oh maybe it's that shift on shift+1=64 overflows. I did not know that.

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -0,0 +1,703 @@
...
+    unsafe fn allocate_temporary_partition(&mut self) -> &mut Partition {
+        for partition in &mut self.partitions {
+            if partition.is_completely_free() {
+                debug_assert_eq!(partition.dynamic_size, 0);
+                partition.free = false;
+                partition.temporary = true;
+                return partition;
+            }
+        }
+        rts_trap_with("Cannot grow memory");
```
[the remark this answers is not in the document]
crusso: Yeah, I see it is. nvm.

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -0,0 +1,703 @@
...
+            self.allocate_in_new_partition(mem, size)
+        }
+    }
+
+    // Significant performance gain by not inlining.
+    #[inline(never)]
+    unsafe fn allocate_in_new_partition<M: Memory>(&mut self, mem: &mut M, size: usize) -> Value {
+        #[cfg(feature = "memory_check")]
+        self.allocation_partition().clear_free_remainder();
+
```
crusso, line 582:
```suggestion

        debug_assert!(size <= PARTITION_SIZE);
```
?

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -0,0 +1,703 @@
...
+        }
+    }
+
+    pub fn plan_evacuations(&mut self) {
+        for partition in &mut self.partitions {
+            debug_assert!(!partition.evacuate);
+            partition.evacuate = self.allocation_index != partition.index
+                && !partition.is_free()
+                && !partition.has_large_content()
+                && partition.dynamic_space_start() < partition.end_address()
```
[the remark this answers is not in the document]
crusso: Oh, maybe for a purely static partition. Ok.

rts/motoko-rts/src/gc/incremental/phases/mark_increment.rs
```diff
@@ -0,0 +1,134 @@
...
+    mark_stack: MarkStack,
+    complete: bool,
+}
+
+pub struct MarkIncrement<'a, M: Memory> {
+    mem: &'a mut M,
+    time: &'a mut BoundedTime,
+    heap: &'a mut PartitionedHeap,
+    mark_stack: &'a mut MarkStack,
+    complete: &'a mut bool,
```
crusso, line 25:
```suggestion
    mark_state: &'a mut MarkState
```
Would this be possible for clarity, i.e. just alias the MarkState, not its fields? Or does it make it slower?

rts/motoko-rts/src/gc/incremental.rs
```diff
@@ -1,18 +1,424 @@
...
+    /// Regular GC increment invoked when the call stack is guaranteed to be empty.
+    /// As the GC cannot scan or use write barriers on the call stack, we need to ensure:
+    /// * The mark phase can only be started on an empty call stack.
+    /// * The update phase can only be completed on an empty call stack.
+    pub unsafe fn empty_call_stack_increment(&mut self, roots: Roots) {
+        assert!(self.state.phase != Phase::Stop);
+        if self.pausing() {
+            self.start_marking(roots);
+        }
+        self.increment();
```
crusso, line 205: Any reason this call to self.increment() is outside the conditional, but the others are not?

rts/motoko-rts/src/gc/incremental/phases/evacuation_increment.rs
```diff
@@ -0,0 +1,109 @@
...
+        }
+    }
+
+    unsafe fn evacuate_object(&mut self, original: *mut Obj) {
+        debug_assert!(original.tag() >= TAG_OBJECT && original.tag() <= TAG_NULL);
+        debug_assert!(!original.is_forwarded());
+        let size = block_size(original as usize);
+        let new_address = self.mem.alloc_words(size);
+        let copy = new_address.get_ptr() as *mut Obj;
+        memcpy_words(copy as usize, original as usize, size);
```
crusso, line 83: Is there a problem when this is a large array? Or can that not happen since a large arrays would be  large object and just not get evacuated at all? I guess `size` is at most one partition here.

rts/motoko-rts/src/gc/incremental/phases/evacuation_increment.rs
```diff
@@ -0,0 +1,109 @@
...
+        self.time
+            .advance(1 + (size.as_usize() as f64 / TIME_FRACTION_PER_WORD) as usize);
+
+        #[cfg(feature = "memory_check")]
+        Self::clear_object_content(original);
+    }
+
+    #[cfg(feature = "memory_check")]
+    unsafe fn clear_object_content(original: *mut Obj) {
+        let object_size = block_size(original as usize);
```
crusso, line 103: Would it make sense to change the tag too? Or do you deliberately want to leave behind an inconsistent object

rts/motoko-rts/src/gc/incremental/phases/update_increment.rs
```diff
@@ -0,0 +1,125 @@
...
+    heap: &'a PartitionedHeap,
+    iterator: &'a mut PartitionedHeapIterator,
+    updates_needed: bool,
+}
+
+impl<'a> UpdateIncrement<'a> {
+    pub unsafe fn start_phase(state: &mut State) {
+        debug_assert!(state.iterator_state.is_none());
+        let heap = &mut state.partitioned_heap;
+        state.iterator_state = Some(PartitionedHeapIterator::new(heap));
```
crusso, line 24:
```
        state.iterator_state = Some(PartitionedHeapIterator::new(heap));
```
Silly question, but where do these Rust objects get allocated and how do we know the GC won't mess with them? IIRC Rust just allocates into Motoko heap blobs - how is the GC prevented from trashing them?

src/codegen/compile.ml
```diff
@@ -1577,6 +1580,7 @@ module Tagged = struct
     | Null -> 27l
     | OneWordFiller -> 29l
     | FreeSpace -> 31l
+    | ArraySliceMinimum -> 32l
```
crusso, line 1583: I wonder if we need to extend `can_have_tag` although it appears to be unused to me (referenced once from an unused function).

crusso: I think it's fine as is, since the tag shouldn't occur in generated code and the conservative approximation in can_have_tag should do the right thing.

crusso:
I've mode some suggestions in #3927

crusso:
One thing I don't quite understand is the maximum heap usage. This is still a copying GC at heart, right? Since we can't drop partitions until update is complete, does this mean that we can use at most half the heap for live data if we want GC to succeed? For example, if half+1 parititions are 95% full and all are scheduled for GC, will GC fail to complete because we can't compact the data into the remaining (half - 1) free partitions? Or am I misunderstanding something?

rts/motoko-rts/src/gc/incremental/mark_bitmap.rs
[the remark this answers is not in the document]
luc-blaeser: Yes, and worse, in release mode, it is even undefined behavior (according to the Rust language specification)...

rts/motoko-rts/src/gc/incremental/mark_stack.rs
```diff
@@ -0,0 +1,119 @@
...
+                self.last = (*self.last).next;
+            }
+            self.top = 0;
+        }
+        debug_assert!(self.top < STACK_TABLE_CAPACITY);
+        (*self.last).entries[self.top] = value;
+        self.top += 1;
+    }
+
+    pub unsafe fn pop(&mut self) -> Option<Value> {
```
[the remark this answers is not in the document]
luc-blaeser: Thank you very much. This makes sense. This would also apply to the generational and compacting GC. Maybe it will  even offer a small performance gain. I will implement this.

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -0,0 +1,703 @@
...
+/// 2. Lowest WASM memory size (i.e. for minimum heap size).
+pub const SURVIVAL_RATE_THRESHOLD: f64 = 0.85;
+
+/// Heap partition of size `PARTITION_SIZE`.
+pub struct Partition {
+    index: usize,        // Index of the partition `0..MAX_PARTITIONS`.
+    free: bool,          // Denotes a free partition (which may still contain static space).
+    large_content: bool, // Specifies whether a large object is contained that spans multiple partitions.
+    marked_size: usize,  // Total amount of marked object space in the dynamic space.
+    static_size: usize,  // Size of the static space.
```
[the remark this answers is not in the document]
luc-blaeser: Thanks, good point. I will improve this.

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -0,0 +1,703 @@
...
+/// -> Large partitions above 32 MB are a waste for small programs, since the WASM memory is
+///    allocated in that granularity and GC is then triggered later.
+pub const PARTITION_SIZE: usize = 32 * 1024 * 1024;
+
+/// Total number of partitions in the memory.
+/// For simplicity, the last partition is left unused, to avoid a numeric overflow when
+/// computing the end address of the last partition.
+const MAX_PARTITIONS: usize = (WASM_MEMORY_BYTE_SIZE.0 / PARTITION_SIZE as u64) as usize - 1;
+
+/// Partitions are only evacuated if the space occupation of alive object in the partition
```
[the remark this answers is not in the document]
luc-blaeser: Thank you.

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -0,0 +1,703 @@
...
+/// is greater than this threshold.
+/// Based on benchmark measurements, this rate is tuned to optimize the following metrics
+/// in the order of occurrence:
+/// 1. Lowest heap size (most reclamation).
+/// 2. Lowest WASM memory size (i.e. for minimum heap size).
+pub const SURVIVAL_RATE_THRESHOLD: f64 = 0.85;
+
+/// Heap partition of size `PARTITION_SIZE`.
+pub struct Partition {
+    index: usize,        // Index of the partition `0..MAX_PARTITIONS`.
```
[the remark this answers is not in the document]
luc-blaeser: True, the `index` is only for convenience. Probably, I could eliminate it (and pass it separately). I will try this.

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -0,0 +1,703 @@
...
+}
+
+/// Partitioned heap used by the incremental GC.
+pub struct PartitionedHeap {
+    partitions: [Partition; MAX_PARTITIONS],
+    heap_base: usize,
+    allocation_index: usize, // Index of the partition currently used for allocations.
+    evacuating: bool,
+    reclaimed: u64,
+    bitmap_allocation_pointer: usize, // Free pointer for allocation the next mark bitmap.
```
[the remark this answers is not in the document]
luc-blaeser: Thank you.

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -0,0 +1,703 @@
...
+    gc_running: bool, // Create bitmaps for partitions whn allocated during active GC.
```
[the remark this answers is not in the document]
luc-blaeser: Thank you very much. Sorry for these mistakes. I will run a spellchecker next time...

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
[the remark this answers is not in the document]
luc-blaeser: Maybe we could differentiate the error messages a little bit in future (and relax the tests accordingly to deal with the different GCs)?

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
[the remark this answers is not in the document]
luc-blaeser: Thank you very much for the review.

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
luc-blaeser: I was unsure, as it is already checked in `allocate_normal_object()` and I only introduced this helper function for `allocate_normal_object` as optimization (for non-inlining the infrequent execution path). Worst case, `allocate_free_partition` would trap with "out of memory" if the size invariant would be violated.

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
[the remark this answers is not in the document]
luc-blaeser: Yes, but probably I should improve the code that it becomes more self-explanatory...

rts/motoko-rts/src/gc/incremental/phases/mark_increment.rs
luc-blaeser: I introduced it primarily for convenience for the accesses inside the mark increment phase (to skip the navigation via state). I believe it should not be slower but will measure it.

rts/motoko-rts/src/gc/incremental.rs
luc-blaeser: The motivation is to unconditionally perform at least one increment regardless of the phase (after the GC has been started). However, if there are pending GC phase switches, subsequent increments are triggered to possibly continue the next phase. 

I noticed a slight performance gain against the better readable and more symmetric version:
```
if self.pausing() {
  self.start_marking(roots);
}
if self.state.phase == Phase::Mark {
  self.mark_increment();
}
if self.mark_completed() {
  self.start_evacuating();
}
if self.state.phase == Phase::Mark {
  self.evacuation_increment();
}
if self.evacuation_completed() {
  self.start_updating();
}
if self.state.phase == Phase::Update {
  self.update_increment();
}
if self.updating_completed() {
  self.complete_run(roots);
}
```
I will remeasure the performance difference.

rts/motoko-rts/src/gc/incremental/phases/evacuation_increment.rs
luc-blaeser: Yes, large arrays (larger than partition size) would reside in specific partitions and will not be moved. This is important for ensuring (possibly quite generous) incremental limits on large arrays.

rts/motoko-rts/src/gc/incremental/phases/evacuation_increment.rs
luc-blaeser: This would make sense. I would have to adjust the memory sanity check that it does not unconditionally load the tag. (In an earlier version, this would have been more complicated since I performed a linear partition scan in the update phase, which has now become redundant thanks to the mark bitmap).
I will implement this.

rts/motoko-rts/src/gc/incremental/phases/update_increment.rs
luc-blaeser: This is a very important question. The static variables are declared in the WASM static data. The Motoko linker allocates this in static space and resizes the heap base accordingly.

src/codegen/compile.ml
luc-blaeser: Maybe we should remove this unused function (in a separate PR) as it could be a pitfall for the future (if we further adjust the tag logic)?

luc-blaeser:
> One thing I don't quite understand is the maximum heap usage. This is still a copying GC at heart, right? Since we can't drop partitions until update is complete, does this mean that we can use at most half the heap for live data if we want GC to succeed? For example, if half+1 parititions are 95% full and all are scheduled for GC, will GC fail to complete because we can't compact the data into the remaining (half - 1) free partitions? Or am I misunderstanding something?

You are absolutely right. This is a disadvantage of this GC (and all copying GCs). However, we have more control over this:
* We could lower the threshold. Shenandoah GC uses 40%.
* We could limit the work to a fix number of partitions (with most garbage). 
While this would reduce the additional allocation during GC, it may still lead to a waste of memory due to garbage in non-collected partitions.
I will investigate further: Maybe we should limit the number of collected partitions to the available free space such that the GC does not run out of memory.

test/run.sh
luc-blaeser: Maybe, cancelling `--force-gc` would be the nicest solution. And activating it in the critical test cases that test the GC. I will adjust this.

test/run-drun/throw-on-async-failure.mo
luc-blaeser: **Update**: I adjusted the test run to use `--force-gc` explicitly. Hopefully, this simplifies it a little bit...

rts/motoko-rts/src/gc/incremental/phases/update_increment.rs
crusso: So it actually reserves space for a large struct and overwrites it on new?

src/codegen/compile.ml
crusso: Yeah, I have not clue why it exists. I guess git blame might reveal who introduced it.

rts/motoko-rts/src/gc/incremental/phases/update_increment.rs
luc-blaeser: Yes. `PartitionedHeapIterator::new()` first creates the struct memory on the call stack and then overwrites the static variable space (in static heap).

crusso:
> You are absolutely right. This is a disadvantage of this GC (and all copying GCs). However, we have more control over this:

But I was strangley reassured by the gcbench figures showing the incremental GC scaling to 4GB. How can that be, or did I misinterpret those numbers?

luc-blaeser:
> > You are absolutely right. This is a disadvantage of this GC (and all copying GCs). However, we have more control over this:
> 
> But I was strangley reassured by the gcbench figures showing the incremental GC scaling to 4GB. How can that be, or did I misinterpret those numbers?

The 4GB in the scalability tests only means the total memory occupation, including GC overheads. Similarly, the copying GC could allocate 4GB (for only copying a 2GB heap space with only alive objects) but unfortunately, the copying GC usually hits the instruction limit first (due to the extensive copying).

The current scalability benchmark cases continuously allocate and retain objects until reaching a limit (memory limit or message instruction limit):
* For test cases with a lot of garbage (`rb-tree`, `trie-map`), it is scaling much higher than the other GCs because it can reclaim the garbage incrementally and the cases do no longer hit the instruction limit.
* For test cases with little garbage, it skips the evacuation phase for many low-garbage partitions. This is because the objects surviving a GC run are usually non-temporary retained objects, like old objects staying in their partitions.

Of course, one could create other test cases, that allocate a lot and suddenly let the objects become garbage at once: Here, the scalability could be less than the other incremental GC with central object table. However, I have such test cases in all of the data-structure-centric performance benchmark cases, but only a smaller scale, such that all GCs can run them (and we can also compare runtime difference). 

However, I believe, more fine-tuning is needed to address the case of memory shortage,  by limiting the number of partitions selected for evacuation based on available free space...

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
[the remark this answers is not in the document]
luc-blaeser: **Update**: It seems that keeping the immutable `index` inside the `Partition` structure is quite convenient for computing the `start_address()`, `end_address()`, and `free_size()` of the partition. If it is okay, I keep it. Or maybe there is another better approach?

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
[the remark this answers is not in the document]
luc-blaeser: I adjusted the logic as a side-effect of the changes with the new partition planning that considers the available free space.

rts/motoko-rts/src/gc/incremental/mark_bitmap.rs
```diff
@@ -0,0 +1,174 @@
...
+//! The bitmap memory is allocated in one or multiple separate bitmap partitions, that are freed on completion
+//! of a GC run.
+//!
+//! A bitmap is represented in raw memory (no object header, no blob) of the fixed size
+//! `PARTITION_SIZE / WORD_SIZE / u8::bits`, which is `PARTITION_SIZE /32` bytes. For simplicity, the bitmaps
+//! is not shortened for partitions that also accommodate static space besides dynamic space.
+//!
+//! To access a mark bit of an object, the corresponding bitmap and address offset inside the object's partition
+//! needs to be first determined. The corresponding bit is then accessed at the byte with index
+//! `offset / WORD_SIZE / u8::BITS` and at the bit index `offset % u8::BITS`.
```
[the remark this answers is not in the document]
luc-blaeser: Yes, indeed. Thank you.

rts/motoko-rts/src/gc/incremental/mark_bitmap.rs
```diff
@@ -0,0 +1,174 @@
...
+                return;
+            }
+
+            // Move on to next word, always within a 64-bit boundary.
+            self.next_bit_index += self.leading_zeros;
+            if self.next_bit_index == BIT_INDEX_END {
+                return;
+            }
+            let word64_index = self.next_bit_index / size_of::<u64>();
+            self.current_word = unsafe { *(self.bitmap_pointer.add(word64_index) as *const u64) };
```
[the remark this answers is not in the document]
luc-blaeser: Indeed, the units are wrong (with accidentally matching values). I have fixed this. Thank you very much for having seen this issue.

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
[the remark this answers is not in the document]
luc-blaeser: I tried the early return and measured the performance: It seems to be identical (same generated code which is good).

rts/motoko-rts/src/gc/incremental/phases/mark_increment.rs
luc-blaeser: I measured it and the difference is negligibly small. Interestingly, the combined state with the extra indirection is sometimes a little bit faster (<0.1%).

rts/motoko-rts/src/gc/incremental/phases/evacuation_increment.rs
luc-blaeser: I have now implemented this additional check. Thanks for the suggestion.

src/codegen/compile.ml
luc-blaeser: I have put this on my backlog for future code refactoring (in a separate PR).

rts/motoko-rts/src/gc/incremental.rs
luc-blaeser: **Update**: I remeasured and the performance seems the same. So, I refactored to code to have the better readable and symmetric version (mentioned above). Thank you very much for this question.

rts/motoko-rts/src/gc/incremental/sort.rs
```diff
@@ -0,0 +1,41 @@
...
+        if left > right {
+            break;
+        }
+    }
+    if right > 0 {
+        quicksort(&mut array[..right + 1], compare);
+    }
+    if left < array.len() - 1 {
+        quicksort(&mut array[left..], compare);
+    }
```
crusso, line 40:
```suggestion
    if right < array.len() - 1 {
        quicksort(&mut array[right + 1..], compare);
    }
    if left > 0  {
        quicksort(&mut array[..left], compare);
    }
```
These recursive calls look wrong to me, but perhaps I'm misunderstanding Rust or this version of quicksort. Shouldn't it be something like the above?

crusso (commented):
With your selective use of the force-gc fllag, do we now still have a testing mode where most tests are run with --force-gc? I think the rationale for doing this originally was to flush out gc bugs.

rts/motoko-rts/src/gc/incremental/sort.rs
luc-blaeser: Interesting point: It is indeed a special version I have adopted from Niklaus Wirth (https://people.inf.ethz.ch/wirth/AD.pdf, pages 65/66): After the partitioning, the invariant is `left > right`. The original version uses variable names `i` and `j`. I will add a comment for clarification. Maybe, I should also choose better variable names instead?
Side note: I noticed that the examples in our samples could run in endless recursion if the pivot would not be chosen as the first element (which is bad for sorted or nearly sorted arrays, regardless of the order). Here, the pivot is the middle element which is good for (nearly) sorted arrays, which could be helpful in scenarios that follow the generational hypothesis where low-garbage old generations cold be populated in linear fashion. Although, one can always create opposite examples...

luc-blaeser:
> With your selective use of the force-gc fllag, do we now still have a testing mode where most tests are run with --force-gc? I think the rationale for doing this originally was to flush out gc bugs.

Good point: I will improve this and add the `--force-gc` flag to all test cases except those where the GC is too expensive. Thank you for noticing this.

rts/motoko-rts/src/gc/incremental/sort.rs
```diff
@@ -8,34 +8,40 @@ pub fn sort<Compare: Fn(usize, usize) -> Ordering>(array: &mut [usize], compare:
     }
 }
 
+/// Special version of quicksort by Niklaus Wirth: https://people.inf.ethz.ch/wirth/AD.pdf, pages 65/66.
```
crusso, line 11: Thanks for commenting and renaming this!

luc-blaeser: Thank you very much for reviewing. And also thanks for suggesting to improve the naming.

ulan (approved):
The new GC lgtm as far as I can see. Great work and good luck with launching it!

luc-blaeser:
> The new GC lgtm as far as I can see. Great work and good luck with launching it!

Thank you very much Ulan for the review of these complex PRs and the many very valuable design discussions.

crusso (approved):
Let's ship this puppy!

iclighthouse:
This is the best news I heard this weekend! Great job!

ByronBecker:
🥳 🎉 🎉 🎉

Issue #3852, opened 2023-03-01
perf: optimize compile.ml Arry.toBlob/ofBlob


crusso:
Both do bounds checked arithmetic and explicit indexing. A while loop bumping a pointer would be safer and faster.

crusso:
``` ocam
let ofBlob env =
    Func.share_code1 env "Arr.ofBlob" ("blob", I32Type) [I32Type] (fun env get_blob ->
      let (set_len, get_len) = new_local env "len" in
      let (set_r, get_r) = new_local env "r" in
      get_blob ^^ Blob.len env ^^ set_len ^^
      get_len ^^ alloc env ^^ set_r ^^

      get_len ^^ from_0_to_n env (fun get_i ->
        get_r ^^ get_i ^^ idx env ^^                          (* bounds check! *)
        get_blob ^^ Blob.payload_ptr_unskewed ^^
        get_i ^^ G.i (Binary (Wasm.Values.I32 I32Op.Add)) ^^
        G.i (Load {ty = I32Type; align = 0; offset = 0l; sz = Some Wasm.Types.(Pack8, ZX)}) ^^
        TaggedSmallWord.msb_adjust Type.Nat8 ^^
        store_ptr
      ) ^^
      get_r
    )
  let toBlob env =
    Func.share_code1 env "Arr.toBlob" ("array", I32Type) [I32Type] (fun env get_a ->
      let (set_len, get_len) = new_local env "len" in
      let (set_r, get_r) = new_local env "r" in

      get_a ^^ Heap.load_field len_field ^^ set_len ^^

      get_len ^^ Blob.alloc env ^^ set_r ^^

      get_len ^^ from_0_to_n env (fun get_i ->
        get_r ^^ Blob.payload_ptr_unskewed ^^
        get_i ^^ G.i (Binary (Wasm.Values.I32 I32Op.Add)) ^^
        get_a ^^ get_i ^^ idx env ^^ (* bound check *)
        load_ptr ^^
        TaggedSmallWord.lsb_adjust Type.Nat8 ^^
        G.i (Store {ty = I32Type; align = 0; offset = 0l; sz = Some Wasm.Types.Pack8})
      ) ^^
      get_r
    )
```

PR #3853, opened 2023-03-02
Backend: Avoid pointless bounds checking array access

nomeata:
this has been in the back of my mind for a while: Originally, I intentionally used safe array access almost everywhere in the backend, even in places where the bounds should always be fine, so that bugs can be reported more efficiently. It seems that people have not found bugs related to that code, so we can drop the bounds check in compiler-generated loops over arrays (mostly in serialization/deserialization).

Turns out all array accesses that need to be bounds-checked go via the bignum index function anyways, so I moved the bounds check there.

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -3576,17 +3576,12 @@ module Arr = struct
   (* Static array access. No checking *)
   let load_field n = Heap.load_field Int32.(add n header_size)
 
-  (* Dynamic array access. Returns the address (not the value) of the field.
-     Does bounds checking *)
-  let idx env =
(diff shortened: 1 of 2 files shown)


crusso:
There's a small diff in the stack trace in drun causing the test to fail. I think it's benign,

Some of those indexing loops can also be optimized to just increment a pointer, I think. See https://github.com/dfinity/motoko/issues/3852.

nomeata:
>  In terms of gas, 2 tests regressed, 1 tests improved and the mean change is +0.3%.

That’s surprising. How can this change regress in gas usage?

crusso (approved):
LGTM, but our cost model is big flakey I suspect.

PR #3768, opened 2023-02-02
Multiple stable memory regions (and a new primitive type `Region`).

matthewhammer:
Runtime-system implementation of the API described in https://github.com/dfinity/motoko-base/pull/516

Checklist:
- [x] pass existing tests for existing (experimental) stable memory, using region0 as that stable memory.
- [x] distinct regions are isolated (see `test/run-drun/stable-regions-are-isolated.mo`).
- [x] serialize/deserialize `Region` type.
- [x] migration path for existing (experimental) stable memory, into region0 of new region manager.
- [x] trap when Region.new fails?
- [x] add version to metadata
- [x] basic perf test
- [x] maybe remove (most of) Region0.rs
- [x] align errors
- [ ] improve test region0/stable-mem-big-blog.mo to check contents of large read/writes (done in halves)
- [ ] restrict access to new somehow (can be done later, but nicer to have on release).
- [x] rts_stable_mem_size?
- [x] max-stable-pages (limits physical pages, i.e. include meta-data) 
- [ ] registers not stack allocation?
- [x] clean up asserts on tag ranges
- [x] lazy metadata allocation?
- [x] add Region.mo to next_moc
- [x] document Region.mo 
- [ ] perf maybe: turn ic_mem_fns into ordinary externs where poss - affects perf IIRC from experience of @luc-blaeser (NOTE, I couldn't see this myself after a simple experiment)
- [x] elim Region0.rs operations wrappers and just do the conditional compilation in compile.ml - bench shows inlining  should be a win.
- [ ] simplify chunked read writes to iterate over vec_pages entries directly should be cheaper.
- [x] specify Region type in manual

TODO separately:
- [ ] add stable-regions.md to portal doc (sidebars.js)
- [ ] gracefully fail compilation of stablemem/region ops by generating trapping functions
- [ ] protect access to Region.new 

See also:
 - [Forum post, for community discussion.](https://forum.dfinity.org/t/motoko-stable-regions/19182)
 - **StableBTree** edits needed to upgrade to this system:
   - https://github.com/sardariuss/MotokoStableBTree/pull/4
   - https://github.com/sardariuss/MotokoStableBTreeTest/pull/1

PR #3860, opened 2023-03-07
fix call to region_init

crusso:
Fixes #3768

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -10501,7 +10501,7 @@ and main_actor as_opt mod_env ds fs up =
         G.nop
     end ^^
     IC.init_globals env ^^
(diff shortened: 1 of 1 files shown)


src/codegen/compile.ml
```diff
@@ -10501,7 +10501,7 @@ and main_actor as_opt mod_env ds fs up =
         G.nop
     end ^^
     IC.init_globals env ^^
-    G.i (Call (nr (E.built_in env "region_init"))) ^^
+    E.call_import env "rts" "region_init" ^^
     (* Continue with decls *)
```
matthewhammer, line 10505: 🎉 Makes sense.  Thanks @crusso!

matthewhammer: Just merged and checked locally.  This works.  Thanks again!

PR #3865, opened 2023-03-09
Support multi-value and unboxed returns from switch statement

nomeata:
so that `switch` behaves like `if` does since #3566.

Only kicks in if _all_ branches produce a manifest tuple or call a function that returns
tuples; if only one branch has an expression that is a first-class tuple, all will return a
first-class tuple (could be improved by adding cases to `compile_exp_as`)

Should have made parts of
https://github.com/dfinity/motoko-base/pull/535 not necessary.

Example code:
```
func multi_value_switch(n : Nat) : (Nat, Nat) {
  switch n {
    case 0 { returns_tuple() };
    case 1 { returns_tuple() };
    case _ { (5,6) };
  };
};
```

Changes (calls to `to_2_tuple` are allocating):
```diff
   (func $multi_value_switch (type 1) (param $clos i32) (param $n i32)
-    (local $switch_in i32) (local $switch_out i32) (local $n.1 i32)
+    (local $switch_in i32)
     local.get $n
     local.set $switch_in
-    block (result i32)  ;; label = @1
-      local.get $switch_in
-      i32.const 0
-      call $B_eq
-      if  ;; label = @2
-      else
-        i32.const 0
-        br 1 (;@1;)
-      end
-      i32.const 0
-      call $returns_tuple
-      global.get 17
-      global.get 18
-      call $to_2_tuple
-      local.set $switch_out
-      i32.const 1
-    end
-    if  ;; label = @1
-    else
+    block  ;; label = @1
       block (result i32)  ;; label = @2
         local.get $switch_in
-        i32.const 2
+        i32.const 0
         call $B_eq
         if  ;; label = @3
         else
           i32.const 0
           br 1 (;@2;)
         end
         i32.const 0
         call $returns_tuple
-        global.get 17
-        global.get 18
-        call $to_2_tuple
-        local.set $switch_out
-        i32.const 1
+        br 1 (;@1;)
       end
       if  ;; label = @2
       else
-        i32.const 2097763
-        local.set $switch_out
+        block (result i32)  ;; label = @3
+          local.get $switch_in
+          i32.const 2
+          call $B_eq
+          if  ;; label = @4
+          else
+            i32.const 0
+            br 1 (;@3;)
+          end
+          i32.const 0
+          call $returns_tuple
+          br 2 (;@1;)
+        end
+        if  ;; label = @3
+        else
+          i32.const 10
+          i32.const 12
+          global.set 18
+          global.set 17
+          br 2 (;@1;)
+        end
       end
-    end
-    local.get $switch_out
-    call $from_2_tuple)
+    end)
```


nomeata:
>  In terms of gas, 4 tests improved and the mean change is -0.3%.

At least noticable :-)

src/codegen/compile.ml
```diff
@@ -9789,21 +9807,28 @@ and compile_exp (env : E.t) ae exp =
...
-      | {it={pat; exp=e}; _}::cs ->
-          let (ae1, code) = compile_pat_local env ae pat in
-          orElse ( CannotFail get_i ^^^ code ^^^
-                   CannotFail (compile_exp_vanilla env ae1 e) ^^^ CannotFail set_j)
-                 (go env cs)
-          in
-      let code2 = go env cs in
-      code1 ^^ set_i ^^ orTrap env code2 ^^ get_j
+
+    (* compile subexpressions and collect the provides stack rep *)
```
ggreif, line 9813:
```suggestion
    (* compile subexpressions and collect the provided stack reps *)
```

ggreif:
I think the
```
       if  ;; label = @2
       else
```
is completely redundant (as well as the `i32.const 0`) because we never push a nonzero value!

ggreif (approved):
Looks awesome!

test/run/multi-value-block.mo
```diff
@@ -0,0 +1,59 @@
...
+let a7 = Prim.rts_total_allocation();
+
+assert((x1,y1) == (1,2));
+assert((x2,y2) == (1,2));
+assert((x3,y3) == (1,2));
+assert((x4,y4) == (3,4));
+assert((x5,y5) == (1,2));
+assert((x6,y6) == (1,2));
+assert((x7,y7) == (5,6));
+assert(a1 == a0);
```
ggreif, line 53: this could be rolled into line 32, etc.

nomeata:
> I think the
> 
> ```
>        if  ;; label = @2
>        else
> ```
> 
> is completely redundant (as well as the `i32.const 0`) because we never push a nonzero value!

Yes, I noticed. It is an artifact of how `PatCode.orElse` I think this is because in the way the `switch` statement is compiled now, we always branch on a successful case, but I think it’s not the case for other patterns (e.g. function patterns, `let`, etc.). Or at least I am unsure. Let's not worry for this PR, maybe open an issue to remember to look into this.

test/run/multi-value-block.mo
nomeata: 🤷🏻 It’s a test file, not worth worrying.
(Initially I wanted to compare and use `Debug.print` to get _all_ failures in nice errors, but then I didn’t bother.)

nomeata:
This `timer.mo` flaky test is a huge PITA. How can you deal with that on a daily basis.

PR #3866, opened 2023-03-09
CI: Run ic-ref-run only once

nomeata:
We run the tests in `test/run-drun` multiple times: drun or ic-ref-run; normal or dbug or two GC variants.

It does not seem to be that useful to run debug or GC variants on multiple backends. So let’s drop these jobs on ic-ref-test.


ggreif (commented):
You are throwing out two items from an attribute set. Since `nix` is lazy, these wouldn't run anyway...

nomeata:
But we run all attribute in `tests`:
```
  all-systems-go = nixpkgs.releaseTools.aggregate {
    name = "all-systems-go";
    constituents = [
      moc
      mo-ide
      mo-doc
      didc
      deser
      samples
      rts
      base-src
      base-tests
      base-doc
      docs
      report-site
      ic-ref-run
      shell
      check-formatting
      check-rts-formatting
      check-generated
      check-grammar
      check-error-codes
    ] ++
    builtins.attrValues tests ++
    builtins.attrValues js;
  };
```

ggreif:
> But we run all attribute in `tests`

Aaah, cool, I didn't know that is possible. `nix` is awesome!

ggreif (approved):
LGTM!

PR #3867, opened 2023-03-09
backend: Push expected stackrep into branches of `IfE` and `SwitchE`

nomeata:
This is a refinement upon #3865: If we know we’ll take apart a tuple (e.g. in tail-position in a function returning a tuple), we should use the unboxed-on-stack convention in every branch of a an `if`/`switch`, even if some branches don’t manifest tuples right away.

This worked with #3865:
```
func multi_value_switch(n : Nat) : (Nat, Nat) {
  switch (n){
    case (0) { returns_tuple(n) };
    case (1) { returns_tuple(n) };
    case (_) { (n,n+1) };
  };
};
```
but this one did not yet, and does now:
```
func multi_value_switch_mixed(n : Nat) : (Nat, Nat) {
  switch (n){
    case (0) { returns_tuple(n) };
    case (1) { let first_class = (n, n+1); first_class; };
    case (_) { (n,n+1) };
  };
};
```

Change to the generated code:
```diff
   (func $multi_value_switch_mixed (type 1) (param $clos i32) (param $n i32)
     (local $switch_in i32) (local $first_class i32)
     local.get $n
     local.set $switch_in
-    block (result i32)  ;; label = @1
+    block  ;; label = @1
       block (result i32)  ;; label = @2
         local.get $switch_in
         i32.const 0
         call $B_eq
         if  ;; label = @3
         else
           i32.const 0
           br 1 (;@2;)
         end
         i32.const 0
         local.get $n
         call $returns_tuple
-        global.get 17
-        global.get 18
-        call $to_2_tuple
         br 1 (;@1;)
       end
       if  ;; label = @2
       else
         block (result i32)  ;; label = @3
           local.get $switch_in
           i32.const 2
           call $B_eq
           if  ;; label = @4
           else
             i32.const 0
             br 1 (;@3;)
           end
           local.get $n
           local.get $n
           i32.const 2
           call $B_add
           call $to_2_tuple
           local.tee $first_class
+          call $from_2_tuple
           br 2 (;@1;)
         end
         if  ;; label = @3
         else
           local.get $n
           local.get $n
           i32.const 2
           call $B_add
-          call $to_2_tuple
+          global.set 18
+          global.set 17
           br 2 (;@1;)
         end
       end
       unreachable
-    end
-    call $from_2_tuple)
+    end)
```

Effectively, the `call $from_2_tuple` is moved into every branch (and eliminated where possible).
It can lead to large code sizes if there are many branches that still need the `call $from_2_tuple`. But probably still the right thing to do, as it seems to match expectations better (as discussed with @crusso).


test/run/multi-value-block.mo
```diff
@@ -1,59 +1,74 @@
 import Prim "mo:⛔";
 
-func returns_tuple() : (Nat, Nat) {
-  return (1,2);
+func returns_tuple(n : Nat) : (Nat, Nat) {
+  return (n,n+1);
```
crusso, line 4:
```suggestion
  return (n, n+1);
```

test/run/multi-value-block.mo
```diff
@@ -1,59 +1,74 @@
...
   if (n == 0) {
-    returns_tuple()
+    returns_tuple(0)
   } else {
     if (n == 1) {
-      returns_tuple()
+      returns_tuple(1)
     } else {
-      (3,4);
+      (n,n+1);
```
crusso, line 14:
```suggestion
      (n, n+1);
```

test/run/multi-value-block.mo
```diff
@@ -1,59 +1,74 @@
...
 };
 
+func multi_value_if_mixed(n : Nat) : (Nat, Nat) {
+  if (n == 0) {
+    returns_tuple(0)
+  } else {
+    if (n == 1) {
+      let first_class = (n, n+1); first_class;
+    } else {
+      (n,n+1);
```
crusso, line 26:
```suggestion
      (n, n+1);
```

test/run/multi-value-block.mo
```diff
@@ -1,59 +1,74 @@
...
+};
 
 func multi_value_switch(n : Nat) : (Nat, Nat) {
   switch (n){
-    case (0) { returns_tuple() };
-    case (1) { returns_tuple() };
-    case (_) { (5,6) };
+    case (0) { returns_tuple(n) };
+    case (1) { returns_tuple(n) };
+    case (_) { (n,n+1) };
```
crusso, line 35:
```suggestion
    case (_) { (n, n+1) };
```

test/run/multi-value-block.mo
```diff
@@ -1,59 +1,74 @@
...
   };
 };
 
+func multi_value_switch_mixed(n : Nat) : (Nat, Nat) {
+  switch (n){
+    case (0) { returns_tuple(n) };
+    case (1) { let first_class = (n, n+1); first_class; };
+    case (_) { (n,n+1) };
```
crusso, line 43:
```suggestion
    case (_) { (n, n+1) };
```

test/run/multi-value-block.mo
```diff
@@ -1,59 +1,74 @@
...
+  };
+};
 
+func test(fun : Nat -> (Nat, Nat), arg : Nat ) : Bool {
+  let a0 = Prim.rts_total_allocation();
+  let (x,y) = fun(arg);
```
crusso, line 49:
```suggestion
  let (x, y) = fun(arg);
```

crusso (approved):
LGTM, though I have low confidence in this part of the backend. @ggreif should take a look as well.

ggreif (approved):
I agree, this is the right thing to do.

Issue #3864, opened 2023-03-08
let-else: `()` cannot produce expected type `None`


di-wu:
> src/oops.mo
```motoko
func oops() : Nat {
  var x = 0;
  loop {
    if (x >= 10) return let y = x else {
      x += 1;
    }
  };
};

assert(oops() == 10);
```
Works in `moc v0.8.2`, breaks in `moc v0.8.3` (`moc -r src/oops.mo`).

ggreif:
Yes, `let` can have an `else` control flow now, whereas in 0.8.2 that belonged to the `if`.
Looks like a contrived example, but if you really want it that way, just make it `return {let y = x} else ...`.

di-wu:
Yes, there are easy workarounds, I just did not expect things to break after upgrading.
It this is expected, then I'll close this.

ggreif:
Claudio came up with a neat fix that unbreaks your code. #3869 will be part of the 0.8.4 release.

PR #3868, opened 2023-03-09
Fix `let-else` capturing `else` block from an `if`

kentosugama:
#3864

```
if false let _ = 4 else return;
assert false;
```

This snippet returns on the old version of `let` but fails on the new version.


ggreif:
This is resolved by #3869.

PR #3869, opened 2023-03-10
grammar fix for `let-else`

### LINKED ISSUE #3864: let-else: `()` cannot produce expected type `None`

crusso:
Builds on repro #3868 for change in semantics of 

`if exp let pat = exp else ...`

- restricts let-else construct to dec production, excluding use as expression
- removes need for precedence declaration.

Seem to preserve old semantics:

```
[nix-shell:~/motoko/test/run]$ rlwrap moc
Motoko compiler (source 0.8.3-14-g6e9f32125-dirty)
> func bar() { if false let _ = 4 else return; assert false};
let bar : () -> () = func
> bar();
> 
```

Fixes #3864


luc-blaeser (approved):
Great. I believe this is a very good pragmatic solution to the syntax backwards problem.

ggreif (approved):
Neat!

crusso:
Let's see if @kentosugama agrees too.

kentosugama (approved):
LGTM!

ggreif:
Closes #3868.

Issue #3857, opened 2023-03-04
Cancelling a recurring timer that I have lost track of


ByronBecker:
Is there a good way to cancel a recurring timer for which I've lost track of the timerId? (Without upgrading the canister in question)

Just curious about if this is possible right now as I ran into this issue earlier today.

ggreif:
Even if _you_ knew what the Id is, how would you force the _canister_ to `cancelTimer <Id>`?

ByronBecker:
I have an API endpoint that essentially does this

```
public shared ({ caller }) func cancelTimer() : async Result.Result<(), Text> {
    switch (timer) {
      case (?timerId) {
        cancelTimer(timerId);
        timer := null;
        #ok;
      };
      case null #err("No timer to cancel");
    };
  };
```

ggreif:
You could install an (almost) identical canister that also records the timer creations in an ordered data structure, and then makes that queryable. This should give you the Id. If this doesn't work (due to dynamic timer creation from internal data) then I lament, but you have to reinstall.

ByronBecker:
Instead of reinstalling, my current solution is just to stop the canister and then restart it.

ByronBecker:
@ggreif I'm running into a weird issue with recurring timers now. Here are the steps I take to get into this spot:

1. Create a function "A" which starts a recurring timer that executes every few (5-10) seconds (in my case the job is making an inter-canister call, but maybe this isn't important).
2. Call function "A"
3. Stop the canister that contains the timer.
4. Wait for a 30-60 seconds.
5. Restart the canister containing the timer. (timer should not be running).
6. Call function "A" again (attempting to start up a **new** timer that executions the same function).

After step 6, it seems like multiple timer executions will kick off back to back, almost like the old timer starts running again until it has "caught up" with the current time, at which point (once "caught up") the rapid executions will stop, and the "newly created timer" will take over.

ggreif:
Interesting findings, but I think it is easy to explain and needs some adjustments to the library.
# stopping the canister seems to reset the _global_ timer
# no timer gets cancelled when the a canister is stopped
# to prevent jitter, recurring timers will add the delay to the desired expiration time (_not_ the wall time of the expiration callback)

[...]

PR #3871, opened 2023-03-10
feat: when re-adding recurrent timers, skip over past expirations

ggreif:
See motivation in #3857 and forum post with discussion https://forum.dfinity.org/t/stopped-canisters-and-recurring-timers-discussion/18873.


src/prelude/internals.mo
```diff
@@ -532,9 +532,10 @@ func @timer_helper() : async () {
       if (n.expire[0] > 0 and n.expire[0] <= now and gathered < thunks.size()) {
         thunks[gathered] := ?(n.job);
         switch (n.delay) {
-          case (?delay) if (delay != 0) {
-            // re-add the node
-            let expire = n.expire[0] + delay;
+          case (null or ?0) ();
+          case (?delay) {
+            // re-add the node, skipping past expirations
+            let expire = n.expire[0] + delay * (1 + (now - n.expire[0]) / delay);
```
luc-blaeser, line 538: Could this maybe be simplified to `let expire = now + delay`?
I am not sure about the case of `delay == 0`: In the current logic, the expiration could then still be in the past?

luc-blaeser: Thank you. I think I now understand. The difference arises from the integer division that brings us back the recurring grid.

luc-blaeser (approved):
This looks good. And thanks for the explanation for the question I had.

PR #3873, opened 2023-03-10
perf: peephole-optimise `const 1; and; if t e` to `ctz; if e t`

ggreif:
Masking the LSBit happens often when checking for _vanilla_ pointers, so this is probably a profitable optimisation.
_Indeed_: this improves cycle counts about 1% on our test set.

Instead of masking with 1, we count the trailing zeroes. We expect a subsequent `if`, but its legs must be switched, since a masked 1 corresponds to _no_ trailing zeroes.

I expect bignum operations and `Nat/Int/32/64` operations to be significantly cheaper now.

Kento checked that `binaryen`'s `wasm-opt` doesn't perform this substitution currently, but I think it could be contributed.


ggreif:
similar opportunity:
``` wasm
              i32.const 1
              i32.and
              i32.eqz
              br_if 0 (;@5;)
```
to
``` wasm
              i32.ctz
              br_if 0 (;@5;)
```

test/run-drun/ok/perf.drun-run.ok
```diff
@@ -1,4 +1,4 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
-debug.print: 53_219
+debug.print: 51_218
```
ggreif, line 3: This is _noice_!

nomeata (approved):
Very nice

PR #3878, opened 2023-03-12
perf: identify const variants

ggreif:
Variants with constant payloads should be constant and end up in the static heap. This will speed up all tree-like data structures that have unit-payload leaves (not to speak of allocation wins!). See https://github.com/dfinity/motoko-base/issues/543 for motivation.

For below program
``` Motoko
import P = "mo:⛔️";

type Tree = { #leaf; #node : (Tree, Tree) };

func leaf() : (Tree, (Nat, Nat)) = (#leaf, (5, 42));

P.debugPrint(debug_show leaf());
```
the function `leaf()` now looks like
``` wasm
  (func $leaf (type 5) (param $clos i32)
    i32.const 2097267
    i32.const 2097279
    global.set 18
    global.set 17)
```
looks as expected. It also runs happily
```
$ wasmtime tree.wasm 
(#leaf, (5, 42))
```

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -138,6 +138,7 @@ module Const = struct
     | Obj of (string * t) list
     | Unit
     | Array of t list (* also tuples, but not nullary *)
+    | Tag of (string * t)
     | Lit of lit
 
   (* A constant known value together with a vanilla pointer.
@@ -1749,6 +1750,13 @@ module Variant = struct
   let test_is env l =
     get_variant_tag ^^
     compile_eq_const (hash_variant_label env l)
+
+  let vanilla_lit env i ptr =
+    E.add_static env StaticBytes.[
+        I32 Tagged.(int_of_tag Variant);
+        I32 (hash_variant_label env i);
+        I32 ptr
+      ]
 
 end (* Variant *)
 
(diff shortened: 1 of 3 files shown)


src/codegen/compile.ml
```diff
@@ -10307,8 +10318,12 @@ and compile_const_exp env pre_ae exp : Const.t * (E.t -> VarEnv.t -> unit) =
   | PrimE (ArrayPrim (Const, _), es)
   | PrimE (TupPrim, es) ->
     let (cs, fills) = List.split (List.map (compile_const_exp env pre_ae) es) in
-    Const.t_of_v (Const.Array cs),
+    Const.(t_of_v (Array cs)),
     (fun env ae -> List.iter (fun fill -> fill env ae) fills)
+  | PrimE (TagPrim i, [e]) ->
+    let (object_ct, fill) = compile_const_exp env pre_ae e in
+    Const.(t_of_v (Tag (i, object_ct))),
+    fill
```
nomeata, line 10326: It's not necessarily an object, so maybe
```suggestion
  | PrimE (TagPrim i, [e]) ->
    let (arg_ct, fill) = compile_const_exp env pre_ae e in
    Const.(t_of_v (Tag (i, arg_ct))),
    fill
```

nomeata (approved):
Looks good so far

crusso:
Nice - I'm surprised we didn't do this already, but I guess variants came later.

PR #3879, opened 2023-03-13
Gabor/modular deser

ggreif:
Issues to be aware of

- function name (`@deserialize<`, `deserialize_go` etc.) needs to be distinguished when present twice (type hash already enters, but there is a slight chance that things collide!)
- [x] use first-class modules? — yes!
- [ ] can incremental GC move the `buf`'s underlying `Blob` around?

## Assumptions and shortcuts

We need to limit the scope of this PR so that this can be implemented realistically. I.e.
- while stable memory is 64-bit addresses, we'll limit ourselves to 32 bit (offsets from a start address)
- we don't change the portions of the RTS (Rust code), which means that the Candid header needs to be read from the working heap (speculatively a few pages, presumably a configurable amount)
- we don't zero out the entire stable memory hosting the stable variables, just the front part


luc-blaeser:

PR #3880, opened 2023-03-14
chore: remove FIXME from `timer-initial.mo` test

ggreif:
Looks like we have a conforming `drun` now.


crusso (approved):
BTW, do we still have the problem that Timer.mo are non-det?

ggreif:
@MergifyIO help

PR #3882, opened 2023-03-15
doc: fix definition of subtyping of object types in `language-manual.md`

### LINKED ISSUE #3846: Docs: should actor subtyping be mentioned in the Quick Reference?
There is no mention of actor subtyping [in the quick ref](https://internetcomputer.org/docs/current/motoko/main/language-manual#subtyping) 

But @rossberg [clearly states](https://forum.dfinity.org/t/using-variant-variables-in-the-canister-public-interface-return-value-breaks-composability/16960/19?u=samer) that subtyping behaves the same as objects

crusso:
Fixes #3846


ggreif (approved):
Rubber-stamping!

PR #3883, opened 2023-03-15
bugfix: check `@nextExpiration` after the changes are committed

ggreif:
There was a slightly buggy behaviour with setting the global timer from `@timer_helper`. We observed that for each one-off timer set there were two expirations happening. The reason is that we were computing the next expiration before the current batch of expired timers were expunged. This re-added the same time as before, and thus caused the second (in-vain) wakeup.

The fix is of course to compute the next global expiration _after_ the already expired nodes were expunged.

### DIFF
--- a/src/prelude/internals.mo
+++ b/src/prelude/internals.mo
@@ -519,15 +519,8 @@ func @timer_helper() : async () {
   func Array_init<T>(len : Nat,  x : T) : [var T] {
     (prim "Array.init" : <T>(Nat, T) -> [var T])<T>(len, x)
   };
+
   let now = (prim "time" : () -> Nat64)();
-  let exp = @nextExpiration @timers;
-  let prev = (prim "global_timer_set" : Nat64 -> Nat64) exp;
-
-  // debug { assert prev == 0 };
-
-  if (exp == 0) {
-    return
-  };
 
   var gathered = 0;
   let thunks = Array_init<?(() -> async ())>(10, null); // we want max 10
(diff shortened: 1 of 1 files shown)


src/prelude/internals.mo
```diff
@@ -565,6 +558,10 @@ func @timer_helper() : async () {
 
   gatherExpired(@timers);
 
+  let exp = @nextExpiration @timers;
+  let prev = (prim "global_timer_set" : Nat64 -> Nat64) exp;
```
ggreif, line 562:
```suggestion
  ignore (prim "global_timer_set" : Nat64 -> Nat64) exp;
```

src/prelude/internals.mo
```diff
@@ -565,6 +558,10 @@ func @timer_helper() : async () {
 
   gatherExpired(@timers);
 
+  let exp = @nextExpiration @timers;
+  ignore (prim "global_timer_set" : Nat64 -> Nat64) exp;
+  if (exp == 0) @timers := null;
+
   for (k in thunks.keys()) {
     ignore switch (thunks[k]) { case (?thunk) ?thunk(); case _ null };
   }
```
crusso, line 567:
```suggestion
  for (o in thunks.vals()) {
     switch o { 
       case (?thunk) { ignore thunk() } 
       case _ { } 
    };
  }
```
(or use `let-else` and `continue` from `else` if you are brave)

ggreif: Yes I was pondering that. But `let-else` has a slight performance penalty due to evaluation the pattern _twice_ (currently).

PR #3888, opened 2023-03-20
Backend: `trap` PrimOP does not return

### LINKED ISSUE #3887: bug: possible miscompilation of switch in 0.8.4
Example code:

```
 import Prim "mo:⛔";
 
 actor {
   func f() {};
 
   func g(name : Text) : () -> () {
     switch (name) {
       case "some" f;
       case _ Prim.trap("Not found");
     };
   };
 
   public func init(name : Text) {
     g(name)();
   };
 };
```

Using a return keyword fixes it:

```
      case ("some") return f;
```

There is a difference between 'case ("some") f;' and 'case ("some") return f; '  (for 0.8.4).

The error message is "Canister v32cj-3iaaa-aaaaa-aaa2a-cai violated contract: function invocation does not match its signature". Does that mean that without the return keyword it tries to invoke f instead of returning f?

Reported by @timohanke  fyi @nomeata

nomeata:
fixes #3887

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -9504,7 +9504,7 @@ and compile_prim_invocation (env : E.t) ae p es at =
     IC.performance_counter env
 
   | OtherPrim "trap", [e] ->
-    SR.unit,
+    SR.Unreachable,
     compile_exp_vanilla env ae e ^^
     IC.trap_text env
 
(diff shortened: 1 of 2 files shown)


ggreif (approved):
Cool fix! Thanks a lot!

PR #3893, opened 2023-03-21
perf: optimise `from_word128_ptr` by speculation

ggreif:
...that the upper 64 bits of the cycle count are all zeros. Since 1T cycles is about one dollar, having 1 in the upper word will have a value more than 18M dollars. I guess that sending such amounts is rare.
However doing the bignum arithmetic to construct, shift and add the upper word is costly, see https://dfinity.github.io/canister-profiling/heartbeat/Motoko_heartbeat.svg

Of course the slow paths should be tested well :-) — Done by `test/run-drun/basic-cycles.mo`!

---------------
TODO
- [ ] We could explore an ultra-fast path too, when the whole amount is 0.

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -4441,22 +4441,24 @@ module Cycles = struct
     (fun env get_ptr ->
      let set_lower, get_lower = new_local env "lower" in
      get_ptr ^^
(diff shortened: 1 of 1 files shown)


kentosugama (approved):
LGTM

PR #3894, opened 2023-03-22
Special GC (Incremental GC with Central Object Table)

luc-blaeser:
Do not merge - this is only an experiment for comparison.

# Special GC (Incremental GC with Central Object Table)

Incremental, generational, and compacting garbage collector based on a central indirection table for object addresses.

**Objective**: An incremental GC with a different design than the evacuation-compacting GC (https://github.com/dfinity/motoko/pull/3837).
* No partitioned heap like in the evacuation-compact GC.
* Easily combinable with generational collection.
* Simple in-place full-heap compaction.
* No pointer update phase.

**Properties**:
* Fast reclamation of short-lived objects with generational collection.
* Full-heap incremental snapshot-at-the-beginning marking.
* Incremental compaction enabled by a central object table.

# Design

The key concept of this garbage collector is to introduce a central object table over which all references to an object are indirected. This allows atomic movement of objects without having to search and update their incoming pointers. Based on this mechanism, the GC can perform incremental heap compacting, similar to a simple mark-and-compact GC.

## Central Object Table

Each object has its object id that serves to reference the object. Thereby, all references are indirected via a central object table, that maps an object id (object reference) to the corresponding object address.
This enables fast moving of objects in the incremental GC by only updating the address of the corresponding object in the table. Objects also carry their id in the header to allow fast reverse lookup, i.e. determining the object id for a given object address. 
The table object is allocated in the heap and can also be moved (e.g. when growing). A global pointer denotes the current table location. Allowing relocation of the table is more expensive in terms of object address lookup (extra indirection via the global object table pointer), but allows a significantly simpler implementation, since the table can grow at any time without having to move other objects.
The ids of reclaimed garbage objects are recycled in the object table using an inlined free list.
When running out of free object ids, the table is currently relocated to double-sized new table at a different location in the heap. Pre-amortized growth could be used in future to obtain `O(1)` worst-case per insertion.

## Garbage Collection 

Garbage collection is performed in two steps:

### Young Generation Collection (Blocking)

The young generation collection runs non-incrementally, blocking the mutator. This is acceptable as the generation tends to be small and the GC work should adapt to the allocation rate

It is scheduled:
* Before each GC increment of the old generation, for simplifying incremental collection.
* After a certain amount of new allocations, i.e. when young generation has exceeded a threshold.

Young generation collection aims at fast reclamation of short-lived objects, to reduce GC latency, which is particularly relevant in an incremental GC. 

The young generation collection requires an extra root set of old-to-young pointers. Those pointers are caught by a write barrier and recorded in a remembered set. The remembered set lives in the young generation and is freed during the young generation collection.

The compaction phase can use the simple object movement enabled by the central object table provided for the incremental GC.

### Old Generation Collection (Incremental)

Old generation collection runs incrementally in multiple GC increments, where the mutator can resume work in between. It is scheduled to run after young generation collection, such that it can ignore generational aspects.

New objects are always first allocated in the young generation and become only visible to the incremental GC after they have been promoted to the old generation (i.e. survived the preceding young generation collection).

The incremental collection of the old generation is based on two phases:
1. **Mark**: Incremental snapshot-at-the-beginning marking. The GC marks at least all objects that have been reachable at the start time of the incremental GC run. A pre-update write-barrier detects relevant concurrent pointer overwrites by the mutator between GC increments and marks the corresponding objects to guarantee snapshot-at-the-beginning consistency. Concurrent allocations to the old generation (being promotions from the young generation) are conservatively marked.
2. **Compact**: Incremental compaction supported by the central object table. After the mark phase, the marked objects of the old generation are compacted towards the bottom of the generation. Objects can be easily moved one-by-one since all incoming references to a moved object can be atomically updated in the central object table. Concurrent allocations to the old generation during this phase (promotions from the young generations) are marked because they have to be retained and compacted during this incremental GC run.

GC increments are currently only scheduled on empty call stack such that we can ignore pointers on the call stack for the root set.

## In-Header Mark Bit
The GC uses a mark bit in the object header instead of mark bitmaps since there is no performance advantage by using the mark bitmap for skipping garbage objects. This is because the object ids of garbage objects need to be freed in the object table and therefore, the compaction phase must visit all objects (marked and unmarked).

## GC Configuration

* **Increment limit**: Regular increment bounded to 4,000,000 steps (approximately 400 million instructions).
* **Young generation collection**: When exceeding 8 MB young generation size.
* **Old generation collection**: When the old generation has grown more than 2.5 times its size or passed the critical limit of 3.5GB (on 4GB heap size).

# Measurement

The following results have been measured on the GC benchmark with `dfx` 0.12.1. `Special` denotes this new garbage collector, while `Incremental` here refers to the other incremental evacuation-compact GC (https://github.com/dfinity/motoko/pull/3837).

The `Copying`, `Compacting`, `Generational` GC are based on the original runtime system ***without*** the header extension (for object id). `No` denotes the disabled GC based on the runtime system also ***with*** the header extension. Measurement results are rounded to two significant figures.

## Scalability

**Summary**: Like the other incremental GC, the special GC scales up to the full heap size.

Average amount of allocations for the benchmark limit cases, until reaching a limit (instruction limit, heap limit, `dfx` cycles limit). 

| **Special**       | **140e6**               |
| Incremental       | 140e6                   |

## Performance

| **Special**       | **2.9e10**              |
| Incremental       | 2.0e10                  |  
| Generational      | 1.9e10                  | 
| Compacting        | 2.2e10                  | 
| Copying           | 2.0e10                  | 

45% slower than the other incremental and copying GC. 31% slower than the compacting GC. 52% slower than the basic generational GC.

## Memory Size

| **Special**       | **260 MB**              |
| Incremental       | 290 MB                  |
| No                | 500 MB                  |
| Generational      | 190 MB                  |
| Compacting        | 190 MB                  |
| Copying           | 270 MB                  |

12% smaller than the other incremental GC. 4% smaller than the copying GC. 37% larger than the compacting and generational GC.

## Overheads

* **Object table indirection**
    - Dereferencing requires an extra look up in the object table.
    - Two more instructions (`add` and `load`) for each field or array element access (and object header access).
* **Write barrier**: 
    - Pre-update barrier during the mark phase: Marking the target of overwritten pointers.
    - Post-update barrier: Recording old-to-young generation pointers in the remembered set.

Around 45% mutator overhead for the object table indirection and write barrier.

# Testing

The special GC is enabled by `--incremental-gc` compiler flag (like the other incremental GC).

1. RTS unit tests

    In Motoko repo folder `rts`:
    ```
    make test
    ```

2. Motoko test cases

    ```
    export EXTRA_MOC_ARGS="--sanity-checks --incremental-gc"
    make
    ```

3. GC Benchmark cases

    In `gcbench` repo: 
    ```
    ./measure-all.sh
    ```

# Discussion

## Advantage

* **Low-latency reclamation**: Fast memory reclamation due to the use of generational collection on top of the incremental collection. No update phase is needed in the incremental GC.

## Limitations

* **Object table growth**: Currently, for implementation simplicity, the object table is extended by copying the entries to a double-sized copy, which causes an unbound interruption in the incremental GC. This could be improved by using a pre-amortized table growth implementation.
* **Object table shrinking**: Due to the fragmentation, the object table cannot be easily compacted without having to update all references.
* **Indirection costs**: The performance costs of central object table iteration is noticeable by around 45%.
* **Iterating objects**: In contrast to other GCs using a mark bitmap, all free objects need to be also iterated for freeing the object id.

## Conclusion

The runtime costs of central object table indirection are not extremely high but relevant at around 45%. Moreover, the growth of the central object table is non-trivial. Both incremental GCs have similar code size complexity with around 5,200 - 5,500 LOC (additions and deletions): While the special GC avoids the partitioned heap, it introduces the central object table.

Except for the higher memory footprint, the incremental evacuation-compact GC shows superior performance according to the benchmark.

## Related PRs

* Incremental GC (based on evacuation-compaction): https://github.com/dfinity/motoko/pull/3837 
* Generational GC: https://github.com/dfinity/motoko/pull/3495


crusso:
I wonder if this approach would be much more viable if we could just use a second, isolated memory for the object table. Unfortunately, the replica doesn't support this I think, but wasm does. Or use stable memory, once fast enough, but probably not efficient enough in combination with regions. @luc-blaeser

PR #3896, opened 2023-03-22
Array index with constant index: Avoid going via vanilla representation

nomeata:
There is some benefit. Let’s see what the perf report will say
```diff
           local.get $s
-          i32.const 14
-          call $Array.idx_bigint
+          i32.const 7
+          call $Array.idx
           local.get $s
-          i32.const 14
-          call $Array.idx_bigint
+          i32.const 7
+          call $Array.idx
           i32.load offset=1
```
Fixes #3895.

PR #3897, opened 2023-03-23
Refactor: Common up array indexing code

nomeata:
Extracted from #3896, just some nice clean-up worth having.


src/codegen/compile.ml
```diff
@@ -8812,6 +8812,14 @@ let rec compile_lexp (env : E.t) ae lexp =
     SR.Vanilla,
     store_ptr
 
+(* Common code for a[e] as lexp and as exp.
+Traps or puts the pointer to the field on the stack
```
ggreif, line 8816:
```suggestion
Traps or pushes the pointer to the element on the stack
```

PR #3901, opened 2023-03-24
bugfix: compiler crash when matching on constant variants

### LINKED ISSUE #3899: backend: `destruct_const_pat` and constant variants
We currently have
``` OCaml
| AltP (p1, p2) -> destruct_const_pat ae p1 const
```
in `destruct_const_pat`, but this looks `p1`-biased. I think we could do better and catch exceptions for `p1` and retry with `p2`.

The possible exceptions originate from
``` OCaml
-  | TagP _ -> raise (Invalid_argument "TagP in static irrefutable pattern")
```
but we now have `Const` denotations for variants too, so we could have a compile-time pattern matching mechanism.

I am trying to come up with a patch.

ggreif:
This fixes a regression about `switch` on constant variants that crashes the compiler (the repro and symptoms are in https://github.com/dfinity/motoko/pull/3901/commits/e138d932b5de102c1910a25c45bf278890006a6e). The regression was introduced in #3878.

We now additionally allow certain refutable patterns in constant expression matches (`TagE` for now).

Fixes #3899.


src/codegen/compile.ml
```diff
@@ -10355,6 +10355,14 @@ and compile_const_decs env pre_ae decs : (VarEnv.t -> VarEnv.t) * (E.t -> VarEnv
         (fun env ae -> fill1 env ae; fill2 env ae) in
   go pre_ae decs
 
+and const_exp_matches_pat env pat exp : bool =
```
nomeata, line 10358: `pat_matches_const_exp`? Not sure who matches who…

test/run/variants.mo
```diff
@@ -55,3 +55,7 @@ assert (debug_show (#foo (#bar)) == "#foo(#bar)");
 
 assert ([#Monday, #Tuesday, #Wednesday, #Thursday, #Friday, #Saturday, #Sunday].size() == 7);
 
+// check compilation of pattern matching on constant variants
+let (#const a) = #const 42;
```
nomeata, line 59: Also a check for a failed constant match?

ggreif: I can add a test. But the new predicate should fail there too, and as such the old path will be taken.

ggreif: You are probably right, and the exceptions will be caught in the new predicate, but that also means that the runtime pattern match will be generated with the same problems :-(

To make this bottom-aware we need a three-way predicate.

src/codegen/compile.ml
```diff
@@ -10200,9 +10200,14 @@ and compile_dec env pre_ae how v2en dec : VarEnv.t * G.t * (VarEnv.t -> scope_wr
     G.( pre_ae1, nop, (fun ae -> fill env ae; nop), unmodified)
 
   (* A special case for constant expressions *)
-  | LetD (p, e) when Ir_utils.is_irrefutable p && e.note.Note.const ->
-    let extend, fill = compile_const_dec env pre_ae dec in
-    G.( extend pre_ae, nop, (fun ae -> fill env ae; nop), unmodified)
+  | LetD (p, e) when e.note.Note.const && (Ir_utils.is_irrefutable p || None <> const_exp_matches_pat env p e) ->
+    if Ir_utils.is_irrefutable p || Option.get (const_exp_matches_pat env p e) then
```
ggreif, line 10204: this needs refactoring

ggreif: ffcd2c9b08cf34d25f6eb75a75f6ee76b63bad28 is an attempt at that.

ggreif:
This still fails the line
```
  let (#a or _) = #b;
```
in `test/run/coverage.mo`, but that is due to the `or`-pattern not backtracking on failure (instead only considering the first alternative for constants).

ggreif:
We could extend this mechanism to other refutable patterns too: `Lit` and `Option`.

src/codegen/compile.ml
ggreif: Maybe @crusso can suggest a better naming.

nomeata:
Yes, indeed. I think I originally didn't do this because I expected it to be rare to have a statically known trap in the code, and wanted to avoid the complexity in rarely executed code paths. But now that it's there, no reason not to have it consistently (if the logic can be kept nicely encapsulated)

src/codegen/compile.ml
```diff
@@ -10188,6 +10190,19 @@ and compile_dec env pre_ae how v2en dec : VarEnv.t * G.t * (VarEnv.t -> scope_wr
...
+    lazy begin
+        let[@warning "-8"] LetD (p, e) = dec.it in
+        const_exp_matches_pat env p e
+      end,
+    lazy begin
+        let[@warning "-8"] LetD (p, _) = dec.it in
+        Ir_utils.is_irrefutable p
+      end in
+  let compile_time_matchable pred = Lazy.(force irrefutable_helper || pred (force const_exp_helper)) in
+
```
nomeata, line 10205: This looks awfully complicated to me. Isn't irrefutably just a special case of a sure-to-pass match?

So maybe have a single function for matching a pattern against a const value that returns the ternary result, and no longer treat irrefutably separately?

src/codegen/compile.ml
```diff
@@ -10188,6 +10190,19 @@ and compile_dec env pre_ae how v2en dec : VarEnv.t * G.t * (VarEnv.t -> scope_wr
   (fun (pre_ae, alloc_code, mk_code, wrap) ->
        G.(pre_ae, with_region dec.at alloc_code, fun ae body_code ->
           with_region dec.at (mk_code ae) ^^ wrap body_code)) @@
+
+  (* Set up some helpers, for exclusive usage by the "constant expressions" special case below *)
+  let const_exp_helper, irrefutable_helper =
```
ggreif, line 10195: I guess this could be a lazy pair instead of a pair of lazies.

src/codegen/compile.ml
nomeata: Also the higher order pred seems a bit roundabout. Maybe introduce a data type for will_match, will_fail or unknown, call it once, and then match on the result?

src/codegen/compile.ml
ggreif: done in bf1bbd2b0208a5af30f43dc007256437dfd999db

src/codegen/compile.ml
ggreif: PTAL!

nomeata (approved):

crusso (commented):
This code looks pretty complicated to me. Can we polish it a bit? 
The use of lazy to avoid a probably small repeated computation seems a bit convoluted.
The code might be cleaner if deconstruct_const_pat returned an optional VarEnv.t rather than using exceptions, but perhaps that affects too much other code?

ggreif:
Thanks for the comment. Yes, I am not very happy with this code, but I am working on two other PRs that will touch the area, so I prefer doing the cleanup a bit later. I was thinking making `compile_const_exp` to return an `Option` instead of crashing  on non-const. A `Some` result would then indicate that the pattern match should be performed at runtime, and then `destruct_const_pat` would give a definitive answer of having a match or bottom. (Currently `destruct_const_pat` throws — indicating bottom — where it should potentially match at compile time, see e.g. #3903).

ggreif:
I had a SGM with @crusso today. Turns out that _all_ const expressions should be matchable, so all the testing for irrefutable patterns and top-level constructors is unnecessary. In the course of a few PR this will be cleaned up.

PR #3903, opened 2023-03-26
compile-time match literals and options

ggreif:
This PR makes constant expressions fully matchable. We now also cover
- `LitP` and
- `OptP`

in the matching machinery. Thus there is no need to differentiate if the pattern is irrefutable.

Simplifies the logic quite a bit (as it should!)

-----------
TODO
- [x] tests for const `Opt` and `Lit`
- [x] bottoming cases too
- [x] check that `Opt` shortcutting works as intended — already in `iter-no-alloc.mo` for kernel present and `aardvark.mo` for kernel missing.

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -100,15 +100,17 @@ module Const = struct
     | Word64 of int64
     | Float64 of Numerics.Float.t
     | Blob of string
+    | Null
 
   let lit_eq = function
-    | (Vanilla i, Vanilla j) -> i = j
-    | (BigInt i, BigInt j) -> Big_int.eq_big_int i j
-    | (Word32 i, Word32 j) -> i = j
-    | (Word64 i, Word64 j) -> i = j
-    | (Float64 i, Float64 j) -> i = j
-    | (Bool i, Bool j) -> i = j
-    | (Blob s, Blob t) -> s = t
+    | Vanilla i, Vanilla j -> i = j
+    | BigInt i, BigInt j -> Big_int.eq_big_int i j
+    | Word32 i, Word32 j -> i = j
+    | Word64 i, Word64 j -> i = j
+    | Float64 i, Float64 j -> i = j
+    | Bool i, Bool j -> i = j
+    | Blob s, Blob t -> s = t
+    | Null, Null -> true
     | _ -> false
 
   (* Inlineable functions
@@ -7134,13 +7136,14 @@ module StackRep = struct
   *)
   let materialize_lit env (lit : Const.lit) : int32 =
     match lit with
-      | Const.Vanilla n  -> n
-      | Const.Bool n     -> Bool.vanilla_lit n
-      | Const.BigInt n   -> BigNum.vanilla_lit env n
-      | Const.Word32 n   -> BoxedSmallWord.vanilla_lit env n
-      | Const.Word64 n   -> BoxedWord64.vanilla_lit env n
-      | Const.Float64 f  -> Float.vanilla_lit env f
-      | Const.Blob t     -> Blob.vanilla_lit env t
+    | Const.Vanilla n  -> n
+    | Const.Bool n     -> Bool.vanilla_lit n
+    | Const.BigInt n   -> BigNum.vanilla_lit env n
+    | Const.Word32 n   -> BoxedSmallWord.vanilla_lit env n
+    | Const.Word64 n   -> BoxedWord64.vanilla_lit env n
+    | Const.Float64 f  -> Float.vanilla_lit env f
+    | Const.Blob t     -> Blob.vanilla_lit env t
+    | Const.Null       -> Opt.null_vanilla_lit env
 
   let rec materialize_const_t env (p, cv) : int32 =
     Lib.Promise.lazy_value p (fun () -> materialize_const_v env cv)
@@ -8157,7 +8160,7 @@ let nat64_to_int64 n =
   then sub_big_int n (power_int_positive_int 2 64)
   else n
 
-let const_lit_of_lit env : Ir.lit -> Const.lit = function
+let const_lit_of_lit : Ir.lit -> Const.lit = function
   | BoolLit b     -> Const.Bool b
   | IntLit n
   | NatLit n      -> Const.BigInt (Numerics.Nat.to_big_int n)
@@ -8170,13 +8173,13 @@ let const_lit_of_lit env : Ir.lit -> Const.lit = function
   | Int64Lit n    -> Const.Word64 (Big_int.int64_of_big_int (Numerics.Int_64.to_big_int n))
   | Nat64Lit n    -> Const.Word64 (Big_int.int64_of_big_int (nat64_to_int64 (Numerics.Nat64.to_big_int n)))
   | CharLit c     -> Const.Vanilla Int32.(shift_left (of_int c) 8)
-  | NullLit       -> Const.Vanilla (Opt.null_vanilla_lit env)
+  | NullLit       -> Const.Null
   | TextLit t
   | BlobLit t     -> Const.Blob t
   | FloatLit f    -> Const.Float64 f
 
-let const_of_lit env lit =
-  Const.t_of_v (Const.Lit (const_lit_of_lit env lit))
+let const_of_lit _env lit =
+  Const.t_of_v (Const.Lit (const_lit_of_lit lit))
 
 let compile_lit env lit =
   SR.Const (const_of_lit env lit), G.nop
@@ -10363,7 +10366,7 @@ and compile_const_exp env pre_ae exp : Const.t * (E.t -> VarEnv.t -> unit) =
       | _, Const.Array cs -> cs
       | _ -> fatal "compile_const_exp/ProjE: not a static tuple" in
     (List.nth cs i, fill)
-  | LitE l -> Const.(t_of_v (Lit (const_lit_of_lit env l))), (fun _ _ -> ())
+  | LitE l -> Const.(t_of_v (Lit (const_lit_of_lit l))), (fun _ _ -> ())
   | PrimE (TupPrim, []) -> Const.t_of_v Const.Unit, (fun _ _ -> ())
   | PrimE (ArrayPrim (Const, _), es)
   | PrimE (TupPrim, es) ->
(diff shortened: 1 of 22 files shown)


src/codegen/compile.ml
```diff
@@ -10384,34 +10393,39 @@ and const_exp_matches_pat env pat exp : bool option =
...
   | _ when Ir_utils.is_irrefutable pat -> Some true
+  | LitE _
   | PrimE (TagPrim _, _) ->
      let c, _ = compile_const_exp env VarEnv.empty_ae exp in
-     (try ignore (destruct_const_pat VarEnv.empty_ae pat c); Some true with Invalid_argument _ -> Some false)
+     (try ignore (destruct_const_pat env VarEnv.empty_ae pat c); Some true with Invalid_argument _ -> Some false)
   | _ -> None
 
-and destruct_const_pat ae pat const : VarEnv.t = match pat.it with
+and destruct_const_pat env ae pat const : VarEnv.t = match pat.it with
```
nomeata, line 10402: Why the `env` parameter? Better not have side effects (adding locals) here, if we end up not doing the constant match anyways. Maybe some function you need to call doesn't need the `env` parameter after all?

ggreif: I was hoping that too. `const_lit_of_lit` needs the `env` only for
``` OCaml
| NullLit       -> Const.Vanilla (Opt.null_vanilla_lit env)
```

nomeata: I'm ok with living with that side effect, the likelihood that this is the only null constant in the program is very high. So maybe leave a comment that this is the only side effect.

Or add a constructor `Const.Null`, and defer `null_vanilla_lit` to where where const values are made manifest? Probably cleaner anyways.

src/codegen/compile.ml
```diff
@@ -101,6 +101,15 @@ module Const = struct
     | Float64 of Numerics.Float.t
     | Blob of string
 
+  let lit_eq = function
+    | (Vanilla i, Vanilla j) -> i = j
+    | (BigInt i, BigInt j) -> Big_int.eq_big_int i j
+    | (Word32 i, Word32 j) -> i = j
+    | (Word64 i, Word64 j) -> i = j
+    | (Float64 i, Float64 j) -> i = j
```
ggreif, line 109: this seems fishy, maybe we shouldn't use `=` for this?

src/codegen/compile.ml
```diff
@@ -101,6 +101,15 @@ module Const = struct
     | Float64 of Numerics.Float.t
     | Blob of string
 
+  let lit_eq = function
+    | (Vanilla i, Vanilla j) -> i = j
```
ggreif, line 105: Do we have a guarantee that there is no duplication in the vanilla space?

src/codegen/compile.ml
nomeata: Just as fishy as matching on float literals in the first place?

src/codegen/compile.ml
nomeata: Can we get rid of this constructor? Is it used for more than `null`?

src/codegen/compile.ml
nomeata: Vanilla wouldn't be used for `null` anymore, would it?

src/codegen/compile.ml
ggreif: Phew!

src/codegen/compile.ml
ggreif: Yeah, hopefully.

src/codegen/compile.ml
ggreif: Does `Numerics.Float.t` provide a `=` at all?

src/codegen/compile.ml
ggreif: Looks like it worked: 84d59c8d0

src/codegen/compile.ml
```diff
@@ -7149,6 +7165,9 @@ module StackRep = struct
     | Const.Tag (i, c) ->
       let ptr = materialize_const_t env c in
       Variant.vanilla_lit env i ptr
+    | Const.Opt c ->
+      let ptr = materialize_const_t env c in
+      Opt.vanilla_lit env ptr
```
ggreif, line 7170: Use `ptr` if `c` is _not_ `Const.Opt` or `Const.Lit Null`.

ggreif: @nomeata this _should_ be the simplest shortcut, or am I mistaken?

ggreif: My solution is 709f0ad40.

test/run/refuted-const-option.mo
```diff
@@ -0,0 +1,5 @@
+// a failing pattern match that can be compiled to a trap
+let null = ?42;
```
ggreif, line 2: Sadly this does a runtime `if`. I have to figure out why.

src/codegen/compile.ml
```diff
@@ -8148,7 +8169,7 @@ let nat64_to_int64 n =
   then sub_big_int n (power_int_positive_int 2 64)
   else n
 
-let const_lit_of_lit env : Ir.lit -> Const.lit = function
+let const_lit_of_lit : Ir.lit -> Const.lit = function
```
nomeata, line 8172: Nice!

nomeata (commented):
`const.ml` still calls `Ir_utils.is_irrefutable p` - still needed?

src/codegen/compile.ml
```diff
@@ -10380,14 +10393,10 @@ and compile_const_decs env pre_ae decs : (VarEnv.t -> VarEnv.t) * (E.t -> VarEnv
...
+and const_exp_matches_pat env ae pat exp : bool =
   assert exp.note.Note.const;
-  match exp.it with
-  | _ when Ir_utils.is_irrefutable pat -> Some true
-  | PrimE (TagPrim _, _) ->
-     let c, _ = compile_const_exp env VarEnv.empty_ae exp in
-     (try ignore (destruct_const_pat VarEnv.empty_ae pat c); Some true with Invalid_argument _ -> Some false)
-  | _ -> None
+  let c, _ = compile_const_exp env ae exp in
+  (try ignore (destruct_const_pat VarEnv.empty_ae pat c); true with Invalid_argument _ -> false)
```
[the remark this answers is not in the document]
ggreif: 7ac29fd18

src/codegen/compile.ml
```diff
@@ -7150,6 +7166,11 @@ module StackRep = struct
       let ptr = materialize_const_t env c in
       Variant.vanilla_lit env i ptr
     | Const.Lit l -> materialize_lit env l
+    | Const.Opt c ->
+      let ptr = materialize_const_t env c in
+      match c with
+      | (_, Const.(Opt _ | Lit Null)) -> Opt.vanilla_lit env ptr
+      | _ -> ptr (* not option shaped, we can short-circuit *)
```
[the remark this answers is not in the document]
ggreif: Here is it: 9de688ebd

nomeata:
> Removing constraints from the const-producer will generate more of them, right? But this PR is mostly about consuming consts.

Hmm, right. With the current state of the PR, in
```
let #foo(x) = #foo(5);
x
```
the pattern match will be compiled statically, but `x` itself will not be const, whereas in
```
let (x,y) = (5,6);
x
```
it will.

But you are right, can probably come one after another.

src/codegen/compile.ml
ggreif: Okay, `test/run/refuted-const-float.mo` checks that `=` returns false, and I have manually checked that `let 3.14 = 3.14` is a `nop`. So this is settled.

src/codegen/compile.ml
```diff
@@ -8161,13 +8173,13 @@ let const_lit_of_lit env : Ir.lit -> Const.lit = function
...
   | CharLit c     -> Const.Vanilla Int32.(shift_left (of_int c) 8)
-  | NullLit       -> Const.Vanilla (Opt.null_vanilla_lit env)
+  | NullLit       -> Const.Null
   | TextLit t
   | BlobLit t     -> Const.Blob t
   | FloatLit f    -> Const.Float64 f
 
-let const_of_lit env lit =
-  Const.t_of_v (Const.Lit (const_lit_of_lit env lit))
+let const_of_lit _env lit =
```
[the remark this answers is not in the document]
ggreif: Done in 2dde40aa3

src/codegen/compile.ml
```diff
@@ -10219,15 +10231,13 @@ and compile_dec env pre_ae how v2en dec : VarEnv.t * G.t * (VarEnv.t -> scope_wr
...
-    let is_compile_time_bottom =
-      let lazy const_exp_matches = const_exp_helper in
-      not (Option.get const_exp_matches) in
-    if is_compile_time_bottom then (* refuted *)
-      (pre_ae, G.nop, (fun _ -> PatCode.patternFailTrap env), unmodified)
-    else (* not refuted *)
+  | LetD (p, e) when e.note.Note.const ->
+    (* constant expression matching with patterns is fully decidable *)
+    if const_exp_matches_pat env pre_ae p e then (* not refuted *)
       let extend, fill = compile_const_dec env pre_ae dec in
```
ggreif, line 10237: this is fine above, morally, already

src/codegen/compile.ml
```diff
@@ -10380,39 +10394,47 @@ and compile_const_decs env pre_ae decs : (VarEnv.t -> VarEnv.t) * (E.t -> VarEnv
...
-    end
+    let l = destruct_const_pat ae p1 const in
+    if l = None then destruct_const_pat ae p2 const
+    else l
   | TupP ps ->
-    let cs = match const with (_ , Const.Array cs) -> cs | (_, Const.Unit) -> [] | _ -> assert false in
-    List.fold_left2 destruct_const_pat ae ps cs
-  | LitP _ -> raise (Invalid_argument "LitP in static irrefutable pattern")
-  | OptP _ -> raise (Invalid_argument "OptP in static irrefutable pattern")
+    let cs = match const with (_, Const.Array cs) -> cs | (_, Const.Unit) -> [] | _ -> assert false in
```
ggreif, line 10418: What about `Array` sizes? Was it already refutable before?

kentosugama (approved):

PR #3904, opened 2023-03-27
Combining Tag and Forwarding Pointer (Incremental GC)

luc-blaeser:
# Combining Tag and Forwarding Pointer (Incremental GC)
Do not merge - this is an experiment of a design alternative for the Incremental GC: https://github.com/dfinity/motoko/pull/3837.

## Idea
Instead of reserving an additional header field for the forwarding pointer (Brooks pointer [1]), the forwarding pointer can be directly encoded in the tag because:
* Ordinary tag values are small values that do not denote valid object addresses.
* The tag of a forwarded object is no longer used since it is stored at the new object location.

Thanks to @ggreif for this truly interesting idea.

## Forwarding Resolution

The new logic involves a conditional branch as sketched by the following pseudo-code:

```
let tag = (*object.as_obj()).header.tag;
if tag <= MAX_TAG {
    // no forwarding
    object
} else {
    // forwarded, tag encodes the new location
    tag
}
```

For the sake of symmetry with the other pointer values and simplicity for the compiler implementation, the forwarding pointers is also represented as a skewed value in the tag, with a value > `MAX_TAG`, the maximum value of an ordinary tag.

## Other Implications

Some specific other RTS logic also relies on the tag to temporarily encode RTS state. These locations had to be adjusted:
* **Array slicing** during incremental marking: The GC temporarily recorded the next slice offset in the tag. This is now stored on the mark stack.
* **Stabilization** on canister upgrades: The tag of sharable objects is modified to store a marker (`StableSeen`) and a stream offset. This implementation had to be adjusted: `StableSeen` is specifically checked during stabilization. Moreover, stream offsets are encoded as scalar (unskewed) values, such that they can be distinguished from forwarding pointers. For efficiency, these specific additional cases are only tested during stabilization while keeping the standard forwarding pointer resolution as lean as possible. Alike the original incremental GC (https://github.com/dfinity/motoko/pull/3837), the GC is stopped when the stabilization starts and tags are modified to encode serialization state.

## Advantages

* **Smaller object sizes**: Saving one word per object, no header extension needed.
* **Faster object allocations**: No need for initializing the forwarding pointer field.

## Disadvantages

* **More expensive forwarding**: Need for a conditional forwarding branch instead of the unconditional jump as implemented in the classical Brooks pointer technique [1].
* **More complicated serialization***: The stabilization logic involves higher complexity in order not to interfere with the forwarding pointer.

## Performance 

As measured on the GC benchmark with `dfx` 0.13.1. The incremental GC with separate forwarding pointer is denoted "Original", while this version with the combined tag and forwarding pointer is denoted "New".

### Runtime

| Version           | Avg. Total Instructions |
| **New**           | **2.55e10**             |
| Original          | 2.00e10                 |

Performance degrade of **27.5%**.

### Memory

| **New**           | **288 MB**              |
| Original          | 294 MB                  |

Memory saving of **2%**

(The final heap size is nearly identical with <1% difference.)

### Allocations

Total allocated size (heap size + reclaimed size):

| **New**           | **466 MB**              |
| Original          | 502 MB                  |

Reduction by **7%**.

## Discussion

The original solution of classical Brooks pointer [1] implementation seems to be significantly faster, while only adding some limited memory overhead in the presence of the incremental GC.

Nevertheless, the idea of combining the tag and forwarding pointer seems interesting - and maybe this PR could still be optimized?

## Reference

[1] R. A. Brooks. Trading Data Space for Reduced Time and Code Space in Real-Time Garbage Collection on Stock Hardware. ACM Symposium on LISP and Functional Programming, LFP'84, New York, NY, USA, 1984.


luc-blaeser:
> It's surprising (to me) that the heap size is only 2% smaller. Were you expecting that? I would have thought shaving a word off every heap value would be a noticeable improvement.

Thank you for asking this.

Actually, comparing the memory size is a little bit difficult because it depends on the time when the GC is triggered which changes with the increased header sizes. 

Therefore, I now also added the metrics of total allocation size to the PR summary. The saving there is around 7% which is in line with my initial measurements when I experimented with the header extension and also compared the memory overhead again the classical current GCs (that do not have this extra header field). 

Of course, this is only the benchmark average and the results may vary for specific use cases. In the theoretical worst case, the additional header field could accumulate to 33% memory overhead if the program only creates smallest objects with minimum size 8 bytes (without an extra forwarding pointer field) and probably also needing an additional 4 byte reference from somewhere to the object to keep it alive, so 4 byte header overhead per 12 byte.

It seems however, that the difference is relatively moderate on average program cases.

PR #3906, opened 2023-03-29
feat: region serialization

crusso:
Tries to implement stable serialization of stateful Region objects consisting of three words
-  id: unboxed word 16 (padded to word32) (immutable)
- page_count: unboxed word 32 (mutable)
- vanilla Blob (mutable) (holding an array of allocated block ids)

These region objects are stateful descriptors of stable memory isolated regions. 

Unfortunately, the simplest example fails to upgrade properly.

```
[nix-shell:~/motoko/test/run-drun]$ more stable-regions-one.mo
import P "mo:⛔";
import Region "stable-region/Region";
import Region0 "stable-mem/StableMemory";

actor {
  stable var r0 = Region0.region();
  stable var r1 = Region.new();

  system func preupgrade() {
    P.debugPrint("pre");
  };

  system func postupgrade() {
    P.debugPrint("post");
  };

  public func sanityTest() {
  };
}

//SKIP run
//SKIP run-low
//SKIP run-ir
// too slow on ic-ref-run:
//SKIP comp-ref
//CALL ingress sanityTest "DIDL\x00\x00"
//CALL upgrade ""
//CALL ingress sanityTest "DIDL\x00\x00"
```

Produces:

```
[nix-shell:~/motoko/test/run-drun]$ ../run.sh -d stable-regions-one.mo
WARNING: Could not run ic-ref-run, will skip running some tests
stable-regions-one: [tc] [comp] [valid] [drun-run]
--- stable-regions-one.drun-run (expected)
+++ stable-regions-one.drun-run (actual)
@@ -0,0 +1,6 @@
+debug.print: pre
+ingress Err: IC0503: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped explicitly: read_alias unexpected idl_typ
Some tests failed:
stable-regions-one.mo
```
@nomeata any clue what I'm doing wrong here?

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -6311,12 +6311,12 @@ module MakeSerialization (Strm : Stream) = struct
           begin
           get_arg_typ ^^
           compile_eq_const (Int32.neg (Option.get (to_idl_prim (Prim Region)))) ^^
-          E.else_trap_with env "read_alias unexpected idl_typ" ^^ (* FAILS! *)
+          E.else_trap_with env "read_alias unexpected idl_typ" ^^
           Region.alloc env (compile_unboxed_const 0l) (compile_unboxed_const 0l) (Blob.lit env "") ^^ set_region ^^
           on_alloc get_region ^^
           get_region ^^ ReadBuf.read_word32 env get_data_buf ^^ Heap.store_field Region.id_field ^^
           get_region ^^ ReadBuf.read_word32 env get_data_buf ^^ Heap.store_field Region.page_count_field ^^
-          get_region ^^ (* with_blob_typ env*) (read_blob ()) ^^ Heap.store_field Region.vec_pages_field ^^
+          get_region ^^ read_blob () ^^ Heap.store_field Region.vec_pages_field ^^
           get_region ^^ Region.check_region "read_alias" env
           end
         )
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -3586,18 +3586,21 @@ module Region = struct
   let alloc env get_id get_pagecount get_vec_pages =
     Tagged.obj env Tagged.Region [ get_id; get_pagecount; get_vec_pages ]
 
-  let check_region s env =
-    let (set_region, get_region) = new_local env "region" in
-    set_region ^^
-    get_region ^^ Tagged.load ^^
-    compile_eq_const Tagged.(int_of_tag Region) ^^
-    get_region ^^ Tagged.load ^^
-    compile_eq_const Tagged.(int_of_tag StableSeen) ^^
-    G.i (Binary (Wasm.Values.I32 I32Op.Or)) ^^
-    E.else_trap_with env ("Internal error: bad region tag "^ s) ^^
-    get_region ^^ Heap.load_field vec_pages_field ^^ Tagged.load ^^
-    compile_eq_const Tagged.(int_of_tag Blob) ^^
-    E.else_trap_with env ("Internal error: bad region.vec_pages" ^ s)
+  let sanity_check s env =
+    if !Flags.sanity then
+    Func.share_code1 env ("check_region_" ^ s) ("val", I32Type) [I32Type]
+      (fun env get_region ->
+         get_region ^^ Tagged.load ^^
+         compile_eq_const Tagged.(int_of_tag Region) ^^
+         get_region ^^ Tagged.load ^^
+         compile_eq_const Tagged.(int_of_tag StableSeen) ^^
+         G.i (Binary (Wasm.Values.I32 I32Op.Or)) ^^
+         E.else_trap_with env ("Internal error: bad region tag "^ s) ^^
+         get_region ^^ Heap.load_field vec_pages_field ^^ Tagged.load ^^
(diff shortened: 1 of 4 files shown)


src/codegen/compile.ml
```diff
@@ -6253,6 +6288,19 @@ module MakeSerialization (Strm : Stream) = struct
             remember_failure get_val ^^
             get_val ^^ store_ptr
           )
+          )
+      | Prim Region ->
+         read_alias env (Prim Region) (fun get_arg_typ on_alloc ->
+          let (set_region, get_region) = new_local env "region" in
+          get_arg_typ ^^
```
matthewhammer, line 6295: `get_arg_typ` feels misplaced here.  Given the name and other uses, my hunch is that it's only meaningful when the type we are reading is parameterized by a type argument, and `Region` is a primitive with no arguments.

crusso: Might be, but I think arg_type might refer to the argument of the idl_alias constructor which is inserted to indicate the preservation of sharing. Worth investigating though. I only added the equality test as a sanity check but disabling it leads to other failures.

matthewhammer: > disabling it leads to other failures

I noticed those too.  The `check_region` routine fails for me, saying that `vec_pages` is bad.

nomeata: Woudln’t call it `get_arg_typ`, it’s more `get_idl_typ`. (If it were composite, we could extract the arg type from it, as in `Array Mut` above).

src/codegen/compile.ml
```diff
@@ -5299,7 +5331,7 @@ module MakeSerialization (Strm : Stream) = struct
       | Prim Blob ->
         add_typ Type.(Array (Prim Nat8))
       | Prim Region ->
-        add_typ (Type.Prim Region)
+        add_sleb128 idl_alias; add_idx t
```
nomeata, line 5334: This is a bit fishy: `add_typ` isn’t used for primitive type before (see below), but I think you wanted to make `Region` primitive?

But it needs the alias type tag _and_ something to put int. So far, these are always different types `Array (Mut t)` vs. `Array t`; `Mut t` vs `t`. Here you want to use `Prim Region` for both the aliased thing and the thing inside, I guess. So maybe fine.

src/codegen/compile.ml
```diff
@@ -6253,6 +6288,19 @@ module MakeSerialization (Strm : Stream) = struct
...
+         read_alias env (Prim Region) (fun get_arg_typ on_alloc ->
+          let (set_region, get_region) = new_local env "region" in
+          get_arg_typ ^^
+          compile_eq_const (Int32.neg (Option.get (to_idl_prim (Prim Region)))) ^^
+          E.else_trap_with env "read_alias unexpected idl_typ" ^^ (* FAILS! *)
+          Region.alloc env (compile_unboxed_const 0l) (compile_unboxed_const 0l) (Blob.lit env "") ^^ set_region ^^
+          on_alloc get_region ^^
+          get_region ^^ ReadBuf.read_word32 env get_data_buf ^^ Heap.store_field Region.id_field ^^
+          get_region ^^ ReadBuf.read_word32 env get_data_buf ^^ Heap.store_field Region.page_count_field ^^
+          get_region ^^ with_blob_typ env (read_blob ()) ^^ Heap.store_field Region.vec_pages_field ^^
```
nomeata, line 6302: Why `with_blob_typ`? This tries to read bytes off the type table, but I don’t think you are explicitly adding an entry for the `blob` type to the type table (in `add_typ`), do you? (I don’t think you should, but then remove the `with_blob_typ` here).

crusso:
@matthewhammer, ok, I think this is working enough to merge in.

src/codegen/compile.ml
```diff
@@ -6253,6 +6313,19 @@ module MakeSerialization (Strm : Stream) = struct
...
             get_val ^^ store_ptr
           )
+          )
+      | Prim Region ->
+         read_alias env (Prim Region) (fun get_region_typ on_alloc ->
+          let (set_region, get_region) = new_local env "region" in
+          (* sanity check *)
+          get_region_typ ^^
+          compile_eq_const (Int32.neg (Option.get (to_idl_prim (Prim Region)))) ^^
+          E.else_trap_with env "read_alias unexpected idl_typ" ^^
```
nomeata, line 6323:
```suggestion
          E.else_trap_with env "deserialize_go (Region): unexpected idl_typ" ^^
```

nomeata (commented):
Might be fine, if it works :-)

matthewhammer:
@crusso I still get errors when I run the unit tests, like this

```
ingress Err: IC0503: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped explicitly: read_alias unexpected idl_typ
```

crusso:
@matthewhammer https://github.com/dfinity/motoko/pull/3906/commits/0bbea9b00cc5371bcabbab830140345f55a128fd is a test that shows what goes wrong if we expose region0 and don't take extra steps to ensure all region0 objects are identified.
One fix is to not exposing region0 at all, the second might be to ensure correct aliasing for region0 on deserialization. It think the former might be cleanest.

Issue #3910, opened 2023-03-31
bug: crash compiling in ledger proxy.mo


crusso:
Hi! The compiler tells me that I've found a bug:
```
OOPS! You've triggered a compiler bug.
Please report this at https://github.com/dfinity/motoko/issues/new with the following details:
Motoko 0.8.5 (source 5lid2g0r-27a8ks7s-siak993r-7saawzd8)
Fatal error: exception File "ir_passes/await.ml", line 148, characters 4-10: Assertion failed
Raised at Ir_passes__Await.t_exp' in file "ir_passes/await.ml", line 148, characters 4-50
```
Full log: https://dash.idx.dfinity.network/invocation/634853e9-0fdd-4ad3-888d-f0bc0fc238ef?target=%2F%2Frs%2Frosetta-api%2Ficp_ledger%3Aledger_proxy
That's the source:
https://sourcegraph.com/github.com/dfinity/ic@59e2a295149b1010cd9e79e4dacde13828b087bc/-/blob/rs/rosetta-api/icp_ledger/wasm/Proxy.mo

https://dfinity.slack.com/archives/CPVUFG8UV/p1680034435578979

faragly:
I am getting similar error text with moc 0.8.6
```
Stderr:
OOPS! You've triggered a compiler bug.
Please report this at https://github.com/dfinity/motoko/issues/new with the following details:

Motoko 0.8.6 (source 4kx0ibq8-zh7m746m-8f29d3sk-zbhm0vj1)

Fatal error: exception File "ir_passes/await.ml", line 148, characters 4-10: Assertion failed
Raised at Ir_passes__Await.t_exp' in file "ir_passes/await.ml", line 148, characters 4-50
Called from Ir_passes__Await.t_exp in file "ir_passes/await.ml", line 83, characters 18-36
Called from Ir_passes__Await.t_dec' in file "ir_passes/await.ml", line 181, characters 34-51
Called from Ir_passes__Await.t_dec in file "ir_passes/await.ml", line 178, characters 17-38
Called from Stdlib__list.map in file "list.ml", line 92, characters 20-23
Called from Stdlib__list.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__list.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__list.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__list.map in file "list.ml", line 92, characters 32-39
[...]

PR #3911, opened 2023-03-31
bugfix: issue #3910

crusso:
Fixes a regression introduced by the `async*`/`await*` feature and changes to the await translation.
The bug was trigged by a pattern matching local `async` function. `Desugar.ml` pushes the pattern match into the body of the function, but outside the `async` expression, obscuring the `async` body from `await.ml`.

The fix here is to push the pattern match into the body of the async instead during lowering, as we do for shared functions already. Unfortunately, this does change the semantics - delaying the pattern match (failure) to the async computation where it is reported as an error result, not an eager trap - do we care?

Update: I think we would care since the ast interpreter and ir semantics won't agree unless we 
- [ ]  reject refutable pattern in local async functions.

Or figure out a better solution.

- [ ] minimize test and write some more, including eval


crusso:
Superseded by #3916

PR #3913, opened 2023-03-31
Separate RTS Builds (Incremental and Non-Incremental GC)

luc-blaeser:
# Separate RTS Builds (Incremental and Non-Incremental GC)

Using different memory layouts determined at compile time:
* Incremental GC: 
   - Extended header with forwarding pointer field.
   - Partitioned heap.
* Non-incremental GC (copying, compacting, and generational GC):
   - Small header only comprising the object tag.
   - Linear heap.

## Runtime System Changes

Separate RTS builds by introducing the feature `"incremental_gc"` for incremental GC memory layout.

Helper macros:
* `#[incremental_gc]`: macro attribute, equivalent to `#[cfg(feature = "incremental_gc")]`
* `#[non_incremental_gc]`: macro attribute, equivalent to `#[cfg(not(feature = "incremental_gc"))]`
* `is_incremental_gc!()`: procedure macros, equivalent to `cfg!(feature = "incremental_gc")`

Different builds:
* `rts.wasm`: Release build with non-incremental GCs (containing the copying, compacting, and generational GC).
* `rts-debug.wasm`: Debug build with non-incremental GCs (containing the copying, compacting, and generational GC).
* `rts-incremental.wasm`: Release build with only the incremental GC (no other GCs).
* `rts-incremental-debug.wasm`: Debug build with only the incremental GC (no other GCs).

## Compiler Changes

GC-dependent compilation:
* Conditional header layout with or without forwarding pointer.
* Linking the corresponding matching RTS build, with build-specific imports.

Switch based on the condition `!Flags.gc_strategy == Flags.Incremental`.

## Performance

Comparing the the following designs:

* **Non-Incremental RTS**: Original RTS in `master branch` without incremental GC changes.
* **Combined RTS**: Combining incremental and non-incremental GC in one RTS build, PR: https://github.com/dfinity/motoko/pull/3837
* **Separate RTS**: This PR.

### Binary Size

Size of the release RTS binary files.

| GC              | Non-Incremental RTS | Combined RTS | Separate RTS |
| --------------- | --------------------| ------------ | ------------ |
| non-incremental | 174 KB              | 194 KB       | 174 KB       |
| incremental     | -                   | 194 KB       | 175 KB       | 

**11%** reduction.

### Total allocations

GC benchmark results with `dfx 0.13.1`. 
Total amount of allocated memory (heap size + reclaimed memory) at runtime, average across benchmark cases.

| GC           | Non-Incremental RTS  | Combined RTS | Separate RTS |
| ------------ | -------------------- | ------------ | ------------ |
| copying      | 459 MB               | 496 MB       | 459 MB       |
| compacting   | 459 MB               | 496 MB       | 459 MB       |
| generational | 480 MB               | 517 MB       | 480 MB       |
| incremental  | -                    | 502 MB       | 502 MB       |

For non-incremental GCs: **8%** reduction compared to the combined RTS, same like non-incremental RTS.

### Memory Size

GC benchmark results with `dfx 0.13.1`. 
Allocated WASM memory size at runtime, average across benchmark cases.

| GC           | Non-Incremental RTS | Combined RTS | Separate RTS |
| ------------ | ------------------- | ------------ | ------------ |
| copying      | 271 MB              | 281 MB       | 271 MB       |
| compacting   | 188 MB              | 195 MB       | 188 MB       |
| generational | 191 MB              | 201 MB       | 194 MB       |
| incremental  | -                   | 294 MB       | 294 MB       |

For non-incremental GCs: **4%** reduction compared to the combined RTS, same like non-incremental RTS.

### Total Instructions

GC benchmark results with `dfx 0.13.1`. 
Number of executed instructions, average across benchmark cases. 

| GC           | Non-Incremental RTS | Combined RTS | Separate RTS |
| ------------ | ------------------- | ------------ | ------------ |
| copying      | 2.05e10             | 2.12e10      | 2.07e10      |
| compacting   | 2.20e10             | 2.24e10      | 2.21e10      |
| generational | 1.91e10             | 1.93e10      | 1.92e10      |
| incremental  | -                   | 1.95e10      | 1.95e10      |

For non-incremental GCs: Around **2%** reduction compared to combined RTS, around *1%** overhead compared to non-incremental RTS.

# Conclusion

Advantages of this PR:
* Avoiding performance degrades for the classical GCs by introducing the incremental GC support.
* Smaller binary sizes.


crusso:
Oh, this is great - thanks for doing it. I guess I'll review this after the incremental GC?

rts/motoko-rts-macros/src/lib.rs
```diff
@@ -136,3 +136,34 @@ pub fn ic_mem_fn(attr: TokenStream, input: TokenStream) -> TokenStream {
     )
     .into()
 }
+
+/// Feature macro for incremental GC features, in particular forwarding pointers.
+/// Equivalenyt to using the attribute `#[cfg(feature = "incremental_gc")]`.
```
crusso, line 141:
```suggestion
/// Equivalent to using the attribute `#[cfg(feature = "incremental_gc")]`.
```

rts/motoko-rts-macros/src/lib.rs
```diff
@@ -136,3 +136,34 @@ pub fn ic_mem_fn(attr: TokenStream, input: TokenStream) -> TokenStream {
...
+    let block = syn::parse_macro_input!(input as syn::Item);
+    quote!(
+        #[cfg(feature = "incremental_gc")]
+        #block
+    )
+    .into()
+}
+
+/// Feature macro for non-incremental GC features, in particular forwarding pointers.
+/// Equivalenyt to using the attribute `#[cfg(not(feature = "incremental_gc"))]`.
```
crusso, line 154:
```suggestion
/// Equivalent to using the attribute `#[cfg(not(feature = "incremental_gc"))]`.
```

luc-blaeser: Thanks!

luc-blaeser:
> Oh, this is great - thanks for doing it. I guess I'll review this after the incremental GC?

Thank you also very much for reviewing this growing pile of large PRs. I also think it makes sense to review it after the incremental GC. If the PR is considered useful, it would need to be split and merged with the different PRs of the incremental GC, namely the forwarding pointer PR (https://github.com/dfinity/motoko/pull/3829) and the incremental GC PR (https://github.com/dfinity/motoko/pull/3837).

rts/Makefile
```diff
@@ -260,30 +288,45 @@ EXPORTED_SYMBOLS=\
...
 	cargo fmt --verbose --manifest-path motoko-rts-tests/Cargo.toml
+	cargo fmt --verbose --manifest-path motoko-rts-macros/Cargo.toml
 
 clean:
 	rm -rf \
 	  _build \
 	  mo-rts.wasm \
 	  mo-rts-debug.wasm \
+    mo-rts-incremental.wasm \
+	  mo-rts-incremental-debug.wasm \
```
crusso, line 328:
```suggestion
	  mo-rts-incremental.wasm \
	  mo-rts-incremental-debug.wasm \
```

rts/Makefile
```diff
@@ -260,30 +288,45 @@ EXPORTED_SYMBOLS=\
...
 	  motoko-rts/target \
-	  motoko-rts-tests/target \
+    motoko-rts-tests/target \
+	  motoko-rts-macros/target \
```
crusso, line 331:
```suggestion
	  motoko-rts-tests/target \
	  motoko-rts-macros/target \
```

rts/Makefile
crusso: Wrong indentation?

crusso (changes requested):
Looks good to me. Thanks for pushing this through.

I guess this increases the complexity of the code a little, but avoiding the regression for the older gc seems worth it to me. Do you agree or do you think we can just live with the regression? If the maintenance burden is too high, please say so.

Made some minor suggestions.

rts/Makefile
luc-blaeser: Thanks!

rts/motoko-rts-tests/src/gc.rs
```diff
@@ -5,36 +5,27 @@
 //
 // To convert an offset into an address, add heap array's address to the offset.
 
+#[cfg(not(feature = "incremental_gc"))]
 mod compacting;
+#[cfg(not(feature = "incremental_gc"))]
 mod generational;
 mod heap;
+#[cfg(feature = "incremental_gc")]
```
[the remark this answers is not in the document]
luc-blaeser: Actually, I did not import the macro in the RTS unit tests, only in the main runtime system. Except for three source location, it is consistently used in about 27 locations in the RTS build. I will try to also import it in the RTS unit tests.

rts/motoko-rts/src/types.rs
```diff
@@ -446,16 +460,32 @@ pub const TAG_ARRAY_SLICE_MIN: Tag = 32;
...
 
 impl Obj {
+    #[incremental_gc]
+    pub fn init_forward(&mut self, value: Value) {
+        self.forward = value;
+    }
+
+    #[non_incremental_gc]
+    pub fn init_forward(&mut self, _value: Value) {}
+
```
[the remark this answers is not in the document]
luc-blaeser: The Rust compiler usually inlines quite eagerly (regardless of the compile feature attribute). I will double check this.

rts/motoko-rts-tests/src/gc.rs
```diff
@@ -424,24 +433,25 @@ fn check_continuation_table(mut offset: usize, continuation_table: &[ObjectIdx],
 }
 
 impl GC {
-    fn run(&self, heap: &mut MotokoHeap, round: usize) -> bool {
-        let heap_base = heap.heap_base_address() as u32;
+    fn run(&self, heap: &mut MotokoHeap, _round: usize) -> bool {
+        let _heap_base = heap.heap_base_address() as u32;
```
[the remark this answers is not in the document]
luc-blaeser: Yes, this is due to the conditional compilation of the two GC modes, where some variables/parameters are unused in certain RTS builds.

src/codegen/compile.ml
```diff
@@ -1769,29 +1782,34 @@ module Tagged = struct
...
+    let non_incremental_gc_data = tag ^ payload in
     let forward = bytes_of_int32 skewed_ptr in (* forwarding pointer *)
-    let data = tag ^ forward ^ payload in
-    E.write_static_memory env unskewed_ptr data;
+    let incremental_gc_data = tag ^ forward ^ payload in
+    (if !Flags.gc_strategy = Flags.Incremental then
+      E.write_static_memory env unskewed_ptr incremental_gc_data
+    else
+      E.write_static_memory env unskewed_ptr non_incremental_gc_data
+    );
```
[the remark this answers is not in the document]
luc-blaeser: Thank you for this suggestion. I have improved this.

src/codegen/compile.ml
```diff
@@ -1816,9 +1834,10 @@ module MutBox = struct
        └──────┴─────┴─────────┘
 
      The object header includes the obj tag (MutBox) and the forwarding pointer.
+     The forwarding pointer is only reserved if compiled for the incremental GC.
   *)
 
-  let field = Tagged.header_size
+  let field env = Tagged.header_size env
```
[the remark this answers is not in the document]
luc-blaeser: Thank you. I think it is simpler with your suggestion. (We can still adjust if needed.)

luc-blaeser:
> Looks good to me. Thanks for pushing this through.
> 
> I guess this increases the complexity of the code a little, but avoiding the regression for the older gc seems worth it to me. Do you agree or do you think we can just live with the regression? If the maintenance burden is too high, please say so.
> 
> Made some minor suggestions.

Thank you very much for also reviewing this fourth PR. I believe the complexity could be justified in this case, and for that reason, I would limit it to having only the two RTS builds like in this PR. The alternative would be to have a separate RTS assembly per GC, which would be too fine-granular and of limited additional performance benefit (runtime and binary size). If we see that the maintenance is a burden, we could still remove this separation again. Also, when we make the incremental GC the default GC in future, we should simplify it again to one RTS build.

rts/motoko-rts/src/types.rs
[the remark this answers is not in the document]
luc-blaeser: *Update*: I double checked and the functions get inlined in the release builds.

src/codegen/compile.ml
```diff
@@ -1586,10 +1591,15 @@ module Tagged = struct
...
+  (* Declare `env` for lazy computation of the header size when the compile environment with compile flags are defined *)
+  let header_size env = 
+    if !Flags.gc_strategy == Flags.Incremental then 2l else 1l
+  
   (* The tag *)
-  let header_size = 2l
   let tag_field = 0l
-  let forwarding_pointer_field = 1l
+  let forwarding_pointer_field = 
+    assert (!Flags.gc_strategy != Flags.Incremental);
```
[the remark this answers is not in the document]
luc-blaeser: Thanks for noticing this. I improved this by adding `env` as argument.

rts/motoko-rts-tests/src/gc.rs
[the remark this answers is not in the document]
luc-blaeser: **Update**: I refactored the RTS unit tests to also use the same GC-dependent procedure macros like in the RTS.

rts/motoko-rts-tests/src/gc.rs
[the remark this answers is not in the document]
luc-blaeser: **Update**: I also refactored this a little bit, to avoid the underscrore prefix for the local variables.

src/codegen/compile.ml
```diff
@@ -1026,11 +1036,10 @@ module RTS = struct
     E.add_func_import env "rts" "stream_shutdown" [I32Type] [];
     E.add_func_import env "rts" "stream_reserve" [I32Type; I32Type] [I32Type];
     E.add_func_import env "rts" "stream_stable_dest" [I32Type; I64Type; I64Type] [];
-    E.add_func_import env "rts" "post_write_barrier" [I32Type] [];
-    E.add_func_import env "rts" "write_with_barrier" [I32Type; I32Type] [];
-    E.add_func_import env "rts" "allocation_barrier" [I32Type] [I32Type];
-    E.add_func_import env "rts" "stop_gc_on_upgrade" [] [];
-    E.add_func_import env "rts" "running_gc" [] [I32Type];
+    if !Flags.gc_strategy == Flags.Incremental then
```
crusso, line 1039:
```suggestion
    if !Flags.gc_strategy = Flags.Incremental then
```
`==` is Ocaml's physical equality

src/codegen/compile.ml
```diff
@@ -1586,10 +1591,15 @@ module Tagged = struct
     | CoercionFailure -> 0xfffffffel
     | StableSeen -> 0xffffffffl
 
+  (* Declare `env` for lazy computation of the header size when the compile environment with compile flags are defined *)
+  let header_size env = 
+    if !Flags.gc_strategy == Flags.Incremental then 2l else 1l
```
crusso, line 1596:
```suggestion
    if !Flags.gc_strategy = Flags.Incremental then 2l else 1l
```
use structural equality

src/codegen/compile.ml
```diff
@@ -1600,15 +1610,18 @@ module Tagged = struct
       set_object ^^ get_object ^^
       compile_unboxed_const (int_of_tag tag) ^^
       Heap.store_field tag_field ^^
-      get_object ^^ (* object pointer *)
-      get_object ^^ (* forwarding pointer *)
-      Heap.store_field forwarding_pointer_field ^^
+      (if !Flags.gc_strategy == Flags.Incremental then
```
crusso, line 1613:
```suggestion
      (if !Flags.gc_strategy == Flags.Incremental then
```
```suggestion
      (if !Flags.gc_strategy = Flags.Incremental then
```

crusso, line 1613:
```suggestion
      (if !Flags.gc_strategy = Flags.Incremental then
```
use structural equality

crusso (approved):
But don't forget to fix those uses of physical equality

src/codegen/compile.ml
luc-blaeser: Thank you very much. I fixed these cases, including one that I had introduced in `https://github.com/dfinity/motoko/pull/3837`.

src/codegen/compile.ml
luc-blaeser: Thank you very much. I fixed this (including the other cases).

src/codegen/compile.ml
```diff
@@ -1586,10 +1591,15 @@ module Tagged = struct
...
+  (* Declare `env` for lazy computation of the header size when the compile environment with compile flags are defined *)
+  let header_size env = 
+    if !Flags.gc_strategy == Flags.Incremental then 2l else 1l
+  
   (* The tag *)
-  let header_size = 2l
   let tag_field = 0l
-  let forwarding_pointer_field = 1l
+  let forwarding_pointer_field env = 
+    assert (!Flags.gc_strategy == Flags.Incremental);
```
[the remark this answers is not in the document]
luc-blaeser: Thank you very much. I fixed this (including the other cases).

src/codegen/compile.ml
luc-blaeser: Thank you very much. I fixed this (including the other cases).

PR #3916, opened 2023-04-03
bugfix: fix for #3910 (bad compilation of local async functions that pattern match on arguments)

crusso:


src/ir_passes/await.ml
```diff
@@ -127,9 +127,14 @@ and t_exp' context exp =
   | DefineE (id, mut ,exp1) ->
     DefineE (id, mut, t_exp context exp1)
   | FuncE (x, (T.Local as s1), c, typbinds, pat, typs,
-      ({ it = AsyncE _; _} as body)) ->
+      ({ it = AsyncE _; _} as async)) ->
     FuncE (x, s1, c, typbinds, pat, typs,
-      t_async context body)
+      t_async context async)
+  | FuncE (x, (T.Local as s1), c, typbinds, pat, typs,
+      ({it = BlockE (ds, ({ it = AsyncE _; _} as async)); _} as wrapper))->
```
nomeata, line 134: Is this doing the right thing even if the user writes a source function that matches that shape (a local function with a block and an `async` block in tail position)?

crusso: Yeah, I worried about that, but I don't think they can by typing, right?

crusso: It would actually be nice if desugaring didn't push the pattern match into the body but somehow kept the pattern as part of the function ast, but that's not the current setup.

nomeata: Yeah, feels bad to have these non-compositional invariants around, but for now, LGTM

nomeata (approved):

src/ir_passes/await.ml
ggreif, line 134:
```suggestion
      ({it = BlockE (ds, ({ it = AsyncE _; _} as async)); _} as wrapper)) ->
```

ggreif: Does this also cover `func () : async () = async { .. }`-style definitions?

crusso: I'll add a test

src/ir_passes/await.ml
```diff
@@ -127,9 +127,14 @@ and t_exp' context exp =
   | DefineE (id, mut ,exp1) ->
     DefineE (id, mut, t_exp context exp1)
   | FuncE (x, (T.Local as s1), c, typbinds, pat, typs,
-      ({ it = AsyncE _; _} as body)) ->
+      ({ it = AsyncE _; _} as async)) ->
```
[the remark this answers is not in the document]
crusso: Yeah, bit gratuitous here so I've removed it (and renamed below).

src/ir_passes/await.ml
crusso: Done

crusso: > Yeah, feels bad to have these non-compositional invariants around, but for now, LGTM

Agreed. But if you think this one's bad, you should see the existing case for oneways below.

PR #3918, opened 2023-04-03
feat(mo-doc): nested modules

### LINKED ISSUE #1727: [mo-doc] Documentation isn't generated for nested modules
Shouldn't be too tricky to do, we can probably follow how classes nest declarations.

rvanasa:
Renders nested modules in the HTML, ADoc, and Markdown output formats. 

Fixes #1727.

### DIFF
--- a/src/docs/extract.ml
+++ b/src/docs/extract.ml
@@ -138,7 +138,7 @@ struct
           _;
         } -> (
         match rhs with
-        | Source.{ it = Syntax.ObjE (sort, fields); _ } ->
+        | Source.{ it = Syntax.ObjBlockE (sort, fields); _ } ->
             let mk_field_xref xref = mk_xref (Xref.XClass (name, xref)) in
             Some
               ( mk_xref (Xref.XType name),
--- a/src/docs/extract.ml
+++ b/src/docs/extract.ml
@@ -12,7 +12,7 @@ and declaration_doc =
   | Value of value_doc
   | Type of type_doc
   | Class of class_doc
-  | Module of module_doc
+  | Object of object_doc
   | Unknown of string
 
 and function_doc = {
@@ -50,7 +50,7 @@ and class_doc = {
   sort : Syntax.obj_sort;
 }
 
-and module_doc = { name : string; fields : doc list }
+and object_doc = { name : string; fields : doc list; sort : Syntax.obj_sort }
 
 let un_prog prog =
   let comp_unit = Mo_def.CompUnit.comp_unit_of_prog true prog in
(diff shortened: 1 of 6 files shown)


src/docs/extract.ml
```diff
@@ -12,6 +12,7 @@ and declaration_doc =
   | Value of value_doc
   | Type of type_doc
   | Class of class_doc
+  | Module of module_doc
```
crusso, line 15: You could consider adding a `sort : Type.obj_sort` (object|module|actor) and then using the sort field of `ObjBlockE` below to populate it. Then you could emit doc for objects and actors too.

src/docs/extract.ml
```diff
@@ -133,8 +136,19 @@ struct
         {
           it = Syntax.LetD ({ it = Syntax.VarP { it = name; _ }; _ }, rhs, _);
           _;
-        } ->
-        Some (mk_xref (Xref.XValue name), extract_let_doc rhs name)
+        } -> (
+        match rhs with
+        | Source.{ it = Syntax.ObjBlockE (sort, fields); _ } ->
```
crusso, line 141: This is the sort field you want to use. You can assert on sort Type.Memory, which is a hacky label only used to describe stable variable records internally.

crusso (approved):
Nice.

I think this can go in as is, but you could also easily generalize it to handle nested object declarations and actors (if we ever support nested actor on the IC rather than just in the interpreter). 

If it goes in as is, maybe just open an issue to extend this PR in future for whoever wants to tackle it.

src/docs/extract.ml
rvanasa: Gotcha; thanks. I was 50/50 on generalizing this for the initial PR, but I went ahead and set everything up (along with a few improvements for module and actor classes).

src/docs/plain.ml
```diff
@@ -77,6 +77,16 @@ let plain_of_func_sort : Buffer.t -> Syntax.func_sort -> unit =
...
 
+let plain_of_obj_sort_title : Buffer.t -> Syntax.obj_sort -> unit =
+ fun buf sort ->
+  Buffer.add_string buf
+    Mo_types.Type.(
+      match sort.it with
+      | Object -> "Object "
+      | Actor -> "Actor "
+      | Module -> "Module "
+      | Memory -> "Memory ")
```
crusso, line 88:
```suggestion
      | Memory -> "Memory ")
```
This case shouldn't really arise but fine to support it.

crusso (approved):
Still LGTM. Thanks for  generalizing!

src/docs/plain.ml
rvanasa: Makes sense. For some reason this was already included, so I just kept the original logic.

PR #3922, opened 2023-04-06
chore: document `base` doc update step

ggreif:


crusso (commented):
That's not what I get on master. Did you do make base in doc, after make src?

ggreif:
> That's not what I get on master. Did you do make base in doc, after make src?

Nope, forgot that. But now this is documented and should be future-proof!

crusso (approved):

PR #3925, opened 2023-04-06
Apply `*.drun` tests to all GCs

luc-blaeser:
Currently, the `*.drun` test cases (upgrades etc.) are only applied to the copying GC.
I would be good to also test the other GCs for these test cases.


crusso:
Well spotted!

crusso (approved):

PR #3927, opened 2023-04-10
review: incremental gc suggestions

crusso:
Some suggestions to #3837, reviewing potential missing cases of TAG_ARRAY_SLICE_MINIMUM and formatting in compile.ml

@Luc, if you want them, you can just merge directly into your branch.

Note: in `compile.ml` I wasn't sure why you were using signed comparison against ArraySliceMinimum, so added a comment.

### DIFF
--- a/rts/motoko-rts/src/debug.rs
+++ b/rts/motoko-rts/src/debug.rs
@@ -165,7 +165,7 @@ pub(crate) unsafe fn print_boxed_object(buf: &mut WriteBuf, p: usize) {
             let obj_ind = obj as *const ObjInd;
             let _ = write!(buf, "<ObjInd field={:#x}>", (*obj_ind).field.get_raw());
         }
-        TAG_ARRAY => {
+        TAG_ARRAY | TAG_ARRAY_SLICE_MIN.. => {
             let array = obj as *mut Array;
             let _ = write!(buf, "<Array len={:#x}", (*array).len);
 
--- a/rts/motoko-rts/src/types.rs
+++ b/rts/motoko-rts/src/types.rs
@@ -802,7 +802,9 @@ pub(crate) unsafe fn block_size(address: usize) -> Words<u32> {
 
         TAG_OBJ_IND => size_of::<ObjInd>(),
 
-        TAG_ARRAY | TAG_ARRAY_SLICE_MIN.. => {
+        // `block_size` is not used during the incremental mark phase and
+        // therefore, does not support array slicing.
+        TAG_ARRAY => {
             let array = address as *mut Array;
             let size = array.len();
             size_of::<Array>() + Words(size)
--- a/rts/motoko-rts/src/types.rs
+++ b/rts/motoko-rts/src/types.rs
@@ -802,7 +802,7 @@ pub(crate) unsafe fn block_size(address: usize) -> Words<u32> {
 
         TAG_OBJ_IND => size_of::<ObjInd>(),
 
-        TAG_ARRAY => {
+        TAG_ARRAY | TAG_ARRAY_SLICE_MIN.. => {
             let array = address as *mut Array;
             let size = array.len();
             size_of::<Array>() + Words(size)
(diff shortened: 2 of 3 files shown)


src/codegen/compile.ml
```diff
@@ -5644,20 +5644,20 @@ module MakeSerialization (Strm : Stream) = struct
...
-        set_temp ^^ 
+        set_temp ^^
         get_temp ^^ compile_unboxed_const Tagged.(int_of_tag StableSeen) ^^
         G.i (Compare (Wasm.Values.I32 I32Op.Ne)) ^^
         get_temp ^^ compile_unboxed_const Tagged.(int_of_tag CoercionFailure) ^^
         G.i (Compare (Wasm.Values.I32 I32Op.Ne)) ^^
         G.i (Binary (Wasm.Values.I32 I32Op.And)) ^^
         get_temp ^^ compile_unboxed_const Tagged.(int_of_tag ArraySliceMinimum) ^^
-        G.i (Compare (Wasm.Values.I32 I32Op.GeS)) ^^
+        G.i (Compare (Wasm.Values.I32 I32Op.GeS)) ^^ (*crusso: why signed? *)
```
crusso, line 5660:
```suggestion
        G.i (Compare (Wasm.Values.I32 I32Op.GeS)) ^^ (* crusso: why signed? *)
```

rts/motoko-rts/src/debug.rs
```diff
@@ -165,7 +165,7 @@ pub(crate) unsafe fn print_boxed_object(buf: &mut WriteBuf, p: usize) {
             let obj_ind = obj as *const ObjInd;
             let _ = write!(buf, "<ObjInd field={:#x}>", (*obj_ind).field.get_raw());
         }
-        TAG_ARRAY => {
+        TAG_ARRAY | TAG_ARRAY_SLICE_MIN.. => {
```
luc-blaeser, line 168: Thank you very much. This is indeed a missing case of `TAG_ARRAY_SLICE_MIN` during the incremental mark phase.

rts/motoko-rts/src/types.rs
```diff
@@ -802,7 +802,7 @@ pub(crate) unsafe fn block_size(address: usize) -> Words<u32> {
 
         TAG_OBJ_IND => size_of::<ObjInd>(),
 
-        TAG_ARRAY => {
+        TAG_ARRAY | TAG_ARRAY_SLICE_MIN.. => {
```
luc-blaeser, line 805: I was also thinking about this change. However, the `block_size` should not be called during incremental mark phase. It will be used during incremental evacuation and updating phase, where the array slice information should already be restored. I will add a comment, that this is on purpose.

src/codegen/compile.ml
luc-blaeser: Thanks, good catch: It should be unsigned comparison.

src/codegen/compile.ml
```diff
@@ -5824,7 +5824,12 @@ module MakeSerialization (Strm : Stream) = struct
           get_tag ^^ compile_eq_const Tagged.(int_of_tag ObjInd) ^^
           E.then_trap_with env "unvisited mutable data in serialize_go (ObjInd)" ^^
           get_tag ^^ compile_eq_const Tagged.(int_of_tag Array) ^^
-          E.then_trap_with env "unvisited mutable data in serialize_go (Array)" ^^
+            E.then_trap_with env "unvisited mutable data in serialize_go (Array)" ^^
```
luc-blaeser, line 5827: The difficulty here is that the tags have already been modified to encode the stream offset, conflicting with the range `>= ArraySliceMinimum`. Therefore, unfortunately, we cannot check whether the value is an original array slice tag or the stream offset but need to rely on `clear_array_slicing` been called at the beginning of the serialization.

src/codegen/compile.ml
```diff
@@ -5824,7 +5824,12 @@ module MakeSerialization (Strm : Stream) = struct
...
+          (if !Flags.gc_strategy = Flags.Incremental then
+            get_tag ^^ compile_unboxed_const Tagged.(int_of_tag ArraySliceMinimum) ^^
+            G.i (Compare (Wasm.Values.I32 I32Op.GeS)) ^^ (* crusso: why signed? *)
+            E.then_trap_with env "unvisited mutable data in serialize_go (ArraySliceMinimum)"
+           else G.nop) ^^
```
luc-blaeser, line 5832:
```suggestion
          E.then_trap_with env "unvisited mutable data in serialize_go (Array)" ^^
```

src/codegen/compile.ml
crusso: Oh wow, that's subtle.

PR #3932, opened 2023-04-13
Test CI Build (Previous Nix Version)

luc-blaeser:
Explicitly set the nix version to `2.13.3` to run the CI builds and tests again.

PR #3933, opened 2023-04-13
chore: specify `nix` version on CI by pinning v2.13.3

ggreif:
solution by @crusso

Recent bump of `nix` to 2.15 broke our CI. Pinning 2.13.3, so that we can restore our manoeuvrability. It remains to be evaluated if it is only us or the general public that is affected.

See also #3932.


luc-blaeser (approved):
Thank you very much, Gabor. 

This solves our CI builds for the moment - and we can analyze independently what the recent nix changes are and how we could adjust the CI build scripts.

PR #3935, opened 2023-04-13
feat: implement `ic0.is_controller` as a primitive

ggreif:


chenyan-dfinity (approved):
LGTM. We can resume the ic-ref tests once they support is_controoler.

Issue #3892, opened 2023-03-21
Why the shared requirement on async*


nomeata:
The `async*` blocks are merely about abstracting over stuff that calls, so I wonder why we still have the `shared` requirement for their types. There is never going to be serialization happening at an `await*`, right?

crusso:
Yeah, I'd like to get rid of that requirement for both async and async*, actually. And do them together to avoid confusion.

crusso:
I was assuming just removing the requirement on `async*` wouldn't be useful enough, but perhaps I'm wrong. If you can already write useful abstractions (and use them!) with just `async*` maybe we should just relax type `async* <typ>`.

crusso:
@nomeata So I've been thinking about how to relax this for `async` (since it works, modulo an arity bug, for `async*`).

Unfortunately, it's a little harder because of the way we currently compile shared functions. Instead of representing them as lambdas that send data and return an async, we actually represent shared functions as simple pairs of (principal x methodname) and, for applications, generate the marshalling code at the call site, based on the type of the function *at the call-site*. In a polymorphic context, this may not be not be a fully known type (i.e. involve type parameters).

My gut feeling is that, in order to relax the shared type constraint on `async <typ>`, we actually need to compile shared functions as proper or special lambdas, that encapsulate their argument serialization and result deserialization (where the types are known) at the definition site, so the call-sites can just be compiled to applications with no extra type info required. 
Unfortunately, this would also mean representing actors as records of functions, not just principals, and adjusting deserialization of actor and function references to introduce the required actor/function proxies, driven by the expected Motoko type during deserialization (I guess). 

For reference equality, we probably want shared functions to be slightly special lambdas, with fixed principal and method name fields, so we can test their equality by  field comparison (and perhaps even with a special GC tag).

Another variation would be to represent shared function not by full lambdas, but by 4-tuples, with the new fields carrying argument/result serialization/deserialization code, but I'm not sure that buys us much.

nomeata:
> For reference equality, we probably want shared functions to be slightly special lambdas, with fixed principal and method name fields

Hmm, good point. And not just for equality, but also for passing them arguments to a shared function, i.e. for serialization?

PR #3936, opened 2023-04-13
experiment: allow non-shared `async*` types (only)

crusso:
So that we can write parametric abstractions, in the absence of a shared type constraint.

By popular demand. Fixes (part of) #3892.

Example:
```
import Debug "mo:base/Debug";
import Array "mo:base/Array";

actor a {

  func expect<T>(a : () -> async* T) : async* T {
    await* a();
  };


  func mapPar<A,B>(args : [A], f : A -> async* B) : async* [B] {
     let vs = Array.init<?B>(args.size(), null);
     for (i in args.keys()) {
        vs[i] := ?(await* f(args[i]));
     };
     Array.tabulate<B>(args.size(), func i { let ?v = vs[i]; v });
  };

  public func f(a : Nat) : async Text { debug_show(a) };

  public func go() : async () {
     let x = await* expect<Nat>(func () : async* Nat {await async 1});
     assert x == 1;

     let vs = await* mapPar([0,1,2], func (a : Nat) : async* Text { await f(a) });

     Debug.print(debug_show vs);
  }
};
```

Pros: just a minor tweak

Cons:
* still requires ugly eta-expansion of `shared` functions calls.
* can't just abstract over `async* T` (must be a function ` U -> async* T`) to avoid scope violations.
  (Perhaps `async* T` should be implicitly scope-polymorphic (since the computation is delayed anyway)).
* can't use implicit lambda arguments because we don't insert scope parameters for those (a separate, but annoying, issue: our sugar to do so is (syntactically) type driven, which obviously doesn't work when types are omitted.
 
TODO
- [ ] test run-drun/do-async-poly.mo needs revisiting
- [ ] actually implement `mapPar` with parallel waiting. This version ain't that.

### DIFF
--- a/src/ir_def/check_ir.ml
+++ b/src/ir_def/check_ir.ml
@@ -228,7 +228,7 @@ let rec check_typ env typ : unit =
     check_typ env typ2;
     check env no_region env.flavor.Ir.has_async_typ "async in non-async flavor";
     let t' = T.promote typ2 in
-    check_shared env no_region t'
+    if s == T.Fut then check_shared env no_region t'
   | T.Obj (sort, fields) ->
     List.iter (check_typ_field env (Some sort)) fields;
     check_field_hashes env "object" no_region fields;
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -515,7 +515,7 @@ and check_typ' env typ : T.typ =
   | AsyncT (s, typ0, typ) ->
     let t0 = check_typ env typ0 in
     let t = check_typ env typ in
-    if not env.pre && not (T.shared t) then
+    if not env.pre && s = T.Fut && not (T.shared t) then
       error_shared env t typ.at
         "M0033" "async has non-shared content type%a"
         display_typ_expand t;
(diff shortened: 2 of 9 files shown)


crusso:
Oh crap, that's actually `mapSeq` (sequential map), not `mapPar` (parallel map), due to lazy semantics of `async*`. Maybe this doesn't give us what we want unless we also do non-shared async types after all. Will need to experiment some more first.

nomeata:
That's annoying.

Does it have to be lazy?

crusso:
> Does it have to be lazy?

I think it does, if you want await* to demarcate the effects. But I'm open to better suggestions. I think the way to solve this is to lift the restriction on vanilla `async t`. Still scratching my head here.

nomeata:
I’m not thinking this through, but could `async* { e }` eagerly start evaluating `e` until the first actual `await`? A bit like `async { e }` “eagerly” (but in a separate threat) starts evaluating `e`?

crusso:
> I’m not thinking this through, but could `async* { e }` eagerly start evaluating `e` until the first actual `await`? A bit like `async { e }` “eagerly” (but in a separate threat) starts evaluating `e`?

That would be the semantics of the 'do async e' I originally proposed, except with a separate type. I guess the problem then is that the effects happen even without 'await*'. With 'async e' nothing happens unless you 'await',

nomeata:
Ah, right

luc-blaeser:

PR #3948, opened 2023-04-26
chore: simplify `base-src` derivation

ggreif:


nomeata (approved):
This has the effect that one will download all of base, instead of just src, when building this locally. Minor point of course.

Issue #3951, opened 2023-04-26
Optimise variant rebuilds to casts


ggreif:
A `Result<A, B>` pattern match of this form
``` Motoko
switch res {
    case ...;
    case (#err e) { ... (#err e) ... }
}
```
where the RHS `#err e` has type `Result<C, B>` can always be optimised to
``` Motoko
switch res {
    case ...;
    case (#err _) { ... (prim "cast" : Result<A, B> -> Result<C, B>) res ... }
}
```
Generally this transformation is valid for variants as long as the alternative is not affected by the type change (i.e. has neither sub-values of type `A` or `C`).

This can probably be implemented as an IR pass, and can potentially exhibit nice gains w.r.t. cycles and size in code that uses monadic error propagation a lot.

ggreif:
These are quite ubiquitous: `mapOk`, `chain` and `flatten` in `mo:base/Result` show this pattern.

nomeata:
It’d be a bit sad to have a IR-to-IR pass that has to resert to `prim cast` because the resulting program can no longer be typechecked. That’s incidentially why GHC doesn't do the optimization you mention at the Core level (but at the STG level it does, I think).

We could improve the type checker so that in the body of a `switch (v) {case (#tag e) …}`, the type of `v` is refined to `{ #tag : … }`, or something.

Of course, if the experiment in #3955 shows that that the perf impact is too large to ignore, that can change the story a bit.

ggreif:
Refining the scrutinee to a subtype sounds like a solution, but we cannot do it in the way GHC does it for GADT matches (type-directed). We'd have to resort to operational evidence (pointing to the case pattern that performed the match) and verifying (type-checking) such a thing sounds hairy to me.

PR #3955, opened 2023-04-28
perf: avoid rebuilding already scrutinised values

ggreif:
Contributes to #3951, covering simple special cases.

To keep this limited in scope, I have refrained from re-using scrutinees in sub-contexts. To make that happen one would need the environment to carry the substitution actions and correctly handle shadowing of variables (e.g. by killing the affected substitutions from the environment before descending into the tainted scopes).

----------
TODO:
- [x] can we get rid of the cast under certain conditions? We can, sometimes, see f09bac6766d0dbb0538a92928e18d04d79d44026.
- [x] `(SwitchE (VarE foo) ... case (TagP t (VarP o)) (TagE t (VarE o)))` --> `(SwitchE (VarE foo) ... case (TagP t WildP) (VarE foo))`
- [x] `(SwitchE (VarE foo) ... case (VarP o) (VarE o))` --> `(SwitchE (VarE foo) ... case WildP (VarE foo))`
- [x] `FileCheck`
- [ ] `case (#none) #none` and `case () ()`


ggreif:
So far the results are promising!
Compared as
```
$ ./compare-wat.sh -f origin/master -t HEAD run/recombine.mo
Building old moc (rev ae7dd1da1af2999ed4fbf8cbda8c4c7ca53a7991)..
Building new moc (rev 2c010175b69fb6efcf8fa0bc7394fb690b5233d8)..
Binary files compare-out/recombine.old/recombine.wasm and compare-out/recombine.new/recombine.wasm differ
diff -r -N -u10 compare-out/recombine.old/recombine.wat compare-out/recombine.new/recombine.wat
--- compare-out/recombine.old/recombine.wat	2023-04-28 22:05:27.097985602 +0200
+++ compare-out/recombine.new/recombine.wat	2023-04-28 22:05:27.248967448 +0200
```
we have eliminated locals and a bunch of instructions, as well as removed all consing:
``` wasm diff
   (func $matchX (type 14) (param $clos i32) (param $foo i32) (result i32)
-    (local $switch_in i32) (local $n i32) (local $tag_scrut i32) (local $heap_object i32) (local $o i32)
+    (local $switch_in i32) (local $tag_scrut i32)
     local.get $foo
     local.set $switch_in
     block (result i32)  ;; label = @1
         local.get $switch_in
         local.tee $tag_scrut
         i32.load offset=5
         i32.const 5097222
         i32.eq
         if  ;; label = @3
-          local.get $tag_scrut
-          i32.load offset=9
-          local.set $n
         else
           i32.const 0
           br 1 (;@2;)
         end
-        i32.const 3
-        call $alloc_words
-        local.tee $heap_object
-        i32.const 15
-        i32.store offset=1
-        local.get $heap_object
-        i32.const 5097222
-        i32.store offset=5
-        local.get $heap_object
-        local.get $n
-        i32.store offset=9
-        local.get $heap_object
+        local.get $foo
         br 1 (;@1;)
       end
       if  ;; label = @2
       else
-        local.get $switch_in
-        local.tee $o
+        local.get $foo
         br 1 (;@1;)
       end
       unreachable
     end)
```

ggreif:
@nomeata `local.tee $tag_scrut` appears totally redundant as it never gets read again :-)

Hopefully `wasm-opt` removes it just like `$switch_in`, which is basically just `$foo`. @kentosugama can you confirm? (I am happy to send you the wasm.)

src/ir_passes/recomb.ml
```diff
@@ -0,0 +1,135 @@
+(* Turns rebuilt variants into casts of the same thing.
+   Also turns catchall cases into wildcard ones when the scrutinee is already a variable. *)
+
+open Ir_def
```
ggreif, line 4: This whole file is mostly copied from `src/ir_passes/eq.ml`

src/ir_passes/recomb.ml
```diff
@@ -0,0 +1,139 @@
...
+       heartbeat = heartbeat';
+       timer = timer';
+       inspect = inspect'
+      }, typ)
+
+(* Entry point for the program transformation *)
+
+let transform (cu, flavor) =
+  assert (not flavor.has_typ_field); (* required for hash_typ *)
+  (t_comp_unit cu, {flavor with has_poly_eq = false})
```
crusso, line 139:
```suggestion
  (t_comp_unit cu, {flavor with has_poly_eq = false})
```
Copy paster remnant? (has_poly_eq)

src/ir_passes/recomb.ml
```diff
@@ -0,0 +1,139 @@
...
+
+and case env scrutinee case = match scrutinee, case with
+  (* switch v { ...; case (#err e) { #err e } } --> switch v { ...; case (#err _) { (prim cast) v } } *)
+  | { it = VarE _; _ }, { it = { pat = { it = TagP (ptag, ({it = VarP pid; _} as pv)); _ } as pat
+                               ; exp = { it = PrimE (TagPrim etag, [{ it = VarE eid; _}]); _ } as exp }; _ }
+    when ptag = etag && pid = eid ->
+    { case with it = { pat = { pat with it = TagP (ptag, {pv with it = WildP})}
+                     ; exp = { exp with it = cast scrutinee.note.Note.typ exp.note.Note.typ
+                                             {(t_exp env scrutinee) with at = exp.at} } } }
+  (* switch v { ...; case p p } --> switch v { ...; case _ v } *)
```
crusso, line 30: This isn't safe unless it's the last case, but I don't see where you require that.

crusso, line 30:
```suggestion
  (* switch x { ...; case y y } --> switch x { ...; case _ x } *)
```

src/ir_passes/recomb.ml
```diff
@@ -0,0 +1,139 @@
...
+ - spot `switch var {`
+ - where a variant match happens, that immediately recombines:
+    `case (#err err) { #err err }`
+ - replace with `case (#err _) { (prim cast) var }`
+ *)
+
+let rec t_exps env = List.map (t_exp env)
+
+and case env scrutinee case = match scrutinee, case with
+  (* switch v { ...; case (#err e) { #err e } } --> switch v { ...; case (#err _) { (prim cast) v } } *)
```
crusso, line 23:
```suggestion
  (* switch x { ...; case (#err y) { #err y } } --> switch x { ...; case (#err _) { (prim cast) x } } *)
```

crusso: I'm not super thrilled about inserting unsafe casts for optimization, but maybe we can this optimization last in the pipeline.

src/ir_passes/recomb.ml
ggreif: not sure, `Ir_def.prog` is `comp_unit * flavor`, I guess we don't chang the flavour, so
``` OCaml
let transform (cu, flavor) =
  (t_comp_unit cu, flavor)
```
?

ggreif: Fixed: 863946604

nomeata (commented):
I'm a bit skeptical if this is worth the complexity, compile time cost, maintenance cost and losing some of the value of having a typed it.

ggreif:
In `Trie.mo` there is code like this:
``` Motoko
    switch (left, right) {
      case (#empty, #empty) {
        #empty
      };
      case (#leaf(leftLeaf), #empty) {
        #leaf(leftLeaf)
      };
      case (#empty, #leaf(rightLeaf)) {
        #leaf(rightLeaf)
      };
```
I have manually rewritten it in https://github.com/dfinity/motoko-base/pull/563/commits/c65868b5ecdfd7ec534de6235bd9f47534a220b7, but obviously the compiler could do the same thing. The complication is that the variant match is not at toplevel.

ggreif:
Random thought... The Wasm backend (`compile.ml`) introduces a lot of local variables while doing the pattern match. These could be reused, but it is probably an order of magnitude more complicated, so probably not worth it.

nomeata:
The backend is written under the assumption that locals are very cheap. If that doesn't hold anymore the pattern match code wouldn't be my first worry… :-)

ggreif:
Oh, by "reused" I meant returning (instead of reconstructing values) one of those Wasm locals that arise while pattern matching (as opposed to user-written Motoko locals).

crusso:
> The backend is written under the assumption that locals are very cheap. If that doesn't hold anymore the pattern match code wouldn't be my first worry… :-)

I fear that's actually not true, because the replica disables all optimizations. ic-wasm's wasm-opt step mitigates this, at the cost of a large untrusted optimizer....

nomeata:
Too bad. We need to support more targets then to be able to justify out design post-hoc.

The other reasoning is that we optimize for cycle consumption (and size), not speed, and at least so far locals are cheap in that way.

ggreif:
Now that we have a `dao` performance check, it turns out that this transform has teeth!

luc-blaeser:

PR #3956, opened 2023-04-28
experiment: syntactic sugar for pipes

crusso:
Defines pipes as syntactic sugar for supplying the receiver as the first argument.

```
e |> f <inst> ()  --> f <inst> e

e |> f <inst> (e1, ...,en) --> f<inst>(e, e1, ..., en), n>=1
```

TODO:
- [ ] preserve left-to-right evaluation order by let-binding e on rhs
- [ ] actually lower precedence of `|>` (TIL lowest comes first in Menhir) to just above ':'

Caveat: argument must be a manifest tuple for this work. A type directed solution using a dedicated AST node could probably do better and allow first-class tuple arguments.

Example:
```
[nix-shell:~/motoko/src]$ rlwrap moc --package base $MOTOKO_BASE
Motoko compiler (source 0.8.7-16-gae7dd1da1-dirty)
> import C = "mo:base/Char";
let C : module {...}
> import I = "mo:base/Iter";
let I : module {...}
> import L = "mo:base/List";
let L : module {...}
> "hello".chars() |> I.map(C.toNat32) |> I.toList() |> L.reverse();
?(111, ?(108, ?(108, ?(101, ?(104, null))))) : List<Nat32>
```

The last line is equivalent to:
```
> L.reverse(I.toList(I.map("hello".chars(), C.toNat32)));
?(111, ?(108, ?(108, ?(101, ?(104, null))))) : List<Nat32>
```

### DIFF
--- a/doc/md/examples/grammar.txt
+++ b/doc/md/examples/grammar.txt
@@ -184,7 +184,7 @@
     <exp_bin> 'and' <exp_bin>
     <exp_bin> 'or' <exp_bin>
     <exp_bin> ':' <typ_nobin>
-    <exp_bin> PIPE <exp_post> ('<' <list(<typ>, ',')> '>')? '(' <list(<exp>, ',')> ')'
+    <exp_bin> '|>' <exp_post> ('<' <list(<typ>, ',')> '>')? '(' <list(<exp>, ',')> ')'
 
 <exp_nondec> ::= 
     <exp_bin>
--- a/doc/md/examples/grammar.txt
+++ b/doc/md/examples/grammar.txt
@@ -184,6 +184,7 @@
     <exp_bin> 'and' <exp_bin>
     <exp_bin> 'or' <exp_bin>
     <exp_bin> ':' <typ_nobin>
+    <exp_bin> PIPE <exp_post> ('<' <list(<typ>, ',')> '>')? '(' <list(<exp>, ',')> ')'
 
 <exp_nondec> ::= 
     <exp_bin>
--- a/src/gen-grammar/grammar.sed
+++ b/src/gen-grammar/grammar.sed
@@ -72,6 +72,7 @@ s/RBRACKET/\']\'/g
 s/QUEST/\'?\'/g
 s/BANG/\'!\'/g
 s/QUERY/\'query\'/g
+s/PIPE/\'|>\'/g
 s/PUBLIC/\'public\'/g
 s/PRIVATE/\'private\'/g
 s/POWOP/\'**\'/g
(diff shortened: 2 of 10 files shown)


crusso:
Note to self:
Variants - use `_` to indicate hole on rhs (not necessarily the first argument)

```
"hello".chars() |> I.map(_, C.toNat32) |> I.toList _ :> L.reverse _;
```

Variant: use a  full pattern (more flexible, but also more verbose).
```
"hello".chars() as cs in I.map(cs, C.toNat32) as ns in I.toList ns as l in L.reverse l;
```

roman-kashitsyn:
> "hello".chars() |> I.map(_, C.toNat32) |> I.toList _ |> L.reverse _;

That's my favorite option! It's really easy to understand what's going on. I believe it's essential to make `_` linear and forbid expressions like

```
"hello".chars() |> Iter.zip(_, _)
```

ggreif:
> Variants - use `_` to indicate hole on rhs (not necessarily the first argument)

Yes, this is a very compact notation for lambdas, got for it!

rossberg:
May I kindly suggest not going there? Motoko's syntax design is intentionally conservative curly-braces-and-semicolons, in order to make "normal" programmers feel at home. There is little that is more off-putting than encountering obscure symbolic syntax (shoehorned from a completely different language culture with curried functions). If this feature actually became widely used, it would hence undermine a central design premise of Motoko. And if not, it's obviously not worth it.

Not a big fan of underscores for implicit lambdas either. Without an additional delimiter that marks their binding point they are completely ambiguous syntactically: for example, `f(_, y)` could legitimately mean any of `func x ⇒ f(x, y)`, `f(func x  ⇒ (x, y))` or `f(func x ⇒ x, y)`.

crusso:
> May I kindly suggest not going there? Motoko's syntax design is intentionally conservative curly-braces-and-semicolons, in order to make "normal" programmers feel at home. There is little that is more off-putting than encountering obscure symbolic syntax (shoehorned from a completely different language culture with curried functions). If this feature actually became widely used, it would hence undermine a central design premise of Motoko. And if not, it's obviously not worth it.

We've actually had a few requests for something like this to make code more legible.

> Not a big fan of underscores for implicit lambdas either. Without an additional delimiter that marks their binding point they are completely ambiguous syntactically: for example, `f(_, y)` could legitimately mean any of `func x ⇒ f(x, y)`, `f(func x ⇒ (x, y))` or `f(func x ⇒ x, y)`.

Nor am I, for the same reason, but this isn't an implicit lambda, but a `let`, and I don't see a similar ambiguity arising (but do point it out if there is one). I actually prefer the variant in #3968 which is more explicit and selective about the argument being filled.

crusso:
https://github.com/tc39/proposal-pipeline-operator has some decent motivation for pipes.

Hack has something similar  to the underscore syntax, but without the linearity restriction. On the other hand, the placeholder can be used anywhere, not just in immediate argument position, which means the argument can be anything, not just an immediate tuple with a place holder.

Roman seemed keen on the linearity restriction to prevent duplicating piped iterators.

crusso:
OMG https://github.com/tc39/proposal-pipeline-operator/blob/main/HISTORY.md

PR #3960, opened 2023-05-02
ci: add an artifact job

chenyan-dfinity:
When "build_artifacts" label is applied, CI builds the `moc` binary and uploads to artifact.

- [x] generate a nightly.link to download the artifact.


ggreif (approved):
This looks useful to me for benchmarking purposes, thanks! We can refine this (refactoring to common rules, to combat duplication) as we go.

Issue #966, opened 2019-12-02
Quo vadis `blob`


nomeata:
In #963 I am introducing `Blob` type. This is initially just scaffolding to not have #789 blocked on the question of whether we want a (more) abstract `Entity`/`Principal`/`Caller`/`Canister`/`User` type, and be able to write some code.

Presumably, the caller will turn into some kind of abstract type, possibly with conversions to and from `Text` and/or `Blob`.

But what about `Blob` itself? I see a few directions this can take:
1. We keep `Blob` as a primitive type, just like `Text`, and add operations (various conversion to and from text, maybe even indexing) as needed.
2. We define `type Blob = [Word8]`, and live with the size penalty due to the uniform representation.
3. We define `type Blob = [Word8]`, but change the representation to pack arrays of 8, 16, 32 and 64bit sized types upon construction, with separate heap object types for these. All array access would dynamically dispatch as needed. We’d be trading execution speed for space usage (but may be a good deal on our platform).

With 2 and 3 there is the problem that we would lose the equality and comparison operators, which are really kinda useful.

Somewhat independent there is the question if we want blob literal syntax.

skilesare:
I'd still love concatenation.  Current use case from the IC Spec:

Representation-independent hashing of structured data

Structured data, such as (recursive) maps, are authenticated by signing a representation-independent hash of the data. This hash is computed as follows (using SHA256 in the steps below):

For each field that is present in the map (i.e. omitted optional fields are indeed omitted):

concatenate the hash of the field's name (in ascii-encoding, without terminal \x00) and the hash of the value (with the encoding specified below).
Sort these concatenations from low to high

Concatenate the sorted elements, and hash the result.

The resulting hash of 256 bits (32 bytes) is the representation-independent hash.

The following encodings of field values as blobs are used

Binary blobs (canister_id, arg, nonce, module) are used as-is.

Strings (request_type, method_name) are encoded in UTF-8, without a terminal \x00.

Natural numbers (compute_allocation, memory_allocation, ingress_expiry) are encoded using the shortest form [Unsigned LEB128](https://en.wikipedia.org/wiki/LEB128#Unsigned_LEB128) encoding. For example, 0 should be encoded as a single zero byte [0x00] and 624485 should be encoded as byte sequence [0xE5, 0x8E, 0x26].

Arrays (paths) are encoded as the concatenation of the hashes of the encodings of the array elements.

Maps (sender_delegation) are encoded by recursively computing the representation-independent hash.

[...]

PR #3965, opened 2023-05-05
feat: add blob get and slice prim functions

q-uint:
Hi 👋🏼

Here is my first attempt to start contributing to Motoko, and writing OCaml.

This PR adds two new primitives to the Prim library: `blobGet` and `blobSlice`. These primitives allow for (more) efficient byte manipulation with a blob.

This is my first time writing OCaml, so there may be room for improvement in the implementation. However, I have tested these functions and they work as intended.

---

The `blobGet` function takes a blob `b` and an offset `offset` as arguments and returns the byte at that offset in the blob.

The `blobSlice` function creates a new blob from a portion of an existing blob. It takes a blob `b`, a starting offset `from`, and an ending offset `to` as arguments. The new blob will contain the bytes of the original blob from the from offset (inclusive) to the to offset (exclusive).

---

I welcome any feedback on the implementation and look forward to improving my OCaml skills 🐫.

### DIFF
--- a/src/mo_values/prim.ml
+++ b/src/mo_values/prim.ml
@@ -285,6 +285,23 @@ let prim trap =
     k (Blob (String.of_seq (Seq.map (fun v ->
       Char.chr (Nat8.to_int (Value.as_nat8 !(Value.as_mut v)))
     ) (Array.to_seq (Value.as_array v)))))
+  (* returns a single nat8 at the given index *)
+  | "blobGet" -> fun _ v k ->
+    (match Value.as_tup v with
+    | [a; b] -> let (a, b) = (Value.as_blob a, Int.to_int (as_int b)) in
+      if b < 0 || b >= String.length a (* check bounds *)
+      then trap.trap "blob index out of bounds"
+      else k (Nat8 (Nat8.of_int (Char.code (String.get a b))))
+    | _ -> assert false)
+  (* returns a new blob that contains a subset [from,to) of the given blob *)
+  (* in the future negative values could be used to offset from the end, nat :< int *)
+  | "blobSlice" -> fun _ v k ->
+    (match Value.as_tup v with
+    | [a; b; c] -> let (a, b, c) = (Value.as_blob a, Int.to_int (as_int b), Int.to_int (as_int c)) in
+      if c < b || b < 0 || b > String.length a (* check bounds, b <= c *)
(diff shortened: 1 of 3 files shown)


src/mo_values/prim.ml
```diff
@@ -285,6 +285,23 @@ let prim trap =
     k (Blob (String.of_seq (Seq.map (fun v ->
       Char.chr (Nat8.to_int (Value.as_nat8 !(Value.as_mut v)))
     ) (Array.to_seq (Value.as_array v)))))
+  (* returns a single nat8 at the given index *)
+  | "blobGet" -> fun _ v k ->
```
ggreif, line 289: This would be better implemented using an overloaded array indexing construct:
``` Motoko
> ("\00":Blob)[0];
stdin:3.2-3.12: type error [M0075], expected array type, but expression produces type
  Blob
```

nomeata: We should make a decision over at https://github.com/dfinity/motoko/issues/966 first. If we merge Blob and Array types somehow, we might not need these prims?

Otoh, it's be easy to keep them around for backwards compat, so maybe ok?

ggreif: we can always remove prims if we want without notice

q-uint:
Seems like I am missing some pieces to let the CI pass.

Running the following in `nix-shell`: `./test/run.sh test/run/blobs.mo` I get the following error:

```
+TODO: compile_prim_invocation
+Error: failed to run main module `_out/blobs.wasm`
+
+Caused by:
+    1: wasm trap: unreachable
+       wasm backtrace:
+         0: init
+         1: _start
+       
```

Are there other files I need to edit/add?

---

@ggreif @crusso

ggreif:
> Are there other files I need to edit/add?

Yes. You have implemented the interpreter(s), but not the code generator. Search for that keyword in `src/codegen/compile.ml`. You can use #3935 as a blueprint as of what to touch there.

src/mo_values/prim.ml
crusso: Yes, I think it's fine to just a function for this at first. I'm not even sure I want to overload array subscripting, though its arguably already overloaded for [var T] and [T]. A good choice of library name can come later. `Blob.nth`, `Blob.sub`, `Blob.index`?

crusso:
Thanks @q-uint !

PR #3968, opened 2023-05-06
experiment: pipe syntax with explicit linear placeholder

crusso:
Alternative to #3956, using `_` to nominate the piped argument, as preferred by @roman-kashitsyn.

TODO:
- [x] let-bind argument when necessary to preserve eval order.
- [x] tests
- [x] don't optimize variable on the lhs, always introduce a let, since lhs might be an lvalue, changing the order of the read.
- [ ] actually lower precedence of '|>' to just above ':'

Examples:
```
[nix-shell:~/motoko/src]$ rlwrap moc --package base $MOTOKO_BASE
Motoko compiler (source 0.8.7-19-ga9dd5b32d-dirty)
> import L = "mo:base/List";
let L : module {...}
> import I = "mo:base/Iter";
let I : module {...}
> import L = "mo:base/List";
let L : module {...}
> import C = "mo:base/Char";
let C : module {...}

> "hello".chars() |> I.map(_, C.toNat32) |> I.toList _ |> L.reverse _;
?(111, ?(108, ?(108, ?(101, ?(104, null))))) : List<Nat32>

> "hello".chars() |> I.map(_, C.toNat32) |> I.toList _ |> L.reverse (_); // handle (_) as _
?(111, ?(108, ?(108, ?(101, ?(104, null))))) : List<Nat32>

> C.toNat32 |> I.map("hello".chars(), _) |> I.toList _ |> L.reverse _; // pipe to second argument
?(111, ?(108, ?(108, ?(101, ?(104, null))))) : List<Nat32>
```


rvanasa:
I love this syntax in the sense of it being a clever and concise generalization. That being said, looking at this through the eyes of a newer Motoko developer, I think the original style #3956 would be more intuitive, and the added opinionation would help keep libraries more consistent with each other. 

Either option would be great for different use cases, so I'm just hoping to add an extra opinion to the mix.

crusso:
> I love this syntax in the sense of it being a clever and concise generalization. That being said, looking at this through the eyes of a newer Motoko developer, I think the original style #3956 would be more intuitive, and the added opinionation would help keep libraries more consistent with each other.

Yeah, I'm a little torn. The thing I didn't like about that #3956 is that, though concise, it changes the number of arguments of the operation, which might be its own source of confusion.

```
"hello".chars() |> I.map(_, C.toNat32)
```
vs.
```
"hello".chars() |> I.map(C.toNat32)
```

Notice how `I.map` now takes one argument, even though its type `(Iter<T>,T->U) -> Iter<U>` requires two.

The underscore not only makes the argument clear but also lets you to select a different one than just the first.

> Either option would be great for different use cases, so I'm just hoping to add an extra opinion to the mix.

No, that's cool, thanks for the feedback. Let's see if others voice a preference.

ByronBecker:
A few more examples would be nice to showcase what's supported and the edges of the _ intended usage.

ggreif:
Some other language is going along these lines too https://twitter.com/code_report/status/1657058727239794688

ggreif:
> is _not_ legal, because the `_` is nested.

Aaah, sorry, that misunderstanding is originating from me then :-O

But is there a hard reason why `_` cannot be in a nested position, or is this just a restriction for the PoC?

crusso:
> Some other language is going along these lines too https://twitter.com/code_report/status/1657058727239794688

That links to an interesting survey here:

https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2672r0.html#to-placehold-or-not-to-placehold

src/mo_frontend/parser.mly
```diff
@@ -641,6 +642,32 @@ exp_un(B) :
     { OrE(e1, e2) @? at $sloc }
   | e=exp_bin(B) COLON t=typ_nobin
     { AnnotE(e, t) @? at $sloc }
+  | e1=exp_bin(B) PIPE f = exp_post(B) inst=inst mk_arg=pipe_arg
+    { match e1.it with
+      | VarE x ->
+        CallE(f, inst, mk_arg e1) @? at $sloc
+      | _ ->
```
crusso, line 649: Don't optmize for variables - it could be a lvalue!

crusso:
> > is _not_ legal, because the `_` is nested.
> 
> Aaah, sorry, that misunderstanding is originating from me then :-O
> 
> But is there a hard reason why `_` cannot be in a nested position, or is this just a restriction for the PoC? I can easily envision a `let`-based desugaring that allows nesting _and_ non-linearity.

It was to ensure linearity (as requested by @roman-kashitsyn) without requiring a full linear type system. 

A non-linear version allowing nesting should indeed be easy using just 'let', but that wasn't the goal here.

ggreif:
@crusso can this be closed, now that an alternative design got merged?

PR #3973, opened 2023-05-12
updates to Motoko mentions of the SDK and DFX

dprats:
Updates to the Motoko docs so it shows the correct naming for IC SDK


ggreif (approved):
Thanks for the fixes!

PR #3947, opened 2023-04-26
chore: bump `rustc` nightly

ggreif:
and a few deps too:
- `wasi-libc`

needs patched toolset from mozilla (PRs submitted: https://github.com/mozilla/nixpkgs-mozilla/pull/309)

TODO:
- [x] `motoko-rts-test` fails in `stream flushing` with unaligned memory access — see #3981

PR #3979, opened 2023-05-15
Fix unaligned dereferencing issue in `rustc` upgrade

luc-blaeser:
Analysis:
* Unaligned pointer dereferencing has undefined behavior in Rust: https://doc.rust-lang.org/reference/behavior-considered-undefined.html
* Accessing the `64-bit` fields in `Stream` are potentially unaligned because allocation granularity is 4 byte. However, no error occurred until now.
* Accessing the function pointer `Stream.outputter` seems to assume 8-byte alignment since the last Rust compiler version (why?) and thus need to be safely accessed if unaligned.

May be a fix for `https://github.com/dfinity/motoko/pull/3947`.


luc-blaeser:
**Update**: I will open a second PR that is based on the latest `master` (including incremental GC changes).

PR #3988, opened 2023-05-18
doc: convert error code explanations from ADoc to Markdown

rvanasa:
This PR replaces the AsciiDoc error code explanations with equivalent Markdown files. 

Since the goal is to show these explanations in developer tools such as the VS Code extension, embed service, and Motoko Playground, switching the ground truth to Markdown will ensure that everything renders in a predictable way across all these different contexts. 

- This change would have a minimal impact on the `moc --explain` output.
- Converting from Markdown to AsciiDoc is essentially lossless in case anyone is relying on the original file format.
- Markdown also has a [more concise code block syntax](https://docs.asciidoctor.org/asciidoc/latest/asciidoc-vs-markdown/) when specifying a language, which will be convenient when writing more of these explanations.


crusso (commented):
LGTM. Thanks for doing this. I don't think anything current relies on the adoc format. It the past, we may have included the adoc files on the adoc generated website.

Indeed, maybe we should now host these files on the new portal, in the references section for Motoko

PR #3994, opened 2023-05-22
fix #3993

### LINKED ISSUE #3993: Three-way `or` patterns fail to IR type check
``` Motoko
func three(#a () or #b () or #c ()) {};
```
fails with
```
three.mo:1.12-1.26: IR type error [M0000], subtype violation:
  {#a; #b}
  {#a}
```

ggreif:
Update: The IR check was correct.

--------
The subtype check for `or`-patterns was backwards.

The full pattern can have more alternatives than the individual branches, thus the `or`-pattern should be a _supertype_ of the types of the branches.

Another test case would be
``` Motoko
func nums(1 or (_ : Int)) {};
```
Which used to fail with:
```
or-pattern-three.mo:2.10-2.26: IR type error [M0000], subtype violation:
  Int
  Nat
```

### DIFF
--- a/src/ir_def/check_ir.ml
+++ b/src/ir_def/check_ir.ml
@@ -1017,8 +1017,8 @@ and check_pat env pat : val_env =
   | AltP (pat1, pat2) ->
     let ve1 = check_pat env pat1 in
     let ve2 = check_pat env pat2 in
-    t <: pat1.note;
-    t <: pat2.note;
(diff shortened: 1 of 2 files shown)


ggreif:
Closing this, as this change can cause unsound things to happen. See #4000 which attempts to fix `or`-pattern inference instead.

PR #3995, opened 2023-05-22
feat: multiline `Text` literals

kentosugama:
Some discussion here: https://github.com/dfinity/motoko/discussions/3939

This PR extends the current `Text` literal to allow multiline. It does not enable interpolation as `Text` is implemented using ropes under the hood, which already has fast appends.

Example:
```
"here is a 

multiline 
text value"
```

Some other solution for nicely expressing `json`, `html`, `xml` and other similar structures are being considered (aside from using `Text` with interpolation), but that is outside the scope of this PR.


rvanasa:
Would there be any downsides to (also) allowing newlines in the existing `"..."` syntax?

rossberg:
Yeah, I'm also a bit confused what the added value of this syntax is over just allowing newlines in regular text literals.

An extra value of a secondary string literal syntax might be that it allows picking custom delimiters, which helps avoiding any escaping inside the string.

kentosugama:
Yes I think that's a reasonable approach. I will modify the syntax.

src/mo_frontend/source_lexer.mll
```diff
@@ -181,8 +183,6 @@ rule token mode = parse
   | float as s { FLOAT s }
   | char as s { CHAR (char lexbuf s) }
   | text as s { TEXT (text lexbuf s) }
-  | '"'character*('\n'|eof)
-    { error lexbuf "unclosed text literal" }
```
rossberg, line 185: I think you want to keep this for eof.

src/mo_frontend/source_lexer.mll
```diff
@@ -91,13 +91,15 @@ let utf8 = ascii | utf8enc
...
 let byte = '\\'hexdigit hexdigit
-let escape = ['n''r''t''\\''\'''\"']
+let escape = ['t''\\''\'''\"']
 let character =
     [^'"''\\''\x00'-'\x1f''\x7f'-'\xff']
   | utf8enc
   | byte
   | '\\'escape
   | "\\u{" hexnum '}'
+  | '\n'
```
rossberg, line 101: Simply throwing in newline into a token is broken. Every newline requires a call to Lexing.new_line, otherwise line number tracking for errors and warnings will be off.

src/mo_frontend/source_lexer.mll
```diff
@@ -91,13 +91,15 @@ let utf8 = ascii | utf8enc
...
+  | '\r'
```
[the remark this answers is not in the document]
kentosugama: Okay, that's good to know. Thank you!

src/mo_frontend/source_lexer.mll
```diff
@@ -91,15 +91,14 @@ let utf8 = ascii | utf8enc
 let utf8_no_nl = ascii_no_nl | utf8enc
 
 let byte = '\\'hexdigit hexdigit
-let escape = ['t''\\''\'''\"']
+let escape = ['\n''\r''t''\\''\'''\"']
```
rossberg, line 94:
```suggestion
let escape = ['n''r''t''\\''\'''\"']
```

kentosugama:
Apologies this took forever

doc/md/language-manual.md
```diff
@@ -170,6 +170,7 @@ character ::=
   | '\\'escape
   | '\\'hexdigit hexdigit
   | "\\u{" hexnum '}'
+  | "\n"
```
crusso, line 173:
```suggestion
  | `\n`        // literal newline
```

doc/md/language-manual.md
```diff
@@ -170,7 +170,7 @@ character ::=
   | '\\'escape
   | '\\'hexdigit hexdigit
   | "\\u{" hexnum '}'
-  | "\n"
+  | `\n`        // literal newline
```
ggreif, line 173: those are backquotes!

ggreif:
Quick question: I guess these are now also covered:
- `actor "multiline"`
- `"multiline" : Blob

crusso (approved):
I extended the language manual and added a changelog description.

One minor issue might want to address later is that we also use text literals for Blob literals, in which case we should maybe  warn about the multiline literal.

```
let blob : Blob = 
"\FF\FF
\FF\FF"
```
Might not produce what the user expects... I guess one could look at the source region of the literal and, for a blob, warn if it spans more than one line.

ggreif:
> warn if it spans more than one line

Yep, and probably `actor "multiline" ...` too

doc/md/language-manual.md
crusso: Nope.

src/mo_frontend/source_lexer.mll
```diff
@@ -98,6 +99,7 @@ let character =
   | byte
   | '\\'escape
   | "\\u{" hexnum '}'
+  | '\n'
```
crusso, line 102:
```suggestion
  | '\\'escape
  | "\\u{" hexnum '}'
  | '\n'
```

doc/md/language-manual.md
```diff
@@ -170,6 +170,7 @@ character ::=
   | '\\'escape
   | '\\'hexdigit hexdigit
   | "\\u{" hexnum '}'
+  | `\n`        // literal newline
```
crusso, line 173:
```suggestion
  | '\n'        // literal newline
```

doc/md/language-manual.md
```diff
@@ -182,6 +183,8 @@ A text literal is `"`-delimited sequence of characters:
 text ::= '"' character* '"'
 ```
 
+Note, a text literal may span multiple lines.
```
crusso, line 186:
```suggestion
Note that a text literal may span multiple lines.
```

Changelog.md
```diff
@@ -2,6 +2,19 @@
...
 
+  * Allow multiline text literals (#3995).
+    For example,
+
+    ```
+    "A horse walks into a bar.
+    The barman says: `Why the long face?`"
+    ```
+
+    parses as:
```
crusso, line 13:
```suggestion
    parses as:
    
```

PR #4001, opened 2023-05-24
bugfix: compiler crash compiling inferred `or`-patterns

crusso:
The compiler crashes compiling (inferred) or-patterns, because the IR checked requires that type of a pattern is always a supertype of the expected type, so the pattern can consume the expected value.
Unfortunately, pattern inference at the source level is producing badly typed patterns that don't preserve the supertyping invariant.

In particular, the problem is that type inference for or-patterns was taking the lub of the types of alternatives.  

For consistency with pattern type checking, both at the IR and source level, we should be inferring the glb of the pattern alternatives, not the lub.

@rossberg, @nomeata  @ggreif can you check my reasoning here?

This is particularly dangerous with full or patterns that bind identifiers:

Consider
```
func bad((((x,y) : (Nat,Int)) or ((y,x) : (Int,Nat)))): Nat { x }
```
If we take the lub, `bad` has inferred type `(Int,Int) -> Nat`, allowing you to coerce an `Int` to a `Nat` which is unsound.
If we take the glb, `bad` receives type `(Nat, Nat) -> Nat`, which is safe.

So much for Motoko being type safe....

Note before we had (removing the unsupported or pattern identifiers):
```
[nix-shell:~/clean/motoko/test/run/ok]$ rlwrap moc
Motoko compiler (source 0.9.1-7-gcf341ee2b)
> func((((_,_) : (Nat,Int)) or ((_,_) : (Int,Nat))) ): Nat { 0 }
  ;
stdin:1.30-1.49: warning [M0146], this pattern is never matched
func : ((Int, Int)) -> Nat
```
Now we obtain:
```
[nix-shell:~/motoko/test/fail]$ rlwrap moc
Motoko compiler (source 0.9.1-8-g3af059077-dirty)
> func((((_,_) : (Nat,Int)) or ((_,_) : (Int,Nat))) ): Nat { 0 }
  ;
stdin:1.30-1.49: warning [M0146], this pattern is never matched
func : ((Nat, Nat)) -> Nat
```
UPDATE:

I was hoping to use this counter-example:
```
func f({a=true} or {}) { };
f{}
```
to demonstrate that inferring the lub is unsound,  but using the glb is ok.
I was hoping Motoko would infer `f : {} -> ()` and that `f{}` would try to access an unavailable field.

However, Motoko rejects the record or-pattern because the types `{ a : Bool }` and `{}` are incompatible (`T.incompatible`).
Disabling the compatibility check to see what happens cause coverage checking to crash. Why do we have this predicate and why do we rely on it?
The relevant PR that introduced the compatibility check is https://github.com/dfinity/motoko/pull/538 (but not much explanation is given).
@rossberg can you recall what was going on here? Could the complexity be avoided by using the glb?

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -1866,7 +1866,7 @@ and infer_pat' env pat : T.typ * Scope.val_env =
   | AltP (pat1, pat2) ->
     let t1, ve1 = infer_pat env pat1 in
     let t2, ve2 = infer_pat env pat2 in
-    let t = T.lub t1 t2 in
+    let t = T.glb t1 t2 in
     if not (T.compatible t1 t2) then
       error env pat.at "M0104"
         "pattern branches have incompatible types,\nleft consumes%a\nright consumes%a"
--- a/test/fail/ok/pat-inconsistent.tc.ok
+++ b/test/fail/ok/pat-inconsistent.tc.ok
@@ -2,20 +2,24 @@ pat-inconsistent.mo:2.13-2.21: warning [M0145], this pattern of type
   Nat
 does not cover value
   0 or 3 or _
+pat-inconsistent.mo:3.19-3.21: warning [M0146], this pattern is never matched
 pat-inconsistent.mo:3.13-3.22: warning [M0145], this pattern of type
-  Int
+  Nat
 does not cover value
   0 or 2 or _
 pat-inconsistent.mo:5.13-5.29: warning [M0145], this pattern of type
-  Int
+  Nat
 does not cover value
-  0 or -1 or _
+  0 or 3 or _
 pat-inconsistent.mo:6.31-6.44: warning [M0146], this pattern is never matched
 pat-inconsistent.mo:7.34-7.50: warning [M0146], this pattern is never matched
+pat-inconsistent.mo:8.14-8.21: warning [M0146], this pattern is never matched
+pat-inconsistent.mo:8.15-8.17: warning [M0146], this pattern is never matched
+pat-inconsistent.mo:8.25-8.32: warning [M0146], this pattern is never matched
+pat-inconsistent.mo:8.29-8.31: warning [M0146], this pattern is never matched
 pat-inconsistent.mo:8.13-8.33: warning [M0145], this pattern of type
-  (Int, Int)
+  (Nat, Nat)
 does not cover value
-  (-1, 0 or 1 or _) or
   (2, 0 or 1 or _) or
   (0 or 1 or _, _)
 pat-inconsistent.mo:9.20-9.22: warning [M0146], this pattern is never matched
--- a/test/run/ok/or-pattern-bug.comp.ok
+++ b/test/run/ok/or-pattern-bug.comp.ok
@@ -1,5 +1,0 @@
-Ill-typed intermediate code after Desugaring (use -v to see dumped IR):
-or-pattern-bug.mo:6.9-6.19: IR type error [M0000], subtype violation:
-  {#a; #b}
-  {#a}
-
--- a/test/run/ok/or-pattern-bug.comp.ok
+++ b/test/run/ok/or-pattern-bug.comp.ok
@@ -1,0 +1,5 @@
+Ill-typed intermediate code after Desugaring (use -v to see dumped IR):
+or-pattern-bug.mo:6.9-6.19: IR type error [M0000], subtype violation:
+  {#a; #b}
+  {#a}
+
--- a/test/run/ok/or-pattern-bug.comp.ret.ok
+++ b/test/run/ok/or-pattern-bug.comp.ret.ok
@@ -1,0 +1,1 @@
+Return code 1
--- a/test/run/ok/or-pattern-bug.comp.ret.ok
+++ b/test/run/ok/or-pattern-bug.comp.ret.ok
@@ -1,1 +1,0 @@
-Return code 1
--- a/test/run/ok/or-pattern-bug.diff-ir.ok
+++ b/test/run/ok/or-pattern-bug.diff-ir.ok
@@ -1,0 +1,8 @@
+--- or-pattern-bug.run
++++ or-pattern-bug.run-ir
+@@ -0,0 +1,5 @@
++Ill-typed intermediate code after Desugaring (use -v to see dumped IR):
++or-pattern-bug.mo:6.9-6.19: IR type error [M0000], subtype violation:
++  {#a; #b}
++  {#a}
++
--- a/test/run/ok/or-pattern-bug.diff-ir.ok
+++ b/test/run/ok/or-pattern-bug.diff-ir.ok
@@ -1,8 +1,0 @@
---- or-pattern-bug.run
-+++ or-pattern-bug.run-ir
-@@ -0,0 +1,5 @@
-+Ill-typed intermediate code after Desugaring (use -v to see dumped IR):
-+or-pattern-bug.mo:6.9-6.19: IR type error [M0000], subtype violation:
-+  {#a; #b}
-+  {#a}
-+
--- a/test/run/ok/or-pattern-bug.diff-low.ok
+++ b/test/run/ok/or-pattern-bug.diff-low.ok
@@ -1,0 +1,8 @@
+--- or-pattern-bug.run
++++ or-pattern-bug.run-low
+@@ -0,0 +1,5 @@
++Ill-typed intermediate code after Desugaring (use -v to see dumped IR):
++or-pattern-bug.mo:6.9-6.19: IR type error [M0000], subtype violation:
++  {#a; #b}
++  {#a}
++
--- a/test/run/ok/or-pattern-bug.diff-low.ok
+++ b/test/run/ok/or-pattern-bug.diff-low.ok
@@ -1,8 +1,0 @@
---- or-pattern-bug.run
-+++ or-pattern-bug.run-low
-@@ -0,0 +1,5 @@
-+Ill-typed intermediate code after Desugaring (use -v to see dumped IR):
-+or-pattern-bug.mo:6.9-6.19: IR type error [M0000], subtype violation:
-+  {#a; #b}
-+  {#a}
-+
--- a/test/run/ok/or-pattern-bug.run-ir.ok
+++ b/test/run/ok/or-pattern-bug.run-ir.ok
@@ -1,5 +1,0 @@
-Ill-typed intermediate code after Desugaring (use -v to see dumped IR):
-or-pattern-bug.mo:6.9-6.19: IR type error [M0000], subtype violation:
-  {#a; #b}
-  {#a}
-
--- a/test/run/ok/or-pattern-bug.run-ir.ok
+++ b/test/run/ok/or-pattern-bug.run-ir.ok
@@ -1,0 +1,5 @@
+Ill-typed intermediate code after Desugaring (use -v to see dumped IR):
+or-pattern-bug.mo:6.9-6.19: IR type error [M0000], subtype violation:
+  {#a; #b}
+  {#a}
+
--- a/test/run/ok/or-pattern-bug.run-ir.ret.ok
+++ b/test/run/ok/or-pattern-bug.run-ir.ret.ok
@@ -1,0 +1,1 @@
+Return code 1
--- a/test/run/ok/or-pattern-bug.run-ir.ret.ok
+++ b/test/run/ok/or-pattern-bug.run-ir.ret.ok
@@ -1,1 +1,0 @@
-Return code 1
--- a/test/run/ok/or-pattern-bug.run-low.ok
+++ b/test/run/ok/or-pattern-bug.run-low.ok
@@ -1,5 +1,0 @@
-Ill-typed intermediate code after Desugaring (use -v to see dumped IR):
-or-pattern-bug.mo:6.9-6.19: IR type error [M0000], subtype violation:
-  {#a; #b}
-  {#a}
-
--- a/test/run/ok/or-pattern-bug.run-low.ok
+++ b/test/run/ok/or-pattern-bug.run-low.ok
@@ -1,0 +1,5 @@
+Ill-typed intermediate code after Desugaring (use -v to see dumped IR):
+or-pattern-bug.mo:6.9-6.19: IR type error [M0000], subtype violation:
+  {#a; #b}
+  {#a}
+
--- a/test/run/ok/or-pattern-bug.run-low.ret.ok
+++ b/test/run/ok/or-pattern-bug.run-low.ret.ok
@@ -1,0 +1,1 @@
+Return code 1
--- a/test/run/ok/or-pattern-bug.run-low.ret.ok
+++ b/test/run/ok/or-pattern-bug.run-low.ret.ok
@@ -1,1 +1,0 @@
-Return code 1
--- a/test/run/ok/or-pattern-bug.tc.ok
+++ b/test/run/ok/or-pattern-bug.tc.ok
@@ -1,0 +1,11 @@
+or-pattern-bug.mo:9.17-9.19: warning [M0146], this pattern is never matched
+or-pattern-bug.mo:11.10-11.33: warning [M0145], this pattern of type
+  {#a; #b; #c}
+does not cover value
+  #c
+or-pattern-bug.mo:12.11-12.13: warning [M0146], this pattern is never matched
+or-pattern-bug.mo:12.17-12.19: warning [M0146], this pattern is never matched
+or-pattern-bug.mo:12.10-12.27: warning [M0145], this pattern of type
+  {#d}
+does not cover value
+  #d
--- a/test/run/or-pattern-bug.mo
+++ b/test/run/or-pattern-bug.mo
@@ -1,0 +1,12 @@
+func ok1(((#a : {#a;#b}) or (#b : {#a;#b})) : {#a;#b}) {}; // accepted
+func ok2(((#a : {#a;#b}) or (#b : {#a;#b})) : {#}) {}; // accepted
+
+//func wrong(((#a : {#a}) or (#b : {#b})) : {#a;#b}) {}; // rejected
+
+func bad(#a or #b) {};
+
+
+func bad1(#a or #b : {#a}) {}; // accepted
+func bad2(#a or #b : {#a;#b}) {}; // accepted
+func bad3(#a or #b : {#a;#b;#c}) {}; //accepted
+func bad4(#a or #b : {#d}) {}; //accepted


rossberg:
Hm, with glb,
```
func (#a or #b) {}
```
would receive type `{#} -> ()`, which is not what you want, I think. Similarly,
```
func (-1 or 1) {}
```
would become `Nat -> ()`, which seems very odd.

For similar reasons, I don't see how this:
> the IR checked requires that type of a pattern is always a supertype of the expected type, so the pattern can consume the expected value
can be correct. Or at least I don't understand what notion of "type of a pattern" it is based on.

In a sense, the whole point of pattern matching is that each pattern tests for certain (possibly singleton) subranges of the value set of the scrutinee, i.e., a subtype.

For bindings OTOH I believe the pointwise lub needs to be taken on the inferred variable environment. E.g., for
```
func(#a(x : T) or #b(x : U)) { ...x... }
```
the two inferred environments have to be merged into `x : lub(T, U)`, while the function has type `{#a : T; #b : U} -> ()`.

The compatibility check, then, is necessary to rule out examples like
```
func(1 or true) {...}
```
The runtime representation is lossy, so there may be no way to distinguish `1` and `true` at runtime.

I'm not 100% sure why objects of different shape needed to be considered incompatible. It probably was exactly to rule out cases like
```
func ({a = _} or {}) {}
```
since that would be given type `{} -> ()` under lub, and therefor would allow to discover the presence of fields at runtime, i.e, provide a form of type reflection through the back door, in a way that cannot be written with `switch`.

nomeata:
Heh, I remember long debates from 6 years ago about the typing of patterns, and whether the type should describe the set of values the patten can scrutinize (expect), or the set of values it accepts (possibly singleton). Claudio above thinks about the first kind, and Andreas about the second kind, right?

My thought from the side lines: good luck hashing this out without actually describing a pattern using two types (one which gets lubed and one which gets glbed upon or'ing).

rossberg:
> I remember long debates from 6 years ago about the typing of patterns, and whether the type should describe the set of values the patten can scrutinize (expect), or the set of values it accepts (possibly singleton).

I'm not sure, in the presence of subtyping, the "set of values the patten can scrutinize" could always be taken to be Any, so what does that mean?

The problem only comes up when patterns occur in negative position, such that their type needs to be inferred. That is, as function arguments. In that situation, I think you naturally want to infer (an approximation of) the smallest type that includes all values in the pattern.

The notion of compatibility is independent, I think, since it (or at least a subrelation of it) is relevant regardless of the inference issue. It's also needed in the checking direction, to reject examples like
```
switch (x : Any) { case 1 ...; case true ...}
```

nomeata:
> I'm not sure, in the presence of subtyping, the "set of values the patten can scrutinize" could always be taken to be Any, so what does that mean?

I’d say that depends on your operational semantics: Does it get stuck matching `true` against the pattern `1`, or does it proceed to the next pattern. As you said above – our operational semantics is lossy, hence we not every pattern can be safely applied to every value.

In the last example, I’d say that `1` can scrutinze a value of type `Int`, and a value of type `Any` is given, so the program is rightfully rejected.

rossberg:
> In the last example, I’d say that 1 can scrutinze any value of type Int (but no other values), and a value of type Any is given, so the program is rightfully rejected.

I think you are essentially building in the notion of compatibility into your intuitive definition of "can scrutinize" here. ;)  I think it makes more sense to avoid such coupling and apply the compatibility check separately from type inference itself, analogous to checking mode.

rossberg:
Here is, in a way, the most fundamental example, that shows why compatibility really is a separate notion:
```
func f((_ : T) or (_ : U)) {...}
```
What type would you expect f to have?

That's almost a rhetoric question, since the syntax already suggests that it should be `(T or U) -> ()`, not `(T and U) -> ()`. Compatibility merely controls whether this combination is legal for an or-pattern, not what the resulting type is.

rossberg:
Stepping back, here is a formal argument. I think the core property you want for inference on patterns is that it is sound wrt checking, i.e.:

If C ⊢ p => t, then C ⊢ p <= t.

Looking at the checking rule for or-patterns,

C ⊢ p1 <= t
C ⊢ p2 <= t
------------------
C ⊢ p1 or p2 <= t

clearly, the above property can't hold when you take the glb in the corresponding inference rule. As a counter example, consider `#a or #b`.

But does it actually hold with lub?

My conjecture is that it does, but only because of the compatibility restriction on record types. Consider `{a = 0} or {}`. If we allowed to infer `{}` for this, then the property would fail, because `C ⊢ {a = 0} <= {}` does not hold for patterns.

I admit this all looks a bit odd. There must be something cleaner hiding in there. I wonder if some of the work on focussing could provide answers, but I doubt they consider subtyping. Perhaps the ultimate answer is that neither lub nor glb are actually wanted here, and instead some polarity-aware relation that behaves one way or the other depending on the polarity of each type connective. But that's just speculation.

crusso:
> Stepping back, here is a formal argument. I think the core property you want for inference on patterns is that it is sound wrt checking, i.e.:
> 
> If C ⊢ p => t, then C ⊢ p <= t.
> 
> Looking at the checking rule for or-patterns,
> 
> C ⊢ p1 <= t C ⊢ p2 <= t ––––––––––––– C ⊢ p1 or p2 <= t
> 
> clearly, the above property can't hold when you take the glb in the corresponding inference rule. As a counter example, consider `#a or #b`.
> 
> But does it actually hold with lub?
> 
> My conjecture is that it does, but only because of the compatibility restriction on record types. Consider `{a = 0} or {}`. If we allowed to infer `{}` for this, then the property would fail, because `C ⊢ {a = 0} <= {}` does not hold for patterns.

Doesn't the counter-example I gave above prove it doesn't hold when taking the lub?
```
func bad((((x,y) : (Nat,Int)) or ((y,x) : (Int,Nat)))): Nat { x }
```

crusso:
> The notion of compatibility is mostly independent, I think, since it (or at least a subrelation of it) is relevant regardless of the inference issue. It's also needed in the checking direction, to reject examples like
> 
> ```
> switch (x : Any) { case 1 ...; case true ... }
> ```

The code doesn't actually use `T.compatible` in the checking direction. Instead, it forces that the expected type is an appropriate subtype of the type of the pattern. That rules out these cases, IIUC. 

The only place place we use `T.compatible` is when inferring the type of an or-pattern.

crusso:
> Here is, in a way, the most fundamental example, that shows why compatibility really is a separate notion:
> 
> ```
> func f((_ : T) or (_ : U)) {...}
> ```
> 
> What type would you expect f to have?
> 
> That's almost a rhetoric question, since the syntax already suggests that it should be `(T or U) -> ()`, not `(T and U) -> ()`. Compatibility merely controls whether this combination is legal for an or-pattern, it doesn't affect the resulting type.

Actually,  for
```
func f( (n : Nat) or (n : Int)) {} 
```
I think I'd expect this to have type `Nat -> ()`, i.e. `Nat and Int` = `Nat` otherwise a `Nat` pattern will wind up encountering and `Int`, leading to the `bad` exploit above.

nomeata:
Is there a reason you are talking about `or` as if it’s something new, and not simply the same (wrt to typing and inference at least) as two separate `case`s?

crusso:
> Is there a reason you are talking about `or` as if it’s something new, and not simply the same (wrt to typing and inference at least) as two separate `case`s?
> 
> Ah, because we never do inference base on `cases` in a `switch`. Got it.

Perhaps we just shouldn't allow inference on or-patterns then and require switching to checking mode via an annotation.

rossberg:
> Doesn't the counter-example I gave above prove it doesn't hold when taking the lub?

There are two questions: the type of the pattern and the type of variables bound in the pattern.

With variable bindings, the rule for `or` of course needs to combine the types of the syntesised local variable environments. I was assuming this to happen by lub as well. So the function parameter in your example would have type (Int, Int), and x and y would also both have type Int, rejecting the bad example.

But I now realise that this indeed breaks the property I stated above, because C ⊢ (_ : (Nat, Int)) <= (Int, Int) does not hold. Hm...

> So maybe a different fix is for the desugaring to propagate the expected type into the translated patterns, rather than just preserving the inferred source annotations, but that seems fishy.

Yeah, I think that still wouldn't solve the problem for the source-level type system. I think the property I stated really ought to hold. Because what does inference mean otherwise?

> The code doesn't actually use T.compatible in the checking direction.

It is used indirectly, in a somewhat convoluted fashion: compatible is used in the definition of opaque, which is used in the definition of inconsistent, which is used when checking switch. There'd probably be a more symmetric way of doing this.

rossberg:
@nomeata, if we had ML-style lambdas with cases, like `fun p1 -> e1 | p2 -> e2`, then inference would indeed come up for cases, too. One additional catch of or patterns still exists, though, namely that their variable bindings span multiple alternatives, which is not the case with cases.

rossberg:
The only fix I can think of is tweaking the checking rule for type annotations to take the intersection instead of checking subtyping:

C ⊢ p <= t and t'
––––––––––––
C ⊢ p : t <= t'

Then the pattern `(x : Int : Nat)` is still well-typed, but so is `(x : Nat : Int)`, and in both case x has type Nat, which some may find odd. But I believe this change would preserve the soundness property for inference, while still rejecting the bad example (by taking the lubs on `or`).

Of course, simply disallowing inference for `or` is an option, too, and the most conservative fix.

rossberg:
@nomeata, why should `func(null) {}` have any other type than `Null → ()`? That would implicitly turn an exhaustive pattern into a non-exhaustive one, which seems super-fishy at best.

In the limit, with a precise enough type system including singletons, patterns and types essentially become the same thing (cf. CDuce, for example). It would be completely off if pattern union did not correspond to type union then, let alone do the opposite.

crusso:
> With variable bindings, the rule for `or` of course needs to combine the types of the syntesised local variable environments. I was assuming this to happen by lub as well. So the function parameter in your example would have type (Int, Int), and x and y would also both have type Int, rejecting the bad example.

Read the example more carefully: I switch the order of `x` and `y` precisely so their types don't change when you take the lub of the environment, even-though the lub of the patterns does yield `(Int,Int)`. So the example let's you cast an `Int` to a `Nat`.

ggreif:
I agree with @crusso that re-checking a pattern with an inferred type should always succeed, and that the example he provides (`func f2` above) fails to satisfy this criterion.

This very much sounds to me like the inherited/synthesised attributes problem to me (e.g. `repmin`), where we should accumulate information in `or`-patterns by `lub`, but finally assign the accumulated sum type to all leaf and branch nodes. Of course only when it is indeed a sum type (i.e. `Int`- or variant-like).

crusso:
Going back to the (abstracted) version of `bad`:

Checking does the right thing:

[nix-shell:~/motoko/src]$ rlwrap moc
Motoko compiler (source 0.9.1-9-g47c451ac4)
> func((((_,_) : (Nat,Int)) or ((_,_) : (Int,Nat))) : (Int,Int) ): Nat { 0 }
  ;
stdin:1.8-1.25: type error [M0117], pattern of type
  (Nat, Int)
cannot consume expected type
  (Int, Int)
 ```

but inference is wrong and accepts:

 ```
> func((((_,_) : (Nat,Int)) or ((_,_) : (Int,Nat))) ): Nat { 0 }
  ;
stdin:3.30-3.49: warning [M0146], this pattern is never matched
func : ((Int, Int)) -> Nat
```

**Comment by nomeata** [comment]
> @nomeata, why should func(null) {} have any other type than Null → ()? That would implicitly turn an exhaustive pattern into a non-exhaustive one, which seems super-fishy at best.

Because that’s what papterns do – they are inherently partial. Imagine you’d want to infernence on `case`:
```
switch (x){  case(null) { … }; … }
```
Of course the `null` pattern doesn’t force the scrutinee to be `null`, but you do want to force it to be of type `t <= opt Any`.

But you put your finger in the right would: Intuitively, we want function patterns to be total! So the question is not just “what values can the pattern scrutinze” but also “what values will cause the pattern not fail” (also used for coverage checking). These are two different things! And for `or` pattern or multiple cases, the first one needs to be glb’ed, and the second one needs to be lub’ed.

My conjecture is that 80% of the confusion around this stems from not separating the “set of values this can be applied to” and “the set of values this will succeed on”.

Also, we do precisely that already – for number literals:
```
> func foo(1) : Bool { true };
stdin:3.9-3.12: warning [M0145], this pattern of type
  Nat
does not cover value
  0 or 2 or _
let foo : Nat -> Bool = func
> foo(2);
stdin:3.9-3.12: execution error, argument value 2 does not match parameter list
```
Note that the type of `foo` isn't `singleton(1) → Bool` but `Nat → Bool`. So why should `func foo(null) : Bool { true }` be any different?

**Comment by crusso** [comment]
In a language with subtyping, the notion of equality is type dependent.. 

Perhaps the same is actually true for pattern matching, and we should just insist on knowing which type of values we are matching against and asking to infer one is just asking for trouble.

**Comment by rossberg** [comment]
@crusso:
> Read the example more carefully: I switch the order of x and y precisely so their types don't change

Ah, yes, I missed that. Clever. Right, so the inference isn't even type-sound.

> I don't think that's true either - for records, compatibility requires equal domains

Ah yes, you are correct. Record patterns are odd, since we allow omitting fields. But I believe it's implied for all other cases.

@nomeata:
> Because that’s what patterns do – they are inherently partial. Imagine you’d want to infer on case:
> ```
> switch (x){  case(null) { … }; … }
> ```

No, they are partial when they are partial. The pattern `null` is perfectly total at type `Null`. That's why you do not want to infer another, unsafe type. Likewise you don't want to infer {#a; #b} for the pattern {#a}.

> So why should func foo(null) : Bool { true } be any different?

Because the type system has a more precise type in this case that is actually safe.

As a general principle, I think you always want to infer the smallest type possible for which the pattern still type-checks as an expected type. Something like:

If C ⊢ p => t, then C ⊢ p <= t and for all t' <: t, ¬ C ⊢ p <= t'.

(I'm not sure if that currently holds for Motoko, though, even without or-patterns. It probably is violated by the rule for type annotations, which would have to return the underlying type, not the annotation. That actually would be more consistent anyway.)

@crusso:
> In a language with subtyping, the notion of equality is type dependent.

I don't think that's true. Subtyping is neither sufficient nor necessary to make equality type-dependent. Rather, I think this effect comes from dependent types or the likes (such as GADTs or singletons). In Motoko or most popular OO languages, equality is not type-dependent, as far as I can tell.

**Comment by nomeata** [comment]
🤷🏻 

Personally, I’d avoid calling this `safe`; Motoko is a safe language even though it infers `Nat` for the `1` pattern. And it’s important to keep this notion separate from safety in the sense of type safety (i.e. not inferring `Bool` for the `1` pattern, which can actually crash, and not just trap-with-incomplete-pattern).

I like your general principle – if applied at the _end_ of the process, i.e. after determining for the whole argument pattern the two sets I mentioned above, and then picking the smallest type that lies between the two sets.

This nicely justified why `func(1)` gets type `Nat →`, but `func(null)` gets type `Null →` and also why `func(#a or #b)` gets type `{#a; #b} →`.

(I don’t think you can do it compositionally at every step of the pattern, that just throws away too much information.)

Put differently: When inferring information about the pattern, gather a lower bound (values it succeds on) and an upper bound (values it can safely look at). When `or` in, the lower bound increases (`lub`, Andreas is happy) and the upper decreases (`glb`, Claudio is happy). At the end, pick a suitable type in between.

**Comment by nomeata** [comment]
> equality is not type-dependent, as far as I can tell

Claudio probably refers to this thing:
```
> {x = 1;} == {x = 2};
false : Bool
> ({x = 1;}:{}) == ({x = 2}:{});
true : Bool
```

**Comment by rossberg** [comment]
> (I don’t think you can do it compositionally at every step of the pattern, that just throws away too much information.)

Can you give an example of why that should not be the case? This essentially is a form of principal types property for patterns. If each pattern form has a principal type, then shouldn't they compose just fine?

(I'm not sure how you would recover it after the fact, or how you'd inductively prove the property in that case.)

> Claudio probably refers to this thing:

Ah, true. Overloading considered harmful.

**Comment by nomeata** [comment]
It's composable and has principal types if you consider a pattern type to consist of the tuple of the “compatible” and the “accepted” information, that much seems clear (to me at least).

Now my conjecture is that you really need both, if you want to infer a safe (doesn't crash), tight (no pattern match failure if avoidable) and useful (no smaller than needed) type for function patterns.

Hmmm, maybe it is possible; maybe only because our compatibility relation is relatively simple and can be inferred from the set of accepted values. But it'd wouldn't be a simple lub or glb then, as `#a or #b` seems to need a lub to allow both tags, but  `#b(5) or #b(true)` calls for an glb of the tag argument to stay safe. And if `#a or 1` has type `None`, then isn't it weird that that's the neutral element of the lub that you expect `or` to do?

My head spins trying to express it using a single type, while (so far, at least), everything seems natural, correct and compositional with two separate types (or sets of values for the accepted values, as used by the coverage checker).

**Comment by rossberg** [comment]
> but  `#b(5) or #b(true)` calls for an glb of the tag argument to stay safe.

That should be handled by compatibility, which rejects {#b : Nat} or {#b : Bool} . That is, in the `or` case, you just compute the lub _and_ check compatibility. You don't need to check compatibility anywhere else. That's what we already do.

In fact, I'd conjecture that our inference rules already produce the principal pattern type. (I was confused about annotated patterns before, those switch to checking mode anyway.)

**Comment by nomeata** [comment]
That might be. What exactly is the change to type annotation checking? Will it mean that `null : opt True` still infers `Null` (as the glb of `Null` and `opt True`)?

Btw, what type would you expect it infers for the pattern `[#a, #b]` (an array pattern)? `[{#}]` or `[{#a;#b}]` or something else?

**Comment by crusso** [comment]
🤯

**Comment by nomeata** [comment]
Ah, I think I have a good example now, small variant from above: Consider pattern 1 `opt true` and pattern 2 `opt ()`.

Presumably “the” pattern type is `opt Bool` and `opt ()` respectively.

What type should `opt true or opt 1` have? It can't just be the lub, as `opt Any` is clearly bogus. But outright rejection because of the compatibility check seems to be too harsh, as the patten _can_ be used at type `Null` or `opt None`, so rejection isn't very principled.

Hmm, but maybe it's pragmatic, in the sense that or-patterns where the type of values the patten can safely look turns to `None` are almost always a programmer mistake, like values glb'ed to `Any`, which we also warn about.

**Comment by crusso** [comment]
> > Edit: Strike that, that's used to check the results, not the pattern. For the patterns, the checking rules require the expected type to be decomposable in the right manner for each pattern form, which implies compatibility.
> 
> I don't think that's true either - for records, compatibility requires equal domains, but checking only requires narrowing of the domain, IIRC.

**Comment by crusso** [comment]
> Ah, I think I have a good example now, small variant from above: Consider pattern 1 `null or opt true` and pattern 2 `null or opt ()`.
> 
> Presumably “the” pattern type is `opt Bool` and `opt ()` respectively.
> 
> What type should `(null or opt true) or (null or opt 1)` have? It can't just be the lub, as `opt Any` is clearly bogus. But outright rejection because of a simple compatibility check seems to be too harsh, as the patten _can_ be used at type `Null` or `opt None`, so rejection isn't very principled.
> 
> Hmm, but maybe it's pragmatic, in the sense that or-patterns where the type of values the patten can safely look turns to `None` are almost always a programmer mistake, like values glb'ed to `Any`, which we also warn about.
> 
> (Edited: phrased the pattern using `null or opt …` to avoid distraction from possible patten match failure.)
On this branch (with glb):
```

[nix-shell:~/clean/motoko/src]$ rlwrap moc
Motoko compiler (source 0.9.1-9-g44342365c)
> func ((null or (? true)) or (null or (? 1))) {};
stdin:1.16-1.24: warning [M0146], this pattern is never matched
stdin:1.29-1.44: warning [M0146], this pattern is never matched
func : Null -> ()

> func (((null or (? true)) or (null or (? 1))) : ? None) {};
stdin:2.17-2.25: warning [M0146], this pattern is never matched
stdin:2.20-2.24: warning [M0146], this pattern is never matched
stdin:2.30-2.45: warning [M0146], this pattern is never matched
func : ?None -> ()

> func (((null or (? true)) or (null or (? 1))) : ? Nat) {};
stdin:3.20-3.24: type error [M0050], literal of type
  Bool
does not have expected type
  Int
> func (((null or (? true)) or (null or (? 1))) : ? Any) {};
stdin:4.20-4.24: type error [M0110], literal pattern cannot consume expected type
  Any
> 
```

**Comment by nomeata** [comment]
Maybe the first question, before “how do we calculate it” is “what do we even want” for patterns in function argument position without an all-surrounding type annotation. Is it:

1. The largest, most permissive type which the program is still type safe.
2. The largest, most permissive type which the program is still type safe and rules out partial patterns 
3. Among all types which are type safe, the smallest one that contains all values for which the pattern accepts.

Andreas seems to expect variant 3, right?


We could also consider disallowing incomplete patterns in function parameters, it would simplify this discussion a lot. And fix the problem with the side effect of trapping appearing in on the wrong side of an async… But we had reasons not do to that, did we?

**Comment by crusso** [comment]
> Maybe the first question, before “how do we calculate it” is “what do we even want” for patterns in function argument position without an all-surrounding type annotation. Is it:
> 
> 1. The largest, most permissive type which the program is still type safe.
> 2. The largest, most permissive type which the program is still type safe and rules out partial patterns
> 3. Among all types which are type safe, the smallest one that contains all values for which the pattern accepts.
> 
> Andreas seems to expect variant 3, right?
> 
> We could also consider disallowing incomplete patterns in function parameters, it would simplify this discussion a lot. And fix the problem with the side effect of trapping appearing in on the wrong side of an async… But we had reasons not do to that, did we?

Which camp am I in? 1, or maybe 2? If there's a simple syntactic characterization patterns that admit principal inferrence, I'd be happy for that that. Maybe patterns consisting of annotated full patterns and inductively tuples, records, variants wildcards and identifiers  would be enough. Not sure.

I've mentally moved on to the next `orrible feature (ICQC).

**Comment by rossberg** [comment]
> Andreas seems to expect variant 3, right?

Probably? Depends on what you mean by "permissive" in this context, given that we are talking about negative positions. Anyway, I thought the property I gave above is pretty explicit.

I want to reemphasise that I think it is crucial that typing of summand patterns is coherent with type union, e.g., the type of a pattern `#a or #b` is exactly `{#a} or {#b}`. You'll have a hard time justifying anything else, because that is the only thing that is coherent with a more precise type system. The treatment of other sorts of sums (e.g., numbers) is just a variation of that. For tuples there isn't much choice, the combination is gonna be pointwise, as it is for sum type arguments.

The only problem are products that have width subtyping, i.e., objects. In fact, if we hadn't allowed the omission of fields then they would be as simple as tuples. And compatibility disallows that in exactly the situation where you use `or`.

**Comment by crusso** [comment]
> > Andreas seems to expect variant 3, right?
> 
> Probably? Depends on what you mean by "permissive" in this context, given that we are talking about negative positions. Anyway, I thought the property I gave above is pretty explicit.
> 
> I want to reemphasise that I think it is crucial that typing of summand patterns is coherent with type union, e.g., the type of a pattern `#a or #b` is exactly `{#a} or {#b}`. You'll have a hard time justifying anything else, because that is the only thing that is coherent with a more precise type system. The treatment of other sorts of sums (e.g., numbers) is just a variation of that. For tuples there isn't much choice, the combination is gonna be pointwise, as it is for sum type arguments.
> 
> The only problem are products that have width subtyping, i.e., objects. In fact, if we hadn't allowed the omission of fields then they would be as simple as tuples. And compatibility disallows that in exactly the situation where you use `or`.

But this doesn't rule out my unsafe counter-example, does it?

**Comment by nomeata** [comment]
> Anyway, I thought the property I gave above is pretty explicit.

Do you mean this one?

> If C ⊢ p => t, then C ⊢ p <= t and for all t' <: t, ¬ C ⊢ p <= t'.

You are a step ahead of me then, I am still trying to get clarity what typing (`C ⊢ p : t `) should be, before thinking about inference or checking algorithms.

It's probably clear to you, but please do help me out. For values it seems clear to me, and for expressions my semantic understanding of typing is something like “If `⊢ e : t`, then if `e` terminates to a value `v`, then `⊢ v : t`. What is the corresponding intuition/goal of pattern typing? Can someone complete the sentence “If `⊢ p : t`…” for me?


I'd lean towards something like “If `⊢ p : t`, then `x : t ⊢ switch (x) {case (p) ()} : ()`” or something like that, but maybe that's too operational - but still better, I think, than some syntactic criterion; surely typing rules ought to be guided by some semantics idea.


Also, if `⊢ p : None` holds (as I think it does and should) doesn't that mean your criterion infers `None` for all patterns?

> Probably? Depends on what you mean by "permissive" in this context, given that we are talking about negative positions.

And with permissive I think I mean largest; preferring inferring `Any` over `None`, to permit the use of the pattern (and hence functions) with its inferred type in as many contexts as possible - indeed the dual of expressions, where `None` is most permissive.

**Comment by nomeata** [comment]
Another puzzle: Assuming we can agree on `_ : Any` (under all ideas of what typing of patterns should mean). What can the type of `#foo or _` be? Taking the lub would mean `#foo or _ : Any` which looks very dangerous

Ah, based on reading [the code](https://github.com/dfinity/motoko/blob/cf5ac77f1fb86b215065ea8fcfef4ecac1012817/src/mo_types/type.ml#L1272), these are reported as incompatible… safe, but not quite what I'd expect. (It keeps drawing me to wanting two judgments for patterns, one which would replace the blunter compatibility check, which I expect will either be unsafe (as Claudio exhibited above with Nat and Int) or too strict (as, one could argue, here).

**Comment by rossberg** [comment]
> I am still trying to get clarity what typing (`C ⊢ p : t`) should be

Well, there is no such relation in a bidirectional system, do you mean C ⊢ p <= t?

> if ⊢ p : None holds (as I think it does and should) doesn't that mean your criterion infers None for all patterns?

Yes, my criterion only defines how => has to relate to  <=.

As for the other, typing all patterns with None obviously is as useless as typing all expressions with Any. For expressions, you want the smallest possible type. Dually, for patterns you want the largest possible type. The question is what "possible" means in terms of match failures.

In an ideal world of a perfectly expressible type system, we could assign each pattern a type precisely describing the domain of values it can handle, so that no match failures are allowed by the type system, but no values are excluded either. But other than CDuce, I'm not aware of a language reaching this ideal.

So we need to approximate. There are two canonical choices: from above or from below.

The approximation that all other languages use for pattern matching is the former, i.e., loosening to the smallest type they can express that includes (i.e., is larger than) the "ideal" type, and handle the loss of accuracy via exhaustiveness warnings.

The alternative would be to go to the largest type that is _included in_ (i.e., is smaller than) the hypothetical ideal type. But that would be both surprising and not very useful -- my pattern (1 or 2) includes a case for 1 but I'm not allowed to match on it?

Note however that with both choices, `#a or #b` would still have the precise type `{#a} or {#b}`. So the glb definitely is the wrong answer for `or`, because that would yield `{#}`, which is almost None and similarly useless.

> Ah, based on reading [the code](https://github.com/dfinity/motoko/blob/cf5ac77f1fb86b215065ea8fcfef4ecac1012817/src/mo_types/type.ml#L1272), these are reported as incompatible… safe, but not quite what I'd expect.

There are two things at play:

1. subtyping describes which values are safe to use in contexts expecting supertypes,
2. compatibility describes which values can be _distinguished_ at run time to allow scrutinising them in the same context.

These are rather different properties. In particular, compatibility is primarily driven by implementation choices (how rich do we want to make our value representation), not semantics per se. And it is only needed to constrain patterns with multiple cases.

**Comment by rossberg** [comment]
Yeah, I would rather formulate it by separate typing rules for case sequences:
```
e : t
c* : t → u
–––––––––––––––––––– [E-swiich]
switch e { c* } : u

p : t      e : u
––––––––––––––––– [C-case]
case p e : t → u

c1* : t1 → u1      c2* : t2 → u2
t1, t2 compatible
–––––––––––––––––––––––––––––––––––– [C-seq]
c1*; c2* : (t1 or t2) → (u1 and u2)


––––––––––––––––––––– [C-empty]
epsilon : None → Any
```

> Isn't it odd if inference rejects #a or _ (incompatibility with Any), but it can be typed with a type annotation?

Perhaps, but all that is saying is that our inference is incomplete, which isn't news. You could of course try to devise an inference that is more powerful and derives {#a} for this, but that would no longer be a simple bidirectional system.

**Comment by rossberg** [comment]
Maybe we should rename "compatible" to "distinguishable" for more clarity.

**Comment by crusso** [comment]
These rules still look unsound to me. Don't they allow?

```
switch -1 {
   case (n : Nat) n;
   case (i  : Int) 0
} : Nat;
```

**Comment by ggreif** [comment]
> These rules still look unsound to me. Don't they:
> 
> ```
> switch -1 {
>    case (n : Nat) n;
>    case (i  : Int) 0
> } : Nat;
> ```
> 
> coercing -1 to a `Nat`?
``` Motoko
  switch (-1) {
       case (n : Nat) n;
       case (i  : Int) 0
    };
```
> stdin:3.12-3.19: type error [M0117], pattern of type
>   Nat
> cannot consume expected type
>   Int

luc-blaeser:

PR #4002, opened 2023-05-24
bugfix: improve error message to include object fields

crusso:
Fixes a bug in error reporting that caused moc to fail to report the expected type of a record pattern, eliding all fields.

Before:
```
[nix-shell:~/clean/motoko/src]$ rlwrap moc
Motoko compiler (source 0.9.1-7-gcf341ee2b)
> func f( {z:Nat} : {x:Nat; y:Bool}) {};
stdin:1.10-1.15: type error [M0119], object field z is not contained in expected type
  {}
> func f( {z:Nat} : {x:Nat; y:Bool}) {};
```

After:
```
[nix-shell:~/clean/motoko/src]$ rlwrap moc
Motoko compiler (source 0.9.1-7-gcf341ee2b-dirty)
> func f( {z:Nat} : {x:Nat; y:Bool}) {};
stdin:1.10-1.15: type error [M0119], object field z is not contained in expected type
  {x : Nat; y : Bool}
> 
```


ggreif (approved):
LGTM! @crusso you have to re-accept `no-timer-canc.mo` `no-timer-set.mo`.

PR #4003, opened 2023-05-25
feat: composite queries

crusso:
Add new form of composite query that:

- can call other queries and composite queries
- can't be called from updates
- can't do async expressions (no way to queue up a callback since state is lost)
- can't call local async or async* (because those might contain update calls)

From the (informal) doc:

# Composite query functions

Although queries can be fast, when called from a frontend, yet trusted though slower, when called from an actor, they are also limited in what they can do. In particular, they cannot themselves issue further messages, including queries.

To address this limitation, the Internet Computer supports another flavour of query function called a composite query. Like plain queries, the state changes made by a composite query are transient, isolated and never committed. Moreover, composite queries cannot call update functions, including those implicit in async expressions (which require update calls under the hood). Unlike plain queries, composite queries can call query functions and composite query functions, on the same and other actors, but only provided those actors reside on the same subnet.

As a contrived example, consider generalising the previous Counter actor to a class of counters. Each instance of the class provides an additional composite query to sum the values of a given array of counters:
```
actor class Counter () {

  var count = 0;

  // ...

  public shared query func peek() : async Nat {
    count
  };

  public shared composite query func sum(counters : [Counter]) : async Nat {
    var sum = 0;
    for (counter in counters.vals())  {
      sum += await counter.peek();
    };
    sum
  }

}
```

Declaring `sum` as a composite query enables it call the peek queries of its argument counters.

While update message can call plain query functions, they cannot call composite query functions. This distinction, which is dictated by the current capabilites of the IC, explains why query functions and composite query functions are regarded as distinct types of shared functions.

Note that the composite query modifier is reflected in the type of a composite query function:
```
  sum : shared composite query ([Counter]) -> async Nat
```

Since only a composite query can call another composite query, you may be wondering how any composite query gets called at all? The answer to this chicken-and-egg problem is that composite queries are initiated outside the IC, typically by an application (such as a browser frontend) sending an ingress message invoking a composite query on a backend actor on the IC.

c.f. https://github.com/dfinity/interface-spec/pull/144

- [X] positive tests
- [X] negative tests
- [X] RTS support (subtyping, IDL validation etc)
- [x] doc
- [x] gc (to GC or not at end of each message)? We currently do a (redundant?) GC for composite queries
- [x] imports
- [x] inspect_message support? Since composite queries can't be called as updates, perhaps they don't matter at all?
- [x] I had to modify cycles adding and refund logic to avoid calling ic0.call_cycles_add128/ic0.cycles_refunded128 unless necessary - these aren't supported for composite query callbacks (see spec).
- [x] More accurate Lifetime modelling - we currently reuse state Lifecycle.InUpdate since callbacks are always in this state (and changeing that statically is tricky without introducing new forms of callback for CompositeQueries. Adding a new state would allow us to suppress GC at the end of each message, if desired, and more elegantly rule out forbidden ops like adding non-zero cycles, observing refunds. Of course, we could also just set a global to disable these until all state-changes are aborted at the end. Might be simpler.
- [x] regenerate ok test output once ic-ref available again by merge with master. At the moment, I can manually run the new tests, but can't accept their output. They pass, trust me.
- [x] better error messages, distinguishing capability  (send update, send query, send composite query, perhaps) required.
- [ ] enable ic-ref tests (once ic-ref supports composite queries)
- [x] add keyword `composite` to idl, local highlight.js


doc/md/examples/grammar.txt
```diff
@@ -12,15 +12,22 @@
...
+    'shared' <mode_opt>
     'query'
+    'composite' 'query'
 
 <shared_pat_opt> ::= 
     <empty>
-    'shared' 'query'? <pat_plain>?
+    'shared' <mode_opt> <pat_plain>?
     'query' <pat_plain>?
+    'composite' 'query' <pat_plain>?
```
crusso, line 30: simplify to
```
<query> ::= composite? query

<func_sort_opt> 
<empty>
shared <query>?
<query>

<shared_pat_opt> ::= 
    <empty>
    'shared' <query>? <pat_plain>?
    <query> <pat_plain>?
```

crusso: Done

doc/md/language-manual.md
```diff
@@ -1264,7 +1272,10 @@ The argument type of `inspect` depends on the interface of the enclosing actor.
 
 -   `arg : Blob`: the raw, binary content of the message argument;
 
--   `msg : <variant>`: a variant of *decoding* functions, where `<variant> == {…​; #<id>: () → T; …​}` contains one variant per shared function, `<id>`, of the actor. The variant’s tag identifies the function to be called; The variant’s argument is a function that, when applied, returns the (decoded) argument of the call as a value of type `T`.
+-   `msg : <variant>`: a variant of *decoding* functions, where `<variant> == {…​; #<id>: () → T; …​}` contains one variant per `shared` or `shared query` function, `<id>`, of the actor
+    (`shared composite query` functions are ignored).
```
luc-blaeser, line 1276: This means that `inspect` is not called for composite queries? Or is the message argument for `shared composite query` simply mapped to `shared query`?

doc/md/language-manual.md
```diff
@@ -83,10 +83,12 @@ All comments are treated as whitespace.
 The following keywords are reserved and may not be used as identifiers:
 
 ``` bnf
-actor and assert async async* await await* break case catch class continue debug
-debug_show do else flexible false for from_candid func if ignore import
-in module not null object or label let loop private public query return
-shared stable switch system throw to_candid true try type var while with
+
+actor and assert async async* await await* break case catch class
+composite continue debug debug_show do else flexible false for
```
luc-blaeser, line 88: Including `composite` as reserved keyword impacts backward compatibility (former identifiers named `composite` become invalid). This would technically be a breaking change, isn't it?

doc/md/language-manual.md
```diff
@@ -2022,6 +2036,17 @@ The context type is a record to allow extension with further fields in future re
 
 :::
 
+Shared functions have different capabilities dependent on their qualification as `shared`, `shared query` or `shared composite query`.
+
+A `shared` function may call any `shared` or `shared query` function, but no `shared composite query` function.
```
luc-blaeser, line 2041: I heard that update calls cannot invoke composite queries. Just for curiosity: Is it a technical reason? Or a conceptual aspect?

doc/md/language-manual.md
```diff
@@ -2022,6 +2036,17 @@ The context type is a record to allow extension with further fields in future re
...
+A `shared query` function may not call any `shared`, `shared query` or `shared composite query` function.
+A `shared composite query` function may call any `shared query` or `shared composite query` function, but no `shared` function.
+
+All varieties of shared functions may call unshared functions.
+
+Composite queries, though composable, can only be called externally (from a frontend) and cannot be initiated from an actor.
+
```
luc-blaeser, line 2048: Maybe we should mention that each composite query call rolls back its state on function exit, and also does not pass state changes to sub-query or sub-composite-query calls. Therefore, repeated calls (which includes recursive calls) have different semantics than in classical procedural execution (like with update calls).

src/codegen/compile.ml
```diff
@@ -7910,6 +7918,8 @@ module FuncDec = struct
         Lifecycle.trans env Lifecycle.Idle
       | Type.Shared Type.Query ->
         Lifecycle.trans env Lifecycle.PostQuery
+      | Type.Shared Type.Composite ->
+        Lifecycle.trans env Lifecycle.Idle
```
luc-blaeser, line 7922: Just to double check: The subsequent state of the lifecycle is `Idle` and not `PostQuery` for composite queries?

src/mo_frontend/typing.ml
```diff
@@ -694,28 +708,30 @@ and check_con_env env at ce =
...
+        (T.Con(c,[])::ts, at::ats)
+     | (C.CompositeAwaitCap c | C.CompositeCap c) when sort = T.Shared T.Query || sort = T.Shared T.Composite ->
+        (T.Con(c,[])::ts, at::ats)
      | C.ErrorCap
      | C.QueryCap _
-     | C.NullCap ->
+     | C.NullCap
+     | _ ->
         error env at "M0047"
           "send capability required, but not available (need an enclosing async expression or function body)"
```
luc-blaeser, line 727: Probably, the error message is not very specific if the call between the modes update/query/composite query is not supported.

luc-blaeser (approved):
I like that Motoko offers composite query calls with static checking.
Nice compiler implementation and documentation of this feature.

doc/md/language-manual.md
crusso: Inspect is indeed not called for composite queries (in my understanding) so we don't even include them in the variant type.

doc/md/language-manual.md
crusso: Yes, it is, but probably won't brake many. Maybe we should bump to 0.10?

src/mo_frontend/typing.ml
crusso: Yes, I was considering revising those (maybe splitting send capability into update query and composite query capability.
Maybe this is the right time to do it.

src/codegen/compile.ml
crusso: Yes, this is an unfortunate hack. For queries, you can just generate the atomic body and enter PostQuery. But composite queries are not atomic, so you need to enter some state after every callback. Unfortunately, the code for generating callbacks always enters InUpdate and its hard to make the transition conditional.
I couldn't see any (easy) way to enter a different state without complicated the rest of the code to track whether we are in an ordinary callback or a composite callback. I could may set and test a global but that seemed gross too.

As far as I know, the Lifecycle stuff is just for sanity checking, so I think we are just losing some accuracy here.

doc/md/language-manual.md
crusso: Will do, I was shying away from this ;->

doc/md/language-manual.md
crusso: Just to clarify, even for querie method, inspect is only called if the method is invoked as an (ingress) update call (with consensus). It's not used for the faster query calls (no-concensus). Very confusing.

Since composite query methods can't be called with (ingress) update calls, just (ingress) query calls, there is no need to include them here.

doc/md/language-manual.md
luc-blaeser: Maybe, we could also just reference the general design description of that feature, because it is not a Motoko-specific aspect?

src/mo_frontend/typing.ml
luc-blaeser: This is nothing important, could be also done later. I guess we can do that as part of an error message improvement work (together as a team across the entire compiler).

doc/md/language-manual.md
crusso: When an update calls a query, the query is actually executed in replicated mode for concensus on the result (state change is still discarded). 
Similary, inspect message is only called for query functions that are executed in update mode (with concensus), not ordinary fast queries.
I suspect that replicated composite queries are much harder to implement and haven't been tackled yet, and might not be.

doc/md/language-manual.md
crusso: Another alternative would be to use `composite_query` (ugly IMO) or `query*` (confusing, subtle IMO)

doc/md/language-manual.md
luc-blaeser: Thanks for the explanation.

doc/md/language-manual.md
crusso: I've added some clarification, plus an example.

doc/md/language-manual.md
luc-blaeser: I think `composite query` is much nicer to read. (Another possibility would be to introduce non-reserved keywords but the complexity is not worth it.) I would keep the current syntax.

doc/md/language-manual.md
luc-blaeser: Great explanation!

crusso:
Luc, if you could take another (quick) look, I've improved the error messages (a bit) and added some more doc about side-effects.

doc/md/language-manual.md
```diff
@@ -2008,12 +2023,17 @@ The function expression `<shared-pat>? func < X0 <: T0, …​, Xn <: Tn > <pat1
...
-Note that a `<shared-pat>` function may itself be `shared <pat>` or `shared query <pat>`:
+Note that a `<shared-pat>` function may itself be `shared <pat>` or `shared query <pat>` or  `shared composite query <pat>`.
 
 -   A `shared <pat>` function may be invoked from a remote caller. Unless causing a trap, the effects on the callee persist beyond completion of the call.
 
 -   A `shared query <pat>` function may be also be invoked from a remote caller, but the effects on the callee are transient and discarded once the call has completed with a result (whether a value or error).
 
+-   A `shared composite query <pat>` function may only be invoked as an ingress message, not from a remote caller.
+    Like a query, the effects on the callee are transient and discarded once the call has completed with a result (whether a value or error).
+    In addition, intermediate state changes made by the call are not obserable by any of its own `query`  or `composite query` callees.
```
crusso, line 2034:
```suggestion
    In addition, intermediate state changes made by the call are not observable by any of its own `query`  or `composite query` callees.
```

test/run-drun/idl-composite.mo
```diff
@@ -0,0 +1,35 @@
...
+
+       // can deserialize as composite query
+       let o1 = from_candid blob : ? (actor { f : composite query () -> async (); });
+       assert o1 == ?This;
+
+       // can't deserialize as query
+       let o2 = from_candid blob : ? (actor { f : /* composite */ query () -> async (); });
+       assert o2 == null;
+
+       // can't deserialized as update
```
crusso, line 19:
```suggestion
       // can't deserialize as update
```

test/run-drun/idl-composite.mo
```diff
@@ -0,0 +1,35 @@
...
+       let o3 = from_candid blob : ? (actor { f : /* composite  query */ () -> async (); });
+       assert o2 == null;
+
+       // can't deserialized as oneway
```
crusso, line 23:
```suggestion
       // can't deserialize as oneway
```

luc-blaeser:
> @luc-blaeser, if you could take another (quick) look, I've improved the error messages (a bit) and added some more doc about side-effects.

Very clear explanation. I would say it is ready to merge...

doc/md/actors-async.md
```diff
@@ -238,3 +239,68 @@ The last two lines above *instantiate* the actor class twice. The first invocati
...
+```
+
+When `state` is `0`, a call to `test` returns
+
+```
+{s0 = 0; s1 = 0; s2 = 0; s3 = 3_000}
+```
+
+because none of the local updates to `state` are visible to any of the callers or callees.
+
```
ggreif, line 305: Maybe add a note that composite queries cannot be run in replicated mode (as opposed to normal queries).

src/idllib/compile_js.ml
```diff
@@ -112,6 +112,7 @@ let pp_mode ppf m =
   match m.it with
   | Oneway -> str ppf "'oneway'"
   | Query -> str ppf "'query'"
+  | Composite -> str ppf "'composite_query'" (* TBR *)
```
crusso, line 115:
```suggestion
  | Composite -> str ppf "'composite_query'"
```

PR #4004, opened 2023-05-26
only warn when bindig types are incompatible

ggreif:


test/run/ok/or-pattern-mismatch.tc.ok
```diff
@@ -0,0 +1,19 @@
...
+  Int
+or-pattern-mismatch.mo:6.5-6.32: warning [M0185], imprecise type for alternative pattern variables a5 given
+  Char
+  Int
+or-pattern-mismatch.mo:6.5-6.21: warning [M0146], this pattern is never matched
+or-pattern-mismatch.mo:6.6-6.20: warning [M0146], this pattern is never matched
+or-pattern-mismatch.mo:9.10-9.41: warning [M0185], imprecise type for alternative pattern variables a6 given
+  Nat
+  Bool
+or-pattern-mismatch.mo:9.10-9.41: warning [M0185], imprecise type for alternative pattern variables a6 given
```
ggreif, line 17: these are duplicated, but there must be a way to suppress the first one

ggreif: f58483c91

src/mo_frontend/typing.ml
```diff
@@ -1874,13 +1874,15 @@ and infer_pat' env pat : T.typ * Scope.val_env =
         display_typ_expand t2;
     if T.Env.keys ve1 <> T.Env.keys ve2 then
       error env pat.at "M0184" "different set of bindings in pattern alternatives";
-    let check_same_bind_type bind t1 t2 = if not (T.eq t1 t2) then
-      error env pat.at "M0185" "types for alternative pattern variables %s differ%a%a"
+    let warn_lossy_bind_type bind t1 t2 = if not (T.eq t1 t2) then
```
crusso, line 1877: Maybe factor our the common code in both inference and checking?

src/mo_frontend/typing.ml
```diff
@@ -1874,13 +1874,15 @@ and infer_pat' env pat : T.typ * Scope.val_env =
...
+      let common = T.lub t1 t2 in
+      if T.(not (eq t1 common || eq common t2)) then
```
crusso, line 1879: I think elsewhere we use predicate `consistent` to  detect information. Maybe follow that pattern, for, err, consistency.

ggreif: If I read the `inconsistent` predicate correctly, it returns `true` when the `lub` is pushed right up to `Any`. I try to warn when both are smaller than the `lub`. So I changed the wording only.

src/mo_frontend/typing.ml
```diff
@@ -135,6 +135,14 @@ let error_in modes env at code fmt =
 
 let plural cs = if T.ConSet.cardinal cs = 1 then "" else "s"
 
+let warn_lossy_bind_type env at bind t1 t2 =
+  let common = T.lub t1 t2 in
+  if T.(not (eq t1 common || eq common t2)) then
```
crusso, line 140: I think else where Andreas used 'inconsistent common [t1; t2]` which I think only warns if the lub degrades to `Any` but none of the branches have type `Any`.

src/mo_frontend/typing.ml
```diff
@@ -135,6 +135,14 @@ let error_in modes env at code fmt =
...
+    warn env at "M0185" "inconsistent type for alternative pattern variables %s given%a%a"
```
crusso, line 141:
```suggestion
    warn env at "M0185" "pattern variabels %a has type%a\nbecause its types in the pattern alternatives are inconsistent,\n type in left pattern is%a\nright type in right pattern is%a"
```

src/mo_frontend/typing.ml
ggreif: Right, but that is too coarse for my taste.

src/mo_frontend/typing.ml
crusso:
```suggestion
    warn env at "M0185" "pattern variable %a has larger type%a\nbecause its types in the pattern alternatives are unrelated smaller types: \ntype in left pattern is%a\ntype in right pattern is%a"
```
(if you aren't using consistency, just subtyping, maybe this is more appropriate).

ggreif: Now you know why I was hesitant to come up with a text :-)

src/mo_frontend/typing.ml
```diff
@@ -135,6 +135,15 @@ let error_in modes env at code fmt =
 
 let plural cs = if T.ConSet.cardinal cs = 1 then "" else "s"
 
+let warn_lossy_bind_type env at bind t1 t2 =
+  let common = T.lub t1 t2 in
+  if T.(not (eq t1 common || eq common t2)) then
+    warn env at "M0185" "pattern variable %s has type%a\nbecause its types in the pattern alternatives are inconsistent:\ntype in left pattern is%a\ntype in right pattern is%a"
```
ggreif, line 141:
```suggestion
   warn env at "M0185" "pattern variable %a has larger type%a\nbecause its types in the pattern alternatives are unrelated smaller types: \ntype in left pattern is%a\ntype in right pattern is%a"
```

src/mo_frontend/typing.ml
rossberg: Right. That would warn for `(x : Nat) or (x : Bool)` (which is inconsistent anyway) but not for `(x : (Int, Nat)) or (x : (Nat, Int))`. In a language with subtyping, it's always a bit fishy to make subtyping an error arbitrarily, though for inferred bindings the situation arguably is different.

As an aside, you can simplify your condition to
```suggestion
  if not T.(sub t1 t2 || sub t2 t1) then
```
which should be slightly more efficient.

ggreif: Thanks, I missed this comment before merging, but it is applied now on the other branch as 1655c8ef1.

rossberg: Thinking about this some more, I'd suggest to change the logic to use `inconsistent` after all. It is weird to accept
```
(x1 : Nat, x2 : Int) or (x1 : Int, x2 : Nat)`
```
without warning, but warn about
```
(x : (Nat, Int)) or (x : (Int, Nat))`
```
That is the kind of thing that violates compositionality, and it is annoying when seemingly innocent refactorings of code can introduce warnings.

PR #4011, opened 2023-05-31
refactoring: define `define_pat` in terms of `match_pat`

crusso:
There was lots of duplication between `define_pat` (used for `let`) and `match_pat`, leading to a strange mutually recursive  implementation in the case of the new variable-binding `or`-patterns.

This refactoring defines `define_pat` in terms of `match_pat`.

- [x] Update test output once `ic-ref` available. The change in output is because `define_pat` no longer pin-points the sub-pattern that failed and just reports the entire pattern as failing to match.


test/run-drun/ok/issue-3910.diff-low.ok
```diff
@@ -5,4 +5,4 @@
  f1
  f2
 -issue-3910.mo:17.10-17.35: execution error, argument value {trap = true} does not match parameter list
-+issue-3910.mo:17.20-17.32: execution error, value true does not match pattern
```
ggreif, line 8: I get that the change in the diagnostic is due to the fact that the whole pattern gets reported whereas before only the failing field did...

ggreif (approved):
Nice refactoring! LGTM.

PR #4012, opened 2023-05-31
temporary: no `or`-pattern inference

ggreif:
As #3993 noted, our algorithm for inferring types of `or`-patterns is flawed, and its result is rejected by the IR type checker. Since IR-checking can be disabled, unsoundness may creep into released binaries (see #4001).

The conservative solution for this is to disallow pattern type inference for function definitions and require argument type annotations for now. This will be reverted as soon as we have a sound inference algorithm in place. (It is possible, as OCaml demonstrates with polymorphic variants).

There are attempts to mitigate the situation, e.g. with #4026.


src/mo_frontend/typing.ml
```diff
@@ -1864,7 +1864,9 @@ and infer_pat' env pat : T.typ * Scope.val_env =
     let t1, ve = infer_pat env pat1 in
     T.Variant [T.{lab = id.it; typ = t1; depr = None}], ve
   | AltP (pat1, pat2) ->
-    let t1, ve1 = infer_pat env pat1 in
+    error env pat.at "M0184"
+        "or-pattern needs annotation";
```
crusso, line 1868:
```suggestion
        "Cannot infer the type of this or-pattern, please add a type annotation";
```

kentosugama (approved):

PR #4016, opened 2023-06-01
chore: bump `nixpkgs` to 23.05

ggreif:
This is needed for tracking `ic-hs` which did the jump recently. `nixpkgs-22.11` got deprecated too.

Wasmtime generates different errors on trapping now.

_Note_: This `nixpkgs` revision is in synchronised with `ic-hs`'.

Remains to be debugged:
- [x] why is `shell` failing in CI, but not on the M1 Mac? — because `unit` tests used a `musl`-based toolset for OCaml builds, without obvious reason.
- [ ] why is the hash updater failing?
- [x] why is `json` version 4.1 still included in the closure? — due to https://github.com/NixOS/nixpkgs/blob/nixos-23.05/pkgs/development/ocaml-modules/logs/default.nix, but that is optional
- [x] setting `dontDetectOcamlConflicts = true;` in the `shell` derivation papers over the problem


default.nix
```diff
@@ -83,8 +83,8 @@ let commonBuildInputs = pkgs:
     pkgs.ocamlPackages.ppxlib
     pkgs.ocamlPackages.ppx_blob
     pkgs.ocamlPackages.ppx_inline_test
-    pkgs.ocamlPackages.ocaml-migrate-parsetree
-    pkgs.ocamlPackages.ppx_tools_versioned
+    # pkgs.ocamlPackages.ocaml-migrate-parsetree
```
ggreif, line 86: this looks like a backporting opportunity

nomeata (approved):
Would be good to have the shell built though

ggreif:
> Would be good to have the shell built though

Sure, I'll debug, but if it remains mysterious then I'll leave an issue. Also I wonder if @crusso can enter `nix-shell` on this branch.

default.nix
```diff
@@ -426,9 +423,9 @@ rec {
       '';
     };
 
-    unit = ocamlTestDerivation {
+    unit = testDerivation {
```
ggreif, line 426: it doesn't seem necessary to run unit tests in `musl` (static) link environment. @nomeata do you agree? It still passes and won't pull in duplicate OCaml library versions into the `moc-shell`.

.github/workflows/test.yml
```diff
@@ -40,6 +40,8 @@ jobs:
         nix-build --max-jobs 1 -A drun
         nix-store --gc --print-roots | grep /motoko/
 
+    - run: nix-store -q --tree $(nix-instantiate -A shell)
```
ggreif, line 43: this can be removed now

ggreif:
> Would be good to have the shell built though

Yep this is fixed now. `unit` tests now use the `.so`-toolset for OCaml. Do you think there is a catch?

crusso (approved):
Tested on my Linux machine.

PR #4018, opened 2023-06-04
fix(mo-doc): show help menu when passing unexpected positional arg

rvanasa:
At the moment, `mo-doc` silently accepts positional arguments, which can lead to seemingly incorrect behavior (such as in https://github.com/dfinity/motoko/issues/4019). This PR fixes this by showing the `--help` usage message when a positional argument is passed to the command.


crusso (approved):
Sorry for the delay, was on holiday and hoped someone else would approve...

PR #4022, opened 2023-06-06
Incremental GC Tuning: Allow collecting the allocation partition

luc-blaeser:
This is another minor optimization for the incremental GC.

This GC heuristics decides whether the new objects in the current allocation partition should be collectable by opening a new allocation partition. This is a tradeoff between allocating a new partition and retaining young objects in the current GC run. 

The current configuration releases the allocation partition when it is an evacuation candidate and it is nearly full, > 95% occupied. 

This saves memory only for a specific scenario, while the average gain is very small. 

Allocated WASM memory:
Case | Unoptimized | Optimized | Saving
-----|------------|----------|-------
asset-storage | 160 MB | 96 MB | 40%
average | 298 MB | 296 MB | 0.6%

This change has no relevant impact on the runtime performance in the benchmark. 

I am not sure whether this optimization is worth it.

Tuning:
* Decreasing the occupation threshold (e.g. to 75%) only reduces memory a little bit (<1%), while increasing runtime costs.
* Too low occupation thresholds (<50%) even increase memory again because of too eager partition switches.

### DIFF
--- a/rts/motoko-rts-tests/src/gc/incremental/partitioned_heap.rs
+++ b/rts/motoko-rts-tests/src/gc/incremental/partitioned_heap.rs
@@ -28,7 +28,7 @@ pub unsafe fn test() {
 }
 
 unsafe fn test_normal_size_scenario() {
-    let mut heap: PartitionedTestHeap = create_test_heap();
+    let mut heap = create_test_heap();
     let occupied_partitions = 2 + heap.heap_pointer() / PARTITION_SIZE;
     test_allocation_partitions(&heap.inner, occupied_partitions);
     test_iteration(&heap.inner, 1024);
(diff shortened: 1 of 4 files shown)


crusso:
I don't really know, tbh. I'll leave the decision up to you.

Curious, is there any way of making the GC more generational by only focussing on the allocation partition when it gets full? It should contain the young objects, right?

luc-blaeser:
Making the incremental GC also more generational is a very interesting idea. However, it would need a remembered set for the allocation partition (denoting the young generation) in order to limit the marking phase and updating phase to the young objects. For simplicity, I left this feature out but it is certainly something we could try as an optimization in the future.

I put this PR in the backlog, as I believe it is maybe a too specific optimization and simplicity is probably better in this case. We can still revive the PR if needed.

PR #4023, opened 2023-06-06
chore: reinstate the `ic-ref-run` test

ggreif:
which has been deactivated in #3935 due to missing `ic0.is_controller` interface. (It was missing because we couldn't import the `ic-hs` changes.)

The change in the test case is due to https://github.com/dfinity/ic-hs/pull/179.


nomeata (approved):

ggreif:
@Mergifyio refresh

PR #4024, opened 2023-06-07
Incremental GC: Fix the live set computation

luc-blaeser:
Objects allocated during a GC run may be reclaimed under a very specific condition:
* Mark stack pages allocated in a partition that is no longer the allocation partition when the evacuation phase starts.

As a consequence, the following may happen:

A debug assertion can be violated:
```
debug_assert!(growth_during_gc.0 <= dynamic_size.as_usize() as u64);
```
In release mode, an arithmetic overflow may cause invalid live set statistics.

A solution is to simplify the computation of the live set, by only considering the final dynamic heap size at GC completion.

This issue has been observed in the PR https://github.com/dfinity/motoko/pull/4022 that changes allocation partitions more eagerly during GC.


crusso (approved):
I'll just trust you on this one.

ggreif:
@luc-blaeser is this worth an entry in the `Changelog.md`?

luc-blaeser:
> @luc-blaeser is this worth an entry in the `Changelog.md`?

Maybe not, this is a very detailed aspect, not sure whether anybody hit this case. Luckily, no serious issue, it only relates to live-set statistics.

PR #4026, opened 2023-06-07
experiment: attempt to fix `or`-pattern inference

ggreif:
When inferring a pattern, simultaneously return a type-setter function that `AltP` can call to propagate the supertype down into the nodes comprising it.

TODO:
- [ ] literals
- [ ] non-sum types (e.g. `(_,_) : (Nat,Int) or (_,_) : (Int,Nat)`)

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -1824,17 +1824,20 @@ and inconsistent t ts =
 (* Patterns *)
 
 and infer_pat_exhaustive warnOrError env pat : T.typ * Scope.val_env =
-  let t, ve = infer_pat env pat in
+  let t, ve, _ = infer_pat env pat in
   if not env.pre then
     coverage_pat warnOrError env pat t;
   t, ve
 
-and infer_pat env pat : T.typ * Scope.val_env =
+and infer_pat env pat : T.typ * Scope.val_env * (T.typ -> unit) =
   assert (pat.note = T.Pre);
   let t, ve = infer_pat' env pat in
   if not env.pre then
     pat.note <- T.normalize t;
-  t, ve
+  let widen = match pat.it with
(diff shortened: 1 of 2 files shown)


crusso:
I'll try it on the counter-example:
```
 func bad((((x,y) : (Nat,Int)) or ((y,x) : (Int,Nat)))): Nat { x }
```
 and report back.

crusso:
Ok, since this branch doesn't support bindings in patterns, I tried it on the simpler:
```
[nix-shell:~/motoko/src]$ moc
Motoko compiler (source 0.9.1-21-gef4204c76)
> func((((_,_) : (Nat,Int)) or ((_,_) : (Int,Nat))) ): Nat { 0 };
stdin:1.30-1.49: warning [M0146], this pattern is never matched
func : ((Int, Int)) -> Nat
> 
```
This is an abstraction of the bad example above and suggests this is still unsound (because it let's you convert a negative `Int` to a `Nat` once you add pattern bindings).

ggreif:
> Ok, since this branch doesn't support bindings in patterns, I tried it on the simpler:
> 
> ```
> [nix-shell:~/motoko/src]$ moc
> Motoko compiler (source 0.9.1-21-gef4204c76)
> > func((((_,_) : (Nat,Int)) or ((_,_) : (Int,Nat))) ): Nat { 0 };
> stdin:1.30-1.49: warning [M0146], this pattern is never matched
> func : ((Int, Int)) -> Nat
> > 
> ```
> 
> This is an abstraction of the bad example above and suggests this is still unsound (because it let's you convert a negative `Int` to a `Nat` once you add pattern bindings).

Because I only support variant types for "backpatching" right now on that branch. Can you try substituting `Int` with `{#a;#b}` and `Nat` with `{#a}` and rerunning your experiment?

luc-blaeser:

PR #4036, opened 2023-06-12
feat: add fields `sender_canister_version` for canister version tracking

ggreif:


src/prelude/internals.mo
```diff
@@ -443,15 +447,17 @@ func @install_actor_helper(
...
-    await @ic00.create_canister({settings = null});
-  await @ic00.install_code({
+    await @ic00.create_canister { settings = null; sender_canister_version };
+  await @ic00.install_code {
     mode = #install;
     canister_id;
     wasm_module;
     arg;
-  });
+    sender_canister_version
```
ggreif, line 459: is this allowed to reuse? In line 438 reusing leads to:
```
ingress Completed: Reject: IC0504: Management canister call payload includes sender canister version 2 that does not match the actual sender canister version 3.
```

test/run-drun/ok/actor-class-mgmt.drun-run.ok
```diff
@@ -1,3 +1,3 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
-ingress Completed: Reply: 0x4449444c0000
+ingress Completed: Reject: IC0302: Management canister has no method 'canister_info'
```
ggreif, line 3: Our `drun` doesn't support this call yet. #3992 is having problems. But we'll be able to test this in the Playground.

src/prelude/internals.mo
```diff
@@ -427,12 +430,13 @@ func @install_actor_helper(
...
     };
-  await @ic00.install_code({
+  await @ic00.install_code {
     mode;
     canister_id;
     wasm_module;
-    arg
-  });
+    arg;
+    sender_canister_version = ?(prim "canister_version" : () -> Nat64)()
```
crusso, line 438: Do we need to recompute canister_version coz it could be different after the first await?

crusso (commented):
Looks fine but maybe let's wait until drun let's us test it before merging...

src/prelude/internals.mo
ggreif: Yes it will, see my other comment.

test/run-drun/ok/actor-class-mgmt.ic-ref-run.ok
```diff
@@ -3,4 +3,4 @@
 => update install_code(record {arg = blob ""; kca_xin = blob "\00asm\01\00\00\00\0...
 <= replied: ()
 => update go()
-<= replied: ()
+<= rejected (RC_CANISTER_REJECT): Unsupported management function canister_info
```
ggreif, line 6: `ic-ref` cannot handle `canister_info` yet.

test/run-drun/actor-class-mgmt.mo
```diff
@@ -3,6 +3,34 @@ import Cycles = "cycles/cycles";
...
+          canister_version : ?Nat64;
+      };
+  };
+
+  type Change_details = {
+      #creation : { controllers : [Principal] };
+      #code_uninstall;
+      #code_deployment : {
+          mode : { #install; #reinstall; #upgrade};
+          module_hash : Blob;
```
ggreif, line 21: suppress this?

test/run-drun/actor-class-mgmt.mo
```diff
@@ -14,6 +42,16 @@ actor a {
...
        }
      } -> async { canister_id : Principal };
+
+      canister_info : {
+          canister_id : Principal;
+          num_requested_changes : ?Nat64;
+      } -> async {
+          total_num_changes : Nat64;
+          recent_changes : [Change];
+          module_hash : ?Blob;
```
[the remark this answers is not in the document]
ggreif: suppressed

test/run-drun/actor-class-mgmt.mo
ggreif: suppressed

src/prelude/internals.mo
```diff
@@ -412,9 +414,10 @@ func @install_actor_helper(
       case (#new settings) {
```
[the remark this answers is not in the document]
ggreif: I wouldn't say it is a breaking change, it is just an _observable_ change. We should support the IC features by default, and offer `#*NoTrace` alternatives for those which prefer secrecy.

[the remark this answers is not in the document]
ggreif: `#newIncognito`, etc., or `#new : Bool`? We can decide later, when we implement this. (The latter would break current code forcing people to adapt, but suffers from boolean blindness.)

crusso (approved):

src/prelude/internals.mo
```diff
@@ -387,13 +387,15 @@ module @ManagementCanister = {
 let @ic00 = actor "aaaaa-aa" :
   actor {
     create_canister : {
-      settings : ?@ManagementCanister.canister_settings
+      settings : ?@ManagementCanister.canister_settings;
```
ggreif, line 390: the user supplies this in the extended mechanism.

src/prelude/internals.mo
```diff
@@ -387,13 +387,15 @@ module @ManagementCanister = {
...
+      sender_canister_version : ?Nat64
```
ggreif, line 391: This is like a new (optional) keyword argument to the call, the only thing a user might reasonably want is suppressing that the controller's version gets passed automatically...

test/run-drun/ok/actor-class-mgmt.drun-run.ok
```diff
@@ -1,3 +1,4 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
+debug.print: {controllers = [rwlgt-iiaaa-aaaaa-aaaaa-cai]; recent_changes = [{canister_version = 0; details = #creation({controllers = [rwlgt-iiaaa-aaaaa-aaaaa-cai]}); origin = #from_canister({canister_id = rwlgt-iiaaa-aaaaa-aaaaa-cai; canister_version = null})}, {canister_version = 1; details = #code_deployment({mode = #install}); origin = #from_canister({canister_id = rwlgt-iiaaa-aaaaa-aaaaa-cai; canister_version = ?18})}, {canister_version = 3; details = #code_deployment({mode = #upgrade}); origin = #from_canister({canister_id = rwlgt-iiaaa-aaaaa-aaaaa-cai; canister_version = ?22})}, {canister_version = 5; details = #code_deployment({mode = #reinstall}); origin = #from_canister({canister_id = rwlgt-iiaaa-aaaaa-aaaaa-cai; canister_version = ?26})}]; total_num_changes = 4}
```
ggreif, line 3: `canister_version = null` is a bit fishy. Where is the original creation code?

ggreif: Is it in line 74?
``` Motoko
let c0 = await
        Cs.C (...
```
?

ggreif: I guess it is rather https://github.com/dfinity/motoko/blob/39d64259f998a993d387afce8c2ee0dc0741c9e3/test/run-drun/actor-class-mgmt.mo#LL92C20-L92C20

ggreif: Indeed, I tried
``` diff
$ git diff
diff --git a/test/run-drun/actor-class-mgmt.mo b/test/run-drun/actor-class-mgmt.mo
index b229bcfc6..ea4d6f8fb 100644
--- a/test/run-drun/actor-class-mgmt.mo
+++ b/test/run-drun/actor-class-mgmt.mo
@@ -35,6 +35,7 @@ actor a {
   let ic00 = actor "aaaaa-aa" :
     actor {
       create_canister : {
+          sender_canister_version : ?Nat64;
         settings : ? {
           controllers : ?[Principal];
           compute_allocation: ?Nat;
@@ -89,7 +90,7 @@ actor a {
 
       Cycles.add(2_000_000_000_000);
       let {canister_id = p} = await
-         ic00.create_canister default_settings;
+        ic00.create_canister { settings = null; sender_canister_version = ?(Prim.canisterVersion()) };
       // no need to add cycles
       let c3 = await
         (system Cs.C)(#install p)(3, null);
```
and found that field filled in.

PR #4041, opened 2023-06-14
feat: allow canister imports of service constructors, silently ignoring the service arguments.

### LINKED ISSUE #3990: FR: allow canister import of service constructor by stripping argument
dfx doesn't strip the canister argument when supplying candid files to Motoko. Though one could fix dfx to do this, it's probably simplest to get Motoko to just drop the service argument of any canister import (perhaps with a warning).

https://forum.dfinity.org/t/import-canister-canistername-error/6225/20

crusso:
Allow canister imports of service constructors, silently ignoring the service arguments. 

A hack that fixes #3990 (for some definition of fix) and duplicate #2319.

Really, dfx should be feeding the idl of the instantiated service, not service constructor, to dependent canisters, but until it's fixed to do so (and it hasn't been forever now), this is a reasonable and simple workaround, avoiding the error-prone and kludgy workaround of rewriting candid files described, for instance, here:

https://dfinityorg.notion.site/ckBTC-example-Encode-Hackathon-0aaf6292e3404dabb49df5d1b5abc797#08a7469beaf14d6ba35e8827e363e160

and also used here:

https://github.com/letmejustputthishere/ckbtc-payments

via npm scripting hacks (!).

Also see here for the pain this caused:

https://forum.dfinity.org/t/confusing-type-error-when-crossing-canisters-expression-of-type-mytype-cannot-produce-expected-type-mytype-1/20636

UPDATE: I turned the previous error into a warning to nag us to fix dfx.

UPDATE: Potentially obviated by https://github.com/dfinity/sdk/pull/3138, which teaches dfx to strip the service argument.

UPDATE: @chenyan-dfinity thinks we should still merge for other scenarios (old dfx, new compiler, I guess).


Changelog.md
```diff
@@ -1,5 +1,9 @@
 # Motoko compiler changelog
 
+* motoko (`moc`)
+
+  * Allow canister imports of service constructors, silently ignoring the service arguments to import the instantiated service. Improves DX using dfx. #4041
```
crusso, line 5:
```suggestion
  * Allow canister imports of service constructors, ignoring the service arguments to import the instantiated service instead (with a warning). Improves DX using dfx. (#4041)
```

Changelog.md
```diff
@@ -1,5 +1,9 @@
 # Motoko compiler changelog
 
+* motoko (`moc`)
+
+  * Allow canister imports of service constructors, ignoring the service arguments to import the instantiated service instead (with a warning). Improves DX using dfx. (#4041)
```
ggreif, line 5: I'd mention _Candid_ here.

src/mo_idl/idl_to_mo.ml
```diff
@@ -97,24 +97,34 @@ and check_meth env occs (m: typ_meth) =
...
+    | Some {it=ServT ms; _} ->
+      List.map (check_meth env occs) ms
+    | Some {it=ClassT (ts1, t); at; _} ->
+      (*@HACK: import service constructors as instantiated services *)
+      (*TODO: fix dfx to derive the correct instantiated candid instead *)
+      begin
+        let t' = check_typ' env occs t in
+        match M.normalize t' with
+        | M.Obj (M.Actor, fs) ->
+          Diag.print_messages [Diag.warning_message at "M0163" "import"
```
ggreif, line 110: Maybe obsolete the old error code and create a new one? Aren't they designed to be immutable?

crusso: Ok.

Changelog.md
crusso: Ok

crusso: Done

src/mo_idl/idl_to_mo.ml
crusso: Done

crusso:
I guess I should close this without merging, right @chenyan-dfinity @lwshang?

chenyan-dfinity:
I think it's still worth merging. There are legacy canisters we cannot control. It's a good idea to make this as warning, and urge users to use the new dfx.

crusso:
> I think it's still worth merging. There are legacy canisters we cannot control. It's a good idea to make this as warning, and urge users to use the new dfx.

When would this apply? Perhaps when someone isn't using dfx and has some Rust candid file they need to use? I'll need an approval to merge...

chenyan-dfinity (approved):
> When would this apply? Perhaps when someone isn't using dfx and has some Rust candid file they need to use? I'll need an approval to merge...

Yep, notably NNS canisters.

PR #4042, opened 2023-06-14
bugfix: allow optional terminal semicolons in Candid imports

### LINKED ISSUE #4037: candid: motoko parser and candid spec appear to disagree on require terminal semicolon after service defn
- [ ] add repro of motoko failing to parse actor .did with terminal semi.

crusso:
Fixes #4037

Terminal semicolons seem to be required by the (informal) syntax in the candid spec, allowed (but not emitted) by Rust, and disallowed by Motoko.

Discovered by user editing and reformatting .did code in VSCode (https://forum.dfinity.org/t/confusing-type-error-when-crossing-canisters-expression-of-type-mytype-cannot-produce-expected-type-mytype-1/20636/10).

We now parse optional semicolons, but don't emit them, to avoid breaking older motoko clients. Should we change the spec to allow this @chenyan-dfinity @nomeata.


nomeata:
No opinions here

chenyan-dfinity (approved):
LGTM

chenyan-dfinity:
The spec looks fine?
```
<prog>  ::= <def>;* <actor>;?
```
The last ; is optional.

crusso:
> The spec looks fine?
> 
> ```
> <prog>  ::= <def>;* <actor>;?
> ```
> 
> The last ; is optional.

I read that as the actor def is optional, e.g. for included files.

PR #4043, opened 2023-06-14
chore: perform scheduled builds of `drun` in a sparse artefact environment

ggreif:
Recently imports of `drun` tend to fail building, because the storage is very constrained on Linux builders, and `nix` preloads all necessary binary artefacts for the whole build process. This meant that when the `drun` build started there were already `OCaml` and `ghc` as well as many instances of `llvm`, `clang` and `rustc` present.

This PR detects when `drun` is about to be built, does it in a frugal way, and also cleans up after itself.

If this method turns out to be effective in the middle term, we can adopt it for `rustc` and other bigger chunks too.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,7 +36,9 @@ jobs:
 
     - name: "pre-build `drun`"
       if: startsWith(github.head_ref, 'gabor/pre-warm')
-      run: nix-build --max-jobs 1 -A tidbits.drun
+      run: |
+        nix-build --max-jobs 1 -A tidbits.drun
+        nix-store --gc --print-live
 
     - name: "nix-build"
       run: nix-build-uncached --max-jobs 4 -A all-systems-go -build-flags -L
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,8 +35,8 @@ jobs:
 
     - run: nix-env -iA nix-build-uncached -f nix/
 
-    - name: "pre-puild `drun`"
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -34,6 +34,10 @@ jobs:
 
     - run: nix-env -iA nix-build-uncached -f nix/
 
+    - name: "pre-build `drun`"
+      if: startsWith(github.head_ref, 'update/ic-')
+      run: nix-build --max-jobs 1 -A tidbits.drun
```
ggreif, line 39: maybe collect garbage?

ggreif: yes, let's see if it works...

ggreif: `drun` seems to survive in the `/nix/store`.

default.nix
```diff
@@ -775,6 +775,10 @@ rec {
     sha256 = "sha256-debC8ZpbIjgpEeISCISU0EVySJvf+WsUkUaLuJ526wA=";
   };
 
+  tidbits = with nixpkgs; {
+    drun = drun;
+  };
+
```
nomeata, line 781: Why the extra `tidbits`? We already have plenty of tidbids in the toplevel, e.g.
```
inherit (nixpkgs) wabt wasmtime wasm
```
just add it there.

ggreif: Ah, okay, I didn't see that!

ggreif:
Btw, I did a test (2bd1c254c7ceeb50ec4933a11c436353ee8ae4c3) and it worked well...

nomeata (approved):

PR #4055, opened 2023-06-20
chore: remove unnecessary `ppx` dependencies

ggreif:
Does anybody know why these were needed in first place?

Anyway, if they are now needed for CI, we can remove them, right?


crusso:
No idea either.

ggreif:
This was the commit that introduced those lines: https://github.com/dfinity/motoko/commit/b5573dfceaf963256d244eb1e9f5468415f1678b
(when the release process produces bad binaries, we'll know which commit to revert...)

PR #4058, opened 2023-06-22
chore: exercise in `nix` overlays

ggreif:
This is mostly a (learning) experiment about how to use overlays and scoped package substitutions. Actually it is not that complicated :-)

Here I customise the `logs` package globally to _not_ use `js_of_ocaml`.

But still I think that this can go in, so that we have a "tutorial" for future modifications.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,6 +40,8 @@ jobs:
         nix-build --max-jobs 1 -A drun
         nix-store --gc --print-roots | grep /motoko/
 
+    - run: nix-store -q --tree $(nix-instantiate -A shell)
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -40,6 +40,8 @@ jobs:
         nix-build --max-jobs 1 -A drun
         nix-store --gc --print-roots | grep /motoko/
 
+    - run: nix-store -q --tree $(nix-instantiate -A shell)
```
ggreif, line 43: this is just for debugging, will be reverted

luc-blaeser:

PR #4064, opened 2023-06-23
perf: keep the heap pointer in a dedicated Wasm global

ggreif:
We now optimise the bump pointer allocator by keeping the heap pointer in a Wasm global, so that the Motoko side can easily access it. This opens up the possibility that for fixed-size allocations we can simply examine a few lower bits (comparing to zero) in order to figure out that a page-crossing has occurred. If so, we only need to "commit" the new page by calling `alloc_words 0`. For the Rust side accessor functions to the `HP` are provided.

### How it works

We construct a mask at compile-time to detect page boundary crossings of the bumped pointer:
- for a (byte) increment `0b1xxxxx` (where the `xxxxx` part is arbitrary)
- and `< 0x8000` (so that the bit 15 is clear, when the LSBit is the 0th)
- we treat this as a 16-bit unsigned integer
- we count its leading zeros (will be `>= 1` and `< 16`), this is `N`.
- the 32-bit HP mask the will be `0b000000000000000011..100..0 : u32`, where there are 16 upper zeros, then `N` ones, closing with `16-N` zeros
- the bumped HP is masked with said number, and when zero, then the addition that happened while bumping the HP caused a rippling carry, and the page boundary has been crossed
- in this case we have to make sure that the new page is backed physically, so we call `Heap.alloc 0` (and discard the result).

Some benchmarks give cycle savings of 10% 😵‍💫💪!

_Note_: Currently the change is only active for _classical GC_, the incremental part will be tackled in #4078.

Further benefits:
- the accessor functions for HP (that now live on the `codegen` side) can be inlined by `wasm-opt`
- looks like `wasm-opt` (`-O3`) also specialises `call alloc_words 0; drop` to a very minimal kernel, speeding up the cold path too.

## Potential future optimisations:
- keep HP as always _skewed_, then no more per-alloc `sub 1` is needed — see #4095
- extend the mask to the right by one bit — https://github.com/dfinity/motoko/pull/4064/files/4eb3dfeb7fdc10737605c631edff33b3f0976eff#r1250623654
  this is #4096, and I regard it as failed with non-skewed `HP`


src/linking/linkModule.ml
```diff
@@ -541,7 +541,8 @@ let set_table_size new_size : module_' -> module_' = fun m ->
     }
   | _ -> raise (LinkError "Expect one table in first module")
 
-let fill_memory_base_import new_base : module_' -> module_' = fun m ->
+
+let fill_item_import modul item new_base (m : module_') : module_' =
```
ggreif, line 545:
```suggestion
let fill_item_import module_name item_name new_base (m : module_') : module_' =
```

src/linking/linkModule.ml
```diff
@@ -817,7 +797,7 @@ let link (em1 : extended_module) libname (em2 : extended_module) =
 
   (* Link globals *)
   let global_required1 = find_imports is_global_import libname em1.module_ in
-  let global_required2 = find_imports is_global_import "env" dm2 in
+  let global_required2 = find_imports is_global_import "env" dm2 @ find_imports is_global_import "GOT.mem" dm2 in
```
ggreif, line 800: revert
```suggestion
  let global_required2 = find_imports is_global_import "env" dm2  in
```

src/codegen/compile.ml
```diff
@@ -1606,9 +1615,28 @@ module Tagged = struct
...
+      (if size < 64l then
+        Heap.(
+        GC.get_heap_pointer env ^^
+        set_object ^^ (* speculate *)
+        GC.get_heap_pointer env ^^
+        compile_add_const Int32.(mul size word_size) ^^ (* FIXME: forwarding? *)
+        GC.set_heap_pointer env ^^
+        GC.get_heap_pointer env ^^
+          (* mask FIXME use logarithm *)
+        compile_bitand_const 0xFF00l ^^
```
luc-blaeser, line 1628: Not sure: Shouldn't this be `0xFFFF_FF00`?

src/codegen/compile.ml
```diff
@@ -1052,8 +1052,11 @@ module GC = struct
     E.call_import env "ic0" "performance_counter"
 
   let register_globals env =
-    (E.add_global64 env "__mutator_instructions" Mutable 0L;
-     E.add_global64 env "__collector_instructions" Mutable 0L)
+    begin
```
ggreif, line 1055: probably not even needed!

src/linking/linkModule.ml
ggreif, line 545: this is just deduplication

rts/motoko-rts/src/memory/ic/linear_memory.rs
```diff
@@ -39,7 +42,7 @@ impl Memory for IcMemory {
         let delta = u64::from(bytes.as_u32());
 
         // Update heap pointer
-        let old_hp = u64::from(HP);
+        let old_hp = u64::from(getHP());
```
ggreif, line 45: It would be interesting to see how many cycles could be saved if we migrate this function back to the `compile.ml` side.

ggreif: I think a better way is to inline the `getHP` and `setHP` functions at link time into the code directly.

ggreif: I'll revert my experiment shortly.

src/codegen/compile.ml
```diff
@@ -1052,8 +1052,11 @@ module GC = struct
...
+      E.add_global64 env "__mutator_instructions" Mutable 0L;
+      E.add_global64 env "__collector_instructions" Mutable 0L;
+      E.add_global32 env "HP" Mutable 0l;
```
ggreif, line 1058:
```suggestion
      E.add_global32 env "__HP" Mutable 0l;
```

src/linking/linkModule.ml
ggreif: Moved out into #4083.

rts/motoko-rts/src/memory/ic/linear_memory.rs
```diff
@@ -6,15 +6,18 @@ use crate::types::*;
 /// Amount of garbage collected so far.
 pub(crate) static mut RECLAIMED: Bytes<u64> = Bytes(0);
 
-/// Heap pointer
-pub(crate) static mut HP: u32 = 0;
+// Heap pointer
+extern "C" {
```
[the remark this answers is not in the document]
ggreif: Yes, see #4078.

src/codegen/compile.ml
ggreif: ba239fade

src/codegen/compile.ml
```diff
@@ -1602,12 +1608,29 @@ module Tagged = struct
...
+         compile_sub_const 1l ^^ (* skew *)
+         GC.get_heap_pointer env ^^
+         let size_in_bytes = Int32.(mul size Heap.word_size) in
+         compile_add_const size_in_bytes ^^
+         GC.set_heap_pointer env ^^
+         GC.get_heap_pointer env ^^
+         compile_bitand_const (overflow_mask size_in_bytes) ^^
+         G.if0
+           G.nop (* no page crossing *)
+           (Heap.alloc env 0l ^^ G.i Drop) (* ensure that the page is allocated *)
```
crusso, line 1631: What is this crosses multiple pages, will the intermediate pages also be allocated? If size > 65535 (in bytes) I think you could cross multiple pages

ggreif: It would commit many pages, but it can't due to the size check above (`< 0x2000` words)

test/bench/ok/alloc.drun-run.ok
```diff
@@ -1,8 +1,8 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
-debug.print: (+268_435_456, 2_818_724_162)
+debug.print: (+268_435_456, 2_533_732_672)
```
ggreif, line 3:
```
$ expr 2533732672000 / 2818724162 - 1000
-102
```
This is a 10% improvement! 🥳

test/bench/ok/palindrome.drun-run.ok
```diff
@@ -1,9 +1,9 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
-debug.print: (true, +1_188, 13_165)
-debug.print: (false, +1_188, 12_450)
-debug.print: (false, +1_188, 13_155)
-debug.print: (true, +868, 12_841)
-debug.print: (false, +868, 11_660)
-debug.print: (false, +868, 12_800)
+debug.print: (true, +1_188, 11_988)
```
ggreif, line 3:
```
$ expr 11988000 / 13165 - 1000
-90
```
this is a 9% improvement!

src/codegen/compile.ml
crusso: Right you are. I was reading my hex wrong. That's 32768 bytes.

ggreif: 3f301d74a performs the comparison in bytes, this will work for 8-byte words too.

src/codegen/compile.ml
```diff
@@ -5400,12 +5444,12 @@ module MakeSerialization (Strm : Stream) = struct
 
   module Registers = struct
     let register_globals env =
-     (E.add_global32 env "@@rel_buf_opt" Mutable 0l;
+      E.add_global32 env "@@rel_buf_opt" Mutable 0l;
```
ggreif, line 5447: cosmetics

src/codegen/compile.ml
ggreif: See also https://github.com/dfinity/motoko/pull/4064#issuecomment-1607963648

src/codegen/compile.ml
```diff
@@ -1602,12 +1608,30 @@ module Tagged = struct
...
+      Int32.(logand 0xFFFFl (shift_left minus_one (16 - Numerics.Nat16.(to_int (clz (of_int n)))))) in
+
     Func.share_code0 env name [I32Type] (fun env ->
-      let (set_object, get_object) = new_local env "new_object" in
-      Heap.alloc env size ^^
+      let set_object, get_object = new_local env "new_object" in
+      let size_in_bytes = Int32.(mul size Heap.word_size) in
+      (if !Flags.gc_strategy <> Flags.Incremental && size_in_bytes < 0x8000l then
+         GC.get_heap_pointer env ^^
+         compile_sub_const 1l ^^ (* skew *)
```
luc-blaeser, line 1624:
```suggestion
         compile_add_const ptr_skew ^^
```

src/codegen/compile.ml
```diff
@@ -1602,12 +1608,30 @@ module Tagged = struct
...
     1l
 
-  (* Note: Post allocation barrier must be applied after initialization *)
+  (* Note: post-allocation barrier must be applied after initialization *)
   let alloc env size tag =
+    assert (size > 1l);
     let name = Printf.sprintf "alloc_size<%d>_tag<%d>" (Int32.to_int size) (Int32.to_int (int_of_tag tag)) in
+    let overflow_mask n =
+      let n = Int32.to_int n in
+      Int32.(logand 0xFFFFl (shift_left minus_one (16 - Numerics.Nat16.(to_int (clz (of_int n)))))) in
```
luc-blaeser, line 1617:
```suggestion
      let page_mask = Int32.sub page_size 1l in
      Int32.(logand page_mask (shift_left minus_one (16 - Numerics.Nat16.(to_int (clz (of_int n)))))) in
```

src/codegen/compile.ml
```diff
@@ -1602,12 +1608,30 @@ module Tagged = struct
...
+
     Func.share_code0 env name [I32Type] (fun env ->
-      let (set_object, get_object) = new_local env "new_object" in
-      Heap.alloc env size ^^
+      let set_object, get_object = new_local env "new_object" in
+      let size_in_bytes = Int32.(mul size Heap.word_size) in
+      (if !Flags.gc_strategy <> Flags.Incremental && size_in_bytes < 0x8000l then
```
luc-blaeser, line 1622:
```suggestion
      let half_page_size = Int32.div page_size 2l in
      (if !Flags.gc_strategy <> Flags.Incremental && size_in_bytes < half_page_size then
```

src/codegen/compile.ml
```diff
@@ -1602,12 +1608,30 @@ module Tagged = struct
     assert (!Flags.gc_strategy = Flags.Incremental);
     1l
 
-  (* Note: Post allocation barrier must be applied after initialization *)
+  (* Note: post-allocation barrier must be applied after initialization *)
   let alloc env size tag =
+    assert (size > 1l);
     let name = Printf.sprintf "alloc_size<%d>_tag<%d>" (Int32.to_int size) (Int32.to_int (int_of_tag tag)) in
+    let overflow_mask n =
```
luc-blaeser, line 1615: Maybe it could be helpful to add a comment on how the page crossing logic works. The PR description is very nice.
As an intuition of that logic (if I am not mistaken), it is checking whether the new HP resides at the beginning of the page, i.e. the new address has leading zeros inside the page.

src/codegen/compile.ml
```diff
@@ -1602,12 +1608,30 @@ module Tagged = struct
...
+         GC.get_heap_pointer env ^^
+         compile_sub_const 1l ^^ (* skew *)
+         GC.get_heap_pointer env ^^
+         compile_add_const size_in_bytes ^^
+         GC.set_heap_pointer env ^^
+         GC.get_heap_pointer env ^^
+         compile_bitand_const (overflow_mask size_in_bytes) ^^
+         G.if0
+           G.nop (* no page crossing *)
+           (Heap.alloc env 0l ^^ G.i Drop) (* ensure that the page is allocated *)
```
luc-blaeser, line 1632: Would it be possible to economize this `0` allocation by always having the current page allocated?

src/codegen/compile.ml
```diff
@@ -1602,12 +1608,30 @@ module Tagged = struct
...
+       else
+         Heap.alloc env size) ^^
```
luc-blaeser, line 1634: If I am not mistaken, this case may also happen if the page is not necessarily crossed, e.g. for the first few small allocations in the page, e.g. for an allocation series of 32 bytes, 64 bytes, 128 bytes? Then, the leading zeros of the bump pointer inside the page remains 0?

src/codegen/compile.ml
```diff
@@ -1052,8 +1052,9 @@ module GC = struct
     E.call_import env "ic0" "performance_counter"
 
   let register_globals env =
-    (E.add_global64 env "__mutator_instructions" Mutable 0L;
-     E.add_global64 env "__collector_instructions" Mutable 0L)
+    E.add_global64 env "__mutator_instructions" Mutable 0L;
+    E.add_global64 env "__collector_instructions" Mutable 0L;
+    E.add_global32 env "_HP" Mutable 0l
```
luc-blaeser, line 1057: Maybe we could add a safe guard that this is not used for the incremental GC (as long as this is not supported)?
```suggestion
    if !Flags.gc_strategy <> Flags.Incremental then
      E.add_global32 env "_HP" Mutable 0l
    else
      ()
```

src/codegen/compile.ml
```diff
@@ -1065,6 +1066,11 @@ module GC = struct
   let set_collector_instructions env =
     G.i (GlobalSet (nr (E.get_global env "__collector_instructions")))
 
+  let get_heap_pointer env =
+    G.i (GlobalGet (nr (E.get_global env "_HP")))
+  let set_heap_pointer env =
+    G.i (GlobalSet (nr (E.get_global env "_HP")))
```
luc-blaeser, line 1072:
```suggestion
    if !Flags.gc_strategy <> Flags.Incremental then
      G.i (GlobalSet (nr (E.get_global env "_HP")))
    else
      assert false
```

src/codegen/compile.ml
```diff
@@ -5202,6 +5226,27 @@ module RTS_Exports = struct
...
+    let get_hp_fi = E.add_fun env "__get_hp" (
+      Func.of_body env [] [I32Type] (fun env ->
+        GC.get_heap_pointer env
+      )
+    ) in
+    E.add_export env (nr {
+      name = Lib.Utf8.decode "getHP";
+      edesc = nr (FuncExport (nr get_hp_fi))
+    });
+
```
luc-blaeser, line 5249:
```suggestion
    if !Flags.gc_strategy <> Flags.Incremental then 
      let set_hp_fi = 
        E.add_fun env "__set_hp" (
        Func.of_body env ["new_hp", I32Type] [] (fun env ->
          G.i (LocalGet (nr 0l)) ^^
          GC.set_heap_pointer env
        )
      ) in
      E.add_export env (nr {
        name = Lib.Utf8.decode "setHP";
        edesc = nr (FuncExport (nr set_hp_fi))
      });

      let get_hp_fi = E.add_fun env "__get_hp" (
        Func.of_body env [] [I32Type] (fun env ->
          GC.get_heap_pointer env
        )
      ) in
      E.add_export env (nr {
        name = Lib.Utf8.decode "getHP";
        edesc = nr (FuncExport (nr get_hp_fi))
      });
    else
      ();
```

src/codegen/compile.ml
ggreif: This line is for sizes that are too big...
Assume HP is at the page boundary (`0xPPPP0000`), then an 8-byte alloc would give `0xPPPP0008` which gives a false positive for the current mask (entering the slow path), but would be on the fast path with an extended mask. I considered this optimisation, but never proved it correct. I'll put this on the potential refinements list, as I don't want to change this PR any more.

src/codegen/compile.ml
ggreif: Maybe, but is there a good/better place to do that?

luc-blaeser: I am not sure. Could we allocate the first page in the dynamic heap (in the case the start address would be page-aligned) and then allocate whenever a page is crossed? But maybe it is not worth the complexity...

luc-blaeser: Ok, sorry for the confusion, now I see that the case is only happening when page is crossed (with some false positives).

src/codegen/compile.ml
luc-blaeser: Thanks for the explanation.

src/codegen/compile.ml
ggreif: added 80188facb

src/codegen/compile.ml
ggreif: Okay, let's assume we can extend the mask. Increment
- `0b0000000001111100` (=`0x4c` = 76), augmented mask
- `0b11111111111111000000` (= `0xffc0`)
now suppose the old `HP` is `0x3333fffc`, then the post-bump `HP` is
`0x33340048`. Masking gives `0x0040` saying that no crossing happened, which is clearly wrong.
We can only do the mask extension when the bit after the topmost `1` in the increment is clear!

src/codegen/compile.ml
```diff
@@ -1602,12 +1615,34 @@ module Tagged = struct
...
+         GC.get_heap_pointer env ^^
+         compile_add_const ptr_skew ^^
+         GC.get_heap_pointer env ^^
+         compile_add_const size_in_bytes ^^
+         GC.set_heap_pointer env ^^
+         GC.get_heap_pointer env ^^
+         compile_bitand_const (overflow_mask size_in_bytes) ^^
+         G.if0
+           G.nop (* no page crossing *)
+           (Heap.alloc env 0l ^^ G.i Drop) (* ensure that the page is allocated *)
```
nomeata, line 1643: Could this be more declaratively `Heap.ensure_allocated`? The magic 0 and the dropping of the result are a bit distracting noise for future readers of the code.

ggreif: see 986d1c63d

src/codegen/compile.ml
luc-blaeser: Yes, the bit mask cannot be simply extended.
I assume the version with skewed pointer is working because we use skew = -1. 
If we would encode skew = +1, then it would not work?

ggreif: The mask extension is only possible for `popcnt N == 1` and also works for skewed `HP`s!

luc-blaeser (approved):
Very nice performance improvement.
(I will need to adjust the incremental GC later for this optimization.)

crusso:
Ok, I'm late to the review but think I understand it now. Very clever.

One thing, compile.ml always calls rts.alloc_array/blob/stream for those, so will, in particular, miss the fast path optimization for arrays/tuples  and text/blob, I think. There might be some more wins there.

PR #3009, opened 2021-12-16
Gabor/memcpy

ggreif:
Relevant: https://stackoverflow.com/questions/67068195/does-clang-provide-intrinsics-for-webassemblys-memory-fill-and-memory-copy

PR #4074, opened 2023-06-26
Support Wasm Bulk Memory Operations

luc-blaeser:
Use [WebAssembly bulk memory operations](https://github.com/WebAssembly/bulk-memory-operations/blob/master/proposals/bulk-memory-operations/Overview.md) in the runtime system and compiler.

**Performance Improvement**:
Although Wasm bulk memory operations are mainly beneficial for the IC execution layer, one may notice a performance improvement for the copying GC at the IC cost model.

Runtime measurements, GC benchmark, total Wasm instructions, `dfx 0.14.3`, with latest bump allocator optimization (only applied to non-incremental GCs):

| Configuration | Before | With bulk-memory op |  Difference
| -----|------|----------|------|
| Copying GC | 2.02e10 | 1.92e10 | -5% |
| Incremental GC | 1.85e10 | 1.85e10 | 0 |
| Generational GC | 1.86e10 | 1.86e10 | 0 |
| Compacting GC | 2.16e10 | 2.16e10 | 0 |

**Fallback**: `ic-ref` and `ic-wasm` do not support Wasm bulk memory operations and thus engage classical `libc` `memcpy` with ordinary memory access instructions.

**Design**:
- Bulk memory operations are implemented in the compiler backend. For all platforms, except for `ic-ref` and `ic-wasm`, the compiler generates Wasm bulk-memory operations.
- The runtime system uses these compiler-generated memory operations.
- If Wasm bulk memory operations are not supported (in the case of `ic-ref` or `ic-wasm`), the compiler invokes the legacy operations exposed by the RTS. For this reason, the RTS is **not** compiled with Wasm bulk memory flag, such that it still integrates the legacy `libc` implementations that do not rely on Wasm bulk memory support.


ggreif:
See also https://github.com/dfinity/motoko/pull/3009. But that needs quite some yak-shaving before it can be made working.

luc-blaeser:
> See also #3009. But that needs quite some yak-shaving before it can be made working.

Thanks for pointing this out. I will look at this PR and update.

src/codegen/compile.ml
```diff
@@ -1181,14 +1181,14 @@ module Heap = struct
   (* Convenience functions related to memory *)
   (* Copying bytes (works on unskewed memory addresses) *)
   let memcpy env = 
-    if E.mode env = Flags.ICMode then
+    if E.mode env = Flags.RefMode then
```
ggreif, line 1184: This is clever, I don't see a short-term fix in `winter`.

ggreif:
Looks like you need some `--bulk-memory` flags in the `test/run.sh` script.

luc-blaeser:
Inactive for a while. We can reopen if needed...

PR #4087, opened 2023-06-29
GC Adjustments for Stable Regions

luc-blaeser:
Some GC- and memory-specific adjustments for stable memory regions:
* Using the forwarding pointer of the incremental GC in the compiler and runtime system.
* Checking whether `region0` is defined before access (since dereferencing address 0 would not trap in release mode).
* Initializing the 16-bit `padding` field in the region as it is read by the compiler-generated code.
* Applying write barriers when assigning `Region.vec_pages`.

### DIFF
--- a/rts/motoko-rts/src/region0.rs
+++ b/rts/motoko-rts/src/region0.rs
@@ -1,23 +1,24 @@
 //use crate::region::Region;
 use crate::memory::Memory;
 use crate::region::{region_grow, region_size};
-use crate::types::{Blob, Value};
+use crate::types::Value;
+use core::ptr::null_mut;
 
 use motoko_rts_macros::ic_mem_fn;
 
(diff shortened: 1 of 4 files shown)


rts/motoko-rts/src/region0.rs
```diff
@@ -108,7 +111,7 @@ pub unsafe fn region0_store_float64<M: Memory>(mem: &mut M, offset: u64, val: f6
 
 #[ic_mem_fn]
 pub unsafe fn region0_store_blob<M: Memory>(mem: &mut M, offset: u64, blob: Value) {
-    let blob: *const Blob = blob.as_blob();
+    let blob = blob.as_blob();
```
crusso, line 114: Should one resolve forwarding on the blob before readings its contents?

matthewhammer:
CI passed. Ready to merge?

PR #4093, opened 2023-07-02
Gabor/specials

ggreif:


ggreif:
I performed a few division tests, but can only find few consecutive remainder sequences. I am now pondering multiplying all hashes with a prime _p_ and then looking for a divisor. This theorem gives me hope: https://twitter.com/AlgebraFact/status/1677419576177352704?s=20

rossberg:
Better optimisations for patten matching are good ideas. There are more sophisticated methods for building decision trees for nested patterns as well. Of course, with all these optimisations, it is not immediately clear how big a gain they are in practice and hence how much bang-for-the-bucks they provide.

Re the use of br_table, could it work to modulo/xor the hash with a small number first (some single-digit multiple of the number of cases) and br_table on the result of that, and then resolve cases with collisions in respective branches by secondary comparisons?

PS: Oh, and would it be hard to use more meaningful issue/PR titles? ;)

ggreif:
> Better optimisations for pattern matching are good ideas. There are more sophisticated methods for building decision trees for nested patterns as well. Of course, with all these optimisations, it is not immediately clear how big a gain they are in practice and hence how much bang-for-the-bucks they provide.

We have a `dao` micro-benchmark in the making and I'll check that in before these hit `master`.

> 
> Re the use of `br_table`, could it work to modulo/xor the hash with a small number first (some single-digit multiple of the number of cases) and br_table on the result of that, and then resolve cases with collisions in respective branches by secondary comparisons?

Absolutely, I've been thinking of masking out a small part of the hash, that has sufficient diversity of bits for a subsequent division. (Masking should preferably be a `shlU` and division scaled up by a power-of-2.)
> 
> PS: Oh, and would it be hard to use more meaningful issue/PR titles? ;)

ggreif:
@ulan do you have ideas/references about such optimisations?

src/codegen/compile.ml
```diff
@@ -10376,16 +10417,37 @@ and compile_exp_with_hint (env : E.t) ae sr_hint exp =
...
+    final_sr,
+    (* Run rest in block to exit from *)
+    FakeMultiVal.block_ env (StackRep.to_block_type env final_sr) (fun branch_code ->
+       orsPatternFailure env (List.map (fun (sr, c) ->
+          c ^^^ CannotFail (StackRep.adjust env sr final_sr ^^ branch_code)
+       ) [sr, CannotFail code1 ^^^ pat_code ^^^ CannotFail rhs_code]) ^^
+       G.i Unreachable (* We should always exit using the branch_code *)
+    )
+
+  | SwitchE (e, cs) -> ascending_remainders env e;
```
ggreif, line 10442:
```suggestion
  | SwitchE (e, cs) ->
```

src/codegen/compile.ml
```diff
@@ -10312,6 +10312,47 @@ and compile_exp_as env ae sr_out e =
...
+  (* for a 2-cased variant type, the second second comparison can be omitted when the first pattern
+     (with irrefutable subpattern) didn't match *)
+  | [{it={pat={it=TagP (_, ip);_}; _}; _} as c1; {it={pat={it=TagP (_, pat');_} as pat2; exp}; _} as c2], Type.(Variant [_; _])
+       when Ir_utils.is_irrefutable ip ->
+     [c1; {c2 with it = {exp; pat = {pat2 with it=TagP ("", pat')}}}]
+  | _ -> cs
+
+
+
+and ascending_remainders env e =
```
ggreif, line 10331:
```suggestion
and _ascending_remainders env e =
```

src/codegen/compile.ml
```diff
@@ -10312,6 +10312,47 @@ and compile_exp_as env ae sr_out e =
...
+  (* for a 2-cased variant type, the second second comparison can be omitted when the first pattern
+     (with irrefutable subpattern) didn't match *)
+  | [{it={pat={it=TagP (_, ip);_}; _}; _} as c1; {it={pat={it=TagP (_, pat');_} as pat2; exp}; _} as c2], Type.(Variant [_; _])
+       when Ir_utils.is_irrefutable ip ->
+     [c1; {c2 with it = {exp; pat = {pat2 with it=TagP ("", pat')}}}]
+  | _ -> cs
+
+
+
+and _ascending_remainders env e =
```
ggreif, line 10331: Please disregard this function, I'll remove (will hopefully come back in a more modular fashion in a follow-up PR).

src/codegen/compile.ml
```diff
@@ -10376,16 +10417,37 @@ and compile_exp_with_hint (env : E.t) ae sr_hint exp =
     )
     ^^
    G.i Unreachable
+
+  | SwitchE (e, cs) when single_case e cs ->
+    let code1 = compile_exp_vanilla env ae e in
+    let [@warning "-8"] [{it={pat={it=TagP (_, pat');_} as pat; exp}; _}] = cs in
+    let ae1, pat_code = compile_pat_local env ae {pat with it=TagP ("", pat')} in
```
ggreif, line 10424: Make a descriptively-named helper for `TagP ("", p)`, e.g. `known_tag_pat`.

src/codegen/compile.ml
```diff
@@ -10376,16 +10417,37 @@ and compile_exp_with_hint (env : E.t) ae sr_hint exp =
...
+       orsPatternFailure env (List.map (fun (sr, c) ->
+          c ^^^ CannotFail (StackRep.adjust env sr final_sr ^^ branch_code)
+       ) [sr, CannotFail code1 ^^^ pat_code ^^^ CannotFail rhs_code]) ^^
+       G.i Unreachable (* We should always exit using the branch_code *)
+    )
+
   | SwitchE (e, cs) ->
     let code1 = compile_exp_vanilla env ae e in
     let (set_i, get_i) = new_local env "switch_in" in
-
```
ggreif, line 10382: put this back

test/perf/dao.mo
```diff
@@ -274,4 +274,5 @@ shared({caller = creator}) actor class DAO() = Self {
```
ggreif, line 274: Please ignore

test/perf/ok/dao.drun-run.ok
```diff
@@ -13,4 +13,5 @@ Ok: Reply: 0x4449444c086e016c08dbb7017db793e6d10307abfec2a4060591ecada00804d6a9b
 Ok: Reply: 0x4449444c086e016c08dbb7017db793e6d10307abfec2a4060591ecada00804d6a9bbae0a7cb4bfd4960b68f18d9dc10b078effd6e90e026c03e1edeb4a71b3c4b1f20468c7ebc4d009036d7b6b06ca87d0cd047ffefe8fdc0a7fb2d5bc880b7fe7fadec00c7fbd8bfbbf0e71e1aca5d90f7f6e066c02006801056c01b9ef9380087d010000
```
ggreif, line 13: Please ignore

crusso:
Might be worth looking at how ocaml deals with polymorphic variants. Perhaps they have a good trick.

nomeata:
Is there no compile time algorithm that takes _n_ numbers and returns a small number of numeric operations that map an input to 0..n-1 if it is the corresponding number (and arbitrary otherwise)? But probably nothing easy enough, else other languages with hashing tags would use it already, right?

src/codegen/compile.ml
ggreif: Done.

nomeata:
I can write such a function in general, but it would be a polynomial of degree n, do hardly more efficient than n comparisons :-)

nomeata:
Minor thought: About avoiding irrefutable tag checks: Instead of making the compile.ml code more complicated, we could introduce a `pattern failed` prim, and have an IR-to-IR pass that elaborates pattern matches to be complete (by inserting a `_` match that fails where needed). Then the backend can trust that pass and omit tag checks in the last clause.

ggreif:
> I can write such a function in general, but it would be a polynomial of degree n, do hardly more efficient than n comparisons :-)

Good old Lagrange would rotate in his grave 😂

ggreif:
I have discussed the the `UncheckedTagP` idea with Kento yesterday. I am shying away from it (for now) because it seems overkill for such a small PR. As this idea grows we could move the machinery to IR, indeed. There is already one such PR (#3955, but that got unfavourable comments, so it is on hold). There are many more transformations that can optimise case lists with variants.

src/codegen/compile.ml
```diff
@@ -10315,6 +10315,22 @@ and compile_exp_as env ae sr_out e =
...
+  | [{it={pat={it=TagP (l, _);_}; _}; _}], Type.(Variant [{lab; _}]) -> l = lab
+  | _ -> false
+
+and known_tag_pat p = TagP ("", p)
+
+and simplify_cases e (cs : Ir.case list) =
+  match cs, e.note.Note.typ with
+  (* for a 2-cased variant type, the second second comparison can be omitted when the first pattern
+     (with irrefutable subpattern) didn't match *)
+  | [{it={pat={it=TagP (_, ip);_}; _}; _} as c1; {it={pat={it=TagP (_, pat');_} as pat2; exp}; _} as c2], Type.(Variant [_; _])
```
crusso, line 10329: This isn't quite enough - I think you need to check the pattern tags are in the variant. See variant-specials-bug.

crusso (changes requested):

src/codegen/compile.ml
crusso: Subtyping is a bitch.

src/codegen/compile.ml
```diff
@@ -10315,6 +10315,23 @@ and compile_exp_as env ae sr_out e =
...
+
+and known_tag_pat p = TagP ("", p)
+
+and simplify_cases e (cs : Ir.case list) =
+  match cs, e.note.Note.typ with
+  (* for a 2-cased variant type, the second comparison can be omitted when the first pattern
+     (with irrefutable subpattern) didn't match, and the patternn types line up *)
+  | [{it={pat={it=TagP (_, ip); note=tp1; _}; _}; _} as c1; {it={pat={it=TagP (_, pat'); note=tp2; _} as pat2; exp}; _} as c2], (Type.(Variant [_; _]) as te)
+       when Ir_utils.is_irrefutable ip
+            && Type.sub tp1 te && Type.sub tp2 te ->
```
ggreif, line 10331: @crusso either I got this backwards, or there is a bug in the `sub` predicate, as these didn't change a bit 😥

crusso: Just check both labels are in the type te.

test/run/variant-specials-bug.mo
```diff
@@ -0,0 +1,13 @@
+func specials(two : { #c0; #c1 }) {
+  switch two {
+    case (#c2) ();           // note #c2 <> #c1
+    case (#c2) assert false; // note #c2 <> #c1
+  };
```
ggreif, line 5: now that this traps (`pattern failed`), it won't get to the latter switch. I'll split the test.

src/codegen/compile.ml
```diff
@@ -10315,6 +10315,24 @@ and compile_exp_as env ae sr_out e =
...
+
+and simplify_cases e (cs : Ir.case list) =
+  match cs, e.note.Note.typ with
+  (* for a 2-cased variant type, the second comparison can be omitted when the first pattern
+     (with irrefutable subpattern) didn't match, and the pattern types line up *)
+  | [{it={pat={it=TagP (l1, ip); _}; _}; _} as c1; {it={pat={it=TagP (l2, pat'); _} as pat2; exp}; _} as c2], Type.(Variant [{lab=el1; _}; {lab=el2; _}])
+       when Ir_utils.is_irrefutable ip
+            && (l1 = el1 || l1 = el2)
+            && (l2 = el1 || l2 = el2) ->
+     [c1; {c2 with it = {exp; pat = {pat2 with it = known_tag_pat pat'}}}]
```
crusso, line 10333: I guess you can even drop c2 when l1 = l2. But maybe don't go there.

ggreif: I won't!

src/codegen/compile.ml
```diff
@@ -10315,6 +10315,24 @@ and compile_exp_as env ae sr_out e =
   let sr_in, code = compile_exp_with_hint env ae (Some sr_out) e in
   code ^^ StackRep.adjust env sr_in sr_out
 
+and single_case e (cs : Ir.case list) =
```
crusso, line 10318: Why not just inline the pattern below and test against the type on the pat, not e?

src/codegen/compile.ml
ggreif: did that!

src/codegen/compile.ml
```diff
@@ -10379,6 +10397,28 @@ and compile_exp_with_hint (env : E.t) ae sr_hint exp =
     )
     ^^
    G.i Unreachable
+
+  | SwitchE (e, cs) when single_case e cs ->
+    let code1 = compile_exp_vanilla env ae e in
+    let [@warning "-8"] [{it={pat={it=TagP (_, pat');_} as pat; exp}; _}] = cs in
```
crusso, line 10403:
```suggestion
  | SwitchE (e,  [{it={pat={it=TagP (l1, pat'); note = T.Variant [(l1, _)]}; exp }; _} ]  when l1 = l2 ->
    let code1 = compile_exp_vanilla env ae e in
```

src/codegen/compile.ml
ggreif: I also learned to mistrust the type note on patterns...

src/codegen/compile.ml
ggreif: Dunno. I trust `e`'s type, but not `pat`'s.

src/codegen/compile.ml
ggreif: The first match's tag shouldn't be assumed to be in the scrutinee's type either! I made two test cases, both work as expected now.

src/codegen/compile.ml
crusso: We do check them, so I think you can trust them. Indeed, it was the checker that caught the bad annoations introduced by pattern inference. But ok, I agree the pattern match is a bit inscrutable.

PR #4095, opened 2023-07-03
perf: keep a skewed pointer in `HP`

ggreif:
Keeping the `HP` as _always skewed_ allows us to eliminate the subtraction on each individual allocation, at the expense of (un)skewing in a few seldom operations (which possibly can also be transitioned to skewed pointers sometime). Removing an instruction from an already compact _hot path_ has profound effects, as already has been demonstrated by #4064. This PR realises one of the suggestions from there.

The runtime savings amount to 1.5% on allocation-heavy benchmarks, and 0.3% on average, which is rather nice.

_Note_: Also included are a few `u32` -> `usize` changes, which may come handy in the 64-bit world.


test/bench/ok/palindrome.drun-run.ok
```diff
@@ -1,9 +1,9 @@
...
-debug.print: (false, +1_188, 11_273)
-debug.print: (false, +1_188, 11_978)
-debug.print: (true, +868, 11_936)
-debug.print: (false, +868, 10_755)
-debug.print: (false, +868, 11_895)
+debug.print: (true, +1_188, 11_854)
+debug.print: (false, +1_188, 11_139)
+debug.print: (false, +1_188, 11_844)
+debug.print: (true, +868, 11_834)
+debug.print: (false, +868, 10_653)
```
ggreif, line 7: This is almost a percent better:
```
$ expr 1065300000  / 10755 - 100000
-949
```

test/bench/ok/alloc.drun-run.ok
```diff
@@ -1,8 +1,8 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
-debug.print: (+268_435_456, 2_533_732_672)
+debug.print: (+268_435_456, 2_500_161_858)
```
ggreif, line 3: This is almost 1.5% better:
```
$ expr 2500161858000 / 2533732672 - 1000
-14
```

test/bench/ok/heap-32.drun-run.ok
```diff
@@ -1,5 +1,5 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
-debug.print: (50_227, +91_377_860, 1_433_012_379)
-debug.print: (50_070, +102_586_000, 1_504_942_528)
+debug.print: (50_227, +91_377_860, 1_434_593_325)
```
ggreif, line 3: This is sadly 0.11% worse:
```
expr 14345933250000 / 1433012379  - 10000
11
```

rts/motoko-rts/src/memory/ic/linear_memory.rs
```diff
@@ -6,18 +6,18 @@ use crate::types::*;
...
     pub(crate) fn getHP() -> u32;
 }
 
 /// Heap pointer after last GC
 pub(crate) static mut LAST_HP: u32 = 0;
 
 pub(crate) unsafe fn initialize() {
-    setHP(get_aligned_heap_base());
-    LAST_HP = getHP();
+    LAST_HP = get_aligned_heap_base();
```
ggreif, line 19: maybe `LAST_HP` can be skewed too?

rts/motoko-rts/src/memory/ic/linear_memory.rs
```diff
@@ -6,18 +6,18 @@ use crate::types::*;
 /// Amount of garbage collected so far.
 pub(crate) static mut RECLAIMED: Bytes<u64> = Bytes(0);
 
-// Heap pointer
+// Heap pointer (skewed)
 extern "C" {
-    pub(crate) fn setHP(new_hp: u32);
+    pub(crate) fn setHP(new_hp: u32); // usize????
```
luc-blaeser, line 11: Maybe it is better to use `Value` if it is a skewed pointer?

rts/motoko-rts/src/gc.rs
```diff
@@ -22,5 +22,5 @@ unsafe fn should_do_gc(max_live: crate::types::Bytes<u64>) -> bool {
         (u64::from(LAST_HP) + max_live.0) / 2,
     );
 
-    u64::from(getHP()) >= heap_limit
+    u64::from(getHP() + 1) >= heap_limit
```
luc-blaeser, line 25: Using `unskew()` or `skew()` from `types.rs` could make this more self-explanatory. Or alternatively, probably even better, `getHP()` could return `Value`.

rts/motoko-rts/src/memory/ic/linear_memory.rs
luc-blaeser: This could make sense for similarity. Maybe the same for the heap base pointer. As for performance, this pointer is accessed only a very few times during the classical GCs...

test/bench/ok/alloc.drun-run.ok
luc-blaeser: I can also measure in the GC benchmark.

rts/motoko-rts/src/memory/ic/linear_memory.rs
ggreif: I left it unskewed but changed it to `usize` (with millions of knock-on changes). Hopefully this will help the 64-bit port sometime.

rts/motoko-rts/src/gc.rs
ggreif: Good point, will do. (But not `Value`.)

ggreif: done in 116d73596

rts/motoko-rts/src/memory/ic/linear_memory.rs
ggreif: This is a local function now.

ggreif: and `setHP` takes an `usize`. 27409cc3b74b517c6051a362a59a99ceb455d0a0

rts/motoko-rts/src/gc/mark_compact.rs
```diff
@@ -57,18 +57,18 @@ unsafe fn compacting_gc<M: Memory>(mem: &mut M) {
...
     GetHp: Fn() -> usize,
-    SetHp: Fn(u32),
+    SetHp: Fn(usize),
     NoteLiveSize: Fn(Bytes<u32>),
     NoteReclaimed: Fn(Bytes<u32>),
 >(
     mem: &mut M,
-    heap_base: u32,
+    heap_base: usize,
     get_hp: GetHp,
```
luc-blaeser, line 72: Should we maybe rename this helper function, because it returns the unskewed pointer and has the same name like the skewed `get_hp()` in the compiler-generated code? The same would apply for `set_hp()`, maybe also in the other classical GCs?

rts/motoko-rts/src/memory/ic/linear_memory.rs
```diff
@@ -6,18 +6,25 @@ use crate::types::*;
 /// Amount of garbage collected so far.
 pub(crate) static mut RECLAIMED: Bytes<u64> = Bytes(0);
 
-// Heap pointer
+// Heap pointer (skewed)
 extern "C" {
-    pub(crate) fn setHP(new_hp: u32);
-    pub(crate) fn getHP() -> u32;
+    fn setHP(new_hp: usize);
```
luc-blaeser, line 11: I wonder whether it could make sense to use `Value` for `getHP()` and `setHP()`. Then, it would be more statically checked that one does not mix skewed and unskewed `HP`.

luc-blaeser: Maybe, the functions could be additionally renamed to `setSkewedHeapPointer()` and `getSkewedHeapPointer()` (in the compiler-generated code) because this design is special in comparison to the other unskewed heap boundaries in the RTS.

ggreif: Getting from `Value` to `u32` incurs cost. I am not willing to pay it. Or do you see a way of making it super cheap?

ggreif: Re. `setSkewedHeapPointer`, etc.
`setHP` is now a local function, with only two clients in this file. I am happy with it. (Plus, the motoko side would need a renaming too.) But you wanted to do the _incremental_ adaption, feel free to rename in that PR.

rts/motoko-rts/src/gc/mark_compact.rs
ggreif: `getHP` and `setHP` are confined into the source file where they are imported, I don't see danger of confusion. But see my other comments.

rts/motoko-rts/src/memory/ic/linear_memory.rs
luc-blaeser: > Getting from `Value` to `u32` incurs cost. I am not willing to pay it. Or do you see a way of making it super cheap? Also the arguments passed to `setHP` are more `usize`-like than `Value` like at the callsities.

I thought it is zero cost because `Value` is only a static wrapper. But I may be mistaken.

ggreif: You are right, it is that `from_*` routines that massage stuff. But still, we can change this later when it becomes pressing.

luc-blaeser (approved):
Looks good. Thank you.
I will maybe need to adjust a little bit when integrating the new bump allocator with the incremental GC.

PR #4096, opened 2023-07-03
perf: extend the overflow detection mask to the right if the bump increment is a power of 2

ggreif:
Trying out the idea in https://github.com/dfinity/motoko/pull/4064/files/4eb3dfeb7fdc10737605c631edff33b3f0976eff#r1250740318.

TLDR: The page-crossing detection mask can be extended for power-of-2 increments by one bit (to the right) which allows for accurate detection of not only `HP` remaining on-page, but also page boundary crossings. This enables the property that the slow path gets entered only when necessary.

~~Alas, it is flawed... with a `HP` that can point exactly to the boundary between pages.~~

With #4095 merged this cannot happen. Let's prove it with pen and paper...

Assume the allocation increment is `N` = 8, the smallest possible (byte-)increment that is a power of 2 when allocating objects. In hex `N = 0x0008`, thus the unextended overflow mask is `M = 0xFFF0`, and we assume that such a mask guarantees (by #4095) the implication `(HP + N) & M > 0 => "i8.store (HP + N) x" won't crash` for all `HP` of the form `0bXX..X11`. In this PR we have to strengthen the implication to the extended mask `M' = M | N`, i.e. one bit to the right. We proceed as follows:

We only need to consider the case where `(HP + N) & M = 0`, otherwise we are done by the above assumption. This reduces
- `(HP + N) & M' > 0 => "i8.store (HP + N) x" won't crash`
-  via `(HP + N) & (M | N) > 0 => "i8.store (HP + N) x" won't crash`
- to `(HP + N) & (M | N) = N => "i8.store (HP + N) x" won't crash`

so the bit under the submask `N` must be set and under `M` no bit is set. This gives us a good picture of `HP + N`, namely `0bPPP..P0000000000001X11` where the 16 `P` bits are irrelevant and `X` can be 0 or 1. This also means that `HP` must have been `N` less: `0bPPP..P0000000000000X11`. Since `HP` was the result of a previous increment, and thus `"i8.store HP x" doesn't crash`, writing a byte to `0bPPP..P0000000000001X11` won't crash either (since it is on the same page). This finishes the implication.

Choosing a different `N` — that is a power of 2 and `< 0x8000` — will result in a similar proof, only that there are more `X` bits.


ggreif:
A the tests reveal, there is a gap here. Whan the `HP` is at page boundary before and we add a power-of-two, then we technically didn't change pages, but the page wasn't allocated yet in first place. So this experiment has failed, and has shown that the simple things are often the more correct ones! Closing. @luc-blaeser

ggreif:
Reopening to test the combination with #4095! Worth a try...

src/codegen/compile.ml
```diff
@@ -1622,20 +1622,22 @@ module Tagged = struct
   let alloc env size tag =
     assert (size > 1l);
     let name = Printf.sprintf "alloc_size<%d>_tag<%d>" (Int32.to_int size) (Int32.to_int (int_of_tag tag)) in
-    (* Computes a (conservative) mask for the bumped HP, so that the existence of non-zero bits under it
+    (* Computes a mask for the bumped HP, so that the existence of non-zero bits under it
```
luc-blaeser, line 1625: If I am not mistaken, the mask is still conservative, isn't it? E.g. if we allocate 12, then 18 on a new partition, it is taking two times the slow path?

ggreif: I don't think so. E.g. allocating 28 first will be detected due to its own mask (it is not a power of 2):
```
(lldb) p/x (0x3333ffef + 28) & 0xffe0
(int) $6 = 0x00000000
```

ggreif: I only called the mask conservative because (as you noted in the other PR) it didn't detect that increments `0xPPPP0000` -> `0xPPPP0800` (powers of 2) remained on-page. The situation is now that with spewed HP and extended pow-of-2 masks both _page-hops_ and _remains_ get detected accurately.

luc-blaeser: Yes, indeed. I only wondered whether there are still cases where a sequence of allocations on the same page take the slow path, e.g. 12, then 48. It is not about correctness - these cases still check the page bounds when needed. I am curious about whether the mask extension is general in the sense that we only perform the slow path when necessary.

ggreif:
Okay, let me fool around a bit...
``` Haskell
import Data.List
import Data.Bits

split b | a <- b .&. 1, b' <- b `shiftR` 1, b' /= 0 = Just (head $ show a, b')
split b | 1 <- b .&. 1 = Just ('1', 0)
split _ = Nothing

bin = reverse . unfoldr split
```

Consider increments that are _single isle_, like 48 (= `0b11_000`).

src/codegen/compile.ml
```diff
@@ -1622,12 +1622,15 @@ module Tagged = struct
...
```
ggreif, line 1625:
```suggestion
    (* Computes a (conservative) mask for the bumped HP, so that the existence of non-zero bits under it
```

luc-blaeser (approved):
I followed the proof and it looks good to me.

PR #4103, opened 2023-07-05
feat: allow controllers to call `__motoko_stable_var_info` query endpoint

ggreif:
This was originally spec-ed, but there was no possibility to implement it (a canister couldn't get its own controller list).
In the meantime canisters can get their controller list (via query), but a query cannot call another query 🫤.

So we had to wait until the `ic0.is_controller` syscall became available. Now it is there and the implementation is easy.

(It was technically possible to call it before, but it invariably trapped.)


src/lowering/desugar.ml
```diff
@@ -438,8 +438,9 @@ and export_footprint self_id expr =
   ([ letD (var v typ) (
        funcE v (Shared Query) Promises [bind1] [] [ret_typ] (
            (asyncE T.Fut bind2
-              (blockE [expD (assertE (primE (I.RelPrim (caller, Operator.EqOp))
-                                        [primE I.ICCallerPrim []; selfRefE caller]));
+              (blockE [expD (assertE (orE (primE (I.RelPrim (caller, Operator.EqOp))
```
crusso, line 441: Might be cheaper to call ICCallerPrim just once, but don't think it matters.

crusso:
I guess we could now provide a helper in ExperimentalInternetComputer to call this on another canister than self, but not urgent.

src/lowering/desugar.ml
ggreif: I am not overly concerned :-)

ggreif: This is now #4108!

PR #4106, opened 2023-07-06
Remove reference to global maps

chenyan-dfinity:
Try to fix https://github.com/dfinity/motoko-playground/issues/148

PR #4107, opened 2023-07-06
bugfix: alternative to #4106

crusso:
Alternative to #4106 avoiding use of globals altogether to avoid leak.

Use a local functor to reallocate state - no need to reset afterwards as private so will not be re-used and should be collected by gc.

### DIFF
--- a/src/mo_idl/mo_to_idl.ml
+++ b/src/mo_idl/mo_to_idl.ml
@@ -6,198 +6,202 @@ open Type
 module E = Syntax
 module I = Idllib.Syntax
 
-let env = ref Env.empty
+(* use a functor to allocate temporary shared state *)
+module MakeState() = struct
 
-(* For monomorphization *)
-module Stamp = Type.ConEnv
-let stamp = ref Stamp.empty
+  let env = ref Env.empty
 
-module TypeMap = Map.Make (struct type t = con * typ list let compare = compare end)
-let type_map = ref TypeMap.empty
+  (* For monomorphization *)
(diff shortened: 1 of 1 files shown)


src/mo_idl/mo_to_idl.ml
```diff
@@ -6,217 +6,217 @@ open Type
...
+        | _ -> assert false
+       )
+
+end
 
 let prog (progs, senv) : I.prog =
-  env := Env.empty;
-  stamp := Stamp.empty;
-  type_map := TypeMap.empty;
+  let open MakeState() in
```
ggreif, line 204: Cool, I like this!

src/mo_idl/mo_to_idl.ml
```diff
@@ -6,217 +6,217 @@ open Type
...
+           (match s with
+            | Query -> [I.Query @@ no_region]
+            | Composite -> [I.Composite @@ no_region]
+            | Write -> []),
+           t1, args ts)
+       | _ -> assert false)
+    | Func _ -> assert false
+    | Async _ -> assert false
+    | Mut t -> assert false
+    | Pre -> assert false
```
ggreif, line 136:
```suggestion
    | Func _
    | Async _
    | Mut _
    | Pre -> assert false
```

ggreif (approved):
This is great if it works! I need to remember this technique...

crusso:
@chenyan-dfinity this is a PR to your branch, so just merge it in if you like.

chenyan-dfinity (approved):
Nice!

PR #4123, opened 2023-07-12
chore: work around `heartbeat`-related flakiness in `dao.mo` by voting twice

ggreif:
Execution of proposals happens by `heartbeat`, so the exact timing is non-deterministic. Now we vote twice, thus some time can pass and the state can advance to a stable point.

Alternatively we could just comment out the `heartbeat`.


chenyan-dfinity (approved):
LGTM

PR #4128, opened 2023-07-17
bugfix: support import of Candid composite queries

crusso:
Fixes bug reported by @Dfinity-skaestle.


Dfinity-skaestle:
Thanks @crusso !

ggreif (approved):
This is fine! We'll need to think later of extending it to `ic-ref` as well...

PR #4132, opened 2023-07-18
chore: bump `rustc` versions

ggreif:


ggreif:
Looks like `-Os` is now taken much more seriously by `rustc` at the expense of the cycle consumption. @luc-blaeser what about switching to `-O2` or such?

rts/motoko-rts/Cargo.toml
```diff
@@ -38,4 +38,4 @@ panic = "abort"
 [profile.release]
 panic = "abort"
 codegen-units = 1 # makes debugging easier
-opt-level = "s" # optimize for size
+opt-level = 2 # optimize some
```
ggreif, line 41: https://github.com/dfinity/canister-profiling/pull/63/commits `opt 3` seems to do a good job!

luc-blaeser:
Obsolete

PR #4133, opened 2023-07-18
feat(ci): add macos m1 runners in os matrix for CI jobs

EnzoPlayer0ne:


.github/workflows/test.yml
```diff
@@ -7,7 +7,7 @@ jobs:
   tests:
     strategy:
       matrix:
-        os: [ ubuntu-latest, macos-latest ]
+        os: [ ubuntu-latest, macos-latest, macos-m1-13, apple-silicon-m1 ]
```
ggreif, line 10: what's the difference?

EnzoPlayer0ne: Apparently GitHub has the shared MacOS M1 runners in beta per our account manager. However I could not find the tag available in GitHub, I searched around and tried some tags which I believe belong to other CI providers. 
Having those shared runners would have made it easier to compile on the m1 architecture. 

Instead as those shared runners do not seem to exist, we will make our on-premise M1 runners available to public repositories of the DFINITY foundation. 

I will close this MR as it was only to test the tags.

luc-blaeser:

Issue #3175, opened 2022-03-31
Timo's Motoko wishlist


crusso:
* shift operators for Nat  (and any other operators that are available for `NatX`/`IntX` but missing for `Nat`/`Int`) — See #3112
* full matrix of conversions between fixed size `NatX`/`IntX` or at least between adjacent pairs of types (e.g. `Nat8` <-> `Nat16` etc).
* `NatX` -> Blob operations, for easy access to bytes in little-endian and perhaps big-endian encoding.
* Indexing for `Blob` value (`blob[n]` or `Blob.sub(blob, n)`)
* (Mutable blob type for writing efficient serializers, see perhaps #3149)

@timohanke

nomeata:
Now that we have inlining of prim wrappers (or I guess once we have that, not sure what the status is), maybe we can promise that two conversions via `Nat`/`Int` will always cancel out, instead of providing a full matrix? But may explicit is actually better.

PR #4139, opened 2023-07-20
feat: numerical type conversions between adjacent fixed-width types

kentosugama:
Add type conversions between `Nat`s of adjacent sizes. Similarly for `Int`s.

Namely:
- `Nat8` <--> `Nat16`
- `Nat16` <--> `Nat32`
- `Nat32` <--> `Nat64`
- `Int8` <--> `Int16`
- `Int16` <--> `Int32`
- `Int32` <--> `Int64`

DX feature requested by Zhenya Usenko in the motoko working group and Timo:
partially #3175


src/prelude/prim.mo
```diff
@@ -39,47 +39,61 @@ module Types = {
...
 // for testing
 func idlHash(x : Text) : Nat32 { (prim "idlHash" : Text -> Nat32) x };
 
 // Printing
 
 func debugPrint(x : Text) { (prim "print" : Text -> ()) x };
-func debugPrintNat(x : Nat) { debugPrint (@text_of_Nat x) };
-func debugPrintInt(x : Int) { debugPrint (@text_of_Int x) };
-func debugPrintChar(x : Char) { debugPrint (charToText x) };
+func debugPrintNat(x : Nat) { debugPrint(@ text_of_Nat x) };
```
ggreif, line 55: "@ text_of_Nat" -> "@text_of_Nat"
also in line 56

kentosugama: Yes this is a formatter error. Ryan has been informed

rvanasa: Pushed a fix and updated the VS Code extension.

kentosugama: Thank you!

src/codegen/compile.ml
```diff
@@ -9669,7 +9669,91 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
+      compile_exp_as env ae SR.UnboxedWord32 e ^^
+      G.i (Convert (Wasm.Values.I64 I64Op.ExtendUI32))
+    | Nat16, (Nat8 as pty)->
+      SR.Vanilla,
+      let num_bits = (TaggedSmallWord.bits_of_type pty) in
+      let max_val = Int32.(shift_left (sub (shift_left 1l num_bits) 1l) 16) in
+      compile_exp_vanilla env ae e ^^
+      compile_rel_const I32Op.LeU max_val ^^
+      E.else_trap_with env "losing precision" ^^
+      compile_exp_vanilla env ae e ^^
```
kentosugama, line 9691: FIXME: don't compile `e` twice. Save to local instead

src/codegen/compile.ml
```diff
@@ -9669,7 +9669,91 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
+      compile_exp_as env ae SR.UnboxedWord64 e ^^
+      G.i (Convert (Wasm.Values.I32 I32Op.WrapI64))
+    | Int8, Int16 ->
+      SR.Vanilla,
+      compile_exp_vanilla env ae e ^^
+      compile_shrS_const 8l
+    | Int16, Int32 ->
+      SR.Vanilla,
+      compile_exp_vanilla env ae e ^^
+      compile_shrS_const 15l
```
kentosugama, line 9719: Comment explaining 15

src/codegen/compile.ml
```diff
@@ -9669,7 +9669,91 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
+      let max_val = Int32.(shift_left (sub (shift_left 1l num_bits) 1l) 16) in
+      compile_exp_vanilla env ae e ^^
+      compile_rel_const I32Op.LeU max_val ^^
+      E.else_trap_with env "losing precision" ^^
+      compile_exp_vanilla env ae e ^^
+      compile_shl_const (Int32.of_int num_bits)
+    | Nat32, (Nat16 as pty)->
+      SR.Vanilla,
+      let num_bits = (TaggedSmallWord.bits_of_type pty) in
+      let max_val = Int32.(sub (shift_left 1l num_bits) 1l) in
```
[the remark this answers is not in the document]
kentosugama: @ggreif I think the shift right test doesn't work for `Nat16` --> `Nat8` conversion, since the lower bits just move into the lower 16 positions.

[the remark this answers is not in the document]
kentosugama: NVM Gabor reminded me that I can just shift by 24 🤦‍♂️

test/run/conversions.mo
```diff
@@ -3,14 +3,14 @@ import Prim "mo:⛔";
 // Nat <--> Nat8
 
 func test_Nat_Nat8(n1 : Nat, n2 : Nat8) {
-  assert(Prim.natToNat8 n1 == n2);
-  assert(Prim.intToNat8Wrap n1 == n2);
-  assert(n1 == Prim.nat8ToNat n2);
+  assert (Prim.natToNat8 n1 == n2);
```
ggreif, line 6: `assert` is syntax, not a function, so the parens are redundant. If you feel like it :-)

test/trap/outrange-int64-int32-upper.mo
```diff
@@ -0,0 +1,2 @@
+import Prim "mo:⛔";
+let _ = Prim.int64ToInt32(2 ** 31);
```
ggreif, line 2: comment?

ggreif (approved):
LGTM! A few nits only...

test/run/conversions.mo
kentosugama: Maybe it makes sense to add this rule to the formatter? @rvanasa 

Also, I noticed there's a lot of formatting discrepancies in the motoko repo, since they haven't been ran through the formatter. Maybe it makes sense to format the whole repo and require it in CI (as long as the tool can be used by non-vscode users). 

I only mention this because the formatting changes clutters the `Files Changed` for PR reviews.

ggreif:
> In terms of gas, 1 tests improved and the mean change is -2.9%.

I am also observing bogus percentages in other innocent PRs. Oh well...

src/codegen/compile.ml
```diff
@@ -9669,7 +9669,106 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
+      set_val ^^
+      get_val ^^
+      compile_shl_const (Int32.of_int num_bits) ^^
+      compile_shrS_const (Int32.of_int num_bits) ^^
+      get_val ^^
+      compile_eq env Type.(Prim Nat16) ^^
+      E.else_trap_with env "losing precision" ^^
+      get_val ^^
+      compile_shl_const (Int32.of_int num_bits)
+    | Int32, (Int16 as pty)->
```
ggreif, line 9744:
```suggestion
    | Int32, (Int16 as pty) ->
```

src/codegen/compile.ml
```diff
@@ -9669,7 +9669,106 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
+      SR.Vanilla,
+      let num_bits = (TaggedSmallWord.bits_of_type pty) in
+      let (set_val, get_val) = new_local env "convertee" in
+      compile_exp_as env ae SR.UnboxedWord32 e ^^
+      set_val ^^
+      get_val ^^
+      compile_shl_const (Int32.of_int num_bits) ^^
+      compile_shrS_const (Int32.of_int num_bits) ^^
+      get_val ^^
+      compile_eq env Type.(Prim Nat32) ^^
```
ggreif, line 9754: This should be a single instruction in the output, what you have is correct, albeit a bit indirect.

src/codegen/compile.ml
```diff
@@ -9669,7 +9669,106 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
+      compile_shrS_const (Int32.of_int num_bits) ^^
+      get_val ^^
+      compile_eq env Type.(Prim Nat16) ^^
+      E.else_trap_with env "losing precision" ^^
+      get_val ^^
+      compile_shl_const (Int32.of_int num_bits)
+    | Int32, (Int16 as pty)->
+      SR.Vanilla,
+      let num_bits = (TaggedSmallWord.bits_of_type pty) in
+      let (set_val, get_val) = new_local env "convertee" in
```
ggreif, line 9747: Fun fact: the data constructor is `,`, the parens are a Haskell-ism.

src/codegen/compile.ml
```diff
@@ -9669,7 +9669,106 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
+    | Nat8, Nat16 ->
+      SR.Vanilla,
+      compile_exp_vanilla env ae e ^^
+      compile_shrU_const 8l
+    | Nat16, Nat32 ->
+      SR.Vanilla,
+      compile_exp_vanilla env ae e ^^
+      compile_shrU_const 15l (* resulting Nat32 will always be unboxed *)
+    | Nat32, Nat64 ->
+      SR.UnboxedWord64,
```
ggreif, line 9681: We currently don't have `sr_hint` in this function, but if we had, we could specialise for e.g. `Vanilla` output in a very efficient way (basically a `nop`). Similarly in other places.

src/codegen/compile.ml
kentosugama: I wonder if we can configure `ocamlformatter` to do these things. Seems silly for humans to be checking these manually

src/codegen/compile.ml
kentosugama: You're saying lines 9751 to 9754 should be 1 instruction? How is that possible?

kentosugama: Marking resolved

PR #4144, opened 2023-07-24
Integrate Regions in Random GC Tests

luc-blaeser:
Extending the random GC tests to also allocate regions. 
The same tests can also be used for long-running and larger-scaling tests. 
Therefore, the maximum number of randomly created regions is limited - until region garbage collection will be supported.


matthewhammer:
Thanks @luc-blaeser !

crusso:
Ohoh, we never review nor merged this? Shall I take another look @luc-blaeser?

luc-blaeser:
> Ohoh, we never reviewed nor merged this? Shall I take another look @luc-blaeser?

No worries. This is only an extended test case. I reran the tests after a recent merge. If it passes, you may take another look. Thanks a lot.

crusso (approved):
LGTM, never hurts to have more testing.

PR #4145, opened 2023-07-24
experiment: refactor enforcement of --max-stable-pages

crusso:
builds on #4143 (which builds on #3768)

--max-stable-pages is enforced in one place only, the low-level page allocator StableMem.

It now bounds the physical stable pages occupied by the user code, leaving the rest as reserve for stable variables.

This means some tests need to know if they are running with --stable-regions or not because the overheads are different.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -103,11 +103,13 @@ jobs:
         folder: report-site-copy
         single-commit: true
 
-  artifacts-ubuntu:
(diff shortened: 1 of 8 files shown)


matthewhammer (approved):
LGTM

matthewhammer:
@crusso Looks like only the migration test failed, and only on one arch (like earlier today, during our Zoom call).

Going to merge.

PR #4146, opened 2023-07-25
bugfix: define regions to be stable, but not shared, adding testcase

crusso:
Fix to PR #3768 

Regions, like mutable arrays and object with mutable fields, are stateful, so can't be shared, but should be stable.

That was always the intention but #3768 didn't quite implement it.


src/mo_types/type.ml
```diff
@@ -817,6 +817,7 @@ let serializable allow_mut t =
       match t with
       | Var _ | Pre -> assert false
       | Prim Error -> false
+      | Prim Region -> allow_mut (* stable, but not shared *)
```
matthewhammer, line 820: LGTM

test/fail/region-ops.mo
```diff
@@ -0,0 +1,29 @@
...
+  to_candid(r); // reject
+};
+
+func FromCandid(b : Blob) : ?Region {
+  let r: ?Region = from_candid b; // reject
+};
+
+do {
+   type t = async Region; // reject
+};
```
matthewhammer, line 29:
```suggestion
};

```

PR #4147, opened 2023-07-25
experiment: repro for bug in region.grow

crusso:
@luc-blaeser mentioned seeing  unexpected "stable memory out of bounds" failures.

I think the bug is due to region_grow only checking the page allocation against the logical maximum determented --max-stable-pages, which may be set higher than what is currently available on the replica. In particular, the region_grow implementation fails to check whether the call to ic0.grow actually succeeds and pretends it has, which then let's you read/write unallocated pages.

This test reproduces the error. 

It's actually fixed in https://github.com/dfinity/motoko/pull/4145


test/run-drun/ok/region0-stable-mem-grow-fail.drun-run.ok
```diff
@@ -0,0 +1,3 @@
+ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
+debug.print: 0
+ingress Err: IC0502: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped: stable memory out of bounds
```
crusso, line 3: This should not happen and indicates that StableMemory.grow has incorrectly indicated success, not failure.

PR #4148, opened 2023-07-25
bug: missing overflow check in regions.grow

crusso:
A repro exposing the missing overflow check that we discussed (offset+len-1 can overflow).


matthewhammer:
Thanks @crusso! (This is great, in the sense that it gives a test for those overflow-prone additions we talked about today.)

PR #4158, opened 2023-08-03
Memory Reserve for Queries and Upgrades

luc-blaeser:
# Memory Reserve for Queries and Upgrades

## Current Situation
With the incremental GC, programs can scale to the full 4GB memory space. As a consequence, update calls can allocate the full memory space, such that even a simple query, a simple composite query, or a simple upgrade logic can no longer succeed. This could also happen with classical non-copying GCs if deterministic time slicing is sufficiently extended.

## Provisional Solution 
This PR prevents update calls (including canister initialization, heartbeats, and timers) from allocating the full memory by leaving a reserve for queries, composite queries, and canister upgrades. During queries, composite queries, and canister upgrades, garbage collection is suspended, such that the reserve is available to the mutator code. Callbacks of composite queries can also use the memory reserve.

The current allocation limit for upgrade calls is 3.75 GB and applies to all GCs. This gives a reserve of 224 MB for the incremental GC as the last 32MB partition is unallocated in the current design. The scheduling heuristics of the incremental and generational GC needs to consider the reduced capacity for determining memory shortage.

This PR can be viewed as a temporary measure until a memory reserve is implemented in the IC runtime system.

## Important Note
The memory reserve may **not** be sufficient for a complex canister upgrade logic or large amount of stable heap data. Moreover, the upgrade logic may exceed the instruction limit. Thorough upgrade testing is required for canisters in any case.


src/codegen/compile.ml
```diff
@@ -5242,6 +5242,22 @@ module RTS_Exports = struct
       edesc = nr (FuncExport (nr rts_trap_fi))
     });
 
+    (* Keep a memory reserve when in update or init state. 
+       This reserve can be used by queries, composite queries, and upgrades. *)
+    let keep_memory_reserve_fi = E.add_fun env "keep_memory_reserve" (
+      Func.of_body env [] [I32Type] (fun env ->
+        Lifecycle.get env ^^
+        compile_eq_const Lifecycle.(int_of_state InUpdate) ^^
```
ulan, line 5250: Does this also cover heartbeat and timers?

ulan (approved):
Thanks for implementing it!

src/codegen/compile.ml
luc-blaeser: Yes, as far as I know. I also checked it with some local tests: Heartbeat and timers functions (defined by the user program) are executed as `InUpdate` (inferred from `Shared Write`).
Maybe @ggreif could confirm?

crusso: I think that's correct. The only slight wrinkly is that composite_queries callbacks also execute in this mode, I think, so the callbacks will try to maintain the reserve. Does it matter?

crusso (approved):
LTGM, but why is there no similar logic for the mark-compact and copying GC? I guess the latter should alway have space, but what about the former?

luc-blaeser:
> LTGM, but why is there no similar logic for the mark-compact and copying GC? I guess the latter should alway have space, but what about the former?

The reserve is currently applied to all GCs, although some may currently not need it as they may exceed the instruction limit before reaching the memory limit. However, only the incremental and generational GC implement a special scheduling logic by triggering GC more eagerly on memory scarceness. I am not sure, whether we should maybe add this for the compacting GC too. I only lowered the threshold for those existing "critical memory" scheduling modes as the reserve has been introduced.
Or maybe I miss an aspect. Or do you think of another logic?

src/codegen/compile.ml
luc-blaeser: Good point, thanks for noticing: Are the callbacks of composite queries executed under `InUpdate` and later rolled back? I wonder whether there would be an easy way to also detect them and exempt them from the reserve.

luc-blaeser: Just double checked: Composite queries are already correctly handled in the current version. They have state `InComposite` and can exhaust the memory reserve.

luc-blaeser: Update: I now also detect callbacks of composite queries and allow them to also use the memory reserve.

luc-blaeser:
I will need to remeasure the performance impact of these latest changes...

luc-blaeser:
I had to move the memory reserve `drun` tests to the non-CI folder because these tests sporadically fail on CI Linux with the error "Ingress message did not finish executing within 10000 batches"...

luc-blaeser:
I also had to downscale one of the tests (`stable-region-migration.drun`) because it continuously exceeded the `drun` batch limit on CI (under Linux).

src/codegen/compile.ml
```diff
@@ -8475,9 +8479,25 @@ module FuncDec = struct
...
+  let callback_start env =
+    Lifecycle.is_in env Lifecycle.InComposite ^^
+    G.if0
+      (G.nop)
+      (message_start env (Type.Shared Type.Write))
+
+  let callback_cleanup env =
+    Lifecycle.is_in env Lifecycle.InComposite ^^
+    G.if0
+      (G.nop)
```
crusso, line 8498: So I guess this means we never do GC in a composite query, since the GC increment is only invoked by message_cleanup env (Type.Shared Type.Write). 
I'm not sure - do we want that behaviour or not? I can see the argument for not doing the GC at return from a composite query. But this means we never do any GC during a composite query, I think. It's probably fine, but just wanted to point it out.

crusso: Re-reading the PR description it look like this is deliberate.  Fine by me, I think.

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -118,7 +118,7 @@ static mut OLD_GENERATION_THRESHOLD: usize = 32 * 1024 * 1024;
 static mut PASSED_CRITICAL_LIMIT: bool = false;
 
 #[cfg(feature = "ic")]
-const CRITICAL_MEMORY_LIMIT: usize = (4096 - 512) * 1024 * 1024;
+const CRITICAL_MEMORY_LIMIT: usize = (4096 - 512) * 1024 * 1024 - crate::memory::MEMORY_RESERVE;
```
crusso, line 121: I forget - what is the other -512 for?

crusso: Oh, I see that's when we start doing aggressive GC. Nvm

crusso (approved):
LGTM.

I perf ever becomes and issue, we could (perhaps) have the lifecycle transition just set the limit in a global and pass it directly to memory_grow (to avoid the keep_reserve() helper and (expensive) calls)

src/codegen/compile.ml
luc-blaeser: This a good point. I also thought about this and I do not have a clear answer. 

On the one hand, GC could make in composite query if they are very garbage-intense and have a lot of callbacks which I am not sure whether this is a typical case in practice. 

On the other hand, the state changes are discarded anyway at the end of composite query and the costs of the temporary garbage collection may be unjustified. If I remember right, the state is even reverted between repeated composite query calls to the same canister. 

Because of this short-lived nature, I had a slight preference for skipping the GC in favor of cheaper composite queries...

rts/motoko-rts/src/gc/generational.rs
luc-blaeser: Yes, the `-512` is the critical limit. I lowered it further by the newly introduced memory reserve...

luc-blaeser:
> LGTM.
> 
> I perf ever becomes and issue, we could (perhaps) have the lifecycle transition just set the limit in a global and pass it directly to memory_grow (to avoid the keep_reserve() helper and (expensive) calls)

Thanks a lot for reviewing it again. Very much appreciated.

PR #4160, opened 2023-08-03
Incremental GC: Adjust Total Allocation Statistics

luc-blaeser:
# Incremental GC: Adjust Total Allocation Statistics

This is a minor, mostly cosmetic fix for the incremental GC: The total allocation statistics (`rts_total_allocation()`) was not monotonously growing but temporarily growing and shrinking while a GC run involved multiple increments. The reason was that the evacuation space was incorrectly counted as new allocations.

This PR does not imply any significant performance impact according to the GC benchmark measurements.

Before this PR (example of `cancan` GC benchmark case):

image

**Fix**: The temporary copying space for evacuations is no longer considered in the total allocation metric.

With this PR (example of `cancan` GC benchmark case):

image


crusso (approved):
LGTM

luc-blaeser:

PR #4161, opened 2023-08-04
More Conservative Incremental GC Scheduling

luc-blaeser:
# More Conservative Incremental GC Scheduling

The existing incremental GC scheduling is reclaiming memory effectively up to 4GB according to the tests, the GC benchmark, and also other analyzed user programs so far.
However, it could also make sense, to schedule the GC a little bit more conservatively to reduce the allocated memory size, the reclamation latency, and the memory pressure. 

## Scheduling Refinement
The GC is **additionally** scheduled on every absolute heap growth of 500 MB or more.

## GC benchmark results:

Aspect | Difference
-------- | ----------
Runtime | (none)
Allocated Memory | -6%
Scalability | +28% (*)

(*) The scalability improvement is mostly due to the `scalable-buffer` benchmark case that scales much higher with this more conservative scheduling. For other cases, the scalability is slightly lower. This heavily depends on when GC is scheduled when the memory is nearly full and the mark bitmaps may no longer be allocated.

### DIFF
--- a/rts/motoko-rts/src/gc/incremental.rs
+++ b/rts/motoko-rts/src/gc/incremental.rs
@@ -77,12 +77,12 @@ unsafe fn should_start() -> bool {
         return false;
     }
 
-    const ABSOLUT_GROWTH_THRESHOLD: Bytes<u64> = Bytes(512 * 1024 * 1024);
+    const ABSOLUTE_GROWTH_THRESHOLD: Bytes<u64> = Bytes(512 * 1024 * 1024);
 
     let current_allocations = partitioned_memory::get_total_allocations();
     debug_assert!(current_allocations >= LAST_ALLOCATIONS);
     let absolute_growth = current_allocations - LAST_ALLOCATIONS;
-    if absolute_growth > ABSOLUT_GROWTH_THRESHOLD {
+    if absolute_growth > ABSOLUTE_GROWTH_THRESHOLD {
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


rts/motoko-rts/src/gc/incremental.rs
```diff
@@ -72,22 +72,32 @@ unsafe fn should_start() -> bool {
...
-    let growth_threshold = if heap_size > CRITICAL_HEAP_LIMIT {
-        CRITICAL_GROWTH_THRESHOLD
-    } else {
-        NORMAL_GROWTH_THRESHOLD
-    };
+    if heap_size.as_usize() < PARTITION_SIZE {
+        return false;
+    }
+
+    const ABSOLUT_GROWTH_THRESHOLD: Bytes<u64> = Bytes(512 * 1024 * 1024);
```
luc-blaeser, line 80:
```suggestion
    const ABSOLUTE_GROWTH_THRESHOLD: Bytes<u64> = Bytes(512 * 1024 * 1024);
```

rts/motoko-rts/src/gc/incremental.rs
```diff
@@ -72,22 +72,32 @@ unsafe fn should_start() -> bool {
...
 
     let current_allocations = partitioned_memory::get_total_allocations();
     debug_assert!(current_allocations >= LAST_ALLOCATIONS);
     let absolute_growth = current_allocations - LAST_ALLOCATIONS;
+    if absolute_growth > ABSOLUT_GROWTH_THRESHOLD {
```
luc-blaeser, line 85:
```suggestion
    if absolute_growth > ABSOLUTE_GROWTH_THRESHOLD {
```

luc-blaeser:
Put back for later when needed.

PR #4162, opened 2023-08-04
Incremental GC: Handle Huge Allocation in One Message

luc-blaeser:
# Incremental GC: Handle Huge Allocation in One Message

**To be discussed whether this case should be handled in the GC or the current behavior is fine.**

This PR focuses on a case reported by @chenyan-dfinity observed in the canister-profiling benchmark (https://github.com/dfinity/canister-profiling) for the incremental GC. A reproduction is the `splay-tree` also added to GC benchmark (https://github.com/luc-blaeser/gcbench/tree/luc%2Fsplay-tree). 

## Problem Analysis

Currently, the incremental GC adopts its collection speed to the concurrent allocation rate. 
However, there is a special case not yet handled that can lead to high reclamation latency:
* A program allocates a huge amount of garbage (gigabytes) in one single IC message.
* The GC is then started and performs an increment of normal duration at the end of this IC message.
* If the program now continues to only allocate little memory, the GC keeps running in normal pace, with short increments.
* It then requires several GC increments until the memory is reclaimed. 
* As a consequence, it seems that the footprint of the incremental GC is higher.

## Possible Solution

Observe the heap growth in the current IC message and adjust the increment duration according to this.
This handles both concurrent allocations as well as a large allocations in a single message that starts the GC.

## Performance Impact

A special heuristics can be applied to prevent performance degrades in normal cases: 128 MB allocation tolerance is granted for message without extra GC costs. Without this heuristics, the runtime overhead would be around 4%, as measured by the GC benchmark.

### DIFF
--- a/rts/motoko-rts/src/gc/incremental.rs
+++ b/rts/motoko-rts/src/gc/incremental.rs
@@ -97,7 +97,7 @@ unsafe fn record_gc_start<M: Memory>(state: &mut State) {
     ALLOCATIONS_AT_LAST_GC_RUN = partitioned_memory::get_total_allocations();
 
     // Heuristics: Tolerate a certain heap growth before the start of a GC run without extra GC increment costs.
-    // This only applies to the allocations before a GC start but not for allocations during a running GC.
+    // This only applies to the allocations before a GC start but not to allocations during a running GC.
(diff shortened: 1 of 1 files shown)


luc-blaeser:
Reactivate whenever needed...

PR #4164, opened 2023-08-09
`Text.lowercase` in RTS

matthewhammer:
- [x] Try to implement `lowercase` in the RTS, using what we have at hand.
- [x] Expose it as a new primitive with the compiler.
- [ ] Tests pass -- there tests, but some strange behavior at the moment.


crusso:
Superseded by https://github.com/dfinity/motoko/pull/4216

PR #4167, opened 2023-08-10
chore: `ic-ref` is deprecated, don't use it

ggreif:


crusso:
> This PR simply comments out the use(r)s of ic-ref-run (instead of eradicating them), so that we can reactivate this quickly in case a maintainer should show up. We can leave this in optimistically, or eventually get rid of it.

Perfect!

crusso (approved):

PR #4171, opened 2023-08-13
experiment: lazy migration for stable memory for regions

crusso:
builds on the unwieldly https://github.com/dfinity/motoko/pull/3768.

Introduces lazy migration of stable memory format 0/1 (no ESM, some ESM) to 2 (Regions) by 

1. Adding an explicit global (StableMem.get/set_version) that tracks the current version.
2. Transitioning from 0 to 1 on first (acutally growing) StableMem.grow.
3. Transitioning from 0/1 to 2 on first Region.new().
4. Testing current version on each (legacy) StableMemory operation and dispatching to the correct legacy/region implementation.

As a result programs have zero overhead until they allocate the first region and all programs can access Region.mo without having to set a flag.

In addition, ExperimentalStableMemory retains the same low space overhead as before, at least until the first region is allocated. Each ESM read-write is slightly slower since it has to switch on the version, but I can live with that.

Avoids:
* the need to set a flag to enable Regions (and the need to explain and document that)
* significant overhead of region metadata (1-8MB) until regions are actually used.
* language dependence on compiler flag (c.f. Haskell).

Removes the need for a user facing --stable-regions flag. Instead, for testing, the flag now just forces eager 
(not lazy) migration to version 2.

Drive-by: remove prim stableMemoryRegion since its dangerous (the region returned won't alias the internal region0 after an upgrade without more work).


test/run-drun/regions-pay-as-you-go.mo
```diff
@@ -0,0 +1,37 @@
...
+}
+
+//SKIP run
+//SKIP run-low
+//SKIP run-ir
+// too slow on ic-ref-run:
+//SKIP comp-ref
+//CALL upgrade ""
+//CALL upgrade ""
+//CALL upgrade ""
```
crusso, line 37:
```suggestion
//CALL upgrade ""

```

test/run-drun/stable-mem-pay-as-you-go.mo
```diff
@@ -0,0 +1,36 @@
...
+}
+
+//SKIP run
+//SKIP run-low
+//SKIP run-ir
+// too slow on ic-ref-run:
+//SKIP comp-ref
+//CALL upgrade ""
+//CALL upgrade ""
+//CALL upgrade ""
```
crusso, line 36:
```suggestion
//CALL upgrade ""

```

src/codegen/compile.ml
```diff
@@ -5381,6 +5350,177 @@ module StableMem = struct
 
 end (* StableMemory *)
```
matthewhammer, line 5351: `StableMem` vs `StableMemory` is a bit confusing for people that haven't seen all of the stages of evolution.

Is there a plan to consolidate these in some future PR, I wonder?

matthewhammer (approved):
LGTM, modulo the `StableMem` and `StableMemory` distinction in `compile.ml`.  Maybe we can combine and consolidate these in the future, or explain why they are separate?

matthewhammer:
> Each ESM read-write is slightly slower since it has to switch on the version, but I can live with that.

Okay.  Sounds good.

> Removes the need for a user facing --stable-regions flag. Instead, for testing, the flag now just forces eager
(not lazy) migration to version 2.

Also sounds good.

src/codegen/compile.ml
crusso: Yeah, I agree. I wanted to keep the StableMem name for the old module which is largely unchanged. 

Maybe StableMemoryAdapter (since it adapts between old an new?), or StableMemoryPrims?

crusso: StableMem basically enforces --max-stable-pages and hides from the user any IC stable memory added for stable variables.

StableMemory provides the guts of the StableMemoryPrims, dynamically switching between legacy and regions.

crusso:
@matthewhammer thanks for taking a look!

rts/motoko-rts/src/region.rs
```diff
@@ -648,18 +674,23 @@ pub(crate) unsafe fn region_migration_from_v2plus_into_v2<M: Memory>(mem: &mut M
 //
 // region manager migration/initialization, with pre-existing stable data.
 //
-// Valid cases:
-//  - from version 1 into version 2.
-//  - from version 2 into version 2.
-//
 #[ic_mem_fn]
-pub(crate) unsafe fn region_init<M: Memory>(mem: &mut M, from_version: i32) {
+pub(crate) unsafe fn region_init<M: Memory>(mem: &mut M, use_stable_regions: i32) {
```
luc-blaeser, line 678: Detail: Maybe the `bool` type could be declared for `use_stable_regions` (although it is an integer passed by the compiler-generated code).

rts/motoko-rts/src/region.rs
```diff
@@ -648,18 +674,23 @@ pub(crate) unsafe fn region_migration_from_v2plus_into_v2<M: Memory>(mem: &mut M
...
+                region_migration_from_no_stable_memory(mem);
+            };
+        }
+        VERSION_SOME_STABLE_MEMORY => {
+            assert!(crate::stable_mem::size() > 0);
+            if use_stable_regions != 0 {
+                region_migration_from_some_stable_memory(mem);
+            };
+        }
+        _ => region_migration_from_regions_plus(mem), //check format & recover region0
```
luc-blaeser, line 693: Possible assertion: `use_stable_region` would be `true` (`!=0`) in this case.

src/codegen/compile.ml
```diff
@@ -532,8 +532,17 @@ module E = struct
     | ts -> VarBlockType (nr (func_type env (FuncType ([], ts))))
 
   let if_ env tys thn els = G.if_ (as_block_type env tys) thn els
+
+  (* NB: confuses wasm-opt, don't use for now
+  let _multi_if_ env tys1 tys2 thn els =
```
luc-blaeser, line 537: Just out of curiosity: Is `wasm-opt` failing or creating invalid code?

luc-blaeser (approved):
Great simplification with lazy region metadata allocation.
Also, very nice refactoring regarding `region0`.

src/codegen/compile.ml
crusso: I don't fully recall, but I think it was failing.

crusso: Commit https://github.com/dfinity/motoko/pull/4171/commits/280ed54d50d5fb8020ef2f85aa05ca358d5f42ef produces this error. Basically I was trying to conditionally call one of two functions with args already on the stack and wasm-opt barfed:

```
test-bench> @@ -0,0 +1,4 @@
test-bench> +Error: Failed to read module
test-bench> +
test-bench> +Caused by:
test-bench> +    [parse exception: Block requires more values than are available (at 0:57416)]

File: rts/motoko-rts/src/region.rs (hunk 5 above)
Comment (crusso) [reply rts/motoko-rts/src/region.rs:693]: The flag now means for migration to regions after upgrade, so (I think) it may not be true in this case. I'm mostly keeping the flag to avoid rewriting all the tests, but we could jettison it.

File: src/codegen/compile.ml (hunk 6 above)
Comment (crusso) [reply src/codegen/compile.ml:537]: I should probably be a good citizen and report the bug, but I'd rather make progress..

File: rts/motoko-rts/src/region.rs (hunk 4 above)
Comment (crusso) [reply rts/motoko-rts/src/region.rs]: Yeah, we don't seem to use strong types across the boundary. I made it a u32 for more consistency at least.

File: rts/motoko-rts/src/region0.rs [hunk 7]
```diff
@@ -1,19 +1,9 @@
...
-    assert_ne!(REGION_0, NO_REGION);
-    crate::region::region_load(mem, REGION_0, offset, dst)
-}
-
-unsafe fn region0_store<M: Memory>(mem: &mut M, offset: u64, src: &[u8]) {
-    assert_ne!(REGION_0, NO_REGION);
-    crate::region::region_store(mem, REGION_0, offset, src)
-}
-
 #[ic_mem_fn]
```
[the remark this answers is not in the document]
crusso: Done. Also made an assert into a debug assert for better perf.

PR #4172, opened 2023-08-16
regions: add missing write-barrier plus refactoring

crusso:
* add the missing write_barrier when initializing fields of a deserialized region (see `Region::init_region`)
* as suggested by Luc, use RTS implemented alloc_region, init_region and field access functions to avoid duplication of logic in rts and compile.ml
* render region implementation abstract to compile.ml


rts/motoko-rts/src/region.rs
```diff
@@ -370,34 +370,72 @@ fn write_magic() {
...
+    let r = r.as_region();
+    debug_assert!(id <= meta_data::max::REGIONS as u32);
+    (*r).id = id as u16;
+    debug_assert!(
+        page_count
+            <= (vec_pages.as_blob().len().as_u32() / meta_data::bytes_of::<u16>() as u32)
+                * meta_data::size::PAGES_IN_BLOCK
+    );
+    (*r).page_count = page_count;
+    write_with_barrier(mem, &mut (*r).vec_pages, vec_pages);
```
crusso, line 418: This is the barrier!

crusso:
@luc-blaeser is this ok to merge?

PR #4176, opened 2023-08-18
regions: refactor for 64-bit region ids, dropping region table 

crusso:
Builds on #3768.

With future GC in mind, use inexhaustible 64-bit region id, and drop dense region table, instead encoding size per block, not region, as partial, block level page counts. Regions can be recovered from the sparse representation in the block table.

Region recovery now does two linear scans of block table to recover region page_count followed by block recovery.

Actually metadata memory overhead from 9 to 6 pages, by removing 12 (8 byte pagecount + 4 reserve) bytes per region entry but adding 7 = (64-bit region id + 1 byte size) bytes per block entry.


test/run-drun/repro.mo
```diff
@@ -0,0 +1,60 @@
...
+  stable var r1 = Region.new();
+  stable var r2 = Region.new();
+
+  let block_size_in_pages = 128 : Nat64;
+
+  P.debugPrint "grow three big regions (including region0).";
+  // Interleave growing regions by a block each:
+  do {
+    P.debugPrint "1";
+    ignore StableMemory.grow(block_size_in_pages);
```
crusso, line 19: fails with assert in region.rs

rts/motoko-rts/src/region.rs
```diff
@@ -132,11 +131,11 @@ impl RegionObject {
     }
 
     pub unsafe fn id(&self) -> RegionId {
-        RegionId((*self.0).id)
+        RegionId(read64((*self.0).id_lower, (*self.0).id_upper))
```
luc-blaeser, line 134:
```suggestion
        RegionId(self.0.read_id64())
```

rts/motoko-rts/src/region.rs
```diff
@@ -388,10 +365,8 @@ unsafe fn alloc_region<M: Memory>(
     let region = r_ptr.get_ptr() as *mut Region;
     (*region).header.tag = TAG_REGION;
     (*region).header.init_forward(r_ptr);
-    debug_assert!(id <= meta_data::max::REGIONS as u32);
-    (*region).id = id as u16;
-    // The padding must be initialized with zero because it is read by the compiler-generated code.
-    (*region).zero_padding = 0;
+    debug_assert!(id <= meta_data::max::REGIONS);
+    write64(&mut (*region).id_lower, &mut (*region).id_upper, id);
```
luc-blaeser, line 369:
```suggestion
    region.write_id64(id);
```

rts/motoko-rts/src/region.rs
```diff
@@ -591,12 +571,32 @@ pub(crate) unsafe fn region_migration_from_no_stable_memory<M: Memory>(mem: &mut
     // Region 0 -- classic API for stable memory, as a dedicated region.
     REGION_0 = region_new(mem);
 
-    assert!((*REGION_0.as_region()).id == 0);
+    assert!((*REGION_0.as_region()).id_lower == 0);
+    assert!((*REGION_0.as_region()).id_upper == 0);
```
luc-blaeser, line 575:
```suggestion
    assert_eq!(REGION_0.as_region().read_id64(), 0);
```

rts/motoko-rts/src/region.rs
```diff
@@ -425,9 +400,9 @@ unsafe fn init_region<M: Memory>(
 }
 
 #[ic_mem_fn]
-pub unsafe fn region_id<M: Memory>(_mem: &mut M, r: Value) -> u32 {
-    let r = r.as_untagged_region();
-    (*r).id.into()
+pub unsafe fn region_id<M: Memory>(_mem: &mut M, r: Value) -> u64 {
+    let r = r.as_region();
+    read64((*r).id_lower, (*r).id_upper)
```
luc-blaeser, line 405:
```suggestion
    r.read_id64()
```

rts/motoko-rts/src/region.rs
```diff
@@ -591,12 +571,32 @@ pub(crate) unsafe fn region_migration_from_no_stable_memory<M: Memory>(mem: &mut
...
-    assert!((*REGION_0.as_region()).id == 0);
+    assert!((*REGION_0.as_region()).id_lower == 0);
+    assert!((*REGION_0.as_region()).id_upper == 0);
 
     // Regions 1 through LAST_RESERVED_REGION_ID, reserved for future use by future Motoko compiler-RTS features.
     region_reserve_id_span(mem, Some(RegionId(1)), RegionId(LAST_RESERVED_REGION_ID));
 }
 
+pub fn block_page_count(rank: u16, block_count: u32, page_count: u32) -> u8 {
+    use meta_data::size::PAGES_IN_BLOCK;
```
luc-blaeser, line 582: Idea (not sure whether really useful): A static check that `PAGES_IN_BLOCK` fits into u8 could prevent unwanted changes of `max::PAGES_IN_BLOCK` and also explain that the downcasts are safe.

rts/motoko-rts/src/region.rs
```diff
@@ -671,19 +671,23 @@ pub(crate) unsafe fn region_migration_from_some_stable_memory<M: Memory>(mem: &m
      * Initialize region0's table entries, for "recovery".
      */
 
-    /* head block is physically last, but logically first. */
-    let head_block_id = region0_blocks - 1; /* invariant: region0_blocks is non-zero. */
-
-    meta_data::block_region_table::set(BlockId(head_block_id as u16), Some((RegionId(0), 0)));
-    meta_data::region_table::set(&RegionId(0), Some(RegionSizeInPages(region0_pages as u64)));
+    /* head block is physically last block, but logically first block in region. */
+    let head_block_id: u16 = (region0_blocks - 1) as u16; /* invariant: region0_blocks is non-zero. */
```
luc-blaeser, line 675: Idea (not very important): Maybe a check that `region0_blocks` fits into u16 would make the downcast look safer. This relies on the assumption that `max::BLOCKS` is never changed to exceed `u16::MAX`.

luc-blaeser (approved):
Looks very good. Not having to recycle region IDs simplifies a lot.

rts/motoko-rts/src/region.rs
crusso: how do I do a static assert in Rust - do we have other examples in the rts you can point me at?

luc-blaeser: I believe an assert statement would be optimized away if it can be statically evaluated.
Explicit static assertions can be written like:
`const _: () = assert!(size_of::<usize>() == size_of::<u32>());`

crusso: Thanks! Done

rts/motoko-rts/src/region.rs
crusso: Added the assert

crusso:
@luc-blaeser PTAL

rts/motoko-rts/src/region.rs
```diff
@@ -491,54 +470,59 @@ pub unsafe fn region_new<M: Memory>(mem: &mut M) -> Value {
...
+                None => {}
+                Some((rid_, _rank, block_page_count)) => {
+                    if &rid_ == rid {
+                        page_count += block_page_count as u32;
+                    }
+                }
+            }
+            block_id += 1;
+        }
+    };
```
luc-blaeser, line 513: This would be quadratic logic if one needs to reconstruct all regions? But I guess this is the theoretical fallback case that we hopefully never need. So, most probably, this is fine.

rts/motoko-rts/src/region.rs
```diff
@@ -491,54 +470,59 @@ pub unsafe fn region_new<M: Memory>(mem: &mut M) -> Value {
...
+            match meta_data::block_region_table::get(BlockId(block_id)) {
+                None => {}
+                Some((rid_, _rank, block_page_count)) => {
+                    if &rid_ == rid {
+                        page_count += block_page_count as u32;
+                    }
+                }
+            }
+            block_id += 1;
+        }
```
luc-blaeser, line 512: A matter of taste, maybe a for loop would be simpler:

```suggestion
        for block_id in 0..tb as u16 {
            match meta_data::block_region_table::get(BlockId(block_id)) {
                None => {}
                Some((rid_, _rank, block_page_count)) => {
                    if &rid_ == rid {
                        page_count += block_page_count as u32;
                    }
                }
            }
        }
```

rts/motoko-rts/src/region.rs
```diff
@@ -315,58 +310,44 @@ mod meta_data {
...
+                    write_u64(region_offset, RegionId::into_u64(None));
+                    write_u16(rank_offset, 0);
+                    write_u8(page_count_offset, 0);
+                }
+                Some((r, j, c)) => {
+                    write_u64(region_offset, RegionId::into_u64(Some(r)));
+                    write_u16(rank_offset, j);
+                    write_u8(page_count_offset, c);
                 }
             }
```
luc-blaeser, line 348: Just cosmetics: Maybe, the write duplicates could be reduced:
(sorry, code suggestion does not work here as it spans code line additions and deletions)
```
            let (region_id, rank, allocated_pages) = match r {
                None => (None, 0, 0),
                Some((r, j, c)) => (Some(r), j, c),
            };
            write_u64(region_offset, RegionId::into_u64(region_id));
            write_u16(rank_offset, rank);
            write_u8(page_count_offset, allocated_pages);
```

rts/motoko-rts/src/region.rs
crusso: Yes, but we only use if for one region at the moment. For several, I think one could stack allocate an array of max::BLOCKS regions, compute the size of any non-empty regions, allocate region objects for the non-empty ones with correctly size blobs and then populate the blobs. Or something more frugal, but still linear time (ish).

rts/motoko-rts/src/region.rs
crusso: I'll do that

PR #4178, opened 2023-08-19
Experiment: add doc comments to generated Candid files

### LINKED ISSUE #3120: As a Motoko developer, I want to be able to annotate my Candid declarations using comments
**Is your feature request related to a problem? Please describe.**
It would be nice to support a syntax for special comments in Motoko that are automatically added to the candid interface, similar to (but not necessarily as full-featured as) [JSDoc](https://en.wikipedia.org/wiki/JSDoc).

**Describe the solution you'd like**
Adding a specially formatted comment to a type or a method of an actor will allow also be reflected in the candid interface. This will be helpful for providing context in auto-generated candid files.

Ideally a similar pattern could be developed for Rust down the line.

**Describe alternatives you've considered**
The alternatives would be to more strongly encourage developers to write and maintain their own candid files, relying less on auto-generation

rvanasa:
One possible approach to passing documentation into Candid interfaces generated from a corresponding Motoko file.

**Progress:**
- [x] Move `Trivia` into `lang_utils`
- [x] Add a trivia table in the representation of Candid programs (same as `prog` in `Mo_def.Syntax`)
- [x] Refactor the Candid pretty printer to use an optional trivia table for doc comments
- [x] Adjust `mo_to_idl` to include a source region for `prog` AST nodes
- [x] Update the Motoko-to-Candid logic and moc.js `candid()` function to explicitly include doc comments
- [x] Expand `trivia.ml` for processing doc comments with concrete syntax information
- [x] Pass Motoko source regions to the corresponding Candid fields in `mo_to_idl.ml`
- [x] Convert Motoko `/** */` block comments to line comments
- [x] Add unit tests (via `moc.js`)

At the moment, this works as expected for doc comments at the top of a Motoko file and for individual methods, which are now included in the generated Candid interface. 

It's also worth noting that Candid will only use `///` line comments for documentation, so `/** */` block comments in Motoko are converted to equivalent line comments. This simplifies language tooling, avoids ambiguity around leading whitespace, and improves consistency when viewing Candid files. 

---

Here's a quick shell command for manually testing via `moc.js`:

```sh
make -C src moc.js && node -e "try { const {Motoko} = require('./src/moc.js'); Motoko.saveFile('result-Main.mo', '/// Line comment\n/** Block comment\n    multi-line */\nimport Prim \"mo:prim\"; actor { /** Field comment */ public func abc() : async Nat { 0 }; }'); console.log(Motoko.candid('result-Main.mo').code) } catch(err) { console.error('Error:', err.message || err) }"
```

Resolves #3120.


crusso:
Nice work, this looks promising! and its nice the way you encapsulated the state to avoid leaks.

I don't think we actually type check the generated idl, so you may be able to just re-use the source regions from the motoko defining occurrences of identifiers, in the idl identifiers. Not sure, and a bit of hack, but might be simplest.

The way Motoko propagates deprecation doc comments (ASFAIK) is to just have a every variant and record type field carry an optional string that is irrelevant to subtyping etc, but is used to carry the warning, if any, to where it is needed. Perhaps it's just a matter of extending the idl record/variant field and actor method types accordingly. That might (eventually) also give you a way to import the trivia of source files back into Motoko types, which could be handy, except that we assume the Motoko info is always a deprecation, not a general comment (IIRC).


The bash script worked for me btw (so I see the method comments are not propagated yet).

rvanasa:
Update: this PR is currently sufficient to unblock @tiagoicp's developer grant.

Input Motoko file:

```motoko
/// Program line comment
/** Block comment
    multi-line */
import Prim "mo:prim";

actor {
    /// Type comment
    public type T = Nat;
    /** Block comment */
    /// Line comment
    public func abc() : async T { 0 };
};
```

Generated Candid file:

```candid
type T = nat;
/// Program line comment
/// Block comment
///     multi-line
service : {
  /// Block comment
  /// Line comment
  abc: () -> (T);
}
```

It would be nice to include doc comments for type definitions; I'm currently thinking of splitting this feature into a separate PR due to requiring another relatively large refactor (and seeming lower priority than the program and method doc comments).

krpeacock:
This resolves #3120

crusso (approved):
LGTM - thanks for pulling on this thread.

PR #4187, opened 2023-09-01
regions: reduce metadata overhead of code using Regions before ESM from 8MB to 1MB (16 pages)

crusso:
In the current design, the first 8MB block of IC stable memory is largely unused to accommodate migration from users of legacy Experimental Stable Memory (migration from ESM moves just the first block of ESM to a higher block to make way for Regions metadata (6 pages), while keeping the other blocks of ESM in-place, requiring blocks to be aligned at 8MB offsets).

This means that code that only uses Regions (or uses them first before using ESM) unnecessarily  wastes the first 8MB block of IC  stable memory instead of just 6 pages.

This PR dynamically sets the base offset for the first allocated block to 1M (for a canister that doesn't yet use ESM) or 8MB (for code that does), reducing the overhead for code that only use regions, or uses regions before ESM.

The base offset (along with fixed block size in pages, in case we change that in future) is recorded in the meta-data, for resumption after upgrade.

We reserve a full 16, not just 6 pages, for future use (e.g. halving block size would double metadata space to 12 pages)

### DIFF
--- a/rts/motoko-rts/src/region.rs
+++ b/rts/motoko-rts/src/region.rs
@@ -245,10 +245,14 @@ mod meta_data {
         pub const BLOCK_IN_BYTES: u64 = PAGE_IN_BYTES * (PAGES_IN_BLOCK as u64);
 
         // Static memory footprint, ignoring any dynamically-allocated pages.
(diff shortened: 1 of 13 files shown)


luc-blaeser (approved):
Very nice saving of stable memory in the case of direct migration to stable regions.

PR #4196, opened 2023-09-06
tweak `Buf::advance`

ggreif:


crusso (commented):
Maybe it would be even better to check for overflow rather than assuming it can't happen?

luc-blaeser:

PR #4207, opened 2023-09-14
experiment: selectively inline share_code helpers

crusso:
to mitigate regression of new cost model, selectively inline share_code helpers in the backend. NB: some recursive ones cannot be unshared/inlined (e.g. seriliazation code and code that explicitly returns rather than returning control flow)

For profiling results ( (sans wasm-opt) against base with new cost-model, see:

https://github.com/dfinity/canister-profiling/pull/80

_Overall Statistics
binary_size: 7.50% [5.67%, 9.33%]
max_mem: no change
cycles: -7.15% [-8.32%, -5.99%]_

Thanks to @chenyan-dfinity for producing a series of these!

~(Note, this includes some Rust benchmarks that are, of course, unchanged. Typical Motoko only improvements are higher)~

Coming soon... wasm-opt numbers.

## aggressive wasm-opt 
https://github.com/dfinity/canister-profiling/pull/82

Overall Statistics
binary_size: 93.01% [67.73%, 118.30%]
max_mem: no change
cycles: -14.23% [-16.95%, -11.50%]

## wasm-opt O4

https://github.com/dfinity/canister-profiling/pull/84

Overall Statistics
binary_size: -10.97% [-12.49%, -9.45%]
max_mem: no change
cycles: -7.97% [-9.17%, -6.78%]

## wasm-opt O3 
https://github.com/dfinity/canister-profiling/pull/83

Overall Statistics
binary_size: -10.50% [-11.82%, -9.18%]
max_mem: no change
cycles: -8.05% [-9.12%, -6.97%]

PR #4208, opened 2023-09-14
experiment: #4207 (with artifacts)

crusso:
Like #4207 but builds artifacts without waiting for tests.


crusso:
superseded by #4212

PR #4209, opened 2023-09-15
Luc/inlining for new metering

crusso:


crusso:
Just opened for a diff

PR #4212, opened 2023-09-16
feat: selectively inline share_code helpers including allocation

crusso:
To mitigate cycle perf regression of new cost model, selectively inline `share_code` helpers in the backend using an additional argument `Never | Always` (i.e. always inline vs never inline). Also, add compiler flags to explicitly opt-in or disable the inlining optimization.

NB: some recursive share_code cannot be unshared/inlined (e.g.  recursive serialization code and code that explicitly returns rather than returning control flow). 

Similar to #4207, but also inlines all heap object allocation and adds compiler flags to enable (default)/disable the optimization.
Note users may want to disable the optimization if they can't accept the increase in code size.

# Profiling data 

## new metering, sans wasm-opt
https://github.com/dfinity/canister-profiling/pull/85

Overall Statistics
binary_size: 10.68% [8.58%, 12.79%]
max_mem: no change
cycles: -8.32% [-9.67%, -6.97%]

## new metering with wasm-opt 03

https://github.com/dfinity/canister-profiling/pull/86

Overall Statistics
binary_size: -6.28% [-7.53%, -5.03%]
max_mem: no change
cycles: -18.21% [-20.07%, -16.36%]

## new metering, master (no-inlining) and wasm-opt 03

https://github.com/dfinity/canister-profiling/pull/83

Overall Statistics
binary_size: -13.96% [-14.64%, -13.28%]
max_mem: no change
cycles: -12.46% [-13.51%, -11.41%]

(UPDATE: revised stats after @chenyan-dfinity updates to PRs)


.github/workflows/test.yml
```diff
@@ -105,7 +105,7 @@ jobs:
 
   artifacts:
     if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository && contains(github.event.pull_request.labels.*.name, 'build_artifacts')
-    needs: tests
+    #needs: tests #nomore
```
crusso, line 108:
```suggestion
    needs: tests
```

crusso:
@chenyan-dfinity any idea why this PR doesn't have a build-artifacts action? I'm stumped.

src/codegen/compile.ml
```diff
@@ -3588,7 +3618,7 @@ module Object = struct
   (* Returns a pointer to the object field (without following the field indirection) *)
   let idx_hash_raw env low_bound =
     let name = Printf.sprintf "obj_idx<%d>" low_bound  in
-    Func.share_code2 env name (("x", I32Type), ("hash", I32Type)) [I32Type] (fun env get_x get_hash ->
+    Func.share_code2 Func.Always env name (("x", I32Type), ("hash", I32Type)) [I32Type] (fun env get_x get_hash ->
```
luc-blaeser, line 3621: Just a note, not for this PR: This is probably a very frequently called function. Even if we would replace `Return` by loop exit, it is maybe too heavy-weight for inlining. A potential optimization for the future could be to use a (perfect) hash table lookup...

src/codegen/compile.ml
```diff
@@ -840,71 +841,100 @@ module Func = struct
...
+    Always (* i.e. never inline *)
+  | Never  (* i.e. always inline *)
 
+  (* (Almost) transparently lift code into a function and call this function,
+     unless sharing = Never and not (!Flags.share_code) in which case the code
+     is inlined.
+     NB: inlined code must not be recursive nor `return`.
+  *)
+  (* Also add a hack to support multiple return values *)
+  let share_code sharing env name params retty mk_body =
```
luc-blaeser, line 854: Nice refactoring, reducing code duplication...

luc-blaeser (approved):
Very useful optimization for reducing the regression gap with the new metering.
I can also remeasure the performance improvement with the GC benchmark.

luc-blaeser:
I measured the following results with the GC benchmark:

Instruction costs including dirty page charging.
Average across benchmark cases.

## Optimization Gain

Unoptimized: New metering without this PR.
Optimized: New metering with this PR.

Metric | Unoptimized | Optimized | Difference
-----|-----|-----|-----
Instructions | 2.46e10 | 2.29e10 | **-7.4%**
Code size | 184 KB | 212 KB | **+15%**

## Metering Comparison

Old metering: Old cost model without this PR.
New optimized: New metering with this PR.

Metric | Old Metering | New Optimzed | Difference
----- | -------------- | ------------------------ | ---------
Instructions | 2.03e10 | 2.29e10 | +13%

src/codegen/compile.ml
crusso: To avoid confusion, Func.Always means always share, never inline. So I think we are in agreement.

src/codegen/compile.ml
```diff
@@ -9292,7 +9322,7 @@ let powInt64_shortcut fast env get_a get_b slow =
 
 (* kernel for Int64 arithmetic, invokes estimator for fast path *)
 let compile_Int64_kernel env name op shortcut =
-  Func.share_code2 env (prim_fun_name Type.Int64 name)
+  Func.share_code2 Func.Always env (prim_fun_name Type.Int64 name)
```
chenyan-dfinity, line 9325: I guess we could also inline these primitives, but they further increase the code size?

chenyan-dfinity (approved):
LGTM!

src/codegen/compile.ml
crusso: Only if used. So yeah, we could inline these and some more probably

crusso: Should I just merge as is and we can revisit later?

chenyan-dfinity: Sure, this is already pretty good.

PR #4216, opened 2023-09-20
feat: implement a simple Rust allocator for transient allocations and access to more of Rust std lib; implement Text.toLowercase/toUppercase via Rust

crusso:
One of the obstacles to using std functions in the rts is the lack of a (dynamic) allocator.

This PR implements a dead simple, POC allocator using the Motoko heap for allocations and uses it to implement
Text.to_lowercase in terms of (std's) String::to_lowercase.

The alloctor must only be used for transient allocations that need not (and cannot) survive the next GC.

- [x] support Layout.align` fully (by padding blobs).
- [x] to_uppercase 
- [x] check binary size change (went from 16K to 17K, so not too bad)
- [x] conditional compilation to compile to native/test target (native test target seem to build fine - perhaps no change needed?)
- [x] base changes https://github.com/dfinity/motoko-base/pull/590

PR #4221, opened 2023-09-25
repro: for strange test failure in run/idl-ops.mo

crusso:
branch of #4216 to keep repro for run/idl-ops.mo failure (the test will made less sensitive in #4216 to enable progress).


luc-blaeser:

PR #4223, opened 2023-09-25
chore: modify CI to free linux disk-space before tests action

crusso:
My linux CI jobs have been failing due to lack of disk space.

I stumbled across this suggestion on the web

https://saturncloud.io/blog/github-action-ecr-optimizing-disk-space/#handling-or-maximizing-github-runner-out-of-disk-space-error

and it seems to help.

Maybe it's enough to just use the GH action https://github.com/jlumbroso/free-disk-space.

Ok, just this gives us back an extra 28GB of disk space. Hopefully that can keep us going for  a while more.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,23 +27,23 @@ jobs:
         name: ic-hs-test
         authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
 
-    - name: clean unncessary files to save space
-      if: matrix.os == 'ubuntu-latest'
-      run: |
-          docker rmi `docker images -q`
-          sudo rm -rf /usr/share/dotnet /etc/mysql /etc/php /etc/sudo apt/sources.list.d
-          sudo apt -y autoremove --purge
-          sudo apt -y autoclean
-          sudo apt clean
-          rm --recursive --force "$AGENT_TOOLSDIRECTORY"
-          df -h
-
       # Free up disk space on Ubuntu
     - name: Free Disk Space (Ubuntu)
       uses: jlumbroso/free-disk-space@main
       if: matrix.os == 'ubuntu-latest'
       with:
-        # This might remove tools that are actually needed, if set to "true" but frees about 6 GB
+        # this might remove tools that are actually needed,
+        # if set to "true" but frees about 6 GB
+        tool-cache: false
+
+        # all of these default to true, but feel free to set to
+        # "false" if necessary for your workflow
+        android: true
+        dotnet: true
+        haskell: true
+        large-packages: true
+        docker-images: true
+        swap-storage: true
         tool-cache: false
         large-packages: true
         swap-storage: true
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -27,6 +27,27 @@ jobs:
...
+
+        # all of these default to true, but feel free to set to
+        # "false" if necessary for your workflow
+        android: true
+        dotnet: true
+        haskell: true
+        large-packages: true
+        docker-images: true
+        swap-storage: true
+        tool-cache: false
```
crusso, line 47:
```suggestion
```

src/codegen/compile.ml
```diff
@@ -8,6 +9,7 @@ instruction lists, as it takes care of (1) source locations and (2) labels.
 This file is split up in a number of modules, purely for namespacing and
 grouping. Every module has a high-level prose comment explaining the concept;
 this keeps documentation close to the code (a lesson learned from Simon PJ).
+
```
crusso, line 12:
```suggestion
```

src/codegen/compile.ml
```diff
@@ -1,4 +1,5 @@
 (*
+
```
crusso, line 2:
```suggestion
```

chenyan-dfinity (approved):

.github/workflows/test.yml
```diff
@@ -27,6 +27,24 @@ jobs:
         name: ic-hs-test
         authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
 
+      # Free up disk space on Ubuntu
+    - name: Free Disk Space (Ubuntu)
+      uses: jlumbroso/free-disk-space@main
```
ggreif, line 32:
```suggestion
      uses: jlumbroso/free-disk-space@v1.2.0
```

ggreif: Actually, this might be a better (maintained?) alternative: https://github.com/marketplace/actions/disk-space-reclaimer

.github/workflows/test.yml
```diff
@@ -27,6 +27,24 @@ jobs:
...
+      if: matrix.os == 'ubuntu-latest'
```
ggreif, line 33:
```suggestion
      if: startsWith(matrix.os, 'ubuntu-')
```
just a nit, in case we want other versions in the matrix sometime...

ggreif (commented):
WTF! Why doesn't GitHub add a warning to CI (when a build fails because of tight space) that this action exists?

Thanks for finding this, this is genius! I stumbled over an alternative that claims to be better maintained, though.

.github/workflows/test.yml
crusso, line 32:
```suggestion
      uses: insightsengineering/disk-space-reclaimer@v1
```

crusso: Ok, I took the maintained one instead (just a fork)

Issue #4229, opened 2023-09-29
bug: strict candid validation should reject IDL_Ext_Region


crusso:
But might not due to current definition of `is_primitive_typ` in idl.sub. It should probably take an `extended` argument and act accordingly.

Possibly, but unlikely, a security hole since we need to validate untrusted candid before decoding.

PR #4230, opened 2023-09-29
bugfix: fixes issue #4229

crusso:
The regions PR added a new (extended) Candid type, IDL_EXT_REGION, that is only valid in extended Candid type tables.

This PR fixes the type table validation logic to properly accept/reject IDL_EXT_REGION occurrences accordingly.

(side hustle: don't delete Linux swap files on CI builds - seems to have been a source of non-(or-delayed)-termination in CI, leading to timeouts)


rts/motoko-rts/src/idl.rs
```diff
@@ -41,8 +41,11 @@ const IDL_CON_func: i32 = -22;
 const IDL_CON_service: i32 = -23;
 
 const IDL_REF_principal: i32 = -24;
+
+// extended Candid only
 const IDL_EXT_region: i32 = -128;
 
+// extended Candid only
```
crusso, line 48:
```suggestion
// Extended Candid only
const IDL_EXT_region: i32 = -128;

// Extended Candid only
```

rts/motoko-rts/src/lib.rs
```diff
@@ -74,6 +74,7 @@ extern "C" {
     fn rts_trap(msg: *const u8, len: Bytes<u32>) -> !;
 }
 
+#[no_mangle]
```
luc-blaeser, line 77: I assume this is only needed for test outputs?

luc-blaeser (approved):
The fix looks good. 
Maybe we need to distinguish the test output between debug and release mode, because it seems that `issue-4229.mo` has a different stack trace depending on the build configuration. Or alternatively, we could ignore the stack trace details in the comparison (except maybe the top-level function).

PR #4231, opened 2023-09-29
expose print_deps in moc.js

chenyan-dfinity:
so that we can collect usage about base library in playground.


crusso (approved):
(rubberstamping)

PR #4233, opened 2023-10-03
RTS extension for main memory-to-region snapshot.

matthewhammer:
Experimental feature: MainMemory-to-Region Snapshots

Gives all Motoko canisters an automatic (compiler-generated) way to take snapshots of their own memories and store these into regions for further analysis, and data recovery.

- [x] extend RTS with some candidate logic for walking over heap and writing into a region.
- [x] expose as a temporary primitive for tests
- [x] use test primitive to see how large of a memory we can snapshot without running out of cycles.
  - [x] Easy to take many snapshots of a small heap.
  - [x] Created a very large main memory (3.8Gb) and was able to snapshot it in a single message, into a region, within the 4GB default for max stable memory.


rts/motoko-rts/src/region.rs
```diff
@@ -948,6 +948,19 @@ pub(crate) unsafe fn region_load<M: Memory>(_mem: &mut M, r: Value, offset: u64,
...
+pub unsafe fn region_write_heap_shapshot<M: Memory>(mem: &mut M, r: Value) {
+    use crate::constants::WASM_PAGE_SIZE;
+    use core::arch::wasm32;
+
+    let heap_size: usize = wasm32::memory_size(0) * (WASM_PAGE_SIZE.as_u32() as usize);
+    if wasm32::memory_size(0) as u64 > region_size(mem, r) {
+        let new_pages = wasm32::memory_size(0) as u64 - region_size(mem, r);
+        region_grow(mem, r, new_pages);
+    };
+    let heap_src = core::slice::from_raw_parts(core::ptr::null(), heap_size);
```
crusso, line 960: I found that slices can't be big enough to copy up to 4GB of data (that's why Blob's are read/written in two halves). I think the size argument is signed, so at most 2^31-1 (Int32::MAX). 

Not sure, but eventually, we may have to hard-code this function in Wasm to avoid having a rust frame for the function on the Rust shadow stack.

rts/motoko-rts/src/region.rs
```diff
@@ -948,6 +948,32 @@ pub(crate) unsafe fn region_load<M: Memory>(_mem: &mut M, r: Value, offset: u64,
     }
 }
 
+pub unsafe fn region_write_heap_shapshot<M: Memory>(mem: &mut M, r: Value) {
```
luc-blaeser, line 951: This looks very good. If the the Rust function `from_raw_parts` would not artificially limit the size to `isize::MAX`, it would be even simpler...

rts/motoko-rts/src/region.rs
```diff
@@ -948,6 +948,32 @@ pub(crate) unsafe fn region_load<M: Memory>(_mem: &mut M, r: Value, offset: u64,
...
+    use crate::constants::WASM_PAGE_SIZE;
+    use core::arch::wasm32;
+
+    let heap_size: usize = wasm32::memory_size(0) * (WASM_PAGE_SIZE.as_u32() as usize);
```
luc-blaeser, line 955: The variable could maybe be named `memory_size` as it covers the entire allocated memory (including unused space beyond the heap end). Perhaps, the function itself could be named `memory_snapshot` or `main_memory_snapshot`. Alternatively, we could also only focus on the heap (static and dynamic heap).

rts/motoko-rts/src/region.rs
```diff
@@ -948,6 +948,32 @@ pub(crate) unsafe fn region_load<M: Memory>(_mem: &mut M, r: Value, offset: u64,
...
+    if wasm32::memory_size(0) as u64 > region_size(mem, r) {
+        let new_pages = wasm32::memory_size(0) as u64 - region_size(mem, r);
+        region_grow(mem, r, new_pages);
+    };
+    if heap_size < isize::MAX as usize {
```
luc-blaeser, line 960: Micro detail: Maybe a short comment could be helpful saying why this special case handling is needed (limitation of `from_raw_parts`).

matthewhammer: Sounds good.  I'll also add the same comment to the two other existing instances of this pattern (store/load Blob).

rts/motoko-rts/src/region.rs
matthewhammer: 💯

matthewhammer:
I, matthewhammer, hereby agree to the DFINITY Foundation's CLA.

luc-blaeser:

PR #4237, opened 2023-10-05
Updated Documentation to include Mops/Vessel info

jessiemongeon1:
This PR adds information about Mops/Vessel to the Modules and Imports documentation page.


doc/md/modules-and-imports.md
```diff
@@ -63,6 +63,37 @@ You can define dependencies for a project using the Vessel package manager or in
...
+    "build": {
+      "packtool": "mops sources"
+    }
+  }
+}
+```
+
+For Vessel, use `vessel sources`. 
+
+Then, download a package with the `mops` or `vessel` CLI tools, such as:
```
rvanasa, line 84: It might be worth mentioning that for Vessel, it's necessary to edit the `vessel.dhall` file (since there isn't an equivalent to `mops add`). Perhaps we could even just use the same instructions as in [this PR](https://github.com/dfinity/portal/pull/1975/files#diff-2ffe11fc134e5fb5877533a43e097a13b71fd0fdcd974a0e4ab456529b2387e8R11-R51) if that saves any time.

PR #4238, opened 2023-10-08
bugfix: fix for unexpected variant decoding issue

crusso:
Repro on playground (from forum bug report: https://forum.dfinity.org/t/icdevs-org-voting-incident-report-and-stable-api-discussion/23510/10?u=claudio).

Calling `say("hello")` traps with "byte not 0 or 1" after decoding an unexpected `Action` variant with a payload but should just default to `null` due to an enclosing option.

I think the moc idl-decoder is failing to skip the payload arg before defaulting to null (but am not super confident about my fix).

Uploading the re-compiled wasm fixes the trap.

- [x] add smaller repro to be sure
- [x] check for other issues (can this happen with records for example)?

@nomeata does the fix smell ok?


src/codegen/compile.ml
```diff
@@ -7194,7 +7194,8 @@ module MakeSerialization (Strm : Stream) = struct
                 continue
             )
             ( sort_by_hash vs )
-            ( coercion_failed "IDL error: unexpected variant tag" )
+            ( skip get_arg_typ ^^ (* HERE *)
+              coercion_failed "IDL error: unexpected variant tag" )
```
nomeata, line 7198: Well spotted. The fix looks reasonable.

nomeata: According to git blame, I made that mistake in https://github.com/dfinity/motoko/commit/5388dc3af52b16c2b34440f64f8fd221cda060d5 and it hasn't surfaced in three years. Oh dear.

crusso: Great, thanks for checking! I'll try to produce a smaller test case

nomeata: Ideally include some tests in the official Candid test suite, so that other implementations don't make that mistake (although I believe all others first parse and then coerce, and thus are less prone to this bug).

An alternative implementation would be to return quickly, and when backtracking in `opt` reset the data and type but pointers and skip the whole thing in one go. Might be more reliable and simpler to implement, especially for records.

crusso: Yeah, I was wondering about that but don't trust myself to rewrite this code. Is the current invariant that the recursive call needs to finish parsing the entire value, leaving the buffer and data pointers just past the end of the value?

nomeata: Yes, and I think there is a comment to that end somewhere.

src/codegen/compile.ml
```diff
@@ -6693,7 +6693,7 @@ module MakeSerialization (Strm : Stream) = struct
           begin code0
           end begin
             get_b ^^ compile_eq_const 1l ^^
-            E.else_trap_with env "IDL error: byte tag not 0 or 1" ^^
+            E.else_trap_with env ("IDL error: byte tag not 0 or 1") ^^
```
crusso, line 6696:
```suggestion
            E.else_trap_with env "IDL error: byte tag not 0 or 1" ^^ 
```

src/codegen/compile.ml
```diff
@@ -6693,7 +6693,7 @@ module MakeSerialization (Strm : Stream) = struct
           begin code0
           end begin
             get_b ^^ compile_eq_const 1l ^^
-            E.else_trap_with env "IDL error: byte tag not 0 or 1" ^^
+            E.else_trap_with env "IDL error: byte tag not 0 or 1" ^^
```
crusso, line 6696:
```suggestion
            E.else_trap_with env "IDL error: byte tag not 0 or 1" ^^
```

nomeata (approved):

PR #4240, opened 2023-10-09
bugfix: fix candid decoding at immutable array types to support opt defaulting

crusso:
Another Candid decoding bug, seems independent of #4238. 

I think the code for decoding immutable arrays was never extended to handle backtracking correctly.
When recovery is an option (haha) and the type descriptor does not match, the current code would continue to try to decode the array length and elements. When recovery is not an option, the previous trap would render this code unreachable. But when recovery is enabled, we need to bail and not try to read the array size and elements but just return the sentinel value instead.

- [ ] Consider applying the same refactoring to mutable array, if possible. 
      Since these aren't Candid, but extended Candid (for stable variables), and don't need to support recovery, the change is (probably) not necessary but nice for uniformity only.
- [ ] restore or delete original repro - I must have edited it while experimenting


test/run/idl-decoding-bug.mo
```diff
@@ -0,0 +1,17 @@
...
+    {a=?#a} */
+  ], 0) :
+  (?[{a:?{#a;#b:{};#c}}], Nat));
+
+let o = (from_candid b) : ?([{a:?{#a}}],Nat);
+//assert(o == ?[?#a, null, null, ?#a, null, ?#a]);
+
+//SKIP run
+//SKIP run-ir
+//SKIP run-low
```
crusso, line 17:
```suggestion
//SKIP run-low

```

test/run/idl-opt-tests.mo
```diff
@@ -0,0 +1,78 @@
...
+    ?[{#b:Int}]
+  );
+  let o = (from_candid b) :?[{#b:Nat}];
+  P.debugPrint(debug_show(o));
+  assert o == null;
+}; // traps due to broken back-tracking in array decoding
+
+//SKIP run
+//SKIP run-ir
+//SKIP run-low
```
crusso, line 78:
```suggestion
//SKIP run-low

```

src/codegen/compile.ml
```diff
@@ -7102,7 +7104,9 @@ module MakeSerialization (Strm : Stream) = struct
         let (set_x, get_x) = new_local env "x" in
         let (set_val, get_val) = new_local env "val" in
         let (set_arg_typ, get_arg_typ) = new_local env "arg_typ" in
-        with_composite_typ idl_vec (ReadBuf.read_sleb128 env) ^^ set_arg_typ ^^
+        with_composite_typ idl_vec (fun get_typ_buf ->
+        ReadBuf.read_sleb128 env get_typ_buf ^^
+        set_arg_typ ^^
```
nomeata, line 7109: Looks reasonable. Maybe indent the code now inside the `(`?

It seems that array is using `remember_failure`, so the full array is deserialized, even if there are failures. That’s good.

src/codegen/compile.ml
```diff
@@ -7102,7 +7104,9 @@ module MakeSerialization (Strm : Stream) = struct
...
         ReadBuf.read_leb128 env get_data_buf ^^ set_len ^^
         get_len ^^ Arr.alloc env ^^ set_x ^^
         get_len ^^ from_0_to_n env (fun get_i ->
```
crusso, line 7112:
```suggestion
          ReadBuf.read_sleb128 env get_typ_buf ^^
          set_arg_typ ^^
          ReadBuf.read_leb128 env get_data_buf ^^ set_len ^^
          get_len ^^ Arr.alloc env ^^ set_x ^^
          get_len ^^ from_0_to_n env (fun get_i ->
```

src/codegen/compile.ml
crusso: (now indented)

src/codegen/compile.ml
```diff
@@ -7062,6 +7062,8 @@ module MakeSerialization (Strm : Stream) = struct
           let (set_x, get_x) = new_local env "x" in
           let (set_val, get_val) = new_local env "val" in
           let (set_arg_typ, get_arg_typ) = new_local env "arg_typ" in
+          (* TODO: if possible refactor to match new Array t code,
+             (though probably unnecessary for extended candid due to lack of fancy opt subtyping) *)
```
crusso, line 7066:
```suggestion
             (perhaps unnecessary for extended candid due to lack of fancy opt subtyping) *)
```
stupid edit to trigger CI

src/codegen/compile.ml
```diff
@@ -6693,7 +6693,7 @@ module MakeSerialization (Strm : Stream) = struct
           begin code0
           end begin
             get_b ^^ compile_eq_const 1l ^^
-            E.else_trap_with env "IDL error: byte tag not 0 or 1" ^^
+            E.else_trap_with env ("IDL error: byte tag not 0 or 1") ^^
```
crusso, line 6696:
```suggestion
            E.else_trap_with env "IDL error: byte tag not 0 or 1" ^^
```

ggreif (approved):
Rubber-stamping!

PR #4243, opened 2023-10-11
experiment: refactor mutable array deserialization for uniformity with immutable arrays.

crusso:
Less sure of this one...


nomeata (commented):
Also unsure. With aliasing, is there a risk that one fails and the other doesn't?

Or rather, since these are invariant, could there ever be a subtyping issue? So maybe the difference between the two implementations isn't observable?

crusso:
> Also unsure. With aliasing, is there a risk that one fails and the other doesn't?
> 
> Or rather, since these are invariant, could there ever be a subtyping issue? So maybe the difference between the two implementations isn't observable?



> Also unsure. With aliasing, is there a risk that one fails and the other doesn't?
> 
> Or rather, since these are invariant, could there ever be a subtyping issue? So maybe the difference between the two implementations isn't observable?

I don't think we can assume that the types are related - instead we are verifying this as we go. So maybe someone could try to  upgrade
from
stable let x = ?[var Int] 
stable let x = ?[var Nat]
even-though the types aren't (Motoko) compatible.
I guess we could argue the behaviour is undefined and a random trap or anything else is ok (if we are lucky enough to trap).

I'm actually half tempted to refine deserialization so that it does regular option typing for extended candid and traps on mutable stuff in ordinary candid, so defaulting never applies, but even then I think we should probably check the contents are (Motoko)-equal types before descending. 

We don't support efficient equality checking at the moment, but Luc has implemented such a beast elsewhere.

Does the change look dangerous? I guess its better to be conservative until someone reports an actual bug.

crusso:
Viewed as too risky a change.

PR #4256, opened 2023-10-18
experiment: emulate stable memory in WASIMode and WasmMode (to support testing of stable data structures on non-IC targets)

crusso:
This PR exploits wasmtime's support for multiple memory and bulk memory instructions to emulate the IC stable memory API (i.e. ExperimentalStableMemory.mo and Regions.mo).

The basic idea is to introduce an extra Wasm memory and emulate the low-level  `ic0.stable64_grow/size/read/write` operations, using bulk MemoryCopy instructions for data transfer.

The goal is to allow users to easily unit-test stable data structures without having to resort to the full IC environment.
For example, the third-party StableBTree implementation went to a lot of trouble and was less efficient as a result, just to support testing outside the IC.

@rvanasa would this work with your VSCode unit testing support? It does require enabling some additional features on wasmtime (no idea about compatibility with wasmer, if we still use that).

The current implementation only adds a 32-bit memory, but it should be possible to add a 64-bit one instead. I just want to get this working first. 32-bit might be all we can support in browsers anyway (not sure).

- [x] enable wasmtime flags in base repo so tests compile (this is one reason CI is failing). Alternatively, only add the stable memory if it's used.
- [ ] PR to mops to enable additional wasmtime flags


Building.md
```diff
@@ -112,7 +112,6 @@ branch to the `next-moc` branch.
 Make a PR off of that branch and merge it using a _normal merge_ (not
 squash merge) once CI passes. It will eventually be imported into this
 repo by a scheduled `niv-updater-action`.
-
 Finally tag the base release (so the documentation interpreter can do the right thing):
```
crusso, line 115:
```suggestion

Finally tag the base release (so the documentation interpreter can do the right thing):
```

src/wasm-exts/customModuleEncode.ml
```diff
@@ -471,6 +471,12 @@ let encode (em : extended_module) =
       | MemorySize -> op 0x3f; u8 0x00
       | MemoryGrow -> op 0x40; u8 0x00
 
+      | StableSize -> op 0x3f; u8 0x01
+      | StableGrow -> op 0x40; u8 0x01
+      | StableRead -> op 0xfc; vu32 0x0al; u8 0x00; u8 0x01
+      | StableWrite -> op 0xfc; vu32 0x0al; u8 0x01; u8 0x00
+
+
```
crusso, line 479:
```suggestion
```

src/codegen/compile.ml
```diff
@@ -4630,7 +4630,7 @@ module IC = struct
         | _  -> "mem"
       );
       edesc = nr (MemoryExport (nr 0l))
-    });
+      });
```
crusso, line 4633:
```suggestion
    });
```

rvanasa:
> @rvanasa would this work with your VSCode unit testing support? It does require enabling some additional features on wasmtime (no idea about compatibility with wasmer, if we still use that).

 It looks like Wasmer supports the newly included flags (based on [this documentation](https://docs.rs/wasmer/latest/wasmer/struct.Features.html)):

We can also enable NaN canonicalization ([documentation](https://docs.rs/wasmer-compiler-cranelift/latest/wasmer_compiler_cranelift/struct.Cranelift.html)) if that's important.

I'll make sure this works with the extension, either by enabling the multi-memory / bulk memory flags or replacing Wasmer with Wasmtime if that ends up being necessary.

crusso:
> > @rvanasa would this work with your VSCode unit testing support? It does require enabling some additional features on wasmtime (no idea about compatibility with wasmer, if we still use that).
> 
> It looks like Wasmer supports the newly included flags (based on [this documentation](https://docs.rs/wasmer/latest/wasmer/struct.Features.html)).

I just tried the latest wasmer 4.2.2 and it seems to support bulk-memory, but not multi-memory, even with `wasmer compile --enable-all foo.wasm`.

I also couldn't figure out how to enable these features for Node - perhaps one has to pass a V8 option. As a result, I had to slightly cripple our small moc-js tests to omit compilation of the binary.

I'm now worried that requiring all these flags to run moc code will decrease, not increase, DX. 

Perhaps I should only  add the extra memory when any stable memory ops are used, and emit a warning to enable the feature on the corresponding runtime.

Will mull it over, but advice appreciated.

crusso:
> > I just tried the latest wasmer 4.2.2 and it seems to support bulk-memory, but not multi-memory, even with wasmer compile --enable-all foo.wasm.
> 
> Just saw this (apologies for the delay in responding). Yeah, it's unfortunate that this doesn't work for Wasmer, and that's also a good point about Node / `moc.js`.
> 
> > Perhaps I should only add the extra memory when any stable memory ops are used, and emit a warning to enable the feature on the corresponding runtime.
> 
> This seems like a reasonable temporary solution, and we can see whether it's worth doing something more complex based on how many people encounter this limitation in practice.
> 
> Since `mo-dev` and `mops test` both use Wasmtime, I could change the VS Code extension to prioritize Wasmtime when available (and show a notification with installation instructions when relevant). In this case, it could make sense to include a custom section (or some other hint) so that tooling can efficiently check whether it's possible to use a generic Wasm runtime. This would of course be less ideal than somehow being able to support all the current runtimes, but we could make it work and revisit based on feedback.

Ok, I'll try to do that. Instead of adding a custom section, it might just be enough to detect if the wasm binary declares more than 1 memory. But perhaps that's awkward.

rvanasa:
That should work!

src/wasm-exts/ast.ml
```diff
@@ -110,6 +110,10 @@ and instr' =
   | Store of storeop                  (* write memory at address *)
   | MemorySize                        (* size of linear memory *)
   | MemoryGrow                        (* grow linear memory *)
+  | StableSize                        (* size of stable memory *)
```
luc-blaeser, line 113: Maybe a short comment could be helpful, indicating that these are custom instructions (different to Wasm spec/proposal).

src/exes/candid_tests.ml
```diff
@@ -254,7 +255,7 @@ let () =
             match run_cmd "moc -Werror -wasi-system-api tmp.mo -o tmp.wasm" with
             | ((Fail | Timeout), stdout, stderr) -> CantCompile (stdout, stderr, src)
             | (Ok, _, _) ->
-              match must_not_trap, run_cmd "timeout 10s wasmtime --disable-cache tmp.wasm" with
+              match must_not_trap, run_cmd ("timeout 10s wasmtime "^ _WASMTIME_OPTIONS_ ^" tmp.wasm") with
```
luc-blaeser, line 258: Nice, seems that our current version misses the `--enable-cranelift-nan-canonicalization`.

src/codegen/compile.ml
```diff
@@ -5059,6 +5059,79 @@ end (* Cycles *)
 *)
 module StableMem = struct
 
+
+  let conv32 env get64 =
```
luc-blaeser, line 5063: Maybe `conv_u32` (unsigned)?

test/random/Embedder.hs
```diff
@@ -36,7 +36,7 @@ addCompilerArgs (WasmTime _) = ("-wasi-system-api" :)
 addCompilerArgs Drun = id
 
 addEmbedderArgs Reference = id
-addEmbedderArgs (WasmTime _) = ("--disable-cache" :)
+addEmbedderArgs (WasmTime _) = ("--disable-cache" :) . ("--enable-cranelift-nan-canonicalization" :) . ("--wasm-features" :) . ("multi-memory,bulk-memory" :)
```
luc-blaeser, line 39: Also here, seems that I have forgotten to add the nan-canicalization flag in our current code base...

test/run/stable-mem/StableMemory.mo
```diff
@@ -0,0 +1,39 @@
+import Prim "mo:⛔";
+
+module {
+
+  public let size = Prim.stableMemorySize;
+  //public let region = Prim.stableMemoryRegion;
```
luc-blaeser, line 6:
```suggestion
```

test/run/stable-memory.mo
```diff
@@ -0,0 +1,30 @@
...
+let 0 = Prim.stableMemoryGrow(16);
+let 16 = Prim.stableMemorySize();
+Prim.debugPrint(debug_show {size = Prim.stableMemorySize()});
+let 0 = Prim.stableMemoryLoadNat8(0);
+Prim.debugPrint(debug_show {read = Prim.stableMemoryLoadNat8(0)});
+Prim.stableMemoryStoreNat8(0,66);
+Prim.debugPrint(debug_show {read = Prim.stableMemoryLoadNat8(0)});
+let 66 = Prim.stableMemoryLoadNat8(0);
+Prim.debugPrint(debug_show {read = Prim.stableMemoryLoadNat8(0)});
+
```
luc-blaeser, line 13: Maybe testing all basic read/write access functions could be useful (to be on the safe side)

luc-blaeser (approved):
This is very nice PR, enabling way more local testing. It even unifies and simplifies a lot of code...

src/codegen/compile.ml
```diff
@@ -11972,7 +12013,15 @@ and conclude_module env set_serialization_globals start_fi_o =
...
 
-  let memories = [nr {mtype = MemoryType {min = E.mem_size env; max = None}} ] in
+  let memories =
+    match E.mode env with
+    | Flags.ICMode
+    | Flags.RefMode ->
+       [nr {mtype = MemoryType {min = E.mem_size env; max = None}} ]
+    | Flags.WASIMode | Flags.WasmMode ->
+       [nr {mtype = MemoryType {min = E.mem_size env; max = None}};
+        nr {mtype = MemoryType {min = Int32.zero; max = None}} ] (* faux stable memory *)
```
luc-blaeser, line 12023:
```suggestion
    nr {mtype = MemoryType {min = E.mem_size env; max = None}}
    ::
    match mode env with
    | Flags.WASIMode | Flags.WasmMode ->
      [ nr {mtype = MemoryType {min = Int32.zero; max = None}} ] (* faux stable memory *)
    | _ -> []
```

luc-blaeser: Mini refactoring analogous to the other PR that builds upon this...

src/wasm-exts/customModuleEncode.ml
```diff
@@ -471,6 +471,11 @@ let encode (em : extended_module) =
       | MemorySize -> op 0x3f; u8 0x00
       | MemoryGrow -> op 0x40; u8 0x00
 
+      | StableSize -> op 0x3f; u8 0x01
```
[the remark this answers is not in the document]
crusso: Added

src/codegen/compile.ml
crusso: Done by merge

src/codegen/compile.ml
crusso: Done

test/random/Embedder.hs
crusso: Yep, but I don't know if we even have tests for floats...

test/run/stable-memory.mo
crusso: Indeed it was...

PR #4259, opened 2023-10-26
experiment: selectively emulate stable memory in wasi/wasm mode

crusso:
builds on #4256 (emulating stable memory on wasmtime) but:

* avoids adding second memory unless necessary (so most programs continue to run w/o extra wasmtime option)
* adds optional custom section wasm_features with comma separate string of extra wasmtime flags (so that  driver could discover the options to pass)
* emits a warning when compiling code that requires extra wasmtime flags.

However, @Luc points out wasmtime should already be run with flag --enable-cranelift-nan-canonicalization and our users probably aren't doing that...

Perhaps we should just specify the options to use via documentation rather than the complicated mechanism here:

@luc-blaeser @rvanasa  thoughts?

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -328,7 +328,7 @@ module E = struct
 
     features : FeatureSet.t ref; (* Wasm features using wasmtime naming *)
 
-    (* StableMemory *)
+    (* requires stable memory, and emulation on wasm targets *)
     requires_stable_memory : bool ref;
   }
 
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -261,6 +261,7 @@ module E = struct
   module NameEnv = Env.Make(String)
   module StringEnv = Env.Make(String)
   module LabSet = Set.Make(String)
+  module FeatureSet = Set.Make(String)
 
   module FunEnv = Env.Make(Int32)
   type local_names = (int32 * string) list (* For the debug section: Names of locals *)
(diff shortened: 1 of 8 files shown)


src/codegen/compile.ml
```diff
@@ -324,6 +325,11 @@ module E = struct
     (* Mutable *)
     locals : value_type list ref; (* Types of locals *)
     local_names : (int32 * string) list ref; (* Names of locals *)
+
+    features : FeatureSet.t ref; (* Wasm features using wasmtime naming *)
+
+    (* StableMemory *)
```
crusso, line 331:
```suggestion
    (* requires stable memory, and emulation on wasm targets *)
```

src/codegen/compile.ml
```diff
@@ -324,6 +325,11 @@ module E = struct
     (* Mutable *)
     locals : value_type list ref; (* Types of locals *)
     local_names : (int32 * string) list ref; (* Names of locals *)
+
+    features : FeatureSet.t ref; (* Wasm features using wasmtime naming *)
+
+    (* requires stable memory, and emulation on wasm targets *)
```
crusso, line 331:
```suggestion
    (* requires stable memory (and emulation on wasm targets) *)
```

luc-blaeser:
This is a nice implementation, with still quite modest code complexity for this specific handling.
I think (but this is completely subjective and I may be wrong as what are the developer's expectations for this) that simply requiring these flags for `wasmtime` execution would be sufficient. Maybe, we could provide a wrapper or some instruction on how to launch it locally, e.g. that developers also use Nan-canonicalization.

Issue #4267, opened 2023-11-05
Bug: Principal.fromBlob accepts blobs larger than 29 bytes


crusso:
This is problematic because candid principal deserialization rejects principals with more than 29 bytes, leading to failed upgrades.

We should either prevent creation of large principles, a (minor?) breaking change, or, perhaps, allow larger principals for stable variable deserialization and (maybe also) Candid deserialization.

See

https://forum.dfinity.org/t/when-upgrading-code-or-deployment-an-error-message-is-reported-and-the-error-message-is-vague/24392/15

(Candid does not actually limit principal sizes but the IC does)

nomeata:
I'd say that the “principal” in candid refers to the principal as defined by the public spec, and that has an explicit size limit. Would suggest to limit it everywhere, including the Candid spec.

timohanke:
I would also suggest to limit the length everywhere. I would then expose the textual encoding separately as a different module. If people have something that is not a Principal to encode or decode, including something longer, then they can use that module instead of abusing a Principal.fromText/toBlob combination or the reverse.

crusso:
Interestingly, we discussed this in the PR that added principalOfBlob...
https://github.com/dfinity/motoko/pull/3050

nomeata:
Oh, good that I am not completely contradicting myself :-D

crusso:
> I would also suggest to limit the length everywhere. I would then expose the textual encoding separately as a different module. If people have something that is not a Principal to encode or decode, including something longer, then they can use that module instead of abusing Principal.fromText/toBlob in combination or the reverse of that.

Is that actually useful, more generally?

PR #4268, opened 2023-11-06
bugfix: repro and fix for issue #4267 (reject too long principals in Principal.fromBlob and actor literals)

crusso:
Fixes issues #4267:

  * BREAKING CHANGE (Minor): values of type `Principal` are now constrained to contain
    at most 29 bytes, matching the IC's notion of principal (#4268).

    In particular:

    * An actor `import` will be statically rejected if the binary representation of the (aliased) textually encoded
    principal contains strictly more than 29 bytes.

    * `Principal.fromBlob(b)` will trap if `b` contains strictly more than 29 bytes.

    *  The actor literal, `actor <exp>`, will trap if the binary representation of
    of the textually encoded principal `<exp>` contains strictly more than 29 bytes.


test/run-drun/bad-actor-import.mo
```diff
@@ -0,0 +1,5 @@
+//MOC-FLAG --actor-idl bad-actor-import
+import c "ic:yvtf6-waaae-bagba-faydq-qcikb-mga2d-qpcai-reeyu-culbo-gazdi-nryhi";
+actor {
+
+}
```
crusso, line 5:
```suggestion
}

```

test/run-drun/bad-actor-import/yvtf6-waaae-bagba-faydq-qcikb-mga2d-qpcai-reeyu-culbo-gazdi-nryhi.did
```diff
@@ -0,0 +1,3 @@
+service a : {
+
+}
```
crusso, line 3:
```suggestion
}

```

crusso:
Part of me thinks it would have been cleaner just to remove the candid check and rely on the IC checking principal lengths... that would avoid changing the candid spec too....

nomeata (approved):
LGTM

PR #4270, opened 2023-11-08
bugfix: repro and fix for unimplemented region ops

crusso:
Looks like some region operations were NYI.

This PR fixes the implementation of load/store for `Int8`, `Int16` and `Float`, adding a test of all operations.

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -10656,64 +10656,34 @@ and compile_prim_invocation (env : E.t) ae p es at =
     compile_exp_as env ae SR.Vanilla e2 ^^
     Region.store_blob env
 
-  | OtherPrim ("regionLoadNat8"), [e0; e1] ->
+  | OtherPrim (("regionLoadNat8" | "regionLoadInt8" as p)), [e0; e1] ->
     SR.Vanilla,
     compile_exp_as env ae SR.Vanilla e0 ^^
     compile_exp_as env ae SR.UnboxedWord64 e1 ^^
     Region.load_word8 env ^^
-    TaggedSmallWord.msb_adjust Type.Nat8
+    TaggedSmallWord.msb_adjust Type.(if p = "regionLoadNat8" then Nat8 else Int8)
 
-  | OtherPrim ("regionLoadInt8"), [e0; e1] ->
-    SR.Vanilla,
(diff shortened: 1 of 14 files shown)


test/run-drun/stable-memory-test.mo
```diff
@@ -0,0 +1,211 @@
+import Prim "mo:⛔";
+import StableMemory "stable-mem/StableMemory";
+
+let 0 = StableMemory.size();
+let 0 = StableMemory.grow(64);
```
crusso, line 5:
```suggestion
let 0 = StableMemory.grow(4);
```

test/run-drun/region-test.mo
```diff
@@ -0,0 +1,212 @@
+import Prim "mo:⛔";
+import Region "stable-region/Region";
+
+let r = Region.new();
+let 0 = Region.size(r);
+let 0 = Region.grow(r, 64);
```
crusso, line 6:
```suggestion
let 0 = Region.grow(r, 4);
```

src/codegen/compile.ml
```diff
@@ -10663,6 +10663,13 @@ and compile_prim_invocation (env : E.t) ae p es at =
     Region.load_word8 env ^^
     TaggedSmallWord.msb_adjust Type.Nat8
 
+  | OtherPrim ("regionLoadInt8"), [e0; e1] ->
```
ggreif, line 10666: why not in line 10659:
``` ocaml
  OtherPrim (("regionLoadNat8" | "regionLoadInt8") as prim), [e0; e1] ->
```
and in line 10664
``` ocaml
    TaggedSmallWord.msb_adjust Type.(if prim = "regionLoadNat8" then Nat8 else Int8)
```

ggreif: and similarly below ...

ggreif (approved):
Rubber-stamping, but see my previous comments.

src/codegen/compile.ml
crusso: Sure, that works too, just didn't like the repeated test but maybe that's better than repeated code.

crusso: Done (also for existing stable memory ops)

PR #4271, opened 2023-11-09
fix: don't coalesce tag with underscore

ggreif:
... in the coverage warnings.

Repro:
``` Motoko
type A = { #a : { f : Int; g : Text }; #b : (Char, Char); #c : ?Nat; #d : Nat };

var a : A = #d 42;
switch a {case (#d 1) ()};
```
Gives
> stdin:13.1-13.26: warning [M0145], this switch of type
>   {#a : {f : Int; g : Text}; #b : (Char, Char); #c : ?Nat; #d : Nat}
> does not cover value
>   #d(0 or 2 or _) or
>   #a(_) or **#b_** or #c(_)

### DIFF
--- a/src/mo_frontend/coverage.ml
+++ b/src/mo_frontend/coverage.ml
@@ -140,9 +140,8 @@ let rec string_of_desc t = function
     "?(" ^ string_of_desc t' desc ^ ")"
   | Tag (desc, l) ->
     let t' = T.lookup_val_field l (T.as_variant_sub l t) in
-    let sep = function "_" -> " _" | s -> s in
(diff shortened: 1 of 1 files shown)


crusso:
And reason not to just prefix every `_` with a space?

ggreif:
> Any reason not to just prefix every wildcard `_` with a space?

Sure, why not? I wanted to address the problem, space looks good to me, lemme do it...

crusso:
Odd, I wondered why I can't update the branch?

ggreif:
this PR is fubar. Will open a new one.

PR #4274, opened 2023-11-11
fix: don't coalesce tag with underscore

ggreif:
... in the coverage warnings.

Repro:
``` Motoko
type A = { #a : { f : Int; g : Text }; #b : (Char, Char); #c : ?Nat; #d : Nat };

var a : A = #d 42;
switch a {case (#d 1) ()};
```
Gives
> stdin:13.1-13.26: warning [M0145], this switch of type
>   {#a : {f : Int; g : Text}; #b : (Char, Char); #c : ?Nat; #d : Nat}
> does not cover value
>   #d(0 or 2 or _) or
>   #a(_) or **#b_** or #c(_)


ggreif (commented):
I like this solution.

crusso (approved):

PR #4225, opened 2023-09-26
Enhanced Orthogonal Persistence (64-Bit without Graph Copy)

luc-blaeser:
### PR Stack
Enhanced orthogonal persistence support is structured in four PRs to ease review:
1. https://github.com/dfinity/motoko/pull/4488
2. https://github.com/dfinity/motoko/pull/4475
3. https://github.com/dfinity/motoko/pull/4225 **<-- this PR**
4. https://github.com/dfinity/motoko/pull/4193

# Enhanced Orthogonal Persistence (64-Bit without Graph Copy)

This implements the vision of **enhanced orthogonal persistence** in Motoko that combines:
* **Stable heap**: Persisting the program main memory across canister upgrades.
* **64-bit heap**: Extending the main memory to 64-bit for large-scaled persistence.

As a result, the use of secondary storage (explicit stable memory, dedicated stable data structures, DB-like storage abstractions) will no longer be necessary: Motoko developers can directly work on their normal object-oriented program structures that are automatically persisted and retained across program version changes.

## Advantages
Compared to the existing orthogonal persistence in Motoko, this design offers:
* **Performance**: New program versions directly resume from the existing main memory and have access to the memory-compatible data.
* **Scalability**: The upgrade mechanism scales with larger heaps and in contrast to serialization, does not hit IC instruction limits.

Compared to the explicit use of stable memory, this design improves:
* **Simplicity**: Developers do not need to deal with explicit stable memory.
* **Performance**: No copying to and from the separate stable memory is necessary.

## Design
The enhanced orthogonal persistence is based on the following main properties:
* Extension of the IC to retain main memory on upgrades.
* Supporting 64-bit main memory on the IC.
* A long-term memory layout that is invariant to new compiled program versions.
* A fast memory compatibility check performed on each canister upgrade.
* Incremental garbage collection using a partitioned heap.

### Memory Layout
In a co-design between the compiler and the runtime system, the main memory is arranged in the following structure, invariant of the compiled program version:
* Lower 4MB: Rust call stack.
* Space between 4MB and 4.5MB: Limited reserved space Wasm data segments, only used for the Motoko runtime system.
* Between 4.5MB and 5MB: Persistent metadata.
* Thereafter: Dynamic heap space. Fix start address at 5MB.

### Persistent Metadata
The persistent metadata describes all anchor information for the program to resume after an upgrade. 

More specifically, it comprises:
* A stable heap version that allows evolving the persistent memory layout in the future.
[...]

PR #4286, opened 2023-11-16
Incremental Graph-Copy-Based Stabilization

luc-blaeser:
# Incremental Graph-Copy-Based Stabilization

Using graph copying instead of Candid-based serialization for stabilization, to save stable variables across upgrades. By way of an optional incremental process, this supports arbitrarily large data beyond the upgrade instruction limit.

## Goals

* **Stop-gap solution until enhanced orthogonal persistence**: More scalable stabilization than the current Candid(ish) serialization.
* **With enhanced orthogonal persistence**: Upgrades in the presence of memory layout changes introduced by future compiler versions.

## Design

Graph copy of sub-graph of stable objects from main memory to stable memory and vice versa on upgrades.
The copy supports arbitrarily large data by logically extending the IC upgrade in a controlled way (see below).

## Properties
* Preserve sharing for all objects like in the heap.
* Allow the serialization format to be independent of the main memory layout.
* Limit the additional main memory needed during serialization and deserialization.
* Avoid deep call stack recursion (stack overflow).
* Allows arbitrarily long large stabilization/destabilization due to incremental mechanism (see below).

## Memory Compatibility Check
Apply a memory compatibility check analogous to the enhanced orthogonal persistence, since the upgrade compatibility of the graph copy is not identical to the Candid subtype relation.

## Incremental Upgrade

Supporting arbitrarily large upgrades beyond the instruction limit:
* Splitting the stabilization/destabilization in multiple asynchronous messages.
* Limiting the stabilization work units to fit the update or upgrade messages.
* Blocking other messages during the explicit incremental stabilization.
* Restricting the upgrade functionality to the canister owner and controllers.
* Stopping the GC during the explicit incremental upgrade process.

**Note**: This is only an extra option for large amounts of data if the work does not fit into the instruction limit of an upgrade. Otherwise, the usual upgrade process can be used.

### Usage

For large upgrades:
1. Initiate the explicit stabilization before the upgrade:
    
```
```

* An assertion first checks that the caller is the canister owner or a canister controller.
* All other messages to the canister will be blocked until the upgrade has been successfully completed.
* The GC is stopped.
* If defined, the actor's pre-upgrade function is called before the explicit stabilization.
* The stabilzation runs in possibly multiple asynchronous messages, each with a limited number of instructions.

2. Run the actual upgrade:

```
dfx deploy CANISTER_ID
```

* Run and complete the stabilization if not yet done in advance. 
* Perform the actual upgrade of the canister on the IC.
* Start the destabilization with a limited number of steps to fit into the upgrade message.
* If destabilization cannot be completed, the canister does not start the GC and does not accept messages except step 3.

3. Complete the explicit destabilization after the upgrade:

```
```

* An assertion checks that the caller is the canister owner or a canister controller.
* All other messages remain blocked until the successful completion of the destabilization.
* The destabilzation runs in possibly multiple asynchronous messages, each with a limited number of instructions.
* If defined, the actor's post-upgrade function is called at the end of the explicit destabilization.
* The GC is restarted.

### Remarks

* Steps 1 (explicit stabilization) and/or 2 (explicit destabilization) may not be needed if the corresponding operation fits into the upgrade message.
* Stabilization and destabilization steps are limited to the increment limits:

    Operation | Message Type | IC Instruction Limit | **Increment Limit**
    ----------|--------------|----------------------|--------------------
    **Explicit (de)stabilization step** | Update | 20e9 | **16e9**
    **Actual upgrade** | Upgrade | 200e9 | **160e9**

## Graph Copy Algorithm
Applying Cheney’s algorithm [1, 2] for both serialization and deserialization:

### Serialization
* Cheney’s algorithm using main memory as from-space and stable memory as to-space: 
* Focusing on stable variables as root (sub-graph of stable objects).
* The target pointers and Cheney’s forwarding pointers denote the (skewed) offsets in stable memory.
* Using streaming reads for the `scan`-pointer and streaming writes for the `free`-pointer in stable memory.

### Deserialization
* Cheney’s algorithm using stable memory as from-space and main memory as to-space: 
* Starting with the stable root created during the serialization process.
* A scan stack is used in the main memory to remember the deserialized objects for later scanning.
* Objects are allocated in main memory using the default allocator.
* Using random read/write access on the stable memory.

## Stable Format
For a long-term perspective, the object layout of the serialized data in the stable memory is fixed and independent of the main memory layout.
* Pointers support 64-bit representations, even if only 32-bit pointers are used in current main memory address space.
* The Brooks forwarding pointer is omitted (used by the incremental GC).
* The pointers encode skewed stable memory offsets to the corresponding target objects.
* References to the null objects are encoded by a sentinel value.
* `BigInt` are explicitly serialized in a defined portable little endian representation, without that the serialization or deserialization allocates temporary objects.

## Specific Aspects
* The null object is handled specifically to guarantee the singleton property. For this purpose, null references are encoded as sentinel values that are decoded back to the static singleton of the new program version.
* Field hashes in objects are serialized in a blob. On deserialization, the hash blob is allocated in the dynamic heap. Same-typed objects that have been created by the same program version share the same hash blob.
* Stable records can dynamically contain non-stable fields due to structural sub-typing. A dummy value can be serialized for such fields as a new program version can no longer access this field through the stable types.
* For backwards compatibility, old Candid destabilzation is still supported when upgrading from a program that used older compiler version.
* Incremental GC: Serialization needs to consider Brooks forwarding pointers (not to be confused with the Cheney's forwarding information), while deserialization can deal with partitioned heap that can have internal fragmentation (free space at partition ends).
* The partitioned heap prevents linear scanning of the heap, especially in the presence of large objects that can be placed at a higher partition than subsequently allocated normal-sized objects. For this reason, a scan stack is allocated in main memory, remembering the deserialized objects that still need to be scanned. With this, the deserialization does not need to make any assumptions of the heap structure (e.g. monotonically increasing allocations, free space markers, empty heap on deserialization start etc.).
* If actor fields are promoted to the `Any` type in a new program version, their content is released in that variable to allow memory reclamation.

## Open Aspects
* Unused fields in stable records that are no longer declared in a new program versions should be removed. This could be done during garbage collection, when objects are moved/evacuated.
* The scan stack used during destabilization involves dynamic allocations.

## Buffer Experiments
Measurments with buffered stable memory accesses proved to offer no relevant performance gains. On contrary, too coarse-grained buffers may even incur higher runtime costs.
Explored design alternatives:
* Page-buffered accesses, always streaming to stable memory, even during deserialization: About 3x performance degrade. https://github.com/dfinity/motoko/tree/luc%2Foriginal-graph-copy
* Selective small buffering for repeated word-read/scans of array elements and object fields: Very limited 2% improvement. https://github.com/dfinity/motoko/tree/luc%2Fgraph-copy-small-buffering 

## Related PRs

* Motoko Enhanced Orthogonal Persistence: https://github.com/dfinity/motoko/pull/4225
* Motoko Incremental Garbage Collector: https://github.com/dfinity/motoko/pull/3837

## References

[1] C. J. Cheney. A Non-Recursive List Compacting Algorithm. Communications of the ACM, 13(11):677-8, November 1970.

[2] R. Jones and R. Lins. Garbage Collection: Algorithms for Automatic Dynamic Memory Management. Wiley 2003. Algorithm 6.1: Cheney's algorithm, page 123.


crusso:
This is cool! Looking forward to reviewing it!

design/GraphCopyStabilization.md
```diff
@@ -0,0 +1,67 @@
...
+## Stable Format
+Currently, for ease of implementation, the object layout in stable memory is largely identical to the main memory with the following differences:
+* The Brooks forwarding pointer is omitted (used by the incremental GC).
+* The pointers encode skewed stable memory offsets to the corresponding target objects.
+* References to the null objects are encoded by a sentinel value.
+
+## Allocator Rules
+The destabilzation uses bulk-copying to main memory which is quite invasive. For correctness, the following invariants need to be met:
+* The main memory allocator used during deserialization must not write to the heap (only compute the target addresses).
+* The allocator must yield monotonously growing addresses during deserialization. Free space gaps are allowed to complete partitions.
```
crusso, line 57:
```suggestion
* The allocator must yield monotonically growing addresses during deserialization. Free space gaps are allowed to complete partitions.
```

rts/motoko-rts/src/stabilization.rs
```diff
@@ -0,0 +1,644 @@
...
+    /// * The source and target layout must use the same size for addresses, e.g. 32-bit.
+    /// * The allocator must be contiguously growing. Free space must be inserted when the
+    ///   allocator uses internal fragmentation, e.g. for the partitioned heap.
+    fn copy(&mut self, object: S) -> T;
+
+    /// Read an object at the `scan` position in the to-space, process all its potential
+    /// pointer fields by invoking `patch_field` for each of them.
+    /// Note:
+    /// * The default implementation assumes identical object body layout between main memory
+    /// and the serialized stable memory. Only the header layout differs insofar tha the main
```
crusso, line 126: ### 
```suggestion
    /// and the serialized stable memory. Only the header layout differs insofar that the main
```

rts/motoko-rts/src/stabilization.rs
```diff
@@ -0,0 +1,644 @@
...
+        let mut header = Obj::default();
+        header.tag = tag;
+        header.init_forward(target);
+        self.to_space.write(&header);
+        target
+    }
+
+    fn insert_free_space(&mut self, target: Value) {
+        let new_allocation = target.get_ptr();
+        // Heap allocations must be monotonously growing, such that their image
```
crusso, line 425:
```suggestion
        // Heap allocations must be monotonically increasing, such that their image
```

rts/motoko-rts/src/stabilization.rs
```diff
@@ -0,0 +1,490 @@
...
+    }
+
+    /// Check if the object has been forwarded.
+    /// Returns `None` if not forwarded, or otherwise, the new target address.
+    fn get_forward_address(&self, object: S) -> Option<T>;
+
+    /// Mark the object as forwarded and record its new target address.
+    fn set_forward_address(&mut self, object: S, target: T);
+
+    /// Allocate the object in the to-space by bumbing the `free` pointer.
```
crusso, line 105:
```suggestion
    /// Allocate the object in the to-space by bumping the `free` pointer.
```

rts/motoko-rts/src/stabilization.rs
```diff
@@ -0,0 +1,490 @@
...
+    use compatibility::{memory_compatible, TypeDescriptor};
+    use metadata::StabilizationMetadata;
+
+    let mut new_type_descriptor = TypeDescriptor::new(new_candid_data, new_type_offsets, 0);
+    let metadata = StabilizationMetadata::load(mem);
+    let mut old_type_descriptor = metadata.type_descriptor;
+    if !memory_compatible(mem, &mut old_type_descriptor, &mut new_type_descriptor) {
+        rts_trap_with("Memory-incompatible program upgrade");
+    }
+    // There may already exist some objects before destabilization.
```
crusso, line 470: Is this really true? is that because of deserialized initialization arguments?

rts/motoko-rts/src/stabilization/layout.rs
```diff
@@ -0,0 +1,375 @@
...
+//!
+//! Each object uses a `StableTag` as header an is followed by
+//! the object payload as outlined in the corresponding Rust structs.
+//! Some objects, such as `StableArray`, `StableObject`, `StableBlob`,
+//! and `StableBigNum`
+//!
+//! Not all heap memory object types are stabilized because some
+//! of them are not stable types. New object types can be added
+//! with backwards compatibility but encoding changes to existing stable
+//! data types must be handled extra care to ensure backwards compatibility.
```
crusso, line 22:
```suggestion
//! data types must be handled with extra care to ensure backwards compatibility.
```

rts/motoko-rts/src/stabilization/layout.rs
```diff
@@ -0,0 +1,375 @@
...
+//! a 32-bit version, while the downgrade from 64-bit to a 32-bit program
+//! is not supported and detected.
+//!
+//! Each object uses a `StableTag` as header an is followed by
+//! the object payload as outlined in the corresponding Rust structs.
+//! Some objects, such as `StableArray`, `StableObject`, `StableBlob`,
+//! and `StableBigNum`
+//!
+//! Not all heap memory object types are stabilized because some
+//! of them are not stable types. New object types can be added
```
crusso, line 20: So are they stabilized to a dummy value?
Say we have a statically stable record that (dynamically) contains a (non)-stable field of function type? That field needs to either be (somehow) elided or (I would expect) serialized to some dummy value. My gut feeling is that would be ok because the field would not be statically reachable so could never be used.

rts/motoko-rts/src/stabilization/layout.rs
```diff
@@ -0,0 +1,375 @@
...
+            TAG_MUTBOX => StableTag::MutBox,
+            TAG_OBJECT => StableTag::Object,
+            TAG_BLOB => StableTag::Blob,
+            TAG_BITS32 => StableTag::Bits32,
+            TAG_BITS64 => StableTag::Bits64,
+            TAG_REGION => StableTag::Region,
+            TAG_VARIANT => StableTag::Variant,
+            TAG_CONCAT => StableTag::Concat,
+            TAG_BIGINT => StableTag::BigInt,
+            TAG_OBJ_IND => StableTag::ObjInd,
```
crusso, line 86: I'm not sure if this is used to stabilize from main memory into stable memory, but it its, then I thing it might be missing a case for TAG_CLOSURE.
Consider:

```
stable let x : {} = { f = func () {} );
```
The type of x is an empty record, so stable, but its value is a non-empty record containing a closure. It the graph copy has to either drop the field, or translate it a (fake) closure or other dummy value. My hope it that a fake closure or dummy value is fine because it will be statically unreachable anyway (the static type doesn't reveal the field and we don't allow down-casts).

rts/motoko-rts/src/stabilization.rs
luc-blaeser: Yes, unfortunately, I did not yet analyze all objects, but I believe some are the actor initialization arguments, and I believe there is also the canister object itself. Some temporary objects, such as the type descriptors can be discarded, but I noticed there are other objects that need to be retained. I can analyze which objects these are such that we know better.

rts/motoko-rts/src/stabilization/layout.rs
luc-blaeser: My thought was that since stabilization only focuses on the dynamically reachable objects from stable actor variables, those objects would all be of stable types due to the type system. I assume that stable records cannot contain declarations of flexible (non-stable) records because this would indeed complicate things (especially also for enhanced orthogonal persistence). But maybe I miss an aspect?

rts/motoko-rts/src/stabilization/layout.rs
luc-blaeser: Thanks a lot for noticing this. It is related to your previous question, maybe more such types are affected. Probably, we could simply omit fields with non-stable types, as those are unreachable by the stable type definitions.

rts/motoko-rts/src/stabilization/layout.rs
luc-blaeser: *Update*: I fixed this issue by placing a dummy value in those non-stable fields. Thanks again for pointing out this bug. Still, the deletion of unused fields in new program version is open aspect.

rts/motoko-rts/src/stabilization/layout.rs
crusso: Great!

Re dropping fields, the static compatibility  check actually does not let you drop stable variables, to avoid resurrecting them later at an incompatible type (IIRC). The most you can do is promote them to type Any. Perhaps your type table based compatibility check missed that aspect.  But you are probably right  and the issue remains that the data will now never be removed. With Candid, once promoted to Any, the data would essentially be erased on destabilisation. With graph copy, it's retained. Unless we can somehow arrange that stable fields of type Any are not considered as roots by the GC.

luc-blaeser: Oh, this is an interesting point: Currently, the static compatibility check allows dropping fields in records (meaning declaring less fields than the original definition). I thought this would be safe because it would be a sub-type and thus compatible from memory perspective. Of course, there could be sharing, and other references may still point to the object by its full type. Deleting a field would only be conservatively possible, we can be sure that such a field is not declared in any (compatible) record type in that program, e.g. by collecting the field hashes used in a program version and checking whether the deserialized or resurrected object contains such old fields. But maybe a miss a point here?

rts/motoko-rts/src/stabilization.rs
crusso: That's probably possible, if anyone can still understand the somewhat convoluted code in desugar.ml.  Or maybe we should just re-arrange the backend a bit so it does the destabilization early, before any user code has a chance to execute, perhaps storing the destabilized record in some well-known global where it can be fetched later. Not sure.

rts/motoko-rts/src/stabilization/layout.rs
crusso: I finally got a round to taking a look at the fix, but I think just fixing record fields is not enough, and  the change should actually occur elsewhere, when scanning (?) a closure object.

To see why, I think these are also cases the must be considered:

```
func f() {};
stable let tuple : (Int, Any) = (0, f);
stable let variant : {#tag : Any} = #tag f;
stable let record : { lab : Any} = {lab = f};
stable let vector : [Any] = [f];
stable let array : [var Any] = [var f];
stable let opt : ?Any = ? f;
//etc?
```
Basically, any structured data, even of stable type, could (dynamically) contain a closure, by virtue of subtyping.
So I think the full solution is not  to special case records, but to serialize any TAG_CLOSURE value to some dummy value (perhaps just a scalar is good enough).




stable let

rts/motoko-rts/src/stabilization/layout/stable_object.rs
```diff
@@ -58,6 +58,16 @@ impl StaticScanner<Value> for Object {
...
+const NON_STABLE_OBJECT_TAGS: [Tag; 1] = [TAG_CLOSURE];
+
+fn has_non_stable_type(old_field: Value) -> bool {
+    unsafe { old_field.is_ptr() && NON_STABLE_OBJECT_TAGS.contains(&old_field.tag()) }
+}
+
+// Dummy value used for non-stable fields are potentially contained in stable records because of structural sub-typing.
+// Must be a non-skewed value such that the GC also ignores this value.
+const DUMMY_VALUE: StableValue = StableValue(0);
+
```
crusso, line 70: I don't think this is sufficient to handle all closures that could occur in stable values (see PR comment https://github.com/dfinity/motoko/pull/4286#discussion_r1404674463

rts/motoko-rts/src/stabilization/layout.rs
luc-blaeser: Thanks a lot. Indeed, my implementation was wrong and only considered structural sub-typing. I fixed this.

rts/motoko-rts/src/stabilization/layout/stable_object.rs
luc-blaeser: Yes, this is indeed insufficient. I fixed it. Thanks a lot for noticing and sorry for the error.

crusso:
For the record, I did a bit of digging and for bignum serialization, we could also use the functions provided by LibTomMath (if you don't already).

See section 10.2 of the pdf manual https://github.com/libtom/libtommath/releases/download/v1.2.0/bn-1.2.0.pdf

There's to/from_sbin and libgmp style pack unpack avialable. Might actually be easier/cheaper than sleb encoding.

rts/motoko-rts/src/stabilization/layout/stable_bigint.rs
```diff
@@ -0,0 +1,85 @@
+use crate::stabilization::layout::{checked_to_usize, write_padding_u64};
+use crate::stabilization::reader_writer::{ScanStream, StableMemorySpace, WriteStream};
+use crate::stabilization::StableMemoryAccess;
+use crate::tommath_bindings::{mp_digit, mp_int};
+use crate::types::{BigInt, Bytes, Obj, Value, TAG_BIGINT};
+
+use super::{checked_to_u32, round_to_u64, Serializer, StableValue, StaticScanner};
+
+// TODO: Use a format that is independent of Tom's math library, e.g. by using LEB128/SLEB128 encoding.
```
crusso, line 9: I did a bit of digging and for bignum serialization, we could also use the functions provided by LibTomMath 

See section 10.2 of the pdf manual https://github.com/libtom/libtommath/releases/download/v1.2.0/bn-1.2.0.pdf

There's to/from_sbin (bigendian) and libgmp style pack unpack available. Might actually be easier/cheaper than sleb encoding, plus there's a function for determining the size in bytes upfront.

rts/motoko-rts/src/stabilization/metadata.rs
```diff
@@ -0,0 +1,224 @@
...
+//! Type descriptor address M:
+//!   Candid type table
+//!     Byte length (u32)
+//!     Data
+//!   Type offset table
+//!     Byte length (u32)
+//!     Data
+//!   (possible zero padding)
+//! -- Last physical page (metadata):
+//!   (zero padding to align at page end)
```
crusso, line 19: Perhaps we could store the instruction count of serialization here, and then use it to compute the cost of upgrade to store and retrieve from a global, if we want such stats.

crusso (commented):
Some more interim comments

rts/motoko-rts/src/stabilization/reader_writer.rs
```diff
@@ -0,0 +1,287 @@
...
+/// Optimized for Cheney's algorithm.
+///
+/// The buffer supports two location-independent streams:
+/// * Streamed reading and updating, used for scanning and patching pointers.
+/// * Streamed writing, used for allocating new objects.
+pub struct StableMemorySpace {
+    /// The pointers used in the serialized stable memory layout are
+    /// relative to this start address of the to-space.
+    base_address: u64,
+    /// Used fore reading and updating.
```
crusso, line 93:
```suggestion
    /// Used for reading and updating.
```

rts/motoko-rts/src/stabilization/reader_writer.rs
```diff
@@ -0,0 +1,287 @@
...
+    scan_address: u64,
+    /// Used for writing.
+    free_address: u64,
+    // TODO: Use a way to not allocate a 128K empty data segment.
+    cache: [StableMemoryPage; CACHE_SIZE],
```
crusso, line 98: Could you just use a global array? But we have 2MB of shadow stack, so 128K not so bad, unless you are worried about the initialization cost. Maybe you could just use physical page 0 and 1, and (pray) the stack doesn't grow into it.

rts/motoko-rts/src/stabilization/metadata.rs
luc-blaeser: Great idea, this would simplify upgrade benchmarking.

rts/motoko-rts/src/idl.rs
```diff
@@ -59,6 +60,11 @@ unsafe fn is_primitive_type(extended: bool, ty: i32) -> bool {
         && (ty >= IDL_PRIM_lowest || ty == IDL_REF_principal || (extended && ty == IDL_EXT_region))
 }
 
+// Only used for memory-compatibility checks during stabilzation using graph-copy-based serialization.
+unsafe fn is_extended_primitive_type(ty: i32) -> bool {
+    is_primitive_type(true, ty) || ty < 0 && ty == IDL_EXT_blob
```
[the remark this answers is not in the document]
luc-blaeser: Indeed, I should refactor this by better naming/structuring. There are three modes: 
1. Proper Candid serialization (using `is_primitive_type() with `extended = false`, allowing neither regions not low-level blobs).
2. Candid(ish) for the old stabilization (using `is_primitive_type` with `extended=true`, allowing regions, but not low-level blobs).
3. The memory-compatibility check for the new graph-based stabilization (using `is_extended_primitive_type()`, considering both regions and low-level blobs). 

The same refactoring would be also beneficial for the enhanced orthogonal persistence PR.

rts/motoko-rts/src/stabilization.rs
luc-blaeser: I agree, the nicer solution would be to ensure an empty heap on destabilization. I will try to rearrange the code. Perhaps, it could be quite easy, as it is only about one special case.

rts/motoko-rts/src/idl.rs
```diff
@@ -543,8 +549,320 @@ unsafe fn is_opt_reserved(typtbl: *mut *mut u8, end: *mut u8, t: i32) -> bool {
...
+    t2: i32,
+    main_actor: bool,
+) -> bool {
+    if t1 >= 0 && t2 >= 0 {
+        let t1 = t1 as u32;
+        let t2 = t2 as u32;
+        if recurring_memory_check(rel, variance, t1, t2) {
+            return true;
+        };
+        remember_memory_check(rel, variance, t1, t2);
```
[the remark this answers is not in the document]
luc-blaeser: Thank you. Indeed, I assume non-recursiveness of main actor type, while not considering this during the cache logic. I will improve this and check the non-recursiveness of the main actor type.

rts/motoko-rts/src/stabilization.rs
```diff
@@ -0,0 +1,526 @@
...
+}
+
+impl<'a, M: Memory> StableMemoryAccess for Deserialization<'a, M> {
+    fn to_space(&mut self) -> &mut StableMemorySpace {
+        &mut self.to_space
+    }
+}
+
+fn clear_stable_memory(start: u64, length: u64) {
+    const CHUNK_SIZE: usize = 1024;
```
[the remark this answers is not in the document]
luc-blaeser: Good idea. I will need to check that the zeroed chunk block is not allocated as a huge Wasm data segment in the generated Rust code.

rts/motoko-rts/src/stabilization/layout.rs
```diff
@@ -0,0 +1,390 @@
...
+//! A 32-bit program stores `StableValue` that fit into 32-bit, and
+//! scalar values beyond this width are boxed.
+//! This means that a 32-bit and 64-bit program can always upgrade from
+//! a 32-bit version, while the downgrade from 64-bit to a 32-bit program
+//! is not supported and detected.
+//!
+//! Each object uses a `StableTag` as header an is followed by
+//! the object payload as outlined in the corresponding Rust structs.
+//! Some objects, such as `StableArray`, `StableObject`, `StableBlob`,
+//! and `StableBigNum`
```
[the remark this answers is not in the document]
luc-blaeser: Oh, yes, I will complete it.

rts/motoko-rts/src/stabilization/layout.rs
```diff
@@ -0,0 +1,390 @@
...
+//!
+//! The stable object graph resides a linear stable memory space.
+//!
+//! Pointers are serialized as skewed offsets in that space.
+//! Scalar and pointer values are serialized as 64-bit `StableValue`
+//! for a long-term perspective, even if main memory operates on 32-bit.
+//! A 32-bit program stores `StableValue` that fit into 32-bit, and
+//! scalar values beyond this width are boxed.
+//! This means that a 32-bit and 64-bit program can always upgrade from
+//! a 32-bit version, while the downgrade from 64-bit to a 32-bit program
```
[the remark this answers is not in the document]
luc-blaeser: This is indeed a problem. I believe we should maybe use two versions for destabilization, starting with 32-bit for the moment. However, the future backwards compatibility (on deserialization) would then make it even more complicated...

rts/motoko-rts/src/stabilization/reader_writer.rs
luc-blaeser: I was primarily worried about a potential Wasm data segment being created by the Rust compiler to define the zeroed memory map. I will check how the compiler behaves, and/or better use a more explicit initialization (not to be surprised in the future by Rust compiler changes).

crusso (commented):
(forgot some pending comments)

src/codegen/compile.ml
```diff
@@ -3958,6 +3681,222 @@ module Blob = struct
 
 end (* Blob *)
 
+
+module Object = struct
```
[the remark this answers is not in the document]
luc-blaeser: Yes, unfortunately, same for the enhanced orthogonal persistence PR.

src/codegen/compile.ml
```diff
@@ -3958,6 +3681,222 @@ module Blob = struct
...
+ 
+     We could alternatively switch to an allocate-first approach in the
+     await-translation of objects, and get rid of this indirection -- if it were
+     not for the implementing of sharing of mutable stable values.
+   *)
+ 
+   let header_size env = Int32.add (Tagged.header_size env) 2l
+ 
+   (* Number of object fields *)
+   let size_field env = Int32.add (Tagged.header_size env) 0l
```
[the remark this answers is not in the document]
luc-blaeser: Indeed, the hash blob is a little bit like a type descriptor object that could carry the size.

rts/motoko-rts/src/stabilization/layout/stable_bigint.rs
luc-blaeser: Thank you. This seems interesting. 
A challenge that I see is that these functions require temporary pre-allocated buffer memory with a size depending on the bignum. The serialization (`to_sbin` calling `to_ubin`) even allocates an additional internal copy of the bignum. This could be problematic for large-sized bignums, but I am not sure how relevant this is in practice.

luc-blaeser:
Just for future reference, if needed:
Branching off the original more complex version of the graph copy that caches stable memory and flips memory on deserialization: https://github.com/dfinity/motoko/tree/luc%2Foriginal-graph-copy

rts/motoko-rts/src/stabilization/reader_writer.rs
luc-blaeser: **Update**: I removed the stable memory cache, as it seems to be less efficient than direct accesses.

luc-blaeser:
**Update**: I changed the serialization of `BigInt` to use the defined binary format of Tom's math library `mb_to_sbin()` and `mb_from_sbin()`. The only aspect I am little bit worried are the temporary allocations performed by the library which could be relevant if memory is short.

rts/motoko-rts/src/stabilization/layout/stable_bigint.rs
luc-blaeser: **Update**: I implemented the more portable serialization using the suggested LibTomMath functions. As mentioned, the only disadvantage are temporary extra allocations involved, but maybe, for moderately-sized big numbers this is acceptable.

crusso:
> **Update**: I changed the serialization of `BigInt` to use the defined binary format of Tom's math library `mb_to_sbin()` and `mb_from_sbin()`. The only aspect I am little bit worried are the temporary allocations performed by the library which could be relevant if memory is short.

I wonder if we could write our own that uses stack allocation instead, by modifying that code.

crusso (commented):
Left some comments for possible (micro) optimizations.

Given the deserialization code is a simple, interruptible loop, recording its progress mostly within the heap/stable-memory, I wonder if we can even use this when there is too much data for the upgrade budget? We "just" do the serialization/deserialization over several ordinary messages, immediately before and after upgrade, rejecting all other messages in the meantime. But maybe I'm missing something and it's not actually so easy to pause and resume the loop iterations between IC messages.

rts/motoko-rts/src/stabilization.rs
luc-blaeser: **Update**: As linear scanning is not feasible with the partitioned heap (in particular due to the large object allocation), I had to change the deserialization, now using a data structure to remember the pending objects to be scanned. As a result, the deserialization does no longer make any assumptions about the heap, such that it does not need to be empty at the beginning of the deserialization process.

src/codegen/compile.ml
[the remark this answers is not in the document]
luc-blaeser: Yes, I also believe the information would be sufficient: The blob size denotes the number of field hashes and thus the size of objects of its type. (If we would ever change to another field lookup data structure, e.g. a perfect hash table with possible few empty hash entries, we could still store the object length in this kind of type descriptor.)

luc-blaeser:
> > **Update**: I changed the serialization of `BigInt` to use the defined binary format of Tom's math library `mb_to_sbin()` and `mb_from_sbin()`. The only aspect I am little bit worried are the temporary allocations performed by the library which could be relevant if memory is short.
> 
> I wonder if we could write our own that uses stack allocation instead, by modifying that code.

Yes, this would probably work for the temporary buffer and the temporary `BigInt` allocation(s) during serialization. An aspect to consider could be large-sized bigints - not sure whether and how often this occurs in practice. Another more tricky aspect could be the `BigInt` allocations during deserialization: While decoding, Tom's math library implementation increases the number by reallocating the object and keeping the last instance.

rts/motoko-rts/src/stabilization/stable_memory_stream.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+}
+
+impl ScanStream for StableMemoryStream {
+    fn scan_completed(&self) -> bool {
+        debug_assert!(self.scan_address <= self.free_address);
+        self.scan_address == self.free_address
+    }
+
+    fn read<T>(&mut self) -> T {
```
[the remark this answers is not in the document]
luc-blaeser: Good point. I will analyze this. Maybe there is an optimization potential.

rts/motoko-rts/src/types.rs
```diff
@@ -710,75 +727,21 @@ impl Blob {
     }
 }
 
-/// Note: Do not declare 64-bit fields, as otherwise, the objects are expected to be 64-bit aligned.
-/// This is not the case in the current heap design.
-/// Moreover, fields would also get 64-bit aligned causing implicit paddding.
-
-#[repr(C)] // See the note at the beginning of this module
-pub struct Stream {
```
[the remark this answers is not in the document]
luc-blaeser: Yes, this is only used Candid stabilization, which would no longer be required by a new stabilization (graph-copy and/or orthogonal persistence). The existing Candid stabilization would still be required for upgrades from older program versions, but it does not use the stream implementation.

rts/motoko-rts/src/stabilization.rs
```diff
@@ -0,0 +1,462 @@
...
+        }
+    }
+
+    /// Note:
+    /// * The deserialized memory may contain free space at a partition end.
+    fn scan(&mut self) {
+        let target_object = unsafe { self.scan_stack.pop() };
+        debug_assert!(target_object != STACK_EMPTY);
+        unsafe {
+            Self::scan_deserialized(self, target_object, &|context, original| {
```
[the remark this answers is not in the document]
luc-blaeser: I will analyze the compiled code, and optimize, if needed.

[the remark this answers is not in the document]
crusso: If Rust does specialization based on traits, then it might be enough to abstract the code using a trait, not a lambda.

crusso:
Probably before doing any optimization we should try to capture the instructions used by upgrade to have some measure to optimize against.

rts/motoko-rts/src/stabilization/layout.rs
luc-blaeser: **Minor improvement**: The implementation now drops the content of `Any`-typed actor fields.

rts/motoko-rts/src/idl.rs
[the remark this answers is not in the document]
luc-blaeser: **Update**: I now fixed it in the simple way by not using the cache for the top-level main actor type comparison. I will apply the same fix for the enhanced orthogonal persistence PR.

rts/motoko-rts/src/idl.rs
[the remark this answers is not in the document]
luc-blaeser: **Update**: I now refactored the three kinds of compatibility modes with the different `is_primitive_type()` logics.

rts/motoko-rts/src/stabilization.rs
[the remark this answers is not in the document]
luc-blaeser: **Update**: I checked it and as expected, Rust does not allocate an unnecessary data segment for the empty chunk array but clears the reserved stack space with `memset`. I now increased it to 64 KB (Wasm page size).

rts/motoko-rts/src/stabilization/layout/stable_bigint.rs
luc-blaeser: **Another update**: I now implemented a custom direct serialization of a bignum to little endian arbitary-length byte sequence, without using Tom's math library. As a result, no temporary allocations are needed.

luc-blaeser:
**Update**: I now implemented a bespoke direct serialization that stores the bignum in little endian arbitrary-length byte array, without causing temporary allocations.

rts/motoko-rts/src/stabilization/stable_memory_stream.rs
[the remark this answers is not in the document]
luc-blaeser: **Update**: I checked it for bulk returrn types and fortunately, the Rust compiler optimzes this by using an output parameter instead of copying the result.

rts/motoko-rts/src/stabilization/layout/stable_array.rs
```diff
@@ -0,0 +1,77 @@
...
+        stable_object: StableValue,
+        target_array: *mut Array,
+    ) {
+        let stable_address = stable_object.payload_address();
+        for index in 0..(*target_array).len {
+            let element_address = stable_address
+                + size_of::<StableArray>().to_bytes().as_usize() as u64
+                + (index * size_of::<StableValue>().to_bytes().as_u32()) as u64;
+            let element = stable_memory.read::<StableValue>(element_address);
+            target_array.set_raw(index, element.deserialize());
```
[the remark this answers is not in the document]
luc-blaeser: Thanks for noticing this. I optimized it in meantime.

src/codegen/compile.ml
```diff
@@ -5236,6 +5177,24 @@ module StableMem = struct
...
+    (G.if0
+      (compile_unboxed_const version_graph_copy_no_regions ^^
+      set_version env)
+      G.nop) ^^
+    get_version env ^^
+    compile_eq_const legacy_version_regions ^^
+    (G.if0
+      (compile_unboxed_const version_graph_copy_regions ^^
+      set_version env)
+      G.nop)
```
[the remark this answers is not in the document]
luc-blaeser: Thank you. I improved this.

rts/motoko-rts/src/stabilization.rs
```diff
@@ -0,0 +1,462 @@
...
+            let object = Self::resolve_gc_forwarding(object);
+            debug_assert!(object.is_obj());
+            let address = self.to_space.written_length();
+            serialize(self.mem, &mut self.to_space, object);
+            StableValue::from_stable_address(address)
+        }
+    }
+
+    fn scan(&mut self) {
+        scan_serialized(self, &|context, original| {
```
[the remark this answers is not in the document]
luc-blaeser: I analyzed the compiled code, and Rust seems to eliminate the lambda and completely inline its body (no extra locals used for `context` and `original`. This is because the lambda is also specified as a generic type, such that the the function is specialized for this specific lambda argument...

rts/motoko-rts/src/stabilization.rs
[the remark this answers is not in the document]
luc-blaeser: I checked this and Rust optimizes these lambdas away (all is static, the function is specialized for the specific lambda).

luc-blaeser:
> Hmm, I took a look at the decompiled wasm for the rts and couldn't find any call_indirect I could attribute to the object visitor: perhaps those lambda calls are indeed all optimized to direct calls.
> 
> It's possible I missed something of course.

Yes, I also checked, Rust optimizes these and seems to be adhere to its "zero-cost abstraction" principle as for the generics, traits, and lambdas (unless the types would be declared `dyn` which we do not use in the RTS).

rts/motoko-rts/src/types.rs
```diff
@@ -598,11 +612,15 @@ impl Region {
 #[repr(C)] // See the note at the beginning of this module
 pub struct Object {
     pub header: Obj,
-    pub size: u32,     // Number of elements
-    pub hash_ptr: u32, // Pointer to static information about object field labels. Not important for GC (does not contain pointers).
+    pub size: u32,        // Number of elements
```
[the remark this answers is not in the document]
luc-blaeser: Good idea. I eliminated this redundancy. This could also be applied to the enhanced orthogonal persistence PR.

src/codegen/compile.ml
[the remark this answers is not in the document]
luc-blaeser: **Update**: I eliminated the size information from the object header, by deriving it from the hash blob size instead. Thanks for this nice suggestion.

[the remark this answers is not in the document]
luc-blaeser: For enhanced orthogonal persistence, it should be relatively simple. I will implement it there.

rts/motoko-rts/src/stabilization/metadata.rs
luc-blaeser: **Update**: I now record the serialization costs and offer `Prim.rts_upgrade_instructions()` to retrieve the total number of instructions of the last upgrade (serialization plus deserialization).

luc-blaeser:
> Left some comments for possible (micro) optimizations.
> 
> Given the deserialization code is a simple, interruptible loop, recording its progress mostly within the heap/stable-memory, I wonder if we can even use this when there is too much data for the upgrade budget? We "just" do the serialization/deserialization over several ordinary messages, immediately before and after upgrade, rejecting all other messages in the meantime. But maybe I'm missing something and it's not actually so easy to pause and resume the loop iterations between IC messages.

I believe this could be a viable approach, and interrupting/resuming the upgrade would be simple enough to realize. As you say, it would be important to block other messages in meantime, including the GC during serialization and deserialization (e.g. serialization replaces objects by forwarding blocks; during deserialization, there exist unpatched references in objects pending to be scanned). A downside could be that the users or DFX or some other external mechanism would need to issue these repeated update increment calls...

luc-blaeser:
> Left some comments for possible (micro) optimizations.
> Given the deserialization code is a simple, interruptible loop, recording its progress mostly within the heap/stable-memory, I wonder if we can even use this when there is too much data for the upgrade budget? We "just" do the serialization/deserialization over several ordinary messages, immediately before and after upgrade, rejecting all other messages in the meantime. But maybe I'm missing something and it's not actually so easy to pause and resume the loop iterations between IC messages.
> 
Thanks for this suggestion. I implemented this "incremental" upgrade process in the additional PR https://github.com/dfinity/motoko/pull/4361

luc-blaeser:
Note: Branching off the original non-incremental graph copy to https://github.com/dfinity/motoko/tree/luc%2Fnon-incremental-graph-copy

luc-blaeser:
Closing PR in favor of https://github.com/dfinity/motoko/pull/4475 (graph copy used in combination with enhanced orthogonal persistence).

PR #4299, opened 2023-11-28
experiment: two additional tests for graph-copy

crusso:
Two additional tests for @luc-blaeser's graph copying branch (luc/graph-copy), Builds on #4286.

- `run-drun/stable-exp-immut.mo` is a version of `stable-exp` that uses immutable arrays, to highlight the sharing of those arrays that would previously explode stabilization (using Candidish) (as suggested in a comment in `stable-exp.mp`)
- `run-drun-non-ci/stable-var-large.mo` attempts to allocated more than 2 GB of stable variables and then upgrade. This should succeed with graph-copy but fail with Candidish due to non-streaming destabilization (which requires copying all stable variable data to the heap before deserialization). Unfortunately, I can't execute it on my machine without causing drun to choke (possible due a lack of diskspace). Disabled for now by placing in attic run-drun-non-ci.

### DIFF
--- a/test/run-drun-non-ci/stable-variable-large.mo
+++ b/test/run-drun-non-ci/stable-variable-large.mo
@@ -1,4 +1,6 @@
 //MOC-FLAG --incremental-gc
+// This would let us test that destablization of >2GB works,
+// if only drun could execute it properly.
 import P "mo:⛔";
 
 actor {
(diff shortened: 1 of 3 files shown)


luc-blaeser (approved):
Thanks a lot. Good tests. Would be great to merge these tests to the graph-copy PR.

luc-blaeser:

PR #4300, opened 2023-11-28
Incremental GC Tuning

luc-blaeser:
# Incremental GC Tuning

Tuning the incremental GC for more conservative collection of larger heaps:

1. GC memory reserve: Keeping a reserve of 512 MB for the incremental GC itself, to be always able to allocate the mark bitmap (128 MB) and have some free space (384 MB) to perform evacuations. This reserve is on top of the 256 MB general reserve for upgrades and query calls.

2. Scheduling the GC more frequently for a medium-sized heap. Until now, there were only two thresholds (critical and non-critical).
    
    Limit    | Heap Size       | Growth Threshold
    ---------|-----------------|------------------
    Small    | 0 .. 1 GB       | 65% 
    Medium   | 1 .. 2.25 GB    | 35%
    Critical | 2.25 .. 3.25 GB | 1%

3. Increasing the GC increment limits by a factor of 3 since DTS (deterministic time-slicing) limits have been increased by factor of 10 since the GC design. Increments now have a limit of about 1.8 billion instructions (while the IC message instruction limit is at 20 billion instructions).

4. Introducing a system-level GC trigger function that can be called by the canister controller or owner to explicitly run a GC increment. This can be helpful if the canister memory is full and the GC received too little time to complete. This can happen e.g. when a large amount of memory is allocated in a single message or when objects are made unreachable shortly before the memory is exhausted.

    ```
    dfx canister call CANISTER_ID __motoko_gc_trigger "()" 
    ```

    This can be called multiple times in a series to complete a GC run.


src/lowering/desugar.ml
```diff
@@ -457,6 +457,30 @@ and export_footprint self_id expr =
...
+  let {lab;typ;_} = motoko_gc_trigger_fld in
+  let v = "$"^lab in
+  let scope_con1 = Cons.fresh "T1" (Abs ([], scope_bound)) in
+  let scope_con2 = Cons.fresh "T2" (Abs ([], Any)) in
+  let bind1  = typ_arg scope_con1 Scope scope_bound in
+  let bind2 = typ_arg scope_con2 Scope scope_bound in
+  let caller = fresh_var "caller" caller in
+  ([ letD (var v typ) (
+        funcE v (Shared Write) Promises [bind1] [] [] (
+            (asyncE T.Fut bind2
```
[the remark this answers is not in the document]
crusso: Looking at the generated code, where the collector is called with a non-empty stack, with some Motoko values on the stack, I think it might be much safer to just do this in the backend only, in the fashion of export_async_helper. 

If that doesn't make sense, I can draft a follow on PR to clarify.

BTW, I also noticed that the primitive PerformGC doesn't update the mutator/collector stats correctly, leading to some  minor misaccounting. I think. Perhaps I missed that when adding the stats (just before you joined IIRC).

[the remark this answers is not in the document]
luc-blaeser: Thanks a lot. The counting is indeed not correct. Indeed, it is much better to implement it in the backend. Many thanks also for the PR https://github.com/dfinity/motoko/pull/4303. I added some minor refinements on top of that PR https://github.com/dfinity/motoko/pull/4305.

luc-blaeser:
> see comments and draft PR #4303

Thanks you very much. I think we should use your version (I have some small adjustments in https://github.com/dfinity/motoko/pull/4305).

test/run-drun/gc-trigger-acl/C.mo
```diff
@@ -0,0 +1,10 @@
+actor class C(gc : actor {
+    __motoko_gc_trigger: () -> async ()
+  })
+{
+
+   // attempt to call passed in gc_trigger
+   public shared func callback() : async () {
+      await gc.__motoko_gc_trigger(); // should fail unless controller or selfaxs
+   }
+}
```
crusso, line 10:
```suggestion
}

```

test/run.sh
```diff
@@ -251,6 +251,15 @@ do
 
   case $ext in
   "mo")
+    if grep -q "//INCREMENTAL-GC-ONLY" $base.mo
```
crusso, line 254: What was the reason for this tweak? Is it not ok to run the test on all gcs?

test/run-drun/gc-trigger.mo
```diff
@@ -0,0 +1,36 @@
+//MOC-NO-FORCE-GC
+//INCREMENTAL-GC-ONLY
```
crusso, line 2: I wasn't sure why you need this INCREMENTAL-GC-ONLY flag..

rts/motoko-rts/src/gc/generational.rs
```diff
@@ -117,7 +117,8 @@ static mut OLD_GENERATION_THRESHOLD: usize = 32 * 1024 * 1024;
 static mut PASSED_CRITICAL_LIMIT: bool = false;
 
 #[cfg(feature = "ic")]
-const CRITICAL_MEMORY_LIMIT: usize = (4096 - 512) * 1024 * 1024 - crate::memory::MEMORY_RESERVE;
+const CRITICAL_MEMORY_LIMIT: usize =
+    (4096 - 512) * 1024 * 1024 - crate::memory::GENERAL_MEMORY_RESERVE;
```
crusso, line 121:
```suggestion
    (4096 - 512) * 1024 * 1024 - crate::memory::GENERAL_MEMORY_RESERVE;
```
Nit: Maybe use your new constants here too? And it definition of OLD_GENERATION_THRESHOLD above?

test/run-drun/gc-trigger.mo
crusso: Keep if needed, I was just wondering if this was debugging residue.

crusso (approved):
But see minor comments

test/run.sh
luc-blaeser: I only used this for the `gc-trigger.mo` test. The test does not work for the generational GC, as it reclaims the garbage immediately, which is nice but then the GC trigger has no effect... Maybe, there is a better way to test the GC trigger, that it reclaims memory in all GC cases.

test/run-drun/gc-trigger.mo
luc-blaeser: I only used it for the `gc-trigger.mo` test case (as the generational GC reclaimed memory earlier than I needed for test).

rts/motoko-rts/src/gc/generational.rs
luc-blaeser: I looked into this and am not sure whether I should reuse the same or similar heuristics as in incremental GC. One aspect is that the GC does not need much extra space except for the mark bitmap (128MB). Moreover, there is no reclamation delay by multiple increments but the free space is entirely collected when the limit is exceeded and/or full collection is initiated. This is why I believe we could keep the more generous limit. But most probably, the full collection would not fit into instruction limit anyways for larger fine-grained heaps. 
I am interested in your thoughts. Maybe there are an additional aspects I should consider?

luc-blaeser:
> But see minor comments

I have a few reply comments/thoughts, and interested in your opinion, especially regarding the application of the incremental GC's scheduling thresholds to the non-incremental GC.

rts/motoko-rts/src/gc/generational.rs
crusso: Oh all I was suggesting is maybe replace     1024 by MB etc, as elsewhere. Totally optional

PR #4303, opened 2023-11-29
experiment: implement gc trigger in backend

crusso:
suggested change to #4300 that avoids unintended duplicate gc calls from non-empty stack

The drawback is yet more code in the compile.ml but could be shrunk with some refactoring..

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -4938,6 +4938,30 @@ module IC = struct
     compile_eq_const 0l ^^
     E.else_trap_with env "not a self-call"
 
+  let assert_caller_self_or_controller env =
+    let (set_len1, get_len1) = new_local env "len1" in
+    let (set_len2, get_len2) = new_local env "len2" in
+    let (set_str1, get_str1) = new_local env "str1" in
+    let (set_str2, get_str2) = new_local env "str2" in
+    system_call env "canister_self_size" ^^ set_len1 ^^
+    system_call env "msg_caller_size" ^^ set_len2 ^^
+    get_len1 ^^ get_len2 ^^ G.i (Compare (Wasm.Values.I32 I32Op.Eq)) ^^
+    E.else_trap_with env "not a self-call" ^^
+
+    get_len1 ^^ Blob.dyn_alloc_scratch env ^^ set_str1 ^^
+    get_str1 ^^ compile_unboxed_const 0l ^^ get_len1 ^^
+    system_call env "canister_self_copy" ^^
+
+    get_len2 ^^ Blob.dyn_alloc_scratch env ^^ set_str2 ^^
+    get_str2 ^^ compile_unboxed_const 0l ^^ get_len2 ^^
+    system_call env "msg_caller_copy" ^^
+
+    get_str1 ^^ get_str2 ^^ get_len1 ^^ Heap.memcmp env ^^
+    compile_eq_const 0l ^^
+    get_str1 ^^ get_len1 ^^ is_controller env ^^
+    G.i (Binary (Wasm.Values.I32 I32Op.Or)) ^^
+    E.else_trap_with env "not a self or controller call"
+
   (* Cycles *)
 
   let cycle_balance env =
@@ -9054,6 +9078,35 @@ module FuncDec = struct
     | _ -> ()
     end
 
+  let export_gc_trigger_method env =
+    let name = Type.(motoko_gc_trigger_fld.lab) in
+    begin match E.mode env with
+    | Flags.ICMode | Flags.RefMode ->
+      Func.define_built_in env name [] [] (fun env ->
+        message_start env (Type.Shared Type.Write) ^^
+        (* Deserialize unit *)
+        Serialization.deserialize env [] ^^
+        IC.get_self_reference env ^^
+        (* Check that we are called from this or a controller *)
+        IC.assert_caller_self_or_controller env ^^
+        Serialization.serialize env [] ^^
+        IC.reply_with_data env ^^
+        (* message_cleanup env (Type.Shared Type.Write), but
+           forces collection *)
+        GC.record_mutator_instructions env ^^
+        E.collect_garbage env true ^^
+        GC.record_collector_instructions env ^^
+        Lifecycle.trans env Lifecycle.Idle
+      );
+
+      let fi = E.built_in env name in
+      E.add_export env (nr {
+        name = Lib.Utf8.decode ("canister_update " ^ name);
+        edesc = nr (FuncExport (nr fi))
+      })
+    | _ -> ()
+    end
+
 end (* FuncDec *)
 
 
@@ -12049,6 +12102,7 @@ and conclude_module env set_serialization_globals start_fi_o =
   RTS_Exports.system_exports env;
 
   FuncDec.export_async_method env;
+  FuncDec.export_gc_trigger_method env;
 
   (* See Note [Candid subtype checks] *)
   Serialization.set_delayed_globals env set_serialization_globals;
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -1425,6 +1425,17 @@ module Stack = struct
     f get_x ^^
     dynamic_free_words env get_n
 
+  let dynamic_with_bytes env name f =
+    let (set_n, get_n) = new_local env "n" in
+    let (set_x, get_x) = new_local env name in
+    (* round up to nearest wordsize *)
+    compile_add_const (Int32.sub Heap.word_size 1l) ^^
+    compile_divU_const Heap.word_size ^^
+    set_n ^^
+    dynamic_alloc_words env get_n ^^ set_x ^^
+    f get_x ^^
+    dynamic_free_words env get_n
+
   (* Stack Frames *)
 
   (* Traditional frame pointer for accessing statically allocated locals/args (all words)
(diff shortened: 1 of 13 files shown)


crusso:
Generated code:
```
  (func $__motoko_gc_trigger (type 0)
    (local $start i32) (local $size i32) (local $len i32) (local $blob i32) (local $len1 i32) (local $len2 i32) (local $str1 i32) (local $str2 i32)
    call $trans_state5
    call 8
    local.set $len1
    call 16
    local.tee $len2
    call $alloc_blob
    i32.const 9
    i32.add
    local.tee $str2
    i32.const 0
    local.get $len2
    call 15
    local.get $len1
    local.get $len2
    i32.eq
    if (result i32)  ;; label = @1
      local.get $len1
      call $alloc_blob
      i32.const 9
      i32.add
      local.tee $str1
      i32.const 0
      local.get $len1
      call 7
      local.get $str1
      local.get $str2
      local.get $len1
      call $memcmp
      i32.eqz
    else
      i32.const 0
    end
    local.get $str2
    local.get $len2
    call 11
    i32.or
    if  ;; label = @1
    else
      i32.const 2099516
      i32.const 29
      call 33
      unreachable
    end
    call 14
    local.tee $len
    call $alloc_blob
    local.tee $blob
    i32.const 9
    i32.add
    i32.const 0
    local.get $len
    call 13
    local.get $blob
    i32.const 0
    call $@deserialize<>
    i32.const 0
    call $@serialize<>
    global.get 19
    global.get 20
    local.set $size
    local.tee $start
    local.get $size
    call 30
    call 31
    i32.const 0
    call 32
    global.set 4
    call $copying_gc <-- the gc call
    i32.const 0
    call 32
    global.get 4
    i64.sub
    global.set 5
    call $trans_state4)
```

luc-blaeser:
Thank you very much. This is much better than my original version.
I prepared a small refinement of this PR: https://github.com/dfinity/motoko/pull/4305.

src/codegen/compile.ml
```diff
@@ -4938,6 +4938,32 @@ module IC = struct
...
     E.else_trap_with env "not a self-call"
 
+  let assert_caller_self_or_controller env =
+    let (set_len1, get_len1) = new_local env "len1" in
+    let (set_len2, get_len2) = new_local env "len2" in
+    let (set_str1, get_str1) = new_local env "str1" in
+    let (set_str2, get_str2) = new_local env "str2" in
+    system_call env "canister_self_size" ^^ set_len1 ^^
+    system_call env "msg_caller_size" ^^ set_len2 ^^
+    get_len2 ^^ Blob.dyn_alloc_scratch env ^^ set_str2 ^^
```
luc-blaeser, line 4948: The only risk that remains is that the GC trigger also allocates dynamic memory and could not work if heap is completely full. I thought that we could may grant the general memory reserve for the GC trigger (https://github.com/dfinity/motoko/pull/4305), but it may pose the risk that (if something would be broken with the GC trigger), a series of GC trigger calls may again eat up that reserve...

test/run-drun/gc-trigger.mo
```diff
@@ -5,7 +5,7 @@ import Prim "mo:prim";
 actor {
     let retained = Prim.Array_init<Nat>(6 * 1024 * 1024, 0);
     // GC is triggered during initialization
-    
+
```
luc-blaeser, line 8: I believe the only adjustment to make this test pass again is to add an extra GC trigger call: https://github.com/dfinity/motoko/pull/4305

luc-blaeser (approved):

src/codegen/compile.ml
crusso: I think I've manage to remove all allocation, by implementing the caller assertion using just stack allocation, not deserializing the trivial argument and only sending a pre-allocated response (not sure that's still pre-allocated in your orthogonal persistence world though). It does mean users can send any old junk payload. One improvement would be to check the argument is literally DIDL\x00\x00 using just stack allocation, but that wouldn't allow for subtyping so is also a hack.

I've added a test for the access control check - wasn't worth refactoring with `assert_calller_self` w/o totally obscuring the logic.

The only thing I'm not sure of is whether you still want to lift the reserve or not, now that this is heap-alloc free. If not, then should be revert the new Lifecycle state? 

Feel free to take over this PR, merge as you see fit (just drop me a message on slack if you do)

test/run-drun/gc-trigger.mo
crusso: Ok, not sure I understand why though - because I removed the second GC call?

luc-blaeser: Yes, this was because the original GC trigger implementation had two GC calls.

src/codegen/compile.ml
luc-blaeser: This is great and I believe safer than using the memory reserve. Thanks a lot.
With this, I can again remove the special memory reserve handling for the GC trigger.

src/codegen/compile.ml
```diff
@@ -1425,6 +1425,17 @@ module Stack = struct
...
+  let dynamic_with_bytes env name f =
+    let (set_n, get_n) = new_local env "n" in
+    let (set_x, get_x) = new_local env name in
+    (* round up to nearest wordsize *)
+    compile_add_const (Int32.sub Heap.word_size 1l) ^^
+    compile_divU_const Heap.word_size ^^
+    set_n ^^
+    dynamic_alloc_words env get_n ^^ set_x ^^
+    f get_x ^^
+    dynamic_free_words env get_n
```
luc-blaeser, line 1437:
```suggestion
    (* round up to nearest wordsize *)
    compile_add_const (Int32.sub Heap.word_size 1l) ^^
    compile_divU_const Heap.word_size ^^
    dynamic_with_words env name f
```

test/run-drun/ok/gc-trigger-acl.drun-run.ok
```diff
@@ -0,0 +1,6 @@
+ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
```
luc-blaeser, line 1: I believe the test source file `gc-trigger-acl.mo` is missing in the change set?
Thanks for having created a test case that checks the permissions.

src/codegen/compile.ml
```diff
@@ -4916,27 +4927,52 @@ module IC = struct
...
+      get_str_self ^^ compile_unboxed_const 0l ^^ get_len_self ^^
+      system_call env "canister_self_copy" ^^
+
+      get_str_caller ^^ compile_unboxed_const 0l ^^ get_len_caller ^^
+      system_call env "msg_caller_copy" ^^
+
+      get_str_self ^^ get_str_caller ^^ get_len_self ^^ Heap.memcmp env ^^
+      compile_eq_const 0l ^^
+      E.else_trap_with env "not a self-call"
+    ))
```
luc-blaeser, line 4947: Not sure whether one could refactor to share the self-call test with both functions. Probably not worth it. It would require two times reading the caller principle...

```suggestion
  let is_self_call env =
    let (set_len_self, get_len_self) = new_local env "len_self" in
    let (set_len_caller, get_len_caller) = new_local env "len_caller" in
    system_call env "canister_self_size" ^^ set_len_self ^^
    system_call env "msg_caller_size" ^^ set_len_caller ^^
    get_len_self ^^ get_len_caller ^^ G.i (Compare (Wasm.Values.I32 I32Op.Eq)) ^^
    G.if1 I32Type
      begin
        get_len_self ^^ Stack.dynamic_with_bytes env "str_self" (fun get_str_self ->
          get_len_caller ^^ Stack.dynamic_with_bytes env "str_caller" (fun get_str_caller ->
            get_str_caller ^^ compile_unboxed_const 0l ^^ get_len_caller ^^
            system_call env "msg_caller_copy" ^^
            get_str_self ^^ compile_unboxed_const 0l ^^ get_len_self ^^
            system_call env "canister_self_copy" ^^
            get_str_self ^^ get_str_caller ^^ get_len_self ^^ Heap.memcmp env ^^
            compile_eq_const 0l))
      end
      begin
        compile_unboxed_const 0l
      end

  let assert_caller_self env =
    is_self_call env ^^
    E.else_trap_with env "not a self-call"
```

src/codegen/compile.ml
```diff
@@ -4916,27 +4927,52 @@ module IC = struct
...
+      begin
+        compile_unboxed_const 0l
+      end ^^
+      G.if0
+        G.nop
+        begin
+          get_str_caller ^^ get_len_caller ^^ is_controller env ^^
+          E.else_trap_with env "not a self-call or call from controller"
+        end
+   ))
```
luc-blaeser, line 4975: Continued from above. As said, probably not worth it. Feel free to ignore.

```suggestion
  let is_controller_call env =
    let (set_len_caller, get_len_caller) = new_local env "len_caller" in
    system_call env "msg_caller_size" ^^ set_len_caller ^^
    get_len_caller ^^ Stack.dynamic_with_bytes env "str_caller" (fun get_str_caller ->
      get_str_caller ^^ compile_unboxed_const 0l ^^ get_len_caller ^^
      system_call env "msg_caller_copy" ^^
      get_str_caller ^^ get_len_caller ^^ is_controller env)
 
  let assert_caller_self_or_controller env =
    is_self_call env ^^
    is_controller_call env ^^
    G.i (Binary (Wasm.Values.I32 I32Op.Or)) ^^
    E.else_trap_with env "not a self-call or call from controller"
```

crusso: Yeah, I was trying to avoid the duplicated call, but couldn't refactor it cleanly. I think it doesn't matter much for gc_trigger so let's go with your version.

test/run-drun/gc-trigger-acl.mo
```diff
@@ -0,0 +1,57 @@
+import Prim "mo:⛔";
```
luc-blaeser, line 1: Nice test case!

PR #4305, opened 2023-11-29
Small Refinements: Implement GC Trigger in Backend

luc-blaeser:
Small refinements for the GC trigger improvement by https://github.com/dfinity/motoko/pull/4303:
* Revert initial non-backend GC trigger changes
* Fix the `gc-trigger` test case
* Allow memory reserve for GC trigger (not sure, to be discussed)

### DIFF
--- a/src/codegen/compile.ml
+++ b/src/codegen/compile.ml
@@ -4419,6 +4419,7 @@ module Lifecycle = struct
     | PostPreUpgrade (* an invalid state *)
     | InPostUpgrade
     | InComposite
+    | InGCTrigger (* grants the general memory reserve for GC trigger *)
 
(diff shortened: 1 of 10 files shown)


crusso (approved):
LGTM - free free to merge both...

Using the reserve seems ideal.

I was actually wondering about skipping the argument deserialization entirely, using stack allocation for the blobs in the assert and replying with static blob literal, but this is much simpler.

crusso:
I was also thinking about refactoring assert_caller_self and assert_caller_self_or_controller into a single function, but couldn't get it right. Now, sleeping on it, I think there is a way using block/breaks for the proper control flow, but perhaps not worth it.

PR #4324, opened 2023-12-14
feat: add repro for issue #4323 and cater for a conformity annotation

ggreif:
reproduce feature request #4323 and implement it

This won't restrict the resulting type ([as pointed out by @rossberg](https://github.com/dfinity/motoko/pull/4327#discussion_r1430238318)), but merely verifies the interface. `public` and type annotations on bindings _inside_ the block are there for shaping the former (just like in `class` definitions).

TODO:
- [x] documentation
- [x] `Changelog.md`
- [ ] resolve old-ish problem now marked with `FIXME` — #4355

Issue #4323, opened 2023-12-14
Type ascription impossible for top-level actors


ggreif:
``` Motoko
actor Main { … } : Types.MainActor
```

is rejected, but should be possible
- to enforce contract
- to avoid publishing too many methods ([as pointed out elsewhere, this should be done by `public`/`private` keywords](https://github.com/dfinity/motoko/pull/4327#discussion_r1430238318))

Another sensible syntax could be
``` Motoko
actor Main : Types.MainActor = { … }
```

This came up in today's "Motoko Developer Working Group" meeting.

ggreif:
It is the `dec_nonvar` production, `s=obj_sort xf=id_opt EQ? efs=obj_body` branch.

crusso:
Yes, something like this would be sensible. 

It should probably be consistent with the semantics of type ascription for classes, which, IIRC, just checks subtyping but doesn't apply it, so unspecified or more general methods shine through.

@rossberg seemed to want that semantics for classes though it seems inconsistent with the static semantics of type annotations. But it's also possible I'm misremembering.

BTW @letmejustputthishere  was asking for this feature on slack and I suggested a temporary workaround.

crusso:
We should of course support this uniformly for actors, object and module literals if we can.

ggreif:
> We should of course support this uniformly for actors, object and module literals if we can.

``` Motoko
> module Moo : module { a : Int } = { public let a = 42 };
let Moo : module {...}
```

ggreif:
Note: `actor Self : <type> {}` is already valid syntax!

PR #4327, opened 2023-12-15
simplify grammar?

crusso:
Tries to simplify grammar by avoiding extra production of #4324, but...

The positive test run-drun/issue-4323.mo fails

I fixed it by extending compUnit.ml, but then don't think this has the right semantics because the hidden methods (see "ohog" ) are still exported because of the way we compile actors.

We would need to filter out the methods or  only allow them to be called from the current actor, but then we can't pass them out as first-class methods that work either...


doc/md/examples/grammar.txt
```diff
@@ -287,15 +284,15 @@
...
 
 <dec_var> ::= 
-    'var' <id> <annot_opt_eq> <exp>
+    'var' <id> (':' <typ>)? '=' <exp>
 
 <dec_nonvar> ::= 
     'let' <pat> '=' <exp>
     'type' <id> ('<' <list(<typ_bind>, ',')> '>')? '=' <typ>
-    <obj_sort> <id>? '='? <obj_body>
+    <obj_sort> <id>? (':' <typ>)? '='? <obj_body>
```
ggreif, line 295: this admits `: module {} {}` which I distaste :-(

crusso: Oh, I see. Yeah, I guess that is consfusing.

test/run-drun/issue-4323.mo
```diff
@@ -0,0 +1,10 @@
+import Prim = "mo:⛔";
+// let-like annotation
+actor Main : actor { beep : () -> async () } {
+
+  public func beep() : async () {};
+
+  public func ohoh() : async () {Prim.debugPrint "ohoh"}; // still exported, despite not in declared interface!
+
+}
+//CALL ingress ohoh "DIDL\x00\x00"
```
crusso, line 10:
```suggestion
//CALL ingress ohoh "DIDL\x00\x00"

```

test/run-drun/issue-4323.mo
```diff
@@ -0,0 +1,10 @@
+import Prim = "mo:⛔";
+// let-like annotation
+actor Main : actor { beep : () -> async () } {
+
+  public func beep() : async () {};
+
+  public func ohoh() : async () {Prim.debugPrint "ohoh"}; // still exported, despite not in declared interface!
```
ggreif, line 7: Not sure if this is a bug or a feature. What is happening with expressions `module {public foo() {}} : { /* empty */ }`

ggreif: Hmm, type ascription does seem to hide public stuff (contrary to what you mention in #4323) :
``` Motoko
> if true {object Foo {public func foo(){}}; Foo : {}};
stdin:4.1-4.53: warning [M0081], this if has type
  Any
because branches have inconsistent types,
true produces
  {}   👈
false produces
  ()
{foo = func} : Any
```

crusso: Yes, that's true `<exp> : <typ>` does restrict the type as you say. But in class declarations, the type annotation on the body does not, which I find odd.

```
actor {

    class C() : object {} // this ascription only checks conformance, doesn't restrict
    {
      public func get () {};
    };

    let c = C();
    c.get(); 

}
```
https://play.motoko.org/?tag=1113634911

crusso: More evidence:

```
actor {

    class C() : object {
      f: () -> Any // note weaker return type
    } // this ascripton only checks conformance, doesn't restrict
    {
      public func get () {};
      public func f() : Nat { 0 };
    };

    let c = C();
    c.get();
    let _ : Nat = c.f();

}
```

https://play.motoko.org/?tag=1445447127

Here. `f` is exported at its actual, not the weaker, specified, type.

doc/md/examples/grammar.txt
ggreif: This form is valid too, so we can choose:
``` Motoko
> module class M() : module { a : Int } = { public let a = 42 };
type M__6 = module {a : Nat}
```

test/run-drun/issue-4323.mo
ggreif: Yeah, this is probably something we should fix soon.

rossberg: The point of only checking conformance is that this ought to work like an `implements` clause, which is what you normally expect and care about for classes and objects. For hiding you already have public/private in Motoko, so no point in having a duplicate mechanism for that. One way in which Motoko is not ML. :)

So I agree that object/actor/module literals should get an annotation, but that should also be a conformance check only.

crusso: Isn't that a bit dangerous though? People might expect hiding, given the behaviour of type annotations, with the same colon syntax, on expressions, but then be bitten when, e.g. third party nevertheless successfully call private methods or see extra result info...

crusso: > One way in which Motoko is not ML. :)

You say that like it's a good thing ;->

crusso: I added a test run-drun/issue-4323-class.mo that shows the current desugaring is inconistent with the way we type classes (where the this reference has the full type, not the annotated type).

One (ugly) fix would be to change the desugaring to insert a dummy expression that checks the type of `this`, without tripping over the definedness check.
```
import Prim = "mo:⛔";
// let-like annotation
actor This  {
  ignore func () : actor { beep : () -> async () }  { This };
  public func beep() : async () {};

  public func ohoh() : async () {Prim.debugPrint "ohoh"}; // still exported, despite not in declared interface!

  public func check() : async ()  { await This.ohoh() }; // check no curtailment this

}
//CALL ingress ohoh "DIDL\x00\x00"
```
The other would be to extend the syntax of object literals to carry an optional type annotation an use it appropriately in the type checker.

rossberg: > People might expect hiding

I believe that would mostly be ML/FP-biased people. More people probably expect "inheritance" in this case, given the precedence of a colon for that in C++ and others. And "structurally typed interface inheritance" is what they get.

I agree it's not perfect, but like other things in Motoko a compromise with mainstream expectations and "familiarity".

(I also fear that sealing semantics are not a good fit for actors, where static type annotations do not actually hide public entry points, which can still be used from the outside. So such a semantics may even produce a wrong sense of encapsulation.)

> You say that like it's a good thing ;→

Neither, just "different". :)

ggreif:
Closing, as #4324 has landed.

PR #4335, opened 2023-12-19
bugfix: fix Region.storeNat32/loadNat32 and Region.storeInt32/loadInt32 

crusso:
bugfix: fix broken implementations of `Region.loadNat32`, `Region.storeNat32`, `Region.loadInt32`, `Region.storeInt32` (#4335). The operations were writing the 32-bit vanilla representation of the Motoko values, not the decoded values. 

Values previously stored with the broken 32-bit operations must be loaded with care.
If bit 0 is clear, the original value can be obtained by an arithmetic shift right by 1 bit.
If bit 0 is set, the value cannot be trusted and should be ignored (it encodes some transient address of a boxed value).

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,13 @@
 # Motoko compiler changelog
 
+* motoko (`moc`)
+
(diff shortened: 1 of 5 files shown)


luc-blaeser (approved):
Thanks for fixing this. 
Luckily, you found this issue.

PR #4346, opened 2024-01-05
extending `ObjBlockE` with optional conformity

ggreif:
Use the same procedure as with `ClassD`, viz. adding an optional type to `ObjBlockE` and checking for it after inferring the body's type.

### DIFF
--- a/doc/md/examples/grammar.txt
+++ b/doc/md/examples/grammar.txt
@@ -295,7 +295,7 @@
 <dec_nonvar> ::= 
     'let' <pat> '=' <exp>
     'type' <id> ('<' <list(<typ_bind>, ',')> '>')? '=' <typ>
-    <obj_sort> <id>? '='? <obj_body>
+    <obj_sort> <id>? <annot_opt_eq>? <obj_body>
     <shared_pat_opt> 'func' <id>? ('<' <list(<typ_bind>, ',')> '>')? <pat_plain> (':' <typ>)? <func_body>
     <shared_pat_opt> <obj_sort>? 'class' <id>? ('<' <list(<typ_bind>, ',')> '>')? <pat_plain> (':' <typ>)? <class_body>
 
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -826,7 +826,7 @@ let rec is_explicit_exp e =
   | ObjE (bases, efs) ->
     List.(for_all is_explicit_exp bases
           && for_all (fun (ef : exp_field) -> is_explicit_exp ef.it.exp) efs)
-  | ObjBlockE (_, dfs) ->
+  | ObjBlockE (_, _, dfs) ->
     List.for_all (fun (df : dec_field) -> is_explicit_dec df.it.dec) dfs
   | ArrayE (_, es) -> List.exists is_explicit_exp es
   | SwitchE (e1, cs) | TryE (e1, cs) ->
@@ -1135,7 +1135,7 @@ and infer_exp'' env exp : T.typ =
         "expected tuple type, but expression produces type%a"
         display_typ_expand t1
     )
-  | ObjBlockE (obj_sort, dec_fields) ->
+  | ObjBlockE (obj_sort, _t, dec_fields) ->
     if obj_sort.it = T.Actor then begin
       error_in [Flags.WASIMode; Flags.WasmMode] env exp.at "M0068"
         "actors are not supported";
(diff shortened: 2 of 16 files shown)


doc/md/examples/grammar.txt
```diff
@@ -295,7 +295,7 @@
 <dec_nonvar> ::= 
     'let' <pat> '=' <exp>
     'type' <id> ('<' <list(<typ_bind>, ',')> '>')? '=' <typ>
-    <obj_sort> <id>? '='? <obj_body>
+    <obj_sort> <id>? <annot_opt_eq>? <obj_body>
```
ggreif, line 298: the big PR will revert this

src/mo_frontend/typing.ml
```diff
@@ -1150,7 +1150,18 @@ and infer_exp'' env exp : T.typ =
...
       else env
     in
-    infer_obj env' obj_sort.it dec_fields exp.at
+    let t = infer_obj env' obj_sort.it dec_fields exp.at in
+    begin match typ_opt with
+      | Some typ ->
+        let t' = check_typ env' typ in
+        if not (T.sub t t') then
+          local_error env exp.at "M0192"
+            "body of type%a\ndoes not match expected type%a"
```
ggreif, line 1159: I don't try to spell out which sort of body it is, as it should be obvious from the inferred type:
``` Motoko
stdin: type error [M0192], body of type
  module {go : () -> ()}
does not match expected type
  module {got : () -> ()}
```

src/mo_frontend/typing.ml
```diff
@@ -2434,7 +2445,7 @@ and infer_dec env dec : T.typ =
             display_typ_expand t'
             display_typ_expand t''
       | Some typ, T.Actor ->
-        local_error env dec.at "M0135" "actor class has non-async return type"
+         local_error env dec.at "M0135"(*FIXME: repeated?*) "actor class has non-async return type"
```
ggreif, line 2448: This is not caught by CI as `M0135` already appears above, but the error string never occurs in the test suite.

test/fail/ok/issue-4323.tc.ok
```diff
@@ -1,4 +1,4 @@
-issue-4323.mo:4.1-6.2: type error [M0096], expression of type
+issue-4323.mo:4.1-6.2: type error [M0192], body of type
   actor {}
-cannot produce expected type
-  actor {beep : shared () -> async ()}
+does not match expected type
+  Main = actor {beep : shared () -> async ()}
```
ggreif, line 4: The `Main =` is due to the conformity being a typedef.

test/fail/ok/issue-4323.tc.ok
```diff
@@ -1,4 +1,4 @@
-issue-4323.mo:4.1-6.2: type error [M0096], expression of type
+issue-4323.mo:4.1-6.2: type error [M0192], body of type
```
ggreif, line 1: Currently the source range extends to the whole thing, would you insist on restricting to the body only?

src/mo_frontend/typing.ml
```diff
@@ -1150,7 +1150,18 @@ and infer_exp'' env exp : T.typ =
         { env with async = C.NullCap; in_actor = true }
       else env
     in
-    infer_obj env' obj_sort.it dec_fields exp.at
+    let t = infer_obj env' obj_sort.it dec_fields exp.at in
+    begin match typ_opt with
```
crusso, line 1154: I wonder if you only want to do the check in the second pass, something like:

```suggestion
   if not (env.pre) then begin match typ_opt with
```
Otherwise you might get repeated local errors on certain occasions. (The type checker has become a mystery to me, tbh).

src/mo_frontend/typing.ml
crusso: Good catch!

test/fail/ok/issue-4323.tc.ok
crusso: Probably doesn't matter match, tbh

src/mo_frontend/typing.ml
crusso, line 1159:
```suggestion
            "body of type%a\ndoes not match expected type%a"
```
Should we call this the `annotated` (not expected) type? Or maybe the `expected interface` type? Then again, I don't we thing we are that informative for the class type annotation either.

src/mo_frontend/typing.ml
```diff
@@ -1150,7 +1150,18 @@ and infer_exp'' env exp : T.typ =
...
+      | Some typ ->
+        let t' = check_typ env' typ in
```
ggreif, line 1156: I am not super confident with this `exp'`, but tried to model this code as close to the `class` check as possible.

test/fail/ok/issue-4323.tc.ok
ggreif: This is like `M0134` treats it as well.

crusso (commented):
Looks ok to me, but will need a change to the manual as well...

crusso (approved):
Approved since you are only merging into another PR, right?

src/mo_frontend/typing.ml
crusso: Yes, I'm not sure what's already in that env' to be honest. For example, does it contain the actor identifier (probably), or  private types defined within the body (hopefully not)?

src/mo_frontend/typing.ml
ggreif: Right. Either this way, or we change both.

src/mo_frontend/typing.ml
ggreif: a81acec6a

PR #4348, opened 2024-01-08
Update `drun` to More Recent Version

luc-blaeser:
Based on `IC` release 2023-12-13.

Adjustments for `nix` build:
* Fix revisions for `ic-btc` dependencies, see https://github.com/dfinity/ic/pull/124
* Patch `rocksdb` to build on MacOS `darwin`.

IC new metering has an effect on the benchmark results.


nix/drun.nix
```diff
@@ -18,29 +18,28 @@ pkgs:
...
-         config.define("ROCKSDB_LIB_IO_POSIX", Some("1"));
-+        if target.contains("aarch64") {
-+            config.define("isSSE42()", Some("0"));
-+            config.define("isPCLMULQDQ()", Some("0"));
-+        }
-     } else if target.contains("android") {
-         config.define("OS_ANDROID", Some("1"));
-         config.define("ROCKSDB_PLATFORM_POSIX", Some("1"));
+@@ -249,6 +249,9 @@ fn build_rocksdb() {
+         config.flag("-Wno-missing-field-initializers");
```
ggreif, line 36: So RocksDB now finally supports the M1?

ggreif (approved):
Looks great!

nix/drun.nix
luc-blaeser: This change actually only supports MacOS with x64 emulation. 
We should also test it on M1 natively...

ggreif: Seems to work!

luc-blaeser: Thanks a lot!

PR #4355, opened 2024-01-10
chore: remove the `FIXME` for duplicated `M0135`

ggreif:
Since the parser already ensures that the return type of actor classes is `AsyncT`, the only way how this can be triggered is by having `T.Cmp`.


crusso (approved):
LGTM

PR #4098, opened 2023-07-04
experiment: tag small words

crusso:
Add RTTI to free bits of unboxed small works, just to get a feel of overhead.

- [x]  there something still wrong with **
- [x] logical operations
- [x] comparisons
- [ ]  some filecheck tests fail, probably because there are looking for incorrect literals now. The tests need to be updated before any merge
- [x] exotics like clz8/16 etc are implemented using overloading of primitives that now need distinct code (to add remove/tags), unless we just save and restore instead. This seems to be the root of the failure of base Int16 bittest failures.
- [ ] char tagging
- [ ] https://github.com/dfinity/motoko/pull/4139 adds numeric conversion between adjacent scalars, but assumes tag bits are clear. Will need adjusting to strip and add them instead.

PR #4357, opened 2024-01-10
experiment: tagging all scalars (merged with master)

crusso:
The Motoko runtime representation of values is largely untyped, distinguishing only between scalar and boxed values 
 using a single bit of the 32-bit value representation.  The tagging is only to support garbage collection, not precise runtime type information.

In the existing value encoding, a Motoko value in vanilla form is a  32-bit value that is either:
* false `(0b0)`,
* true `(0b1)`, 
* a word-aligned (encoded) pointer to a heap allocated value.
  Encoded by subtracting 1 from the pointer value (ensuring the 2 LSBs are 0b11), pointing
  heap allocated value
* null (some well-known skewed pointer).
* a 31-bit scalar value, stored in the top bits of the value with LSB 0.

Scalar values encode `Nat8/16` and `Int8/16` values and chars, and 31-bit subranges of `Nat32`, `Int32`, `Nat64`, `Int64`, `Nat` and `Int`. Large integer values that don't fit  in a 31-bit scalar are boxed on the heap.

Observe that, in Motoko, some types are always scalar (eg. `Nat8`), some types are always boxed (e.g. `Blob`), and some types have a mixed scalar/boxed representation (e.g. `Nat32` and `Nat`), depending on the size of the value.

This PR adds exact runtime type information to all[*] scalar values, making the scalar values self describing.
Making the heap fully self-describing requires refining the heap tags use to identify heap objects, distinguishing boxed `Nat32` from boxed `Int32`, `Blob` from `Principal` and `Text`, tuples from (mutable and immutable) arrays etc. That work of refining heap tags will need to be completed in a follow on or sibling PR, but is hopefully less involved than the changes herein.

To add scalar type info, we extend the scalar tagging scheme with a richer set of (inline) type descriptors, using some of the least significant bits of the 31-bit scalar representation.

To avoid dedicating a fix-length suffix to the scalar tag, scalar tags are actually variable length, using shorter tags for larger payload types, and longer tags for shorter payload types. This gives us a reasonable tag space (set of possible tags), without   reducing the scalar range of mixed representation types too much.

At one extreme, the tag of `Int` (and `Nat`) is just `0b10`, leaving a 30-bit payload for compact bignums, losing just `1` bit from the current representation's 31-bit compact range. This is important because `Int`s are common, and `Nat`s are used to index arrays, so we should avoid boxing more than necessary.

In the middle, the tag of `Nat16`, `Int16` is `0b10(0^12)00` and `0b11(0^12)00`, leaving a 16-bit payload in the MSB.

At the other extreme, tag of the unit value, `()`, is 32-bit `0x01(0^28)00`, occupying the entire value.

The primary motivation is to support value, not type driven, serialization of stable values to a precisely typed stable format, without loss of type information, so that upgrades can still accommodate type dependent changes of representation from one in-memory format to another. Secondary motivations are live and post-mortem heap inspection tools and light-weight debugging tools, that can parse values in locals, arguments and on the heap using tags.

[*] There remain some untagged 31-bit scalars. These are used to encode the state of text and blob iterators, hidden in dedicated iterator closure environments. Not that these are not stable types, so need not be precisely tagged for stabilization.

# Tagging Scheme

   | Value | Type | Payload bits |
   |-------| ------| --------------|
   | `((O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O))` | TBool (* false *) | 0 |
   | `((O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,I))`   | TBool (* true *) | 0 |
   | `((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,I,I))`                                         | TRef    |    30       |  
   | `((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,I,O))`                                       | TNum |  30        |
   | `((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,O,I,O,O))`                                     | TNat64 | 28       |
   | `((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,I,I,O,O))`                                       |  TInt64  |  28      |
   | `((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,O,I,O,O,O))`                                    | TNat32 |  27      |
   | `((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,I,I,O,O,O))`                                     | TInt32   | 27        |
   | ... unused tags ....                             | ...   | ...       |
   | `((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,O,I,O), (O,O,O,O,O,O,O,O))` | TChar | 21 |
  | ... unused tags ....                             | ...   | ...       |
   | `((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (O,I,O,O,O,O,O,O), (O,O,O,O,O,O,O,O))`| TNat16 | 16 |
   | `((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (I,I,O,O,O,O,O,O), (O,O,O,O,O,O,O,O))` | TInt16 | 16 |
  | ... unused tags ....                             | ...   | ...       |
   | `((_,_,_,_,_,_,_,_), (O,I,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O))` | TNat8 | 8 |
   | `((_,_,_,_,_,_,_,_), (I,I,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O))` | TInt8 | 8 |
  | ... unused tags ....                             | ...   | ...       |
   | `((O,I,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O))` |  TUnit | 0 |

# Implementation

The implementation was carried out in a number of precursor PRs:

#4098: added tags to small values.
#4278: made the payload size for scalar values configurable using a fixed compile time constant.
#4279: added tags to compact `Nat/32/64` and `Int/32/64`, making the payload size type dependent.
#4279: extended the range of compact `Int/Nat` from 29 to 30-bit.
#4354: improved the tagging scheme to use the longest possible tags for the required payload size, 
            upping the ranges of unused tags (for future use)
#4357: (this PR) fix bugs in sanity checking of tags, revealing more bugs and fixing them.

# Overheads

Cycle counts differences measured using `test/bench` `test/perf`, against master.

## perf

Master |   |   | Branch |   |  
-- | -- | -- | -- | -- | --
gas/assetstorage | 10,013,950.00 |   | gas/assetstorage | 10,013,950.00 | 0.00%
size/assetstorage | 187,086.00 |   | size/assetstorage | 187,368.00 | 0.15%
gas/dao | 4,413,634,512.00 |   | gas/dao | 4,413,756,015.00 | 0.00%
size/dao | 268,104.00 |   | size/dao | 269,092.00 | 0.37%
gas/qr | 1,302,744,564.00 |   | gas/qr | 1,306,084,636.00 | 0.26%
size/qr | 256,683.00 |   | size/qr | 257,603.00 | 0.36%
gas/reversi | 80,920,993.00 |   | gas/reversi | 81,291,914.00 | 0.46%
size/reversi | 176,357.00 |   | size/reversi | 176,863.00 | 0.29%
gas/sha256 | 14,487,064,533.00 |   | gas/sha256 | 15,569,983,217.00 | 7.48%
size/sha256 | 179,736.00 |   | size/sha256 | 181,078.00 | 0.75%

## bench
Master |   |   | Branch |   |  
-- | -- | -- | -- | -- | --
gas/alloc | 9,243,068,120.00 |   | gas/alloc | 10,353,374,958.00 | 12.01%
size/alloc | 181,066.00 |   | size/alloc | 181,621.00 | 0.31%
gas/bignum | 130,604,743.00 |   | gas/bignum | 130,608,493.00 | 0.00%
size/bignum | 184,420.00 |   | size/bignum | 184,959.00 | 0.29%
gas/heap-32 | 1,610,218,447.00 |   | gas/heap-32 | 1,696,355,149.00 | 5.35%
size/heap-32 | 182,167.00 |   | size/heap-32 | 182,695.00 | 0.29%
gas/nat16 | 61,393,031.00 |   | gas/nat16 | 111,725,211.00 | 81.98%
size/nat16 | 181,010.00 |   | size/nat16 | 181,595.00 | 0.32%
gas/palindrome | 10,131,340.00 |   | gas/palindrome | 10,133,674.00 | 0.02%
size/palindrome | 185,338.00 |   | size/palindrome | 185,883.00 | 0.29%
gas/region0-mem | 6,402,149,937.00 |   | gas/region0-mem | 6,704,153,332.00 | 4.72%
size/region0-mem | 181,898.00 |   | size/region0-mem | 182,444.00 | 0.30%
gas/region-mem | 5,974,331,587.00 |   | gas/region-mem | 6,276,334,982.00 | 5.06%
size/region-mem | 181,539.00 |   | size/region-mem | 182,085.00 | 0.30%
gas/stable-mem | 3,885,566,188.00 |   | gas/stable-mem | 4,187,556,443.00 | 7.77%
size/stable-mem | 181,896.00 |   | size/stable-mem | 182,442.00 | 0.30%

The largest overhead is seen in `Nat16.mo`, which does loads of `Nat16` arithmetic in a tight loop.
This overhead is greatly *reduced* when the test is rewritten using `Nat32`. 

Master |   |   | Branch |   |  
-- | -- | -- | -- | -- | --
gas/nat16 | 61,393,031.00 |   | gas/nat16 | 111,725,211.00 | 81.98%
size/nat16 | 181,010.00 |   | size/nat16 | 181,595.00 | 0.32%
gas/xxx-nat32 | 57,198,791.00 |   | gas/xxx-nat32 | 59,296,369.00 | 3.67%
size/xxx-nat32 | 181,001.00 |   | size/xxx-nat32 | 181,491.00 | 0.27%

Why? Compact `Nat32`s are represented with stack representation `UnboxedWord32 Nat32` on the stack, suggesting that a similar untagged stack representation `UnboxedWord32 Nat8` (not `Vanilla`) for tagged `Nat8` and other small tagged values might reduce the new tagging overhead considerably. Worth investigating, especially once we move to 64-bit where using 64-bit vanilla stack slots for small values seems wasteful.

# TODOs

[ ] Some rts functions still return untagged 31-bit scalars (eg. bigint_of_float) which are, for now, correctly fixed up at the call-sites in compile.ml.
[ ] BitTagged.can_tag_i32 and ilk are not as clever as they could be and do a naive range check rather than funky XOR/CTZ tricks


rts/motoko-rts/src/bigint.rs
```diff
@@ -261,6 +261,7 @@ unsafe extern "C" fn bigint_of_float64(j: f64) -> Value {
     // can be represented as `Int` without resorting to heap allocation, i.e.
     // in the range `-1073741824 == 0xc0000000 <= j as i32 <= 0x3fffffff == 1073741823`
     if j < 1073741824.0 && j > -1073741825.0 {
+        // TODO: FIX
```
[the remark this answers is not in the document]
crusso: Actually, it's ok. This function returns the value as untagged 31-bit scalar when possible, and the caller in compile.ml takes care of boxing it further if necessary. Clunky but ok.

crusso, line 264:
```suggestion
        // TODO: return as 30-bit compact tagged Int (not 31-bit untagged scalar) and fix call-site in compile.ml
```

src/codegen/compile.ml
```diff
@@ -11119,15 +11592,26 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
     const_sr SR.Vanilla (TaggedSmallWord.btst_kernel env Type.Nat16)
+  | OtherPrim "btstInt8", [_;_] ->
+    const_sr SR.Vanilla (TaggedSmallWord.btst_kernel env Type.Int8)
+  | OtherPrim "btstInt16", [_;_] ->
+    const_sr SR.Vanilla (TaggedSmallWord.btst_kernel env Type.Int16)
   | OtherPrim "btst32", [_;_] ->
-    const_sr SR.UnboxedWord32 (TaggedSmallWord.btst_kernel env Type.Nat32)
+     const_sr (SR.UnboxedWord32 Type.Nat32) (TaggedSmallWord.btst_kernel env Type.Nat32)
+  | OtherPrim "btstInt32", [_;_] ->
+    const_sr (SR.UnboxedWord32 Type.Int32) (TaggedSmallWord.btst_kernel env Type.Nat32)
```
crusso, line 11602:
```suggestion
    const_sr (SR.UnboxedWord32 Type.Int32) (TaggedSmallWord.btst_kernel env Type.Int32)
```
this is a (benign) bug

PR #4361, opened 2024-01-12
Incremental Graph-Copy-Based Upgrades

crusso:
# Incremental Graph-Copy-Based Upgrades

Refinement of https://github.com/dfinity/motoko/pull/4286

Supporting arbitrarily large graph-copy-based upgrades beyond the instruction limit:
* Splitting the stabilization/destabilization in multiple asynchronous messages.
* Limiting the stabilization work units to fit the update or upgrade messages.
* Blocking other messages during the explicit incremental stabilization.
* Restricting the upgrade functionality to the canister owner and controllers.
* Stopping the GC during the explicit incremental upgrade process.

## Usage

For large upgrades:
1. Initiate the explicit stabilization before the upgrade:
    
```
```

* An assertion first checks that the caller is the canister owner or a canister controller.
* All other messages to the canister will be blocked until the upgrade has been successfully completed.
* The GC is stopped.
* If defined, the actor's pre-upgrade function is called before the explicit stabilization.
* The stabilzation runs in possibly multiple asynchronous messages, each with a limited number of instructions.

2. Run the actual upgrade:

```
dfx deploy CANISTER_ID
```

* Run and complete the stabilization if not yet done in advance. 
* Perform the actual upgrade of the canister on the IC.
* Start the destabilization with a limited number of steps to fit into the upgrade message.
* If destabilization cannot be completed, the canister does not start the GC and does not accept messages except step 3.

3. Complete the explicit destabilization after the upgrade:

```
```

* An assertion checks that the caller is the canister owner or a canister controller.
* All other messages remain blocked until the successful completion of the destabilization.
* The destabilzation runs in possibly multiple asynchronous messages, each with a limited number of instructions.
* If defined, the actor's post-upgrade function is called at the end of the explicit destabilization.
* The GC is restarted.

## Remarks

* Steps 1 (explicit stabilization) and/or 2 (explicit destabilization) may not be needed if the corresponding operation fits into the upgrade message.
* Stabilization and destabilization steps are limited to the increment limits:

    Operation | Message Type | IC Instruction Limit | **Increment Limit**
    ----------|--------------|----------------------|--------------------
    **Explicit (de)stabilization step** | Update | 20e9 | **16e9**
    **Actual upgrade** | Upgrade | 200e9 | **160e9**

* The stabilization code in the RTS has been restructured to be less monolithic.


src/codegen/compile.ml
```diff
@@ -9112,6 +8945,47 @@ module FuncDec = struct
...
+        IC.assert_caller_self_or_controller env ^^
+        (* Skip argument deserialization to avoid allocations *)
+        GraphCopyStabilization.stabilization_increment env ^^
+        (G.if0
+          G.nop
+          (call_async_stabilization env)) ^^
+        (* Send static reply *)
+        Blob.lit_ptr_len env "DIDL\x00\x00" ^^
+        IC.reply_with_data env
+        (* Skip garbage collection *)
```
crusso, line 8977: I think this should just exit, with replying. Instead, the call_back should test if there is more work to be done and either do it and issue another call with same callback and exit, or (if done), issue a reply and exit. Not sure about the error handling but it could probably just reject with the error message and exit.

luc-blaeser: Thank you for the suggestion. I implemented it in this way.

luc-blaeser:
Trying to reopen this PR to trigger the usual CI steps...

PR #4363, opened 2024-01-14
experiment: used untagged stack rep for TaggedSmallWords to avoid tagging overhead in numeric ops

crusso:
Experiment with untagged stack rep for small words to avoid tagging overhead in most operations.

Improves perf and simplifies arithmetic code back to zero-padded conventions.

Perf https://docs.google.com/spreadsheets/d/1zC2Hsl9gGUzJESQmSABPiu-XIsICEw1I3O-JKHNWVQs/edit?usp=drivesdk

PR #4369, opened 2024-01-24
feat: precise tagging of scalar values

crusso:
The Motoko runtime representation of values is largely untyped, distinguishing only between scalar and boxed values 
 using a single bit of the 32-bit value representation.  The tagging is only to support garbage collection, not precise runtime type information.

In the existing value encoding, a Motoko value in vanilla form is a  32-bit value that is either:
* false `(0b0)`,
* true `(0b1)`, 
* a word-aligned (encoded) pointer to a heap allocated value.
  Encoded by subtracting 1 from the pointer value (ensuring the 2 LSBs are 0b11), pointing
  heap allocated value
* null (some well-known skewed pointer).
* a 31-bit scalar value, stored in the top bits of the value with LSB 0.

Scalar values encode `Nat8/16` and `Int8/16` values and chars, and 31-bit subranges of `Nat32`, `Int32`, `Nat64`, `Int64`, `Nat` and `Int`. Large integer values that don't fit  in a 31-bit scalar are boxed on the heap.

Observe that, in Motoko, some types are always scalar (eg. `Nat8`), some types are always boxed (e.g. `Blob`), and some types have a mixed scalar/boxed representation (e.g. `Nat32` and `Nat`), depending on the size of the value.

This PR adds exact runtime type information to all[*] scalar values, making the scalar values self describing.
Making the _entire_ heap fully self-describing requires refining the heap tags use to identify heap objects, distinguishing boxed `Nat32` from boxed `Int32`, `Blob` from `Principal` and `Text`, tuples from (mutable and immutable) arrays etc. That work of refining heap tags will need to be completed in a follow on or sibling PR, but is hopefully less involved than the changes herein.

To add precise scalar type info, we extend the scalar tagging scheme with a richer set of (inline) type descriptors, using some of the least significant bits of the 31-bit scalar representation.

To avoid dedicating a fix-length suffix (say 1 byte) to the scalar tag, scalar tags are actually variable length, using shorter tags for larger payload types, and longer tags for shorter payload types. This gives us a reasonable tag space (set of possible tags, some still unused), without  reducing the scalar range of mixed representation types too much.

At one extreme, the tag of `Int` (and `Nat`) is just `0b10`, leaving a 30-bit payload for compact `Nat/Int`, losing just `1` bit from the current representation's 31-bit compact range. This is important because `Int`s are common, and `Nat`s are used to index arrays, so we should avoid boxing more than necessary.

In the middle, the tag of `Nat16`, `Int16` is `0b10(0^12)00` and `0b11(0^12)00`, leaving a 16-bit payload in the MSB.

At the other extreme, the tag of the unit value, `()`, is 32-bit `0x01(0^28)00`, occupying the entire value.

The primary motivation of this work is to support value, not type driven, serialization of stable values to a precisely typed stable format, without loss of type information, so that upgrades can still accommodate type dependent changes of representation from one in-memory format to another. Secondary motivations are live and post-mortem heap inspection tools and light-weight debugging tools, that can parse values in locals, arguments and on the heap using tags.

[*] There remain some raw, untagged 31-bit scalars whose type is only known to the compiler. These are used to encode the state of text and blob iterators, hidden in dedicated iterator closure environments. Note that these are not stable types, so need not be precisely tagged for stabilization.

# Tagging Scheme

   | Value | Type | Payload bits |
   |-------| ------| --------------|
   | `((O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O))` | TBool (* false *) | 0 |
   | `((O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,I))`   | TBool (* true *) | 0 |
   | `((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,I,I))`                                         | TRef    |    30       |  
   | `((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,I,O))`                                       | TNum |  30        |
   | `((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,O,I,O,O))`                                     | TNat64 | 28       |
   | `((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,I,I,O,O))`                                       |  TInt64  |  28      |
   | `((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,O,I,O,O,O))`                                    | TNat32 |  27      |
   | `((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,I,I,O,O,O))`                                     | TInt32   | 27        |
   | ... unused tags ....                             | ...   | ...       |
   | `((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,O,I,O), (O,O,O,O,O,O,O,O))` | TChar | 21 |
  | ... unused tags ....                             | ...   | ...       |
   | `((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (O,I,O,O,O,O,O,O), (O,O,O,O,O,O,O,O))`| TNat16 | 16 |
   | `((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (I,I,O,O,O,O,O,O), (O,O,O,O,O,O,O,O))` | TInt16 | 16 |
  | ... unused tags ....                             | ...   | ...       |
   | `((_,_,_,_,_,_,_,_), (O,I,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O))` | TNat8 | 8 |
   | `((_,_,_,_,_,_,_,_), (I,I,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O))` | TInt8 | 8 |
  | ... unused tags ....                             | ...   | ...       |
   | `((O,I,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O))` |  TUnit | 0 |

# Implementation

The implementation was carried out in a number of precursor PRs:

* #4098: Added 1-byte tags to small  values, untagging an retagging on every operation, with many code changes.
* #4278: Made the payload/tag size for scalar values configurable using a fixed compile time constant.
* #4322: Added tags to compact `Nat32/Int32` and `Nat32/Nat64`, making the payload size type-dependent.
            The previously untyped _StackReps_ `UnboxedWord32` and `UnboxedWord64` were extended to carry a type 
argument. The argument is used to remember and re-introduce the precise tag on unboxing and boxing. 
            It can also be used to verify the tag on unboxing, for sanity checking.
* #4345: Tag compact Int and Nat (both as Int due to subtyping) 
* #4353: Extended the range of compact `Int/Nat` from 29 to 30-bit, by adjusting the tagging scheme. This is just 1 bit less
            than with the existing scheme (31-bit, untagged scalars).
* #4354: Improved the tagging scheme to use the longest possible tags for the required payload size, upping the ranges of unused tags (for future use)
* #4357: Merge with master, fixs bugs in sanity checking of tags. Fix bugs revealing by more stringent sanity checks.
* #4363: Uses the `UnboxedWord32/Word64` stack reps also for untagged, 0-right-padded  small tagged values, 
           tagging/untagging only on exit to and from stack. 
           This alone reduces the (large) 80% overhead in bench/nat16.mo to 55%.
           It also has the advantage of reverting almost all changes to the arithmetic code, 
           which can now (again) assume values are right, 0-padded as it did previously,
* #4369: (this PR) does a small tweak so that mutable locals containing small tagged values in untagged form, extending
           the existing optimization done for mutable locals containing unboxed `Nat32`/`Int32` and `Int64`/`Nat64`.
           This reduces the `bench/nat16.mo` overhead from 55% to just 6% (the benchmark use repeated in-place updates in a tight loop so benefits greatly).
           This PR also makes use of the previously unused bit in the the compact representation of `Nat32s` and `Nat64s` which previously had to concur with the representation of `Int32` and `Nat64` and could only represent half the unsigned range.
           With the typed StackRep, we now know whether the values are signed or not and can choose distinct compact
           representation for `Nat32` vs `Int32`, and `Nat64` vs `Int64` rather that shared ones.
           Note however, that the compact representation for `Nat` cannot recover the missing bit because of subtyping.
           A compact `Nat` **must** have the same representation as a compact `Int` to support non-coercive subtyping.
* #4375 (incoming): rewrite array iter optimization to respect compact bignum representation invariants.
* #4400 : gate feature behind `Mo_config.Flags.rtti (default off)`, avoiding overhead for now.
* added (unadvertised) flag `--experimental-rtti` to enable feature for performance feedback from users.

# Overheads

These are the cycle count and code size differences measured using `test/bench` and  `test/perf`, compared against master (see spreadsheet for perf of interim PRs).

Summarized from:

https://docs.google.com/spreadsheets/d/1zC2Hsl9gGUzJESQmSABPiu-XIsICEw1I3O-JKHNWVQs/edit?usp=sharing

## test/perf

Master |   |   | Widening |   | Widening vs Master |   | Gated |   | Gated vs Master
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
gas/assetstorage | 10013950 |   | gas/assetstorage | 10013950 | 0.00% |   | gas/assetstorage | 10013950 | 0.00%
size/assetstorage | 186455 |   | size/assetstorage | 186705 | 0.13% |   | size/assetstorage | 186520 | 0.03%
gas/dao | 4413634512 |   | gas/dao | 4413744976 | 0.00% |   | gas/dao | 4413743944 | 0.00%
size/dao | 265797 |   | size/dao | 266385 | 0.22% |   | size/dao | 265922 | 0.05%
gas/qr | 1302744688 |   | gas/qr | 1305067118 | 0.18% |   | gas/qr | 1302750018 | 0.00%
size/qr | 256049 |   | size/qr | 256925 | 0.34% |   | size/qr | 256285 | 0.09%
gas/reversi | 80920993 |   | gas/reversi | 81019001 | 0.12% |   | gas/reversi | 80927129 | 0.01%
size/reversi | 175956 |   | size/reversi | 176421 | 0.26% |   | size/reversi | 176084 | 0.07%
gas/sha224 | 460197621 |   | gas/sha224 | 498978947 | 8.43% |   |   |   |  
size/sha224 | 191929 |   | size/sha224 | 192859 | 0.48% |   |   |   |  
gas/sha256 | 14487063673 |   | gas/sha256 | 15568532694 | 7.47% |   | gas/sha256 | 14486916565 | 0.00%
size/sha256 | 179075 |   | size/sha256 | 180167 | 0.61% |   | size/sha256 | 179223 | 0.08%

## test/bench

Master |   |   | Widening |   | Widening vs Master |   | Gated |   | Gated vs Master
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
gas/alloc | 9,243,068,120.00 |   | gas/alloc | 10,350,366,461.00 | 11.98% |   | gas/alloc | 9243068126 | 0.00%
size/alloc | 181,066.00 |   | size/alloc | 180,759.00 | -0.17% |   | size/alloc | 180464 | -0.33%
gas/bignum | 130,604,743.00 |   | gas/bignum | 130,606,013.00 | 0.00% |   | gas/bignum | 130604779 | 0.00%
size/bignum | 184,420.00 |   | size/bignum | 184,093.00 | -0.18% |   | size/bignum | 183790 | -0.34%
gas/heap-32 | 1,610,218,447.00 |   | gas/heap-32 | 1,695,702,521.00 | 5.31% |   | gas/heap-32 | 1609469958 | -0.05%
size/heap-32 | 182,167.00 |   | size/heap-32 | 181,856.00 | -0.17% |   | size/heap-32 | 181556 | -0.34%
gas/nat16 | 61,393,031.00 |   | gas/nat16 | 65,587,813.00 | 6.83% |   | gas/nat16 | 61393019 | 0.00%
size/nat16 | 181,010.00 |   | size/nat16 | 180,727.00 | -0.16% |   | size/nat16 | 180408 | -0.33%
gas/palindrome | 10,131,340.00 |   | gas/palindrome | 10,133,866.00 | 0.02% |   | gas/palindrome | 10131268 | 0.00%
size/palindrome | 185,338.00 |   | size/palindrome | 185,024.00 | -0.17% |   | size/palindrome | 184695 | -0.35%
gas/region0-mem | 6,402,149,937.00 |   | gas/region0-mem | 6,452,495,054.00 | 0.79% |   | gas/region0-mem | 6402149955 | 0.00%
size/region0-mem | 181,898.00 |   | size/region0-mem | 181,602.00 | -0.16% |   | size/region0-mem | 181281 | -0.34%
gas/region-mem | 5,974,331,587.00 |   | gas/region-mem | 6,024,676,752.00 | 0.84% |   | gas/region-mem | 5974331605 | 0.00%
size/region-mem | 181,539.00 |   | size/region-mem | 181,252.00 | -0.16% |   | size/region-mem | 180931 | -0.33%
gas/stable-mem | 3,885,566,188.00 |   | gas/stable-mem | 3,935,898,195.00 | 1.30% |   | gas/stable-mem | 3885566206 | 0.00%
size/stable-mem | 181,896.00 |   | size/stable-mem | 181,600.00 | -0.16% |   | size/stable-mem | 181279 | -0.34%
gas/xxx-nat32 | 57,198,791.00 |   | gas/xxx-nat32 | 57,199,237.00 | 0.00% |   | gas/xxx-nat32 | 57198779 | 0.00%
size/xxx-nat32 | 181,001.00 |   | size/xxx-nat32 | 180,694.00 | -0.17% |   | size/xxx-nat32 | 180399 | -0.33%


src/codegen/compile.ml
```diff
@@ -7091,35 +7483,38 @@ module MakeSerialization (Strm : Stream) = struct
           ReadBuf.read_float64 env get_data_buf ^^
           Float.box env
         end
-      | Prim (Int64|Nat64) ->
+      | Prim ((Int64|Nat64) as pty)->
```
ggreif, line 7486:
```suggestion
      | Prim ((Int64|Nat64) as pty) ->
```

luc-blaeser:
I wonder whether we should also distinguish between `Nat` and `Int` (while these types currently share the common tag `0b...10`)?

luc-blaeser:
Some performance results by the GC benchmark: https://github.com/luc-blaeser/gcbench

In this measurements, the instruction cost overheads also seem really moderate. The memory size difference is negligible.

## Instructions
Total instructions consumed, using copying GC.

Scenario	       | Original	| Tagged	| Difference
-------------------|------------|-----------|------------
asset-storage	   | 2.04E+08	| 2.04E+08	| 0
blobs	           | 1.52E+10	| 1.52E+10	| 0
btree-map	       | 5.35E+10	| 5.73E+10	| +7%
buffer	           | 2.53E+10	| 2.71E+10	| +7%
cancan	           | 4.43E+10	| 5.05E+10	| +14%
extendable-token   | 3.39E+06	| 3.48E+06	| +3%
game-of-life	   | 2.33E+07	| 2.33E+07	| 0
graph	           | 1.92E+10	| 2.02E+10	| +5%
imperative-rb-tree | 3.55E+10	| 3.62E+10	| +2%
linked-list	       | 3.75E+10	| 3.77E+10	| +1%
qr-code            | 1.16E+09	| 1.16E+09	| 0
random-maze	       | 2.72E+08	| 2.91E+08	| +7%
rb-tree	           | 1.76E+10	| 1.78E+10	| +1%
reversi	           | 2.59E+07	| 2.90E+07	| +12%
scalable-buffer	   | 6.75E+10	| 7.14E+10	| +6%
sha256	           | 3.58E+10	| 3.85E+10	| +8%
trie-map	       | 3.29E+10	| 3.36E+10	| +2%
**Average**	       | **2.27E+10** | **2.40E+10** | **+6%**

## Memory
Allocated memory size, using copying GC.

Scenario	       | Original	| Tagged	| Difference
-------------------|------------|-----------|------------
asset-storage	   | 5 MB	    | 5 MB	    | 0
blobs	           | 2128 MB	| 2128 MB	| 0
btree-map	       | 38 MB	    | 38 MB	    | 0
buffer	           | 57 MB	    | 57 MB	    | 0
cancan	           | 1246 MB	| 1246 MB	| 0
extendable-token   | 4 MB	    | 4 MB	    | 0
game-of-life	   | 75 MB	    | 81 MB	    | +8%
graph	           | 150 MB     | 150 MB	| 0
imperative-rb-tree | 65 MB      | 65 MB	    | 0
linked-list	       | 262 MB	    | 262 MB	| 0
qr-code	           | 53 MB	    | 53 MB	    | 0
random-maze	       | 5 MB	    | 5 MB	    | 0
rb-tree	           | 37 MB	    | 37 MB	    | 0
reversi	           | 3 MB	    | 3 MB	    | 0
scalable-buffer	   | 51 MB      | 51 MB	    | 0
sha256	           | 131 MB	    | 132 MB	| +1%
trie-map	       | 195 MB	    | 195 MB	| 0
**Average**	       | **265 MB** | **265 MB** | **0**

rts/motoko-rts/src/bigint.rs
```diff
@@ -129,7 +129,7 @@ extern "C" {
     pub(crate) fn bigint_trap() -> !;
 }
 
-#[inline]
+#[inline(always)]
```
luc-blaeser, line 132: Probably, `#[inline]` is sufficient (or even not specifying the attribute at all). I measured no real difference, even a very slight degrade...

src/codegen/compile.ml
```diff
@@ -29,6 +29,100 @@ let page_size = Int32.of_int (64*1024)
...
+    match u32 with
+    | ((O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O)) -> TBool (* false *)
+    | ((O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,I)) -> TBool (* true *)
+    | ((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,I,I)) -> TRef  (* 30 bit *)
+    | ((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,I,O)) -> TNum   (* 30 bit *)
+    | ((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,O,I,O,O)) -> TNat64 (* 28 bit *)
+    | ((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,I,I,O,O)) -> TInt64
+    | ((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,O,I,O,O,O)) -> TNat32 (* 27 bit *)
+    | ((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,I,I,O,O,O)) -> TInt32
+    | ((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,O,I,O), (O,O,O,O,O,O,O,O)) -> TChar
```
luc-blaeser, line 87: Nice! 
Maybe worth a brief comment: The 21 bits payload of `TChar` is sufficient to cover all UTF8 code points...

src/codegen/compile.ml
```diff
@@ -1534,7 +1628,7 @@ module BitTagged = struct
      tell it apart from a pointer by looking at the last bits: if set, it is a
      pointer.
 
-     Small here means -2^30 ≤ x < 2^30, and untagging needs to happen with an
+     Small here means -2^sbits ≤ x < 2^sbits, and untagging needs to happen with an
```
luc-blaeser, line 1631: For unsigned fix-sized numbers, the range of "small" values would be `-2^ubits <= x < 2^ubits`?
I believe, the comment below maybe needs be a adjusted a little bit, e.g. "Note that {Nat,Int}{8,16} do not need to be explicitly bit-tagged" would no longer be valid.

src/codegen/compile.ml
```diff
@@ -1574,58 +1668,223 @@ module BitTagged = struct
...
+  let tag_const pty i = Type.(
+    match pty with
+    |  Nat | Int | Int64 | Int32 ->
+      Int32.shift_left (Int64.to_int32 i) (32 - ubits_of pty)
+      (* tag *)
+      |> Int32.logor (TaggingScheme.tag_of_typ pty)
+    |  Nat64 | Nat32 ->
+      Int32.shift_left (Int64.to_int32 i) (32 - ubits_of pty) (* TBR *)
+      (* tag *)
+      |> Int32.logor (TaggingScheme.tag_of_typ pty)
```
luc-blaeser, line 1700: Maybe a brief not could be helpful (if I understood correctly):
For signed small (=non-boxed) numbers, the bit at position `ubits_of pty - 1` serves as the sign bit. 
Therefore, the tagging logic is the same for signed and unsigned small numbers.

```suggestion
    |  Nat | Int | Int64 | Int32 | Nat64 | Nat32 ->
      Int32.shift_left (Int64.to_int32 i) (32 - ubits_of pty)
      (* tag *)
      |> Int32.logor (TaggingScheme.tag_of_typ pty)
```

src/codegen/compile.ml
```diff
@@ -1574,58 +1668,223 @@ module BitTagged = struct
...
+      (Func.share_code1 Func.Always env name ("v", I32Type) [I32Type] (fun env get_n ->
+         get_n ^^
+         compile_bitand_const tag_mask ^^
+         compile_eq_const (TaggingScheme.tag_of_typ ty) ^^
+         E.else_trap_with env "unexpected tag" ^^
+         get_n))
+    else G.nop
 
-  let untag env =
-    compile_shrS_const 1l ^^
```
luc-blaeser, line 1609: Just for my own understanding: We spent one extra bit in the old approach for `Nat32` and `Nat64` (in addition to the tagging), leaving bit 30 zero, such that the common signed untagging logic could be applied?

src/codegen/compile.ml
```diff
@@ -1574,58 +1668,223 @@ module BitTagged = struct
...
+  let if_can_tag_i32 env pty retty is1 is2 = Type.(match pty with
+    |  Nat | Int | Int64 | Int32 ->
+      Func.share_code1 Func.Never env
+        (prim_fun_name pty "if_can_tag_i32") ("x", I32Type) [I32Type] (fun env get_x ->
+          (* checks that all but the low sbits are both either 0 or 1 *)
+          get_x ^^
+          get_x ^^ compile_shrS_const (Int32.of_int ((32 - ubits_of pty) - 1)) ^^
+          G.i (Binary (Wasm.Values.I32 I32Op.Xor)) ^^
+          compile_shrU_const (Int32.of_int (sbits_of pty)) ^^
+          G.i (Test (Wasm.Values.I32 I32Op.Eqz)) ^^
```
luc-blaeser, line 1843: Just a detail: As most of the logic is symmetric to the 64-bit tagging except for this case, I wonder whether we could apply this logic also to the 64-bit signed tagging test.

src/codegen/compile.ml
```diff
@@ -2356,42 +2621,58 @@ module BoxedSmallWord = struct
...
+      (prim_fun_name pty "box") ("n", I32Type) [I32Type] (fun env get_n ->
+      get_n ^^ compile_shrU_const (Int32.of_int (BitTagged.sbits_of pty)) ^^
       G.i (Unary (Wasm.Values.I32 I32Op.Popcnt)) ^^
       G.i (Unary (Wasm.Values.I32 I32Op.Ctz)) ^^
       G.if1 I32Type
-        (get_n ^^ BitTagged.tag_i32)
-        (compile_box env get_n)
+        (get_n ^^ BitTagged.tag_i32 env pty)
+        (compile_box env pty get_n)
+    )
```
luc-blaeser, line 2655: This can probably be deleted?

src/codegen/compile.ml
```diff
@@ -2470,15 +2769,16 @@ module TaggedSmallWord = struct
     in lsb_adjust ty ^^ set_b ^^ lsb_adjust ty ^^
        compile_unboxed_one ^^ get_b ^^ clamp_shift_amount ty ^^
        G.i (Binary (Wasm.Values.I32 I32Op.Shl)) ^^
-       G.i (Binary (Wasm.Values.I32 I32Op.And))
+       G.i (Binary (Wasm.Values.I32 I32Op.And)) ^^
+       msb_adjust ty
```
luc-blaeser, line 2773: I am not sure whether this `msb_adjust ty` is really needed because the result should be a boolean (not `ty`).

src/codegen/compile.ml
```diff
@@ -2470,15 +2769,16 @@ module TaggedSmallWord = struct
...
 
-  (* Code points occupy 21 bits, so can always be tagged scalars *)
-  let untag_codepoint = compile_shrU_const 8l
-  let tag_codepoint = compile_shl_const 8l
+  (* Code points occupy 21 bits, so can always be tagged scalars *)  (* TODO: rename (not tagging) *)
```
luc-blaeser, line 2775: Is the TODO already resolved by the renaming of the character adjustment functions?

src/codegen/compile.ml
```diff
@@ -3057,8 +3392,9 @@ module MakeCompact (Num : BigNumType) : BigNumType = struct
           get_amount ^^ compile_rel_const I32Op.LeU 42l ^^
           G.i (Binary (Wasm.Values.I32 I32Op.And)) ^^
           G.if1 I32Type
-            get_res
-            (get_n ^^ compile_shrS_const 1l ^^ Num.from_word30 env ^^ get_amount ^^ Num.compile_lsh env)
+            (get_res ^^ compile_bitor_const (TaggingScheme.tag_of_typ Type.Int))
+            (get_n ^^ compile_shrS_const (Int32.of_int (32 - BitTagged.ubits_of Type.Int)) ^^
+             Num.from_word30 env ^^ get_amount ^^ Num.compile_lsh env) (* TBR: from_word30? *)
```
luc-blaeser, line 3397: I am not sure whether I understand `from_word30`:
* In the original version, it maybe represented the 30 bits payload of the unboxed signed integer (32 minus 1 bit for the old tag and minus the sign bit, even if it was Nat).
* In the new version of precise tagging, it would probably be 29 bits (as the tag now occupies two bits), as there is no distinguishing between signed and unsigned big numbers? 
Maybe we could rename `from_word30` to e.g. `tag_small_bignum`?

src/codegen/compile.ml
```diff
@@ -3170,51 +3513,84 @@ module MakeCompact (Num : BigNumType) : BigNumType = struct
     try_unbox I32Type (fun _ -> match n with
         | 32 | 64 -> G.i Drop ^^ Bool.lit true
         | 8 | 16 ->
-          compile_bitand_const Int32.(logor 1l (shift_left minus_one (n + 1))) ^^
+          (* Please review carefully! *)
+          compile_bitand_const Int32.(logor 1l (shift_left minus_one (n + (32 - BitTagged.ubits_of Type.Int)))) ^^
```
luc-blaeser, line 3517: Regarding your comment, I believe this is correct:
- The shift serves for testing that the payload **including** the tag fits the desired bit width. This is now `n + 2` for `Type.Int`.
- The `or` with `1l` is to test the absence of the pointer skew bit, meaning that the is an unboxed big number.

But it is certainly good, if @ggreif double checks this.

src/codegen/compile.ml
```diff
@@ -3170,51 +3513,84 @@ module MakeCompact (Num : BigNumType) : BigNumType = struct
...
+
   let fits_signed_bits env n =
     let set_a, get_a = new_local env "a" in
     try_unbox I32Type (fun _ -> match n with
         | 32 | 64 -> G.i Drop ^^ Bool.lit true
         | 8 | 16 ->
+          (* Please review carefully! *)
            set_a ^^
-           get_a ^^ get_a ^^ compile_shrS_const 1l ^^
+           get_a ^^ get_a ^^ compile_shrS_const 1l ^^ (*Review 1l*)
```
luc-blaeser, line 3554: I believe that the `1l` is also correct. The `xor` with shift by 1 serves to check whether all bits beyond the signed payload are either all 1 or all 0. Maybe we could add a comment.

src/codegen/compile.ml
```diff
@@ -3338,60 +3728,61 @@ module MakeCompact (Num : BigNumType) : BigNumType = struct
...
     set_a ^^
-    get_a ^^ BitTagged.if_can_tag_i64 env [I32Type]
-      (get_a ^^ BitTagged.tag)
+    get_a ^^ BitTagged.if_can_tag_i64 env Type.Int [I32Type]
+      (get_a ^^ BitTagged.tag env Type.Int) (*TBR*)
       (get_a ^^ Num.from_signed_word64 env)
 
-  let from_word30 env =
-    compile_shl_const 1l
+  let from_word30 env = (* TBR: don't think we've got 30 bits! *)
```
luc-blaeser, line 3742: I think it would be 29 bits with the new big number tagging. Maybe we could rename this function (cf. comment above).

src/codegen/compile.ml
```diff
@@ -3338,60 +3728,61 @@ module MakeCompact (Num : BigNumType) : BigNumType = struct
...
+    (*    compile_shl_const (Int32.sub 32l BitTagged.ubitsl) ^^ *) (* TBR *)
```
luc-blaeser, line 3743: I believe this shift back to give room for the tag is already part of `BitTagged.tag_i32`.
```suggestion
```

src/codegen/compile.ml
```diff
@@ -4370,7 +4762,7 @@ module Tuple = struct
 
   (* We represent the boxed empty tuple as the unboxed scalar 0, i.e. simply as
      number (but really anything is fine, we never look at this) *)
-  let unit_vanilla_lit = 0l
+  let unit_vanilla_lit = TaggingScheme.unit_tag (* all tag, trivial payload *)
```
luc-blaeser, line 4765: Just a note for myself: Resume review from here... :-)

src/codegen/compile.ml
```diff
@@ -8296,19 +8691,32 @@ module StackRep = struct
...
+    (* TaggedSmallWord types *)
+    | UnboxedWord32 (Type.(Int8 | Nat8 | Int16 | Nat16 | Char) as pty), Vanilla ->
+       TaggedSmallWord.tag env pty
+    | Vanilla, UnboxedWord32 (Type.(Nat8 | Int8 | Nat16 | Int16 | Char) as pty) ->
+       TaggedSmallWord.untag env pty
 
-    | UnboxedWord32, Vanilla -> BoxedSmallWord.box env
-    | Vanilla, UnboxedWord32 -> BoxedSmallWord.unbox env
+    (* BoxedSmallWord types *)
+    (* TODO: constrain pty *)
```
luc-blaeser, line 8704: This would be `(Type.(Int32 | Nat32) as pty)`?

src/codegen/compile.ml
```diff
@@ -10224,21 +10633,23 @@ and compile_prim_invocation (env : E.t) ae p es at =
   | NextArrayOffset spacing, [e] ->
     let advance_by =
       match spacing with
-      | ElementSize -> Arr.element_size
-      | One -> 2l (* 1 : Nat *) in
+      | ElementSize
```
luc-blaeser, line 10636: Maybe we should rename `ElementSize` or discard it completely because (if I understand it correctly), the iterator would now advance by an index of 1 and recompute the word offset (by multiplying the word size) before dereferencing the array element.
(Perhaps, briefly repeating the comment that this index value is untagged (but shifted) could be helpful. Or in other words, it uses the specific tag `0b00` as an optimization.)

luc-blaeser: (Maybe a short comment could be added: The `advance_by` value is not tagged but only shifted as an optimzation because it is directly added to the already tagged previous iterator index.)

src/codegen/compile.ml
```diff
@@ -10224,21 +10633,23 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
-    compile_exp_vanilla env ae e1 ^^
-    compile_exp_vanilla env ae e2 ^^
+    compile_exp_vanilla env ae e1 ^^ BitTagged.untag_i32 __LINE__ env Type.Int ^^
+    compile_exp_vanilla env ae e2 ^^ BitTagged.untag_i32 __LINE__ env Type.Int ^^
     G.i (Compare (Wasm.Values.I32 I32Op.LtU))
   | DerefArrayOffset, [e1; e2] ->
     SR.Vanilla,
     compile_exp_vanilla env ae e1 ^^ (* skewed pointer to array *)
     Tagged.load_forwarding_pointer env ^^
     compile_exp_vanilla env ae e2 ^^ (* byte offset *)
```
luc-blaeser, line 10650: If I am not mistaken, the value would be the tagged index in the array:
```suggestion
    compile_exp_vanilla env ae e2 ^^ (* element index *)
```

src/codegen/compile.ml
```diff
@@ -10249,8 +10660,8 @@ and compile_prim_invocation (env : E.t) ae p es at =
   | GetPastArrayOffset spacing, [e] ->
     let shift =
       match spacing with
-      | ElementSize -> compile_shl_const 2l (* effectively a multiplication by word_size *)
-      | One -> BigNum.from_word30 env in    (* make it a compact bignum *)
+      | ElementSize
+      | One -> BigNum.from_word30 env in    (* make it a compact bignum *) (* TBR: from_word30? *)
```
luc-blaeser, line 10664: If I am not mistaken, it would be the compact integer value with 29 bit payload, excluding the sign bit and the tag...

src/codegen/compile.ml
```diff
@@ -10352,97 +10766,116 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
+      E.call_import env "rts" "bigint_of_float64" ^^
+      let set_b, get_b = new_local env "b" in
+      (* adjust 31 to ubit scalar, done here to avoid touching bigint.rs *)
+      set_b ^^
+      get_b ^^
+      BitTagged.if_tagged_scalar env [I32Type]
+        (get_b ^^
+         compile_shrS_const 1l ^^
+         BigNum.from_signed_word32 env)
+        (get_b)
```
luc-blaeser, line 10814: The analogous logic is used in `compile_load_from_data_buf` (covering both signed and unsigned cases). Maybe, this logic could be extracted as a common "bigint result sanitizer" helper function.

src/codegen/compile.ml
```diff
@@ -10352,97 +10766,116 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
+      compile_exp_as env ae (SR.UnboxedWord32 Nat8) e ^^
+      TaggedSmallWord.lsb_adjust Nat8 ^^
+      TaggedSmallWord.msb_adjust Nat16
     | Nat16, Nat32 ->
-      SR.Vanilla,
-      compile_exp_vanilla env ae e ^^
-      compile_shrU_const 15l (* resulting Nat32 will always be unboxed *)
+      SR.UnboxedWord32 Nat32,
+      compile_exp_as env ae (SR.UnboxedWord32 Nat16) e ^^
+      TaggedSmallWord.lsb_adjust Type.Nat16
```
luc-blaeser, line 10847: 👍

src/codegen/compile.ml
```diff
@@ -10451,36 +10884,36 @@ and compile_prim_invocation (env : E.t) ae p es at =
       get_val ^^
       G.i (Convert (Wasm.Values.I32 I32Op.WrapI64))
     | Int8, Int16 ->
-      SR.Vanilla,
-      compile_exp_vanilla env ae e ^^
+      SR.UnboxedWord32 Int16,
+      compile_exp_as env ae (SR.UnboxedWord32 Int8) e ^^
       compile_shrS_const 8l
```
luc-blaeser, line 10889: I think this is the short version of the following logic. The 8 bit shift only works because the tags of `Int8` and `Int16` are identical except that there is a shift by 8 bits. Probably, a more explicit logic would be safer to maintain:
```suggestion
      TaggedSmallWord.lsb_adjust Int8 ^^
      TaggedSmallWord.msb_adjust Int16
```

src/codegen/compile.ml
```diff
@@ -10451,36 +10884,36 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
     | Int16, Int32 ->
-      SR.Vanilla,
-      compile_exp_vanilla env ae e ^^
-      compile_shrS_const 15l (* resulting Int32 will always be unboxed *)
+      SR.UnboxedWord32 Int32,
+      compile_exp_as env ae (SR.UnboxedWord32 Int16) e ^^
+      compile_shrS_const 16l (* lsb_adjust? *)
```
luc-blaeser, line 10893:
```suggestion
      TaggedSmallWord.lsb_adjust Type.Int16
```

or even
```
      TaggedSmallWord.lsb_adjust Type.Int16
      TaggedSmallWord.msb_adjust Type.Int32
```

src/codegen/compile.ml
```diff
@@ -10451,36 +10884,36 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
       let set_val, get_val = new_local env "convertee" in
-      compile_exp_vanilla env ae e ^^
+      compile_exp_as env ae (SR.UnboxedWord32 Int16)  e ^^
       set_val ^^
       get_val ^^
       compile_shl_const (Int32.of_int num_bits) ^^
       compile_shrS_const (Int32.of_int num_bits) ^^
       get_val ^^
-      compile_eq env Type.(Prim Nat16) ^^
+      compile_eq env Type.(Prim Nat16) ^^ (* TBR *)
```
luc-blaeser, line 10908: Although the implementation results in the same code, I am not sure whether this should rather be: 
```suggestion
      compile_eq env Type.(Prim Int16) ^^ 
```

src/codegen/compile.ml
```diff
@@ -10579,7 +11012,7 @@ and compile_prim_invocation (env : E.t) ae p es at =
     SR.Vanilla,
     compile_exp_vanilla env ae e ^^
     Arr.len env ^^
-    BigNum.from_word30 env
+    BigNum.from_word32 env
```
luc-blaeser, line 11015: Good catch: If we would have relaxed the 2 billion element (2 * 1024 * 1024 * 1024) array length limit in `alloc_array` in the RTS, this would have been broken...

src/codegen/compile.ml
```diff
@@ -10941,31 +11382,99 @@ and compile_prim_invocation (env : E.t) ae p es at =
 
 
   | OtherPrim "popcnt8", [e] ->
```
luc-blaeser, line 11384: Maybe, we should internally rename this to `popcntNat8` in symmetry with the new specialized `popcntInt8` etc.

src/codegen/compile.ml
```diff
@@ -10941,31 +11382,99 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
-    SR.Vanilla,
-    compile_exp_vanilla env ae e ^^
+    SR.UnboxedWord32 Type.Nat16,
+    compile_exp_as env ae (SR.UnboxedWord32 Type.Nat16) e ^^
     G.i (Unary (Wasm.Values.I32 I32Op.Popcnt)) ^^
     TaggedSmallWord.msb_adjust Type.Nat16
+  | OtherPrim "popcntInt8", [e] ->
+    SR.UnboxedWord32 Type.Int8,
+    compile_exp_as env ae (SR.UnboxedWord32 Type.Int8) e ^^
+    compile_shrU_const (TaggedSmallWord.shift_of_type Type.Int8) ^^
```
luc-blaeser, line 11397: I am not sure whether this shift is needed (kind of a `lsb_adjust Nat8` skipping the sign extension). It could also be used for `Nat8`. Maybe we could add a comment stating that the bit counting is independent of whether the payload resides in LSB or MSB. The only important part is not to extend the sign bit...

src/codegen/compile.ml
```diff
@@ -10941,31 +11382,99 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
-  | OtherPrim "clz64", [e] -> SR.UnboxedWord64, compile_exp_as env ae SR.UnboxedWord64 e ^^ G.i (Unary (Wasm.Values.I64 I64Op.Clz))
-  | OtherPrim "ctz8", [e] -> SR.Vanilla, compile_exp_vanilla env ae e ^^ TaggedSmallWord.ctz_kernel Type.Nat8
-  | OtherPrim "ctz16", [e] -> SR.Vanilla, compile_exp_vanilla env ae e ^^ TaggedSmallWord.ctz_kernel Type.Nat16
-  | OtherPrim "ctz32", [e] -> SR.UnboxedWord32, compile_exp_as env ae SR.UnboxedWord32 e ^^ G.i (Unary (Wasm.Values.I32 I32Op.Ctz))
-  | OtherPrim "ctz64", [e] -> SR.UnboxedWord64, compile_exp_as env ae SR.UnboxedWord64 e ^^ G.i (Unary (Wasm.Values.I64 I64Op.Ctz))
+  | OtherPrim "popcntInt64", [e] ->
+    SR.UnboxedWord64 Type.Int64,
+    compile_exp_as env ae (SR.UnboxedWord64 Type.Int64) e ^^
+    G.i (Unary (Wasm.Values.I64 I64Op.Popcnt))
+  | OtherPrim "clz8", [e] ->
```
luc-blaeser, line 11422: Maybe also here we could consider internally renaming it to `clzNat8`. Same for `ctzNat8`.

src/codegen/compile.ml
```diff
@@ -11116,18 +11633,29 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
+     const_sr (SR.UnboxedWord32 Type.Nat32) (TaggedSmallWord.btst_kernel env Type.Nat32)
+  | OtherPrim "btstInt32", [_;_] ->
+     const_sr (SR.UnboxedWord32 Type.Int32) (TaggedSmallWord.btst_kernel env Type.Int32) (* ! *)
   | OtherPrim "btst64", [_;_] ->
-    const_sr SR.UnboxedWord64 (
+    const_sr (SR.UnboxedWord64 Type.Nat64) (
       let (set_b, get_b) = new_local64 env "b" in
       set_b ^^ compile_const_64 1L ^^ get_b ^^ G.i (Binary (Wasm.Values.I64 I64Op.Shl)) ^^
-      G.i (Binary (Wasm.Values.I64 I64Op.And))
+      G.i (Binary (Wasm.Values.I64 I64Op.And)) (* TBR *)
```
luc-blaeser, line 11651: The logic looks correct to me. Maybe we could extract a common bit test function for `Word64`.

test/bench/alloc.mo
```diff
@@ -16,7 +16,7 @@ actor alloc {
         };
         let (m1, n1) = counters();
         debugPrint(debug_show (m1 - m0, n1 - n0));
-
+        assert false; // rollback changes
```
luc-blaeser, line 19: Is `drun` suffering from a timeout? Maybe we could also just downsize the test. I believe not rolling back could maybe also be interesting to assess the GC time (with a growing heap).

test/run/idl-ops.mo
```diff
@@ -51,7 +51,8 @@ assert(null == deserArrayNat (serArrayInt arrayInt));
 assert((?arrayInt) == deserArrayInt (serArrayInt arrayInt));
 let heapDifference = Prim.rts_heap_size() : Int - started_with;
 // Difference between incremental and non-incremental GC
-assert(heapDifference <= +8_046);
+// TODO: adjust and restore
+//assert(heapDifference <= +8_046);
```
luc-blaeser, line 55:
```suggestion
assert(heapDifference <= +20_176);
```

test/run/idl.mo
```diff
@@ -47,7 +47,8 @@ assert(arrayInt == deserArrayInt (serArrayInt arrayInt));
 
 let heapDifference = Prim.rts_heap_size() : Int - started_with;
 // Difference between incremental and non-incremental GC
-assert(heapDifference == +4_488 or heapDifference == +4_888);
+// TODO: adjust and restore
+//assert(heapDifference == +4_488 or heapDifference == +4_888);
```
luc-blaeser, line 51:
```suggestion
assert(heapDifference == +13_532 or heapDifference == +14_940);
```

test/run/if-swap.mo
```diff
@@ -3,10 +3,10 @@ func wantSeeSwap(x : Bool) : Nat =
 
 ignore(wantSeeSwap(true));
 
-// CHECK-LABEL: (func $wantSeeSwap
-// CHECK-NEXT: local.get $x
-// CHECK-NEXT: if (result i32)
-// CHECK-NEXT: i32.const 50
-// CHECK-NEXT: else
-// CHECK-NEXT: i32.const 84
+// FHECK-LABEL: (func $wantSeeSwap
```
luc-blaeser, line 6: Ah, this is only a TODO to later adjust the code comparison...

luc-blaeser (approved):
Thank you very much for this very nice PR which does not only improve the self-descriptiveness of the heap values, but even the self-descriptiveness in the compiler implementation!

There are only a few comments I have, some local improvement suggestions, besides the mentioned TODOs above.

src/codegen/compile.ml
```diff
@@ -8296,19 +8691,32 @@ module StackRep = struct
...
     | Vanilla, UnboxedFloat64 -> Float.unbox env
 
     | Const (_, Const.Lit (Const.Bool b)), Vanilla -> Bool.lit b
     | Const c, Vanilla -> compile_unboxed_const (materialize_const_t env c)
-    | Const (_, Const.Lit (Const.Word32 n)), UnboxedWord32 -> compile_unboxed_const n
-    | Const (_, Const.Lit (Const.Word64 n)), UnboxedWord64 -> compile_const_64 n
+    | Const (_, Const.Lit (Const.Vanilla n)), UnboxedWord32 ty ->
+       compile_unboxed_const n ^^
+       TaggedSmallWord.untag env ty
+    | Const (_, Const.Lit (Const.Word32 (_ty1, n))), UnboxedWord32 _ty2 ->
```
luc-blaeser, line 8716: I assume `ty1` and `ty2` should be identical?

```suggestion
    | Const (_, Const.Lit (Const.Word32 (ty1, n))), UnboxedWord32 ty2 when ty1 = ty2 ->
```

src/codegen/compile.ml
```diff
@@ -29,6 +29,100 @@ let page_size = Int32.of_int (64*1024)
...
+  (* Leverage OCaml pattern match compilation to check tagging scheme is injective *)
+  let _decode u32 =
+    match u32 with
+    | ((O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O)) -> TBool (* false *)
+    | ((O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,I)) -> TBool (* true *)
+    | ((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,I,I)) -> TRef  (* 30 bit *)
+    | ((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,I,O)) -> TNum   (* 30 bit *)
+    | ((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,O,I,O,O)) -> TNat64 (* 28 bit *)
+    | ((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,I,I,O,O)) -> TInt64
+    | ((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,O,I,O,O,O)) -> TNat32 (* 27 bit *)
```
luc-blaeser, line 85: Just a thought: If we switch to 64-bit field values, the question is whether we keep these tags (with potentially longer payloads) or use an adjusted tagging scheme. For example, a 32-bit tag could be used `Nat32`, `Int32` in 64-bit words. 

The same question would apply to the stable format of graph copy. If we use a 64-bit format right from the beginning, we could also use this extended tagging scheme there. The serialization from 32-bit main memory would then need to retag scalar values.

test/run/idl.mo
crusso: Thanks!

rts/motoko-rts/src/bigint.rs
crusso, line 132:
```suggestion
#[inline]
```

test/bench/alloc.mo
crusso: Yes, I think I added the rollback when I only had 24 bits compact bignums, leading to loads of boxing. Probably safe to revert now we have 30 bit compact bignums.

test/bench/alloc.mo
```diff
@@ -1,6 +1,6 @@
 //MOC-FLAG --force-gc
 import { performanceCounter; rts_heap_size; debugPrint } = "mo:⛔";
-
+// this test becomes much more expensive due to smaller compact bignums.
```
crusso, line 3:
```suggestion

```

crusso, line 3:
```suggestion

```

test/bench/alloc.mo
crusso: Reverted

crusso: Reverted

src/codegen/compile.ml
crusso: Fixed in #4375

src/codegen/compile.ml
crusso: Fixed in #4375

src/codegen/compile.ml
crusso: Fixed in #4375

src/codegen/compile.ml
crusso: For unsinged, fixed size, its `0 <= x < 2^ubits', I think

I've adjust the comment (which shows how the complexity has grown a fair bit, sigh)

src/codegen/compile.ml
crusso: Done

src/codegen/compile.ml
crusso: That's right. But now that the stack rep distinguished between types, we don't have to waste that bit anymore and can support specialized tagging.

src/codegen/compile.ml
crusso: yep

src/codegen/compile.ml
crusso: prim.mo expects it to return a ty, which is then equated against 0:ty. I think there's an existing todo to fix this....
I think I'll leave it for now...

src/codegen/compile.ml
crusso: Yep

src/codegen/compile.ml
crusso: done

src/codegen/compile.ml
crusso: Done

src/codegen/compile.ml
crusso: Done in #4375

src/codegen/compile.ml
crusso: Deleted

src/codegen/compile.ml
crusso: Done

src/codegen/compile.ml
crusso: Done in #4375

src/codegen/compile.ml
crusso: Done, but there might be more such places...

src/codegen/compile.ml
crusso: Yep, added commnet

src/codegen/compile.ml
crusso: Yep, added comment.

src/codegen/compile.ml
crusso: Yep.

src/codegen/compile.ml
crusso: Yes, I didn't want to do that in case some uses the prim directly (though that's strictly not supported).
Even better would be if we allowed prims to have type parameters, so we could resolve overloading more cleanly, but that's maybe work for a separate PR.

src/codegen/compile.ml
crusso: Ditto for popcntInt8. Fixed both.

src/codegen/compile.ml
crusso: see previous comment about breaking existing uses of prim

src/codegen/compile.ml
crusso: That was the original code, but seems like a benign change. Done.

src/codegen/compile.ml
crusso: I was worried we  might need this to support conversions between equal fixed size signed and unsigned (eg Nat8/Int8) but am trying it out.

src/codegen/compile.ml
crusso: I guess I was expecting to stretch the relevant tags by shifting left 32 bits for small words, make 32 bit values small words and widening the payload for compact 64bit and BigInt. But haven't thought about it that hard. Keeping the tags as is but shifting the payload would also work, at the cost of using shorter tags than necessary I guess. Not sure which is preferable, actually.

src/codegen/compile.ml
crusso: Added an assert instead (also for BoxedWord64)

src/codegen/compile.ml
crusso: Created an issue to fix this https://github.com/dfinity/motoko/issues/4391

src/codegen/compile.ml
crusso: Great explanation, which make me wonder why we mask out the low bit since it must be zero in this case anyway. @ggreif any idea why?

rts/motoko-rts/src/bigint.rs
crusso: Reverted

src/codegen/compile.ml
crusso: I've removed the need for this logic entirely by getting the rts to call a new compile.ml exported function for constructing a motoko `Int` from an i32 (`int_from_i32`), boxing only when necessary.
That actually revealed a subtle bug in `compile_load_from_data_buf` which was mistakenly calling `Num.from_(signed)_word32`, not `BigNum.from_(signed)_word32`, causing it to heap allocate even small deserialized bignums, undoing @ggreif's hard work to optimize this.

The change fixes the regression in idl.mo.

src/codegen/compile.ml
crusso: Done.

src/codegen/compile.ml

src/codegen/compile.ml
```diff
@@ -29,6 +29,100 @@ let page_size = Int32.of_int (64*1024)
...
+  (*
+     Enable for development only to sanity check value tags and
+     locate unexpected tag errors to compile.ml source lines.
+     Flags.sanity_check will check tags, but not further locate them.
+  *)
+
+  let debug = false (* should never be true in master! *)
+
+  type bit = I
+           | O
```
crusso, line 61:
```suggestion
  type bit = I | O
```
trival edit to trigger CI and produce artifacts

luc-blaeser:
I remeasured the performance in GC benchmark with the latest tagging changes: https://github.com/luc-blaeser/gcbench

Generally low or moderate performance regression.

## Instructions
Total instructions consumed, using copying GC

Scenario	       | Original	| Tagged	| Difference
-------------------|------------|-----------|------------
asset-storage	   | 2.04E+08	| 2.04E+08	| 0
blobs	           | 1.52E+10	| 1.52E+10	| 0
btree-map	       | 5.35E+10	| 5.79E+10	| +8%
buffer	           | 2.53E+10	| 2.71E+10	| +7%
cancan	           | 4.43E+10	| 5.05E+10	| +14%
extendable-token   | 3.39E+06	| 3.41E+06	| +1%
game-of-life	   | 2.33E+07	| 2.33E+07	| 0
graph	           | 1.92E+10	| 2.03E+10	| +6%
imperative-rb-tree | 3.55E+10	| 3.77E+10	| +6%
linked-list	       | 3.75E+10	| 3.99E+10	| +6%
qr-code            | 1.16E+09	| 1.16E+09	| 0
random-maze	       | 2.72E+08	| 2.92E+08	| +7%
rb-tree	           | 1.76E+10	| 1.87E+10	| +6%
reversi	           | 2.59E+07	| 2.86E+07	| +10%
scalable-buffer	   | 6.75E+10	| 7.20E+10	| +7%
sha256	           | 3.58E+10	| 3.85E+10	| +8%
trie-map	       | 3.29E+10	| 3.54E+10	| +8%
**Average**	       | **2.27E+10** | **2.44E+10** | **+7%**

## Memory Size
No changes since the last measurement. Except for one case (`game-of-life`, +8%) no relevant Wasm allocated memory increase compared to `master`.

crusso:
Excellent - thanks for  that!

src/codegen/compile.ml
```diff
@@ -2237,33 +2525,39 @@ module BoxedWord64 = struct
...
 
-  let unbox env = Func.share_code1 Func.Never env "unbox_i64" ("n", I32Type) [I64Type] (fun env get_n ->
+  let unbox env pty =
+    Func.share_code1 Func.Never env
+      (prim_fun_name pty "unbox64") ("n", I32Type) [I64Type] (fun env get_n ->
       get_n ^^
       BitTagged.if_tagged_scalar env [I64Type]
-        ( get_n ^^ BitTagged.untag env)
+        ( get_n ^^ BitTagged.untag __LINE__ env pty)
         ( get_n ^^ Tagged.load_forwarding_pointer env ^^ Tagged.load_field64 env (payload_field env))
```
ggreif, line 2561:
```suggestion
        (get_n ^^ BitTagged.untag __LINE__ env pty)
        (get_n ^^ Tagged.load_forwarding_pointer env ^^ Tagged.load_field64 env (payload_field env))
```

src/codegen/compile.ml
ggreif: `if_tagged_scalar` only chooses this fast path when the bottom bit is 0, so the `logor 1l` is indeed redundant.

src/codegen/compile.ml
```diff
@@ -3170,51 +3534,84 @@ module MakeCompact (Num : BigNumType) : BigNumType = struct
     try_unbox I32Type (fun _ -> match n with
         | 32 | 64 -> G.i Drop ^^ Bool.lit true
         | 8 | 16 ->
-          compile_bitand_const Int32.(logor 1l (shift_left minus_one (n + 1))) ^^
+          (* Please review carefully! *)
```
ggreif, line 3537: can remove

src/codegen/compile.ml
```diff
@@ -10393,56 +10850,67 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
+      TaggedSmallWord.lsb_adjust Nat8 ^^
+      TaggedSmallWord.msb_adjust Nat16
     | Nat16, Nat32 ->
-      SR.Vanilla,
-      compile_exp_vanilla env ae e ^^
-      compile_shrU_const 15l (* resulting Nat32 will always be unboxed *)
+      SR.UnboxedWord32 Nat32,
+      compile_exp_as env ae (SR.UnboxedWord32 Nat16) e ^^
+      TaggedSmallWord.lsb_adjust Type.Nat16 ^^
+      TaggedSmallWord.msb_adjust Nat32 (* NB: a nop for 32-bit present, but not for 64-bit future*)
```
ggreif, line 10882:
```suggestion
      TaggedSmallWord.msb_adjust Nat32 (* NB: a nop for 32-bit present, but not for 64-bit future *)
```

ggreif (approved):
Holy shit! This is a gargantuan one. I didn't see anything suspicious, but also my eyesight is not as good as in my youth years :-)

Obviously mandatory unboxing/reboxing makes `arg1 ⨶ arg2`-style simple arithmetic very expensive. A pity that such simple ones are the rule and not the exception.

I see the point that we need tagging if we want graph-serialisation *and* 64-bit heap cells. But when will that come?

Anyway I suggest a somewhat extended beta phase to get some experience with this.

src/mo_values/prim.ml
```diff
@@ -146,7 +146,8 @@ let prim trap =
      | _ -> assert false)
   | "fexp" -> fun _ v k -> k (via_float Stdlib.exp v)
   | "flog" -> fun _ v k -> k (via_float Stdlib.log v)
-
+  (* TODO: refine exotic cases below to catch more errors *)
+  | "popcntInt8" | "popcntInt16" | "popcntInt32" | "popcntInt64"
   | "popcnt8" | "popcnt16" | "popcnt32" | "popcnt64" ->
```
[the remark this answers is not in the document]
ggreif: Oh, they are, for `Nat*`.

src/lowering/desugar.ml
```diff
@@ -269,33 +269,46 @@ and transform_for_to_while p arr_exp proj e1 e2 =
...
-     let size = arr.size(e1) ;
-     var indx = 0 ;
-     label l loop {
-       if indx < size
-       then { let p = arr[indx]; e2; indx += 1 }
-       else { break l }
-     } *)
+     let last = arr.size(e1) : Int - 1 ;
+     var indx = 0;
+     if (last == -1) { }
```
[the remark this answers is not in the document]
crusso: It's only done once. Does it matter? We can always improve later. I just didn't want to ever exceed the range of (now) 30 bit compact bignum.

src/mo_values/prim.ml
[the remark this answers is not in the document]
crusso: Yep. Better would be to just have designated prims with a type argument, or generalize all OtherPrims to carry type parameters when necessary - that would simplify alot of boilerplate code in various places. But cane come later...

rts/motoko-rts/src/bigint.rs
```diff
@@ -271,12 +277,9 @@ unsafe extern "C" fn bigint_of_float64(j: f64) -> Value {
 #[cfg(feature = "ic")]
 #[no_mangle]
 unsafe extern "C" fn bigint_to_float64(p: Value) -> f64 {
-    if p.is_scalar() {
-        p.get_signed_scalar() as f64
-    } else {
-        let mp_int = p.as_bigint().mp_int_ptr();
-        mp_get_double(mp_int)
-    }
+    debug_assert!(!p.is_scalar());
```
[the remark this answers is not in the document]
crusso: The RTS only knows about untagged 31 bit scalars, but our compact bignums are actually 30 bit. The idea here is to shift the switch on compact vs boxed to compile.ml, where the scalar tagging scheme is known. That also make it easier to change in the future without touching the rts. So this function should now only be called on proper heap allocated bigints.

src/codegen/compile.ml
```diff
@@ -1534,11 +1627,24 @@ module BitTagged = struct
      tell it apart from a pointer by looking at the last bits: if set, it is a
      pointer.
 
-     Small here means -2^30 ≤ x < 2^30, and untagging needs to happen with an
-     _arithmetic_ right shift. This is the right thing to do for signed
-     numbers, and because we want to apply a consistent logic for all types,
-     especially as there is only one wasm type, we use the same range for
-     signed numbers as well.
```
crusso, line 1541: Was that a typo?

"we use the same range for  signed numbers as well."|-> 
"we use the same range for *un*signed numbers as well."

That always confused me....

src/codegen/compile.ml
```diff
@@ -2434,13 +2754,13 @@ module TaggedSmallWord = struct
     | Type.(Int32|Nat32) -> G.nop
     | Type.(Nat8|Nat16) as ty -> compile_shrU_const (shift_of_type ty)
     | Type.(Int8|Int16) as ty -> compile_shrS_const (shift_of_type ty)
+    | Type.Char as ty -> compile_shrU_const (shift_of_type ty)
```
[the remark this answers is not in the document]
crusso: Probably not because always had dedicated untag_codepoint/tag_codepoint functions and would use those.

src/codegen/compile.ml
crusso, line 3537:
```suggestion
          (* The shift serves for testing that the payload including the tag fits the desired bit width. 
              E.g. this is now n + 2 for Type.Int. *)
```

src/codegen/compile.ml
```diff
@@ -3170,51 +3534,84 @@ module MakeCompact (Num : BigNumType) : BigNumType = struct
     try_unbox I32Type (fun _ -> match n with
         | 32 | 64 -> G.i Drop ^^ Bool.lit true
         | 8 | 16 ->
-          compile_bitand_const Int32.(logor 1l (shift_left minus_one (n + 1))) ^^
+          (* Please review carefully! *)
+          compile_bitand_const Int32.(logor 1l (shift_left minus_one (n + (32 - BitTagged.ubits_of Type.Int)))) ^^
```
crusso, line 3538:
```suggestion
          compile_bitand_const Int32.(shift_left minus_one (n + (32 - BitTagged.ubits_of Type.Int))) ^^
```
remove redundant logor 1L

src/codegen/compile.ml
crusso: I change the comment to explain the shift (and removed the logor 1l)

crusso:
> Holy shit! This is a gargantuan one. I didn't see anything suspicious, but also my eyesight is not as good as in my youth years :-)

It was even bigger before I used the typed stackrep to do the tagging and untagging in one place... There each operation had to strip and add the tags, which was many changes...

> Obviously mandatory unboxing/reboxing makes `arg1 ⨶ arg2`-style simple arithmetic very expensive. A pity that such simple ones are the rule and not the exception.

I guess we could measure that and see the diff. My guess it that it is  3 extra instructions per op but less for compound expressions.

> I see the point that we need tagging if we want graph-serialisation _and_ 64-bit heap cells. But when will that come?

We could disable it for now, by just changing the TaggingScheme module and using all-zero tags, but then we would need to keep candid serialization around for the final upgrade to a tagged world. I think @luc-blaeser  has already ripped that out in his successor PR adding graph serialization deserialization, leaving just candid deserialization for the migration. But I could be wrong. 

> Anyway I suggest a somewhat extended beta phase to get some experience with this.

Yes maybe we should make a release and ask people to try it without including in dfx for a while.

crusso:
@ggreif @luc-blaeser 

What should I do about the filecheck tests? I could modify them for the new tagging scheme (but then we can't disable it easily) or just leave as is.

They feel pretty fragile to me.

crusso:
@ggreif @luc-blaeser  I started  PR #4400 to toggle the scalar tagging via a flag in case we are nervous. Widening the compact integer ranges back to 31 bits (from 30/28/26 for `Nat/Int`, `Nat64/Int64`  `Nat32/Int32`)  revealed what I think was a bug in the `if_can_tag_i32/64`  tests and sanity checks. 

When `ubits pty` is 31, I think the broken tests would just clear the signs bits and thus always succeed and tag the numbers, even when out of range.
When `ubits pty` was less than 31, it think it would fail to xor the lowest sign-bit, and possibly box more than necessary (not entirely sure, tbh).


I've fixed the bug here https://github.com/dfinity/motoko/pull/4369/commits/537876bf3c8eca78ecc71265e10b1d89f5fddaeb
but would appreciate a careful second look from both of you.

PR #4370, opened 2024-01-24
fix: wrong new_pages calculation in StableMultiLog

ninegua:
Fix the wrong calculation on the new pages needed to grow stable memory in the StableMultiLog example used in our documentation.


crusso (approved):
Nice catch!

PR #4371, opened 2024-01-24
EXPERIMENT: research actor modularisation

ggreif:
Research how actors can be chopped into parts and reassembled in a mix-and-match fashion. Everything here is experimental, mostly to figure out which parts of the compiler are affected. When I have a better understanding I'll write up a concrete proposal.

Contains quite some drive-by simplifications and cleanups. These should be broken out separately — Done: #4393.


luc-blaeser:

PR #4375, opened 2024-01-26
experiment: simplify array iter optimization, don't cheat the type system

crusso:
Improves #4369.

Rewrite the array.keys/vals optimization to safely use compact bignums, but without breaking the abstraction of using compact Int/Nat as indices. The old implementation cheated and would allow index arithmetic to overflow the compact range.

Assumes all valid array indices are representable as compact nat/int.


src/codegen/compile.ml
```diff
@@ -3804,7 +3813,7 @@ module BigNumLibtommath : BigNumType = struct
   let truncate_to_word32 env = E.call_import env "rts" "bigint_to_word32_wrap"
   let truncate_to_word64 env = E.call_import env "rts" "bigint_to_word64_wrap"
 
-  let from_word30 env = E.call_import env "rts" "bigint_of_word32"
```
luc-blaeser, line 3807: Interesting that this ever worked. Probably it is/was never used for a signed number?

src/codegen/compile.ml
```diff
@@ -4536,6 +4545,8 @@ module Arr = struct
      No difference between mutable and immutable arrays.
   *)
 
+  let max_size env = Int32.shift_left 1l 29 (* inclusive *)
```
luc-blaeser, line 4548: This is because the array iterator did not and does not support arrays longer than around 500 million elements (1<<29)?
Just curious whether we check the case of longer arrays, since the RTS has a higher limit (around 2 billion elements, 1<<31).

luc-blaeser (approved):
Really nice!

src/codegen/compile.ml
crusso: Would this not have worked? I think it's ok, but probably unreachable anyway (since a word30 would always be compact).

src/codegen/compile.ml
crusso: Hmm, I think that should actually be 
```suggestion
  let max_size env = Int32.shift_left 1l 28 (* inclusive *)
```
if I want max_size to be 2^29 (silly me)

rts/motoko-rts/src/constants.rs
```diff
@@ -3,6 +3,10 @@ use crate::types::{Bytes, Words};
 /// Wasm word size. RTS only works correctly on platforms with this word size.
 pub const WORD_SIZE: u32 = 4;
 
+/// Maximum Motoko array size 2^29 (inclusive)
+/// NB: Must agree with Arr.max_array_size in compile.ml.
+pub const MAX_ARRAY_SIZE: u32 = 1 << 29;
```
ggreif, line 8: Technically the 29 is not a magic, but depends on `WORD_SIZE` (above). Not a biggie.

src/ir_def/ir.ml
```diff
@@ -146,10 +146,10 @@ and prim =
   | SelfRef of Type.typ               (* returns the self actor ref *)
   | SystemTimePrim
   (* Array field iteration/access *)
-  | NextArrayOffset of spacing        (* advance array offset *)
-  | ValidArrayOffset                  (* verify array offset *)
-  | DerefArrayOffset                  (* array offset indexing *)
-  | GetPastArrayOffset of spacing     (* array offset past the last element *)
+  | NextArrayOffset                   (* advance compact array offset,as Nat *)
```
ggreif, line 149:
```suggestion
  | NextArrayOffset                   (* advance compact array offset, as Nat *)
```

src/codegen/compile.ml
```diff
@@ -10630,21 +10642,23 @@ and compile_prim_invocation (env : E.t) ae p es at =
     SR.Vanilla,
     compile_array_index env ae e1 e2 ^^
     load_ptr
-  | NextArrayOffset spacing, [e] ->
```
ggreif, line 10633: I guess I had `spacing` as a parameter to allow `Nat8` (relevant for `Blob`?) and `Float` (but these are currently heap-indirect) increments. We can recall this feature if needed in the future.

src/codegen/compile.ml
```diff
@@ -10657,15 +10671,13 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
-      match spacing with
-      | ElementSize
-      | One -> BigNum.from_word30 env in    (* make it a compact bignum *) (* TBR: from_word30? *)
+  | GetLastArrayOffset, [e] ->
+    assert (BitTagged.can_tag_const Type.Int (Int64.of_int32 (Int32.sub (Arr.max_array_size env) 1l)));
     SR.Vanilla,
     compile_exp_vanilla env ae e ^^ (* array *)
     Arr.len env ^^
-    shift
+    compile_sub_const 1l ^^
```
ggreif, line 10679: Hmm, does this add an extra subtraction in every iteration?

crusso: No, it's only called once before loop entry (see desugar.ml)

crusso: No, it's only called once before loop entry (see desugar.ml)

src/codegen/compile.ml
crusso: I think spacing would increment by 4 or 1 depending on vals or keys iteration. But I can't store a byte offset with only 30 29 bits available, so just shift by 2 in both cases now.

rts/motoko-rts/src/constants.rs
crusso: I depends on the tagging scheme as well as wordsize. I now have 2 tag bits for plus sign bit, leading to 29.

crusso: We've lost 1-bit from compact bignums.

src/lowering/desugar.ml
```diff
@@ -269,33 +269,46 @@ and transform_for_to_while p arr_exp proj e1 e2 =
...
-       if indx < size
-       then { let p = arr[indx]; e2; indx += 1 }
-       else { break l }
-     } *)
+     let last = arr.size(e1) : Int - 1 ;
+     var indx = 0;
+     if (last == -1) { }
+     else {
+       label l loop {
+         let p = arr[indx]; // sans bound check
```
crusso, line 277:
```suggestion
         let p = arr[indx]; /* sans bound check */
```
trivial edit to trigger CI

ggreif (approved):
Okay! LFTM!

PR #4379, opened 2024-02-02
Upgrade the Wasm Profiler to the New IC Cost Model

luc-blaeser:
# Upgrade the Wasm Profiler to the New IC Cost Model

Upgrade the instrumentation of the `wasm-profiler` to use the new IC cost model.
* Also consider the IC call costs.
* Support both static and dynamic instruction costs.

This profiler is only used for the flamegraph generation (`tests/profile-report.sh`) and is **independent** of the CI performance difference report. 

## Limitation
* `parity_wasm` seems to be no longer maintained. It also lacks support of various new Wasm features (e.g. several instructions, tables, function references, Memory64 etc.). => We need to think about replacing it by `wasmparser` and `wasm-encoder` similar like in the current IC implementation.


chenyan-dfinity (commented):
I think we also need to update the system API cost: https://github.com/dfinity/ic-wasm/blob/main/src/utils.rs#L50

chenyan-dfinity:
There is a new repo https://github.com/dfinity/bench for using `drun` in `cargo bench`. I wonder if we can somehow adopt it for our use case.

luc-blaeser:
> I think we also need to update the system API cost: https://github.com/dfinity/ic-wasm/blob/main/src/utils.rs#L50

Thank you very much. The IC system API costs were indeed not even considered in the original profiler version. I now added the system API costs based on your implementation.

luc-blaeser:
> There is a new repo https://github.com/dfinity/bench for using `drun` in `cargo bench`. I wonder if we can somehow adopt it for our use case.

This is very interesting and could be a good solution for replacing our specific instrumentation, especially because the used library `partity_wasm` seems to be no longer maintained and lacks some Wasm feature support. 

We would probably need to at least adjust the scripts a little bit to generate the flamegraph and the performance delta report.

Maybe we should directly try to switch to that solution and not temporarily lift the self-built instrumentation?

chenyan-dfinity (approved):
LGTM. I think we can merge this first, and explore the use of `ic-bench` repo for another time (the repo is not public yet).

luc-blaeser:
> LGTM. I think we can merge this first, and explore the use of `ic-bench` repo for another time (the repo is not public yet).

Thank you very much, Yan. I also think this is a pragmatic approach.

PR #4386, opened 2024-02-06
remove `__get_candid_interface_tmp_hack` endpoint

chenyan-dfinity:
Now that candid are in the canister metadata for quite a while, `tmp_hack` is redundant. It also cannot be set to private like we do in metadata.


ggreif (approved):
Thanks! This looks great! Can you add to the `Changelog.md`?

chenyan-dfinity:
@ggreif Why does it cause gas change at all?

crusso:
> @ggreif Why does it cause gas change at all?

I think this is due to the static heap being different which can affect deserialization of bignums due to @Gabor's speculative decoding of compact bignums. @luc-blaeser look familiar to you?

I kicked off the mac build again since it failed due to drun non-determinism.

crusso (approved):
LGTM, but I wonder if any external folk are relying on it....

Thanks for cleaning this up!

crusso:
Hmm, I'm not sure why the perf changed, tbh. I think my original hunch was wrong, actually.

chenyan-dfinity:
The benchmark may contain deserialization, but I don't know why the perf changed.

rdobrik:
We are still using it in our Java Agent so we need to fix it and replace with metadata. Any example how to get idl from running canister using metadata?

chenyan-dfinity:
If you already have an agent library, you can fetch the state tree with path `/canister/<canister_id>/metadata/candid:service`. See the spec: https://internetcomputer.org/docs/current/references/ic-interface-spec/#state-tree-canister-information.

Concretely, here is the code in Rust: https://github.com/dfinity/agent-rs/blob/main/ic-agent/src/agent/mod.rs#L994 and code in agent-js: https://github.com/dfinity/agent-js/blob/main/packages/agent/src/canisterStatus/index.ts#L144

rdobrik:
> If you already have an agent library, you can fetch the state tree, e.g., `read_state`, with path `/canister/<canister_id>/metadata/candid:service`. See the spec: https://internetcomputer.org/docs/current/references/ic-interface-spec/#state-tree-canister-information.
> 
> Concretely, here is the code in Rust: https://github.com/dfinity/agent-rs/blob/main/ic-agent/src/agent/mod.rs#L994 and code in agent-js: https://github.com/dfinity/agent-js/blob/main/packages/agent/src/canisterStatus/index.ts#L144

Thank you, I will fix our Java code.

rdobrik:
Worked. Fixed our Java Agent. We will publish fixed library ASAP. Thanks for help.

PR #4395, opened 2024-02-13
perf: add sha224 benchmark

crusso:
Another perf benchmark, using Timo's optimized sha224 as used in base.


luc-blaeser:

PR #4396, opened 2024-02-14
Merge Preparation: Precise Tagging + Graph-Copy-Based Stabilization

luc-blaeser:
Preparing merging https://github.com/dfinity/motoko/pull/4369 in https://github.com/dfinity/motoko/pull/4286.

Includes an explicit translation of the tagging scheme between the 32-bit runtime system and the 64-bit long-term stable format.


luc-blaeser:
Currently not needed for https://github.com/dfinity/motoko/pull/4225 because the stable data format uses the same tags as the 64-bit main memory representation.

PR #4407, opened 2024-02-21
tweaks: detect unused declarations

crusso:
tweaks to #4377:

* don't warn about unused declarations in code from packages (assuming packaces are third party you can't silence them anyway):
  * annotate LibPath Ast nodes with source package, if any, as tracked and determined during import resolution.
  * predicate unused declaration warnings on package origin.
* don't reject unused declarations in the repl treating top-level  code as belonging to fake package "<top-level>" (a mild hack).
   The repl can't know the rest of the interaction so any warning is premature and a nuisance. 
* change terminology of declarations/variables to bindings/indentifiers (for consistency with rest of code)
* add error-code description in M0194.md
* add changelog entry.

Future: we could suppress all warnings, not just unused declarations - from imported package code this way, should we want to.  A --lint mode could re-enable them for further auditing. The rationale is that the warnings are of interest to and actionable on only by the author of the package, not the client.


src/lang_utils/error_codes/M0194.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+or just prefix the identifier with an underscore:
+
+```motoko
+let _nickname = "klutz";
+// code that never uses `nickname`
+```
+
+
```
luc-blaeser, line 33: 👍

luc-blaeser (approved):
Thank you very much. This contains lots of nice improvement over the initial version.

src/lang_utils/error_codes/M0194.md
crusso, line 33:
```suggestion
```

crusso, line 33:
```suggestion
```motoko
let _nickname = "klutz";
// code that never uses `nickname`
```
```

Comment (crusso) [inline src/lang_utils/error_codes/M0194.md:33]:
```suggestion
```

crusso:
```suggestion
```

PR #4424, opened 2024-03-04
feat: custom RTS functions using a Wasm custom section

rvanasa:
Alternative to #4413 using a [Wasm custom section](https://webassembly.github.io/spec/core/appendix/custom.html) in place of compiler flags. This makes it simpler to use Rust bindings without relying on external tooling to pass the expected Rust function names. 

The current format of the custom section is simply a list of RTS function names separated by semicolon characters (`;`), ignoring whitespace adjacent to each name, and ignoring empty strings (e.g. `hello; echo;`). 

In the future, we could extend this format to specify the expected Motoko type in a backwards-compatible way (for example, `hello; echo : <T>(T) -> T;`)


src/wasm-exts/customModuleDecode.ml
```diff
@@ -821,15 +821,16 @@ let utf8 sec_end s =
   let pos = pos s in
   let bytes = get_string (sec_end - pos) s in
   try
-    let _ = Utf8.decode (string s) in
+    let _ = Utf8.decode bytes in
```
rvanasa, line 824: (This is a bugfix in the decoding logic)

PR #4430, opened 2024-03-06
chore: add additional steps in `Building.md` for releasing `base` towards `mops`

ggreif:
Improve the release instructions to make the `mops` upload complete.

@ZenVoich the upload process did not check the `CHANGELOG.md` entry, but somewhere I read that it wouldn't let the upload proceed without a current one. What went wrong?

### DIFF
--- a/Building.md
+++ b/Building.md
@@ -103,10 +103,11 @@ branch to the `next-moc` branch.
 * Change into `motoko-base`
 * `git switch next-moc; git pull`
 * `git switch -c $USER/update-moc-0.11.$MOC_MINOR`
-* Update the `moc_version` env variable in `.github/workflows/{ci, package-set}.yml`
+* Update the `CHANGELOG.md` file with an entry at the top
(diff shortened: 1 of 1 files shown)


ZenVoich:
> @ZenVoich the upload process did not check the `CHANGELOG.md` entry, but somewhere I read that it wouldn't let the upload proceed without a current one. What went wrong?

mops publish doesn't fail if changelog.md file or entry in that file is not found, only a warning is displayed.

It seems base package was published in this action https://github.com/dfinity/motoko-base/actions/runs/8164518203/job/22319915173

So changelog entry should be in `motoko-base` repo https://github.com/dfinity/motoko-base/blob/master/CHANGELOG.md

Is it possible to maintain different changelogs for `moc` compiler and `base` package?
For example `base 0.11.0` package related changelog is only this line
```
Added Option.equal function (thanks to ByronBecker) (dfinity/motoko-base⁠#615).
```

In `base 0.10.4` there is no changes in base package.

chenyan-dfinity (approved):

PR #4432, opened 2024-03-06
experiment: bash scripts to typecheck mops files

crusso:
Some unpolished but working bash scripts to type check (manually listed) mops package against $MOTOKO_BASE.

Might come in handy in future


luc-blaeser:

PR #4434, opened 2024-03-07
bugfix: fix bug affecting `--experimental-rtti`

crusso:
I introduced a bug when making rtti a compiler option.

In particular, in `compile.ml`, the value `BigNum.tag_padded` did not test the new compiler flag properly - it was missing the conventional `env` argument and produced code that was failing to tag compact `Int` and `Nat` values (doing a logical or with the non-rtti tag, i.e. 0).

It would probably be safer to read the flag from the env argument, not directly from `Flags`, but that can come in a follow-on PR.

The good news is that the bug is benign for vanilla moc (and we don't even advertise `--experimental-rtti`)

Ì told you it was experimental.


luc-blaeser (approved):
Good analysis. Luckily, it only affected the experimental mode.
I agree, putting the flag behind `env` would be more robust.

ggreif (approved):
Nice find. This was a sneaky one...

ggreif (commented):
I fixed the `.ok` file

crusso:
I've opened #4440 for good measure (run CI with rtti enabled by default)

PR #4193, opened 2023-09-05
Enhanced Orthogonal Persistence

luc-blaeser:
# Enhanced Orthogonal Persistence
This implements the vision of **enhanced orthogonal persistence** in Motoko that combines:
* **Stable heap**: Persisting the program main memory across canister upgrades.
* **64-bit heap**: Extending the main memory to 64-bit for large-scaled persistence.

As a result, the use of secondary storage (explicit stable memory, dedicated stable data structures, DB-like storage abstractions) will no longer be necessary: Motoko developers can directly work on their normal object-oriented program structures that are automatically persisted and retained across program version changes.

## Activation
Enhanced orthogonal persistence is currently offered for **beta testing** via the compiler flag `--enhanced-orthogonal-persistence`.

To activate enhanced orthogonal persistence under `dfx`, the following command-line argument needs to be specified in `dfx.json`:

```
...
    "type" : "motoko"
    ...
    "args" : "--enhanced-orthogonal-persistence"
...
```

## Advantages
Compared to the existing orthogonal persistence in Motoko, this design offers:
* **Performance**: New program versions directly resume from the existing main memory and have access to the memory-compatible data.
* **Scalability**: The upgrade mechanism scales with larger heaps and in contrast to serialization, does not hit IC instruction limits.

Compared to the explicit use of stable memory, this design improves:
* **Simplicity**: Developers do not need to deal with explicit stable memory.
* **Performance**: No copying to and from the separate stable memory is necessary.

## Design
The enhanced orthogonal persistence is based on the following main properties:
* Extension of the IC to retain main memory on upgrades.
* Supporting 64-bit main memory on the IC.
* A long-term memory layout that is invariant to new compiled program versions.
* A fast memory compatibility check performed on each canister upgrade.
* Incremental garbage collection using a partitioned heap.

### Integration with Classical Persistence
Enhanced orthogonal persistence is offered for **beta testing** via the compiler flag `--enhanced-orthogonal-persistence`.
Classical persistence with 32-bit main memory and Candid stabilization currently remains the default mode.
See `design/PersistenceModes.md` for more information.

### Memory Layout
In a co-design between the compiler and the runtime system, the main memory is arranged in the following structure, invariant of the compiled program version:
* Lower 4MB: Rust call stack.
* Space between 4MB and 4.5MB: Limited reserved space Wasm data segments, only used for the Motoko runtime system.
* Between 4.5MB and 5MB: Persistent metadata.
* Thereafter: Dynamic heap space. Fix start address at 5MB.

[...]

PR #4437, opened 2024-03-08
Enhanced Orthogonal Persistence (64-Bit with Pointer Compression)

luc-blaeser:
Note: This requires adjustments of the IC runtime and execution layers - **the PR is thus not yet ready to be merged.**

# Enhanced Orthogonal Persistence (64-Bit with Pointer Compression)

* **64-bit heap with pointer compression**: Supporting 32GB of main memory without doubling pointers sizes.

## Advantages
* **Instantenous upgrades**: New program versions simply resume on the existing main memory and have access to the memory-compatible data.
* **Scalable upgrades**: The upgrade mechanism scales with larger heaps and in contrast to serialization, does not hit IC instruction limits.

## Design
The stable heap is based on the following main properties:

### IC Extension
The necessary IC extensions are implemented in a separate PR: https://github.com/dfinity/ic/pull/143
This PR is based on these extensions.

### Memory Layout
* Lower 4MB: Rust call stack.

### Compact Pointers
By aligning objects to 16 bytes, object addresses have the lowest 4 bits zero. One bit is spent for pointer tagging (skewing), such that 3 free bits can be used to extend the addressable space by factor of `2**3 = 8`, resulting in 32GB. 

A 64-bit address `a` in the range of 32GB is encoded as a compact pointer `a >> 3 - 1` of 32-bit width.
The compact pointer `p` is again decoded to a 64-bit address `(p + 1) << 3`.

The alignment to 16 bytes increases internal fragmentation (potential padding inside the object). However, the internal fragmentation overhead is expected to  less than generally extending the word size to 64-bit. Without alignment, the minimum object size is 12 bytes (except for the `Null` singleton that fits into 8 bytes).

### Persistent Metadata
* The reference to the null singleton object.

### Compatibility Check
Upgrades are only permitted if the new program version is compatible to the old version, such that the runtime system guarantees a compatible memory structure.

Compatible changes for immutable types are equivalent to the allowed Motoko subtype relation, e.g.
* Removing object fields.
* Adding variant fields.
* `Nat` to `Int`.

This compatibility check serves as an additional safety measure on top of the DFX Candid subtype check that can be bypassed by users (when ignoring a warning). Moreoever, the memory compatibility rules is in some aspects different to the Candid sub-type check:
* Types cannot be made optional.
* Mutable types (aliases) are supported with type invariance.

### Garbage Collection

The garbage collection state needs to be persisted and retained across upgrades. This is because the GC may not yet be completed at the time of an upgrade, such that object forwarding is still in use. The partition table is stored as part of the GC state. 
This is because the GC may not yet be completed at the time of an upgrade, such that object forwarding is still in use. The partition table is stored as part of the GC state.

* The null singleton object.
* The stable type table.

### Main Actor
On an upgrade, the main actor is recreated and existing stable variables are recovered from the persistent root.
The remaining actor variables, the flexible fields as well as new stable variables, are (re)initialized.
As a result, the GC can collect unreachable flexible objects of previous canister versions. 
Unused stable variables of former versions can also be reclaimed by the GC.

### No Static Heap
The static heap is abandoned and former static objects need to be allocated in the dynamic heap.
This is because these objects may also need to survive upgrades and must not be not overwritten by new data segments. 

The incremental GC also operates on these objects, meaning that forwarding pointer resolution is also necessary for these objects.

The runtime systems avoids any global Wasm variables for state that needs to be preserved on upgrades.
Instead, such global runtime state is stored in the persistent metadata.

Sharing optimization (pooling) is possible for compile-time-known objects, see below.

### Wasm Data Segments
Only passive Wasm data segments are used by the compiler and runtime system. In contrast to ordinary active data segments, passive segments can be explicitly loaded to a dynamic address.

This simplifies two aspects: 
* The generated Motoko code can contain arbitrarily large data segments which can loaded to dynamic heap when needed.
* The IC can simply retain the main memory on an upgrade without needing to patch the active data segments of the new program version to the persistent memory.

However, more specific handling is required for the Rust-implemented runtime system:
The Rust-generated active data segments of the runtime system is changed to passive and loaded to the expected static address at the program start (canister initialization and upgrade).
The location and size of the RTS data segments is therefore limited to a defined reserve, see above. 
This is acceptable because the RTS only uses small size for data segments (e.g. 54KB) that is independent of the compiled Motoko program.

### Null Singleton
As an optimization, the top-level `null`` singleton is allocated once in the dynamic heap and remembered in the persistent metadata across upgrades. 
This is necessary to implement null checks by pointer comparison (however, by first resolving pointer forwarding before the comparison).
The null singleton needs to be part of the persistent root set.

### Migration Path
When migrating from the old serialization-based stabilization to the new stable heap, the old data is deserialized one last time from stable memory and then placed in the new stable heap layout.
Once operating on the stable heap, the system prevents downgrade attempts to the old serialization-based persistence.

### Old Stable Memory
The old stable memory remains equally accessible as secondary memory with the new support.

## Current Limitations
* Optimization of statically known objects: Currently, compile-time-known objects are always dynamically allocated. For an improved performance and size, they could be shared in the dynamic heap by remembering them in an additional pool table. The pool table needs to be registered as a transient GC root and recreated on canister upgrades.
* The compact pointer representation only allows 32 GB. Moreover, the GC is also limited to 32 GB because it uses a static partition table.
* Workaround for Rust needed to build PIC (position-independent code) libraries. Explicit use of `emscripten` via LLVM IR. 
* `ic-wasm` would need to be extended to support memory64 and passive data segments. The Wasm optimizations in `test/bench` are thus currently deactivated.
* The Wasm profiler is no longer applicable because the underlying `parity-wasm` crate seems deprecated. A re-implementation of the profiler would be needed.

## Related PRs
* IC support for enhanced orthogonal persistence: https://github.com/dfinity/ic/pull/143
* IC deterministic working set limit for 64-bit main memory: https://github.com/luc-blaeser/ic/pull/1
* Motoko base library 64-bit support: https://github.com/dfinity/motoko-base/pull/589
* Full-fledged 64-bit enhanced orthogonal persistence: https://github.com/dfinity/motoko/pull/4225

Underlying partial implementations:
* 32-bit enhanced orthogonal persistence: https://github.com/dfinity/motoko/pull/4193
* 64-bit support for Motoko (transient heap): https://github.com/dfinity/motoko/pull/4136


luc-blaeser:
This was an experiment. It is only a compromise solution. Probably, for ease of maintenance, we should go for only two modes, the classical support and new full-fledged 64-bit enhanced orthogonal persistence.

PR #4438, opened 2024-03-09
experiment: custom RTS functions

rvanasa:
This PR adds utility macros, traits, and functions which can be used to implement Rust FFI bindings in the RTS. 

I'm currently focusing on just a few Motoko types (primarily `Blob`, `Array`, tuples, and numeric primitives). It's relatively simple to expand support by implementing `FromValue` and `IntoValue` for other Rust types. 

**Rust functions (included in PR for now):**

```rust
#[motoko]
unsafe fn empty() {}

#[motoko]
unsafe fn identity(value: Value) -> Value {
    value
}

#[motoko]
unsafe fn div_rem(a: u32, b: u32) -> (u32, u32) {
    (a / b, a % b)
}

#[motoko]
unsafe fn array_concat(a: Vec<Value>, b: Vec<Value>) -> Vec<Value> {
    [a, b].concat()
}

#[motoko]
unsafe fn blob_modify(mut blob: BlobVec) -> BlobVec {
    blob.0.push('!' as u8);
    blob
}

#[motoko]
unsafe fn manual_alloc(#[memory] mem: &mut impl Memory) -> Value {
    // Low-level access to memory allocation
    let value = alloc_blob(mem, Bytes(3 as u32));
    let blob = value.as_blob_mut();
    let mut dest = blob.payload_addr();
    for i in 0..3 {
        *dest = (i + 1) * 0x11;
        dest = dest.add(1);
    }
    allocation_barrier(value)
}

#[motoko]
unsafe fn bool_swap(a: bool, b: bool) -> (bool, bool) {
    (b, a)
}

#[motoko]
unsafe fn check_numbers(
    a: u8,
    b: i8,
    c: u16,
    d: i16,
    e: u32,
    f: i32,
    g: u64,
    h: i64,
) -> (u8, i8, u16, i16, u32, i32, u64, i64) {
    (a, b, c, d, e, f, g, h)
}
```

**Motoko usage (`ffi.mo`):**

```motoko
import Prim "mo:prim";
import Array "mo:base/Array";
import Blob "mo:base/Blob";

// Rust bindings
func empty() : () = (prim "rts:empty" : () -> ())();
func identity<T>(value : T) : T = (prim "rts:identity" : T -> T)(value);
func blob_modify(value : Blob) : Blob = (prim "rts:blob_modify" : Blob -> Blob)(value);
func array_concat<T>(a : [T], b : [T]) : [T] = (prim "rts:array_concat" : ([T], [T]) -> [T])(a, b);
func manual_alloc() : Blob = (prim "rts:manual_alloc" : () -> Blob)();
func div_rem(a : Nat32, b : Nat32) : (Nat32, Nat32) = (prim "rts:div_rem" : (Nat32, Nat32) -> (Nat32, Nat32))(a, b);
func bool_swap(a : Bool, b : Bool) : (Bool, Bool) = (prim "rts:bool_swap" : (Bool, Bool) -> (Bool, Bool))(a, b);
type Numbers = (Nat8, Int8, Nat16, Int16, Nat32, Int32, Nat64, Int64);
func check_numbers(a : Nat8, b : Int8, c : Nat16, d : Int16, e : Nat32, f : Int32, g : Nat64, h : Int64) : Numbers = (prim "rts:check_numbers" : Numbers -> Numbers)(a, b, c, d, e, f, g, h);

// `empty`
assert empty() == ();

// `identity`
let echoValue = identity(5);
Prim.debugPrint(debug_show echoValue);
assert echoValue == 5;

// `div_rem`
let (div, rem) = div_rem(7, 2);
assert (div, rem) == (3, 1);

// `array_concat`
let a = Array.freeze(Array.init<Nat8>(10_000_000, 123 : Nat8));
let b = Array.freeze(Array.init<Nat8>(500_000, 234 : Nat8));
let concat = array_concat(a, b);
assert concat.size() == a.size() + b.size();
assert concat[0] == 123;
assert concat[concat.size() - 1] == 234;

// `blob_modify`
let inputBlob = Blob.fromArray(Array.freeze(Array.init<Nat8>(10_000_000, 123 : Nat8)));
let blob = blob_modify(inputBlob);
let array = Blob.toArray(blob);
assert array[0] == 123;
assert array[array.size() - 1] == 33; // '!'
assert blob.size() == inputBlob.size() + 1;

// `manual_alloc`
let allocValue = manual_alloc();
assert Blob.toArray(allocValue) == [0x11, 0x22, 0x33];

// `bool_swap`
for (a in [true, false].vals()) {
    for (b in [true, false].vals()) {
        assert bool_swap(a, b) == (b, a);
    }
};

// `check_numbers`
let numbers: Numbers = (1, -2, 3333, -4444, 5_000_000, -5_000_000, 0, -1_000_000_000_000_000);
assert check_numbers(numbers) == numbers;
```

Try this yourself (with placeholders `ffi.mo` and `../motoko-base`):

```bash
MOC_UNLOCK_PRIM=1 moc -c ffi.mo -wasi-system-api --package base ../motoko-base/src && wasmtime ffi.wasm
```

**Changes:**

* [x] `rts_sections` in Wasm module decoder
* [x] `custom_rts_functions` field in compilation environment
* [x] `"rts:*"` primitive functions which refer to names in the custom section
* [x] Bugfix for decoding custom sections with UTF-8 content
* [x] `FromValue` and `IntoValue` traits in RTS
* [x] `#[motoko]` procedural macro attribute which wraps `#[ic_mem_fn]` and generates a custom section
* [x] Bump `proc-macro2` and `syn` in the `motoko-rts-macros` crate
* [x] Example RTS functions using `#[motoko]` attribute
* [x] Macro to implement `FromValue` and `IntoValue` for tuples
* [ ] Re-vendor Cargo dependencies in Nix (are [these instructions](https://github.com/dfinity/motoko/tree/master/rts#rust-build) up to date?)
* [ ] Type checking or runtime error for unknown `"rts:*"` primitive expressions?
* [ ] Convert examples into tests


src/wasm-exts/customModuleDecode.ml
```diff
@@ -821,15 +821,15 @@ let utf8 sec_end s =
   let pos = pos s in
   let bytes = get_string (sec_end - pos) s in
   try
-    let _ = Utf8.decode (string s) in
+    let _ = Utf8.decode bytes in
```
rvanasa, line 824: (This is a bugfix in the decoding logic)

rts/motoko-rts/src/custom.rs
```diff
@@ -0,0 +1,326 @@
+// Custom RTS function utilities
+
+use alloc::vec::Vec;
```
luc-blaeser, line 3: This would potentially involve dynamic allocation using the Rust-interop `EmphermalAllocator` in `allocator.rs` allocating `Blob` behind the scenes. Did you check how many implicit allocations happen? I wonder whether we could possibly avoid the `Vec` and use custom convenience wrapper around `Blob` and `Array` for implementing FFI functions.

rts/motoko-rts/src/custom.rs
```diff
@@ -0,0 +1,326 @@
...
+};
+
+extern "C" {
+    fn unit() -> Value;
+
+    fn u8_from_nat8(value: Value) -> u8;
+    fn u16_from_nat16(value: Value) -> u16;
+    fn u32_from_nat32(value: Value) -> u32;
+    fn u64_from_nat64(value: Value) -> u64;
+
```
luc-blaeser, line 19: Maybe, we could also add `float` conversions?

rts/motoko-rts/src/custom.rs
```diff
@@ -0,0 +1,326 @@
...
+        let vec = self.0;
+        let value = alloc_blob(mem, Bytes(vec.len() as u32));
+        let blob = value.as_blob_mut();
+        let mut dest = blob.payload_addr();
+        for item in vec.into_iter() {
+            *dest = item;
+            dest = dest.add(1);
+        }
+        Ok(allocation_barrier(value))
+        // ^^ Is this allocation barrier necessary?
```
luc-blaeser, line 193: Yes, the barrier would be essential, as otherwise a returned `Blob` could be discarded by the GC (if allocated concurrently during a GC run), potentially resulting in dangling pointers.

rts/motoko-rts/src/custom.rs
```diff
@@ -0,0 +1,326 @@
...
+impl<T: FromValue> FromValue for Vec<T> {
+    unsafe fn from_value(value: Value, mem: &mut impl Memory) -> MotokoResult<Self> {
+        match value.tag() {
+            TAG_ARRAY => {
+                let array = value.as_array();
+                let len = array.len();
+                (0..len)
+                    .into_iter()
+                    .map(|i| T::from_value(array.get(i), mem))
+                    .collect()
```
luc-blaeser, line 206: Not sure whether a dynamic allocation here happens via our `EmphemeralAllocator`.

rts/motoko-rts/src/custom.rs
```diff
@@ -0,0 +1,326 @@
...
+
+// Implement `FromValue` and `IntoValue` for tuples
+#[tuple_macro]
+macro_rules! tuple_impl {
+    ($len:expr; $($name:ident = $index:tt),+) => {
+        impl<$($name: FromValue),+> FromValue for ($($name),+) {
+            unsafe fn from_value(value: Value, mem: &mut impl Memory) -> MotokoResult<Self> {
+                const LENGTH: u32 = $len;
+                match value.tag() {
+                    TAG_ARRAY => {
```
luc-blaeser, line 244: Potentially, during GC the tag could also be a tag above `TAG_ARRAY_SLICE_MIN` (due to the incremental array slicing during the marking and updating phase in the GCs).

luc-blaeser, line 244:
```suggestion
                    TAG_ARRAY | TAG_ARRAY_SLICE_MIN.. => {
```

rts/motoko-rts/src/custom.rs
```diff
@@ -0,0 +1,326 @@
...
+                }
+            }
+        }
+        impl<$($name: IntoValue),+> IntoValue for ($($name),+) {
+            unsafe fn into_value(self, mem: &mut impl Memory) -> MotokoResult<Value> {
+                const LENGTH: u32 = $len;
+                let value = alloc_array(mem, LENGTH);
+                let array = value.as_array();
+                let dest = array.payload_addr();
+                $(*(dest.add($index)) = $name::into_value(self.$index, mem)?;)+
```
luc-blaeser, line 261: Here, we would have to use `Array.initialize()` as otherwise a initialization barrier may be missing (resolving forwarding pointers during incremental update phase).

luc-blaeser, line 261:
```suggestion
                $(array.initialize($index, $name::into_value(self.$index, mem)?, mem);)+
```

src/codegen/compile.ml
```diff
@@ -6154,18 +6171,84 @@ module RTS_Exports = struct
...
+    export_fun env "int64_from_i64" (
+      Func.of_body env ["v", I64Type] [I32Type] (fun env ->
+        G.i (LocalGet (nr 0l)) ^^ BoxedWord64.box env Type.Int64));
+    export_fun env "nat64_from_u64" (
+      Func.of_body env ["v", I64Type] [I32Type] (fun env ->
+        G.i (LocalGet (nr 0l)) ^^ BoxedWord64.box env Type.Nat64));
+
+    export_fun env "i8_from_int8" (
+      Func.of_body env ["v", I32Type] [I32Type] (fun env ->
+        G.i (LocalGet (nr 0l)) ^^ TaggedSmallWord.lsb_adjust Type.Int8));
```
luc-blaeser, line 6226: Probably, `TaggedSmallWord.untag` should be applied too before `lsb_adjust`, e.g. to sanity check that the tags are correct.

src/codegen/compile.ml
```diff
@@ -11730,6 +11813,22 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
+    else (
+      (* TODO: type checking error *)
+      Printf.printf "%s" Diag.(string_of_message {
+        sev = Error;
+        code = "M0199";
+        at;
+        cat = "RTS";
+        text = Printf.sprintf "custom function '%s' not found" s';
+      });
+      exit 1)
```
luc-blaeser, line 11830: As mentioned in the TODO, raising an error would be the nicer way. This would need some handling by the compiler pipeline.

rts/motoko-rts-macros/src/lib.rs
```diff
@@ -167,3 +167,160 @@ pub fn non_incremental_gc(attr: TokenStream, input: TokenStream) -> TokenStream
...
+            output = quote!(
+                #output
+                #macro_ident!(#length_syn; #parts);
+            );
+        }
+    }
+
+    // return syn::Error::new_spanned(quote!(), format!("Macro expansion:\n{}", output))
+    //     .to_compile_error()
+    //     .into();
```
luc-blaeser, line 212: Just cosmetics: The commented code could be removed.

rts/motoko-rts-macros/src/lib.rs
```diff
@@ -167,3 +167,160 @@ pub fn non_incremental_gc(attr: TokenStream, input: TokenStream) -> TokenStream
...
+    Memory,
+}
+
+/// Utility macro to implement traits for n-length tuples.
+#[proc_macro_attribute]
+pub fn tuple_macro(_attr: TokenStream, input: TokenStream) -> TokenStream {
+    // Currently hard-coded. If we use this anywhere else, it could make
+    // sense to configure these in the attribute, e.g. `#[tuple_impl(2, 20)]`
+    let min_length: u32 = 2;
+    let max_length: u32 = 20;
```
luc-blaeser, line 182: Keeping it configurable would probably be nicer.

rts/motoko-rts-macros/src/lib.rs
```diff
@@ -167,3 +167,160 @@ pub fn non_incremental_gc(attr: TokenStream, input: TokenStream) -> TokenStream
...
+    let args: Vec<proc_macro2::TokenStream> = fn_params
+        .iter()
+        .map(|(sort, ident, _)| match sort {
+            ArgSort::Value => quote!(#ident),
+            ArgSort::Memory => quote!(#memory_ident),
+        })
+        .collect();
+
+    // Motoko return value
+    let ret = quote!(crate::types::Value);
```
luc-blaeser, line 289: I am not sure whether all FFI functions would have a `Value` return type. Maybe there are some functions without return type (e.g. that implement checks with traps or do some updates).

rts/motoko-rts-macros/src/lib.rs
```diff
@@ -167,3 +167,160 @@ pub fn non_incremental_gc(attr: TokenStream, input: TokenStream) -> TokenStream
...
+            crate::custom::IntoValue::into_value(ret, #memory_ident).unwrap()
+        }
+
+        #[link_section = "rts:custom-functions"]
+        static #custom_section_ident: [u8; #custom_section_len] = *#custom_section_bytes;
+    );
+
+    // return syn::Error::new_spanned(quote!(), format!("Macro expansion:\n{}", output))
+    //     .to_compile_error()
+    //     .into();
```
luc-blaeser, line 323: Cosmetics: Perhaps remove the commented code

luc-blaeser:
Very nice PR, Ryan. Thanks a lot. This offers a well-structured, convenient small framework for FFI implementations. As you say, it would still require advanced knowledge to implement FFI. (Especially, also the GC aspects, e.g. keep Rust pointers only temporarily, applying the right GC barriers etc.).
The only worry I have is that users could easily break the memory safety and that we would then get issue reports of memory corruptions in Motoko (which could be time-consuming to invest and also maybe influence the safety reputation of Motoko). I guess people could still do this today by adjusting Motoko compiler/RTS on their own, but I believe now it would be easier. I wonder if we could reduce this risk, i.e. instruct users about all the safety/security rules and aspects for FFI functions, have an explicit opt-in for this, and/or apply additional steps when triaging issue reports that we can filter out Motoko code where users apply FFI functions (e.g. having a question before reporting to confirm no FFI was used). Maybe my worry is exaggerated. I am interested what others team colleagues think, @crusso , @ggreif , @chenyan-dfinity.

luc-blaeser:
PS: I believe we could add some more tests for the FFI. I could also do some more stress testing with the GC, e.g. composing an additional GC random test or benchmark case that makes use of FFI.

rts/motoko-rts/src/custom.rs
crusso: 'char' would be useful too.

rts/motoko-rts/src/custom.rs
crusso: If we don't define this trait, do we steer people into avoiding arguably dubious examples like array_concat?

rts/motoko-rts/src/custom.rs
```diff
@@ -0,0 +1,326 @@
...
+    }
+}
+
+impl FromValue for BlobVec {
+    unsafe fn from_value(value: Value, _mem: &mut impl Memory) -> MotokoResult<Self> {
+        match value.tag() {
+            TAG_BLOB => {
+                let blob = value.as_blob();
+                let len = blob.len().as_u32();
+                Ok(BlobVec((0..len).into_iter().map(|i| blob.get(i)).collect()))
```
crusso, line 176: This is probably also hiding uses of the ephemeral allocator. I assume collect is allocating a Vec?

rts/motoko-rts-macros/src/lib.rs
rvanasa: Replaced the commented code with conditional evaluation based on a `DEBUG` constant. There might be a better way to check the macro expansion, but this is the simplest solution that I've been able to find for gaining visibility into the generated code.

rvanasa:
> The only worry I have is that users could easily break the memory safety and that we would then get issue reports of memory corruptions in Motoko (which could be time-consuming to invest and also maybe influence the safety reputation of Motoko).

This is a really good point @luc-blaeser, especially given the precedent of doing something similar in the past and getting bug reports due to misuse of the low-level access. Because safety is a key part of Motoko's brand, this by itself makes a fairly strong case for developers to avoid this FFI approach if possible. One possibility could be to repurpose this PR as an internal refactor (implementing the built-in RTS functions using the `#[motoko]` macro). We could potentially keep the `rts:custom-functions` Wasm custom section to give advanced developers the option to extend the RTS where it would otherwise be impossible to use Motoko for their use case. 

While this functionality is currently opt-in via the `MOC_*_RTS` environment variables, I suppose we could also include a compiler flag or something that explicitly allows custom RTS functions (or maybe even switch back to using the original logic in https://github.com/dfinity/motoko/pull/4413). Also interested to hear more opinions from the rest of the team about how we could address this.

luc-blaeser:

PR #4449, opened 2024-03-13
fix: passing `<system>` to class methods from bodies

### LINKED ISSUE #4447: Motoko 0.11 `<system>` not working for class bodies?
From what I understand the classes should just put a <system> type as the first type and calling system functions should work.
But I have this code and 
```motoko
public class Dao<system, TProposalContent>(
        data : StableData<TProposalContent>,
        onExecute : Proposal<TProposalContent> -> async* (),
        onReject : Proposal<TProposalContent> -> async* (),
    ) {
     // ....

        public func resetEndTimers<system>() {
            for (proposal in proposals.vals()) {
                switch (proposal.endTimerId) {
                    case (null) ();
                    case (?id) Timer.cancelTimer(id);
                };
                proposal.endTimerId := null;
                if (proposal.status == #open) {
                    let proposalDurationNanoseconds = durationToNanoseconds(proposalDuration);
                    let endTimerId = createEndTimer<system>(proposal.id, proposalDurationNanoseconds);
                    proposal.endTimerId := ?endTimerId;
                };
            };
        };
// ...
        resetEndTimers<system>(); // Always reset end timers on init
 ````
It gives me an error with resetEndTimers about `system` capability required, but not available
image

Am I doing this right? It seems from the docs that classes are allowed

ggreif:
This (hopefully) resolves #4447.

The problem was that the two modes of bidirectional type checking behaved differently, because of a bug in setting up the checking environment. That has been fixed by @crusso in #4453, where he also corrected the issue with the error message showing up twice in the test case.

When calling `setTimer` from the class body type _checking_ happens, while when calling a self-method it is _infer_.

### DIFF
--- a/test/fail/issue-4447.mo
+++ b/test/fail/issue-4447.mo
@@ -1,10 +1,21 @@
 import { debugPrint; setTimer } = "mo:⛔";
 
-func() {
+func _f() {
 
     func indirect<system>() {
         ignore setTimer<system>(0, false, func () : async () { debugPrint "YEP!" });
     };
 
-    indirect<system>()
+    indirect<system>();
+    ignore indirect<system>()
+};
+
+class _C() {
+
+    func indirect<system>() {
+        ignore setTimer<system>(0, false, func () : async () { debugPrint "YEP!" });
+    };
+
+    indirect<system>();
+    ignore indirect<system>()
 }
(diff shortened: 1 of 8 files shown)


ggreif:
closing, as @crusso has cracked it with #4453!

Issue #4418, opened 2024-02-26
Bug: Making actor from awaited string fails.


AStepanov25:
NOTE: It works on [https://embed.smartcontracts.org/](https://embed.smartcontracts.org/), but fails locally.
The code:
```
actor {
  public shared func a() : async () {
    let a = actor (await async { "" }) : actor {};
  };
};
```
Causes:
```
Ill-typed intermediate code after Desugaring (use -v to see dumped IR):
/home/sample/src/main.mo:3.13-3.39: IR type error [M0000], inferred effect not a subtype of expected effect
Raised at Ir_def__Check_ir.error.(fun) in file "ir_def/check_ir.ml", line 95, characters 30-92
Called from Ir_def__Check_ir.check_exp in file "ir_def/check_ir.ml", line 386, characters 2-100
Called from Ir_def__Check_ir.check_dec in file "ir_def/check_ir.ml", line 1100, characters 4-21
Called from Stdlib__list.iter in file "list.ml", line 110, characters 12-15
Called from Ir_def__Check_ir.check_exp in file "ir_def/check_ir.ml", line 703, characters 4-22
Called from Ir_def__Check_ir.check_exp in file "ir_def/check_ir.ml", line 745, characters 4-23
Called from Ir_def__Check_ir.check_exp in file "ir_def/check_ir.ml", line 786, characters 4-40
Called from Ir_def__Check_ir.check_dec in file "ir_def/check_ir.ml", line 1100, characters 4-21
Called from Stdlib__list.iter in file "list.ml", line 110, characters 12-15
Called from Ir_def__Check_ir.check_comp_unit in file "ir_def/check_ir.ml", line 1159, characters 4-23
Called from Ir_def__Check_ir.check_prog in file "ir_def/check_ir.ml", line 1178, characters 6-28
```
Dfx: 0.17.0

timohanke:
It works in the interpreter. So maybe a compiler issue.

crusso:
That's a bug, thanks for reporting it!

crusso:
Fixed in #4450

PR #4450, opened 2024-03-13
bugfix: repro and fix for #4418

crusso:
Fixes issue #4418 caused by unsound effect inference for actor references.


luc-blaeser (approved):
Thanks for fixing this.

ggreif (approved):
LGTM!

Issue #4448, opened 2024-03-12
Motoko 11 used actor class marked as `unused idetifier`


Gekctek:
Im getting this warning that I am not using `StadiumActor` identifier
![Yellow](https://github.com/dfinity/motoko/assets/1934953/86bb29d3-4118-4f2f-928b-72b2b06c12ad)

But i know I am using it here, in another file

![Use](https://github.com/dfinity/motoko/assets/1934953/bb48b2b9-611b-4c17-91b9-5b06097e9651)

But if I remove it as the warning suggests, it breaks

crusso:
Yep, that's annoying and wrong. Thanks for reporting it!

luc-blaeser:
Thanks for reporting. This is fixed by https://github.com/dfinity/motoko/pull/4452

PR #4452, opened 2024-03-13
Exempt top-level actor class identifier from unused detection

luc-blaeser:
A possible fix for https://github.com/dfinity/motoko/issues/4448.

The unnecessary warning appears because the files are analyzed separately in VSCode. When compiled, e.g. with `dfx`, the actor class would not be reported unused if it is instantiated in the main program or a used module.

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -2529,6 +2529,8 @@ and infer_dec env dec : T.typ =
       let cs, tbs, te, ce = check_typ_binds env typ_binds in
       let env' = adjoin_typs env te ce in
       let in_actor = obj_sort.it = T.Actor in
+      (* Top-level actor class identifier is implicitly public and thus considered used. *)
+      if env.in_prog && in_actor then use_identifier env id.it;
       let t_pat, ve =
         infer_pat_exhaustive (if in_actor then error else warn) env' pat
       in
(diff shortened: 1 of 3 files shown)


crusso (commented):
LGTM

I wonder if we can just the same thing for other top-level declaration to avoid complaining about unused library/actor names

luc-blaeser:
> LGTM
> 
> I wonder if we can just the same thing for other top-level declaration to avoid complaining about unused library/actor names

Thanks. I agree, we should exempt the other top-level declarations too. I will prepare the corresponding adjustment.

crusso:
> > LGTM
> > I wonder if we can just the same thing for other top-level declaration to avoid complaining about unused library/actor names
> 
> Thanks. I agree, we should exempt the other top-level declarations too. I will prepare the corresponding adjustment.

This is a good example:
```
class A() {};
actor B {};
```
(issue-1938.mo)

crusso:
Running ./check.sh  in
motoko/doc/md/examples

should be a good test (lots of name top-level actors in there)

luc-blaeser:
**Update**: I thought to limit the exemption to actor classes, actors, and modules, if occurring in top-level program scope. With this, we can still find unused variables, functions etc. in the top-level scope. And in particular, we also detect unused imports.

luc-blaeser:
Update: I will open a separate PR for improving the warning message of unused field patterns.

crusso (commented):
LGTM, thanks for doing this chore!

luc-blaeser:
Thanks @crusso  for reviewing!

PR #4469, opened 2024-03-22
fix: add special case for displaying actor principal in `moc.js` output

rvanasa:
This PR includes a special case in the `run_stdin_from_file` console output to address a quirk with `moc.js` in the developer documentation.


rvanasa:
Closing in favor of #4471.

PR #4471, opened 2024-03-22
feat: use static type information for value pretty-printing logic

### LINKED ISSUE #4470: FR: Render actor ids using IC textual format, not blob format, in interpreter output.
This is mostly to improve the repl output on IC docs.

https://dfinity.slack.com/archives/CPL67E7MX/p1711059148385989

image

rvanasa:
* Refactors `pp_val` and related functions to receive type information (using `Type.Any` by default in some cases)
* Displays the text representation of actor IDs with surrounding backticks, such as `` `ys6dh-5cjiq-5dc` : actor {...}``

Resolves #4470.

### DIFF
--- a/src/mo_values/value.ml
+++ b/src/mo_values/value.ml
@@ -1,4 +1,5 @@
 open Numerics
+module T = Mo_types.Type
 
 (* Environments *)
 
@@ -180,7 +181,9 @@ let comma ppf () = fprintf ppf ",@ "
 
 let semi ppf () = fprintf ppf ";@ "
 
-let rec pp_val_nullary d ppf = function
+let rec pp_val_nullary d ppf (t, v : T.typ * value) =
+  let t = T.normalize t in
+  match t, v with
   | _, Null -> pr ppf "null"
   | _, Bool b -> pr ppf (if b then "true" else "false")
   | _, Int n when Int.(ge n zero) -> pr ppf (Int.to_pretty_string n)
@@ -195,24 +198,27 @@ let rec pp_val_nullary d ppf = function
   | _, Float f -> pr ppf (Float.to_pretty_string f)
   | _, Char c ->  pr ppf (string_of_string '\'' [c] '\'')
   | _, Text t -> pr ppf (string_of_string '\"' (Lib.Utf8.decode t) '\"')
-  | Obj (Actor, _), Blob b -> pr ppf (string_of_string '`' (Lib.Utf8.decode (Ic.Url.encode_principal b)) '`')
+  | T.Obj _, Blob b -> pr ppf (string_of_string '`' (Lib.Utf8.decode (Ic.Url.encode_principal b)) '`')
   | _, Blob b -> pr ppf ("\"" ^ Blob.escape b ^ "\"")
-  | Tup ts, Tup vs ->
+  | T.Tup ts, Tup vs ->
     fprintf ppf "@[<1>(%a%s)@]"
-      (pp_print_list ~pp_sep:comma (pp_val d ppf)) (List.combine ts vs)
+      (pp_print_list ~pp_sep:comma (pp_val d)) (List.combine ts vs)
       (if List.length vs = 1 then "," else "")
-  | _, Obj ve ->
+  | T.Obj (_, fs), Obj ve ->
     if d = 0 then pr ppf "{...}" else
     fprintf ppf "@[<hv 2>{@;<0 0>%a@;<0 -2>}@]"
-      (pp_print_list ~pp_sep:semi (pp_field d)) (Env.bindings ve)
-  | Array t, Array a ->
+      (pp_print_list ~pp_sep:semi (pp_field d)) (List.map (fun (lab, v) ->
+          let t = (List.find_map (fun (T.{ lab = lab'; typ; _ }) -> if lab = lab' then Some typ else None)) fs in
+          (lab, Option.value t ~default:T.Any, v))
+        (Env.bindings ve))
+  | T.Array t, Array vs ->
     fprintf ppf "@[<1>[%a]@]"
-      (pp_print_list ~pp_sep:comma (fun ppf v -> pp_val d ppf (t, v))) (Array.to_list a)
+      (pp_print_list ~pp_sep:comma (pp_val d)) (List.map (fun v -> (t, v)) (Array.to_list vs))
   | _, Func (_, _) -> pr ppf "func"
   | _, Comp _ -> pr ppf "async*"
   | t, v ->
     (* "(" ^ string_of_val d v ^ ")" *)
-    fprintf ppf "@[<1>(%a)@]" (pp_val d ppf) (t, v)
+    fprintf ppf "@[<1>(%a)@]" (pp_val d) (t, v)
 
 and pp_field d ppf (lab, t, v) =
     fprintf ppf "@[<2>%s =@ %a@]" lab (pp_val d) (t, v)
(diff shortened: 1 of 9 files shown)


src/mo_values/value.ml
```diff
@@ -180,75 +181,98 @@ let comma ppf () = fprintf ppf ",@ "
...
+  | _, Variant (l, Tup []) -> fprintf ppf "#%s" l
+  | t, Variant (l, v) ->
+    let fs = match t with  T.Variant fs -> fs | _ -> [] in
+    let t' = Option.value ~default:T.Any
+      (List.find_map (fun T.{ lab; typ; _ } -> if lab = l then Some typ else None) fs) in
+    (match v with
+    | Tup vs -> fprintf ppf "@[#%s@;<0 1>%a@]" l (pp_val d) (t', Tup vs)
+    | _ -> fprintf ppf "@[#%s@;<0 1>(%a)@]" l (pp_val d) (t', v))
+  | t, Async {result; waiters = []} ->
+    let t' = match t with T.Opt t' -> t' | _ -> T.Any in
```
crusso, line 250:
```suggestion
    let t' = match t with T.Async _, t' -> t' | _ -> T.Any in
```

src/mo_values/value.ml
```diff
@@ -180,75 +181,98 @@ let comma ppf () = fprintf ppf ",@ "
...
+    fprintf ppf "@[<2>async@ %a@]" (pp_res d) (t', result)
+  | t, Async {result; waiters} ->
+    let t' = match t with T.Opt t' -> t' | _ -> T.Any in
```
crusso, line 253:
```suggestion
    let t' = match t with T.Async _, t' -> t' | _ -> T.Any in
```

src/mo_values/value.ml
```diff
@@ -180,75 +181,98 @@ let comma ppf () = fprintf ppf ",@ "
...
+    let fs = match t with  T.Obj (_, fs) -> fs | _ -> [] in
     if d = 0 then pr ppf "{...}" else
     fprintf ppf "@[<hv 2>{@;<0 0>%a@;<0 -2>}@]"
-      (pp_print_list ~pp_sep:semi (pp_field d)) (Env.bindings ve)
-  | Array a ->
+      (pp_print_list ~pp_sep:semi (pp_field d)) (List.map (fun (lab, v) ->
+          let t = (List.find_map (fun T.{ lab = lab'; typ; _ } -> if lab = lab' then Some typ else None)) fs in
+          (lab, Option.value t ~default:T.Any, v))
+        (Env.bindings ve))
+  | t, Array vs ->
```
crusso, line 219: I wonder if we can do better here an print out a [var ...] or [ ... ] when t' is T.Mut t or other

src/mo_values/value.ml
```diff
@@ -180,75 +181,98 @@ let comma ppf () = fprintf ppf ",@ "
...
       (if List.length vs = 1 then "," else "")
-  | Obj ve ->
+  | t, Obj ve ->
+    let fs = match t with  T.Obj (_, fs) -> fs | _ -> [] in
     if d = 0 then pr ppf "{...}" else
     fprintf ppf "@[<hv 2>{@;<0 0>%a@;<0 -2>}@]"
-      (pp_print_list ~pp_sep:semi (pp_field d)) (Env.bindings ve)
-  | Array a ->
+      (pp_print_list ~pp_sep:semi (pp_field d)) (List.map (fun (lab, v) ->
+          let t = (List.find_map (fun T.{ lab = lab'; typ; _ } -> if lab = lab' then Some typ else None)) fs in
```
crusso, line 216:
```suggestion
          let t = Type.lookup_val_field_opt lab fs in
```
Lookup_val_field_opt correctly distinguishes between type fields and value fields -these actually share the same namespace. I think your code could be confused by `object { type x = int; val  x = 1}`.

crusso: I guess this also raises the question of printing out type fields or suppressing them. Before, when we only had the value to hand, we could only print out the value fields. But now that we have the type available, we can print out both.

src/pipeline/pipeline.ml
```diff
@@ -53,16 +53,16 @@ let print_dyn_ve scope =
       Format.printf "@[<hv 2>%s %s :@ %a =@ %a@]@."
         (if t == t' then "let" else "var") x
```
crusso, line 54: That's such a hack. I think that was Andreas...

src/mo_values/value.ml
```diff
@@ -180,75 +181,98 @@ let comma ppf () = fprintf ppf ",@ "
...
+  | _, Int32 i -> pr ppf (Int_32.(pos_sign (gt i zero) ^ to_pretty_string i))
+  | _, Int64 i -> pr ppf (Int_64.(pos_sign (gt i zero) ^ to_pretty_string i))
+  | t, Opt v ->
+    let t' = match t with T.Opt t' -> t' | _ -> T.Any in
+    fprintf ppf "@[<1>?%a@]" (pp_val_nullary d) (t', v)
+  | _, Variant (l, Tup []) -> fprintf ppf "#%s" l
+  | t, Variant (l, v) ->
+    let fs = match t with  T.Variant fs -> fs | _ -> [] in
+    let t' = Option.value ~default:T.Any
+      (List.find_map (fun T.{ lab; typ; _ } -> if lab = l then Some typ else None) fs) in
```
crusso, line 245:
```suggestion
      (Type.find_val_field_opt l fs) in
```

crusso (commented):
Instead of treating Any specially, I wonder if it would be better to have two functions: one that pretty prints the raw values (for use in tracing where types aren't avialable), and another which only prints what's visible from the type.

rvanasa:
> Instead of treating Any specially, I wonder if it would be better to have two functions: one that pretty prints the raw values (for use in tracing where types aren't available), and another which only prints what's visible from the type.

(We talked about this in a Zoom meeting and decided to stick with the current approach)

crusso:
Ryan, I created a follow on PR here #4474. See what you think.

src/mo_values/value.ml
```diff
@@ -180,75 +181,113 @@ let comma ppf () = fprintf ppf ",@ "
...
+    | Async {result; waiters} ->
+      let t' = match t with T.Async (_, _, t') -> t' | _ -> T.Non in
+      fprintf ppf "@[<2>async[%d]@ %a@]"
+        (List.length waiters) (pp_res d) (t', result)
+    | Mut r ->
+      let t' = match t with T.Mut t' -> t' | _ -> T.Non in
+      pp_val d ppf (t', !r)
+    | v -> pp_val_nullary d ppf (t, v)
+
+and pp_res d ppf (t, result) =
```
crusso, line 276:
```suggestion
and pp_res d ppf t result =
```
Same here

src/mo_values/value.mli
```diff
@@ -113,8 +113,8 @@ val compare : value -> value -> int
 
 (* Pretty Printing *)
 
-val pp_val : int -> Format.formatter -> value -> unit
-val pp_def : int -> Format.formatter -> def -> unit
+val pp_val : int -> Format.formatter -> (Type.typ * value) -> unit
+val pp_def : int -> Format.formatter -> (Type.typ * def) -> unit
```
crusso, line 117:
```suggestion
(* NB: pass Type.Non to print at dynamic, not static, type *)
val pp_val : int -> Format.formatter -> Type.typ  -> value -> unit
val pp_def : int -> Format.formatter -> Type.typ -> def -> unit
```

src/mo_values/value.mli
```diff
@@ -113,8 +113,8 @@ val compare : value -> value -> int
...
 
-val string_of_val : int -> value -> string
-val string_of_def : int -> def -> string
+val string_of_val : int -> Type.typ -> value -> string
```
crusso, line 119:
```suggestion
(* NB: pass Type.Non to print value at full dynamic, not static, type *)
val string_of_val : int -> Type.typ -> value -> string
```

crusso (approved):
LGTM!

PR #4475, opened 2024-03-27
Enhanced Orthogonal Persistence (64-Bit with Graph Copy)

luc-blaeser:
Note: This requires adjustments of the IC runtime and execution layers - **the PR is thus not yet ready to be merged.**

### PR Stack
Enhanced orthogonal persistence support is structured in three PRs to ease review:
1. https://github.com/dfinity/motoko/pull/4286 **<-- this PR**
2. https://github.com/dfinity/motoko/pull/4225
3. https://github.com/dfinity/motoko/pull/4193

# Enhanced Orthogonal Persistence (64-Bit with Graph Copy)

* **64-bit heap**: Extending the main memory to 64-bit for large-scaled persistence.

## Advantages
* **Performance**: New program versions directly resume from the existing main memory and have access to the memory-compatible data.
* **Scalability**: The upgrade mechanism scales with larger heaps and in contrast to serialization, does not hit IC instruction limits.

## Design
The enhanced orthogonal persistence is based on the following main properties:
* Supporting 64-bit main memory on the IC.

### IC Extension
The necessary IC extensions are implemented in a separate PR: https://github.com/dfinity/ic/pull/139. This PR is based on these extensions.

### Memory Layout
* Lower 4MB: Rust call stack.

### Persistent Metadata

### Compatibility Check
Upgrades are only permitted if the new program version is compatible with the old version, such that the runtime system guarantees a compatible memory structure.

Compatible changes for immutable types are largely analogous to the allowed Motoko subtype relation, e.g.
* Removing object fields.
* Adding variant fields.
* `Nat` to `Int`.

This compatibility check serves as an additional safety measure on top of the DFX Candid subtype check that can be bypassed by users (when ignoring a warning). Moreover, in some aspects, the memory compatibility rules differ to the Candid sub-type check:
* Support of variable (MutBox) with type invariance.
* Types cannot be made optional (no insertion of Option).
* Same arity for function parameters and function return types.
* Records cannot introduce additional optional fields.
* Same arity for tuple types.
* Records and tuples are distinct.

### Garbage Collection

The garbage collection state needs to be persisted and retained across upgrades. This is because the GC may not yet be completed at the time of an upgrade, such that object forwarding is still in use. The partition table is stored as part of the GC state.

* The stable type table.

### Main Actor
On an upgrade, the main actor is recreated and existing stable variables are recovered from the persistent root. The remaining actor variables, the flexible fields as well as new stable variables, are (re)initialized. 

As a result, the GC can collect unreachable flexible objects of previous canister versions. Unused stable variables of former versions can also be reclaimed by the GC.

### No Static Heap
The static heap is abandoned and former static objects need to be allocated in the dynamic heap. This is because these objects may also need to survive upgrades and the persistent main memory cannot accommodate a growing static heap of a new program version in front of the existing dynamic heap. The incremental GC also operates on these objects, meaning that forwarding pointer resolution is also necessary for these objects. 

For memory and runtime efficiency, object pooling is implemented for compile-time-known constant objects (with side-effect-free initialization), i.e. those objects are already created on program initialization/upgrade in the dynamic heap and thereafter the reference to the corresponding prefabricated object is looked up whenever the constant value is needed at runtime.

The runtime system avoids any global Wasm variables for state that needs to be preserved on upgrades. Instead, such global runtime state is stored in the persistent metadata.

### Wasm Data Segments
Only passive Wasm data segments are used by the Motoko compiler and runtime system. In contrast to ordinary active data segments, passive segments can be explicitly loaded to a dynamic address.

This simplifies two aspects: 
* The generated Motoko code can contain arbitrarily large data segments (to the maximum that is supported by the IC). The segments can be loaded to the dynamic heap when needed.
* The IC can simply retain the main memory on an upgrade without needing to patch any active data segments of the new program version to the persistent main memory.

However, more specific handling is required for the Rust-implemented runtime system (RTS): The Rust-generated active data segment of the runtime system is changed to the passive mode and loaded to the expected static address on the program start (canister initialization and upgrade). The location and size of the RTS data segments is therefore limited to a defined reserve of 512 KB, see above. This is acceptable because the RTS only requires a controlled small amount of memory for its data segments, independent of the compiled Motoko program.

### Null Sentinel
As an optimization, the top-level `null` pointer is represented as a constant sentinel value pointing to the last unallocated Wasm page. This allows fast null tests without involving forwarding pointer resolution of potential non-null comparand pointers.

### Migration Path
When migrating from the old serialization-based stabilization to the new persistent heap, the old data is deserialized one last time from stable memory and then placed in the new persistent heap layout. Once operating on the persistent heap, the system should prevent downgrade attempts to the old serialization-based persistence. 

Assuming that the persistent memory layout needs to be changed in the future, the runtime system supports serialization and deserialization to and from stable memory in a defined data format using graph copy. 

### Graph Copy

The graph copy is an alternative persistence mechanism that will be only used in the rare situation when the persistent memory layout will be changed in the future. Arbitrarily large data can be serialized and deserialized beyond the instruction and working set limit of upgrades: Large data serialization and deserialization is split in multiple messages, running before and/or after the IC upgrade to migrate large heaps. Of course, other messages will be blocked during this process and only the canister owner or the canister controllers are permitted to initiate this process. 

Graph copying needs to be explicitly initiated before an upgrade to new Motoko version that is incompatible to the current enhanced orthogonal persistent layout. For large data, the graph copy needs to be manually completed after the actual upgrade.

```
dfx deploy CANISTER_ID
```

More detailed information and instructions on graph copy are contained in `design/GraphCopyStabilization.md`.

### Old Stable Memory
The old stable memory remains equally accessible as secondary (legacy) memory with the new support.

## Current Limitations
* Freeing old object fields: While new program versions can drop object fields, the runtime system should also delete the redundant fields of persistent objects of previous program versions. This could be realized during garbage collection when objects are copied. For this purpose, the runtime system may maintain a set of field hashes in use and consult this table during garbage collection. Another, probably too restrictive solution could be to disallow field removal (subtyping) on object upgrades during the memory compatibility check.
* The incremental GC only allows 64 GB. Transitioning to a dynamic partition table would be necessary to go beyond this limit. This is to be be implemented in a separate PR.
* Workaround for Rust needed to build PIC (position-independent code) libraries. Explicit use of `emscripten` via LLVM IR. 
* `ic-wasm` would need to be extended to Wasm64. The Wasm optimizations in `test/bench` are thus currently deactivated.
* The Wasm profiler (only used for the flamegraphs) is no longer applicable because the underlying `parity-wasm` crate is deprecated before Wasm Memory64 support. It also lacks full support of passive data segments. A re-implementation of the profiler would be needed.

## Related PRs

* IC support for enhanced orthogonal persistence: https://github.com/dfinity/ic/pull/143
* IC deterministic working set limit for 64-bit main memory: https://github.com/luc-blaeser/ic/pull/1
* Motoko base library 64-bit support: https://github.com/dfinity/motoko-base/pull/589

Underlying partial implementations:
* 32-bit enhanced orthogonal persistence without graph copy: https://github.com/dfinity/motoko/pull/4193
* 64-bit enhanced orthogonal persistence without graph copy: https://github.com/dfinity/motoko/pull/4225


design/GraphCopyStabilization.md
```diff
@@ -0,0 +1,129 @@
...
+* Perform the actual upgrade of the canister on the IC.
+* Detects that graph-copying is in use.
+* Clears the heap if enhanced orthogonal persistence is active.
+* Start the destabilization with a limited number of steps to fit into the upgrade message.
+* If destabilization cannot be completed, the canister does not start the GC and does not accept messages except step 3.
+
+3. Complete the explicit destabilization after the upgrade:
+
+```
+dfx canister call CANISTER_ID __motoko_destabilze_after_upgrade "()"
```
crusso, line 69:
```suggestion
dfx canister call CANISTER_ID __motoko_destabilze_after_upgrade "()"
```
```suggestion
dfx canister call CANISTER_ID __motoko_destabilize_after_upgrade "()"
```

design/GraphCopyStabilization.md
```diff
@@ -0,0 +1,129 @@
...
+* The Brooks forwarding pointer used by the incremental GC is omitted.
+* The pointers encode skewed stable memory offsets to the corresponding target objects.
+* References to the null objects are encoded by a defined null sentinel value.
+* `BigInt` are explicitly serialized in a defined portable little endian representation, without that the serialization or deserialization allocates temporary objects.
+The format is also versioned to allow future refinements of the graph copy algorithm.
+
+## Specific Aspects
+* Field hashes in objects are serialized in a blob. On deserialization, the hash blob is allocated in the dynamic heap. Same-typed objects that have been created by the same program version share the same hash blob.
+* Stable records can dynamically contain non-stable fields due to structural sub-typing. A dummy value can be serialized for such fields as a new program version can no longer access this field through the stable types.
+* For backwards compatibility, old Candid destabilzation is still supported when upgrading from a program that used older compiler version.
```
crusso, line 116:
```suggestion
* For backwards compatibility, old Candid destabilization is still supported when upgrading from a program that used older compiler version.
```

design/GraphCopyStabilization.md
```diff
@@ -0,0 +1,129 @@
...
+* Incremental GC: Serialization needs to consider Brooks forwarding pointers (not to be confused with the Cheney's forwarding information), while deserialization can deal with partitioned heap that can have internal fragmentation (free space at partition ends).
+* The partitioned heap prevents linear scanning of the heap, especially in the presence of large objects that can be placed at a higher partition than subsequently allocated normal-sized objects. For this reason, a scan stack is allocated in the main memory, remembering the deserialized objects that still need to be scanned. With this, the deserialization does not need to make any assumptions of the heap structure (e.g. monotonically increasing allocations, free space markers, empty heap on deserialization start etc.).
```
crusso, line 118: Do we need to worry about the scan stack overflowing at all? I assume not, but just checking..

src/ir_def/ir.ml
```diff
@@ -232,13 +232,17 @@ let full_flavor () : flavor = {
   has_poly_eq = true;
 }
 
+type actor_type = {
+  transient_actor_type: Type.typ;
+  stable_actor_type: Type.typ
```
crusso, line 237: I guess stable_actor_type is the record of initial stable field value (probably without mutability disinctions) - what  is transient_actor_type?

src/ir_def/ir.ml
```diff
@@ -232,13 +232,17 @@ let full_flavor () : flavor = {
...
+}
 
 (* Program *)
 
 type comp_unit =
   | LibU of dec list * exp
   | ProgU of dec list
-  | ActorU of arg list option * dec list * field list * system * Type.typ (* actor (class) *)
+  | ActorU of arg list option * dec list * field list * system * actor_type * exp (* actor (class) *)
```
crusso, line 245: Curious what `exp` will be for...

src/lowering/desugar.ml
```diff
@@ -500,7 +500,7 @@ and build_actor at ts self_id es obj_typ =
     let vs = fresh_vars "v" (List.map (fun f -> f.T.typ) fields) in
     blockE
       ((match call_system_func_opt "preupgrade" es obj_typ with
-        | Some call -> [ expD (primE (I.ICPerformGC) []); expD call]
+        | Some call -> [ expD call]
```
crusso, line 503: In the integrated PR, this might change the behaviour of the classic compiler too. I'm not sure why we inserted the GC call, but it was probably to free up space to do any copying needed in preupgrade - probably a bad idea anyway....

src/lowering/desugar.ml
```diff
@@ -513,9 +513,9 @@ and build_actor at ts self_id es obj_typ =
                ) fields vs)
             ty)) in
   let footprint_d, footprint_f = export_footprint self_id (with_stable_vars (fun e -> e)) in
-  I.(ActorE (footprint_d @ ds', footprint_f @ fs,
+  (ty, I.(ActorE (footprint_d @ ds', footprint_f @ fs,
      { meta;
-       preupgrade = with_stable_vars (fun e -> primE (I.ICStableWrite ty) [e]);
+       preupgrade = (primE (I.ICStableWrite ty) []);
```
crusso, line 518: Curious what happened to the argument `e` - I guess the semantics of ICStableWrite has changed a bit...

src/ir_def/ir.ml
crusso: Ah, ok, it's just the original actor type.

src/ir_def/ir.ml
```diff
@@ -72,7 +72,7 @@ and exp' =
   | FuncE of                                   (* function *)
       string * Type.func_sort * Type.control * typ_bind list * arg list * Type.typ list * exp
   | SelfCallE of Type.typ list * exp * exp * exp (* essentially ICCallPrim (FuncE shared…) *)
-  | ActorE of dec list * field list * system * Type.typ (* actor *)
+  | ActorE of dec list * field list * system * Type.typ * exp (* actor *)
```
crusso, line 75: Hmm, I wonder if a slightly cleaner refactoring might have been to put the additional type and exp fields in `ActorE` and `ActorU` into the `system` record. Don't bother now, it's just a note to self.

src/ir_def/ir.ml
crusso: Ok, exp coughs up the stable variable record after running the preupgrade system method.

crusso (commented):
I'm curious: what happens if you call __motoko_stabilize_before_upgrade() followed by __motoko_stabilize_after_upgrade(), without any upgrade in between? I guess it should just trap.

rts/motoko-rts/src/stabilization.rs
```diff
@@ -0,0 +1,74 @@
+//! Graph-copy-based stabilzation on upgrades, serializing the entire stable object graph into
+//! stable memory by using a defined long-term stable storage format.
+//!
+//! This is to support potentially radical changes of the persistent main memory layout, e.g.
+//! introducing a new GC or rearranging persistent metadata. This also relies on precise value
+//! tagging to allow more advanced changes that require value metadata, e.g. specializing arrays for
+//! small element types or even downgrading to 32-bit heap layouts (provided that the amount of live
+//! data fits into a 32-bit memory).
+//!
+//! A memory compatibility check similar to enhanced orthogonal persistence has to be performed.
```
crusso, line 10: I actually wish we did this for the legacy stable format too! But maybe in another life...

crusso: Also the old candid destabilization is at least somewhat defensive and will catch many errors...

rts/motoko-rts/src/stabilization.rs
```diff
@@ -0,0 +1,74 @@
...
+            ic0_stable64_write(position, &empty_chunk as *const u8 as u64, size);
+        }
+        position += size;
+    }
+}
+
+fn grant_stable_space(byte_size: u64) {
+    debug_assert!(byte_size < u64::MAX - PAGE_SIZE - 1);
+    let required_pages = (byte_size + PAGE_SIZE - 1) / PAGE_SIZE;
+    let available_pages = stable_mem::size();
```
crusso, line 63: I forget, are these the operations for growing physical or virtual stable memory. Will need to check.

rts/motoko-rts/src/stabilization/graph_copy.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    /// Determin whether the copy algorithm is completed,
+    /// i.e. a sufficient amount of copy increments has been invoked.
+    fn is_completed(&self) -> bool;
+
+    /// Complete the graph copy algorithm.
+    fn complete(&mut self);
+
+    /// Copy reachable objects in a time-bounded work step. with a synthetic time bound.
+    /// This allows to spread the incremtnal graph copy where the work is
```
crusso, line 38:
```suggestion
    /// This allows to spread the incremental graph copy where the work is
```

rts/motoko-rts/src/stabilization/graph_copy.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// while !copy_algorithm.is_completed() {
+    ///     copy_algorthm.copy_increment();
+    /// }
+    /// ```
+    fn start<M: Memory>(&mut self, mem: &mut M, root: S) {
+        self.evacuate(mem, root);
+    }
+
+    /// Determin whether the copy algorithm is completed,
+    /// i.e. a sufficient amount of copy increments has been invoked.
```
crusso, line 31:
```suggestion
    /// i.e. a sufficient amount of copy increments have been invoked.
```

rts/motoko-rts/src/stabilization/graph_copy.rs
crusso, line 1: Well, that's nice an simple!

rts/motoko-rts/src/stabilization.rs
crusso: The fix might be to call ic0_stable64_grow and and ic_stable64_size instead.

rts/motoko-rts/src/stabilization/layout/stable_array.rs
```diff
@@ -0,0 +1,130 @@
...
+            array_length: array.len() as u64,
+        }
+    }
+
+    unsafe fn serialize_dynamic_part(
+        stable_memory: &mut StableMemoryStream,
+        main_array: *mut Array,
+    ) {
+        for index in 0..main_array.len() {
+            let main_element = main_array.get(index);
```
crusso, line 39: So this loop has to complete in one increment --- is it possible that large array will cause serialization to fail? Should we maybe worry about  making this incremental too?

rts/motoko-rts/src/stabilization/layout/stable_array.rs
```diff
@@ -0,0 +1,130 @@
...
+        _main_memory: &mut M,
+        stable_memory: &StableMemoryAccess,
+        stable_object: StableValue,
+        target_array: *mut Array,
+    ) {
+        let stable_address = stable_object.payload_address();
+        let mut element_address =
+            stable_address + size_of::<StableArray>().to_bytes().as_usize() as u64;
+        for index in 0..(*target_array).len {
+            let element = stable_memory.read::<StableValue>(element_address);
```
crusso, line 84: Ditto here:this loop has to complete in one increment --- is it possible that a large array will cause deserialization to fail? Should we maybe worry about making this incremental too?

rts/motoko-rts/src/stabilization/layout/stable_bigint.rs
```diff
@@ -0,0 +1,202 @@
...
+#[repr(C)]
+pub struct StableBigInt {
+    is_negative: bool, // Sign bit.
+    number_of_bits: Bits, // Number of used bits in the payload.
+
+                       // Dynamically sized payload of the big integer stored in little endian format.
+                       // Zero padding to align to the next `u64`.
+}
+
+const USED_BITS_PER_ELEMENT: u32 = 28;
```
crusso, line 32: Is this 28 on both 64- and 32-bit?

rts/motoko-rts/src/stabilization/layout/stable_blob.rs
```diff
@@ -0,0 +1,92 @@
...
+            byte_length: main_object.len().as_usize() as u64,
+        }
+    }
+
+    unsafe fn serialize_dynamic_part(
+        stable_memory: &mut StableMemoryStream,
+        main_object: *mut Blob,
+    ) {
+        let byte_length = main_object.len().as_usize();
+        stable_memory.raw_write(main_object.payload_addr() as usize, byte_length);
```
crusso, line 49: Like array serialization, this raw_write is non-incremental and I guess could fail to fit in the increment...

rts/motoko-rts/src/stabilization/layout/stable_blob.rs
```diff
@@ -0,0 +1,92 @@
...
+        _main_memory: &mut M,
+        stable_memory: &StableMemoryAccess,
+        stable_object: StableValue,
+        target_blob: *mut Blob,
+    ) {
+        let stable_address = stable_object.payload_address();
+        let source_payload = stable_address + size_of::<StableBlob>().to_bytes().as_usize() as u64;
+        let target_payload = target_blob.payload_addr() as usize;
+        let blob_length = self.byte_length as usize;
+        stable_memory.raw_read(source_payload, target_payload, blob_length);
```
crusso, line 90: Ditto: like array deserialization, this raw_read is non-incremental and I guess could fail to fit in the increment...

rts/motoko-rts/src/stabilization/layout.rs
```diff
@@ -0,0 +1,404 @@
...
+//! Pointers are serialized as 64-bit skewed offsets in that space.
+//! The stable encoding of scalars is currently identical to the
+//! main memory encoding, requiring the precise scalar tag as metadata.
+//! If the main memory scalar representation changes in future, the
+//! values need to be explicitly translated to the expected stable format.
+//!
+//! Each object uses a `StableTag` as header and is followed by
+//! the object payload as outlined in the corresponding Rust structs.
+//! Some objects, such as `StableArray`, `StableObject`, `StableBlob`,
+//! and `StableBigNum` have a dynamic payload body in an addition to a static
```
crusso, line 14:
```suggestion
//! and `StableBigNum` have a dynamic payload body in addition to a static
```

rts/motoko-rts/src/stabilization/layout/stable_bits32.rs
crusso, line 1: The struct `StableBits32` (and this file) might actually be unused now - hangover from 32-bit initial implementation?

crusso: It it were used, would you need to align to 64-bit boundary?

rts/motoko-rts/src/stabilization/layout.rs
```diff
@@ -0,0 +1,404 @@
...
+mod stable_some;
+mod stable_variant;
+
+/// Different kinds of objects used in the stable format.
+/// Note: The tags are encoded as `StableTag` to be represented as `u64`.
+/// Native Rust enums would be encoded as `i32`, leading to unwanted alignment
+/// when directly serializing them as tags to the 64-bit stable format.
+#[repr(C)]
+#[derive(Clone, Copy, PartialEq)]
+pub enum StableObjectKind {
```
crusso, line 71: I wonder if you could use
```
[#repr(u64)] 
pub enum ...
  Array = 1, 
```
etc to avoid declaring u64 constants `STABLE_TAG_ARRAY` but perhaps that doesn't work for u64...

rts/motoko-rts/src/stabilization/layout.rs
```diff
@@ -0,0 +1,404 @@
...
+            TAG_OBJECT => StableObjectKind::Object,
+            TAG_BLOB => StableObjectKind::Blob,
+            TAG_BITS64 => StableObjectKind::Bits64,
+            TAG_REGION => StableObjectKind::Region,
+            TAG_VARIANT => StableObjectKind::Variant,
+            TAG_CONCAT => StableObjectKind::Concat,
+            TAG_BIGINT => StableObjectKind::BigInt,
+            TAG_OBJ_IND => StableObjectKind::ObjInd,
+            TAG_SOME => StableObjectKind::Some,
+            _ => unimplemented!("tag {tag}"),
```
crusso, line 154: Is this actually unimplemented with more case to consider or impossible?

rts/motoko-rts/src/stabilization/serialization/stable_memory_stream.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    fn raw_read(&mut self, data_address: usize, length: usize) {
+        debug_assert!(self.scan_address + length as u64 <= self.free_address);
+        unsafe {
+            ic0_stable64_read(data_address as u64, self.scan_address, length as u64);
+        }
+        self.scan_address += length as u64;
+    }
+
+    fn rewind(&mut self, length: usize) {
```
crusso, line 107: This might be unused?

rts/motoko-rts/src/stabilization/serialization/stable_memory_stream.rs
```diff
@@ -0,0 +1,149 @@
...
+
+pub trait ScanStream {
+    // Determines whether the stream has reached the end.
+    fn scan_completed(&self) -> bool;
+    // Read a value from the stream.
+    fn read<T>(&mut self) -> T;
+    // Read raw data from the stream.
+    fn raw_read(&mut self, data_address: usize, length: usize);
+    // Go back in the stream.
+    fn rewind(&mut self, length: usize);
```
crusso, line 37: This might be unused?

src/lowering/desugar.ml
crusso: Hmm, perhaps this dates from the original implementation, before we did streaming stablitization direct to stable memory, and wanted to free up space for a temporary, in-memory, candid blob. Ok to discard I think.

rts/motoko-rts/src/stabilization.rs
crusso: Maybe it isn't a bug.

I just ran this test, that allocates 3 stable memory pages and the count is preserved across upgrades.
```
//MOC-FLAG --stabilization-instruction-limit=10000

import Prim "mo:prim";

actor {

    if (Prim.stableMemorySize() == 0) {
      Prim.debugPrint("growing stable memory");
      ignore Prim.stableMemoryGrow(3);
    };
    assert Prim.stableMemorySize() == 3;
    Prim.debugPrint(debug_show {stableMemorySize = Prim.stableMemorySize()});
    stable var largeArray = Prim.Array_tabulate<Nat>(100_000, func(index) { index });

    public func check() : async () {
        Prim.debugPrint("Array of length " # debug_show (largeArray.size()));
        var index = 0;
        while (index < largeArray.size()) {
            assert (largeArray[index] == index);
            index += 1;
        };
    };

    system func preupgrade() {
        Prim.debugPrint("PRE-UPGRADE HOOK!");
    };

    system func postupgrade() {
        Prim.debugPrint("POST-UPGRADE HOOK!");
    };
};

//CALL ingress check "DIDL\x00\x00"
//CALL ingress __motoko_stabilize_before_upgrade "DIDL\x00\x00"
//CALL upgrade ""
//CALL ingress __motoko_destabilize_after_upgrade "DIDL\x00\x00"
//CALL ingress check "DIDL\x00\x00"
//CALL ingress __motoko_stabilize_before_upgrade "DIDL\x00\x00"
//CALL upgrade ""
//CALL ingress __motoko_destabilize_after_upgrade "DIDL\x00\x00"
//CALL ingress check "DIDL\x00\x00"

//SKIP run
//SKIP run-ir
//SKIP run-low
```
output:
```
[nix-shell:~/clean/motoko/test/run-drun]$ ../run.sh -d stabilize-large-array-esm.mo
WARNING: Could not run ic-ref-run, will skip running some tests
stabilize-large-array-esm: [tc] [comp] [comp-ref] [valid] [valid-ref] [drun-run]
--- stabilize-large-array-esm.drun-run (expected)
+++ stabilize-large-array-esm.drun-run (actual)
@@ -0,0 +1,22 @@
+debug.print: growing stable memory
+debug.print: {stableMemorySize = 3}
+debug.print: Array of length 100_000
+debug.print: PRE-UPGRADE HOOK!
+debug.print: {stableMemorySize = 3}
+debug.print: POST-UPGRADE HOOK!
+debug.print: Array of length 100_000
+debug.print: PRE-UPGRADE HOOK!
+debug.print: {stableMemorySize = 3}
+debug.print: POST-UPGRADE HOOK!
+debug.print: Array of length 100_000
```

Maybe its fine because you write the ESM page count to metadata before doing the actual graph copy.

crusso: (But I haven't verified that, it's my guess at why this still works)

crusso: Ok, turns out this is a real bug because stablemem_grow, unlike ic0_stable64_grow, caps growth to --max-stable-pages, and will prevent stabilization of larger heap data.

See #4540

rts/motoko-rts/src/stabilization.rs
```diff
@@ -0,0 +1,74 @@
...
+    if required_pages > available_pages {
+        let additional_pages = required_pages - available_pages;
+        debug_assert_ne!(additional_pages, u64::MAX);
+        let result = stable_mem::grow(additional_pages);
```
crusso, line 67: You want to use the physical, not logical grow here.

rts/motoko-rts/src/stabilization/graph_copy/limit.rs
```diff
@@ -0,0 +1,32 @@
...
+        let elapsed = current - self.start;
+        elapsed > self.limit
+    }
+
+    pub fn reset(&mut self) {
+        self.start = Self::instruction_counter();
+    }
+
+    fn instruction_counter() -> u64 {
+        unsafe { ic0_performance_counter(0) }
```
crusso, line 30: This might actually be quite expensive to call frequently (IIRC like 200 instructions per call?)

src/codegen/compile.ml
```diff
@@ -9351,8 +9510,371 @@ module FuncDec = struct
...
+    compile_comparison I64Op.LeU
+
+  let use_graph_destabilization env =
+    get_persistence_version env ^^
+    compile_eq_const StableMem.version_graph_copy_no_regions ^^
+    get_persistence_version env ^^
+    compile_eq_const StableMem.version_graph_copy_regions ^^
+    G.i (Binary (Wasm_exts.Values.I64 I64Op.Or))
+
+  let use_enhancecd_orthogonal_persistence env =
```
crusso, line 9830:
```suggestion
  let use_enhanced_orthogonal_persistence env =
```

src/codegen/compile.ml
```diff
@@ -5106,7 +5159,7 @@ module IC = struct
 
   let async_method_name = Type.(motoko_async_helper_fld.lab)
   let gc_trigger_method_name = Type.(motoko_gc_trigger_fld.lab)
-
+
```
crusso, line 5162:
```suggestion
 
```

src/codegen/compile.ml
```diff
@@ -8407,15 +8533,38 @@ module Persistence = struct
...
+    E.call_import env "rts" "is_graph_stabilization_started" ^^ Bool.from_rts_int32
+
+  let start_graph_stabilization env actor_type =
+    EnhancedOrthogonalPersistence.create_type_descriptor env actor_type ^^
+    E.call_import env "rts" "start_graph_stabilization"
+
+  let graph_stabilization_increment env =
+    E.call_import env "rts" "graph_stabilization_increment" ^^ Bool.from_rts_int32
+
+  let start_graph_destabilization env actor_type complete_initialization =
```
crusso, line 8558: `complete_initialization` is unused?

src/codegen/compile.ml
```diff
@@ -9351,8 +9510,371 @@ module FuncDec = struct
...
+    IC.system_call env "call_perform" ^^
+    G.i (Convert (Wasm_exts.Values.I64 I64Op.ExtendUI32)) ^^
+    E.then_trap_with env "Async stabilization increment call failed" 
+
+  let define_async_stabilization_reply_callback env =
+    Func.define_built_in env async_stabilization_reply_callback_name ["env", I32Type] [] (fun env ->
+      is_stabilization_completed env ^^
+      E.if0
+        begin
+          (* Sucessful completion of the async stabilzation sequence. *)
```
crusso, line 9573:
```suggestion
          (* Sucessful completion of the async stabilization sequence. *)
```

src/codegen/compile.ml
```diff
@@ -9351,8 +9510,371 @@ module FuncDec = struct
...
+    Func.define_built_in env async_destabilization_reply_callback_name ["env", I32Type] [] (fun env ->
+      get_destabilized_actor env ^^
+      compile_test I64Op.Eqz ^^
+      E.if0
+        begin
+          (* Trigger next async destabilization increment. *)
+          call_async_destabilization env
+        end
+        begin
+          (* Send static reply of sucessful async destabilzation sequence. *)
```
crusso, line 9692:
```suggestion
          (* Send static reply of sucessful async destabilization sequence. *)
```

src/codegen/compile.ml
```diff
@@ -9351,8 +9510,371 @@ module FuncDec = struct
...
+
+  let export_async_destabilization_method env actor_type =
+    let name = async_destabilization_method_name in
+    begin match E.mode env with
+    | Flags.ICMode | Flags.RefMode ->
+      Func.define_built_in env name [] [] (fun env ->
+        IC.assert_caller_self_or_controller env ^^
+        (* Stay in lifecycle state `InDestabilization` if not yet completed. *)
+        destabilization_increment env actor_type ^^
+        IC.static_nullary_reply env
```
crusso, line 9727: Because this calls IC.static_nullary_reply env, I think the asynchronous call will immediately complete, like oneway message, triggering the destablization to run in the background.

Maybe that's intentional but, if you wanted this to only respond once completed,  I think you only want to issue the reply in `complete_graph_destabilization env`.

design/GraphCopyStabilization.md
luc-blaeser: Thank you!

design/GraphCopyStabilization.md
luc-blaeser: Thanks!

design/GraphCopyStabilization.md
luc-blaeser: The scan stack grows in the heap. My assumption is that there is sufficient space in the heap. I can measure the maximum stack depth for some complex practical cases but I assume it is relatively low (maybe a few MB). Moreover, array slicing during the scanning (`scan_deserialized`) helps to reduce the stack depth for large arrays.

luc-blaeser: (TODO for @luc-blaeser : Measure the scan stack depth.)

src/lowering/desugar.ml
luc-blaeser: Good point: This has impact on the classical support in the integrated PR.
I also think that with the streaming Candid serialization, the GC call should not be necessary.

src/lowering/desugar.ml
luc-blaeser: Yes, I had to change the semantics of ICStableWrite. The stable actor record to persist can no longer be passed in as argument (`e`) because explicit incremental stabilization may start before the actual upgrade and already needs the to build the stable record (and run preupgrade) in advance. Therefore, the persistence logic builds the stable record when needed (inversion of control). `ICStableWrite` now means "complete the incremental stabilization" or "prepare EOP upgrade".

src/ir_def/ir.ml
luc-blaeser: I agree. I added a comment and made a temporary small code refactoring making it more obvious what `exp` stands for. We can then later combine it with `system`.

luc-blaeser:
> I'm curious: what happens if you call __motoko_stabilize_before_upgrade() followed by __motoko_stabilize_after_upgrade(), without any upgrade in between? I guess it should just trap.

I double checked, it traps: 
```
+ingress Completed: Reject: IC0503: Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped explicitly: RTS error: No destabilization needed
```

The same happens when `__motoko_destabilize_after_upgrade()` is applied on a running canister.

rts/motoko-rts/src/stabilization.rs
luc-blaeser: I agree, it has different compatibility semantics, and type-driven Candid destabilization is written in a way that it should not lead to corrupt memory state even if the `dfx` incompatibility warning is skipped.

crusso (commented):
virtual -> physical mem_size and grow function changes LGTM. Thanks!

design/GraphCopyStabilization.md
crusso: I think you are probably fine since your stack is not fixed size but linked segments, right?

crusso (commented):
Try
stabilize-large-blob.mo
```
//MOC-FLAG --stabilization-instruction-limit=10000 --max-stable-pages 65536

import Prim "mo:prim";

actor {

    let pages : Nat64 = 65536;
    if (Prim.stableMemorySize() == 0) {
      Prim.debugPrint("growing stable memory");
      ignore Prim.stableMemoryGrow(pages);
    };
    assert Prim.stableMemorySize() == pages;
    stable let blob = Prim.stableMemoryLoadBlob(0, Prim.nat64ToNat pages * 65536);

    public func check() : async () {
        Prim.debugPrint(debug_show (blob.size()))
    };

    system func preupgrade() {
        Prim.debugPrint("PRE-UPGRADE HOOK!");
    };

    system func postupgrade() {
        Prim.debugPrint("POST-UPGRADE HOOK!");
    };
};

//CALL ingress check "DIDL\x00\x00"
//CALL ingress __motoko_stabilize_before_upgrade "DIDL\x00\x00"
//CALL upgrade ""
//CALL ingress __motoko_destabilize_after_upgrade "DIDL\x00\x00"
//CALL ingress check "DIDL\x00\x00"
//CALL ingress __motoko_stabilize_before_upgrade "DIDL\x00\x00"
//CALL upgrade ""
//CALL ingress __motoko_destabilize_after_upgrade "DIDL\x00\x00"
//CALL ingress check "DIDL\x00\x00"

//SKIP run
//SKIP run-ir
//SKIP run-low
```
produces 
```
[nix-shell:~/clean/motoko/test/run-drun]$ ../run.sh -d stabilize-large-blob.mo
WARNING: Could not run ic-ref-run, will skip running some tests
stabilize-large-blob: [tc] [comp] [comp-ref] [valid] [valid-ref] [drun-run]
--- stabilize-large-blob.drun-run.ret (expected)
+++ stabilize-large-blob.drun-run.ret (actual)
@@ -0,0 +1 @@
+Return code 101
--- stabilize-large-blob.drun-run (expected)
+++ stabilize-large-blob.drun-run (actual)
@@ -0,0 +1,18 @@
+debug.print: growing stable memory
+debug.print: 4_294_967_296
+debug.print: PRE-UPGRADE HOOK!
+debug.print: POST-UPGRADE HOOK!
+debug.print: 4_294_967_296
+debug.print: PRE-UPGRADE HOOK!
+debug.print: POST-UPGRADE HOOK!
+thread 'main' panicked at rs/drun/src/lib.rs:394:5:
+Ingress message did not finish executing within 10000 batches, panicking
+note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Some tests failed:
stabilize-large-blob.mo

```
That seems ok to me - I think the drun failure is just drun crappiness.

rts/motoko-rts/src/stabilization.rs
luc-blaeser: Thanks a lot for having noticed this bug and all the analysis and the repro.
It should now be fixed, using physical memory size and grow.

rts/motoko-rts/src/stabilization/graph_copy.rs
```diff
@@ -0,0 +1,92 @@
...
+    fn is_completed(&self) -> bool;
+
+    /// Complete the graph copy algorithm.
+    fn complete(&mut self);
+
+    /// Copy reachable objects in a time-bounded work step. with a synthetic time bound.
```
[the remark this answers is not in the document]
luc-blaeser: Thanks!

rts/motoko-rts/src/stabilization/graph_copy.rs
luc-blaeser: Thanks!

rts/motoko-rts/src/stabilization/graph_copy.rs
luc-blaeser: Thank you.

rts/motoko-rts/src/stabilization/layout/stable_array.rs
luc-blaeser: I believe the only practically critical loop is the scanning of the array, as this triggers Cheney's algorithm to potentially evacuate all the objects referenced in the elements. Therefore, I had to slice that logic to become incremental. The price is however complexity of such self-designed co-routines.

The other loops may take long, but I believe, even for large practical cases, they should still fit in the quite generous DTS-supported instruction limits (40 billion instructions for stabilization increments that are normal update messages, and 200 billion instructions for the actual upgrade).

I will remeasure very large array cases to see if this is practical problem.

rts/motoko-rts/src/stabilization/layout/stable_array.rs
luc-blaeser: I will check, but I believe, even for very large arrays, it should be okay. The critical loop is the scanning which is incremental.

rts/motoko-rts/src/stabilization/layout/stable_blob.rs
luc-blaeser: I will check whether this is an issue in practice for large blobs. I tried to avoid the complexity of incremental loops, except for the necessary cases (which I have seen was the array scanning in Cheney's algorithm).

rts/motoko-rts/src/stabilization/layout/stable_bigint.rs
luc-blaeser: The encoding is still 32-bit in Tom's math library. The number consists of `mp_digit` that is `u32` even on 64-bit. I therefore also compute the elements of the stable bigints as `i32` and pad it to 64-bit.

rts/motoko-rts/src/stabilization/layout/stable_blob.rs
luc-blaeser: I will analyze very large blobs but believe it could fit in the rather generous instruction limit - different to the array scanning logic (triggering sub-sequent evacuation of potentially all elements).

rts/motoko-rts/src/stabilization/layout.rs
luc-blaeser: Thank you!

rts/motoko-rts/src/stabilization/graph_copy.rs
luc-blaeser: Yes, but this is only the skeleton, the details make the complexity...

rts/motoko-rts/src/stabilization/layout/stable_bits32.rs
luc-blaeser: Oh, yes, it is a leftover and would need to be 64-bit aligned. Thanks for noticing. I will remove it.

rts/motoko-rts/src/stabilization/layout.rs
luc-blaeser: Actually, it is a good idea. I refactored it accordingly. Thanks for the suggestion. The only remaining problem is that the conversion from `u64` to enum still requires a custom, quite cumbersome, mapping function using constants. (It seems that there is no easy approach, maybe some libraries, but it is not worth taking that complexity/risk).

rts/motoko-rts/src/stabilization/layout.rs
luc-blaeser: Oh, it is indeed impossible. I need to improve this, by using `unreachable`

rts/motoko-rts/src/stabilization/serialization/stable_memory_stream.rs
luc-blaeser: Thanks. Indeed, this is redundant. I have removed it.

rts/motoko-rts/src/stabilization/serialization/stable_memory_stream.rs
luc-blaeser: Thanks. Indeed, this is redundant. I have removed it.

src/codegen/compile.ml
```diff
@@ -8013,7 +8099,7 @@ invariant type constructors in a single pass.
 end (* Serialization *)
 
 (* OldStabilization as migration code: 
-  Deserializing a last time from explicit stable memory into the stable heap:
+  Deserializing a last time from Candid-serailized stable objects into the stable heap:
```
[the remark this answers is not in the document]
luc-blaeser: Thanks!

rts/motoko-rts/src/stabilization.rs
luc-blaeser: Indeed. I fixed this bug. Thank a lot for having identified this.

src/codegen/compile.ml
luc-blaeser: Thanks. I fixed it (including the other occurrences).

rts/motoko-rts/src/stabilization/graph_copy/limit.rs
luc-blaeser: Right, I will measure the overhead incurred. Maybe we need a more coarse-grained counter.

src/codegen/compile.ml
luc-blaeser: Thanks. This is a leftover that can be deleted. I moved the completion logic to  `destabilization_increment`.

src/codegen/compile.ml
luc-blaeser: Thank you!

src/codegen/compile.ml
luc-blaeser: Thanks!

src/codegen/compile.ml
luc-blaeser: Thanks for checking. The logic is indeed quite complicated: This is only an internal async helper function which gets called by another outer async function `call_async_destabilization` in an async loop/chain style. The `call_async_destabilization` that is initiated by the external `__motoko_destabilize_after_upgrade`. The outer async function should only return when the entire destabilization has finished.

luc-blaeser:
> virtual -> physical mem_size and grow function changes LGTM. Thanks!

Thank you too for having found the bug!

design/GraphCopyStabilization.md
luc-blaeser: It is only limited by heap free space...

rts/motoko-rts/src/stabilization/layout/stable_array.rs
luc-blaeser: I tried a 100_000_000-long array and it takes around 8.6 billion instructions to serialize. It is not cheap but probably practically feasible.

rts/motoko-rts/src/stabilization/layout/stable_array.rs
luc-blaeser: For an array of length 100_000_000, I measured 6 billion instructions for deserialization. Probably, this is fine for the 40 billion instruction update message in which the increment runs.

crusso (approved):
Recent changes look good to me.

rts/motoko-rts/src/stabilization/layout/stable_bigint.rs
luc-blaeser: **Update**: I now use the 64-bit version of Tom's math library for EOP. This uses 60bits per 64-bit word, thus being  more space-efficient.

rts/motoko-rts/src/stabilization/layout/stable_blob.rs
luc-blaeser: I measured the serialization of a 2 GB blob: It takes about 2.7 billion instructions. I believe this is still practically small enough to fit into a stabilzation increment.

rts/motoko-rts/src/stabilization/graph_copy/limit.rs
luc-blaeser: Indeed, this can be quite expensive, like 2 billion instructions for an array of length of 10 million elements. I will try to implement an improvement based on synthetic estimated counting with occasional calibration to the real costs.

rts/motoko-rts/src/stabilization.rs
```diff
@@ -0,0 +1,82 @@
...
+    pub fn moc_stabilization_instruction_limit() -> u64;
+    fn ic0_performance_counter(number: u32) -> u64;
+}
+
+// Dummy value used for non-stable objects that are potentially reachable from
+// stable variable because of structural subtyping or `Any`-subtyping.
+// Must be a non-skewed value such that the GC also ignores this value.
+const DUMMY_VALUE: StableValue = StableValue::from_raw(0);
+
+fn clear_stable_memory(start: u64, length: u64) {
```
luc-blaeser, line 46: Note to myself: This needs to be incremental when called at the end of deserialization, as it may be very large length (multiple GC) and will otherwise exceed the instruction limit.

rts/motoko-rts/src/stabilization/deserialization.rs
```diff
@@ -0,0 +1,186 @@
...
+/// while !deserialization.is_completed() {
+///     deserialization.copy_increment();
+/// }
+/// ```
+impl Deserialization {
+    /// Start the deserialization, followed by a series of copy increments.
+    pub fn start<M: Memory>(mem: &mut M, stable_start: u64, stable_size: u64) -> Deserialization {
+        let from_space = StableMemoryAccess::open(stable_start, stable_size);
+        let scan_stack = unsafe { ScanStack::new(mem) };
+        let limit = InstructionLimit::new(unsafe { moc_stabilization_instruction_limit() });
```
luc-blaeser, line 60: Note to my self: This limit needs to be reset to the lower update limit when subsequent destabilization increments are needed after the actual upgrade.

rts/motoko-rts/src/stabilization/graph_copy/limit.rs
luc-blaeser: I optimzed it, by calling `ic0_performance_counter` less frequently, only when a certain amount of objects have been copied **or** a certain amount of memory has been processed. 

Together with some more tuning (chunk size of memory clearing), I obtain the following improvements:

### Upgrade Instructions with Incremental Graph Copy

Benchmark      | Before Optimization        | After Optimization         | Difference
---------------|----------------------------|----------------------------|-----------------
Double Sharing | 690_000                    | 260_000                    | -62%
Linked List    | 86_000_000                 | 71_000_000                 | -17%     
Array List     | 5_800_000_000              | 3_400_000_000              | -41%
Small Objects  | 52_000_000_000             | 36_000_000_000             | -31%
Small Texts    | 20_000_000_000             | 15_000_000_000             | -25%
Big Numbers    | 4_100_000_000              | 3_100_000_000              | -24%
Blobs          | 940_000_000                | 950_000_000                | +1%

(The "Blobs" are little bit slower due to another recent change, where I made the stable memory clearing incremental, see below.)

rts/motoko-rts/src/stabilization.rs
luc-blaeser: This is now fixed. I tested it for cases of more than 10 GB of graph-stabilized memory.

rts/motoko-rts/src/stabilization/deserialization.rs
luc-blaeser: This is now fixed.

PR #4481, opened 2024-04-03
experiment: measure cost of repeated subtype tests (to detect regressions in EOP)

crusso:
Looks like #4193 introduces a serious regression in candid subtyping (as suspected). 

See https://github.com/dfinity/motoko/pull/4481/commits/00285ba840936466f545f458ac3fc30eb97e2ba3 for the diff.

I suspect the problem is that each candid subtype test allocates a new blob for the global type table and type index vector, instead of re-using a pre-allocated one.


test/bench/candid-subtype-cost.mo
```diff
@@ -0,0 +1,147 @@
...
+    uninstall_code : shared { canister_id : canister_id } -> async ();
+    update_settings : shared {
+        canister_id : Principal;
+        settings : canister_settings;
+      } -> async ();
+  }
+
+}
+
+//CALL ingress go 0x4449444C0000
```
crusso, line 147:
```suggestion
//CALL ingress go 0x4449444C0000

```

test/bench/candid-subtype-cost.mo
```diff
@@ -0,0 +1,147 @@
+// measure cost of repeated candid subtype checks
+import { Array_tabulate; performanceCounter; debugPrint; rts_heap_size } = "mo:⛔";
+
+actor {
+
+
+// This is a generated Motoko binding.
+// Please use `import service "ic:canister_id"` instead to call canisters on the IC if possible.
+
```
crusso, line 9:
```suggestion
```

PR #4488, opened 2024-04-05
Enhanced Orthogonal Persistence (Complete Integration)

luc-blaeser:
### PR Stack
Enhanced orthogonal persistence support is structured in four PRs to ease review:
1. https://github.com/dfinity/motoko/pull/4488 **<-- this PR**
2. https://github.com/dfinity/motoko/pull/4475
3. https://github.com/dfinity/motoko/pull/4225
4. https://github.com/dfinity/motoko/pull/4193

# Enhanced Orthogonal Persistence (Complete Integration)

* **64-bit heap**: Extending the main memory to 64-bit for large-scaled persistence.

## Advantages
* **Performance**: New program versions directly resume from the existing main memory and have access to the memory-compatible data.
* **Scalability**: The upgrade mechanism scales with larger heaps and in contrast to serialization, does not hit IC instruction limits.

## Design
The enhanced orthogonal persistence is based on the following main properties:
* Supporting 64-bit main memory on the IC.

### Integration with Classical Persistence
Enhanced orthogonal persistence is offered for **beta testing** via the compiler flag `--enhanced-orthogonal-persistence`.
Classical persistence with 32-bit main memory and Candid stabilization currently remains the default mode.
See `design/PersistenceModes.md` for more information.

### Memory Layout
* Lower 4MB: Rust call stack.

### Persistent Metadata

### Compatibility Check
Upgrades are only permitted if the new program version is compatible with the old version, such that the runtime system guarantees a compatible memory structure.

Compatible changes for immutable types are largely analogous to the allowed Motoko subtype relation, e.g.
* Removing object fields.
* Adding variant fields.
* `Nat` to `Int`.

This compatibility check serves as an additional safety measure on top of the DFX Candid subtype check that can be bypassed by users (when ignoring a warning). Moreover, in some aspects, the memory compatibility rules differ to the Candid sub-type check:
* Top-level actor fields (`stable` fields) can change mutability (`let` to `var` and vice-versa).
* Support of variable (MutBox) with type invariance.
* Types cannot be made optional (no insertion of Option).
* Same arity for function parameters and function return types (no removed optional parameters, no additional optional results).
* Records cannot introduce additional optional fields.
* Same arity for tuple types (no insertion of optional items).
* Records and tuples are distinct.

### Garbage Collection

The garbage collection state needs to be persisted and retained across upgrades. This is because the GC may not yet be completed at the time of an upgrade, such that object forwarding is still in use. The heap partition structure is described by a linked list of partition tables that is reachable from the GC state.

* The stable type table.

### Main Actor
On an upgrade, the main actor is recreated and existing stable variables are recovered from the persistent root. The remaining actor variables, the flexible fields as well as new stable variables, are (re)initialized. 

As a result, the GC can collect unreachable flexible objects of previous canister versions. Unused stable variables of former versions can also be reclaimed by the GC.

### No Static Heap
The static heap is abandoned and former static objects need to be allocated in the dynamic heap. This is because these objects may also need to survive upgrades and the persistent main memory cannot accommodate a growing static heap of a new program version in front of the existing dynamic heap. The incremental GC also operates on these objects, meaning that forwarding pointer resolution is also necessary for these objects. 

For memory and runtime efficiency, object pooling is implemented for compile-time-known constant objects (with side-effect-free initialization), i.e. those objects are already created on program initialization/upgrade in the dynamic heap and thereafter the reference to the corresponding prefabricated object is looked up whenever the constant value is needed at runtime.

The runtime system avoids any global Wasm variables for state that needs to be preserved on upgrades. Instead, such global runtime state is stored in the persistent metadata.

### Wasm Data Segments
Only passive Wasm data segments are used by the Motoko compiler and runtime system. In contrast to ordinary active data segments, passive segments can be explicitly loaded to a dynamic address.

This simplifies two aspects: 
* The generated Motoko code can contain arbitrarily large data segments (to the maximum that is supported by the IC). The segments can be loaded to the dynamic heap when needed.
* The IC can simply retain the main memory on an upgrade without needing to patch any active data segments of the new program version to the persistent main memory.

However, more specific handling is required for the Rust-implemented runtime system (RTS): The Rust-generated active data segment of the runtime system is changed to the passive mode and loaded to the expected static address on the program start (canister initialization and upgrade). The location and size of the RTS data segments is therefore limited to a defined reserve of 512 KB, see above. This is acceptable because the RTS only requires a controlled small amount of memory for its data segments, independent of the compiled Motoko program.

### Null Sentinel
As an optimization, the top-level `null` pointer is represented as a constant sentinel value pointing to the last unallocated Wasm page. This allows fast null tests without involving forwarding pointer resolution of potential non-null comparand pointers.

### Memory Capacity
The canister has no upfront knowledge of the maximum allocatable Wasm main memory in 64-bit address space, as there is no IC API call to query the main memory limit. This limit may also be increased in future IC releases.

Therefore, a mechanism is implemented to deal with an unknown and dynamically increasable main memory capacity offered by the IC. This is needed in two cases:

* GC reserve (strict): The runtime system ensures sufficient free space to allow garbage collection at all times, even if the heap is full. For this purpose, the runtime system already pre-allocates the reserve, to be sure that the reserve is available despite the unknown capacity. As an optimization, this pre-allocation is skipped when the memory demand including the GC reserve is below a guaranteed minimum Wasm memory limit of the IC, e.g. 4GB or 6GB.
* GC scheduling (heuristic): The GC schedules at high frequency when memory is becoming scarce. For this purpose, the GC maintains an assumption of the minimum memory limit and probes the supposed limit when the heap size approaches this limit. If the allocation succeeds, the assumed limit is increased. Otherwise, the critical high-frequency GC scheduling rate is activated.

In both cases, the runtime system tries to reduce Wasm memory allocations as much as possible, i.e. not pre-allocating memory for small heap sizes, and not probing an allocation in certain memory ranges by assuming that the IC only offers main memory of a certain granularity, e.g. multiples of 2GB. To save instructions, the critical GC scheduling is only activated when reaching the actual memory limit. Moreover, the mechanism can handle an increased memory capacity at runtime, e.g. when the IC is upgraded to a new release with a higher memory limit.

### Migration Path
When migrating from the old serialization-based stabilization to the new persistent heap, the old data is deserialized one last time from stable memory and then placed in the new persistent heap layout. Once operating on the persistent heap, the system should prevent downgrade attempts to the old serialization-based persistence. 

Assuming that the persistent memory layout needs to be changed in the future, the runtime system supports serialization and deserialization to and from stable memory in a defined data format using graph copy. 

### Graph Copy

The graph copy is an alternative persistence mechanism that will be only used in the rare situation when the persistent memory layout will be changed in the future. Arbitrarily large data can be serialized and deserialized beyond the instruction and working set limit of upgrades: Large data serialization and deserialization is split in multiple messages, running before and/or after the IC upgrade to migrate large heaps. Of course, other messages will be blocked during this process and only the canister owner or the canister controllers are permitted to initiate this process. 

Graph copying needs to be explicitly initiated before an upgrade to new Motoko version that is incompatible to the current enhanced orthogonal persistent layout. For large data, the graph copy needs to be manually completed after the actual upgrade.

```
dfx deploy CANISTER_ID
```

More detailed information and instructions on graph copy are contained in `design/GraphCopyStabilization.md`.

### Old Stable Memory
The old stable memory remains equally accessible as secondary (legacy) memory with the new support.

## Current Limitations
* The memory footprint of a program increases with 64 bit as the word size for scalars and pointers are doubled. In turn, in some cases, boxing can be avoided due to larger word size which again reduces memory demand.
* Freeing old object fields: While new program versions can drop object fields, the runtime system should also delete the redundant fields of persistent objects of previous program versions. This could be realized during garbage collection when objects are copied. For this purpose, the runtime system may maintain a set of field hashes in use and consult this table during garbage collection. Another, probably too restrictive solution could be to disallow field removal (subtyping) on object upgrades during the memory compatibility check.
* The Wasm profiler (only used for the flamegraphs) is no longer applicable because the underlying `parity-wasm` crate is deprecated before Wasm Memory64 support. It also lacks full support of passive data segments. A re-implementation of the profiler would be needed.


rts/motoko-rts/src/gc/copying.rs
crusso, line 1: Not new - reviewed in #4563

rts/motoko-rts/src/gc/generational.rs
crusso, line 1: Not new - diff reviewed in https://github.com/dfinity/motoko/pull/4563

rts/motoko-rts/src/gc/generational/mark_stack.rs
crusso, line 1: Not new - diff reviewed in https://github.com/dfinity/motoko/Not new - diff reviewed in https://github.com/dfinity/motoko/pull/4563pull/4563

rts/motoko-rts/src/gc/generational/sanity_checks.rs
crusso, line 1: Not new - diff reviewed in https://github.com/dfinity/motoko/pull/4563

rts/motoko-rts/src/gc/generational/write_barrier.rs
crusso, line 1: Not new - diff reviewed in https://github.com/dfinity/motoko/pull/4563

src/codegen/compile_classical.ml
crusso, line 1: GOT HERE: still need to review all .ml code (everything else done) 07062024

rts/motoko-rts/src/leb128.rs
```diff
@@ -81,7 +93,13 @@ pub unsafe fn sleb128_decode_checked(buf: *mut Buf) -> Option<isize> {
 
         // Overflow check ported from Wasm reference implementation:
         // https://github.com/WebAssembly/spec/blob/f9770eb75117cac0c878feaa5eaf4a4d9dda61f5/interpreter/binary/decode.ml#L89-L98
-        if shift == 63 && (byte & 0b0111_1111 != 0 && byte & 0b0111_1111 != 0b0111_1111) {
+        let overflow = if uses_enhanced_orthogonal_persistence!() {
```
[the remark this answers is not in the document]
luc-blaeser: Good idea. I will adjust this.

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -1,66 +1,90 @@
...
 }
 
-#[no_mangle]
-pub unsafe extern "C" fn get_max_live_size() -> Bytes<usize> {
-    crate::gc::incremental::get_max_live_size()
+/// Page allocation. Ensures that the memory up to, but excluding, the given pointer is allocated.
+/// Ensure a memory reserve of at least one Wasm page depending on the canister state.
+/// `memory_reserve`: A memory reserve in bytes ensured during update and initialization calls.
+//  For use by queries and upgrade calls. The reserve may vary depending on the GC and the phase of the GC.
+unsafe fn grow_memory(ptr: u64, memory_reserve: usize) {
```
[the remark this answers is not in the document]
luc-blaeser: Yes, this is used in both 32-bit and 64-bit. In 32-bit it is an optimization (w.r.t. to the IC cost model) to avoid  overflows in `alloc_words` for linear memory (classical GCs only). I thought to keep that logic, although it looks asymmetric.

```
let old_hp = getHP() as u64;
let new_hp = old_hp + delta;
```

rts/motoko-rts/src/types.rs
```diff
@@ -265,35 +278,62 @@ impl Value {
...
+                    || self.forward().forward().get_ptr() == self.forward().get_ptr()
+            );
+        }
     }
 
     /// Check whether the object's forwarding pointer refers to a different location.
     pub unsafe fn is_forwarded(self) -> bool {
-        self.check_forwarding_pointer();
-        self.forward().get_ptr() != self.get_ptr()
+        if is_incremental_gc!() {
```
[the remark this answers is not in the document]
luc-blaeser: It is only used for sanity checking with non-incremental GCs. I believe it is also optimized away for classical GCs in sanity checks.

test/run/float-ops.mo
```diff
@@ -29,8 +29,9 @@ assert (not isNegative(positiveNaN));
 assert (isNegative(negativeNaN));
 assert (isNegative(negate(positiveNaN)));
 assert (not isNegative(negate(negativeNaN)));
-Prim.debugPrint(debug_show(positiveNaN));
-Prim.debugPrint(debug_show(negativeNaN));
+// Difference between 32-bit and 64-bit version. 64-bit is even more accurate (with correct sign bit on `nan`).
```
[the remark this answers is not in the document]
luc-blaeser: Good idea. I will try how easy it is to align them between 32-bit and 64-bit.

src/codegen/compile_classical.ml
luc-blaeser: Thank you very much!

rts/motoko-rts/src/memory.rs
```diff
@@ -2,8 +2,29 @@
...
+use motoko_rts_macros::enhanced_orthogonal_persistence;
 use motoko_rts_macros::ic_mem_fn;
 
+#[cfg(feature = "ic")]
+use crate::constants::MB;
+
+#[enhanced_orthogonal_persistence]
+use crate::constants::GB;
+
+// TODO: Redesign for 64-bit support by using a dynamic partition list.
```
crusso, line 15: Addressed in #4556, right?

rts/motoko-rts/src/stream.rs
```diff
@@ -0,0 +1,211 @@
...
+use crate::bigint::{check, mp_get_u32, mp_isneg, mp_iszero};
+use crate::mem_utils::memcpy_bytes;
+use crate::memory::{alloc_blob, Memory};
+use crate::rts_trap_with;
+use crate::tommath_bindings::{mp_div_2d, mp_int};
+use crate::types::{size_of, Blob, Bytes, Stream, Value, TAG_BLOB};
+
+use motoko_rts_macros::ic_mem_fn;
+
+const MAX_STREAM_SIZE: Bytes<usize> = Bytes((1 << 30) - 1);
```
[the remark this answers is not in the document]
luc-blaeser: **Update**: I added a static assertion for 32-bit and added some comments, also in `types.rs` for the `Stream` declaration.

rts/motoko-rts/src/memory.rs
luc-blaeser: Yes, this PR would lift the restriction.

src/codegen/compile_enhanced.ml
```diff
@@ -5938,7 +5938,23 @@ module RTS_Exports = struct
...
+        Bool.to_rts_int32
+      )
+    ) in
+    E.add_export env (nr {
+      name = Lib.Utf8.decode "keep_memory_reserve";
+      edesc = nr (FuncExport (nr keep_memory_reserve_fi))
+    });
 
     let rts_trap_fi = E.add_fun env "rts_trap" (
       (* `libc` stil uses 32-bit length parameter for `rts_trap` *)
```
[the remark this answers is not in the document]
luc-blaeser: Thank you!

src/codegen/compile_classical.ml
crusso, line 1: Reviewed as diff to compile.ml on master. 
Most changes are trivial changes in offset from Int32 to Int64, plus a modification for obtaining the stable actor record.

LGTM!

src/linking/linkModule.ml
```diff
@@ -801,25 +833,29 @@ let link (em1 : extended_module) libname (em2 : extended_module) =
...
-  let max_rts_stack_size = 4 * 1024 * 1024 in
-  let max_rts_data_segment_size = 512 * 1024 in
-  (if (Int32.to_int new_heap_start) > max_rts_stack_size + max_rts_data_segment_size then
-    (raise (TooLargeDataSegments (Printf.sprintf "The Wasm data segment size exceeds the supported maxmimum of %nMB." max_rts_data_segment_size)))
-  else
-    ()
-  );
+  if uses_memory64 em1.module_ then
+  begin
+    (* The RTS data segments must fit below 4.5MB according to the persistent heap layout.
```
crusso, line 838: I was worried that 512KB might not be enough if we decide to support unicode and need more static data. But QuickJS seems to do this with just 40KB of additional data so perhaps this is enough.

src/prelude/internals.mo
```diff
@@ -444,8 +446,13 @@ func @install_actor_helper(
         (#reinstall, (prim "cast" : (actor {}) -> Principal) actor1)
       };
       case (#upgrade actor2) {
+        let wasm_memory_persistence = if enhanced_orthogonal_persistence { 
+          ?(#Keep)
```
crusso, line 450: What will happen here if the old actor doesn't support eop. Will it fail or just degrade to candid destabilization?

I'm thinking about the distributed map example. If we upgrade the main node from non-eop to eop, and it tries to upgrade it previously deployed non-eop actor class instances, will they fail to upgrade?

src/rts/rts.ml
```diff
@@ -1,8 +1,8 @@
 (*
-This source file loads the RTS (`mo-rts.wasm` etc.) via the environment
+This source file loads the RTS Wasm files via the environment
 variables. This is for local development (e.g. inside `nix-shell`). The nix
 build of `moc` will statically replace this file with one that just embeds
-`mo-rts.wasm` etc. as a static string, to produce a fully self-contained `moc`
+RTS Wasm file as a static string, to produce a fully self-contained `moc`
```
crusso, line 5:
```suggestion
RTS Wasm file as a static string, to produce a fully self-contained `moc`
```
```suggestion
RTS Wasm files as static strings, to produce a fully self-contained `moc`
```

test/run-drun/actor-class-mgmt.mo
```diff
@@ -107,37 +98,17 @@ actor a {
...
       let c5 = await
-        (system Cs.C)(#upgrade c4)(5, null);
-      await useIncrementalStabilization(c5).__motoko_destabilize_after_upgrade();
-      assert ({args = 5; upgrades = 2} == (await c5.observe()));
+        (system Cs.C)(#reinstall c4)(5, null);
+      assert ({args = 5; upgrades = 0} == (await c5.observe()));
       assert (c5 == c4);
 
-      let c6 = await
-        (system Cs.C)(#upgrade_with_persistence { wasm_memory_persistence = #Keep ; canister = c5 })(6, null);
```
crusso, line 125: Is there no more need for #upgrade_with_persistence? Is just subsumed by #upgrade?

crusso (approved):
I've reached the summit!

All looks good to me, but obviously I could have missed something.

In future, we could even consider adding support for a programmatic graph-copy upgrade or fall back to one if the direct path fails due to version mismatch...

src/pipeline/pipeline.ml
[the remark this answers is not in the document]
luc-blaeser: Yes, I considered such but one thought that I had is that the memory management is also quite different beyond the GC (no static heap, fix metadata, specific roots etc.). Therefore, I thought it could maybe be useful to have an explicit flag for EOP, but I assume the difference is small and we can still adjust it.

test/run-drun/actor-class-mgmt.mo
luc-blaeser: This is only test restructuring because of the integration: I removed this part from this test file as it runs on both classical persistence and EOP. There is another test `actor-class-mgmt-enhanced.mo` that specifically runs for EOP and contains the upgrade with the explicit persistence mode.

src/prelude/internals.mo
luc-blaeser: This will fall back to upgrade by Candid stabilization, meaning the old actor is automatically migrated to EOP.
I added a test case `map-mixed-upgrades.drun` that tests the various scenarios of keeping and dropping main memory (including graph copy).

src/rts/rts.ml
luc-blaeser: Thank you!

luc-blaeser:
> I've reached the summit!
> 
> All looks good to me, but obviously I could have missed something.
> 
> In future, we could even consider adding support for a programmatic graph-copy upgrade or fall back to one if the direct path fails due to version mismatch...

Thank you  very much for this review marathon! I really appreciate your work.

I was thinking of the automatic fallback of graph copy: I guess it is possible but probably with a little bit of extra complexity: As one would detect the version mismatch only in the post-upgrade, one would need to move the stabilization logic to the new version, understanding the old version(s), whereas, currently, each compiler version only implements its own version stabilization and destabilization.

PR #4494, opened 2024-04-10
Docs: Motoko docs revisions & restructure

jessiemongeon1:


doc/md/writing-motoko/message-restrictions.md
crusso, line 1: This file is empty. Was it supposed to contain stuff from the original async-data?

doc/md/actors-async.md
```diff
@@ -1,306 +0,0 @@
...
-``` motoko no-repl
-  public shared func bump() : async Nat {
-    await inc();
-    await read();
-  };
-```
-
-does *not* have the same semantics and allows another client of the actor to interfere with its operation: each `await` suspends execution, allowing an interloper to change the state of the actor. By design, the explicit `await`s make the potential points of interference clear to the reader.
-
-## Traps and Commit Points
```
crusso, line 133: It's important not to lose this section and traps and commit points. But I guess you are in the middle of moving stuff around

crusso: Ah, I see it's in traps.md now. Nice!

crusso:
Be aware that lots of the file textually include examples from doc/md/examples. The examples are extracted so that we can type-check them with moc using doc/md/examples/check.sh to avoid shipping broken examples.

If you are in a nix-shell, you can generate a local doc preview using docusaurus using

```
nix-shell
cd doc
make preview
```
that might help find broken links.

The base library documention in doc/md/base is autogenerated from the base sources available in nix-shell. Don't edit those since your edits will get clobbered next time we do `make base`.

doc/md/writing-motoko/message-restrictions.md
jessiemongeon1: Yep, it'll contain that info.

jessiemongeon1:
- All docs revised to use the Dev Docs style guide.
- Docs reorganized into different categories 
- Sidebar configuration set for each doc, so in the portal repo we can set 'autogenerate' for the Motoko docs.
- Preview of new sidebar categories:

Screenshot 2024-04-18 at 12 55 44 PM

To-do in portal repo:
- Add redirects for renamed/moved files
- Update any links to old/moved URL
- Update sidebar.js to use autogenerate
- Update top nav bar link for 'Motoko' button
-

doc/md/canister-maintenance/cycles.md
```diff
@@ -1,10 +1,12 @@
 # Managing cycles
```
crusso: I looks like this section wasn't revised yet to use the new <system> capability when adding (and perhaps accepting) cycles. We'll need to revise that eventually though the examples should still compile with a warning.

doc/md/motoko-grammar.md
crusso, line 1: I don't think we should lose this file, instead move it to reference/motoko-grammar.md

doc/md/canister-maintenance/cycles.md
```diff
@@ -0,0 +1,111 @@
...
+
+:::
+
+## The `ExperimentalCycles` Library
+
+The `ExperimentalCycles` library provides imperative operations for observing an actor’s current balance of cycles, transferring cycles and observing refunds.
+
+The library provides the following operations:
+
+``` motoko no-repl
```
crusso, line 30: Will these long one line comments render nicel? If not, I'd go back to just showing the function signatures, and then using bullets underneath to explain each one.

doc/md/canister-maintenance/cycles.md
```diff
@@ -0,0 +1,111 @@
...
+func balance() : (amount : Nat) // Returns the actor’s current balance of cycles as `amount`. Function `balance()` is stateful and may return different values after calls to `accept(n)`, calling a function after `add`ing cycles, or resuming from `await` which reflects a refund.
+
+func available() : (amount : Nat) // Returns the currently available `amount` of cycles. This is the amount received from the current caller, minus the cumulative amount `accept`ed so far by this call. On exit from the current shared function or `async` expression via `return` or `throw` any remaining available amount is automatically refunded to the caller.
```
crusso, line 33:
```suggestion
func available<system>() : (amount : Nat) // Returns the currently available `amount` of cycles. This is the amount received from the current caller, minus the cumulative amount `accept`ed so far by this call. On exit from the current shared function or `async` expression via `return` or `throw` any remaining available amount is automatically refunded to the caller. Requires `system` capability.
```

doc/md/canister-maintenance/cycles.md
```diff
@@ -0,0 +1,111 @@
...
+
+func accept(amount : Nat) : (accepted : Nat) // Transfers `amount` from `available()` to `balance()`. It returns the amount actually transferred, which may be less than requested, for example, if less is available, or if canister balance limits are reached.
```
crusso, line 35:
```suggestion
func accept<system>(amount : Nat) : (accepted : Nat) // Transfers `amount` from `available()` to `balance()`. It returns the amount actually transferred, which may be less than requested, for example, if less is available, or if canister balance limits are reached. Requires `system` capability.
```

doc/md/canister-maintenance/cycles.md
```diff
@@ -0,0 +1,111 @@
...
+
+func add(amount : Nat) : () // Indicates the additional amount of cycles to be transferred in the next remote call, i.e. evaluation of a shared function call or `async` expression. Upon the call, but not before, the total amount of units `add`ed since the last call is deducted from `balance()`. If this total exceeds `balance()`, the caller traps, aborting the call.
```
crusso, line 37:
```suggestion
func add<system>(amount : Nat) : () // Indicates the additional amount of cycles to be transferred in the next remote call, i.e. evaluation of a shared function call or `async` expression. Upon the call, but not before, the total amount of units `add`ed since the last call is deducted from `balance()`. If this total exceeds `balance()`, the caller traps, aborting the call. Requires `system` capability.
```

doc/md/canister-maintenance/upgrades.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Overview
+
+One key feature of ICP is its ability to persist canister state using WebAssembly memory and globals rather than a traditional database. This means that the entire state of a canister is restored before and saved after each message, without explicit user instruction. This automatic and user-transparent preservation of state is called orthogonal persistence.
+
+Orthogonal persistence allows a canister to save and restore data to dedicated **stable memory** that, unlike ordinary canister memory, is retained across an upgrade and allows a canister to transfer data in bulk across upgrades.
+
+Motoko provides high-level support for preserving state that leverages stable memory. This feature, called **stable storage**, is designed to accommodate changes to both the application data and the Motoko compiler used to produce the application code.
+
+Utilizing stable storage depends the developer anticipating and indicating the data to retain after an upgrade. Depending on the application, the data you decide to persist might be some, all, or none of a given actor’s state.
```
crusso, line 15: I think the shortened overview is actually less clear than the original and conflates orthogonal persistence with stable memory, which is misleading.
In CS orthogonal persistence refers to  completely transparent preservation of state without any user intervention/special coding. This is what the IC provides to canister between messages - the state of the canister is transparently restored before the message and saved after without the user having to read/write stuff to a database as in traditional 3-tier (frontend/backend/DB) system. Instead, all data is just stored in memory. That's the OP bit.

Upgrades actually break the illusion of OP because the user has to explicitly arrange for the state of the old canister to be saved from the old version and then restored in the new version of the canister.  In Motoko, they can do this with reasonable convenience by explicitly declaring stable variables, or less conveniently by explicitly using stable memory or regions. But as soon as explicit measures are required, it's not really orthogonal persistence.

doc/md/canister-maintenance/compatibility.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+An upgrade from `v2` to `v3`'s stable types requires consuming an `Int` as a `Nat`, which is a **type error**.
+
+``` motoko no-repl file=../examples/count-v3.most
+```
+
+## Dual interface evolution
+
+An upgrade is safe provided that the Candid interface evolves to a subtype and the stable interface evolves to a compatible one, such as a variable to supertype or a new type.
```
crusso, line 77:
```suggestion
An upgrade is safe provided that the Candid interface evolves to a subtype and the stable interface evolves to a compatible one: a stable variable must either be newly declared, or re-declared at a super type of its old type.
```

doc/md/compatibility.md
```diff
@@ -1,253 +0,0 @@
...
-E.g. the upgrade from `v2` to `v3` fails this check:
-```
-> moc --stable-compatible v2.most v3.most
-(unknown location): Compatibility error [M0170], stable variable state of previous type
-  var Int
-cannot be consumed at new type
-  var Nat
-```
-
-## Examples in the wild
```
crusso, line 190: The example seems to have been lost in the move but its actually a common case and deserves mention somewhere. Users are constantly trying to add fields to records on upgrade and that's just not sound (because the upgrade would have to insert fields, making up values for them out of thin air).

doc/md/canister-maintenance/upgrades.md
```diff
@@ -0,0 +1,154 @@
...
+
+:::
+
+:::note
+
+The stable-compatible relation is quite conservative. In the future, it may be relaxed to accommodate a change in field mutability and/or abandoning fields from `<stab-sig1>` but with a warning.
+
+:::
+
+## Preupgrade and postupgrade system methods
```
crusso, line 92: I would maybe move this back just after the Section on ##Typing since its really about how to persist data that isn't inherently stable. Then the rest of the section doc is about upgrade safety and how to check it.

doc/md/writing-motoko/modules-and-imports.md
```diff
@@ -1,23 +1,29 @@
...
-import Result "mo:base/Result";
+import Debug "mo:base/Debug";
+Debug.print("hello world");
 ```
 
-Notice that the import declaration includes the `mo:` prefix to identify the module as a Motoko module and that the declaration does not include the `.mo` file type extension.
+This example illustrates how to import Motoko code—indicated by using the `mo:` prefix to identify the module as a Motoko module. The declaration does not include the `.mo` file type extension. Then, it uses the `base/` base library path and the module name `Debug`.
 
-Above example uses an identifier pattern to import modules wholesale, but you can also selectively import a subset of symbols from a module by resorting to the object pattern syntax:
+You can also selectively import a subset of symbols from a module by resorting to the object pattern syntax:
```
crusso, line 26:
```suggestion
You can also selectively import a subset of named values from a module by using the object pattern syntax:
```

doc/md/writing-motoko/modules-and-imports.md
```diff
@@ -1,23 +1,29 @@
...
+
+This section provides examples of different scenarios for using the `module` and `import` keywords.
 
-## Importing from the Motoko base library
+## Importing from the base library
 
-One of the most common import scenarios is one that you see illustrated in the examples in this guide, in the Motoko projects in the examples repository, and in the tutorials involves importing modules from the Motoko base library. Importing modules from the base library enables you to re-use the values, functions and types defined in those modules rather than writing similar ones from scratch.
+You can find source code and documentation for Motoko base modules in the [motoko-base](https://github.com/dfinity/motoko-base) open source repository. There are instructions in the repository for generating a local copy of the current documentation for the Motoko base library.
```
crusso, line 15: Shouldn't we first point them the base library documentation on the portal (or here in doc/md/base)?

doc/md/writing-motoko/modules-and-imports.md
```diff
@@ -1,23 +1,29 @@
+---
+sidebar_position: 16
+---
+
 # Modules and imports
```
crusso, line 5: I wonder if we should actually be referring to base more consistently as a package (a collection of modules), not a library, or at least indicate that we use the terms library and package interchangeably.

doc/md/base-intro.md
```diff
@@ -1,24 +0,0 @@
-# Using Motoko base modules
-
-The design of Motoko strives to minimize built-in types and operations. Instead of built-in types, Motoko provides a base library of modules to handle many kinds of common operations and make the language feel complete. This base library is still evolving with modules that support core features and all of the base library APIs are subject to change over time to varying degrees. You should note, in particular, that the size and number of modules and functions included in the base library is likely to increase dramatically and updates to the base library modules might introduce breaking changes that require you to update your programs to remain compatible.
```
crusso, line 3: Ok, so this content now lives in modules-and-imports.md.
Make sense.

doc/md/motoko-grammar.md
jessiemongeon1: This content is now in the Motoko style guide. Should it have its own file, or is adding it into the style guide ok?

doc/md/getting-started/dev-env.md
```diff
@@ -0,0 +1,48 @@
...
+
+Before you start developing Motoko, verify the following:
+
+- [x] You have an internet connection and access to a shell terminal on your local macOS or Linux computer.
+
+- [x] You have a command line interface (CLI) window open. This window is also referred to as the 'terminal' window.
+
+- [x] You have downloaded and installed the IC SDK package as described in the [installing the IC SDK](/docs/current/developer-docs/getting-started/install/) page.
+
+- [x] You have a code editor installed. The [VSCode IDE](https://code.visualstudio.com/download) is a popular choice.
```
rvanasa, line 44:
```suggestion
- [x] You have a code editor installed. The [VS Code IDE](https://code.visualstudio.com/download) (with the [Motoko extension](https://marketplace.visualstudio.com/items?itemName=dfinity-foundation.vscode-motoko)) is a popular choice.
```

doc/md/getting-started/dev-env.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Container environments
+
+Developers may want to setup a containerized environment for Motoko and other ICP-related development. Container environments are especially useful for Windows-based systems, since `dfx` is not natively supported on Windows.
+
+Learn more about [developer containers](/docs/current/developer-docs/developer-tools/ide/dev-containers) and [Docker containers](/docs/current/developer-docs/developer-tools/ide/dev-containers#using-docker-directly) for Motoko development.
+
+## Motoko playground
+
+Motoko playground is a browser-based developer environment that allows for temporary deployment and testing of canister smart contracts. Motoko playground can also be run locally, or it can be utilized through the `dfx deploy --playground` command via the CLI.
```
rvanasa, line 28:
```suggestion
[Motoko playground](https://play.motoko.org/) is a browser-based developer environment that allows for temporary deployment and testing of canister smart contracts. Motoko playground can also be utilized through the `dfx deploy --playground` command via the CLI.
```

CC @chenyan-dfinity

doc/md/getting-started/dev-env.md
```diff
@@ -0,0 +1,48 @@
+---
+sidebar_position: 3
+---
+
+
+# Developer environment
+
+## Overview
+
+To develop and deploy Motoko canister smart contracts, you will need a developer environment that contains the Motoko compiler and base library. It is recommended to use the IC SDK, which includes Motoko, along with `dfx`, a command-line tool used to create, build, and deploy canisters on ICP.
```
rvanasa, line 10:
```suggestion
To develop and deploy Motoko canister smart contracts, you will need a developer environment that contains the Motoko compiler and base library. It is recommended to use the [IC SDK](https://github.com/dfinity/sdk#readme), which includes Motoko, along with `dfx`, a command-line tool used to create, build, and deploy canisters on ICP.
```

doc/md/getting-started/dev-env.md
```diff
@@ -0,0 +1,48 @@
...
+
+Developer environments come in several types and formats, making developing flexible and accessible.
+
+## Cloud environments
+
+Cloud developer environments, such as Gitpod or GitHub Codespaces, are browser-based developer environments that can be configured with the necessary tools to develop Motoko code.
```
rvanasa, line 16:
```suggestion
[Gitpod](https://www.gitpod.io/) and [GitHub Codespaces](https://github.com/features/codespaces) are browser-based development environments that can be used to build, test, and run Motoko smart contracts. 

Here are some starter projects for online Motoko canister development:

* [ICP Hello World Motoko](https://github.com/dfinity/icp-hello-world-motoko#readme)
* [Vite + React + Motoko](https://github.com/rvanasa/vite-react-motoko#readme)
```

rvanasa (commented):
Found a few places where we could include links, adjust phrasing, etc. on the "Developer environment" page. Added some loose suggestions. LGTM in general!

doc/md/canister-maintenance/cycles.md
jessiemongeon1: addressed in https://github.com/dfinity/motoko/pull/4512

doc/md/canister-maintenance/upgrades.md
jessiemongeon1: adressed in https://github.com/dfinity/motoko/pull/4512

doc/md/compatibility.md
jessiemongeon1: adressed in https://github.com/dfinity/motoko/pull/4512

jessiemongeon1:
Working out of https://github.com/dfinity/motoko/pull/4512

PR #4497, opened 2024-04-11
experiment: remove regression in idl_sub 

crusso:
Builds on #4193 

Builds the absolute global type table just once on entry to (candid) deserializer, not once on every call to idl_sub, avoiding 
repeated allocation.

The global type table data is stored in additional "Registers" (i.e. globals) to avoid passing them down the stack, but must be reset between GC runs as the type descriptors can be moved by GC.

Compare perf output
Master/EOP unpooled: [00285ba](https://github.com/dfinity/motoko/pull/4497/commits/00285ba840936466f545f458ac3fc30eb97e2ba3)
EOP unpooled/EOP pooled: [34937d9](https://github.com/dfinity/motoko/pull/4497/commits/34937d965f27e8b53a1aa6bf0019df847bd71472)
EOP pooled/this branch: [52c5f7f](https://github.com/dfinity/motoko/pull/4497/commits/52c5f7f3355ff577d8ff5d03285bd2e57a07a168)

TODO:
[ ] Should we remove the new benchmark? It's totally artificial...

### DIFF
--- a/rts/motoko-rts/src/idl.rs
+++ b/rts/motoko-rts/src/idl.rs
@@ -1208,23 +1208,33 @@ unsafe extern "C" fn idl_sub_buf_init(rel_buf: *mut u32, typtbl_size1: u32, typt
 }
 
 #[ic_mem_fn]
-unsafe fn idl_sub<M: Memory>(
+unsafe fn idl_alloc_typtbl<M: Memory>(
     mem: &mut M,
+    candid_data: Value,
+    type_offsets: Value,
+    typtbl_out: *mut *mut *mut u8,
+    typtbl_end_out: *mut *mut u8,
+    typtbl_size_out: *mut u32,
+) {
+    let mut type_descriptor = TypeDescriptor::new(candid_data, type_offsets);
+    *typtbl_out = type_descriptor.build_type_table(mem);
+    *typtbl_end_out = type_descriptor.type_table_end();
+    *typtbl_size_out = type_descriptor.type_count() as u32;
+}
+
+#[no_mangle]
+unsafe extern "C" fn idl_sub(
     rel_buf: *mut u32, // a buffer with at least 2 * typtbl_size1 * typtbl_size2 bits
     typtbl1: *mut *mut u8,
+    typtbl2: *mut *mut u8,
     typtbl_end1: *mut u8,
+    typtbl_end2: *mut u8,
     typtbl_size1: u32,
-    candid_data2: Value,
-    type_offsets2: Value,
+    typtbl_size2: u32,
     t1: i32,
     t2: i32,
 ) -> bool {
     debug_assert!(rel_buf != (0 as *mut u32));
-
-    let mut type_descriptor2 = TypeDescriptor::new(candid_data2, type_offsets2);
-    let typtbl2 = type_descriptor2.build_type_table(mem);
-    let typtbl_end2 = type_descriptor2.type_table_end();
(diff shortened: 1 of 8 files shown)


rts/motoko-rts/src/idl.rs
```diff
@@ -1208,24 +1208,34 @@ unsafe extern "C" fn idl_sub_buf_init(rel_buf: *mut u32, typtbl_size1: u32, typt
...
     rel_buf: *mut u32, // a buffer with at least 2 * typtbl_size1 * typtbl_size2 bits
     typtbl1: *mut *mut u8,
+    typtbl2: *mut *mut u8,
     typtbl_end1: *mut u8,
+    typtbl_end2: *mut u8,
     typtbl_size1: u32,
-    candid_data2: Value,
-    type_offsets2: Value,
+    typtbl_size2: u32,
     t1: i32,
```
luc-blaeser, line 1234: Maybe we could add null checks for the type tables too and reset the registers to null after the deserialization call. Just as safety guard that future compiler backend extensions do not use the pointers in the registers across GC increments.

src/codegen/compile.ml
```diff
@@ -7759,7 +7783,10 @@ module Serialization = struct
                 get_ref_buf ^^ Registers.set_ref_buf env ^^
                 get_typtbl_ptr ^^ load_unskewed_ptr ^^ Registers.set_typtbl env ^^
                 get_maintyps_ptr ^^ load_unskewed_ptr ^^ Registers.set_typtbl_end env ^^
-                get_typtbl_size_ptr ^^ load_unskewed_ptr ^^ Registers.set_typtbl_size env
+                get_typtbl_size_ptr ^^ load_unskewed_ptr ^^ Registers.set_typtbl_size env ^^
+                get_global_typtbl_ptr ^^ load_unskewed_ptr ^^ Registers.set_global_typtbl env ^^
+                get_global_typtbl_end_ptr ^^ load_unskewed_ptr ^^ Registers.set_global_typtbl_end env ^^
+                get_global_typtbl_size_ptr ^^ load_unskewed_ptr ^^ Registers.set_global_typtbl_size env
```
luc-blaeser, line 7789: Maybe, we could reset the pointers to `null` after the deserialization.

test/run-drun/max-stack-variant.mo
```diff
@@ -2,7 +2,7 @@
 import { errorMessage; debugPrint; } = "mo:⛔";
 
 actor {
-    let expectedMinimumSize = 26_000;
+    let expectedMinimumSize = 31_000;
```
luc-blaeser, line 5: I just noticed that the version before this PR already supported 31_000. So, there seems to be another issue that the test does not scale to 62_000 with doubled stack size in EOP.

luc-blaeser (approved):
Great PR! Thanks a lot for the improvement.

test/run-drun/max-stack-variant.mo
luc-blaeser: I believe this is the Wasm stack overflow. So, probably, the limit 31_000 is also to be expected for this PR. (I changed the limit back to 31_000 like in `master`.)

rts/motoko-rts/src/idl.rs
crusso: Yeah, that would be sensible I guess

src/codegen/compile.ml
crusso: Makes sense

PR #4500, opened 2024-04-15
feat: `motoko-san` contributions by Serokell

int-index:
RFP-6: The second iteration of Motoko-san

High-level outline of new features:
* Support more Motoko syntax: return values, patter matching, 
  polymorphic private functions, loop labels
* Support build-in types: arrays, tuples, option, variants, records
  (immutable), `Nat`, `Text`
* Extend annotation syntax: quantifiers, loop invariants, `Prim.Ret`
  bind for return value, function post-conditions
* Add showcases: `test/viper/{reverse.mo,todo_record.mo}`

For more details see the diff of the `src/viper/README.md`


aterga:
Could you add instructions here on how to get the current version of the tool working with the motoko-san extension, please?

src/viper/pretty.ml
```diff
@@ -206,13 +207,19 @@ and pp_stmt' ppf = function
       fprintf ppf "@[%a@]" (pp_print_list ~pp_sep:comma pp_exp) args
     in
     fprintf ppf ")"
-  | LabelS (_, _) -> failwith "LabelS?"
+  | LabelS lbl ->
+      fprintf ppf "@[label %s@]" lbl.it
+  | GotoS lbl ->
```
aterga, line 212: I see that you are generating goto statements which look redundant. What' your main motivation for adding them to the translation?

int-index: `goto` is used to translate return statements, consider e.g.

```
  public func abs(n : Int) : async Int {
    if (n < 0) {
      return (0 - n);
    };
    return n;
  };
```

Here we have early exit in the `if` branch, which I decided to encode with a `goto` to a label at the end of the function body. Generated Viper code:

```
method abs($Self: Ref, n: Int)
     returns ($Res: Int)
    requires $Perm($Self)
    requires $Inv($Self)
    ensures $Perm($Self)
    ensures $Inv($Self)
    { 
      if ((n < 0))
         { 
           $Res := (0 - n)
           goto $Ret 
         }
      $Res := n
      goto $Ret
      label $Ret 
    }
```

crusso: Hmm, if Viper doesn't support early exit via return then this seems sensible. The other alternative is to disallow early return from the Motoko fragment (returns are optional and a method can just exit with a final value).

crusso: @aterga is this document https://github.com/viperproject/silver/blob/master/documentation/syntax/sil-syntax.txt current? It hasn't been touched in a decade and I (perhaps incorrectly) recall having to read the scala parser to figure out what the grammar of .vpr files actually is...

aterga: I don't this that document is current. 

Viper does have support for arbitrary goto expressions, so even loops could theoretically be encoded via goto, I think. I recall that it's even possible to annotate Viper gotos with invariants (exactly for supporting loops).

Using gotos to model early return statements should work, but I don't have experience with that. I think no goto invariants would be needed in this case, as no new looping would be introduced to the control flow.

As Claudio says, I'd also recommend not spending the time implementing generic early return statements as that may be involved.

Feel free to rewrite the Motoko examples you wirk with so they do not need to return early.

int-index: > I'd also recommend not spending the time implementing generic early return statements as that may be involved.

They seem to work as implemented. What am I missing?

src/viper/prelude.ml
```diff
@@ -0,0 +1,16 @@
...
+  axiom $all_diff { forall a: Array, i: Int :: {$loc(a, i)} $first($loc(a, i)) == a && $second($loc(a, i)) == i }
+  axiom $size_nonneg { forall a: Array :: $size(a) >= 0 }
+}
+define $array_acc(a, t, p) forall j: Int :: 0 <= j && j < $size(a) ==> acc($loc(a, j).t, p)
+define $array_untouched(a, t) forall j: Int :: 0 <= j && j < $size(a) ==> $loc(a, j).t == old($loc(a, j).t)
+field $int: Int
+field $bool: Bool
+field $ref: Ref
+field $array: Array
+/* END PRELUDE */|prelude}
```
crusso, line 16:
```suggestion
/* END PRELUDE */|prelude}

```

src/viper/prep.ml
```diff
@@ -0,0 +1,145 @@
...
+let prep_unit (u : comp_unit) : comp_unit =
+  let { imports; body } = u.it in
+  match body.it with
+  | ActorU(id_opt, decs) ->
+    let decs' = mono_dec_fields decs in
+    (* let _ = List.map (fun g -> print_endline (string_of_mono_goal g)) goals in *)
+    let body' = ActorU(id_opt, decs') in
+    (* let _ = List.map (fun d -> print_endline (Wasm.Sexpr.to_string 80 (Arrange.dec_field d))) decs' in *)
+    { u with it = {imports; body = { body with it = body' } } }
+  | _ -> u
```
crusso, line 145:
```suggestion
  | _ -> u
  
```

src/viper/prep.ml
```diff
@@ -0,0 +1,145 @@
...
+  String.concat "$" (g.mg_id :: List.map (fun t ->
+    match T.normalize t with
+    | T.Prim T.Int  -> "Int"
+    | T.Prim T.Nat  -> "Nat"
+    | T.Prim T.Bool -> "Bool"
+    | _ -> unsupported Source.no_region (Mo_types.Arrange_type.typ t)) g.mg_typs)
+
+let mono_calls_visitor (stk : mono_goal Stack.t) : visitor =
+  { visit_exp = (function
+    | {it = CallE({it = VarE v; at = v_at; note = v_note},inst,e); at; note} ->
```
crusso, line 33: Curious: how would you handle first-class uses of a polymorphic functions? Create a record of monomorphic ones?

test/fail/verification-asserts.mo
```diff
@@ -5,3 +5,4 @@ assert:return true;
 
 // `invariant` is a keyword in verification mode only
 assert:invariant true;
+assert:loop:invariant true;
```
crusso, line 8:
```suggestion
assert:loop:invariant true;

```

test/viper/loop-invariant.mo
```diff
@@ -0,0 +1,15 @@
...
+    var j : Int = 0;
+    while (i < n) {
+      assert:loop:invariant (i * 2 == j);
+      assert:loop:invariant (j - i == i);
+      i := i + 1;
+      j := j + 2;
+    };
+    return j;
+  };
+}
```
crusso, line 15:
```suggestion
}

```

src/viper/prep.ml
int-index: Yes, in case of a higher-rank usage that would be a plausible strategy. A record of monomorphic functions could indeed represent a finite set of instantiations of a polymorphic function.

However, a first-class use is not necessarily higher-rank. E.g. in `map(f, xs)` it wouldn't be necessary to turn `f` into a record of functions because all elements are of a single type. So we will end up with e.g. `map$Int(f$Int, xs)` if the elements are integers.

Actual higher-rank usage is rare in practice and I'm not sure if we should invest effort into supporting it.

src/mo_def/syntax.ml
```diff
@@ -186,6 +186,7 @@ and exp' =
   | LabelE of id * typ * exp                   (* label *)
   | BreakE of id * exp                         (* break *)
   | RetE of exp                                (* return *)
+  | ResVarE                                    (* name of result *)
```
ggreif, line 189: If I remember correctly, Motoko allows to name values of result tuples (unary too?). I.e.
``` Motoko
func foo() : (first : Int, second : Char) { ... }
```
or such. I doubt however that these names are available in the AST. If they are maintained, one could refer to them from invariants, etc. and the `ResVarE` could be made redundant.

src/viper/README.md
```diff
@@ -185,42 +286,118 @@ Below, we summarize the language features that _Motoko-san_ currently supports.
...
+        return a;   // conditional early exit
+      };
+      return a;     // normal exit
+      ```
+
+    * `while` loops and loop invariants
+
+      ```motoko
+      while (cond) {
+          assert:loop:invariant (e1);
```
ggreif, line 370: Parens necessary here?

src/mo_def/syntax.ml
ggreif: Surprise! It appears and is named:
``` Scheme
$ rlwrap moc -dp
Motoko compiler (source 0.11.1-112-g1afaf35e3)
> func foo() : (first : Int) = (8);
(Prog
  (LetD
    (VarP foo)
    (FuncE
      ???
      Local
      foo
      (TupP)
      (NamedT first (PathT (IdH Int)))
      =
      (LitE (PreLit 8 Nat))
    )
  )
)
let foo : () -> Int = <func>
```

int-index: It's not in scope in the function body though, so you couldn't mention it in an `assert`

src/mo_frontend/parser.mly
```diff
@@ -674,6 +674,8 @@ exp_un(B) :
     { RetE(TupE([]) @? at $sloc) @? at $sloc }
   | RETURN e=exp(ob)
     { RetE(e) @? at $sloc }
+  | VAR COLON RETURN
```
crusso, line 677: Is it possible to move this to `assertions.mly`, to avoid polluting the vanilla Motoko grammar?

src/mo_frontend/typing.ml
```diff
@@ -1115,7 +1117,7 @@ and infer_exp' f env exp : T.typ =
   if not env.pre then begin
     assert (T.normalize t' <> T.Pre);
     let e = A.infer_effect_exp exp in
```
crusso, line 1119: I'm surprised omitting the normalization doesn't break the type checker. I'm curious why you do this. Is it avoid losing type names to better map to Viper's nominal type system?

src/mo_frontend/typing.ml
```diff
@@ -1139,6 +1141,11 @@ and infer_exp'' env exp : T.typ =
     | None ->
       error env id.at "M0057" "unbound variable %s" id.it
     )
+  | ResVarE ->
+    (match env.rets with
```
crusso, line 1145: Maybe fail if not in viper mode?

crusso: Or better if not ( in viper mode and within an assertion)?

src/mo_frontend/typing.ml
```diff
@@ -55,9 +55,10 @@ type env =
     check_unused : bool;
     used_identifiers : S.t ref;
     unused_warnings : unused_warnings ref;
+    viper_mode : bool;
   }
 
-let env_of_scope msgs scope =
+let env_of_scope ?(viper_mode=false) msgs scope =
```
crusso, line 61: I wonder if we could, instead of additing this optional argument a bunch of places, just add a field to scope and use that appropriately, to avoid changing so many function signatures.

src/prelude/prim.mo
```diff
@@ -478,3 +478,13 @@ func regionStoreBlob(r : Region, offset : Nat64, val :  Blob) : () =
 let call_raw = @call_raw;
 
 func performanceCounter(counter : Nat32) : Nat64 = (prim "performanceCounter" : (Nat32) -> Nat64) counter;
+
+// predicates for motoko-san
+
+func forall<T>(f: T -> Bool): Bool {
```
crusso, line 484: I wonder what happens if these are called from compiled code rather that in viper mode. I expect the compiler will just crash complaining about an unknown Prim. I think we can live with this for now.

Note to self: possible fixes:
1) conditionally extend prim.mo in prelude.ml, the way we add timer support; or 
2) add a special deprecation error that triggers only in viper mode (as  for M0199 in https://github.com/dfinity/motoko/pull/4575)

src/mo_frontend/typing.ml
```diff
@@ -2921,12 +2928,12 @@ and infer_dec_valdecs env dec : Scope.t =
 
 (* Programs *)
 
-let infer_prog scope pkg_opt async_cap prog : (T.typ * Scope.t) Diag.result =
+let infer_prog ?(viper_mode=false) scope pkg_opt async_cap prog : (T.typ * Scope.t) Diag.result =
```
crusso, line 2931: It would be nice (but not essential) to avoid the `?(viper_mode=false)` argument (here and below) by passing an extra field in scope instead.  
`

src/pipeline/pipeline.ml
```diff
@@ -180,10 +180,10 @@ let async_cap_of_prog prog =
      else
        Async_cap.initial_cap()
 
-let infer_prog pkg_opt senv async_cap prog : (Type.typ * Scope.scope) Diag.result =
+let infer_prog ?(viper_mode=false) pkg_opt senv async_cap prog : (Type.typ * Scope.scope) Diag.result =
```
crusso, line 183: I would be nice to get rid of the ?(viper_mode=false) arguments perhaps by adding a field to senv

src/mo_def/syntax.ml
ggreif: Right, IIRC we pondered the "final" syntax of Motoko-san to have postconditions written _before_ the function body (and after the return declaration), so it could be in scope there. Please ignore my comment and stay with the big kludge of `assert:*` that I came up with. I am not proud of it 🥹

src/mo_frontend/typing.ml
ggreif: Maybe it is just a leftover from early development stages?

src/viper/README.md
```diff
@@ -124,8 +125,9 @@ After modifying the code and recompiling `moc`, don't forget to test the changes
 [nix-shell:motoko]$ make -C test/viper
 ```
 
-Each test case consists of a (formally specified) Motoko source file, say, `$TEST` (e.g., `invariant.mo`) and the expected test results, represented via a triplet of files:
+Each test case consists of a (formally specified) Motoko source file, say, `$TEST` (e.g., `invariant.mo`) and the expected test results, represented via a set of files:
 * `test/viper/ok/$TEST.vpr.ok` — what the Motoko compiler is expected to generate; this should be a Viper program.
+* `test/viper/ok/$TEST.vpr.stderr.ok` — diagnostic messages (warnings) emitted by the Motoko compiler. No file if stderr is empty.
```
ggreif, line 130: I suggest to make these `.vpr-stderr.ok` or better `.vpr-diag.ok`

src/viper/README.md
```diff
@@ -185,42 +286,118 @@ Below, we summarize the language features that _Motoko-san_ currently supports.
...
+        }
+        ```
 
         Supporting `let y = do { let y = 1 : Int; y + y };` is _simple_.
 
     * Runtime assertions: `assert i <= MAX;`
 
-    * Assignments (to local variables and actor fields): `x := x + 1`
-    
+    * Assignments (to local variables, actor fields, and array elements):
```
ggreif, line 343: record `var` fields too?

src/viper/prep.ml
```diff
@@ -0,0 +1,146 @@
...
+module Arrange = Mo_def.Arrange
+
+(*
+  Monomorphisation goal. For example, a function call f<Int,Bool>(a,b,c)
+  yields a mono_goal of {id="f", typs=[Int,Bool]}
+*)
+type mono_goal = { mg_id : string; mg_typs : T.typ list; }
+
+let compare_mono_goal (g1 : mono_goal) (g2 : mono_goal) =
+  match String.compare g1.mg_id g2.mg_id with
```
ggreif, line 17: why not

src/viper/prep.ml
```diff
@@ -0,0 +1,146 @@
...
+
+type dec_field_template = { dft_id : string; dft_mk : T.typ list -> dec_field }
+
+let string_of_mono_goal (g : mono_goal) : string =
+  String.concat "$" (g.mg_id :: List.map (fun t ->
+    match t, T.normalize t with
+    | _, T.Prim T.Int  -> "Int"
+    | _, T.Prim T.Nat  -> "Nat"
+    | _, T.Prim T.Bool -> "Bool"
+    | T.Con (con, []), _ -> Mo_types.Cons.name con (* TODO: encode type arguments *)
```
ggreif, line 29: Is this some emulation of a nominal type convention?

src/viper/README.md
```diff
@@ -240,33 +513,53 @@ Below, we summarize the language features that _Motoko-san_ currently supports.
...
-            };
-            ```
-    
     * `assert:system` — Compile-time assertions
     
-    **Loop invariants** — Extension is _simple_.
+    * `assert:loop:invariant` — Loop invariants
     
-    **Pure functions** — The tool could be easily extended with a keyword, e.g., `@pure`, to specify functions that are verifier to be side-effect free; such functions could be used inside other code specifications, e.g., `assert:invariant is_okay()` for some `@pure func is_okay() : Bool`.
+    **Expressions in assertions** — Note that using the syntax bellow outside of assertions causes undefined behavior.
```
crusso, line 521:
```suggestion
    **Expressions in assertions** — Note that using the syntax below outside of assertions causes undefined behavior.
```

src/viper/README.md
```diff
@@ -240,33 +513,53 @@ Below, we summarize the language features that _Motoko-san_ currently supports.
...
+          assert:return x < old_x;
+      };
+      ```
+
+    * `Prim.Forall<T>(func x ..)` and `Prim.Exists<T>(func x ..)` — refers ∀ and ∃ quantifiers. For example:
+
+      ```motoko
+      private func fill_array(x: [var Bool], val: Bool): () {
+         assert:return Prim.Forall<Nat>(func i = (0 <= i and i < x.size() implies a[i] == val));
+      }
```
crusso, line 551: Isn't it lowercase Prim.forall and Prim.exists in the implementation now?

src/viper/README.md
```diff
@@ -240,33 +513,53 @@ Below, we summarize the language features that _Motoko-san_ currently supports.
...
+      ```
+    * `Prim.Ret<T>()` — refers to the result value of a function. It's temporal syntax and requires to manually annotate type of the value. Example:
```
crusso, line 553:
```suggestion
    * `Prim.Ret<T>()` — refers to the result value of a function. It's temporary syntax and requires to manually annotate the type of the value. Example:
```

Actually, should it be Prim.ret (lowercase) to match Prim.forall and Prim.exists (as implemented).

crusso (changes requested):
I think we are good to merge in the current state (I can fix up some minor issues later to forbid the new prims in compiled code). But it would be great if the fork could sync with master first and resolve any conflicts. 

You may need see some changes in the test output because of a recent change affected unused identifier warnings...

crusso (approved):
Approving in anticipation of CI succeeding...

ggreif:
@mergifyio refresh

ggreif:
@mergifyio merge

ggreif:
https://github.com/Mergifyio update

PR #4507, opened 2024-04-22
exploration: `finally`

ggreif:
Let's figure out how to interpret `finally` first...

### DIFF
--- a/src/lowering/desugar.ml
+++ b/src/lowering/desugar.ml
@@ -216,7 +216,7 @@ and exp' at note = function
   | S.OldE e -> (oldE (exp e)).it
   | S.IfE (e1, e2, e3) -> I.IfE (exp e1, exp e2, exp e3)
   | S.SwitchE (e1, cs) -> I.SwitchE (exp e1, cases cs)
-  | S.TryE (e1, cs) -> I.TryE (exp e1, cases cs)
+  | S.TryE (e1, cs, None) -> I.TryE (exp e1, cases cs)
   | S.WhileE (e1, e2) -> (whileE (exp e1) (exp e2)).it
   | S.LoopE (e1, None) -> I.LoopE (exp e1)
   | S.LoopE (e1, Some e2) -> (loopWhileE (exp e1) (exp e2)).it
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -942,7 +942,7 @@ let rec is_explicit_exp e =
   | ObjBlockE (_, _, dfs) ->
     List.for_all (fun (df : dec_field) -> is_explicit_dec df.it.dec) dfs
   | ArrayE (_, es) -> List.exists is_explicit_exp es
-  | SwitchE (e1, cs) | TryE (e1, cs) ->
+  | SwitchE (e1, cs) | TryE (e1, cs, None) ->
     is_explicit_exp e1 &&
     List.exists (fun (c : case) -> is_explicit_exp c.it.exp) cs
   | BlockE ds -> List.for_all is_explicit_dec ds
@@ -1513,7 +1513,7 @@ and infer_exp'' env exp : T.typ =
     if not env.pre then
       coverage_cases "switch" env cases t1 exp.at;
     t
-  | TryE (exp1, cases) ->
+  | TryE (exp1, cases, None) ->
     let t1 = infer_exp env exp1 in
     let t2 = infer_cases env T.catch T.Non cases in
     if not env.pre then begin
@@ -1815,7 +1815,7 @@ and check_exp' env0 t exp : T.typ =
     check_cases env t1 t cases;
     coverage_cases "switch" env cases t1 exp.at;
     t
-  | TryE (exp1, cases), _ ->
+  | TryE (exp1, cases, None), _ ->
     check_ErrorCap env "try" exp.at;
     check_exp env t exp1;
     check_cases env T.catch t cases;
--- a/src/mo_interpreter/interpret.ml
+++ b/src/mo_interpreter/interpret.ml
@@ -620,7 +620,7 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
     interpret_exp env exp1 (fun v1 ->
       interpret_cases env cases exp.at v1 k
       )
-  | TryE (exp1, cases) ->
+  | TryE (exp1, cases, None) ->
     let k' = fun v1 -> interpret_catches env cases exp.at v1 k in
     let env' = { env with throws = Some k' } in
     interpret_exp env' exp1 k
--- a/src/mo_def/arrange.ml
+++ b/src/mo_def/arrange.ml
@@ -125,7 +125,8 @@ module Make (Cfg : Config) = struct
     | PrimE p             -> "PrimE"   $$ [Atom p]
     | ImportE (f, _fp)    -> "ImportE" $$ [Atom f]
     | ThrowE e            -> "ThrowE"  $$ [exp e]
-    | TryE (e, cs)        -> "TryE"    $$ [exp e] @ List.map catch cs
+    | TryE (e, cs, None)  -> "TryE"    $$ [exp e] @ List.map catch cs
+    | TryE (e, cs, Some _)-> "TryE FINALLY" $$ [exp e] @ List.map catch cs (* FIXME *)
     | IgnoreE e           -> "IgnoreE" $$ [exp e]))
 
   and exps es = List.map exp es
--- a/src/mo_def/syntax.ml
+++ b/src/mo_def/syntax.ml
@@ -193,7 +193,7 @@ and exp' =
   | AnnotE of exp * typ                        (* type annotation *)
   | ImportE of (string * resolved_import ref)  (* import statement *)
   | ThrowE of exp                              (* throw exception *)
-  | TryE of exp * case list                    (* catch exception *)
+  | TryE of exp * case list * exp option       (* catch exception / finally *)
   | IgnoreE of exp                             (* ignore *)
 (*
   | FinalE of exp * exp                        (* finally *)
--- a/src/mo_frontend/definedness.ml
+++ b/src/mo_frontend/definedness.ml
@@ -117,7 +117,8 @@ let rec exp msgs e : f = match e.it with
   | OldE e              -> exp msgs e
   | IfE (e1, e2, e3)    -> exps msgs [e1; e2; e3]
   | SwitchE (e, cs)     -> exp msgs e ++ cases msgs cs
-  | TryE (e, cs)        -> exp msgs e ++ cases msgs cs
+  | TryE (e, cs, None)  -> exp msgs e ++ cases msgs cs
+  | TryE (e, cs, Some f)-> exps msgs [e; f] ++ cases msgs cs
   | WhileE (e1, e2)     -> exps msgs [e1; e2]
   | LoopE (e1, None)    -> exp msgs e1
   | LoopE (e1, Some e2) -> exps msgs [e1; e2]
--- a/src/mo_frontend/parser.mly
+++ b/src/mo_frontend/parser.mly
@@ -708,7 +708,7 @@ exp_un(B) :
   | IF b=exp_nullary(ob) e1=exp_nest ELSE e2=exp_nest
     { IfE(b, e1, e2) @? at $sloc }
   | TRY e1=exp_nest c=catch
-    { TryE(e1, [c]) @? at $sloc }
+    { TryE(e1, [c], None) @? at $sloc }
 (* TODO: enable multi-branch TRY (already supported by compiler)
   | TRY e=exp_nest LCURLY cs=seplist(case, semicolon) RCURLY
     { TryE(e, cs) @? at $sloc }
--- a/src/mo_frontend/traversals.ml
+++ b/src/mo_frontend/traversals.ml
@@ -60,8 +60,8 @@ let rec over_exp (f : exp -> exp) (exp : exp) : exp = match exp.it with
      f { exp with it = ObjE (List.map (over_exp f) bases, List.map (over_exp_field f) efs) }
   | IfE (exp1, exp2, exp3) ->
      f { exp with it = IfE(over_exp f exp1, over_exp f exp2, over_exp f exp3) }
-  | TryE (exp1, cases) ->
-     f { exp with it = TryE (over_exp f exp1, List.map (over_case f) cases) }
+  | TryE (exp1, cases, exp2_opt) ->
+     f { exp with it = TryE (over_exp f exp1, List.map (over_case f) cases, Option.map (over_exp f) exp2_opt) }
   | SwitchE (exp1, cases) ->
      f { exp with it = SwitchE (over_exp f exp1, List.map (over_case f) cases) }
   | FuncE (name, sort_pat, typ_binds, pat, typ_opt, sugar, exp1) ->


src/mo_interpreter/interpret.ml
```diff
@@ -620,10 +620,17 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
...
+  | TryE (exp1, cases, None) ->
     let k' = fun v1 -> interpret_catches env cases exp.at v1 k in
     let env' = { env with throws = Some k' } in
     interpret_exp env' exp1 k
+  | TryE (exp1, cases, Some exp2) ->
+    let k' v1 =
+      let cleanup v2 = interpret_exp env exp2 (fun _ -> k v2) in
+      interpret_catches env cases exp.at v1 cleanup in
+    let env' = { env with throws = Some k' } in
+    let k'' v2 = interpret_exp env' exp2 (fun _ -> k v2) in
```
ggreif, line 632: here too!

src/mo_interpreter/interpret.ml
```diff
@@ -620,10 +620,17 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
...
-  | TryE (exp1, cases) ->
+  | TryE (exp1, cases, None) ->
     let k' = fun v1 -> interpret_catches env cases exp.at v1 k in
     let env' = { env with throws = Some k' } in
     interpret_exp env' exp1 k
+  | TryE (exp1, cases, Some exp2) ->
+    let k' v1 =
+      let cleanup v2 = interpret_exp env exp2 (fun _ -> k v2) in
+      interpret_catches env cases exp.at v1 cleanup in
+    let env' = { env with throws = Some k' } in
```
crusso, line 631: I'm not sure this is enough. Unless we disallow returns and break from exp1 and exp2, I think you will need to redefine the rets ands labs continuations too, execute the finally before proceeding. This is getting complicated....

crusso: I think your current code will allow a `return` (or `break`) from `exp1` or `cases` to skip the finally clause - only if the execution falls through `exp1` or `cases` is the finally clause executed. A throw from `cases` will probably also bypass the finally.

crusso: Probably worth having a careful read of this https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/statements#1311-the-try-statement

src/codegen/compile.ml
```diff
@@ -10576,6 +10582,7 @@ and compile_prim_invocation (env : E.t) ae p es at =
 
   begin match p, es with
   (* Calls *)
+  | CallPrim _, [e1; e2; _] (* FIXME just ignore the third (cleanup stack) arg for now *)
```
ggreif, line 10585: undo this

src/ir_passes/await.ml
```diff
@@ -37,6 +37,19 @@ let letcont k scope =
...
+(* pre-compose a continuation with a call to a `finally`-thunk *)
+let precont k thunk =
+  let finally = blockE [expD (thunk -*- unitE ())] in
+  match k with
+  | ContVar k' ->
+     let typ = match typ_of_var k' with
+       | T.(Func (Local, Returns, [], [], _)) -> T.unit
+       | T.(Func (Local, Returns, [], [typ], _)) -> typ
+       | _ -> assert false in
+    MetaCont (typ, fun v -> finally (varE k' -*- varE v))
```
ggreif, line 49: am I careful enough avoiding code duplication here?

ggreif: by using `meta` instead of `MetaCont` directly, we'll assert non-duplication

doc/md/examples/grammar.txt
```diff
@@ -217,6 +217,8 @@
     'if' <exp_nullary> <exp_nest>
     'if' <exp_nullary> <exp_nest> 'else' <exp_nest>
     'try' <exp_nest> <catch>
+    'try' <exp_nest> F'in'ALLY <exp_nest>
```
ggreif, line 220: WTF?

src/ir_passes/await.ml
```diff
@@ -37,8 +37,21 @@ let letcont k scope =
     blockE [funcD k' v e] (* at this point, I'm really worried about variable capture *)
             (scope k')
 
+(* pre-compose a continuation with a call to a `finally`-thunk *)
+let precont k thunk =
+  let finally = blockE [expD (thunk -*- unitE ())] in
+  match k with
+  | ContVar k' ->
+     let typ = match typ_of_var k' with
+       | T.(Func (Local, Returns, [], [], _)) -> T.unit
```
ggreif, line 46: there is no `Type.dom`, only `Type.codom`.

src/ir_passes/await.ml
ggreif: Indeed, I get assertion failures when doing it. Investigating.

src/ir_def/check_ir.ml
```diff
@@ -365,13 +365,13 @@ let rec check_exp env (exp:Ir.exp) : unit =
   (* helpers *)
   let check p = check env exp.at p in
   let (<:) t1 t2 =
-(*  try *)
+  try
       check_sub env exp.at t1 t2
-(*  with e ->
+  with e ->
```
ggreif, line 370: revert!

src/ir_def/check_ir.ml
```diff
@@ -409,7 +409,7 @@ let rec check_exp env (exp:Ir.exp) : unit =
             check_concrete env exp.at t_ret;
           end;
           typ exp2 <: t_arg;
-          t_ret <: t
+          t_ret <: t;
```
ggreif, line 412: undo

src/ir_def/construct.mli
```diff
@@ -127,6 +127,8 @@ val let_no_shadow : var -> exp -> dec list -> dec list
 
 val contT : typ -> typ -> typ
 val err_contT : typ -> typ
+val bail_contT : typ
+val clean_contT : typ
 val answerT : typ -> typ (* answer type of a continuation type *)
 
 val cpsT : typ -> typ -> typ
```
ggreif, line 134: This is never used. Remove?

src/ir_def/construct.ml
```diff
@@ -632,17 +627,20 @@ let contT typ ans_typ = T.(Func (Local, Returns, [], as_seq typ, as_seq ans_typ)
...
+
+let clean_contT = bail_contT (* last-resort replica callback *)
+
 let answerT typ : T.typ =
   match typ with
   | T.Func (T.Local, T.Returns, [], ts1, ts2) -> T.seq ts2
   | _ -> assert false
 
-let cpsT typ ans_typ = T.(Func (Local, Returns, [], [contT typ ans_typ; err_contT ans_typ], as_seq ans_typ))
+let cpsT typ ans_typ = T.(Func (Local, Returns, [], [contT typ ans_typ; err_contT ans_typ; bail_contT], as_seq ans_typ))
```
ggreif, line 639: See `.mli`!

src/prelude/internals.mo
```diff
@@ -285,7 +285,7 @@ func @equal_array<T>(eq : (T, T) -> Bool, a : [T], b : [T]) : Bool {
 };
 
 type @Cont<T> = T -> () ;
-type @Async<T> = (@Cont<T>,@Cont<Error>) -> {
+type @Async<T> = (@Cont<T>, @Cont<Error>, () -> ()) -> {
```
ggreif, line 288: Add `@BailCont` and `@CleanCont` here too?

src/prelude/internals.mo
```diff
@@ -307,7 +307,11 @@ func @getSystemRefund() : @Refund {
   return (prim "cyclesRefunded" : () -> Nat) ();
 };
 
-func @new_async<T <: Any>() : (@Async<T>, @Cont<T>, @Cont<Error>) {
+func @cleanup() {
+    (prim "print" : Text -> ()) "CLEANUP_E"
+};
+
+func @new_async<T <: Any>() : (@Async<T>, @Cont<T>, @Cont<Error>, () -> ()) {
```
ggreif, line 314: `@CleanCont`

src/prelude/internals.mo
ggreif: `@BailCont`

src/prelude/internals.mo
```diff
@@ -342,10 +346,17 @@ func @new_async<T <: Any>() : (@Async<T>, @Cont<T>, @Cont<Error>) {
     };
   };
 
-  func enqueue(k : @Cont<T>, r : @Cont<Error>) : {
+  var cleanup : () -> () = @cleanup;
```
ggreif, line 349: `@BailCont`

src/prelude/internals.mo
```diff
@@ -342,10 +346,17 @@ func @new_async<T <: Any>() : (@Async<T>, @Cont<T>, @Cont<Error>) {
...
+
+  func clean() {
+      cleanup();
+  };
+
+  func enqueue(k : @Cont<T>, r : @Cont<Error>, c : () -> ()) : {
```
ggreif, line 355: `@BailCont`

src/lowering/desugar.ml
```diff
@@ -216,7 +216,16 @@ and exp' at note = function
...
+  | S.TryE (e1, cs, None) -> I.TryE (exp e1, cases cs, None)
+  | S.TryE (e1, cs, Some e2) ->
+    let thunk = T.(funcE ("$FIXME") Local Returns [] [] [] (exp e2)) in
+    assert T.(is_func thunk.note.Note.typ);
+    let th = fresh_var "thunk" thunk.note.Note.typ in
+    let post e1 =
+      let v = fresh_var "res" note.Note.typ in
+      blockE [letD v e1; expD (varE th -*- unitE ())] (varE v) in
+    (blockE [letD th thunk] (* use funcD for thunk? *)
+      { e1 with it = I.TryE (exp e1 |> post, cases_map post cs, Some (varE th)); note }).it
```
ggreif, line 228: Instead of post-composing all the bits, just post-compose the whole:
``` scheme
blockE (res = TryE; cleanup()) res
```

ggreif: Or, maybe better, do all post-composition in `await.ml`.

src/ir_passes/await.ml
```diff
@@ -100,9 +114,9 @@ and t_exp' context exp =
     SwitchE (t_exp context exp1, cases')
   | LoopE exp1 ->
     LoopE (t_exp context exp1)
-  | LabelE (id, _typ, exp1) ->
+  | LabelE (id, typ, exp1) ->
```
ggreif, line 117: why these changes?

src/ir_passes/await.ml
```diff
@@ -359,6 +373,47 @@ and c_exp' context exp k =
         ]
         (c_exp context' exp1 (ContVar k))
     ))
+  | TryE (exp1, cases, Some ({it = VarE id2; _} as exp2)) ->
```
ggreif, line 376: Why not communicate the `var` to this point (from `desugar.ml`)?

src/mo_interpreter/interpret.ml
```diff
@@ -620,10 +620,20 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
...
     interpret_exp env' exp1 k
+  | TryE (exp1, cases, Some exp2) ->
+    let k' v1 =
+      let cleanup v2 = interpret_exp env exp2 (fun _ -> k v2) in
+      interpret_catches env cases exp.at v1 cleanup in
+    let out ret v = interpret_exp env exp2 (fun _ -> ret v) in
+    let env' = { env with throws = Some k'
+                        ; rets = Option.map out env.rets
+                        ; labs = V.Env.map out env.labs } in
+    let k'' v2 = interpret_exp env' exp2 (fun _ -> k v2) in
```
ggreif, line 635: will this `env'` cause double execution of `finally` if it contains `return`?

src/mo_interpreter/interpret.ml
```diff
@@ -620,10 +620,17 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
...
       interpret_cases env cases exp.at v1 k
       )
-  | TryE (exp1, cases) ->
+  | TryE (exp1, cases, None) ->
     let k' = fun v1 -> interpret_catches env cases exp.at v1 k in
     let env' = { env with throws = Some k' } in
     interpret_exp env' exp1 k
+  | TryE (exp1, cases, Some exp2) ->
+    let k' v1 =
+      let cleanup v2 = interpret_exp env exp2 (fun _ -> k v2) in
```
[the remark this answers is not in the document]
ggreif: There is also quite some repetition in this code. Can it be recaptured better?

src/ir_interpreter/interpret_ir.ml
```diff
@@ -498,16 +498,14 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
...
   | TryE (exp1, cases, Some (id, ty)) ->
     assert env.flavor.has_await;
-    let exp2 = Construct.(varE (var id ty)) in
+    let exp2 = Construct.(varE (var id ty) -*- unitE ()) in
     let k' v1 =
-      let cleanup v2 = interpret_exp env exp2 (fun _ -> k v2) in
-      interpret_catches env cases exp.at v1 cleanup in
-    let out ret v = interpret_exp env exp2 (fun _ -> ret v) in
-    let k'' v2 = interpret_exp env exp2 (fun _ -> k v2) in
+      interpret_catches env cases exp.at v1 k in
```
crusso, line 503: Isn't it incorrect to skip the cleanup in the cases?

ggreif: Nope, as this is post-de sugaring which already inserted cleanups into the regular flow (i.e. after the entire `try .. catch`).

src/ir_interpreter/interpret_ir.ml
```diff
@@ -498,16 +498,14 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
     interpret_exp env' exp1 k
   | TryE (exp1, cases, Some (id, ty)) ->
     assert env.flavor.has_await;
-    let exp2 = Construct.(varE (var id ty)) in
+    let exp2 = Construct.(varE (var id ty) -*- unitE ()) in
```
ggreif, line 501: @crusso suggested to not construct syntax on the fly, but interpret the call. Not sure if that is simpler, yet.

src/lowering/desugar.ml
```diff
@@ -216,7 +216,16 @@ and exp' at note = function
   | S.OldE e -> (oldE (exp e)).it
   | S.IfE (e1, e2, e3) -> I.IfE (exp e1, exp e2, exp e3)
   | S.SwitchE (e1, cs) -> I.SwitchE (exp e1, cases cs)
-  | S.TryE (e1, cs) -> I.TryE (exp e1, cases cs)
+  | S.TryE (e1, cs, None) -> I.TryE (exp e1, cases cs, None)
```
crusso, line 219: Hmm, doing the transformation half in desugaring and half later seems a bit dangerous to me. Why not simplify desugaring to just name the thunk and do the rest later...

But I see that is on your TODO list already...

src/mo_interpreter/interpret.ml
ggreif: Indeed, but it is fixed now.

test/run-drun/try-finally.mo
```diff
@@ -0,0 +1,264 @@
...
+        try /*ignore*/ await t0() catch _ {};
+        try await t0r() catch _ {};
+        try await t0d() catch _ {};
+        try await t2t() catch _ {};
+        try await t6c() catch _ {};
+        try await t6t() catch _ {};
+        try await t6d() catch _ {};
+        try await t8t() catch _ {};
+        /// caveat: t7 won't return!
+        try await* t7() catch _ {} finally debugPrint "It's over";
```
ggreif, line 258: @crusso my gut feeling says `"It's over"` should be output, but it doesn't appear...

crusso: Oh dear, that might actually be true. Did you modify the async* value translation to take an additional continuation, like the async value translation? If not, maybe that is what is missing here.

src/ir_passes/await.ml
```diff
@@ -68,12 +81,13 @@ let rec t_async context exp =
    (* add the implicit return label *)
    let k_ret = fresh_cont (typ exp1) T.unit in
    let k_fail = fresh_err_cont T.unit in
+   let k_clean = fresh_cont T.unit T.unit in
    let context' =
```
crusso, line 85: Should context' perhaps contain a Cleanup label? Maybe that's why your other test isn't chaining as expected across await* calls... Not sure.

test/run-drun/try-finally.mo
ggreif: I think `t_async` is missing a key piece. On it.

src/ir_passes/await.ml
ggreif: Yes, ae686d6d5. I also just noticed the asymmetry.

src/ir_passes/await.ml
```diff
@@ -68,12 +81,14 @@ let rec t_async context exp =
...
-     LabelEnv.add Return (Cont (ContVar k_ret))
-       (LabelEnv.add Throw (Cont (ContVar k_fail)) LabelEnv.empty)
+     LabelEnv.add Cleanup (Cont (ContVar k_clean))
```
ggreif, line 86: No idea how to test for uninterrupted `Cleanup` chain for the `Fut` case. "This side" should not trap on message sends. The "other side" (i.e. the execution) is behind a context switch.

src/ir_passes/await.ml
```diff
@@ -370,41 +426,39 @@ and c_exp' context exp k =
...
     assert false (* must have effect T.Triv, handled by first case *)
   | AsyncE (T.Fut, tb, exp1, typ1) ->
     (* add the implicit return label *)
     let k_ret = fresh_cont (typ exp1) T.unit in
     let k_fail = fresh_err_cont T.unit in
+    let k_clean = fresh_cont T.unit T.unit in
     let context' =
-      LabelEnv.add Return (Cont (ContVar k_ret))
-        (LabelEnv.add Throw (Cont (ContVar k_fail)) LabelEnv.empty)
+      LabelEnv.add Cleanup (Cont (ContVar k_clean))
```
ggreif, line 451: Similarly for inline `async { ... }` the send shouldn't trap and the only code executed is RTS. No user code is running. But I am adding it for symmetry reasons, without a test.

ggreif: @crusso what do you think? Could I test this somehow?

src/ir_passes/async.ml
```diff
@@ -248,100 +255,105 @@ let transform prog =
...
+          t_typ (T.open_ [t] t_clean)
         | t -> assert false
       in
       let v_ret = fresh_var "v" t_ret in
       let v_fail = fresh_var "e" t_fail in
-      ([v_ret; v_fail] -->* (callE (t_exp exp1) [t0] (tupE [varE v_ret; varE v_fail]))).it
-    | PrimE (CallPrim typs, [exp1; exp2]) when is_awaitable_func exp1 ->
+      let v_clean = fresh_var "c" t_clean in
+      ([v_ret; v_fail; v_clean] -->* callE (t_exp exp1) [t0] (tupE [varE v_ret; varE v_fail; varE v_clean])).it
+    | PrimE (CallPrim typs, (exp1 :: exp2 :: _)) when is_awaitable_func exp1 ->
```
ggreif, line 323: revert

src/ir_passes/await.ml
```diff
@@ -359,6 +374,47 @@ and c_exp' context exp k =
         ]
         (c_exp context' exp1 (ContVar k))
     ))
+  | TryE (exp1, cases, Some (id2, typ2)) ->
```
ggreif, line 377: Deduplicate with the above if possible

src/ir_interpreter/interpret_ir.ml
```diff
@@ -822,7 +833,7 @@ and interpret_func env at sort x args f c v (k : V.value V.cont) =
 
 and interpret_message env at x args f c v (k : V.value V.cont) =
   let v_caller, v_reply, v_reject = match V.as_tup c with
-    | [v_caller; v_reply; v_reject] -> v_caller, v_reply, v_reject
+    | [v_caller; v_reply; v_reject; _v_cleanup] -> v_caller, v_reply, v_reject
```
ggreif, line 836: We _could_ save the cleanup in a global and run it when there is an error (e.g. `assert`ion). WDYT?

ggreif: Alternatively we could save it into the continuation environment and invoke it before fatal errors!

src/ir_passes/await.ml
```diff
@@ -359,6 +374,47 @@ and c_exp' context exp k =
...
+              | T.Await -> c_exp context exp (ContVar k)
+            in
+            { it = { pat; exp = exp' }; at; note })
+          cases
+        @ [{ it = {pat = varP error; exp = varE f -*- varE error};
+             at = no_region;
+             note = ()
+        }] in
+      let throw = fresh_err_cont (answerT (typ_of_var k)) in
+      let lab = function
```
ggreif, line 400: `pre` would be a better name!

src/ir_passes/await.ml
```diff
@@ -359,6 +374,47 @@ and c_exp' context exp k =
...
+    letcont f (fun f ->
+    letcont k (fun k ->
+    match eff exp1 with
+    | T.Triv -> assert false (* FIXME: maybe lift this later? *)
+    | T.Await ->
+      let error = fresh_var "v" T.catch in
+      let cases' =
+        List.map
+          (fun {it = {pat;exp}; at; note} ->
+            let exp' = match eff exp with
```
ggreif, line 389: Isn't `eff exp` _always_ `Await`?

src/ir_passes/await.ml
```diff
@@ -397,14 +397,17 @@ and c_exp' context exp k =
...
         | Label -> assert false
       in
-      let context' = LabelEnv.mapi (function | Return | Named _ -> lab | Cleanup | Throw -> fun c -> c) context in
+      let context' = LabelEnv.mapi (function | Return | Named _ | Cleanup -> pre
+                                             | Throw -> fun c -> c) context in
       let context'' = LabelEnv.add Throw (Cont (ContVar throw)) context' in
-      let c = match LabelEnv.find_opt Cleanup context'' with Some c -> c | None -> Cont (ContVar (var "@cleanup" bail_contT)) in
-      let context''' = LabelEnv.add Cleanup (lab c) context'' in
+      let c = match LabelEnv.find_opt Cleanup context'' with
+        | None -> Cont (ContVar (var "@cleanup" bail_contT))
```
ggreif, line 408: This was a bad commit, as `Cleanup` is a no-op!

src/mo_frontend/typing.ml
```diff
@@ -1815,11 +1823,21 @@ and check_exp' env0 t exp : T.typ =
...
     check_cases env T.catch t cases;
     coverage_cases "try handler" env cases T.catch exp.at;
+    if not env.pre then
+      begin match exp2_opt with
+      | None -> ()
+      | Some exp2 ->
+        check_exp_strong env T.unit exp2;
+        if exp2.note.note_eff <> T.Triv then
+          local_error env exp2.at "M0199" "a cleanup clause must not send messages";
+        if exp1.note.note_eff <> T.Await then
```
ggreif, line 1838: Undo this, committed by accident!

src/ir_passes/await.ml
ggreif: Nope, this is the `exp` of the `case`!

ggreif:
@robin-kunzler I have added you to the reviewers, as you voiced interest in resolution to [FOLLOW-1438](https://dfinity.atlassian.net/browse/FOLLOW-1438).

src/lowering/desugar.ml
ggreif: This is fixed now, the desugaring step only creates the thunk, which is then lowered in `await.ml`. The interpreters are now functionally identical too :-)

src/mo_interpreter/interpret.ml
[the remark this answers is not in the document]
ggreif: Done!

src/mo_interpreter/interpret.ml
ggreif: Done!

src/mo_interpreter/interpret.ml
ggreif: Right. We now enter the `try` body with all possible continuations (success, `throw`/failure, `return` and `label`) augmented. I also added tests for all these cases.

src/mo_frontend/typing.ml
ggreif: Done!

src/codegen/compile.ml
```diff
@@ -9411,7 +9411,12 @@ module FuncDec = struct
     Func.define_built_in env name ["env", I32Type] [] (fun env ->
         G.i (LocalGet (nr 0l)) ^^
         ContinuationTable.recall env ^^
-        G.i Drop);
+        Arr.load_field env 2l ^^ (* get the cleanup closure *)
```
ggreif, line 9414: This is the meat.

src/ir_passes/await.ml
ggreif: ah, `typ` is *used*, so don't add leading underscore

src/ir_passes/async.ml
```diff
@@ -69,31 +72,34 @@ let new_async t =
...
-      let k = fresh_var "k" (contT u T.unit) in
-      let r = fresh_var "r" (err_contT T.unit) in
-      [k; r] -->* (
+      let k = fresh_var "k" (contT u unit) in
+      let r = fresh_var "r" (err_contT unit) in
+      let c = fresh_var "c" bail_contT in
+      [k; r; c] -->* (
         varE unary_async -*-
           (tupE [
              [v] -->* (varE k -*- varE v);
```
ggreif, line 92: use `-->`

src/ir_passes/async.ml
```diff
@@ -248,84 +255,87 @@ let transform prog =
...
           let ic_reply = (* flatten v, here and below? *)
             let v = fresh_var "v" (T.seq ts1) in
             v --> (ic_replyE ts1 (varE v)) in
           let ic_reject =
-            let e = fresh_var "e" T.catch in
-            [e] -->* (ic_rejectE (errorMessageE (varE e))) in
-          let exp' = callE (t_exp exp1) [t0] (tupE [ic_reply; ic_reject]) in
-          expD (selfcallE ts1 exp' (varE nary_reply) (varE reject))
+            let e = fresh_var "e" catch in
+            [e] -->* ic_rejectE (errorMessageE (varE e)) in
```
ggreif, line 301: `-->`

src/ir_passes/async.ml
```diff
@@ -385,20 +395,22 @@ let transform prog =
...
               let k =
                 let v = fresh_var "v" t1 in
                 v --> (ic_replyE ret_tys (varE v)) in
               let r =
-                let e = fresh_var "e" T.catch in
-                [e] -->* (ic_rejectE (errorMessageE (varE e))) in
-              let exp' = callE (t_exp cps) [t0] (tupE [k;r]) in
-              FuncE (x, T.Shared s', Replies, typbinds', args', ret_tys, exp')
+                let e = fresh_var "e" catch in
+                [e] -->* ic_rejectE (errorMessageE (varE e)) in
```
ggreif, line 409: `-->`

src/lowering/desugar.ml
```diff
@@ -216,7 +216,14 @@ and exp' at note = function
...
+  | S.TryE (e1, cs, Some e2) ->
+    let thunk = T.(funcE ("$cleanup") Local Returns [] [] [] (exp e2)) in
+    assert T.(is_func thunk.note.Note.typ);
+    let th = fresh_var "thunk" thunk.note.Note.typ in
+    (blockE
+       [ letD th thunk ]
```
ggreif, line 225: use `funcD` or (`-->*` above)

src/mo_frontend/typing.ml
```diff
@@ -1529,12 +1532,13 @@ and infer_exp'' env exp : T.typ =
     if not env.pre then
       coverage_cases "switch" env cases t1 exp.at;
     t
-  | TryE (exp1, cases) ->
+  | TryE (exp1, cases, exp2_opt) ->
     let t1 = infer_exp env exp1 in
     let t2 = infer_cases env T.catch T.Non cases in
     if not env.pre then begin
+      Option.iter (check_exp_strong { env with rets = None; labs = T.Env.empty } T.unit) exp2_opt;
```
crusso, line 1539: Nice - no returns no exiting jumps allowed. How are throws prevented though?

crusso:
```suggestion
      Option.iter (check_exp_strong { env with rets = None; labs = T.Env.empty; throws = None } T.unit) exp2_opt;
```
Maybe add a test that incorrectly tries to throw from finally and then make this change.

src/mo_frontend/typing.ml
```diff
@@ -1831,11 +1835,20 @@ and check_exp' env0 t exp : T.typ =
...
     check_exp env t exp1;
     check_cases env T.catch t cases;
-    coverage_cases "try handler" env cases T.catch exp.at;
+    if cases <> []
+    then coverage_cases "try handler" env cases T.catch exp.at;
+    if not env.pre then
+      begin match exp2_opt with
+      | None -> ()
+      | Some exp2 ->
+        check_exp_strong { env with rets = None; labs = T.Env.empty } T.unit exp2;
```
crusso, line 1848:
```suggestion
        check_exp_strong { env with rets = None; labs = T.Env.empty; throws = None } T.unit exp2;
```

src/mo_frontend/typing.ml
```diff
@@ -1831,11 +1835,20 @@ and check_exp' env0 t exp : T.typ =
...
+        if exp2.note.note_eff <> T.Triv then
+          local_error env exp2.at "M0200" "a cleanup clause must not send messages";
+      end;
```
crusso, line 1851: I think a better way to prevent this is to set the env.asyncCap to NullCap so we just allow synchronous code here.

src/mo_interpreter/interpret.ml
```diff
@@ -620,10 +620,17 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
...
       )
-  | TryE (exp1, cases) ->
-    let k' = fun v1 -> interpret_catches env cases exp.at v1 k in
-    let env' = { env with throws = Some k' } in
-    interpret_exp env' exp1 k
+  | TryE (exp1, cases, exp2_opt) ->
+    let k, env = match exp2_opt with
+      | None -> k, env
+      | Some exp2 ->
+        let pre k v = interpret_exp env exp2 (fun _ -> k v) in
```
crusso, line 627:
```suggestion
        let pre k v = interpret_exp env exp2 (fun v2 -> V.as_unit v2; k v) in
```
(Might be a good sanity check)

src/ir_def/check_ir.ml
```diff
@@ -560,7 +560,7 @@ let rec check_exp env (exp:Ir.exp) : unit =
            (match ts2 with
             | [] -> ()
             | _ -> error env exp.at "CPSAwait answer type error");
-           typ kr <: T.Tup [cont_typ; T.Func(T.Local, T.Returns, [], [T.catch], ts2)];
+           typ krc <: T.(Tup Construct.[cont_typ; err_contT (Tup ts2); bail_contT]);
```
crusso, line 563: I'm worried the err_contT arity might be incorrect this way, for example if ts2 = [t2] . Might be safer to revert to original or use T.seq ts2.

src/mo_frontend/typing.ml
ggreif: `throws are prevented by the effect system already (see `test/fail/try-finally`). But this is a good defensive measure.

src/codegen/compile.ml
crusso: I guess we are now open to leaking entries in the closure table if the user-defined cleanup traps.
I don't see how one could avoid this apart from maybe delaying the  user-defined cleanup to run on entry to the next message. Probably too complicated to consider. Let's just live with the possible leak.

Changelog.md
```diff
@@ -1,5 +1,27 @@
...
+
+    A trailing `finally` clause to `try`/`catch` expressions facilitates structured
+    resource deallocation (e.g. acquired locks, etc.) and similar cleanups in the
+    presence of control-flow expressions (`return`, `break`, `continue`, `throw`).
+
+    _Note_: `finally`-expressions that are in scope will be executed even if an execution
+    path _following_ an `await`-expression traps. This behaviour, not available before in Motoko,
+    allows programmers writing last-resort cleanups. For trapping execution paths _without_ an
+    intervening `await`, the replica-provided state rewinding mechanism stays in charge of
+    the cleanup.
```
crusso, line 17:
```suggestion
    _Note_: `finally`-expressions that are in scope will be executed even if an execution
    path _following_ an `await`-expression traps. This feature, not available before in Motoko,
    allows programmers to implement cleanups even in the presence of traps. For trapping 
    execution paths prior to any `await`, the replica-provided state roll back mechanism must 
    ensure no cleanup is required.
```

src/codegen/compile.ml
ggreif: Whatever way to look at it, the last cleanup measure — if failing by trap — will be vulnerable. Leaking a harmless entry (which will never be activated by the replica), and also missing the actual release of lock is the penalty for trapping in this place. But the cross-section of this code is deliberately a small one as it will be hard to pile code into the `finally`. I hope.

doc/md/reference/language-manual.md
```diff
@@ -2547,6 +2550,10 @@ Because the [`Error`](../base/Error.md) type is opaque, the pattern match cannot
 
 :::
 
+The `try` expression can be endowed with a cleanup clause (and optionally omitting the `catch`) to facilitate structured rollback of temporary state changes (e.g. locks).
```
crusso, line 2553:
```suggestion
The `try` expression can be provided with a `finally` cleanup clause to facilitate structured rollback of temporary state changes (e.g. to release a lock). 
The preceding `catch` clause may be omitted in the presence of a `finally` clause.
```

src/ir_def/check_ir.ml
ggreif: Yes, `seq` seems like the right way to do, as `err_contT` will undo it.
```suggestion
           typ krc <: T.(Tup Construct.[cont_typ; err_contT (seq ts2); bail_contT]);
```

src/mo_interpreter/interpret.ml
ggreif: I added a comment above to verify the evaluation result. But `V.as_unit` should suffice here.

src/ir_interpreter/interpret_ir.ml
```diff
@@ -492,10 +492,19 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
...
-    let k' = fun v1 -> interpret_catches env cases exp.at v1 k in
-    let env' = { env with throws = Some k' } in
-    interpret_exp env' exp1 k
+  | TryE (exp1, cases, finally_opt) ->
+    assert env.flavor.has_await;
+    let k, env = match finally_opt with
+      | None -> k, env
+      | Some (id, ty) ->
+        let exp2 = Construct.(varE (var id ty) -*- unitE ()) in
+        let pre k v = interpret_exp env exp2 (fun _ -> k v) in
```
ggreif, line 501:
```suggestion
        let pre k v = interpret_exp env exp2 (fun v2 -> V.as_unit v2; k v) in
```

src/ir_passes/async.ml
ggreif: not here!

src/mo_frontend/typing.ml
ggreif: I'll try!

src/mo_frontend/typing.ml
crusso: See #4622 for the fix

src/mo_frontend/typing.ml
crusso: Done in #4622

src/ir_passes/await.ml
crusso: Did you resolve this?

src/ir_passes/await.ml
```diff
@@ -324,29 +341,45 @@ and c_exp' context exp k =
...
+     let pre k =
+       match finally_opt with
+       | Some (id2, typ2) -> precont k (var id2 typ2)
+       | None -> k in
+    let pre' = function
+      | Cont k -> Cont (pre k)
+      | Label -> assert false in
+    (* All control-flow out must pass through the potential `finally` thunk *)
+    let context = LabelEnv.map pre' context in
+    (* assert that a surrounding `AwaitPrim _` has set up a `Cleanup` cont *)
```
crusso, line 354: I don't grokk this comment

src/ir_passes/await.ml
```diff
@@ -37,8 +39,21 @@ let letcont k scope =
...
-type label = Return | Throw | Named of string
+(* pre-compose a continuation with a call to a `finally`-thunk *)
+let precont k vthunk =
+  let finally e = blockE [expD (varE vthunk -*- unitE ())] e in
+  match k with
+  | ContVar k' ->
+     let typ = match typ_of_var k' with
+       | T.(Func (Local, Returns, [], [], _)) -> T.unit
+       | T.(Func (Local, Returns, [], [typ], _)) -> typ
+       | _ -> assert false in
```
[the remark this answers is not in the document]
crusso: I tried to trigger it but failed, so who knows...

src/ir_passes/await.ml
```diff
@@ -324,29 +341,45 @@ and c_exp' context exp k =
...
+    | T.Await when cases = [] ->
+      c_exp context exp1 (ContVar k)
     | T.Await ->
-      let error = fresh_var "v" T.catch  in
+      let error = fresh_var "v" T.catch in
+      let rethrow = { it = { pat = varP error; exp = varE f -*- varE error };
+                      at = no_region;
+                      note = ()
+                    } in
+      let omit_rethrow = List.exists (fun {it = {pat; exp}; _} -> Ir_utils.is_irrefutable pat) cases in
```
crusso, line 372: This is just an optimization right? I think the default is a actually a hangover from when we had a non-opaque Error type and syntax allowed multiple catch clauses (for more convenient matching of the error code).

crusso (commented):
Some interim comments - will take another pass tomorrow. Still a little confused why the cps translation need to take three arguments since a message a can only reply or reject, but maybe we need it for the internal finally stack...

src/ir_def/check_ir.ml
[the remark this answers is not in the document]
ggreif: I wanted to introduce the convention that `clean_contT` goes towards replica (send message) and `bail_contT` is `await`-related. Many variables with stem "c" (as in `k`-`r`-`c`) should be "b" instead.

src/ir_def/check_ir.ml
```diff
@@ -718,12 +721,13 @@ let rec check_exp env (exp:Ir.exp) : unit =
...
  *)
     check_cases env t1 t cases
-  | TryE (exp1, cases) ->
+  | TryE (exp1, cases, vt) ->
     check env.flavor.has_await "try in non-await flavor";
     check (env.async <> None) "misplaced try";
     check_exp env exp1;
     typ exp1 <: t;
     check_cases env T.catch t cases;
+    Option.iter (fun (_, t) -> t <: Construct.bail_contT) vt
```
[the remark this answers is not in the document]
ggreif: That's not bad...

src/ir_def/construct.ml
```diff
@@ -632,17 +627,20 @@ let contT typ ans_typ = T.(Func (Local, Returns, [], as_seq typ, as_seq ans_typ)
 
 let err_contT ans_typ =  T.(Func (Local, Returns, [], [catch], as_seq ans_typ))
 
+let bail_contT = T.(contT unit unit) (* when `await`ing *)
```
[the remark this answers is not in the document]
ggreif: Bail is for `await`, `clean` is for ICCallPrim and `selfcallE`. They both have `() -> ()` type. `clean_contT` gets invoked by the replica. If there is an `await` pending, then `@new_async` will invoke/delegate to the `bail_contT`.

src/ir_def/construct.mli
ggreif: yes

src/ir_passes/await.ml
```diff
@@ -37,8 +39,21 @@ let letcont k scope =
...
-(* Named labels for break, special labels for return and throw *)
-type label = Return | Throw | Named of string
+(* pre-compose a continuation with a call to a `finally`-thunk *)
+let precont k vthunk =
+  let finally e = blockE [expD (varE vthunk -*- unitE ())] e in
+  match k with
+  | ContVar k' ->
+     let typ = match typ_of_var k' with
+       | T.(Func (Local, Returns, [], [], _)) -> T.unit
+       | T.(Func (Local, Returns, [], [typ], _)) -> typ
```
crusso, line 49:
```suggestion
     let typ = match typ_of_var k' with
       | T.(Func (Local, Returns, [], ts1, _)) -> T.seq ts1
```
Might work and cover all cases

crusso (approved):
Nice work! 

LGTM though its obviously a big hairy change.

The test cases seemed good to me but it's a huge matrix...

Please see comments

test/run-drun/try-finally.mo
```diff
@@ -0,0 +1,334 @@
...
+                throw error "InnerIN6c";
+            } catch _ {
+                debugPrint "InnerCATCH6c";
+                assert false;
+                debugPrint "DEADCATCH6c";
+            } finally { debugPrint "InnerOUT6c" };
+            debugPrint "DEAD6c";
+        }
+        finally { debugPrint "OUT6c" };
+        debugPrint "AFTER6c"
```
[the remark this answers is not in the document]
ggreif: Yeah, now with "DEAD" in that string.

test/run-drun/try-finally.mo
```diff
@@ -0,0 +1,334 @@
...
+                debugPrint "InnerIN6t";
+                await m();
+                debugPrint "InnerLIVE6t";
+                assert false;
+                debugPrint "InnerDEAD6t";
+            } finally { debugPrint "InnerOUT6t" };
+            debugPrint "DEAD6t";
+        }
+        finally { debugPrint "OUT6t" };
+        debugPrint "AFTER6t"
```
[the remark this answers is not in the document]
ggreif: Yep.

src/mo_frontend/typing.ml
```diff
@@ -961,9 +961,12 @@ let rec is_explicit_exp e =
   | ObjBlockE (_, _, dfs) ->
     List.for_all (fun (df : dec_field) -> is_explicit_dec df.it.dec) dfs
   | ArrayE (_, es) -> List.exists is_explicit_exp es
-  | SwitchE (e1, cs) | TryE (e1, cs) ->
+  | SwitchE (e1, cs) | TryE (e1, cs, None) ->
     is_explicit_exp e1 &&
     List.exists (fun (c : case) -> is_explicit_exp c.it.exp) cs
+  | TryE (e1, cs, Some e2) ->
+    is_explicit_exp { e with it = TryE (e1, cs, None) } &&
+    is_explicit_exp e2
```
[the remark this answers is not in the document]
ggreif: I probably don't grok the meaning of "explicit" in this context. Reading up a bit more on it.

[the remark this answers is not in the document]
ggreif: Okay, then 964 needs the same treatment of `cs = []`, right?

src/ir_passes/await.ml
ggreif: Some `AwaitPrim` in the enclosing context is in charge of setting up a trivial-valued `Cleanup` key in the `LabelEnv`. We depend on it being there when mapping over the `context`.

src/ir_passes/await.ml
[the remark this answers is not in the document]
ggreif: By the way how we construct the continuations I very much doubt that we'll hit this `assert`. We always build monomorphic local functions.

src/ir_passes/await.ml
ggreif: No, it is on the TODO list (in the description). I have no idea how to fix it (not a semantic problem) and I prefer to write a follow-on PR to handle it.

src/lowering/desugar.ml
ggreif: `funcD` is unary, that won't work.

src/mo_frontend/typing.ml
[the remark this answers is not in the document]
ggreif: I'd remove `TryE` from 964, then.

src/ir_interpreter/interpret_ir.ml
ggreif: This. But where should we keep the cleanup chain? In a global? Then we are stateful and have to be careful to reset it to `None` (or `@cleanup`) every time we send messages?

src/ir_passes/await.ml
```diff
@@ -12,6 +12,8 @@ let fresh_cont typ ans_typ = fresh_var "k" (contT typ ans_typ)
 
 let fresh_err_cont ans_typ  = fresh_var "r" (err_contT ans_typ)
 
+let fresh_bail_cont ans_typ = fresh_var "b" bail_contT
```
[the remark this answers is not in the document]
ggreif: I _think_ I know what I dislike about the `finally_cont`. It is because the `finally` blocks are composed into all sorts of continuations, but what is meant here is a chain of `finally`-blocks _only_, which is invoked in the specific situation of a trapping `await` continuation. That's why I came up with _bail_, because we are leaving the current context doing only the minimal amount of work possible.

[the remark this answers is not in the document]
ggreif: Of course we can come up with a clean nomenclature any time we wish. Not sure if this naming thing warrants a delay in merging.

src/ir_passes/async.ml
```diff
@@ -416,21 +428,21 @@ let transform prog =
...
-                [e] -->* tupE [] in (* discard error *)
-              let exp' = callE (t_exp cps) [t0] (tupE [k;r]) in
-              FuncE (x, T.Shared s', Returns, typbinds', args', ret_tys, exp')
-            | Returns, _ ->
-              assert false
-            | Replies,_ -> assert false
+                let e = fresh_var "e" catch in
+                e --> tupE [] in
+              let cl =
+                [] -->* unitE () in
```
ggreif, line 442: Wait, this is `varE (var "@cleanup" cleanup_contT)`, right?

ggreif: Heh:
```
$ git grep -nF '[] -->* unitE'
src/ir_passes/async.ml:303:            [] -->* unitE () in
src/ir_passes/async.ml:411:                [] -->* unitE () in
src/ir_passes/async.ml:442:                [] -->* unitE () in
```

ggreif: Changed in 39a65acb9.

src/ir_passes/async.ml
```diff
@@ -248,84 +255,87 @@ let transform prog =
...
             let v = fresh_var "v" (T.seq ts1) in
             v --> (ic_replyE ts1 (varE v)) in
           let ic_reject =
-            let e = fresh_var "e" T.catch in
-            [e] -->* (ic_rejectE (errorMessageE (varE e))) in
-          let exp' = callE (t_exp exp1) [t0] (tupE [ic_reply; ic_reject]) in
-          expD (selfcallE ts1 exp' (varE nary_reply) (varE reject))
+            let e = fresh_var "e" catch in
+            [e] -->* ic_rejectE (errorMessageE (varE e)) in
+          let ic_cleanup =
```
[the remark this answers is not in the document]
ggreif: Good question. Needs some thought. But I also don't have the chain of `finally` thunks around here. So I probably just do the simplest thing possible, in analogy to `ic_reject`. Actually, I think one has to `await` for that cleanup chain to be relevant. However the `selfcallE` _does_ use the `await`-able continuations. So I guess the real answer is that `ic_reject` is invoked when the _other_ side (self-call recipient) traps _after_ an `await`.

src/ir_passes/async.ml
ggreif: Here I know that it is unary.

src/ir_def/check_ir.ml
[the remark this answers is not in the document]
ggreif: See cb0d2366a.

src/ir_def/check_ir.ml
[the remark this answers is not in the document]
ggreif: As explained elsewhere `finallyT` would be rather ambiguous.

src/mo_frontend/typing.ml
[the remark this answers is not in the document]
ggreif: @crusso I _think_ this is the last thing to be resolved. WDYT?

src/prelude/internals.mo
ggreif: See: 1ead3f52f.

src/prelude/internals.mo
ggreif: 1ead3f52f

src/prelude/internals.mo
ggreif: 1ead3f52f

src/prelude/internals.mo
ggreif: 1ead3f52f

src/lowering/desugar.ml
ggreif: Done.

src/ir_interpreter/interpret_ir.ml
ggreif: This is moot now, code is rewritten.

src/ir_interpreter/interpret_ir.ml
ggreif: I left a TODO in the description, maybe in another PR.

src/ir_passes/await.ml
ggreif: done

Changelog.md
```diff
@@ -9,6 +9,8 @@
     A trailing `finally` clause to `try`/`catch` expressions facilitates structured
     resource deallocation (e.g. acquired locks, etc.) and similar cleanups in the
     presence of control-flow expressions (`return`, `break`, `continue`, `throw`).
+    Additionally, in presence of `finally` the `catch` clause becomes optional and
+    potential errors fall through to any enclosing `catch`.
```
crusso, line 13:
```suggestion
    Additionally, in presence of `finally` the `catch` clause becomes optional and
    and any uncaught error from the `try` block will be propagated, after executing the `finally` block.
```

Its probably worth adding a test for this if we don't alread have one. i.e. Somethink like:
```
var done = false
try { throw Prim.error(foo) }
finally { done := true }
assert done
```

test/run-drun/try-finally.mo
```diff
@@ -313,6 +315,21 @@ actor A {
         try await t6t() catch _ {};
         try await t6d() catch _ {};
         try await t8t() catch _ {};
+
+        // other side problem tests
+        try await async {
```
crusso, line 320: Ok, I think this is behaving correctly. Maybe refactor into a separate subtest/function (like the others)

test/bench/ok/heap-32.drun-run.ok
```diff
@@ -1,5 +1,5 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
 debug.print: (50_227, +29_863_068, 769_000_085)
-debug.print: (50_070, +32_992_212, 830_427_314)
+debug.print: (50_070, +32_992_212, 830_427_376)
```
crusso, line 4: Any idea why this has increased so much?

crusso (commented):
See comments

test/bench/ok/heap-32.drun-run.ok
crusso: Oh, weird maybe it hasn't changed. When I was commenting on the recent changes, the diff was more like 10%

src/ir_passes/await.ml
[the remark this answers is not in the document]
crusso: yeah, ok. My only real preference for finally is that it would show up in a grep for finally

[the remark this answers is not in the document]
crusso: happy to live with it.

src/ir_passes/await.ml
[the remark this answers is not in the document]
crusso: (I wasn't worried about the type parameters, which should always be empty, but that ts1 could have any length, not just 0 or 1 as in the previous code.)

src/ir_passes/await.ml
crusso: Hmm, that's a wee bit worrying.

src/ir_passes/await.ml
```diff
@@ -416,19 +446,24 @@ and c_exp' context exp k =
   | PrimE (AwaitPrim s, [exp1]) ->
     let r = match LabelEnv.find_opt Throw context with
       | Some (Cont r) -> r
-      | Some Label
-      | None -> assert false
+      | _ -> assert false
     in
+    let b = match LabelEnv.find_opt Cleanup context with
+      | Some (Cont r) -> r
+      | None -> ContVar (var "@cleanup" bail_contT)
```
crusso, line 453: Actually, can this case (`None`) ever arise? Aren't we always in the context of an AsyncE/AsyncE*, in which case the env  shouldn't contain a Cleanup? Maybe this is a hangover from the code you had before you extended AsyncE*?

crusso (changes requested):
There's something fish in await.ml. Hold off merging for a bit

test/bench/ok/heap-32.drun-run.ok
ggreif: It is just a few cycles. Not a big deal.

test/run-drun/try-finally.mo
ggreif: I'll do this in a separate PR.

crusso (approved):
We can merge but I think it would be good to do the follow up PR addressing code duplication *before* releasing.

PR #4512, opened 2024-04-22
doc: Motoko docs revisions & restructure

crusso:
* organizes docs into sub-folders, unused docs into `doc/md/attic`
* resurrects make preview
* add some new docs (from portal)
* splits some old docs into separate files
* fixes links
* adjusts doc order in nav bar

This PR is the second step in rebuilding #4494, and applies the edits from #4494 to the already moved source files in #4511, improving the diff and easing review.


doc/md/canister-maintenance/compatibility.md
```diff
@@ -1,162 +1,120 @@
...
-In Motoko, we can declare variables to be stable (across upgrades).
 
-``` motoko no-repl file=./examples/count-v1.mo
+``` motoko no-repl file=../examples/count-v1.mo
 ```
 
-Because it’s `stable`, this counter’s `state` is *retained* across upgrades.
-
-(If not marked `stable`, `state` would restart from `0` on upgrade).
+If a variable is not marked `stable`, `state` would restart from `0` on upgrade.
```
crusso, line 32:
```suggestion
If the variable `state` were not declared `stable`, `state` would restart from `0` on upgrade.
```

doc/md/getting-started/basic-concepts.md
crusso, line 1: I actually think we should rewrite this entire section at some point, perhaps by fleshing out the content in overview.md ( a set of slides used to introduce Motoko). But its fine to leave it for now.

doc/md/getting-started/basic-concepts.md
```diff
@@ -309,150 +137,34 @@ In this case, we can omit this annotation without changing the meaning of the pr
...
+However, sometimes type annotations are required by the compiler to infer other assumptions and to check the program as a whole. When they are added and the compiler still accepts the program, you can confirm that the added annotations are consistent with the existing ones.
 
 ## Type errors and messages
 
-Mathematically, the type system of Motoko is *declarative*, meaning that it exists independently of any implementation, as a concept entirely in formal logic. Likewise, the other key aspects of the language definition (for example, its execution semantics) exist outside of an implementation.
-
-However, to design this logical definition, to experiment with it, and to practice making mistakes, we want to interact with this type system, and to make lots of harmless mistakes along the way.
+Mathematically, the type system of Motoko is **declarative**, meaning that it exists independently of any implementation as a concept entirely in formal logic. Likewise, the other key aspects of the language definition, for example its execution semantics, exist outside of an implementation.
```
crusso, line 146:
```suggestion
Motoko is a statically typed programming language. That means that he compiler will reject any program that contains obviously non-sensical code. 

For example, while adding two numbers is allowed, adding a number to a text value makes no sense (to Motoko) and is flagged as an error that must be fixed before the code can be run or deployed. 

The rules that Motoko applies to verify the correctness  of code before compiling and running it are called its type system. Motoko's type system will detect and reject static errors such as applying a function to the wrong number of arguments, or to arguments of the wrong type. 

The type system is a safety feature that prevents a slew of errors that would otherwise have to be detected and reported at runtime, when they would be difficult or impossible to address.
```

doc/md/getting-started/basic-concepts.md
```diff
@@ -309,150 +137,34 @@ In this case, we can omit this annotation without changing the meaning of the pr
...
 
-The error messages of the *type checker* attempt to help the developer when they misunderstand or otherwise misapply the logic of the type system, which is explained indirectly in this book.
+To design this logical definition, experiment with it, and practice making mistakes, you want to interact with this type system. The error messages of the **type checker** attempt to help the developer when they misunderstand or otherwise misapply the logic of the type system.
```
crusso, line 148:
```suggestion
```

doc/md/getting-started/basic-concepts.md
```diff
@@ -309,150 +137,34 @@ In this case, we can omit this annotation without changing the meaning of the pr
...
--->
+Instead, whenever possible, the Motoko base library provides the types and operations that make the language feel complete. However, this base library is still under development, and is still incomplete.
 
-For various practical language engineering reasons, the design of Motoko strives to minimize builtin types and operations.
-
-Instead, whenever possible, the Motoko base library provides the types and operations that make the language feel complete. ***However**, this base library is still under development, and is still incomplete*.
-
-The [Motoko Base Library](./base-intro.md) lists a *selection* of modules from the Motoko base library, focusing on core features used in the examples that are unlikely to change radically. However, all of these base library APIs will certainly change over time (to varying degrees), and in particular, they will grow in size and number.
+The [Motoko base library](../base-intro.md) lists a selection of modules, focusing on core features used in the examples that are unlikely to change radically., although, all of these base library APIs will certainly change over time and in particular, grow in size and number.
```
crusso, line 156:
```suggestion
The [Motoko base library](../base-intro.md) lists a selection of modules, focusing on core features used in the examples that are unlikely to change radically. The base library APIs will likely evolve over time and in particular, grow in size and number as Motoko matures.
```

crusso (commented):
Great stuff. Added some interim comments.

doc/md/getting-started/motoko-introduction.md
```diff
@@ -1,91 +1,145 @@
...
+
+To address the need for correctness without complexity, DFINITY has designed its own Motoko programming language. Motoko provides a simple and expressive alternative to other programming languages that is easy to learn whether you are a new or experienced programmer.
+
+## Motivation and goals
+
+Motoko strives to be a simple and useful language for ICP smart contracts that uses a familiar syntax that incorporates the actor model. Motoko offers seamless integration for ICP features and makes the most out of WebAssembly's present and future functionalities.
+
+Motoko is not, and is not intended to be, the only language for implementing canister smart contracts. Canister development kits exist for Rust, TypeScript, Python, and soon Solidity. ICP's goal is to enable any language to be able to produce canister smart contracts by compiling the canister source code into WebAssembly.
```
crusso, line 33: Who is doing Solidity?

doc/md/reference/style.md
```diff
@@ -1,7 +1,20 @@
...
+
 To increase readability and uniformity of Motoko source code, the style guide provides suggestions for formatting Motoko sources and other basic conventions.
 
+## Motoko grammar
+
+The specification is auto-generated with a tool:
+
+``` bnf file=../examples/grammar.txt
+```
+
```
crusso, line 17:
```suggestion
```

doc/md/reference/style.md
```diff
@@ -728,7 +741,7 @@ Rationale: `g[1]` in particular will be misparsed as an indexing operation.
 
     Limit their use to as small a scope as possible. Rule out the `null` case and use non-option types wherever possible.
 
--   Consider using records (objects with just data) instead of tuples when there are more than 2 or 3 components.
+-   Consider using records, which are objects with just data, instead of tuples when there are more than 2 or 3 components.
```
crusso, line 744:
```suggestion
-   Consider using records instead of tuples when there are more than 2 or 3 components. Records are just simple objects with named fields containing data but not functions.
```

doc/md/writing-motoko/errors.md
```diff
@@ -1,142 +1,136 @@
...
+### How not to handle errors
 
-Asynchronous `Error`s should generally only be used to signal unexpected failures that you cannot recover from, and that you don’t expect many consumers of your API to handle. If a failure should be handled by your caller you should make it explicit in your signature by returning a `Result` instead. For completeness here is the `markDone` example with exceptions:
+An incorrect way of reporting errors is through the use of a sentinel value. For example, for your `markDone` function, you might decide to use the value `-1` to signal that something failed. The callsite then has to check the return value against this special value and report the error. It's easy to not check for that error condition and continue to work with that value in the code. This can lead to delayed or even missing error detection and is strongly discouraged.
```
crusso, line 126:
```suggestion
A generally poor way of reporting errors is through the use of a sentinel value. For example, for your `markDone` function, you might decide to use the value `-1` to signal that something failed. The callsite then has to check the return value against this special value and report the error. It's easy to not check for that error condition and continue to work with that value in the code. This can lead to delayed or even missing error detection and is strongly discouraged.
```

doc/md/getting-started/motoko-introduction.md
```diff
@@ -1,91 +1,145 @@
...
+-   Object-oriented, functional & imperative.
+
+-   Objects as records of functions.
+
+-   `async`/`await` for sequential programming of asynchronous messaging.
+
+-   Structural typing with simple generics and subtyping.
+
+-   Safe arithmetic (both unbounded and checked).
+
```
crusso, line 52:
```suggestion

- Pattern matching with coverage checking

```

doc/md/getting-started/motoko-introduction.md
```diff
@@ -1,91 +1,145 @@
...
+-   Non-nullable types by default.
+
+-   Garbage collected (no manual memory management).
+
+-   JavaScript-like syntax but statically typed & sane.
+
+Semantics of Motoko include:
```
crusso, line 59: I need to rewrite these bullets..

crusso (commented):
More interim comments

doc/md/getting-started/motoko-introduction.md
```diff
@@ -1,91 +1,145 @@
...
+-   No inheritance.
 
-To execute, Motoko statically compiles to [WebAssembly](about-this-guide.md#webassembly), a portable binary format that abstracts cleanly over modern computer hardware, and thus permits its execution broadly on the Internet, and the Internet Computer.
+## Implementation
 
-## Each canister smart contract as an *actor*
+Motoko is implemented in OCaml and leverages the `wasm` library. It uses a simple reference interpreter, but has a more complex compilation into WebAssembly. Compilation uses multipass with typed IR in each pass, and uniform representation, unboxed arithmetic. Motoko offers three garbage collectors (copying, compacting, or generational), and garbage collection is invoked after messages. Motoko uses polymorphism by erasure.
```
crusso, line 75: This is irrelevant to our users, I'd say

doc/md/getting-started/motoko-introduction.md
```diff
@@ -1,91 +1,145 @@
...
+### Code sequentially in direct style
 
-The Internet Computer ensures that each message that is sent receives a response. The response is either success with some value, or an error. An error can be the explicit rejection of the message by the receiving canister, a trap due to an illegal instruction such as division by zero, or a system error due to distribution or resource constraints. For example, a system error might be the transient or permanent unavailability of the receiver (either because the receiving actor is oversubscribed or has been deleted).
+On ICP, canisters can communicate with other canisters by sending asynchronous messages.
 
-### Asynchronous actors
+Asynchronous programming is hard, so Motoko enables you to author asynchronous code in much simpler, sequential style. Asynchronous messages are function calls that return a future, and the `await` construct allows you to suspend execution until a future has completed.
+### Modern type system
```
crusso, line 86:
```suggestion

### Modern type system
```

doc/md/getting-started/motoko-introduction.md
crusso, line 1: The section takes some material from attic/motoko.md but I'm not sure it takes all and seems to have lost all the examples... Did they got elsewhere? Otherwise I'd think it would be better to include them,

doc/md/writing-motoko/arguments.md
```diff
@@ -0,0 +1,131 @@
...
+    str := str # city #", ";
+    };
+    return str # "bon voyage!";
+}
+};
+
+```
+
+You might notice that `Text` in this code example is enclosed by square (`[ ]`) brackets. By itself, `Text` represents a collection of UTF-8 characters. The square brackets around a type indicate that it is an **array** of that type. In this context, therefore, `[Text]` indicates an array of a collection of UTF-8 characters, enabling the program to accept and return multiple text strings.
+
```
crusso, line 92:
```suggestion
You might notice that `Text` in this code example is enclosed by square (`[ ]`) brackets. By itself, `Text` represents a (UTF-8 encoded) sequence of Unicode characters. Placing square brackets around a type describes an **array** of that type. In this context, therefore, `[Text]` indicates an array of text values, enabling the program to accept multiple text values as an array.

```

doc/md/writing-motoko/arguments.md
```diff
@@ -0,0 +1,131 @@
...
+actor {
+
+public func location(cities : [Text]) : async Text {
+    return "Hello, from " # (debug_show cities) # "!";
+};
+
+public func location_pretty(cities : [Text]) : async Text {
+    var str = "Hello from ";
+    for (city in cities.vals()) {
+    str := str # city #", ";
```
crusso, line 83:
```suggestion
        str := str # city #", ";
```

doc/md/writing-motoko/arguments.md
crusso, line 1: This doc is more about using dfx to call motoko than anything else. Perhaps it should go in a separate section (or at the end of getting-started/quickstart.md

doc/md/writing-motoko/arguments.md
```diff
@@ -0,0 +1,131 @@
...
+public func apply<A, B>(fs : [A -> B], xs : [A]) : [B] {
+    var ys : [B] = [];
+    for (f in fs.vals()) {
+        ys := append<B>(ys, map<A, B>(f, xs));
+    };
+    ys;
+};
+```
+
+For information about the functions that perform operations on arrays, see the description of the Array module in the Motoko base library or the **Motoko programming language reference**. For another example focused on the use of arrays, see the [quick sort](https://github.com/dfinity/examples/tree/master/motoko/quicksort) project in the [examples](https://github.com/dfinity/examples/) repository.
```
crusso, line 105:
```suggestion
For information about the functions that perform operations on arrays, see the description of the [Array module](../base/Array.md) in the Motoko base library or the **Motoko programming language reference**. For another example focused on the use of arrays, see the [quick sort](https://github.com/dfinity/examples/tree/master/motoko/quicksort) project in the [examples](https://github.com/dfinity/examples/) repository.
```

doc/md/writing-motoko/incomplete-code.md
```diff
@@ -0,0 +1,48 @@
+---
+sidebar_position: 27
+---
+# Writing incomplete code
+
+## Overview
+
+In the midst of writing a program, you may want to run an incomplete version or a version where one or more execution paths are either missing or simply invalid.
+
+To accommodate these situations, Motoko uses the `xxx`, `nyi` and `unreachable` functions from the base `Prelude` library, explained below. Each is a simple wrapper around a more [general trap mechanism](traps.md).
```
crusso, line 10:
```suggestion
To accommodate these situations, Motoko provides the `xxx`, `nyi` and `unreachable` functions from the base `Prelude` library, explained below. Each is a simple wrapper around a more [general trap mechanism](traps.md).
```

doc/md/canister-maintenance/compatibility.md
```diff
@@ -187,9 +148,19 @@ cannot be consumed at new type
   var Nat
 ```
 
-## Examples in the wild
+Upgrades from `v2.wasm` to `v3.wasm` would fail and roll-back, avoiding data loss. If Candid is revised, an upgrade would now "succeed", but with data loss. This is the difference between a fail safe and a silent failure.
```
crusso, line 151:
```suggestion
Because of the compatibility error, you should not attempt to upgrade from `v2.wasm` to `v3.wasm`. The result of upgrading is unpredictable. At best, the upgrade will detect the incompatibility, trap and roll back to the current version, as if the upgrade had never been attempted. At worse, the upgrade will appear to succeed, but lose some or all of the state of the previous version, re-initializing some of the stable variables you intended to preserve.
```

doc/md/writing-motoko/incomplete-code.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Short-term holes
+
+Short-term holes are never committed to a source repository and only ever exist in a single development session for a developer that is still writing the program.
```
crusso, line 14:
```suggestion
```

doc/md/writing-motoko/incomplete-code.md
```diff
@@ -0,0 +1,48 @@
...
+import P "mo:base/Prelude";
+```
+
+The developer can fill any missing expression with the following:
+
+``` motoko include=prelude
+P.xxx()
+```
+
+The result will always type check at compile time, and always trap at run time, if and when this expression executes.
```
crusso, line 28: The expression `xxx()` has type `None`, which is a subtype of every other type. This means the expression can plug any short-term gap in your code. 
`xxx()` will trap when executed.

doc/md/writing-motoko/incomplete-code.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Long-term holes
+
+By convention, longer-term holes can be considered "not yet implemented" (`nyi`) features, and marked as such with a similar function from the Prelude module:
+
+``` motoko include=prelude
+P.nyi()
+```
+
```
crusso, line 37: Like `xxx()`, `nyi()` has type `None`, thus any other type, and will trap on execution.

crusso, line 37:
```suggestion
Like `xxx()`, `nyi()` has type `None`, thus any other type, and will trap on execution.
```

doc/md/writing-motoko/incomplete-code.md
```diff
@@ -0,0 +1,48 @@
...
+## `unreachable` code paths
+
+In contrast to the situations above, sometimes code will never be filled. It will never be evaluated assuming the coherence of the internal logic of the programs' invariants.
```
crusso, line 40:
```suggestion
Sometime a developer will be required to provide code that they know will never be executed, due to the preceding program logic.
```

doc/md/writing-motoko/incomplete-code.md
```diff
@@ -0,0 +1,48 @@
...
+
+To document a code path as logically impossible, or unreachable, use the base library function `unreachable`:
```
crusso, line 42:
```suggestion
To document unreachable code, the developer can use the  `unreachable` function from the prelude:
```

doc/md/writing-motoko/incomplete-code.md
```diff
@@ -0,0 +1,48 @@
...
+
+``` motoko include=prelude
+P.unreachable()
+```
+
+As in the situations above, this function type-checks in all contexts, and when evaluated, traps in all contexts.
```
crusso, line 48:
```suggestion
Like `unreachable()` has type `None` and thus any other type, and will trap on (unexpected!) execution.
```

doc/md/writing-motoko/integers.md
```diff
@@ -0,0 +1,60 @@
...
+# Integers
+
+
+## Overview
+
+Integers `Int` are primitive values that have bounded and unbounded variants. Integers can be negative numbers such as `-2` or `-1`, or positive numbers such as `1` and `2`.
+
+Integers are unbounded and do not overflow. They use representations that grow to accommodate any finite number.
+
+This page will provide an example calculator program that creates a single actor with several public entry-point functions to perform basic arithmetic operations using integers.
```
crusso, line 14:
```suggestion
Motoko offers a variety types to represent integers and natural numbers, with the usual suite of arithmetic operators (`+`, `-`, `*`, '/' etc.) and comparison operators (`==`, `!=`,  `<`, `>`, `<=`, `>=`).

The types `Int` and `Nat` are unbounded, meaning their values can grow to arbitrary size (limited only by memory). 

The type `Nat` is a subtype of `Int`, so you can always supply a value of type `Nat` where and `Int` is expected (but not vice versa).

Motoko also provides bounded, or fixed-size integers and naturals, using a suffix to indicate the size of the type, in bits. 

Thus `Int8`, `Int16` and `Int32` and `Int64` are 8-, 16-, 32- and 64-bit integer types, while `Nat8`, `Nat16`, `Nat32`, and `Nat64` are 8-, 16-, 32- and 64-bit natural types.

An arithmetic operation on a value of a fixed-size type will trap if its result exceeds the bounds of the fixed-size type, either due to overflow or underflow.
For example, `255 : Nat8 + 3` traps, because 258 is too large for a `Nat8`.

Wrapping, non-trapping, versions of the usual arithmetic operations, performing modular arithmentic, are available by suffixing the usual operator with a `%`. For example, `255 : Nat8 +% 3` evaluates to `2`.

The type of an integer or natural constant is determined by the textual context
of the constant, defaulting to `Int` for negative constants and `Nat` for positive ones. 
Otherwise, the type of a constant can be indicated using a type annotation, for example `0 : Nat8`.

Once can force a nonnegative constant to be interpreted as an `Int` using an explicit sign, for example  `+1` is the `Int` one.

For convenience, in-place updates of a variable or array element can be written by  using a compound assignment operator, combining an arithmetic operation with  the assignment operator `=` , e.g. `x += 1` is short-hand for `x = x + 1` and combines `+` with `=`.

Motoko does not provide any implicit conversions between numeric types. Instead, base library functions like `Nat8.toNat`  and `Nat8.fromNat` should be used for explicit conversion.
```

crusso, line 14:
```suggestion
Integers `Int` are primitive values that have bounded and unbounded variants. Integers can be negative numbers such as `-2` or `-1`, or positive numbers such as `1` and `2`.

Integers are unbounded and do not overflow. They use representations that grow to accommodate any finite number.

To illustrate working with numbers, here is calculator program that creates a single actor with several public entry-point functions to perform basic arithmetic operations using integers. 
```

doc/md/writing-motoko/integers.md
```diff
@@ -0,0 +1,60 @@
+---
+sidebar_position: 10
+---
+
+# Integers
```
crusso, line 5:
```suggestion
# Integers
```
```suggestion
# Integers
```
```suggestion
# Integers and Naturals
```
Or maybe just numbers?

doc/md/writing-motoko/integers.md
```diff
@@ -0,0 +1,60 @@
...
+  // Clear the calculator and reset to zero
+  public func clearall() : async Int {
+    if (cell : Int != 0)
+      cell -= cell;
+    return cell
+  };
+ };
+```
+
+You might notice that this sample code uses integer (`Int`) data types, enabling you to use positive or negative numbers. If you wanted to restrict the functions in this calculator code to only use positive numbers, you could change the data type to only allow natural (`Nat`) data.
```
crusso, line 47: Do we use `you` in the Motoko doc?

doc/md/writing-motoko/intercanister-calls.md
```diff
@@ -0,0 +1,73 @@
...
+
+```motoko
+import Prim "mo:prim";
+
+actor {
+    public func getValue() : async Nat {
+        Prim.debugPrint("Hello from canister 2!");
+        return 10;
+    };
+};
```
crusso, line 41:
```suggestion
import Debug "mo:base/Debug";

actor {
    public func getValue() : async Nat {
        Debug.print("Hello from canister 2!");
        return 10;
    };
};
```
We should not encourge use of mo:prim in examples....

doc/md/writing-motoko/intercanister-calls.md
```diff
@@ -0,0 +1,73 @@
...
+
+For example, you might want to create a dapp for professional networking, organizing community events, or hosting fundraising activities. Each of these dapps might have a social component that enables users to identify social relationships based on some criteria or shared interest, such as friends and family or current and former colleagues.
+
+To address this social component, you might want to create a single canister for storing user relationships then write your professional networking, community organizing, or fundraising application to import and call functions that are defined in the canister for social connections. You could then build additional applications to use the social connections canister or extend the features provided by the social connections canister to make it useful to an even broader community of other developers.
+
+This example will showcase a simple way to configure inter-canister calls that can be used as the foundation for more elaborate projects and use-cases such as those mentioned above.
+
+## Example
```
crusso, line 17: This example could be better. Maybe add an issue for that?

doc/md/writing-motoko/local-objects-classes.md
```diff
@@ -1,30 +1,35 @@
...
+# Objects and classes
 
 <!--
 TODO: Move examples into doc/modules/language-guide/examples
 -->
 
-In Motoko, an `object` may encapsulate local state (`var`-bound variables) by packaging this state with `public` methods that access and update it.
+## Overview
+
+Motoko programs benefit from the ability to encapsulate state as objects with abstract types, as an `object` may encapsulate local state as `var`-bound variables by packaging this state with `public` methods that access and update it.
```
crusso, line 13:
```suggestion
Motoko programs benefit from the ability to encapsulate state as objects with abstract types. The [mutable state](mutable-state.md) introduces declarations of  mutable state in the form of `var`-declared variables.  Using such declarations privately in its body, an object can encapsulate the state, declaring public methods that access and update it.
```

doc/md/writing-motoko/local-objects-classes.md
```diff
@@ -1,30 +1,35 @@
...
 
-As in other typed languages, Motoko programs benefit from the ability to encapsulate state as objects with abstract types.
+Motoko objects that include mutable state are not shareable as a critical security-oriented design decision. If they were shareable, that would expose two main security risks: conceptually moving a mobile object’s code among actors and executing it remotely, and sharing state with remote logic.
```
crusso, line 15:
```suggestion
By design, objects with mutable field or methods cannot be sent to remote actors. If that were allowed, a receiver would either have to receive a remote reference to the local object, breaking the isolation of the actor model by allowing remote updates to local state. Or the receiver would have to receive a copy of the local object: then, the effect of any changes to the copy would not be reflected in the original, leading to confusion.
```

doc/md/writing-motoko/local-objects-classes.md
```diff
@@ -1,30 +1,35 @@
...
 
-However, Motoko objects that include mutable state are *not shareable*, and this is a critical security-oriented design decision.
+However, objects may be pure records which are shareable since they are free from mutable state. The [mutable state](mutable-state.md) introduces declarations of private mutable state in the form of `var`-bound variables and mutable array allocation.
```
crusso, line 17:
```suggestion
```

doc/md/writing-motoko/local-objects-classes.md
```diff
@@ -1,30 +1,35 @@
...
+**Object classes** : A family of related objects to perform a task with a customizable initial state. Motoko provides a syntactical construct, called a `class` definition, which simplifies building objects of the same type and implementation.
 
-**Object classes**. Frequently, one needs *a family* of related objects to perform a task. When objects exhibit similar behavior, it makes sense to fabricate them according to the same blueprint, but with customizable initial state. To this end, Motoko provides a syntactical construct, called a `class` definition, which simplifies building objects of the same type and implementation. We introduce these after discussing objects.
+**Actor classes** : An object class that exposes a [service](async-data.md) using asynchronous behavior. The corresponding Motoko construct is an [actor class](actor-classes.md), which follows a similar but distinct design.
```
crusso, line 26: Maybe this should go to the end, after the section on classes.

doc/md/writing-motoko/local-objects-classes.md
```diff
@@ -56,15 +59,13 @@ This object `counter` has the following *object type*, written as a list of fiel
...
 
 In the declaration of `object`, the variable `count` was explicitly declared neither as `public` nor as `private`.
 
-By default, all declarations in an object block are `private`, as is `count` here. Consequently, the type for `count` does not appear in the type of the object, *and* its name and presence are both inaccessible from the outside.
+By default, all declarations in an object block are `private`. Consequently, the type for `count` does not appear in the type of the object. Its name and presence are both inaccessible from the outside.
 
-The inaccessibility of this field comes with a powerful benefit: By not exposing this implementation detail, the object has a *more general* type (fewer fields), and as a result, is interchangeable with objects that implement the same counter object type differently, without using such a field.
-
-## Example: The `byteCounter` object
+By not exposing this implementation detail, the object has a more general type with fewer fields, and as a result, is interchangeable with objects that implement the same counter object type differently without using such a field.
```
crusso, line 68:
```suggestion
By not exposing this implementation detail, the object has a more general type with fewer fields, and as a result, is interchangeable with objects that implement the same counter object type differently without using such a field.
```
```suggestion
By not exposing this implementation detail, the object has a more general type with fewer fields, and as a result, is interchangeable with objects that have the same interface but a different implementation.
```

doc/md/writing-motoko/local-objects-classes.md
```diff
@@ -88,83 +89,15 @@ This object has the same type as the previous one, and thus from the standpoint
...
-Here, we consider a counter named `fullCounter` with a less general type than any given above. In addition to `inc`, `read` and `bump`, it additionally includes `write`, which permits the caller to change the current count value to an arbitrary one, such as back to `0`.
-
-**Object subtyping.** In Motoko, objects have types that may relate by subtyping, as the various types of counters do above. As is standard, types with *more fields* are *less general* (are ***sub**types* of) types with *fewer fields*. For instance, we can summarize the types given in the examples above as being related in the following subtyping order:
-
--   Most general:
-
-``` motoko no-repl
-{ bump : () -> Nat }
-```
+A common type shared among two implementations of an object or service affords the potential for the internal implementation complexity to be factored away from the rest of the application.
```
crusso, line 98:
```suggestion
The common type abstracts the differences in the implementations of the objects, shielding the rest of the application from their implementation details.
 ```

File: doc/md/writing-motoko/incomplete-code.md (hunk 19 above)
Comment (crusso) [inline doc/md/writing-motoko/incomplete-code.md:28]:
```suggestion
The expression `xxx()` has type `None`, which is a subtype of every other type. This means the expression can plug any short-term gap in your code. 
`xxx()` will trap when executed.
```

crusso (commented):
More interim comments...

doc/md/writing-motoko/local-objects-classes.md
```diff
@@ -88,83 +89,15 @@ This object has the same type as the previous one, and thus from the standpoint
...
-  read : () -> Nat ;
-  bump : () -> Nat ;
-  write : Nat -> () ;
-}
-```
-
-If a function expects to receive an object of the first type (`{ bump: () → Nat }`), *any* of the types given above will suffice, since they are each equal to, or a subtype of, this (most general) type.
-
-However, if a function expects to receive an object of the last, least general type, the other two will *not* suffice, since they each lack the needed `write` operation, to which this function rightfully expects to have access.
+Objects can also have [subtypes](object-subtyping.md).
```
crusso, line 100:
```suggestion
Objects can also have [subtypes](object-subtyping.md).
```
```suggestion
Objects types can also have [subtypes](object-subtyping.md), allowing an object with a more specific type to pass as an object of a more general type (for example, to pass as an object with fewer fields).
```

doc/md/getting-started/motoko-introduction.md
jessiemongeon1: Bitfinity: https://internetcomputer.org/docs/current/developer-docs/backend/solidity/

doc/md/attic/motoko-grammar.md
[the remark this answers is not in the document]
jessiemongeon1: Moved to reference section

doc/md/canister-maintenance/compatibility.md
```diff
@@ -215,39 +186,3 @@ actor {
...
-
-Motoko embeds `.did` and `.most` files as wasm *custom sections*, for use by other tools, e.g. dfx.
-
-In future, `dfx canister upgrade` will, by default:
-
-1.  query the IC for a canister’s dual interfaces,
-
-2.  check compatibility of the installed and new binary,
-
-3.  abort the upgrade when unsafe.
```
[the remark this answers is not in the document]
jessiemongeon1: Added section to the end and updated to the correct dfx command

doc/md/canister-maintenance/compatibility.md
```diff
@@ -1,162 +1,120 @@
...
-Unfortunately, when we upgrade this counter (say with itself), its state is lost.
-
-|         |       |         |             |
-|---------|-------|---------|-------------|
-| version | state | success | call        |
-| v0      | 0     | ✓       | inc()       |
-| v0      | 1     | ✓       | inc()       |
-| v0      | 2     | ✓       | upgrade(v0) |
-| v0      | *0*   | *✗*     | inc()       |
-| v0      | 1     |         |             |
```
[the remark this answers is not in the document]
jessiemongeon1: Yes, I think they can be removed.

doc/md/getting-started/motoko-introduction.md
jessiemongeon1: Removed this section

doc/md/getting-started/motoko-introduction.md
jessiemongeon1: For an introduction, including examples isn't ideal. It should be a high level overview of the language's features and design, not showcasing code snippets. We could create a second page that introduces those examples as a follow-up page to this introduction?

doc/md/writing-motoko/arguments.md
jessiemongeon1: Removed the majority of the dfx calls to keep the focus on the code examples

doc/md/writing-motoko/local-objects-classes.md
jessiemongeon1: Moved further now

doc/md/writing-motoko/actors-async.md
```diff
@@ -1,66 +1,58 @@
-# Actors and async data
+---
+sidebar_position: 1
+---
 
-The programming model of the Internet Computer consists of memory-isolated canisters communicating by asynchronous message passing of binary data encoding Candid values. A canister processes its messages one-at-a-time, preventing race conditions. A canister uses call-backs to register what needs to be done with the result of any inter-canister messages it issues.
+# Actors
 
-Motoko abstracts the complexity of the Internet Computer with a well known, higher-level abstraction: the *actor model*. Each canister is represented as a typed actor. The type of an actor lists the messages it can handle. Each message is abstracted as a typed, asynchronous function. A translation from actor types to Candid types imposes structure on the raw binary data of the underlying Internet Computer. An actor is similar to an object, but is different in that its state is completely isolated, its interactions with the world are entirely through asynchronous messaging, and its messages are processed one-at-a-time, even when issued in parallel by concurrent actors.
```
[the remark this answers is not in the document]
jessiemongeon1: Retaining these paragraphs

doc/md/writing-motoko/actors-async.md
```diff
@@ -1,66 +1,58 @@
...
+- Waiting on `f` to be completed is expressed using `await f` to obtain a value of type `T`. To avoid introducing shared state through messaging, for example, by sending an object or mutable array, the data that can be transmitted through shared functions is restricted to immutable, shared types.
 
--   function `bump()` asynchronously increments and reads the counter.
+## Defining an actor
 
-Shared functions, unlike local functions, are accessible to remote callers and have additional restrictions: their arguments and return value must be *shared* types - a subset of types that includes immutable data, actor references, and shared function references, but excludes references to local functions and mutable data. Because all interaction with actors is asynchronous, an actor’s functions must return futures, that is, types of the form `async T`, for some type `T`.
+Consider the following actor declaration:
 
-The only way to read or modify the state (`count`) of the `Counter` actor is through its shared functions.
+```motoko
```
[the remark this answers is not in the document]
jessiemongeon1: Reverted to original example

doc/md/writing-motoko/actors-async.md
```diff
@@ -1,66 +1,58 @@
...
-For now, the only place shared functions can be declared is in the body of an actor or actor class. Despite this restriction, shared functions are still first-class values in Motoko and can be passed as arguments or results, and stored in data structures.
+- Instead of a `print` function, this sample program defines an `actor` with a public greet function that takes a `name` argument with a type of `Text`.
 
-The type of a shared function is specified using a shared function type. For example, the value `inc` has type `shared () → async Nat` and could be supplied as a standalone callback to some other service (see [publish-subscribe](sharing.md) for an example).
+- The program uses the `async` keyword to indicate that it returns an asynchronous message consisting of a concatenated text string constructed using `"Hello, "`, the `#` operator, the `name` argument, and `"!"`.
 
 ## Actor types
 
-Just as objects have object types, actors have *actor types*. The `Counter` actor has the following type:
+Just as objects have object types, actors have actor types. The `Counter` example above has the following type:
```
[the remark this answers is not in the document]
jessiemongeon1: Reverted to original example

doc/md/getting-started/basic-concepts.md
jessiemongeon1: Created an issue: https://github.com/dfinity/motoko/issues/4514

doc/md/getting-started/motoko-introduction.md
jessiemongeon1: Created an issue: https://github.com/dfinity/motoko/issues/4515

doc/md/writing-motoko/intercanister-calls.md
jessiemongeon1: Created an issue: https://github.com/dfinity/motoko/issues/4516

doc/md/getting-started/motoko-introduction.md
crusso: Makes sense. In that case, maybe we should use attic/motoko.md (with the examples) as the more detailed intro to the writing-motoko section?

doc/md/writing-motoko/incomplete-code.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Long-term holes
+
+By convention, longer-term holes can be considered "not yet implemented" (`nyi`) features, and marked as such with a similar function from the Prelude module:
+
+``` motoko include=prelude
+P.nyi()
+```
+Like `xxx()`, `nyi()` has type `None`, thus any other type, and will trap on execution.
+## `unreachable` code paths
```
crusso, line 38:
```suggestion
## Unreachable code

```

Q: the page starts with `you` but then talks about the developer - should it stick with `you` or is this switching ok? Up to you.

doc/md/writing-motoko/message-restrictions.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Overview
+
+ICP places restrictions on when and how canisters are allowed to communicate. These restrictions are enforced dynamically but prevented statically in Motoko, ruling out a class of dynamic execution errors. Two examples are:
+
+-   Canister installation can execute code, but not send messages.
+
+-   A canister query method cannot send messages.
+
+These restrictions are applied in Motoko through which certain expressions can be used.
```
crusso, line 15:
```suggestion
These ICP restrictions are surfaced in Motoko as static constraints on the context in which certain expressions can be used and are enforced by the type system.
```

doc/md/writing-motoko/message-restrictions.md
```diff
@@ -0,0 +1,25 @@
...
+
+In Motoko, an expression occurs in an asynchronous context if it appears in the body of an `async` expression, which may be the body of a shared or local function or a stand-alone expression. The only exception are `query` functions, whose body is not considered to open an asynchronous context.
+
+In Motoko calling a shared function is an error unless the function is called in an asynchronous context. Calling a shared function from an actor class constructor is also an error.
+
+The `await` and `async` constructs are only allowed in an asynchronous context.
+
+It is only possible to `throw` or `try/catch` errors in an asynchronous context. This is because structured error handling is supported for messaging errors only and, like messaging itself, confined to asynchronous contexts.
+
+These rules also mean that local functions cannot directly call shared functions or `await` futures.
```
crusso, line 25:
```suggestion
These rules also mean that local functions cannot directly call shared functions or `await` futures.

(The same restrictions apply to `await*` and `async*` expressions.)
```

doc/md/writing-motoko/incomplete-code.md
jessiemongeon1: Good catch, it should all be `you`. Updated to fix.

doc/md/writing-motoko/traps.md
```diff
@@ -0,0 +1,36 @@
+---
+sidebar_position: 26
+---
+
+# Traps
```
crusso, line 5:
```suggestion
# Traps and commit points
```

doc/md/writing-motoko/traps.md
crusso, line 1: Really, this section is about the interplay between shared functions, await, traps and commit points. I wonder if its original place, in the section on actors-async, together with await - wasn't a better home.

doc/md/writing-motoko/writing-intro.md
crusso, line 1: Thanks for moving this!

doc/md/writing-motoko/pattern-matching.md
```diff
@@ -28,48 +57,40 @@ switch ("Adrienne", #female) {
...
 
-terminology  
-The (usually structured) expression that is being matched is frequently called the *scrutinee* and the patterns appearing behind the keyword `case` are the *alternatives*. When every possible scrutinee is matched by (at least one) alternative, then we say that the scrutinee is *covered*. The patterns are tried in top-down fashion and thus in case of *overlapping* patterns the one higher-up is selected. An alternative is considered *dead* (or *inactive*), if for every value that it matches there is higher-up alternative that is also matched.
+### Terminology
+
+The usually structured expression that is being matched is frequently called the **scrutinee**. The patterns appearing behind the keyword `case` are the **alternatives**. When every possible scrutinee is matched by at least one alternative, then we say that the scrutinee is **covered**. The patterns are tried in top-down fashion and thus in case of overlapping patterns, the one higher-up is selected. An alternative is considered dead or inactive if for every value that it matches there is higher-up alternative that is also matched.
```
crusso, line 72:
```suggestion

The expression whose value is being matched is frequently called the **scrutinee**. The patterns appearing behind the keyword `case` are the **alternatives**. When every possible value of the scrutinee is matched by at least one alternative, then we say that the scrutinee is **covered**. The alternatives are tried in order; in case of overlapping patterns, the earlier one is selected. An alternative is considered dead or redundant if for every value that it matches there is already some earlier alternative that also matches the value.
```

doc/md/writing-motoko/pattern-matching.md
crusso, line 1: This section could be improved, both the prose and the range of examples (e.g. variants and options are missing). See issue https://github.com/dfinity/motoko/issues/4519

crusso (commented):
Ok, I think I'm done for now. I think there's too many issues to address in one PR, so lets get this one in and improve bits later.

doc/md/writing-motoko/traps.md
jessiemongeon1: Moved it back to actors-async

jessiemongeon1 (approved):

doc/md/reference/language-manual.md
```diff
@@ -2610,17 +2619,17 @@ The Candid encoding of a value as a blob is not unique and the same value may ha
 
 The declaration expression `<dec>` has type `T` provided the declaration `<dec>` has type `T`.
 
-Evaluating the expression `<dec>` proceeds by evaluating `<dec>`, returning the result of `<dec>` but discarding the bindings introduced by `<dec>` (if any).
+Evaluating the expression `<dec>` proceeds by evaluating `<dec>`, returning the result of `<dec>` but discarding the bindings introduced by `<dec>`, if any.
 
-(The expression `<dec>` is actually shorthand for the block expression `{ <dec> }`.)
+The expression `<dec>` is actually shorthand for the block expression `{ <dec> }`.
```
crusso, line 2624:
```suggestion
The expression `<dec>` is actually shorthand for the block expression `{ <dec> }`.
```
```suggestion
The expression `<dec>` is actually shorthand for the block expression `do { <dec> }`.
```

doc/md/reference/language-manual.md
```diff
@@ -2610,17 +2619,17 @@ The Candid encoding of a value as a blob is not unique and the same value may ha
...
 
 ### Ignore
 
 The expression `ignore <exp>` has type `()` provided the expression `<exp>` has type `Any` .
 
-The expression `ignore <exp>` evaluates `<exp>` (typically for some side-effect) but discards its value.
+The expression `ignore <exp>` evaluates `<exp>`, typically for some side-effect, but discards its value.
 
-`Ignore` is useful for evaluating an expression within a sequence of declarations when that expression has non-`unit` type (and the simpler `<exp>` declaration would be ill-typed). Then the semantics is equivalent to `let _ = <exp> : Any`.
+`Ignore` is useful for evaluating an expression within a sequence of declarations when that expression has non-`unit` type, and the simpler `<exp>` declaration would be ill-typed. Then the semantics is equivalent to `let _ = <exp> : Any`.
```
crusso, line 2632:
```suggestion
The `ignore` declaration is useful for evaluating an expression within a sequence of declarations when that expression has non-`unit` type, and the simpler `<exp>` declaration would be ill-typed. Then the semantics is equivalent to `let _ = <exp> : Any`.
```

doc/md/reference/language-manual.md
```diff
@@ -2362,43 +2373,43 @@ The evaluation of `break <id> <exp>` evaluates `<exp>` to some result `r`. If `r
...
 
 The expression `return` is equivalent to `return ()`.
 
 The expression `return <exp>` has type `None` provided:
 
--   `<exp>` has type `T` and
+-   `<exp>` has type `T`.
 
--   `T` is the return type of the nearest enclosing function (with no intervening `async` expression), or
+-   `T` is the return type of the nearest enclosing function with no intervening `async` expression.
```
crusso, line 2386:
```suggestion
-  and either one of:

   -  `T` is the return type of the nearest enclosing function with no intervening `async` expression.

   -  `async T` is the type of the nearest enclosing, perhaps implicit, `async` expression with no intervening function declaration. 
```

doc/md/reference/language-manual.md
```diff
@@ -2362,43 +2373,43 @@ The evaluation of `break <id> <exp>` evaluates `<exp>` to some result `r`. If `r
...
 
--   `async T` is the type of the nearest enclosing (perhaps implicit) `async` expression (with no intervening function declaration)
+-   `async T` is the type of the nearest enclosing, perhaps implicit, `async` expression with no intervening function declaration.
```
crusso, line 2388:
```suggestion
```

crusso (commented):
Took a pass through language-manual.md

doc/md/reference/language-manual.md
```diff
@@ -446,80 +453,80 @@ The *visibility* qualifier `<vis>?` determines the accessibility of every field
...
 
--   A stability qualifier *should* appear on `let` and `var` declarations that are actor fields. An absent stability qualifier defaults to `flexible`.
+-   A stability qualifier should appear on `let` and `var` declarations that are actor fields. An absent stability qualifier defaults to `flexible`.
 
 -   `<stab>` qualifiers must not appear on fields of objects or modules.
 
 -   The pattern in a `stable let <pat> = <exp>` declaration must be *simple* where, a pattern `pat` is simple if it (recursively) consists of
 
-    -   a variable pattern `<id>`, or
+    -   A variable pattern `<id>`, or
```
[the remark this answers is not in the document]
jessiemongeon1: It is a style choice rather than a grammar choice, and we use that style across the rest of the Dev Docs. I'll revise the sentence if that helps.

doc/md/reference/language-manual.md
```diff
@@ -134,19 +139,19 @@ The 'e' (or 'E') prefixes a base 10, decimal exponent; 'p' (or 'P') prefixes a b
 
 :::note
 
-the use of decimal notation, even for the base 2 exponent, is in keeping with the established hexadecimal floating point literal syntax of the `C` language.
+The use of decimal notation, even for the base 2 exponent, is keeping with the established hexadecimal floating point literal syntax of the `C` language.
```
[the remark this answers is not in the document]
jessiemongeon1: Revised to `adheres to`

doc/md/reference/language-manual.md
```diff
@@ -321,7 +326,7 @@ The following table defines the relative precedence and associativity of operato
 
 ### Programs
 
-The syntax of a *program* `<prog>` is as follows:
+The syntax of a program `<prog>` is as follows:
```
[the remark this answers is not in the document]
jessiemongeon1: Replacing italics with bold text

doc/md/reference/language-manual.md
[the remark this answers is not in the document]
jessiemongeon1: Revised sentence

doc/md/canister-maintenance/upgrades.md
```diff
@@ -1,39 +1,28 @@
...
+## Overview
 
-The Internet Computer’s persistence model allows a canister to save and restore such data to dedicated *stable memory* that, unlike ordinary canister memory, is retained across an upgrade, allowing a canister to transfer data in bulk to its replacement canister.
+One key feature of ICP is its ability to persist canister state using WebAssembly memory and globals rather than a traditional database. This means that the entire state of a canister is restored before and saved after each message, without explicit user instruction.
 
-For applications written in Motoko, the language provides high-level support for preserving state that leverages Internet Computer stable memory. This higher-level feature, called *stable storage*, is designed to accommodate changes to both the application data and to the Motoko compiler used to produce the application code.
+A canister saves and restores data to dedicated **stable memory** that, unlike ordinary canister memory, is retained across an upgrade and allows a canister to transfer data in bulk across upgrades.
```
crusso, line 11:
```suggestion
```

doc/md/canister-maintenance/upgrades.md
```diff
@@ -1,39 +1,28 @@
...
+---
 
-Though convenient, orthogonal persistence poses a challenge when it comes to upgrading the code of a canister. Without an explicit representation of the canister’s state, how does one transfer any application data from the retired canister to its replacement?
+# Stable variables and upgrade methods
 
-Accommodating upgrades without data loss requires some new facility to *migrate* a canister’s crucial data to the upgraded canister. For example, if you want to deploy a new version of a user-registration canister to fix an issue or add functionality, you need to ensure that existing registrations survive the upgrade process.
+## Overview
 
-The Internet Computer’s persistence model allows a canister to save and restore such data to dedicated *stable memory* that, unlike ordinary canister memory, is retained across an upgrade, allowing a canister to transfer data in bulk to its replacement canister.
+One key feature of ICP is its ability to persist canister state using WebAssembly memory and globals rather than a traditional database. This means that the entire state of a canister is restored before and saved after each message, without explicit user instruction.
```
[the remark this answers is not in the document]
crusso: @jessiemongeon1 perhaps I wasn't clear in my comment [here](https://github.com/dfinity/motoko/pull/4494#discussion_r1572671604) but I think we've lost too much information and I'd like more of the text restored, please.

PR #4522, opened 2024-04-30
ci(Mergify): configuration update to use macos-12 not macos-latest

crusso:
This change has been made by @crusso from Mergify config editor.


jessiemongeon1 (approved):

ggreif:
The `macOS` comparison error seems to come from
```
$ git grep "stdout: '10_000"
test/test-moc_interpreter.js:    stdout: '10_000 : Nat\n'
```
My hunch is that by un-bumping MacOS an older `node` got selected? But node-js is supposed to come from `nix`, so I am somewhat confused...

PR #4534, opened 2024-05-07
chore: update base doc

crusso:
I think @jessiemongeon1 will update the portal today so may as well update the doc too.


luc-blaeser (approved):
Thanks for updating the documentation.

jessiemongeon1 (approved):

PR #4538, opened 2024-05-10
docs: Add Randomness tutorial

jessiemongeon1:


timohanke:
It seems that in this function we don't get the full range all the way to max. I mean not even the open interval [0,max).
```
  func chooseMax(f : Random.Finite, max : Nat) : ? Nat {
    assert max > 0;
    do ? {
      if (max == 1) return ? 0;
      var k = bit(f.coin()!);
      var n = max / 2;
      while (n > 1) {
        k := k * 2 + bit(f.coin()!);
        n := n / 2;
      };
      if (k < max)
        return ? k
      else chooseMax(f, max)!; // retry
    };
  };
  ```
For example say max = 3. Then n = 1, the while loop won't run at all, so we get k = 0 or 1 , which is the interval [0,2) not [0,3).

**Comment by timohanke** [comment]
A different remark: I think we should make clear what requires asynchronous randomness and what doesn't. Asynchronous randomness must be retrieved _after_ bets are closed (we made that clear). But what if we try to shuffle the deck and the entropy is exhausted. Do we then really want to tell the developers they have to get new asynchronous randomness? That's bad for latency and unnecessary. They can extend the one entropy that they already retrieved asynchronously after the bets were closed and feed it into a deterministic PRNG. Since the bets are closed it's ok that this is deterministic. There is no point in getting the next entropy (if additional entropy is required) asynchronously again. And the asynchronous entropy isn't cryptographically "better" than the one from a PRNG either.

I would feed the entropy into sha2 for example to get an infinite stream. Then the shuffling will always complete in one execution round.

**Comment by crusso** [comment]
> It seems that in this function we don't get the full range all the way to max. I mean not even the open interval [0,max).
> 
> ```
>   func chooseMax(f : Random.Finite, max : Nat) : ? Nat {
>     assert max > 0;
>     do ? {
>       if (max == 1) return ? 0;
>       var k = bit(f.coin()!);
>       var n = max / 2;
>       while (n > 1) {
>         k := k * 2 + bit(f.coin()!);
>         n := n / 2;
>       };
>       if (k < max)
>         return ? k
>       else chooseMax(f, max)!; // retry
>     };
>   };
> ```
> 
> For example say max = 3. Then n = 1, the while loop won't run at all, so we get k = 0 or 1 , which is the interval [0,2) not [0,3).

Sorry, Jessie just stole my broken code. Without switching my brain on, is it enough to change the loop guard to
```
while (n > 0) { ... }
```
?

**Comment by crusso** [comment]
> A different remark: I think we should make clear what requires asynchronous randomness and what doesn't. Asynchronous randomness must be retrieved _after_ bets are closed (we made that clear). But what if we try to shuffle the deck and the entropy is exhausted. Do we then really want to tell the developers they have to get new asynchronous randomness? That's bad for latency and unnecessary. They can extend the one entropy that they already retrieved asynchronously after the bets were closed and feed it into a deterministic PRNG. Since the bets are closed it's ok that this is deterministic. There is no point in getting the next entropy (if additional entropy is required) asynchronously again. And the asynchronous entropy isn't cryptographically "better" than the one from a PRNG either.
> 
> I would feed the entropy into sha2 for example to get an infinite stream. Then the shuffling will always complete in one execution round.

Hmm, that suggests that our random maze example could also be improved (and, incidentally, might be suffering from the same bug). The maze generator keeps asking for more entropy when it runs out of bits too.

**Comment by timohanke** [comment]
> `n > 0`

No, but using `max - 1`. The number of coins drawn should equal `bitlength(max - 1)`.

```
  import Random "mo:base/Random";
  
  func bit(b : Bool) : Nat = if(b) 1 else 0;
  
  func chooseMax(f : Random.Finite, max : Nat) : ? Nat {
    assert max > 0;
    do ? {
      var n = max - 1 : Nat;
      var k = 0;
      while (n != 0) {
        k *= 2;
        k += bit(f.coin()!);
        n /= 2;
      };
      k;
    };
  };
  
  let f = Random.Finite("\0a");
  
  chooseMax(f, 256);
  ```

jessiemongeon1:
@crusso Could you suggest some edits to address @timohanke 's feedback?

doc/md/writing-motoko/randomness.md
```diff
@@ -0,0 +1,127 @@
...
+      var n = max / 2;
+      while (n > 1) {
+        k := k * 2 + bit(f.coin()!);
+        n := n / 2;
+      };
+      if (k < max)
+        return ? k
+      else chooseMax(f, max)!; // retry
+    };
+  };
```
crusso, line 67:
```suggestion
  func chooseMax(f : Random.Finite, max : Nat) : ? Nat {
    assert max > 0;
    do ? {
      var n = max - 1 : Nat;
      var k = 0;
      while (n != 0) {
        k *= 2;
        k += bit(f.coin()!);
        n /= 2;
      };
      if (k < max) k else chooseMax(f, max)!;
    };
  };
```

doc/md/writing-motoko/randomness.md
```diff
@@ -0,0 +1,127 @@
+---
+sidebar_position: 22
+---
+
+# Randomness
+
+## Overview
+
+The Motoko feature `Random` can be used for generating random values within smart contracts on ICP. Randomness on ICP is an intricate process, since ICP uses deterministic computing to obtain cryptographic random values.
```
crusso, line 9:
```suggestion
The Motoko [`Random`](../base/Random.md) base library can be used for generating random values within smart contracts on ICP. Randomness on ICP is an intricate process, since ICP uses deterministic computing to obtain cryptographic random values.
```

doc/md/writing-motoko/randomness.md
```diff
@@ -0,0 +1,127 @@
...
+
+At the low level, ICP uses a Verifiable Randomness Function that is exposed by the management canister and Motoko `Random` module. In each execution round, the randomness function is evaluated with number of the current round used as input in order to produce a fresh set of random bytes.
```
crusso, line 11:
```suggestion
At the low level, ICP uses a Verifiable Random Function that is exposed by the management canister and is used by Motoko `Random` module. In each execution round, the randomness function is evaluated with number of the current round used as input in order to produce a fresh set of random bytes.
```

crusso (commented):
Unfortunately, I don't think we can easily implement what Timo suggests using what's available in base (sigh). We'd need a Sha2 implementation that produces an infinite stream of bits from a finite blob and the Random class would need to generalized to work with an infinite bit- or bool iterator to avoid running out of supply.

I made some interim improvements instead...

crusso:
@timohanke is that a little better, if not perfect?

timohanke:
> @timohanke is that a little better, if not perfect?

Yeah, best we can do with only base available.

doc/md/writing-motoko/randomness.md
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+};
+```
+
+View this example on the [Motoko Playground](https://play.motoko.org/?tag=2675232834) or on [GitHub](https://github.com/crusso/card-shuffle/blob/main/src/cards_backend/main.mo).
+
+REMARK: The above solution directly uses the finite blob of 256-random bits returned by the management canister. Class `Random.Finite` uses this finite supply of bits to generate at most 256 coin flips, returning `null` when no more flips are possible.
+ 
+When its current supply of bits is exhausted, the code asynchronously requests another 256-bit blob to continue the shuffle. A more efficient, and equally robust approach would be to use the first 256-bit blob as a seed to a sequential pseudo random number generator, generating an infinite, lazy stream of bits, and then complete the shuffle with a single round of communication.
```
crusso, line 106:
```suggestion
:::tip

The above solution directly uses the finite blob of 256-random bits returned by the management canister. Class `Random.Finite` uses this finite supply of bits to generate at most 256 coin flips, returning `null` when no more flips are possible.
 
When its current supply of bits is exhausted, the code asynchronously requests another 256-bit blob to continue the shuffle. A more efficient, and equally robust approach would be to use the first 256-bit blob as a seed to a sequential pseudo random number generator, generating an infinite, lazy stream of bits, and then complete the shuffle with a single round of communication.

:::

```

Shall we change the remark to a tip so both paragraphs are in the same logical unit?

crusso (approved):
But see suggestion...

PR #4544, opened 2024-05-17
experiment: precise heap tags

crusso:
Implement precise tagging of heap allocated objects.

Splits 

* `TAG_ARRAY` into consecutive `TAG_ARRAY_I`, `TAG_ARRAY_M`, `TAG_ARRAY_T`, `TAG_ARRAY_S` (immutable array, array, tuple, shared function). 
* `TAG_BLOB` into consecutive `TAG_BLOB_B` ,` TAB_BLOB_T` and `TAG_BLOB_P`, `TAG_BLOB_A` (blob, text, principal, actor reference).
*  `TAG_BITS_32` into consecutive `TAG_BITS32_U`, `TAG_BITS32_S` and `TAG_BITS32_F`  ((boxed) Nat32/Int32, Float32 (reserved for future use).
* `TAG_BITS_64` into consecutive `TAG_BITS64_U`, `TAG_BITS64_S` and `TAG_BITS64_F`  ((boxed) Nat64/Int32, Float64).

Modifies array slicing logic to preserve the base tag of the underlying array, by stealing the top 2 bits of the slice start field, relying on the fact that array indices are < 2^30. For clarity, the tag in decompressed on each iteration. For efficiency, it could actually be maintained during slicing compressed form as a sort ( see 448962f).

TODO:
~[ ] print refined tags in debug output?~ (rts debug printing is rarely used and can be adapted later)


rts/motoko-rts/src/types.rs
```diff
@@ -447,29 +457,71 @@ pub type Tag = u32;
...
-pub const TAG_BLOB: Tag = 17;
-pub const TAG_FWD_PTR: Tag = 19; // Only used by the copying GC - not to be confused with forwarding pointer in the header used for incremental GC.
-pub const TAG_BITS32: Tag = 21;
-pub const TAG_BIGINT: Tag = 23;
-pub const TAG_CONCAT: Tag = 25;
-pub const TAG_REGION: Tag = 27;
-pub const TAG_NULL: Tag = 29;
-pub const TAG_ONE_WORD_FILLER: Tag = 31;
-pub const TAG_FREE_SPACE: Tag = 33;
+pub const TAG_ARRAY_I: Tag = 5;
```
luc-blaeser, line 460: (Only a very little stylistic thought: Maybe we could name the tags with a unabbreviated suffix like `TAG_ARRAY_IMMUTABLE` etc. or just have a short comment here at the declaration.)

rts/motoko-rts/src/types.rs
```diff
@@ -447,29 +457,71 @@ pub type Tag = u32;
...
 // purposes of `visit_pointer_fields`.
 // Invariant: the value of this (pseudo-)tag must be
 //            higher than all other tags defined above
-pub const TAG_ARRAY_SLICE_MIN: Tag = 34;
+pub const TAG_ARRAY_SLICE_MIN: Tag = 50;
+
+pub fn slice_tag(array_tag: Tag, slice_start: u32) -> Tag {
+    debug_assert!(array_tag == TAG_ARRAY_I || array_tag == TAG_ARRAY_M || array_tag == TAG_ARRAY_T);
+    debug_assert!(slice_start >= TAG_ARRAY_SLICE_MIN && slice_start < (1 << 30));
+    (((array_tag - TAG_ARRAY_I) / 2) << 30) | slice_start
```
luc-blaeser, line 495: Here, maybe, a debug assertion could be helpful, to ensure that `array_tag - TAG_ARRAY` is even and has a distance of maximum 3 (to be encoded in two bits).
The 30 bits is also related to `MAX_ARRAY_SIZE` in `constant.rs` that it must cover the maximum size of the array, i.e. `MAX_ARRAY_SIZE < (1 << 30)`. Maybe, a local constant could be helpful (just that I double check it for the 64-bit porting, where maximum array size is a little bit larger).

luc-blaeser: Just a thought: Maybe the odd numbering is no longer needed in future if we focus on incremental GC. In this case the division and multiplication by 2 would no longer be necessary...

rts/motoko-rts/src/types.rs
```diff
@@ -447,29 +457,71 @@ pub type Tag = u32;
...
+
+pub fn slice_start(tag: Tag) -> (Tag, u32) {
+    debug_assert!(
+        tag == TAG_ARRAY_I
+            || tag == TAG_ARRAY_M
+            || tag == TAG_ARRAY_T
+            || tag >= TAG_ARRAY_SLICE_MIN
+    );
+    if tag >= TAG_ARRAY_SLICE_MIN {
+        (TAG_ARRAY_I + (tag >> 30) * 2, tag << 2 >> 2)
```
luc-blaeser, line 506: Maybe, the the second and third occurrence of 2 could be expressed as `usize::BITS - 30` (where 30 is the constant of the shift to encode the tag in the slicing info)

rts/motoko-rts/src/types.rs
```diff
@@ -447,29 +457,71 @@ pub type Tag = u32;
...
+
+pub fn base_array_tag(tag: Tag) -> Tag {
+    debug_assert!(
+        tag == TAG_ARRAY_I
+            || tag == TAG_ARRAY_M
+            || tag == TAG_ARRAY_T
+            || tag >= TAG_ARRAY_SLICE_MIN
+    );
+    if tag >= TAG_ARRAY_SLICE_MIN {
+        TAG_ARRAY_I + (tag >> 30) * 2
```
luc-blaeser, line 520: Perhaps a possible sanity check: The result would be `tag == TAG_ARRAY_I || tag == TAG_ARRAY_M || tag == TAG_ARRAY_T` (based on the assumption that the three array tags are consecutively numbered and odd).

rts/motoko-rts/src/types.rs
```diff
@@ -506,7 +558,9 @@ impl Obj {
     }
 
     pub unsafe fn as_blob(self: *mut Self) -> *mut Blob {
```
luc-blaeser, line 560: I wonder whether it could make sense to use dedicated `as_blob` functions, depending on the context, e.g. for texts etc.

rts/motoko-rts/src/types.rs
```diff
@@ -572,6 +626,22 @@ impl Array {
...
+    pub unsafe fn get_slice_start(self: *const Self) -> (Tag, u32) {
+        slice_start((*self).header.tag)
+    }
+
+    pub unsafe fn set_slice_start(self: *mut Self, array_tag: Tag, start: u32) -> () {
+        (*self).header.tag = slice_tag(array_tag, start)
+    }
+
+    pub unsafe fn restore_tag(self: *mut Self, array_tag: Tag) -> () {
+        (*self).header.tag = array_tag;
```
luc-blaeser, line 643: I assume the restored tag would fulfil `array_tag == TAG_ARRAY_I || arraty_tag == TAG_ARRAY_M || array_tag == TAG_ARRAY_T` (no slicing any longer)? Not sure, if a debug assertion would make sense.

rts/motoko-rts/src/types.rs
```diff
@@ -389,7 +399,7 @@ impl Value {
     /// In debug mode panics if the value is not a pointer or the
     /// pointed object is not a `Blob`.
     pub unsafe fn as_stream(self) -> *mut Stream {
-        debug_assert_eq!(self.tag(), TAG_BLOB);
+        debug_assert!(self.is_blob());
```
luc-blaeser, line 402: I believe the stream would always be a `TAG_BLOB_B`? Maybe, a stricter assertion could be applied.

rts/motoko-rts/src/types.rs
```diff
@@ -373,14 +383,14 @@ impl Value {
     /// Get the pointer as `Blob` using forwarding. In debug mode panics if the value is not a pointer or the
     /// pointed object is not a `Blob`.
     pub unsafe fn as_blob(self) -> *const Blob {
```
luc-blaeser, line 385: Just an idea: We could expose distinct access function, on for pure blob and one for text (maybe also for principal) with corresponding specific tag checks, and then call the corresponding specific function depending of the context. This would only serve for making sanity checks stricter.

rts/motoko-rts/src/principal_id.rs
```diff
@@ -5,15 +5,15 @@ use crate::mem_utils::memcpy_bytes;
...
+use crate::types::{Bytes, Value, TAG_BLOB_B, TAG_BLOB_T};
 
 use motoko_rts_macros::ic_mem_fn;
 
 // CRC32 for blobs. Loosely based on https://rosettacode.org/wiki/CRC-32#Implementation_2
 
 #[no_mangle]
 pub unsafe extern "C" fn compute_crc32(blob: Value) -> u32 {
-    if blob.tag() != TAG_BLOB {
+    if !blob.is_blob() {
```
luc-blaeser, line 16: Would this be a case where the blob is a principal? The check could then be constrained to `TAG_BLOB_P`.

rts/motoko-rts/src/principal_id.rs
```diff
@@ -98,7 +98,11 @@ pub unsafe fn base32_of_checksummed_blob<M: Memory>(mem: &mut M, b: Value) -> Va
     let n = b.as_blob().len();
     let mut data = b.as_blob().payload_const();
 
-    let r = alloc_blob(mem, Bytes((n.as_u32() + 4 + 4) / 5 * 8)); // contains padding
+    let r = alloc_blob(
+        mem,
+        TAG_BLOB_T, /*?*/
```
luc-blaeser, line 103: This is a good question. The base32 would be a textual encoding?

rts/motoko-rts/src/principal_id.rs
luc-blaeser: **Errand**: Maybe not. Below it seems to be a `TAG_BLOB_B` that is passed to this function. It is hard to follow...

rts/motoko-rts/src/principal_id.rs
```diff
@@ -225,7 +229,7 @@ unsafe fn base32_to_principal<M: Memory>(mem: &mut M, b: Value) -> Value {
     let mut data = blob.payload_const();
 
     // Every group of 5 characters will yield 6 bytes (due to the hypen)
-    let r = alloc_blob(mem, Bytes(((n.as_u32() + 4) / 5) * 6));
+    let r = alloc_blob(mem, TAG_BLOB_T, Bytes(((n.as_u32() + 4) / 5) * 6));
```
luc-blaeser, line 232: I am not sure: Would the result maybe be a `TAG_BLOB_P`? (Just trying to infer it by the function name.)

rts/motoko-rts/src/text.rs
```diff
@@ -172,7 +174,7 @@ unsafe extern "C" fn text_to_buf(mut s: Value, mut buf: *mut u8) {
 unsafe extern "C" fn stream_write_text(stream: *mut Stream, mut s: Value) {
     loop {
         let s_ptr = s.as_obj();
-        if s_ptr.tag() == TAG_BLOB {
+        if s_ptr.tag() == TAG_BLOB_B || s_ptr.tag() == TAG_BLOB_T {
```
luc-blaeser, line 177: Just curious: The `TAG_BLOB_B` would occur for the Candid streaming. And the `TAG_BLOB_T` because of the principal to text conversions?

rts/motoko-rts-tests/src/gc/heap.rs
```diff
@@ -570,7 +570,7 @@ fn create_static_heap(
     // root.
     let array_addr = u32::try_from(heap.as_ptr() as usize).unwrap();
     let mut offset = 0;
-    write_word(heap, offset, TAG_ARRAY);
+    write_word(heap, offset, TAG_ARRAY_M); // TODO: TBR
```
luc-blaeser, line 573: Alternatively, we could also use `TAG_ARRAY_I` for the static roots. But it is quite uncritical as it is not scanned by graph stabilization.

rts/motoko-rts-tests/src/principal_id.rs
```diff
@@ -36,11 +36,11 @@ pub unsafe fn test() {
 
     let text = text_of_str(&mut heap, "aaaaa-aa");
     let principal = blob_of_principal(&mut heap, text);
-    assert_eq!(text_compare(principal, text_of_str(&mut heap, ""),), 0);
+    assert_eq!(blob_compare(principal, text_of_str(&mut heap, ""),), 0);
 
     let text = text_of_str(&mut heap, "bfozs-kwa73-7nadi");
     assert_eq!(
-        text_compare(
+        blob_compare(
```
luc-blaeser, line 43: If I am not mistaken, this compares a `TAG_BLOB_B` with `TAG_BLOB_T`. Perhaps, the `text_of_str()` is here more like a blob encoder in this specific test...

src/codegen/compile.ml
```diff
@@ -1996,18 +1999,22 @@ module Tagged = struct
      so update all!
    *)
 
+  type bits_sort = U | S | F
```
luc-blaeser, line 2002: (Just cosmetics: Maybe a brief comment could be helpful saying what the abbreviations stand for.)

src/codegen/compile.ml
```diff
@@ -2197,7 +2212,7 @@ module Tagged = struct
     let (set_o, get_o) = new_local env "o" in
     let prep (t, code) = (t, get_o ^^ code)
     in set_o ^^ get_o ^^ branch_default env retty def (List.map prep cases)
-
+(*
```
luc-blaeser, line 2230: Maybe we can delete this dead code...

src/codegen/compile.ml
```diff
@@ -2571,7 +2587,11 @@ module BoxedWord64 = struct
 
   let payload_field = Tagged.header_size
 
-  let heap_tag env pty = Tagged.Bits64 (* TODO *)
+  let heap_tag env pty =
```
luc-blaeser, line 2605: Nice: The TODO is now addressed.

src/codegen/compile.ml
```diff
@@ -4548,7 +4577,12 @@ module Text = struct
     set_blob ^^
     get_blob ^^ Blob.as_ptr_len env ^^
     E.call_import env "rts" "utf8_valid" ^^
-    G.if1 I32Type (Opt.inject_simple env get_blob) (Opt.null_lit env)
+    G.if1 I32Type
+      (get_blob ^^ Blob.as_ptr_len env ^^
+       of_ptr_size env ^^ (* creates text blob *)
```
luc-blaeser, line 4606: Good that you covered this case: We indeed need to create a copy of the blob, as it has a different purpose (and tag)...

src/codegen/compile.ml
```diff
@@ -4658,22 +4692,24 @@ module Arr = struct
...
       ] @ element_instructions)
 
   (* Does not initialize the fields! *)
   (* Note: Post allocation barrier must be applied after initialization *)
-  let alloc env =
-    E.call_import env "rts" "alloc_array"
+  let alloc env array_sort len =
+    compile_unboxed_const Tagged.(int_of_tag (Array array_sort)) ^^
+    len ^^
+    E.call_import env "rts" "alloc_array" (* TODO *)
```
luc-blaeser, line 4736: Just curious about the `TODO`: Do you have a specific open aspect in mind?

src/codegen/compile.ml
```diff
@@ -5338,7 +5402,7 @@ module IC = struct
 
   let error_message env =
     Func.share_code0 Func.Never env "error_message" [I32Type] (fun env ->
-      Blob.of_size_copy env
+      Blob.of_size_copy env Tagged.T
```
luc-blaeser, line 5405: (Not to myself @luc-blaeser ): I got here...

src/codegen/compile.ml
```diff
@@ -7519,11 +7583,11 @@ module MakeSerialization (Strm : Stream) = struct
             *)
             get_thing ^^ set_result ^^
             get_memo ^^ get_result ^^ store_unskewed_ptr ^^
-            get_memo ^^ compile_add_const 4l ^^ Blob.lit env (typ_hash t) ^^ store_unskewed_ptr
+            get_memo ^^ compile_add_const 4l ^^ Blob.lit env Tagged.T (typ_hash t) ^^ store_unskewed_ptr
```
luc-blaeser, line 7586: Alternatively, we could also consider the hash fingerprint (to double check the aliasing deserialization mechanism) as pure blob?

src/codegen/compile.ml
```diff
@@ -8639,7 +8703,7 @@ module GCRoots = struct
     })
 
   let store_static_roots env =
-    Arr.vanilla_lit env (E.get_static_roots env)
+    Arr.vanilla_lit env Tagged.M (E.get_static_roots env)
```
luc-blaeser, line 8706: I wonder whether the static roots could be tagged as immutable, as they reside in static space and should not change at runtime (references to static MutBoxes that are not moved by the GC). This would be different with EOP.

src/codegen/compile.ml
```diff
@@ -8747,7 +8812,10 @@ module StackRep = struct
     | Const.Unit -> Tuple.unit_vanilla_lit env
     | Const.Array cs ->
       let ptrs = List.map (materialize_const_t env) cs in
-      Arr.vanilla_lit env ptrs
+      Arr.vanilla_lit env Tagged.I ptrs
+    | Const.Tuple cs ->
```
luc-blaeser, line 8816: This means that we never optimized for constant tuples before? Good extension.

luc-blaeser (approved):
Very nice PR. Thanks a lot.

I also measured the performance with the GC benchmark and the difference is negligible.

rts/motoko-rts/src/debug.rs
```diff
@@ -204,15 +204,15 @@ pub(crate) unsafe fn print_boxed_object(buf: &mut WriteBuf, p: usize) {
                 (*variant).field.get_raw()
             );
         }
-        TAG_BLOB => {
+        TAG_BLOB_B | TAG_BLOB_T | TAG_BLOB_P => {
```
crusso, line 207: Missing TAG_BLOB_A

crusso, line 207:
```suggestion
        TAG_BLOB_B | TAG_BLOB_T | TAG_BLOB_P | TAG_BLOB_A => {
```

rts/motoko-rts-tests/src/principal_id.rs
crusso: Yes, it is. I changed to blob_compare because text_compare (that deals with ropes) checks the TAG_BLOB_T tags of leaves. bob_compare only checks the contents.

rts/motoko-rts/src/principal_id.rs
crusso: Yeah. I'll remove the comment.

rts/motoko-rts/src/principal_id.rs
crusso: Unfortunately, this also required adjust the tests in crc32.rs

rts/motoko-rts/src/text.rs
crusso: It's because Stream.write_text is is abused to write both the binary header (type table) and textual string contents.

src/codegen/compile.ml

src/codegen/compile.ml
crusso: Nope, deleted comment

src/codegen/compile.ml
crusso: Even more accurate would be as a tuple, since the contents are not uniformly typed.

crusso: (But then Motoko tuples cannot contain mutboxs, so this would be cheating too)

src/codegen/compile.ml
crusso: Sure.

src/codegen/compile.ml
luc-blaeser: You are right. `Tagged.M` seems a pragmatic choice.

crusso: Ok, let's just keep as is

rts/motoko-rts-tests/src/gc/heap.rs
crusso: (won't fix, as discussed)

rts/motoko-rts/src/types.rs
crusso: added assert

rts/motoko-rts/src/types.rs
crusso: done

rts/motoko-rts/src/types.rs
crusso: I've done something similar (and added some assert to static_checks.rs)

rts/motoko-rts/src/types.rs
crusso: I did a grep and many of the as_blob uses are immediatly following allocation with a known tag, so perhaps not that useful. Will take another look though

rts/motoko-rts/src/types.rs
crusso: (See above)

rts/motoko-rts/src/types.rs
crusso: True, maybe we could do that later.
I had version where I compressed the tag just once and keep in compressed form until the final reset, but the code was less clear if slightly more efficient, so I backed out.

src/codegen/compile.ml

rts/motoko-rts/src/types.rs
crusso: Decided against renaming for now, just for the sake of time. But did add comments.

crusso:
I think I've addressed all the issue. PTAL

rts/motoko-rts-tests/src/gc/heap.rs
```diff
@@ -497,7 +497,7 @@ fn create_dynamic_heap(
         let mut heap_offset = continuation_table_offset;
 
         let continuation_table_address = u32::try_from(heap_start + heap_offset).unwrap();
-        write_word(dynamic_heap, heap_offset, TAG_ARRAY);
+        write_word(dynamic_heap, heap_offset, TAG_ARRAY_M); // TODO: TBR
```
luc-blaeser, line 500:
```suggestion
        write_word(dynamic_heap, heap_offset, TAG_ARRAY_M);
```

rts/motoko-rts-tests/src/gc/heap.rs
luc-blaeser, line 573:
```suggestion
    write_word(heap, offset, TAG_ARRAY_M);
```

rts/motoko-rts/src/blob_iter.rs
```diff
@@ -16,7 +16,7 @@ unsafe fn blob_iter<M: crate::memory::Memory>(mem: &mut M, blob: Value) -> Value
 
     // NB. cannot use as_array() here as we didn't write the header yet
     let iter_array = iter_ptr.get_ptr() as *mut Array;
-    (*iter_array).header.tag = TAG_ARRAY;
+    (*iter_array).header.tag = TAG_ARRAY_T;
```
luc-blaeser, line 19: This would be the iterator case, I will need to check whether it is reachable in the graph copy...

rts/motoko-rts/src/text_iter.rs
```diff
@@ -53,7 +53,7 @@ const ITER_TODO_IDX: u32 = 2;
 /// Returns a new iterator for the text
 #[ic_mem_fn]
 pub unsafe fn text_iter<M: Memory>(mem: &mut M, text: Value) -> Value {
-    let iter = alloc_array(mem, 3);
+    let iter = alloc_array(mem, TAG_ARRAY_T, 3);
```
luc-blaeser, line 56: Note to myself: This is the other iterator case of a non-standard array. To be checked that it is not reachable by graph copy...

luc-blaeser (dismissed):
Also the tagging of the actor and shared function references looks really good. Thank you very much. I will now try to merge with the EOP PRs.

rts/motoko-rts/src/memory.rs
```diff
@@ -45,11 +45,11 @@ pub trait Memory {
 /// Note: After initialization, the post allocation barrier needs to be applied to all mutator objects.
 /// For RTS-internal blobs that can be collected by the next GC run, the post allocation barrier can be omitted.
 #[ic_mem_fn]
-pub unsafe fn alloc_blob<M: Memory>(mem: &mut M, size: Bytes<u32>) -> Value {
+pub unsafe fn alloc_blob<M: Memory>(mem: &mut M, tag: u32, size: Bytes<u32>) -> Value {
```
luc-blaeser, line 48:
```suggestion
pub unsafe fn alloc_blob<M: Memory>(mem: &mut M, tag: u32, size: Bytes<u32>) -> Value {
    debug_assert!(tag == TAG_BLOB_B || tag == TAG_BLOB_T || tag == TAG_BLOB_P || tag == TAG_BLOB_A);
```

rts/motoko-rts/src/memory.rs
```diff
@@ -59,7 +59,7 @@ pub unsafe fn alloc_blob<M: Memory>(mem: &mut M, size: Bytes<u32>) -> Value {
 /// Allocate a new array.
 /// Note: After initialization, the post allocation barrier needs to be applied to all mutator objects.
 #[ic_mem_fn]
-pub unsafe fn alloc_array<M: Memory>(mem: &mut M, len: u32) -> Value {
+pub unsafe fn alloc_array<M: Memory>(mem: &mut M, tag: u32, len: u32) -> Value {
```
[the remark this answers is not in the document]
luc-blaeser: Just noticed while merging to EOP, that it is easy to mix the `tag` and `len` arguments...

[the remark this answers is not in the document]
ggreif: Asserting the tag is an effective defence.

rts/motoko-rts/src/types.rs
```diff
@@ -447,29 +457,91 @@ pub type Tag = u32;
...
+// The top two bits encode the original array tag, the remaining bits are the start index of the slice.
 // Invariant: the value of this (pseudo-)tag must be
 //            higher than all other tags defined above
-pub const TAG_ARRAY_SLICE_MIN: Tag = 34;
+pub const TAG_ARRAY_SLICE_MIN: Tag = 54;
+
+pub const TAG_SPACING: Tag = 2;
+
+pub fn is_blob_tag(tag: Tag) -> bool {
+    tag == TAG_BLOB_B || tag == TAG_BLOB_T || tag == TAG_BLOB_P || tag == TAG_BLOB_A
```
ggreif, line 498: one could help the compiler doing such tests by arranging the tags to have values that only differ in the two penultimate bits, e.g. 29..35.

src/codegen/compile.ml
```diff
@@ -1996,18 +1999,33 @@ module Tagged = struct
...
+    | U (* signed *)
+    | S (* unsigned *)
+    | F (* float *)
+  type array_sort =
+    | I (* [ T ] *)
+    | M (* [var T ] *)
+    | T (* (T,+) *)
+    | S (* shared ... -> ... *)
+  type blob_sort =
+    |  B (* Blob *)
```
ggreif, line 2012:
```suggestion
    | B (* Blob *)
```

src/codegen/compile.ml
```diff
@@ -11605,8 +11654,18 @@ and compile_prim_invocation (env : E.t) ae p es at =
     compile_exp_vanilla env ae e ^^
     IC.trap_text env
 
-  | OtherPrim ("blobToArray" | "blobToArrayMut"), e ->
-    const_sr SR.Vanilla (Arr.ofBlob env)
+  | OtherPrim "principalOfBlob", e ->
```
ggreif, line 11657: When a `nop` turns into an alloc (+ linear copy) we should better have a good explanation for the refined tagging scheme :->

src/mo_values/prim.ml
```diff
@@ -292,6 +292,12 @@ let prim trap =
       in go (fun xs -> xs) k 0
     | _ -> assert false
     )
+
+
+  | "blobOfPrincipal" -> fun _ v k -> k v
+  | "principalOfBlob" -> fun _ v k -> k v
```
ggreif, line 298:
```suggestion
  | "blobOfPrincipal"
  | "principalOfBlob"
```

ggreif (approved):
The patch is long, but mostly type-directed, I don't see correctness hazards. Hopefully the blob conversions will not cause perf concerns, as they are mostly small. LGTM

rts/motoko-rts/src/memory.rs
crusso: Added by Luc in EOP PRs.

src/codegen/compile.ml
crusso: Cosmetic (won't fix)

src/mo_values/prim.ml
crusso: Cosmetic (won't fix)

crusso:
@ggreif, since Luc has already merged this branch into his EOP PRs, and a squash merge of this PR to master may cause him trouble merging later, we've decide not to merge this branch directly into master but as part of larger PR. That's also why I'm not addressing the issues you found (which are mostly cosmetic).

rts/motoko-rts/src/types.rs
crusso: True, but is_blob_tag is mostly called from assertions (apart from a relatively cheap call in crc32, that could also be an assert). Won't fix since Luc has already merged this PR into his stack...

crusso:
Not merging into master as already merged into Luc's EOP PRs.

luc-blaeser:
I now merged the PR into the EOP PR stack. I think we could probably now close this PR.
@crusso: Thank you again for this great contribution!

luc-blaeser:
Merged with EOP

PR #4546, opened 2024-05-18
experiment: precise heap tagging (continued)

crusso:
Probably too ugly to use, but mostly working

[x] TODO: refine Const.Array in to Const.Array and Const.Tuple
[x] Adapt array slicing in incremental and generational GC, allowing preservation of precise array tag 
    (no change required for mark-compact that preserves the original tag on the mark stack).


rts/motoko-rts/src/text.rs
```diff
@@ -188,7 +188,7 @@ unsafe extern "C" fn stream_write_text(stream: *mut Stream, mut s: Value) {
 #[ic_mem_fn]
 pub unsafe fn blob_of_text<M: Memory>(mem: &mut M, s: Value) -> Value {
     let obj = s.as_obj();
-    if obj.tag() == TAG_BLOB_T {
+    if obj.tag() == TAG_BLOB_T ||obj.tag() == TAG_BLOB_B || obj.tag() == TAG_BLOB_P {
```
crusso, line 191: restrict back to TAG_BLOB_T only

PR #4549, opened 2024-05-21
Make the Incremental GC the Default

luc-blaeser:
As a preparation step towards enhanced orthogonal persistence.

Also re-measuring the base line of the CI (micro) performance tests.


luc-blaeser:
> Comparing from [ecf39a9](https://github.com/dfinity/motoko/commit/ecf39a990c59fac5047e48662a8ec03cadbd4f53) to [54a4c56](https://github.com/dfinity/motoko/commit/54a4c5650b8e321a2b0e6fc64741b8f5818b56a3): In terms of gas, 5 tests regressed and the mean change is +40.1%. In terms of size, 5 tests regressed and the mean change is +2.5%.

I believe we should run the perf tests without `--force-gc` as this causes unnatural and expensive partition copying for the incremental GC that is not triggered repeatedly in normal mode. I will remeasure the difference in a separate PR.

luc-blaeser:
Put back on hold. We can enable this when EOP has become the default mode.

PR #4553, opened 2024-05-23
chore: replace drun by PocketIC

mraszyk:


test/perf/assetstorage.mo
```diff
@@ -34,7 +34,7 @@ shared({caller = creator}) actor class () {
 };
 
 //CALL query list 0x4449444C0000
-//CALL ingress store 0x4449444C016d7b02710004414243440400010203
+//CALL ingress store RElETAFtewJxAARBQkNEBAABAgM=
```
chenyan-dfinity, line 37: Can we keep the original format? Base64 doesn't seem very readable.

mraszyk:
This MR was merely a draft and I'm closing it now as I'll make drun use PocketIC internally so that we can keep the drun input format.

PR #4556, opened 2024-05-28
Unbounded Number of Heap Partitions for 64-Bit

luc-blaeser:
# Unbounded Number of Heap Partitions for 64-Bit

Supporting dynamic extensions of the heap partition tables to support the entire 64-bit address space.

## Design
To address full memory space, the runtime system maintains a linked list of partition tables:
* The first partition table is located in the static heap descriptor (persistent metadata).
* Subsequent extension partition tables reside in the static space of a partition.
The static locations are necessary to ensure that partition tables are never moved.

A partition table addresses 8 GB of consecutive memory. This allows large objects (arrays/blobs) of a maximum size of 8128 MB (8 GB minus 64 MB partition size). 

## Testing
This PR is tested specifically with small partition table lengths and smaller partitions (128 MB per partition table):
* GC benchmark (with and without sanity checks and memory checks), performance and scalability tests.
* Randomized GC tests (with sanity checks) in a heap range of 8-9 GB.


rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -390,40 +491,125 @@ impl PartitionedHeap {
...
+
+    unsafe fn mutable_partition_table(&mut self, partition_index: usize) -> *mut PartitionTable {
+        let first_table = &mut self.partition_table as *mut PartitionTable;
+        debug_assert_ne!(first_table, null_mut());
+        if partition_index < PARTITIONS_PER_TABLE {
+            first_table
+        } else {
+            self.get_extension_table(partition_index)
+        }
+    }
```
crusso, line 536: NIT: These two functions are almost identical, can one express one in terms of the other?

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -390,40 +491,125 @@ impl PartitionedHeap {
...
     }
 
     fn mutable_partition(&mut self, index: usize) -> &mut Partition {
-        &mut self.partitions[index]
+        unsafe {
+            let table = self.mutable_partition_table(index);
+            let offset = index % PARTITIONS_PER_TABLE;
+            &mut (*table).partitions[offset]
+        }
+    }
```
crusso, line 552: NIT: These two functions are almost identical, can one express one in terms of the other?

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -515,26 +701,40 @@ impl PartitionedHeap {
         }
     }
 
-    fn rank_partitions_by_garbage(&self) -> [usize; MAX_PARTITIONS] {
-        let mut ranked_partitions: [usize; MAX_PARTITIONS] = from_fn(|index| index);
```
crusso, line 519: I'm curious why this function even works? Doesn't it return a pointer to something stack allocated? Or is copying the array?

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -515,26 +701,40 @@ impl PartitionedHeap {
...
-        sort(&mut ranked_partitions, &|left, right| {
-            self.get_partition(left)
-                .garbage_amount()
-                .cmp(&self.get_partition(right).garbage_amount())
-                .reverse()
-        });
+    unsafe fn temporary_array<M: Memory>(mem: &mut M, length: usize) -> *mut usize {
+        // No post allocation barrier as this RTS-internal blob can be collected by the GC.
```
crusso, line 705: I guess you need to use a blob because Rust doesn't let you stack-allocate something of dynamic size `length`?

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -1,14 +1,16 @@
...
-use crate::constants::WASM_PAGE_SIZE;
+use crate::constants::{GB, WASM_PAGE_SIZE};
 use crate::gc::incremental::memory_reserve;
-use crate::gc::incremental::partitioned_heap::MAXIMUM_MEMORY_SIZE;
 use crate::rts_trap_with;
 use crate::types::{Bytes, Value, Words};
 use core::arch::wasm64;
 use core::cmp::min;
 
+// TODO: Can we inspect the maximum main memory limit from the IC?
```
crusso, line 11: I don't think we can... even if you could, you might not get that amount on request.

crusso: (Indeed, if we really want to leave a reliable reserve, perhaps we should actually allocate some extra space beyond ptr unless we are actively using the reserve. But that makes my head hurt)

crusso (approved):
LGTM (but my confidence is not high today (too hot!))

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
luc-blaeser: Thanks, I refactored this.

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
luc-blaeser: The caller stack-allocates the space for the return value and only passes a pointer to that location. This allows bulk return values without copying. (This is also done in C and in Oberon :))

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
luc-blaeser: Yes, unfortunately.

rts/motoko-rts/src/memory/ic.rs
luc-blaeser: An interesting idea of pre-allocating the reserve. A downside would be that it could be quite wasteful if the reserve is large (or we apply some heuristics of probing that limit when getting close to some "magic" boundaries, e.g. 4GB, 6GB, 8GB, ...).

PR #4558, opened 2024-05-31
experiment: precise heap tags for actors and shared functions

crusso:
builds on #4544 

Introduce two new tags:
`TAG_BLOB_A` : a blob representing an actor reference (distinct from a raw principal).
`TAG_ARRAY_S`:  a pair representing a shared function as an actor reference (TAG_BLOB_A) paired with a method name as a text blob (TAG_BLOB_T).

Adding these tags now, to distinguish from other values, will hopefully give us some leverage if we ever want to change their representation in the future (perhaps adding type descriptors to drive dynamic serialization, for example).

Requires some additional coercions primitives to convert from blobs to principals (missing from #4544) and actors to blobs, copying payload with a different tag.

TODO: 

- [x] fix prim "cast" uses
- [x] rename new coercions to match Prim.* names
- [x] implement interpreter prims (and new values?)
- [x] add _LINE_ directives to debug tagging

### DIFF
--- a/rts/motoko-rts-tests/src/gc.rs
+++ b/rts/motoko-rts-tests/src/gc.rs
@@ -289,10 +289,14 @@ fn check_dynamic_heap(
                         tag == TAG_ARRAY_I
                             || tag == TAG_ARRAY_M
                             || tag == TAG_ARRAY_T
+                            || tag == TAG_ARRAY_S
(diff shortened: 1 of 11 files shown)


luc-blaeser (approved):
Very nice! Thanks a lot.

Issue #4508, opened 2024-04-22
bug: incorrect unused identifier warning


crusso:
This code says that e is unused even though it is used.
```
module {
  public func foo() : async () {
    var x = try {} catch (e) { throw e };
    x;
  };
};
```
This example seems somewhat minimal. For example without the var x = the error does not occur. With let instead of var it also does not occur.

Also, and perhaps related:

```
module {
  public func foo() : async () {
    try {} catch (e) { };
  };
};
```
does not report that e is unused.

https://dfinity.slack.com/archives/CPL67E7MX/p1713629264089809

luc-blaeser:
Thank you for reporting. The PR https://github.com/dfinity/motoko/pull/4560 should fix this issue.

christoph-dfinity:
Fixed in https://github.com/caffeinelabs/motoko/pull/4560

PR #4560, opened 2024-06-05
Improve Unused Declaration Detection (Switch/Catch Clauses, No Warnings on Type Check Errors)

luc-blaeser:
Fix for issue: https://github.com/dfinity/motoko/issues/4508
* Correctly detect used and unused declaration in switch and try-catch clauses.

Only warn on unused identifiers if type-checking is error-free (https://github.com/dfinity/motoko/pull/4561)
* Avoid potential spurious cascading warnings in the presence of type checker errors.

Updated:
* Expected warnings in Motoko test cases.
* Refactoring of the Motoko base library to comply with the improved detection: https://github.com/dfinity/motoko-base/pull/636


test/run/unused-parameter.mo
crusso, line 1: Maybe move this test to test/fail - that's where we put typechecking tests.

test/run-drun/unused-exception.mo
crusso, line 1: Maybe move this test to test/fail - that's where we put typechecking tests.

test/run-drun/unused-switch-in-catch.mo
crusso, line 1: Maybe move this test to test/fail - that's where we put typechecking tests.

test/run-drun/used-exception.mo
crusso, line 1: Maybe move this test to test/fail - that's where we put typechecking tests.

src/mo_frontend/typing.ml
```diff
@@ -1975,15 +1975,17 @@ and infer_call env exp1 inst exp2 at t_expect_opt =
 
 (* Cases *)
 
-and infer_cases env t_pat t cases : T.typ =
-  List.fold_left (infer_case env t_pat) t cases
+and infer_cases env check_unused t_pat t cases : T.typ =
+  List.fold_left (infer_case env check_unused t_pat) t cases
 
-and infer_case env t_pat t case =
+and infer_case env check_unused t_pat t case =
```
crusso, line 1981: Do you need this extra `check_unused` parameter? I was about to suggest putting it as a field in env, but env.unused already exists. Can it not be re-used for this or is there something more subtle going on?

src/mo_frontend/typing.ml
```diff
@@ -1975,15 +1975,17 @@ and infer_call env exp1 inst exp2 at t_expect_opt =
...
   let {pat; exp} = case.it in
   let ve = check_pat env t_pat pat in
   let initial_usage = enter_scope env in
   let t' = recover_with T.Non (infer_exp (adjoin_vals env ve)) exp in
-  leave_scope env ve initial_usage;
+  if check_unused then
+    leave_scope env ve initial_usage
```
crusso, line 1987: I wonder if this might lead to false positives error when checking exp succeed to error recovery, before have checked all sub-expressions.


unused-false-positive
```
(switch (true,false) {
  case (a,b) { d # a # b };
}) : Text;
```
produces:

```[nix-shell:~/clean/motoko/test/fail]$ moc -c unused-false-positive.mo
unused-false-positive.mo:4.16-4.17: type error [M0057], unbound variable d
unused-false-positive.mo:4.9-4.10: warning [M0194], unused identifier a (delete or rename to wildcard `_` or `_a`)
unused-false-positive.mo:4.11-4.12: warning [M0194], unused identifier b (delete or rename to wildcard `_` or `_b`)
```
Could this be fixed my only doing leave_scope right after infer_exp (under the recovery_with)

something like:

```
let t' = recover_with T.Non (fun e -> 
   let t = infer_exp (adjoin_vals env ve); if check_unused then leave_scope env ve 
   initial_usage; t) 
   exp 
in
let t'' ...
```

But perhaps that leads to other issues...


Maybe it's not worth fixing. I was just wondering what would happen in the IDE with incorrect code producing lots of unused identifier warnins. We can just wait for the complaints, if any.

src/mo_frontend/typing.ml
```diff
@@ -1998,8 +2000,11 @@ and check_cases env t_pat t cases =
 
 and check_case env t_pat t case =
   let {pat; exp} = case.it in
+  let initial_usage = enter_scope env in
   let ve = check_pat env t_pat pat in
-  recover (check_exp (adjoin_vals env ve) t) exp
+  let t' = recover (check_exp (adjoin_vals env ve) t) exp in
+  leave_scope env ve initial_usage;
```
crusso, line 2006: See above - perhaps leave_scope should only happen under `recover (...)`.

src/mo_frontend/typing.ml
```diff
@@ -1509,13 +1509,13 @@ and infer_exp'' env exp : T.typ =
...
     let t1 = infer_exp_promote env exp1 in
-    let t = infer_cases env t1 T.Non cases in
+    let t = infer_cases env true t1 T.Non cases in
     if not env.pre then
       coverage_cases "switch" env cases t1 exp.at;
     t
   | TryE (exp1, cases) ->
     let t1 = infer_exp env exp1 in
-    let t2 = infer_cases env T.catch T.Non cases in
+    let t2 = infer_cases env false T.catch T.Non cases in
```
crusso, line 1518: I don't quite understand why try and switch are handled differently?

src/mo_frontend/typing.ml
luc-blaeser: I actually tried this yesterday as my first choice, but unfortunately some case failed. I believe it was in the base library, the `Trie.mo` where suddenly a false positive occurred (used declaration of a switch inside a declaration inside another switch was recognized as unused in error)... I will try to extract that case and add it to the test cases too. The reason is that the `env.unused` is applied recursively, while the `check_unused` is applied only at one level. And there is case where `env.unused` is recursively turned off, and then the local `check_unused` in a switch block should not override this. But it is really difficult...

src/mo_frontend/typing.ml
luc-blaeser: Yes, this is indeed a problem and should be fixed, ideally maybe in a separate PR. If I understand correctly, it is not per se related to this fix which addresses an inconsistency in the detection of (correctly typed) the try-catch causes.
I also observed these spurious cascading errors. I will prepare a separate fix for this. Thanks for bringing this up.

src/mo_frontend/typing.ml
luc-blaeser: Or we could postpone the unused warnings after successful type checking. This would probably be a quite general solution.

src/mo_frontend/typing.ml
luc-blaeser: I try to see whether I can simplify this. I believe `check_unused` is actually `not env.pre` in this case. It is independent of `env.check_unused` as I checked yesterday...

src/mo_frontend/typing.ml
luc-blaeser: I believe I can simplify this, see above. The distinction was actually because try-catch block seemed to get inferred multiple times, some with `env.pre`...

crusso (approved):
LGTM

I'll leave it to you whether you want to merge  #4561 into this or separately into master.

I guess  a changelog entry would be good.

src/mo_frontend/typing.ml
luc-blaeser: Now solved by merging in https://github.com/dfinity/motoko/pull/4561. Thank you very much for this PR!

src/mo_frontend/typing.ml
luc-blaeser: Solved by merged https://github.com/dfinity/motoko/pull/4561

PR #4561, opened 2024-06-06
experiment: only warn on unused identifiers if type-checking is (proper) error-free

crusso:
Builds on #4560 

This should avoid false positives due to error recovery. Does however, remove many unused warnings from tests that contain real errors, preventing unused warnings from being reported.


luc-blaeser (approved):
Great! Thank you very much. No longer these spurious warnings when there is a type checking error.

PR #4562, opened 2024-06-06
experiment: test stabilization of text and blob iterators

crusso:
extra tests for #4475, adding test variants derived from on `stabilize-non-stable.drun`

Blob and Text iterators have custom representations that might trip up graph stabilization unless properly hidden behind closures (which they appear to be).

This PR test stable variables that contain text and blob iterators in their  dynamic values, if not their static types (which would make the types unstable).

UPDATE: tests suggests this is ok, even if I make the iterator states public fields in internals.mo (currently they are private).


luc-blaeser:
Thank you very much. I will merge them to the the graph copy PR, if this is okay.

PR #4563, opened 2024-06-07
experiment: integrate enhanced orthogonal persistence (against master)

crusso:


rts/motoko-rts/src/stream.rs
```diff
@@ -42,12 +42,12 @@ use crate::types::{size_of, Blob, Bytes, Stream, Value, TAG_BLOB};
 
 use motoko_rts_macros::ic_mem_fn;
 
-const MAX_STREAM_SIZE: Bytes<u32> = Bytes((1 << 30) - 1);
-const INITIAL_STREAM_FILLED: Bytes<u32> = Bytes(32);
-const STREAM_CHUNK_SIZE: Bytes<u32> = Bytes(128);
+const MAX_STREAM_SIZE: Bytes<usize> = Bytes((1 << 30) - 1);
```
crusso, line 45: Should that 30 be usize::BITS-2 or are we reluing on the fact that this is always only in the 32-bit runtime...

luc-blaeser:
Obsolete. EOP has been merged

PR #4567, opened 2024-06-07
Merge Preparation: Merge Precise Object Tagging to Graph Copy EOP

luc-blaeser:
Preparing merging https://github.com/dfinity/motoko/pull/4544 in https://github.com/dfinity/motoko/pull/4475 (level 3 of PR stack)


rts/motoko-rts-tests/src/stabilization/layout.rs
```diff
@@ -8,11 +8,19 @@ pub unsafe fn test() {
 
 fn test_stable_tags() {
     for object_kind in [
-        StableObjectKind::Array,
+        StableObjectKind::ArrayImmutable,
```
crusso, line 11: Cool!

rts/motoko-rts/src/memory.rs
```diff
@@ -34,11 +34,12 @@ pub trait Memory {
 /// Note: After initialization, the post allocation barrier needs to be applied to all mutator objects.
 /// For RTS-internal blobs that can be collected by the next GC run, the post allocation barrier can be omitted.
 #[ic_mem_fn]
-pub unsafe fn alloc_blob<M: Memory>(mem: &mut M, size: Bytes<usize>) -> Value {
+pub unsafe fn alloc_blob<M: Memory>(mem: &mut M, tag: Tag, size: Bytes<usize>) -> Value {
+    debug_assert!(tag == TAG_BLOB_B || tag == TAG_BLOB_T || tag == TAG_BLOB_P || tag == TAG_BLOB_A);
```
crusso, line 38:
```suggestion
    debug_assert!(is_blob_tag(tag));
```
(might need to import it)

rts/motoko-rts/src/memory.rs
```diff
@@ -48,13 +49,16 @@ pub unsafe fn alloc_blob<M: Memory>(mem: &mut M, size: Bytes<usize>) -> Value {
 /// Allocate a new array.
 /// Note: After initialization, the post allocation barrier needs to be applied to all mutator objects.
 #[ic_mem_fn]
-pub unsafe fn alloc_array<M: Memory>(mem: &mut M, len: usize) -> Value {
+pub unsafe fn alloc_array<M: Memory>(mem: &mut M, tag: Tag, len: usize) -> Value {
+    debug_assert!(
+        tag == TAG_ARRAY_I || tag == TAG_ARRAY_M || tag == TAG_ARRAY_T || tag == TAG_ARRAY_S
+    );
```
crusso, line 55:
```suggestion
    debug_assert!(is_base_array_tag(tag));
    );
```
(Might need to import it)

rts/motoko-rts/src/stabilization/deserialization.rs
```diff
@@ -96,24 +97,16 @@ impl Deserialization {
             |context, field_address| {
                 *field_address = translate(context, *field_address);
             },
-            |context, slice_start, array| {
-                const SLICE_INCREMENT: usize = 127;
```
crusso, line 100: Note the SLICE_INCREMENT in the slice_array is 128, not 127 (in case the change matters)

rts/motoko-rts/src/stabilization/layout.rs
```diff
@@ -349,39 +410,48 @@ pub unsafe fn deserialize<M: Memory>(
     stable_object: StableValue,
 ) -> Value {
     let tag = stable_memory.read::<StableTag>(stable_object.to_stable_address());
-    match tag.decode() {
-        StableObjectKind::Array => {
-            StableArray::deserialize(main_memory, stable_memory, stable_object)
+    let object_kind = tag.decode();
```
crusso, line 413: Most of the cases don't actually need the object_kind except to assert on the object_kind. 
The Array and Blob cases do need, but instead of decoding object_kind a second time, could also just be called with an appropriate Tag value instead. The additional cost is probably negligible, and the uniformity is nice,  but I guess it might be worth optimizing in future....

crusso (approved):
LGTM. There might be a bit of unnecessary overhead passing those mostly redundant object_kinds around but perhaps we shouldn't sweat the small stuff.

rts/motoko-rts/src/stabilization/deserialization.rs
luc-blaeser: Thanks for mentioning. I think the change is fine - it has no relevant impact (just a heuristic config, not too high, not too low - the value must be at least `TAG_ARRAY_SLICE_MIN` to be valid.)

rts/motoko-rts/src/memory.rs
luc-blaeser: Thanks, I addressed it in the lower PRs and merge the refactoring up.

rts/motoko-rts/src/memory.rs
luc-blaeser: Thanks, I addressed it in the lower PRs and merge the refactoring upwards.

rts/motoko-rts/src/stabilization/layout.rs
luc-blaeser: Indeed, the tag is only needed for specific allocations that have multiple sub-tags (like array, blobs and boxed numbers). E.g. `StableArray::deserialize` needs to call `alloc_array` that requires the specific tag.

luc-blaeser:
> LGTM. There might be a bit of unnecessary overhead passing those mostly redundant object_kinds around but perhaps we shouldn't sweat the small stuff.

Thank you very much for the review. Indeed, some `object_kinds` are redundant for specific cases in `deserialize`, but not for all cases. I guess this is the price of having a generic "mini-framework" for serialization/deserialization with a common trait for all the specific stable types.

rts/motoko-rts/src/stabilization/layout.rs
crusso: Ah, I missed the fact that these additional arguments are forced by the trait. No worries.

crusso:
(Still LGTM)

PR #4569, opened 2024-06-10
Merge Preparation: Merge Precise Object Tagging to Integrated EOP

luc-blaeser:
Preparing merging https://github.com/dfinity/motoko/pull/4544 in https://github.com/dfinity/motoko/pull/4488 (level 4 of PR stack)


rts/motoko-rts-tests/src/gc/compacting/mark_stack.rs
```diff
@@ -33,18 +33,28 @@ fn test_push_pop() {
...
     TAG_SOME,
     TAG_VARIANT,
-    TAG_BLOB,
+    TAG_BLOB_B,
+    TAG_BLOB_T,
+    TAG_BLOB_P,
+    TAG_BLOB_A,
     TAG_FWD_PTR,
-    TAG_BITS32,
+    TAG_BITS32_U,
```
crusso, line 55: Welcome back BITS32...

rts/motoko-rts/src/memory.rs
```diff
@@ -55,11 +55,12 @@ pub trait Memory {
 /// Note: After initialization, the post allocation barrier needs to be applied to all mutator objects.
 /// For RTS-internal blobs that can be collected by the next GC run, the post allocation barrier can be omitted.
 #[ic_mem_fn]
-pub unsafe fn alloc_blob<M: Memory>(mem: &mut M, size: Bytes<usize>) -> Value {
+pub unsafe fn alloc_blob<M: Memory>(mem: &mut M, tag: Tag, size: Bytes<usize>) -> Value {
+    debug_assert!(tag == TAG_BLOB_B || tag == TAG_BLOB_T || tag == TAG_BLOB_P || tag == TAG_BLOB_A);
```
crusso, line 59:
```suggestion
    debug_assert!(is_blob_tag(tag));
```
(might need an import)

rts/motoko-rts/src/memory.rs
```diff
@@ -69,13 +70,16 @@ pub unsafe fn alloc_blob<M: Memory>(mem: &mut M, size: Bytes<usize>) -> Value {
 /// Allocate a new array.
 /// Note: After initialization, the post allocation barrier needs to be applied to all mutator objects.
 #[ic_mem_fn]
-pub unsafe fn alloc_array<M: Memory>(mem: &mut M, len: usize) -> Value {
+pub unsafe fn alloc_array<M: Memory>(mem: &mut M, tag: Tag, len: usize) -> Value {
+    debug_assert!(
+        tag == TAG_ARRAY_I || tag == TAG_ARRAY_M || tag == TAG_ARRAY_T || tag == TAG_ARRAY_S
+    );
```
crusso, line 76:
```suggestion
    debug_assert!(is_base_array_tag(tag));
```
(might need an import)

rts/motoko-rts/src/stabilization/deserialization.rs
```diff
@@ -92,24 +93,16 @@ impl Deserialization {
             |context, field_address| {
                 *field_address = translate(context, *field_address);
             },
-            |context, slice_start, array| {
-                const SLICE_INCREMENT: usize = 127;
```
crusso, line 96: From previous PR, note that SLICE_INCREMENT is now 128 (not 127) in slice_array (I think).

src/codegen/compile_enhanced.ml
```diff
@@ -4212,7 +4231,12 @@ module Text = struct
     get_blob ^^ Blob.as_ptr_len env ^^
     E.call_import env "rts" "utf8_valid" ^^
     Bool.from_rts_int32 ^^
-    E.if1 I64Type (Opt.inject_simple env get_blob) (Opt.null_lit env)
+    E.if1 I64Type
+      (get_blob ^^ Blob.as_ptr_len env ^^
+       of_ptr_size env ^^ (* creates text blob *)
+       set_blob ^^
```
crusso, line 4237: (Maybe we could just use Blob.copy env Tagged.B Tagged.T (or similar) here... (and it compile_classical).)

crusso (approved):
LGTM!

(One thing that occurred to me: I wonder if we even need both MutBox and ObjInd in the stable format. It may be the case that MutBox is dedicated to mutable variables (on the stack, and in closure environments), while ObjInd is always used for mutable object fields.
If that is the case, we might not need a stable representation for MutBox, only for ObjInd. But I'm not 100% sure nor am I sure the compiler strictly maintains the distinction.)

rts/motoko-rts/src/memory.rs
luc-blaeser: Thanks!

rts/motoko-rts/src/memory.rs
luc-blaeser: Thank you!

rts/motoko-rts/src/stabilization/deserialization.rs
luc-blaeser: Thanks for mentioning. I think this is fine. Any value `>= TAG_ARRAY_SLICE_MIN` should be correct.

src/codegen/compile_enhanced.ml
luc-blaeser: Thank you for refactoring suggestion!

crusso:
(Still LGTM)

luc-blaeser:
> LGTM!
> 
> (One thing that occurred to me: I wonder if we even need both MutBox and ObjInd in the stable format. It may be the case that MutBox is dedicated to mutable variables (on the stack, and in closure environments), while ObjInd is always used for mutable object fields. If that is the case, we might not need a stable representation for MutBox, only for ObjInd.
> 
> But I'm not 100% sure nor am I sure the compiler strictly maintains the distinction. Indeed judging by this comment in compile.ml, perhaps not:
> 
> ```
>  The field1_data for mutable fields are pointers to either an ObjInd, or a
>  MutBox (they have the same layout). This indirection is a consequence of
>  how we compile object literals with `await` instructions, as these mutable
>  fields need to be able to alias local mutable variables.
> ```
> 
> )

This is a good point. I did some analysis. `ObjInd` is only used for mutable fields that are created during Candid-destabilization. Apart from that, mutable fields are always `MutBox`. I will prepare a PR for `master` to refactor the code and get rid of `ObjInd`, unifying it with `MutBox`.

PR #4570, opened 2024-06-11
docs: Remove ESM usage instructions

jessiemongeon1:


ghost (approved):

doc/md/stable-memory/stablememory.md
```diff
@@ -2,102 +2,20 @@
...
-```
-
-The auxiliary function `ensure(offset)` is used to grow `ExerimentalStableMemory` as necessary to accommodate more data. It computes the 64KiB page of a given offset and ensures enough pages have been allocated to guarantee that offset is within bounds.
-
-The shared `log(t)` function encodes its [`Text`](../base/Text.md) argument as a [`Blob`](../base/Blob.md), allocates enough stable memory to store it, and writes both the blob contents and its size at the next available offset in `ExperimentalStableMemory`, updating `base`.
-
-The shared `readLast(count)` query reads up to `count` messages from the log, traversing the log in reverse from `base`.
-
-Because `StableLog` allocates and maintains its (potentially large) log data directly in stable memory and uses just a small and fixed amount of storage for actual stable variables (here `base`), upgrading `StableLog` to a new implementation (perhaps to provide more functionality) should not consume too many cycles, regardless of the current size of the log.
-
```
luc-blaeser, line 100: I am not sure whether we should remove the packages mentioned below too, especially if they only operate on ESM and not on regions. I did not yet check this.

luc-blaeser (commented):
Thank you very much for this PR!

(Sorry, I just noted that the PR is merged. It is very good, no need to address my comment above)

PR #4572, opened 2024-06-17
chore: add deprecation warning to ESM methods

ghost:
Modifies ESM methods to add a compiler deprecation warning. Instructs users to use the [Stable Regions](https://internetcomputer.org/docs/current/motoko/main/stable-memory/stable-regions/#the-region-library) instead.


crusso:
Ok, turns out this doesn't work out of the box because deprecation comments are only honoured on module fields, not top-level declarations as in prim.mo.

I've hacked the compiler to allow this, but now think this might not be the way to go after all. All user level deprecation messages share the same error code, so it would be hard or ugly to reliably filter out the stable memory ones, and the message now complains about the Prim.stableMemoryXXX operations being used by ESM.

A more direct approach might be to split the  stable memory prims into a separate file, as we do for timers, and only include them when --experimental-stable-memory is enabled on the moc command line, opting in to the old feature and disabling it otherwise. Other code that imports ESM without specifying the flag will get a hard error about an unknown Prim.stableMemoryXXX rather than a warning. Will need to think about some more.

PR #4573, opened 2024-06-19
experiment: deprecate experimental stable memory 

crusso:
Leverages `/// @deprecate` comments to deprecate uses of ExperimentalStableMemory.

Adds moc flag `--experimental-stable-memory <n>` to control access to legacy stable memory:

n=0: remove related prims
n=1: warn on use of prims
n=2: allow use of prims sans warning.

Current default is n=1 to prod users to migrate their code to use preferred `Region.mo`.

The implementation is somewhat hacky and separates the ESM related prims into a prelude/stable-memory-api.mo with deprecations in place.
The prims are selectively added to the special `Prim` module depending on the value of the flag, with the deprecations disabled by a `@`-character replacement with ` ` when n = 2.

The net effect is that, by default, any code that directly or indirectly imports ESM will be reported, as will any code that directly access a stable memory primitive, alerting the user to potential vulnerabilities. The code will still compile, but in future we can prevent that by defaulting n to 0, not 1.

The not-so-nice aspect of this implementation is that:
1: importing ESM will generate several warnings, one for each operation.
2: The errors only indicate that ESM *may* be in use. 
3. They do not pinpoint which importing library is responsible for the warning.


src/mo_frontend/typing.ml
```diff
@@ -55,6 +55,7 @@ type env =
     check_unused : bool;
     used_identifiers : S.t ref;
     unused_warnings : unused_warnings ref;
+    reported_experimental_stable_memory : bool ref;
```
crusso, line 58: Delete me

crusso:
Replaced by #4575

PR #4575, opened 2024-06-21
experiment: deprecate experimental stable memory (improved)

crusso:
Leverages `/// @deprecate` comments to deprecate uses of ExperimentalStableMemory.

Adds moc flag `--experimental-stable-memory <n>` to control access to legacy stable memory:

- n<0: error on use of (ESM) prims.
- n=0: warn on use of prims.
- n=1: warning-less use of prims.

Current default is n=0 to encourage users to migrate their code to use preferred `Region.mo` without preventing compilation.

The implementation adds /// @deprecate M0199 comments Prim.mo and spots this specific deprecation during type-checking, reporting this specific violation just once per compilation unit, as an error, warning or not at all, depending on compiler setting.

The net effect is that, by default, any code that directly or indirectly imports ESM will be reported, as will any code that directly access a stable memory primitive, alerting the user to potential vulnerabilities. The code will still compile, but in future we can prevent that by defaulting n to 0, not 1.

The not-so-nice aspect of this implementation is that:
1: The errors only indicate that ESM *may* be in use. 
2: They do not pinpoint which importing library is responsible for the warning, but adding `/// @deprecate M0199` to corresponding ESM fields directly should achieve that too.


test/run-drun/ok/stable-mem-no-warning.tc.ok
```diff
@@ -0,0 +1,67 @@
+/home/crusso/motoko/src/moc: wrong argument '--force-gc'; option '--experimental-stable-memory' expects an integer.
```
crusso, line 1: FIX THIS

src/mo_def/compUnit.ml
```diff
@@ -61,7 +61,10 @@ let comp_unit_of_prog as_lib (prog : prog) : comp_unit =
       if as_lib
       then
         (* Deprecated syntax, see Typing.check_lib *)
-        let fs = List.map (fun d -> {vis = Public None @@ no_region; dec = d; stab = None} @@ d.at) ds' in
+        let fs = List.map (fun d ->
```
crusso, line 64: undent

src/pipeline/pipeline.ml
```diff
@@ -295,16 +295,18 @@ let prim_error phase (msgs : Diag.messages) =
...
   let parse = Parser.Incremental.parse_prog in
-
   match parse_with Lexer.mode_priv lexer parse prim_name with
   | Error es -> prim_error "parsing" es
   | Ok (prog, _ws) ->
     let open Syntax in
     let open Source in
     let senv0 = initial_stat_env in
-    let fs = List.map (fun d -> {vis = Public None @@ no_region; dec = d; stab = None} @@ d.at) prog.it in
+    let fs = List.map (fun d ->
```
crusso, line 306: indentation

src/pipeline/pipeline.ml
```diff
@@ -295,16 +295,18 @@ let prim_error phase (msgs : Diag.messages) =
   exit 1
 
 let check_prim () : Syntax.lib * stat_env =
-  let lexer = Lexing.from_string (Prelude.prim_module ~timers:!Flags.global_timer) in
+  let lexer = Lexing.from_string (Prelude.prim_module ~timers: !Flags.global_timer) in
```
crusso, line 298:
```suggestion
  let lexer = Lexing.from_string (Prelude.prim_module ~timers:!Flags.global_timer) in
```

src/prelude/prelude.ml
```diff
@@ -3,6 +3,6 @@ let internals = [%blob "prelude/internals.mo"]
 let timers_api = [%blob "prelude/timers-api.mo"]
 let prim_module' = [%blob "prelude/prim.mo"]
 let prim_module ~timers:required =
-  if required
-  then prim_module' ^ timers_api
-  else prim_module'
+   if required
+   then prim_module' ^ timers_api
+   else prim_module'
```
crusso, line 8:
```suggestion
  if required
  then prim_module' ^ timers_api
  else prim_module'
```

src/wasm-exts/customModuleEncode.ml
```diff
@@ -135,7 +135,8 @@ let encode (em : extended_module) =
       let source_code =
         try
           if filename = "prelude" then Prelude.prelude else
-          if filename = "prim" then Prelude.prim_module ~timers:!Mo_config.Flags.global_timer else
+          if filename = "prim" then Prelude.prim_module  ~timers: !Mo_config.Flags.global_timer
+          else
```
crusso, line 139:
```suggestion
          if filename = "prim" then Prelude.prim_module  ~timers: !Mo_config.Flags.global_timer else
```

src/mo_frontend/typing.ml
```diff
@@ -1368,9 +1385,8 @@ and infer_exp'' env exp : T.typ =
         "cannot infer type of forward field reference %s"
         id.it
     | t ->
-      Option.iter
-        (warn env exp.at "M0154" "field %s is deprecated:\n%s" id.it)
-        (T.lookup_val_deprecation id.it tfs);
+      if not env.pre then
```
luc-blaeser, line 1388: Is the purpose of `not env.pre` to avoid some false cascading warnings/errors when there exist other type errors? (Just curious, as the condition is new for this deprecation warning.)

src/prelude/prim.mo
```diff
@@ -335,7 +335,7 @@ func cyclesRefunded() : Nat {
 };
 
 func cyclesAccept<system>(amount : Nat) : Nat {
-  (prim "cyclesAccept" : <system>Nat -> Nat)(amount);
+  (prim "cyclesAccept" : Nat -> Nat)(amount);
```
luc-blaeser, line 338: Some cleanup :)

test/run/ok/deprecated.tc.ok
```diff
@@ -10,9 +10,9 @@ deprecated.mo:19.8-19.11: warning [M0154], field f is deprecated:
 deprecated function
 deprecated.mo:21.13-21.16: warning [M0154], type field T is deprecated:
 also types
-deprecated.mo:23.11-23.12: warning [M0154], type field f is deprecated:
+deprecated.mo:23.11-23.12: warning [M0154], field f is deprecated:
```
luc-blaeser, line 13: Nice!

luc-blaeser (approved):
Very nice PR! Thank you.

src/mo_frontend/typing.ml
crusso: I noticed we were getting duplicate warnings without this, probably because the warning was generated on both the pre and checking passes.

src/prelude/prim.mo
crusso: Yep. Oh, maybe there is another one or two. Lemme check...

crusso: Done

luc-blaeser:
(It seems that some expected test result needs to be updated...)

PR #4576, opened 2024-06-24
Code Refactoring: Unifying `ObjInd` and `MutBox`

luc-blaeser:
Currently, mutable fields are represented as `ObjInd` or `MutBox` in an interchangeable format. 
* For new object literals, mutable fields are allocated as `MutBox`.
* Destabilized object from the Candid stable format have their mutable fields represented as `ObjInd`.

Moreover, `ObjInd` and `MutBox` share the same physical memory layout, except that the object tag differs.

This PR gets rid of `ObjInd` and uniformly represents all mutable fields as `MutBox`.


rts/motoko-rts-tests/src/gc/compacting/mark_stack.rs
```diff
@@ -33,9 +33,8 @@ fn test_push_pop() {
         .unwrap();
 }
 
-static TAGS: [Tag; 14] = [
+static TAGS: [Tag; 13] = [
```
crusso, line 36: Lucky 13

crusso:
Good experiment!

luc-blaeser:
> Good experiment!

Thank you! Do you think we should merge it?

crusso (approved):
Yes, I think we can merge this. And then also remove the graph copy version of MutBox!

Unless @ggreif can think of some other reason we have this distinction?

src/codegen/compile.ml
```diff
@@ -7870,7 +7870,7 @@ module MakeSerialization (Strm : Stream) = struct
       | Mut t ->
         read_alias env (Mut t) (fun get_arg_typ on_alloc ->
           let (set_result, get_result) = new_local env "result" in
-          Tagged.obj env Tagged.ObjInd [ compile_unboxed_const 0l ] ^^ set_result ^^
+          MutBox.alloc env ^^ set_result ^^
```
crusso, line 7873: Does MutBox.alloc contain a valid vanilla value on allocation? Should we worry if doesn't?

luc-blaeser:
> Yes, I think we can merge this. And then also remove the graph copy version of MutBox!
> 
> Unless @ggreif can think of some other reason we have this distinction?

Absolutely, I can't wait to also remove the `ObjInd` of graph copy :)

ggreif (approved):
LGTM!

luc-blaeser:
Thank you Gabor and Claudio for the review!

PR #4580, opened 2024-06-27
experiment: mechanism for interacting with external Wasm components

rvanasa:
This PR introduces a standardized import which can be used for cross-language interoperability via the [Wasm component model](https://component-model.bytecodealliance.org/) specification. 

**Progress:**
- [x] Import `component.call` in compiled Wasm module
- [x] Add `componentCall` prim function
- [x] Enable this feature with the `-import-component` flag (e.g. `moc -wasi-system-api -import-component ...`)
- [x] Use new error code (`M0200`) when trying to use this feature without the `-import-component` flag
- [x] Add unit test
- [x] Replace `Nat32` arg with `Blob`
- [x] Replace `Nat32` return type with `Blob`

**Next steps**
- Generalize `wit:component:call` to support arbitrary component imports?

Note that this only works in runtimes with support for the component model:
* [Wasmtime](https://github.com/bytecodealliance/wasmtime#readme) works out of the box.
* [Wasmi](https://github.com/wasmi-labs/wasmi#readme) can be used in Rust canisters via [`wasm_component_layer`](https://github.com/DouglasDwyer/wasm_component_layer).

An alternate approach would be to enable passing the names of arbitrary Wasm imports. While this seems like it would be beneficial at first glance, the component model architecture tends to favor a pattern of defining a standard interface (for example, the [WASI CLI world](https://github.com/WebAssembly/wasi-cli#readme)) and then "plugging" exports into corresponding imports (see [WAC](https://github.com/bytecodealliance/wac#readme) for how this specifically works). 

The current approach follows this pattern with the following WIT world:

```wit
package motoko:component;

world motoko {
    import component: interface {
        call: func(value: list<u8>) -> list<u8>;
    }
}
```

It's possible to plug a Rust component into this interface using something similar to the following WAC composition:

```wit
package motoko:composition;

let rust = new rust:component {};

let motoko = new motoko:component {
    component: rust,
    ...
};

export motoko.run;
```


crusso:
@rvanasa I took a look and got CI working for you in #4584.

The problem was that moc -t doesn't even enter compile.ml, so you weren't generating the error when just running moc -t (as the tests in test/fail (which are mostly type checker test do).

Feel free to merge into your branch or ignore...

PR #4584, opened 2024-06-28
experiment: reimplement M0200 for #4580

crusso:
Builds on #4580.

Moves M0200 error detection from compile.ml to typing.ml, using special #deprecation M0200  in Prim.ml (the same trick we use to selectively deprecate stable memory prims).


rvanasa (approved):
LGTM; thanks!

PR #4585, opened 2024-06-28
EOP: Support Unknown Main Memory Capacity in 64-Bit

luc-blaeser:
# Support Unknown Main Memory Capacity in 64-Bit

While in 32-bit memory, the entire address space is available for allocations, the allocatable memory in 64-bit address space is limited by the IC configuration. This limit may be increased in future IC releases.
The canister has no upfront knowledge of the maximum allocatable Wasm main memory, as there is no IC API call to query the main memory limit.

This PR implements a mechanism to deal with an unknown and dynamically increasable main memory capacity offered by the IC. This is needed in two cases:
* **GC Reserve** (strict): The RTS ensures sufficient free space to allow garbage collection at all times, even if the heap is full. For this purpose, the runtime system already pre-allocates the reserve, to be sure that the reserve is available despite the unknown capacity. As an optimization, this pre-allocation is skipped when the memory demand including the GC reserve is below a guaranteed minimum Wasm memory limit of the IC, e.g. 4GB or 6GB. The reserve is relaxed for queries, composite queries, and upgrades (focussing on graph-copy stabilization).
* **GC Scheduling** (heuristic): The GC schedules at high frequency when memory is becoming scarce. For this purpose, the GC maintains an assumption of the minimum memory limit and probes the supposed limit when the heap size approaches this limit. If the allocation succeeds, the assumed limit is increased. Otherwise, the critical high-frequency GC scheduling rate is activated.

In both cases, the runtime system tries to reduce Wasm memory allocations as much as possible, i.e. not pre-allocating memory for small heap sizes, and not probing an allocation in certain memory ranges by assuming that the IC only offers main memory of a certain granularity, e.g. multiples of 2GB. To save instructions, the critical GC scheduling is only activated when reaching the actual memory limit. Moreover, the mechanism can handle an increased memory capacity at runtime, e.g. when IC is upgraded to a new release with a higher memory limit.

This has been tested by an IC branch implementing a memory limit of 6GB.


rts/motoko-rts/src/gc/incremental/scheduling.rs
```diff
@@ -0,0 +1,86 @@
...
+//!
+//! Distinction of three scheduling modes:
+//! * Critical: The heap exceeds 80% of the memory capacity (using capacity probing):
+//!   - Schedule a GC start with a frequency of 1% of the heap growth.
+//! * Medium: The heap exceeds 50% of the memory capacity:
+//!   - Schedule a GC start with a frequency of 35% of the heap growth.
+//! * Low: The heap is below 50% of the memory capacity.
+//!   - Schedule a GC start with a frequency of 65% of the heap growth.
+//!
+//! A heuristics for capacity probing is used to determine minimum amount of memory capacity.
```
luc-blaeser, line 12:
```suggestion
//! A heuristics for capacity probing is used to determine the minimum amount of memory capacity.
```

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -6,15 +6,23 @@ use crate::gc::incremental::memory_reserve;
...
 /// `Memory` implementation allocates in Wasm heap with Wasm `memory.grow` instruction.
 pub struct IcMemory;
 
+/// Probe the memory capacity beyond this limit by pre-allocating the reserve.
+/// This is necessary because there is no function to query the IC's main memory capacity for 64-bit.
+/// It is assumed that the capacity is at least the 32-bit space of 4GB, minus the last reserved
+/// Wasm page (for Rust call stack overflow detection, see `compile.ml`).
+const GUARANTEED_MEMORY_CAPACITY: usize = 4 * GB - WASM_PAGE_SIZE.0;
+
+/// Assumption (not correctness-critical): The IC offers main memory in multiple of 2 GB.
```
luc-blaeser, line 21:
```suggestion
/// Assumption (not correctness-critical): The IC offers main memory in multiples of 2 GB.
```

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -25,25 +33,23 @@ impl Memory for IcMemory {
...
+        debug_assert!(reserve <= GUARANTEED_MEMORY_CAPACITY);
+        let memory_demand = if ptr > GUARANTEED_MEMORY_CAPACITY - reserve {
+            // Detect overflow of `ptr + reserve`.
+            if ptr > usize::MAX - reserve {
                 rts_trap_with("Cannot grow memory");
             }
-        }
+            // The reserve will be pre-allocated as a way to check the main memory capacity.
+            // As the reserve can be substantial, this is only done when memory demand has
+            // grown beyond `GUARANTEED_MEMORY_CAPACITY`.
```
luc-blaeser, line 45:
```suggestion
            // As the reserve can be proportionally large for small heaps, this is only done when 
            // the memory demand exceeds `GUARANTEED_MEMORY_CAPACITY`.
```

rts/motoko-rts/src/memory/ic.rs
```diff
@@ -66,3 +72,47 @@ pub unsafe extern "C" fn get_heap_size() -> Bytes<usize> {
 pub unsafe extern "C" fn get_max_live_size() -> Bytes<usize> {
     crate::gc::incremental::get_max_live_size()
 }
+
+/// Supposed minimum memory capacity used for GC scheduling heuristics.
+/// The result may increase after time. This is because the actual capacity is
+/// not known upfront and can only derived by memory allocation probing.
```
luc-blaeser, line 78:
```suggestion
/// not known in advance and can only derived by memory allocation probing.
```

PR #4587, opened 2024-07-01
Refine the Memory Reserve Handling in the Allocator

luc-blaeser:
Refining the allocator’s memory reserve for better maintainability.

The current allocator is based on the assumption that the memory reserve is at least of a WASM page size (64 KB) to guard against overflows of the shadow call stack (used in Rust and in the compiler backend).
This assumption is always met, i.e. the RTS is already safe:
* In the incremental GC: The memory reserve is at least 256 MB. When the GC is inactive, the reserve is increased by additional 640 MB.
* In the classical GCs: The reserve is constantly 256 MB.

This change relaxes this assumption and directly guards the last page in the allocator, independent of the GC’s memory reserve. This is only to increase robustness for future changes, where different memory reserves may be configured, potentially even below a page size.

### DIFF
--- a/rts/motoko-rts/src/memory/ic.rs
+++ b/rts/motoko-rts/src/memory/ic.rs
@@ -39,17 +39,23 @@ pub struct IcMemory;
 /// Page allocation. Ensures that the memory up to, but excluding, the given pointer is allocated.
 /// Ensure a memory reserve of at least one Wasm page depending on the canister state.
 /// `memory_reserve`: A memory reserve in bytes ensured during update and initialization calls.
-//  For use by queries and upgrade calls. The reserve may vary depending on the GC and the phase of the GC.
+/// For use by queries and upgrade calls. The reserve may vary depending on the GC and the phase of the GC.
 unsafe fn grow_memory(ptr: u64, memory_reserve: usize) {
     const LAST_PAGE_LIMIT: usize = 0xFFFF_0000;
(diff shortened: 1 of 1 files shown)


luc-blaeser:
> Comparing from [0d36eff](https://github.com/dfinity/motoko/commit/0d36eff4c567b53ee0a6c654ff448ace8263ffc4) to [5867e19](https://github.com/dfinity/motoko/commit/5867e19671f54018ee68d9adaf5cc197d1f0dbe0): In terms of gas, 1 tests regressed and the mean change is +5.5%. In terms of size, no changes are observed in 5 tests.

Hm, strange, will need to analyze this. Probably the change from a combined if to two separate ifs...

luc-blaeser:
> > Comparing from [0d36eff](https://github.com/dfinity/motoko/commit/0d36eff4c567b53ee0a6c654ff448ace8263ffc4) to [5867e19](https://github.com/dfinity/motoko/commit/5867e19671f54018ee68d9adaf5cc197d1f0dbe0): In terms of gas, 1 tests regressed and the mean change is +5.5%. In terms of size, no changes are observed in 5 tests.
> 
> Hm, strange, I analyzed it: The degrading test is `dao.ml`. And it is due to different inlining behavior of `grow_memory`. I will adjust it.

**Update**: I opened a separate PR with a slight inlining optimization w.r.t. the IC cost model. https://github.com/dfinity/motoko/pull/4589

crusso (approved):
LGTM!

PR #4589, opened 2024-07-01
Refine the Memory Reserve Handling in the Allocator (Optimized)

luc-blaeser:
(Optimized version of https://github.com/dfinity/motoko/pull/4587)

Refining the allocator’s memory reserve for better maintainability.

The current allocator is based on the assumption that the memory reserve is at least of a WASM page size (64 KB) to guard against overflows of the shadow call stack (used in Rust and in the compiler backend).
This assumption is always met, i.e. the RTS is already safe:
* In the incremental GC: The memory reserve is at least 256 MB. When the GC is inactive, the reserve is increased by additional 640 MB.
* In the classical GCs: The reserve is constantly 256 MB.

This change relaxes this assumption and directly guards the last page in the allocator, independent of the GC’s memory reserve. This is only to increase robustness for future changes, where different memory reserves may be configured, potentially even below a page size.


luc-blaeser:
Closing in favor of https://github.com/dfinity/motoko/pull/4590

PR #4594, opened 2024-07-03
DD-221: document option blocks, record extension, `system` capabilities and `let`/`else`

ggreif:
This is covering tasks
- DD-489
- DD-128
- DD-127
- DD-488

and adds an improvement to note pointing at security recommendations


doc/md/getting-started/basic-concepts.md
```diff
@@ -199,3 +213,19 @@ assert n % 2 == 0; // traps when n not even
...
+
+Naturally, the programmer must be careful granting send privileges to third-party libraries (resp. functions therein) and review for potential security breaches if she choses to do so.
+
+There is a special class of functions, however, which can be called from code that does not possess send capability, but registers a callback that does. This kind of *capability elevation* is possible when adding timers (e.g. `setTimer` in `base`). To harden the `actor` against supply-chain attacks (malicious sends masquerading behind a capability-starved call interface to third-party code), Motoko allows to declare functions that can potentially lead to capability elevation to require a `system` capability.
+
+The `system` capability originates from the top-level actor and can be passed to functions (that expect it) by specifying a pseudo-type parameter `<system, ...>` which must appear at the call site. Similarly, functions demanding the `system` capability declare it in their signature:
```
jessiemongeon1, line 225:
```suggestion
### The `system` capability

Smart contracts (i.e., Motoko `actor`s) may generally contain assets that correspond to real-world value, the loss of which can be detrimental. Many such assets are mobile and can be transferred through message sending. In order for a library to be able to send a message, the corresponding function must be imported and possess an `async` (or `async*`) type. Calling such functions is only possible when the send capability is given by the callee, also having such a type.

Naturally, the programmer must be careful when granting send privileges to third-party libraries (resp. functions therein) and reviewing for potential security breaches if they choose to do so.

There is a special class of functions, however, that can be called from code that does not possess send capability but registers a callback that does. This kind of *capability elevation* is possible when adding timers (e.g. `setTimer` in `base`). To harden the `actor` against supply-chain attacks (malicious sends masquerading behind a capability-starved call interface to third-party code), Motoko allows to declare functions that can potentially lead to capability elevation to require a `system` capability.

The `system` capability originates from the top-level actor and can be passed to functions that expect it by specifying a pseudo-type parameter `<system, ...>` which must appear at the call site. Similarly, functions demanding the `system` capability declare it in their signature:
```

doc/md/writing-motoko/pattern-matching.md
```diff
@@ -100,6 +100,22 @@ Some types contain just a single value. We call these singleton types. Examples
...
+
+## Refutable patterns and dealing with non-matching data
+
+Numerous times one is only interested in specially formed data and desires to summarily handle all non-matching forms. The `let`-`else` construct is designed precisely for this pupose. Whereas the regular destructuring `let` allows to focus on a single given pattern, it invariably traps if the right hand side data doesn't match it (this is also warned at compile-time). The `else` clause gives the programmer a way to deal with refuted matches, such as bailing out of the processing or logging a message before trapping. As such, `let`-`else` is similar to a two-`case` `switch` in a compact form, that additionally doesn't force the indentation of the processing logic following it.
+
+The below example illustrates how you can write a non-indenting `if`-`else` by resorting to a `let`-`else` in your code:
```
jessiemongeon1, line 110:
```suggestion
To detect the possibility of such runtime failures, the Motoko compiler checks for the exhaustiveness of pattern matching by keeping track of the covered shape of the scrutinee. The compiler issues a warning for any non-covered scrutinees. Motoko even constructs a helpful example of a scrutinee that is not matched. A useful by-product of the exhaustiveness check is that it identifies and warns the developer about dead or redundant alternatives that can never be matched.

## Refutable patterns and dealing with non-matching data

Developers may only be interested in specially formed data with a desire to directly handle all non-matching forms. The `let`-`else` construct is designed precisely for this purpose. Whereas the regular destructuring `let` allows to focus on a single given pattern, it invariably traps if the right-hand side data doesn't match it, which is warned at compilation. The `else` clause gives the programmer a way to deal with refuted matches, such as bailing out of the process or logging a message before trapping. As such, `let`-`else` is similar to a two-`case` `switch` in a compact form that additionally doesn't force the indentation of the processing logic following it.

The below example illustrates how you can write a non-indenting `if`-`else` by resorting to a `let`-`else` in your code:
```

doc/md/writing-motoko/pattern-matching.md
```diff
@@ -100,6 +100,22 @@ Some types contain just a single value. We call these singleton types. Examples
...
+``` motoko
+let true = isLoggedIn(customer) else return;
+// process message for logged-in customer
+```
+
+The expression (or block) following the `else` must be of type `None`, thus signifying that its execution trajectory mustn't contribute to the code immediately following the `let` declaration.
+
+## Option blocks for streamlined processing of optional data
+
+Pattern matching on optional data (of type `?T`) is a preferred technique of avoiding the dreaded `null`-exeption problems known from other languages. However `switch`ing on several options can lead to tedious coding and deeply nested sources. To remedy these problems, Motoko provides *option blocks* (`do ? { ... }`) that allow safe unwrapping of options using a postfix `!` operator. Every use of the `!` in the block corresponds to a `switch` on some option, with the additional short-cutting behaviour that if `!` is applied to a `null` value, the entire block stops evaluation and immediately returns `null`.
```
jessiemongeon1, line 121:
```suggestion
The expression (or block) following the `else` must be of type `None` signifying that its execution trajectory mustn't contribute to the code immediately following the `let` declaration.

## Option blocks for streamlined processing of optional data

Pattern matching on optional data (of type `?T`) is a preferred technique for avoiding the dreaded `null`-exeption problems known from other languages. However,' switch'ing  on several options can lead to tedious coding and deeply nested sources. To remedy these problems, Motoko provides *option blocks* (`do ? { ... }`) that allow safe unwrapping of options using a postfix `!` operator. Every use of the `!` in the block corresponds to a `switch` on some option, with the additional short-cut behavior that if `!` is applied to a `null` value, the entire block stops evaluation and immediately returns `null`.
```

jessiemongeon1 (approved):

doc/md/getting-started/basic-concepts.md
```diff
@@ -103,6 +111,12 @@ A declaration list is not itself an expression, so you cannot declare another va
 
 **Block expressions** can be formed from a list of declarations by enclosing it with matching curly braces. Blocks are only allowed as sub-expressions of control flow expressions like `if`, `loop`, `case`, etc. A block expression produces a value and, when enclosed in parentheses, can occur within some larger, compound expression.
 
+:::note
+
+A particular form of blocks are provided for convenience when processing data that may be missing or incomplete. These are described under [option blocks](../writing-motoko/pattern-matching.md#option-blocks).
```
ggreif, line 116: Does this link work? Or do I need to add an anchor? But how?

doc/md/writing-motoko/pattern-matching.md
ggreif: Maybe the word I was searching for is _short-circuiting_. @crusso?

doc/md/writing-motoko/pattern-matching.md
```diff
@@ -100,6 +100,22 @@ Some types contain just a single value. We call these singleton types. Examples
...
+``` motoko
+let true = isLoggedIn(customer) else return;
+// process message for logged-in customer
+```
+
+The expression (or block) following the `else` must be of type `None` signifying that its execution trajectory mustn't contribute to the code immediately following the `let` declaration.
+
+## Option blocks for streamlined processing of optional data
+
+Pattern matching on optional data (of type `?T`) is a preferred technique for avoiding the dreaded `null`-exeption problems known from other languages. However,' switch'ing  on several options can lead to tedious coding and deeply nested sources. To remedy these problems, Motoko provides *option blocks* (`do ? { ... }`) that allow safe unwrapping of options using a postfix `!` operator. Every use of the `!` in the block corresponds to a `switch` on some option, with the additional short-cut behavior that if `!` is applied to a `null` value, the entire block stops evaluation and immediately returns `null`.
```
ggreif, line 121:
```suggestion
Pattern matching on optional data (of type `?T`) is a preferred technique for avoiding the dreaded `null`-exeption problems known from other languages. However, `switch`-ing  on several options can lead to tedious coding and deeply nested sources. To remedy these problems, Motoko provides *option blocks* (`do ? { ... }`) that allow safe unwrapping of options using a postfix `!` operator. Every use of the `!` in the block corresponds to a `switch` on some option, with the additional short-circuiting behavior that if `!` is applied to a `null` value, the entire block stops evaluation and immediately returns `null`.
```

doc/md/getting-started/basic-concepts.md
ggreif, line 116:
```suggestion
A particular form of blocks are provided for convenience when processing data that may be missing or incomplete. These are described under [option blocks](../writing-motoko/pattern-matching.md#option-blocks-for-streamlined-processing-of-optional-data).
```

doc/md/getting-started/basic-concepts.md
```diff
@@ -57,7 +57,7 @@ For precise language definitions of primitive and non-primitive values, see the
 
 Objects are aggregate data made from *labeled* constituent data. In their most general form, objects can contain named values (`let` and `var`) as well as methods (`func`) that act on them. Objects are written with the leading keyword `object` followed by an optional name and the block comprising its constituents. Only `public` constituents contribute to the object's type.
 
-In many cases objects are used as simply containers of data, which we refer to as *records*. When building records, Motoko has a simplified syntax to offer, where semicolon-separated labelled fields are placed in braces. The labels are identifiers (with a leading `var` when the field is mutable) followed by `=` and the initial value. All fields are public, and contribute to the record's type.
+In many cases, objects are used as simple containers of data, which are referred to as *records*. When building records, Motoko has a simplified syntax to offer where semicolon-separated label fields are placed in braces. The labels are identifiers (with a leading `var` when the field is mutable), followed by `=` and the initial value. All fields are public and contribute to the record's type.
```
ggreif, line 60: "labeled fields"?

ggreif: "named fields"?

ggreif:
@jessiemongeon1 thanks for the improvements, I accepted most of them!

doc/md/getting-started/basic-concepts.md
jessiemongeon1: Named fields would be best IMO

ggreif, line 60:
```suggestion
In many cases, objects are used as simple containers of data, which are referred to as *records*. When building records, Motoko has a simplified syntax to offer where semicolon-separated named fields are placed in braces. The labels are identifiers (with a leading `var` when the field is mutable), followed by `=` and the initial value. All fields are public and contribute to the record's type.
```

ggreif (commented):
Claudo will add some more descriptive text in another PR.

doc/md/getting-started/basic-concepts.md
ggreif: Done!

doc/md/writing-motoko/pattern-matching.md
```diff
@@ -100,6 +100,22 @@ Some types contain just a single value. We call these singleton types. Examples
...
+## Refutable patterns and dealing with non-matching data
+
+Developers may only be interested in specially formed data with a desire to directly handle all non-matching forms. The `let`-`else` construct is designed precisely for this purpose. Whereas the regular destructuring `let` allows to focus on a single given pattern, it invariably traps if the right-hand side data doesn't match it, which is warned at compilation. The `else` clause gives the programmer a way to deal with refuted matches, such as bailing out of the process or logging a message before trapping. As such, `let`-`else` is similar to a two-`case` `switch` in a compact form that additionally doesn't force the indentation of the processing logic following it.
+
+The below example illustrates how you can write a non-indenting `if`-`else` by resorting to a `let`-`else` in your code:
+
+``` motoko
+let true = isLoggedIn(customer) else return;
+// process message for logged-in customer
+```
```
crusso, line 115: This example isn't very motivating since you could just use 
```
is (!IsLoggedIn(customer)) return';
... 
```
What about something that binds some identifiers? Along the lines of (unchecked):
```
let (?record) = map.get(map, customer) else return Error.error("no such customer")
````

File: doc/md/getting-started/basic-concepts.md [hunk 8]
```diff
@@ -103,6 +111,12 @@ A declaration list is not itself an expression, so you cannot declare another va
 
 **Block expressions** can be formed from a list of declarations by enclosing it with matching curly braces. Blocks are only allowed as sub-expressions of control flow expressions like `if`, `loop`, `case`, etc. A block expression produces a value and, when enclosed in parentheses, can occur within some larger, compound expression.
 
+:::note
```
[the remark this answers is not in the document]
crusso: We actually alread have an example of option blocks here: 
See also http://localhost:3000/docs/writing-motoko/control-flow#option-blocks-and-null-breaks

doc/md/writing-motoko/pattern-matching.md
ggreif: Yeah, this is "parse, don't validate". Good point.

crusso: I'll  use this from Claude...

``` motoko
func getName(optionalName : ?Text) : Text {
  let ?name = optionalName else return "Unknown";
  name
}
```

PR #4597, opened 2024-07-03
Make CI Performance Comparison Test Deterministic

luc-blaeser:
One performance test case, `dao.mo`, uses heartbeats that lead to non-deterministic results in instruction costs, as seen in https://github.com/dfinity/motoko/pull/4587 and https://github.com/dfinity/motoko/pull/4596.

This PR fixes this issue, by calling the former heartbeat function explicitly in the performance test. 

Special thanks to @crusso for giving the hint to this cause.


luc-blaeser:
> Comparing from [0d36eff](https://github.com/dfinity/motoko/commit/0d36eff4c567b53ee0a6c654ff448ace8263ffc4) to [8833b59](https://github.com/dfinity/motoko/commit/8833b59d0eb7faf722310fe3de35da6b0d1ebda7): The produced WebAssembly code seems to be completely unchanged.

This is only because the compiler/RTS binary did not change. Otherwise, the change would be like https://github.com/dfinity/motoko/pull/4596#issuecomment-2206220131

```
Comparing from https://github.com/dfinity/motoko/commit/0d36eff4c567b53ee0a6c654ff448ace8263ffc4 to https://github.com/dfinity/motoko/commit/206f93588a7b32bd68933b9732a371f39ed31dff:
In terms of gas, 1 tests improved and the mean change is -47.8%.
In terms of size, 1 tests improved and the mean change is -0.2%.
```

crusso (approved):

PR #4600, opened 2024-07-04
Allow `try` without `catch`

ggreif:
_UPDATE_: I reverted a lot of it, see commit 1a7b61df1 for an explanation.

see #4507

TODO:
- [x] check that no IR-scrutinising happens — see below
- [ ] use `catch_opt` consistently
- [x] suspicious: IR-interpreter and `await.ml` didn't need changes? — see below

In the IR when none of the `catch` alternatives match, then an artificially added
``` OCaml
@ [{ it = { pat = varP error; exp = varE f -*- varE error };
```
re-throws the `Error`. So df9389cee shortcuts the creation of a silly do-nothing failure continuation when the list of alternatives is empty. It would just re-throw unconditionally anyway!

Optimisation opportunity: only emit this when all of the patterns are refutable! — Done: 2329069a1


src/mo_def/syntax.ml
```diff
@@ -193,7 +193,7 @@ and exp' =
   | AnnotE of exp * typ                        (* type annotation *)
   | ImportE of (string * resolved_import ref)  (* import statement *)
   | ThrowE of exp                              (* throw exception *)
-  | TryE of exp * case list * exp option       (* catch exception / finally *)
+  | TryE of exp * case option * exp option     (* catch exception / finally *)
```
ggreif, line 196: Well, I am not so confident about whether this change is worth it. @crusso what do you think?

PR #4601, opened 2024-07-05
EOP: Using a Dedicated Stable Option Type in Persistent State

luc-blaeser:
# EOP: Using a Dedicated Stable Option Type in Persistent State
Related to enhanced orthogonal persistence, see https://github.com/dfinity/motoko/pull/4193.

As the Rust-native `Option` type may change its implementation and memory representation in future Rust releases, the runtime system for enhanced orthogonal persistent should use a dedicated option type in the internal persistent state. This dedicated option type has a long-term fixed representation (in C-layout convention) and is used in the garbage collector state that is again part of the persistent metadata.

### DIFF
--- a/rts/motoko-rts/src/stable_option.rs
+++ b/rts/motoko-rts/src/stable_option.rs
@@ -1,0 +1,55 @@
+//! Option type with a fixed C-representation.
+//! This is used as a long-term option type in the persistent metata,
+//! such as the garbage collector state.
+//! As the native Rust type may change in future Rust releases,
+//! the Motoko runtime system uses its own dedicated `Option` type
+//! with a fixed long-term representation. This is used in the persistent
+//! metadata, such as in the garbage collector state.
+//! See:
+//! * `gc::incremental::State`
+//! * `gc::partitioned_heap::PartitionedHeapIterator`
+
+/// Stable option type used for orthogonal persistent state.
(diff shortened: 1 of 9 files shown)


rts/motoko-rts/src/stable_option.rs
```diff
@@ -0,0 +1,55 @@
+//! Option type with a fixed C-representation.
+//! This is used as a long-term option type in the persistent metata,
+//! such as the garbage collector state.
+//! As the native Rust type may change in future Rust releases,
+//! the Motoko runtime system uses its own dedicated `Option` type
+//! with a fixed long-term representation. This is used in the persistent
+//! metadata, such as in the garbage collector state.
```
luc-blaeser, line 7:
```suggestion
//! Customized option type with a fixed C-representation.
//! As the native Rust type may change in future Rust releases,
//! the Motoko runtime system uses its own dedicated `StableOption` 
//! type with a fixed long-term representation. This is used in the 
//! persistent metadata, such as in the garbage collector state.
```

crusso (approved):
Seems like a sensible precaution (https://github.com/rust-lang/rfcs/blob/master/text/2195-really-tagged-unions.md).

luc-blaeser:
Thank you for the review, Claudio!

PR #4602, opened 2024-07-05
EOP: Relaxed Upgrade Compatibility Check

luc-blaeser:
# Relaxed Upgrade Compatibility Check
This is used by `dfx` to check the upgrade compatibility with regard to the stable variables.

**Note**: With EOP, the Motoko RTS implements an atomic and authoritative upgrade compatibility check that cannot be bypassed. Hence, the `dfx` check only provides an additional help for the user. 

## Relaxed Rules for Stable Variables
Enhanced orthogonal persistence offers more liberal rules for more flexible upgrades:
- Mutability of stable fields can be changed because they are never aliased.
- Stable fields can be dropped, however, with a warning of potential data loss. For this, we give up the transitivity property of upgrades.

**Note**: Upgrade transitivity means that an upgrade from a program `A` to `B` and then from `B` to `C` has the same effect as directly upgrading from `A` to `C`. Now, if `B` discards a field and `C` re-adds it, this transitivity is no longer maintained. However, rigorous upgrade transitivity was also not guaranteed before, since `B` may contain initialization logic or pre-/post-upgrade hooks that alter the stable data.

### DIFF
--- a/src/mo_frontend/stability.ml
+++ b/src/mo_frontend/stability.ml
@@ -49,7 +49,7 @@ let match_stab_sig tfs1 tfs2 : unit Diag.result =
       | tf1 :: tfs1', [] ->
         (* dropping fields is allowed, but with a warning *)
         warning_discard s tf1;
-        Some ()
+        go tfs1' []
       | tf1::tfs1', tf2::tfs2' ->
         (match Type.compare_field tf1 tf2 with
          | 0 ->
--- a/src/mo_frontend/stability.ml
+++ b/src/mo_frontend/stability.ml
@@ -47,7 +47,7 @@ let match_stab_sig tfs1 tfs2 : unit Diag.result =
       | [], _ ->
         Some () (* no or additional fields ok *)
       | tf1 :: tfs1', [] ->
-        (* dropping fields is allowed, but with a warning *)
+        (* dropped field is allowed with warning, recurse on tfs1' *)
         warning_discard s tf1;
         go tfs1' []
       | tf1::tfs1', tf2::tfs2' ->
@@ -57,7 +57,7 @@ let match_stab_sig tfs1 tfs2 : unit Diag.result =
               error_sub s tf1 tf2;
             go tfs1' tfs2'
          | -1 ->
-           (* dropped field is allowed with warning, recurse of tfs1' *)
+           (* dropped field is allowed with warning, recurse on tfs1' *)
            warning_discard s tf1;
            go tfs1' tfs2 
          | _ ->
--- a/src/mo_frontend/stability.ml
+++ b/src/mo_frontend/stability.ml
@@ -22,7 +22,7 @@ let error_sub s tf1 tf2 =
 
 (* Relaxed rules with enhanced orthogonal persistence for more flexible upgrades.
    - Mutability of stable fields can be changed (because they are never aliased).
-   - Stable fields can be dropped (abdandoning the transitivity property of upgrades).
+   - Stable fields can be dropped (abandoning the transitivity property of upgrades).
 
    Upgrade transitivity means that an upgrade from a program A to B and then from B to C 
    should have the same effect as directly upgrading from A to C. If B discards a field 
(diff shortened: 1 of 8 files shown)


src/mo_frontend/stability.ml
```diff
@@ -29,37 +29,37 @@ let error_sub s tf1 tf2 =
...
+   transitivity was also not guaranteed before, since B may contain initialization logic
+   or pre-/post-upgrade hooks that alter the stable data.
+*)
 let match_stab_sig tfs1 tfs2 : unit Diag.result =
   (* Assume that tfs1 and tfs2 are sorted. *)
   let res = Diag.with_message_store (fun s ->
-    (* Should we insist on monotonic preservation of fields, or relax? *)
     let rec go tfs1 tfs2 = match tfs1, tfs2 with
-      | [], _ ->
+      | [], _ ->
```
luc-blaeser, line 47:
```suggestion
      | [], _ ->
```

src/mo_frontend/stability.ml
```diff
@@ -29,37 +29,37 @@ let error_sub s tf1 tf2 =
...
             if not (sub (as_immut tf1.typ) (as_immut tf2.typ)) then
               error_sub s tf1 tf2;
             go tfs1' tfs2'
-       | -1 ->
-          error_discard s tf1;
-          go tfs1' tfs2
+        | -1 ->
+          (* dropped field is allowed with warning, recurse of tfs1' *)
+          warning_discard s tf1;
+          go tfs1' tfs2
```
luc-blaeser, line 62:
```suggestion
       | -1 ->
         (* dropped field is allowed with warning, recurse of tfs1' *)
         warning_discard s tf1;
         go tfs1' tfs2 
```

src/mo_frontend/stability.ml
```diff
@@ -29,39 +29,39 @@ let error_sub s tf1 tf2 =
...
   let res = Diag.with_message_store (fun s ->
-    (* Should we insist on monotonic preservation of fields, or relax? *)
     let rec go tfs1 tfs2 = match tfs1, tfs2 with
       | [], _ ->
         Some () (* no or additional fields ok *)
       | tf1 :: tfs1', [] ->
-        error_discard s tf1;
-        go tfs1' []
+        (* dropping fields is allowed, but with a warning *)
+        warning_discard s tf1;
```
crusso, line 51: I guess this only warns about the first field, but not the remaining ones in tfs1'. That was fine for an error (just bail on the first error) but is less fine for a warning... 
Maybe replace by `List.iter (warning_discard s) tfs1`?

test/cmp/ok/neg-xyz-x.cmp.ok
```diff
@@ -1,7 +1,4 @@
-(unknown location): Compatibility error [M0169], stable variable y of previous type
+(unknown location): warning [M0169], stable variable y of previous type
   Int
-cannot be discarded; promote to type `Any` instead
-(unknown location): Compatibility error [M0169], stable variable z of previous type
-  Bool
-cannot be discarded; promote to type `Any` instead
-FALSE
+ will be discarded. This may cause data loss. Are you sure?
+TRUE
```
crusso, line 4: Note that we've dropped the error for z entirely.... (though I'm curious why it was produced before...)

src/mo_frontend/stability.ml
crusso: Or just recurse down tfs1' [] as in the original code. That should issue all the warnings, just as it issued all the errors.

crusso (changes requested):
Mostly good, but see comments.

src/mo_frontend/stability.ml
luc-blaeser: Thank you for having found this issue. I re-added the recursion.

test/cmp/ok/neg-xyz-x.cmp.ok
luc-blaeser: Thanks for noticing. This is related to the bug above. Sorry that it slipped my attention. (I initially had no warning, just allowed it, and then later introduced the warning, but half-baked.)

luc-blaeser:
> Mostly good, but see comments.

Thank you for the careful review!

crusso (approved):
LGTM

PR #4608, opened 2024-07-10
feat: parenthetical syntax for message sends (`cycles`, `timeout` etc.)

ggreif:
This PR introduces new syntax (parenthetical notes, or simply _parentheticals_) for `async`-valued calls and `async` self-sends. Two attributes are recognised therein: `cycles : Nat` and `timeout : Nat32`. The type-checker enforces these types (when attributes present) and warns on unknown ones. Having a `timeout` attribute present will make the message send a bounded-wait one (also called best-effort) according to the new replica semantics. Specifying `cycles` works like the (soon to be deprecated) `Cycles.add<system>`, ~but one has to decide which to use (either `Cycles.add` or a parenthetical) since there is no mix-and-match~.

-----------
~`ICCallPrim` (and friends) should carry the fragment to set the `SystemCyclesAddPrim` while the call is being set up towards the replica.~ This now happens in the desugarer by assigning to variables (`@cycles` and `@timeout`).

Parentheticals
- now: `(with cycles = 42_000_000) Actor.call(param)`
- also now: `(with timeout = 10)`
  `ic0.call_with_best_effort_response : (timeout_seconds : i32) -> ()`
- maybe: `(with receiveMax = 50_000)` (to limit the response size)
- maybe: `(with resend = { tries = 5; delay = 3 })` when `SYS_UNKNOWN` reject response (like [Unix `EINTR`](https://unix.stackexchange.com/questions/253349/what-is-the-rationale-behind-eintr), but saves cycles by not re-encoding the arguments)
- support https://github.com/dfinity/ic/pull/1158
- usecase `(with memoryLimit = 1G) ActorClass(<args>)`
- default/suppress call-meta-attributes: `func() : async () = (with) async { ... }` — document this!

TODOs:
- [x] `Changelog.md`
- [x] documentation
- [x] test one-ways
- typecheck
  - [x] warn when an attribute is moot
  - [x] `cycles : Nat` for canister sends (_self_ and _raw_ sends too)
  - [x] `timeout : Nat32` for best-effort
  - [x] parenthetical should have no send cap (test still missing)
- [x] `async` blocks
- [x] `ICCallPrim`, see top
- `ICCallRawPrim`, not directly annotatable (desisting for now, but see #4868)
- [ ] `FIXME`s, `TODO`s
  - [x] `ICCallPrim` should have a non-`option` setup
  - [ ] break out unrelated PRs (optimisations?) and beautifications — see #4890
- [ ] warn on queries?
- [x] best-effort: new error code when deadline passed?
  - [x] doc/md/base/Error.md -> `motoko-base`, see dfinity/motoko-base#686
  - [x] visit all `git grep system_fatal` and amend `#system_unknown`
- [x] decide whether `(with cycles) async` or `async (with cycles)` — decided: keep as prefix!
- [x] write a temporary test that `Cycles.add` still works
- [x] test that parenthetical have no send capability
- [x] test `(with cycles) (func () -> async () {})()` direct application
- [x] test `(with cycles) (system Lib.Class)(...)()`
- [x] test error on `async*` (and call)


src/ir_def/construct.ml
```diff
@@ -312,14 +312,16 @@ let funcE name sort ctrl typ_binds args typs exp =
     note = Note.{ def with typ; eff = T.Triv };
   }
 
+let recordE' = ref (fun _ -> nullE ()) (* gets correctly filled below *)
```
ggreif, line 315: use `and` to allow mutual recursion

src/ir_passes/tailcall.ml
```diff
@@ -94,14 +94,14 @@ and assignEs vars exp : dec list =
...
-  | PrimE (CallPrim insts, [e1; e2])  ->
+  | PrimE (CallPrim (insts, pars), [e1; e2])  ->
     begin match e1.it, env with
     | VarE f1, { tail_pos = true;
                  info = Some { func; typ_binds; temps; label; tail_called } }
          when f1 = func && are_generic_insts typ_binds insts  ->
       tail_called := true;
       (blockE (assignEs temps (exp env e2)) (breakE label (unitE ()))).it
-    | _,_-> PrimE (CallPrim insts, [exp env e1; exp env e2])
+    | _,_-> PrimE (CallPrim (insts, pars), [exp env e1; exp env e2])
```
ggreif, line 104:
```suggestion
    | _,_-> PrimE (CallPrim (insts, exp env pars), [exp env e1; exp env e2])
```

src/mo_frontend/definedness.ml
```diff
@@ -82,7 +82,7 @@ let rec exp msgs e : f = match e.it with
   (* Eager uses are either first-class uses of a variable: *)
   | VarE i              -> M.singleton i.it Eager
   (* Or anything that is occurring in a call (as this may call a closure): *)
-  | CallE (e1, ts, e2)  -> eagerify (exps msgs [e1; e2])
+  | CallE (_FIXME, e1, ts, e2)  -> eagerify (exps msgs [e1; e2])
```
ggreif, line 85:
```suggestion
  | CallE (_FIXME, e1, ts, e2) -> eagerify (exps msgs [e1; e2])
```

crusso:
Notes: 

```
// foo.mo
actor class Foo() {}
```

translates too something like:

```
module {
  type Foo = actor { .. }
  func Foo() : async Foo = async {
    let p = await* Internals.create_helper() {
      let available = Cycles.availabe();
      Cycles.add(availabe);
      IC.create_cansiter()
      IC.install_code(...)
   };
   actor (..p..) : Foo
};

```
```
// test
import Lib "foo"
public func test() : async () {
       addCycles(10);
       let a : Lib.Foo = Lib.Foo();
       ();
  }
}
```

src/ir_interpreter/interpret_ir.ml
```diff
@@ -310,7 +310,7 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
   | PrimE (p, es) ->
     interpret_exps env es [] (fun vs ->
       match p, vs with
-      | CallPrim typs, [v1; v2] ->
+      | CallPrim (typs, _), [v1; v2] ->
```
ggreif, line 313: But this ignores the side-effects therein!

src/ir_interpreter/interpret_ir.ml
```diff
@@ -446,7 +446,7 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
         let reject = Option.get env.rejects in
         let e = V.Tup [V.Variant ("canister_reject", V.unit); v1] in
         Scheduler.queue (fun () -> reject e)
-      | ICCallPrim, [v1; v2; kv; rv] ->
+      | ICCallPrim _, [v1; v2; kv; rv] ->
```
ggreif, line 449: Side-effects?

src/ir_passes/async.ml
```diff
@@ -321,23 +315,32 @@ let transform prog =
...
+        else None in
+                                                                             
+
       (blockE (
-        letP (tupP [varP nary_async; varP nary_reply; varP reject]) def ::
+        letP (tupVarsP [nary_async; nary_reply; reject]) def ::
         let_eta exp1' (fun v1 ->
           let_seq ts1 exp2' (fun vs ->
-            [ expD (ic_callE v1 (seqE (List.map varE vs)) (varE nary_reply) (varE reject)) ]
+              [ expD (ic_callE setup v1 (seqE (List.map varE vs)) (varE nary_reply) (varE reject)) ]
```
ggreif, line 333: unindent

src/codegen/compile.ml
```diff
@@ -11792,7 +11791,9 @@ and compile_prim_invocation (env : E.t) ae p es at =
     let (set_arg, get_arg) = new_local env "arg" in
     let (set_k, get_k) = new_local env "k" in
     let (set_r, get_r) = new_local env "r" in
-    let add_cycles = Internals.add_cycles env ae in
+    let add_cycles = match setup with
+      | None -> Internals.add_cycles env ae
+      | Some exp -> compile_exp_vanilla env ae exp ^^ G.i Drop in
```
ggreif, line 11796: Compile to `Tup`, then we don't need to `Drop`, maybe!

src/codegen/compile.ml
```diff
@@ -10627,7 +10627,9 @@ and compile_prim_invocation (env : E.t) ae p es at =
          StackRep.of_arity return_arity,
 
          code1 ^^
-         compile_unboxed_zero ^^ (* A dummy closure *)
+         Type.(match as_obj par.note.Note.typ with
+         | Object, [] -> compile_unboxed_zero (* a dummy closure *)
```
ggreif, line 10631: Perhaps push a dummy, zero-length closure, so that `ICCyclesPrim` doesn't need to do a null-check?

src/ir_def/construct.ml
```diff
@@ -108,6 +111,7 @@ let primE prim es =
     | SystemCyclesAcceptPrim -> T.nat
     | DeserializePrim ts -> T.seq ts
     | DeserializeOptPrim ts -> T.Opt (T.seq ts)
+    | ICCyclesPrim -> T.(Opt (Obj (Object, [{ lab = "cycles"; typ = nat; src = empty_src}])))
```
ggreif, line 114: Have a common place for these types?

src/ir_def/ir_effect.ml
```diff
@@ -8,11 +8,16 @@ module T = Mo_types.Type
    but I prefer to keep it mostly separate for now *)
 
 let max_eff e1 e2 =
-  match e1,e2 with
+  match e1, e2 with
   | T.Triv, T.Triv -> T.Triv
   | _ , T.Await -> T.Await
   | T.Await, _ -> T.Await
 
+let max_effs' seed = List.fold_left max_eff seed
```
ggreif, line 16: this is separated out to #4649.

src/ir_passes/async.ml
```diff
@@ -24,8 +24,8 @@ module ConRenaming = E.Make(struct type t = con let compare = Cons.compare end)
 
 (* Helpers *)
 
-let selfcallE ts e1 e2 e3 e4 =
-  { it = SelfCallE (ts, e1, e2, e3, e4);
+let selfcallE (cyc : exp) ts e1 e2 e3 e4 =
```
ggreif, line 27:
```suggestion
let selfcallE cyc ts e1 e2 e3 e4 =
```

doc/md/examples/grammar.txt
```diff
@@ -214,6 +214,7 @@
     'break' <id> <exp_nullary>?
     'continue' <id>
     'debug' <exp_nest>
+    '(' <exp_post>? 'with' <list(<exp_field>, ';')> ')' <exp_nest>
```
rvanasa, line 217: Thoughts on using `,` here in place of `;` for consistency with other parenthesized expressions? The formatter uses this invariant (commas in parentheses, semicolons in square brackets and curly braces) for automatically replacing commas with semicolons and vice versa whenever there is otherwise a syntax error that makes the AST unparseable. I can add an exception, but it seems nice to keep this pattern so that it's easier for people to remember which delimiter to use.

ggreif: Sorry for missing this comment for some time... Yeah, it is a nice consistency argument. I was stealing the syntax from the record field separators, but I guess comma works as well. Will try and report back. @crusso any gut feelings about this?

crusso: I'd stick with `;` since these are more like fields. Would be even nicer if they just were fields...

ggreif: Yes, they are fields, but where is it written that they cannot be separated by commas? ;-) Just playing the devil's advocate.

Anyway, I have started a branch to get a feeling for the suggestion: #4782. I am not married to either way.

src/codegen/compile_classical.ml
```diff
@@ -2251,6 +2251,15 @@ module Tagged = struct
     set_tag ^^
     go cases
 
+  (* like branch_default_with but the tag is known statically *)
```
ggreif, line 2254: These changes need to be transported to `%_enhanced.ml`.

src/codegen/compile_classical.ml
```diff
@@ -5500,7 +5510,7 @@ module IC = struct
     | Flags.(ICMode | RefMode) ->
       system_call env "call_cycles_add128"
     | _ ->
-      E.trap_with env "cannot accept cycles when running locally"
+      E.trap_with env "cannot add cycles when running locally"
```
ggreif, line 5513: break this out?

src/ir_def/check_ir.ml
```diff
@@ -803,13 +809,15 @@ let rec check_exp env (exp:Ir.exp) : unit =
...
-  | SelfCallE (ts, exp_f, exp_k, exp_r, exp_c) ->
+  | SelfCallE (cyc, ts, exp_f, exp_k, exp_r, exp_c) ->
     check (not env.flavor.Ir.has_async_typ) "SelfCallE in async flavor";
+    check_exp env cyc;
     List.iter (check_typ env) ts;
     check_exp { env with lvl = NotTopLvl } exp_f;
     check_exp env exp_k;
     check_exp env exp_r;
     check_exp env exp_c;
+    typ cyc <: T.(Opt (Obj (Object, [])));
```
ggreif, line 820: Instead of `Opt`, could we have a (non-user-accessible) field `@cycles`?

src/ir_def/construct.ml
```diff
@@ -377,7 +387,7 @@ let dotE exp name typ =
   { it = PrimE (DotPrim name, [exp]);
     at = no_region;
     note = Note.{ def with
-      typ = typ;
+      typ;
```
ggreif, line 390: break out?

src/ir_passes/async.ml
```diff
@@ -156,8 +160,8 @@ let let_seq ts e d_of_vs =
     (letP p e)::d_of_vs [x]
   | ts ->
     let xs = fresh_vars "x" ts in
-    let p = tupP (List.map varP xs) in
-    (letP p e)::d_of_vs (xs)
+    let p = tupVarsP xs in
+    (letP p e) :: d_of_vs xs
```
ggreif, line 164:
```suggestion
    letP p e :: d_of_vs xs
```

src/ir_passes/await.ml
```diff
@@ -99,7 +99,9 @@ let rec t_async context exp =
        (LabelEnv.add Return (Cont k_ret)
           (LabelEnv.singleton Throw (Cont k_fail)))
    in
-     cps_asyncE s typ1 (typ exp1)
+     cps_asyncE s typ1 (match par_opt with
+                        | Some par -> assert false(*FIXME:; optE par*)
```
ggreif, line 103: strange, why doesn't this trigger?

ggreif:
@Mergifyio refresh

crusso (commented):
Some interim comments

src/lang_utils/error_codes.ml
```diff
@@ -203,5 +203,8 @@ let error_codes : (string * string option) list =
     "M0197", Some([%blob "lang_utils/error_codes/M0197.md"]); (* `system` capability required *)
     "M0198", Some([%blob "lang_utils/error_codes/M0198.md"]); (* Unused field pattern warning *)
     "M0199", Some([%blob "lang_utils/error_codes/M0199.md"]); (* Deprecate experimental stable memory *)
-    "M0200", Some([%blob "lang_utils/error_codes/M0200.md"]) (* Cannot determine subtyping or equality *)
+    "M0200", Some([%blob "lang_utils/error_codes/M0200.md"]); (* Cannot determine subtyping or equality *)
+    "M0202", None; (* parenthetical note must be applied to a message send *)
```
ggreif, line 207: make this contiguous

src/ir_passes/async.ml
```diff
@@ -327,26 +332,37 @@ let transform prog =
...
-      let (nary_async, nary_reply, reject, clean), def =
-        new_nary_async_reply ts2
-      in
+      let (nary_async, nary_reply, reject, clean), def = new_nary_async_reply ts2 in
+      let hasCycles = Type.(sub pars.note.Note.typ (Obj(Object, [{ lab = "cycles"; typ = nat; src = empty_src}]))) in
+      let hasTimeout = Type.(sub pars.note.Note.typ (Obj(Object, [{ lab = "timeout"; typ = nat32; src = empty_src}]))) in
+      let cyclesSetup = if hasCycles
+        then Some (thenE
+                     (natE Mo_values.Numerics.Nat.zero |> assignVarE "@cycles")
+                     (primE SystemCyclesAddPrim [dotE pars "cycles" T.nat]))
```
ggreif, line 341: Bind pars to an object variable? Otherwise we get duplicated effects!

src/codegen/compile_classical.ml
```diff
@@ -12187,6 +12215,29 @@ and compile_prim_invocation (env : E.t) ae p es at =
     SR.Vanilla, Cycles.refunded env
   | SystemCyclesBurnPrim, [e1] ->
     SR.Vanilla, compile_exp_vanilla env ae e1 ^^ Cycles.burn env
+  | ICCyclesPrim, [] ->
```
ggreif, line 12218: This is not an idempotent operation, so we have to be careful to not call it twice. E.g. it fails for paired up environment+parenthetical.

src/codegen/compile_classical.ml
```diff
@@ -12187,6 +12215,29 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
+          [ Tagged.Closure,
+            G.i Drop ^^
+            Opt.null_lit env
+          ; Tagged.(Array T),
+            Opt.inject_simple env (Arr.load_field env 0l) ^^
+            G.i (LocalGet (nr 0l)) ^^
+            Arr.load_field env 1l ^^
+            G.i (LocalSet (nr 0l))
+          ; Tagged.Object,
+            Opt.inject_simple env G.nop
```
ggreif, line 12234: don't care storing back anything as there is no captured environment.

src/codegen/compile_classical.ml
```diff
@@ -12187,6 +12215,29 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
+  | ICCyclesPrim, [] ->
+    SR.Vanilla,
+    G.i (LocalGet (nr 0l)) ^^ (* closed-over bindings *)
+    G.if1 I32Type
+      begin
+        G.i (LocalGet (nr 0l)) ^^
+        Tagged.branch_with env [I32Type]
+          [ Tagged.Closure,
+            G.i Drop ^^
+            Opt.null_lit env
```
ggreif, line 12227: use `br 0` to fall out?

src/ir_def/construct.ml
ggreif: Also, this should be a pair of `Opt`s with `cycles` and `timeout` views!

ggreif: ... with a better name `SendMetaPrim`

src/codegen/compile_classical.ml
ggreif: This should return two options. Possibly just the `Nat` (`cycles`) and `Nat32` (`timeout`). We definitely have to search the attributes.

src/prelude/internals.mo
```diff
@@ -507,6 +528,11 @@ func @create_actor_helper(wasm_module : Blob, arg : Blob) : async Principal = as
 
 // raw calls
 func @call_raw(p : Principal, m : Text, a : Blob) : async Blob {
+  let cycles = (prim "cyclesAvailable" : () -> Nat)();
+  if (cycles != 0) {
+    ignore (prim "cyclesAccept" : Nat -> Nat) cycles;
```
ggreif, line 533: check whether we do something different in class instantiation with sent cycles

src/prelude/internals.mo
```diff
@@ -9,6 +9,16 @@ code, and cannot be shadowed.
 
 type @Iter<T_> = {next : () -> ?T_};
 
+// Function called by backend to add funds to call.
+// DO NOT RENAME without modifying compilation.
+func @pass_cycles(par : ?{ cycles : Nat }) {
+  let ?{ cycles } = par else return;
```
ggreif, line 15: `@reset_cycles()` should be the first thing

src/ir_def/construct.ml
ggreif:
```suggestion
    | ICCyclesPrim -> T.(Opt (Obj (Object, [])))
```
The dynamic type is unknown.

src/ir_def/check_ir.ml
```diff
@@ -556,6 +556,11 @@ let rec check_exp env (exp:Ir.exp) : unit =
       check (T.shared (T.seq ots)) "DeserializeOpt is not defined for operand type";
       typ exp1 <: T.blob;
       T.Opt (T.seq ots) <: t
+
+
+    | ICCyclesPrim, [] -> () (* FIXME *)
```
ggreif, line 561: check for `Opt(Obj(Object, []))`

src/ir_def/ir.ml
```diff
@@ -161,19 +161,21 @@ and prim =
   | SystemCyclesBalancePrim
   | SystemCyclesRefundedPrim
   | SystemCyclesBurnPrim
+  | ICCyclesPrim                      (* cycles to send by parenthetical *)
```
ggreif, line 164: rename to `ICCallAttrsPrim`

src/prelude/internals.mo
ggreif: Done: ff5c16e8ae6b07bdbaba5d92446756174a5b3040

src/prelude/internals.mo
ggreif: Now they are the same!

src/ir_def/ir.ml
ggreif: Done: 92e194c83.

src/ir_def/check_ir.ml
ggreif: Done: cb17a6e5f0bb17221cf85f8f19d5d32011320272.

src/ir_def/check_ir.ml
```diff
@@ -405,7 +405,9 @@ let rec check_exp env (exp:Ir.exp) : unit =
   | PrimE (p, es) ->
     List.iter (check_exp env) es;
     begin match p, es with
-    | CallPrim insts, [exp1; exp2] ->
+    | CallPrim (insts, par), [exp1; exp2] ->
+      check_exp { env with async = None } par;
```
ggreif, line 409: Is this the correct way to ensure that there is no send capability in this part?

src/ir_def/check_ir.ml
```diff
@@ -724,10 +734,6 @@ let rec check_exp env (exp:Ir.exp) : unit =
   | SwitchE (exp1, cases) ->
     check_exp env exp1;
     let t1 = T.promote (typ exp1) in
-(*    if not env.pre then
```
ggreif, line 727: I believe we don't do coverage checking in the IR, right? This must be a leftover...

src/ir_def/check_ir.ml
```diff
@@ -805,13 +813,15 @@ let rec check_exp env (exp:Ir.exp) : unit =
       , tbs, List.map (T.close cs) ts1, List.map (T.close cs) ret_tys
       ) in
     fun_ty <: t
-  | SelfCallE (ts, exp_f, exp_k, exp_r, exp_c) ->
+  | SelfCallE (cyc, ts, exp_f, exp_k, exp_r, exp_c) ->
```
ggreif, line 816:
```suggestion
  | SelfCallE (par, ts, exp_f, exp_k, exp_r, exp_c) ->
```

src/lang_utils/error_codes.ml
```diff
@@ -213,4 +213,8 @@ let error_codes : (string * string option) list =
     "M0207", None; (* Migration consumes, but does not produce, an un-declared field *)
     "M0208", None; (* Missing field migration*)
     "M0209", None; (* Misplaced migration expression on module/object *)
+    "M0210", None; (* Parenthetical note must be applied to a message send *)
+    "M0211", None; (* Parenthetical note has no attributes *)
+    "M0212", Some([%blob "lang_utils/error_codes/M0212.md"]); (* Unrecognised attribute in parenthetical note *)
+    "M0213", None; (* Parenthetical note on public methods is disallowed *)
```
ggreif, line 219:
```suggestion
    "M0213", None; (* Parenthetical note on shared methods is disallowed *)
```

src/ir_def/construct.ml
ggreif: Done by reordering: 8bebec586

src/codegen/compile_classical.ml
ggreif: yes: #4890

src/codegen/compile_classical.ml
```diff
@@ -2541,7 +2551,7 @@ module Closure = struct
       I32Type :: Lib.List.make n_args I32Type,
       FakeMultiVal.ty (Lib.List.make n_res I32Type))) in
     (* get the table index *)
-    Tagged.load_forwarding_pointer env ^^
+    (*Tagged.load_forwarding_pointer env ^^ FIXME: NOT needed, accessing immut slots*)
```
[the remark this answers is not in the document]
ggreif: See: #4890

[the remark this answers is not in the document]
ggreif: #4890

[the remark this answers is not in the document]
ggreif: #4890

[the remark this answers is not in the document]
ggreif: #4890

src/codegen/compile_classical.ml
```diff
@@ -10928,18 +10967,23 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
 
          StackRep.of_arity return_arity,
          code1 ^^ StackRep.adjust env fun_sr SR.Vanilla ^^
+         Closure.prepare_closure_call env ^^ (* FIXME: move to front elsewhere too *)
          set_clos ^^
-         get_clos ^^
-         Closure.prepare_closure_call env ^^
+         Type.(match as_obj par.note.Note.typ, ret_tys with
+               | (Object, []), _ -> get_clos (* just the closure *)
+               | _, [ret] when is_async_fut ret -> Arr.lit env Tagged.T [compile_exp_vanilla env ae par; get_clos] (* parenthetical: pass a pair *)
```
ggreif, line 10985: `is_async_fut` is the wrong predicate! When `T -> async U` `Local` types get lowered, we end up with `(,,) -> { #schedule; #suspend }`

src/mo_types/type.ml
```diff
@@ -586,6 +588,8 @@ let is_unit = function Tup [] -> true | _ -> false
 let is_pair = function Tup [_; _] -> true | _ -> false
 let is_func = function Func _ -> true | _ -> false
 let is_async = function Async _ -> true | _ -> false
+let is_async_cmp = function Async (Cmp, _, _) -> true | _ -> false
```
ggreif, line 591: not needed any more?

src/mo_frontend/typing.ml
```diff
@@ -2544,6 +2569,42 @@ and infer_obj env s exp_opt dec_fields at : T.typ =
...
+       let check want =
+         if not (sub env par.at typ want)
+         then local_error env par.at "M0127"(*FIXME*) "parenthetical field %s is declared with type%a\ninstead of expected type%a" lab
+                display_typ typ
+                display_typ want in
+       match lab with
+       | "cycles" -> check T.nat
+       | "timeout" -> check T.nat32
+       | _ -> () in
+     List.iter check_lab attrs_flds;
```
ggreif, line 2604: This should probably go into `infer_check_bases_fields`, as it currently won't check bases...

ggreif: also, use `checked` above

default.nix
```diff
@@ -512,8 +512,8 @@ rec {
       drun-compacting-gc = snty_compacting_gc_subdir "run-drun" [ moc nixpkgs.drun ] ;
       drun-generational-gc = snty_generational_gc_subdir "run-drun" [ moc nixpkgs.drun ] ;
       drun-incremental-gc = snty_incremental_gc_subdir "run-drun" [ moc nixpkgs.drun ] ;
-      drun-eop-release = enhanced_orthogonal_persistence_subdir "run-drun" [ moc nixpkgs.drun ] ;
-      drun-eop-debug = snty_enhanced_orthogonal_persistence_subdir "run-drun" [ moc nixpkgs.drun ] ;
+      ## FOR NOW drun-eop-release = enhanced_orthogonal_persistence_subdir "run-drun" [ moc nixpkgs.drun ] ;
+      ## FOR NOW drun-eop-debug = snty_enhanced_orthogonal_persistence_subdir "run-drun" [ moc nixpkgs.drun ] ;
```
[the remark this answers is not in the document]
ggreif: I'll patch in the classical changes into enhanced, and then this can be re-enabled.

src/lowering/desugar.ml
```diff
@@ -258,6 +263,23 @@ and exp' at note = function
       { it = I.LetD ({it = I.WildP; at = e.at; note = T.Any}, exp e);
         at = e.at; note = ()}], (unitE ()))
 
+and distill_meta = function
+  | None -> []
+  | Some par ->
+    let cycles, clean_cycles =
+      if T.(sub par.note.note_typ (Obj (Object, [{ lab = "cycles"; typ = nat; src = empty_src }])))
```
ggreif, line 270: field def and label name should be in `type.md`

src/mo_types/type.ml
ggreif: gone: f18adbf41

src/lowering/desugar.ml
ggreif: Done: 10258f0b82641119523be3eb3a45a719a185a940

src/mo_frontend/typing.ml
ggreif: Done: 0c07b52f8

src/lang_utils/error_codes.ml
```diff
@@ -213,4 +213,9 @@ let error_codes : (string * string option) list =
     "M0207", None; (* Migration consumes, but does not produce, an un-declared field *)
     "M0208", None; (* Missing field migration*)
     "M0209", None; (* Misplaced migration expression on module/object *)
+    "M0210", None; (* Parenthetical note must be applied to a message send *)
+    "M0211", None; (* Parenthetical note has no attributes *)
+    "M0212", Some([%blob "lang_utils/error_codes/M0212.md"]); (* Unrecognised attribute in parenthetical note *)
+    "M0213", None; (* Parenthetical note on shared methods is disallowed *)
+    "M0214", None; (* Expected type of filed in arenthetical note differs from inferred *)
```
crusso, line 220:
```suggestion
    "M0214", None; (* Expected type of filed in parenthetical note differs from inferred *)
```

src/lang_utils/error_codes/M0212.md
```diff
@@ -0,0 +1,14 @@
+# M0212
+
+This warning means that you are affixing a parenthetical note to a message send (i.e. either
+a canister method call of a self-send with `async`, or `actor`) that contains an attribute not recognised
+by this version of the Motoko compiler.
+
+Currently following attributes are recognised in parenthetical notes:
```
crusso, line 7:
```suggestion
Currently, the following attributes are permitted in parenthetical notes:
```

src/mo_def/arrange.ml
```diff
@@ -98,7 +98,8 @@ module Make (Cfg : Config) = struct
         Atom (if sugar then "" else "=");
         exp e'
       ]
-    | CallE (e1, ts, e2)  -> "CallE"   $$ [exp e1] @ inst ts @ [exp e2]
+    | CallE (None, e1, ts, e2)  -> "CallE"      $$ [exp e1] @ inst ts @ [exp e2]
+    | CallE (Some par, e1, ts, e2) -> "CallE()" $$ [exp par] @ [exp e1] @ inst ts @ [exp e2]
```
crusso, line 102: It turns out that vscode actually uses arrange.ml to communicate type info with the plugin (not pretty, but true). This means that changing the output format might affect vscode.

To avoid, this it might be better to follow and adapt the current logic for migration expressions (see `include_migration` and `migration eo`. Maybe it's enough to rename these to `include_parenthetical` `parethetical eo` and use in CallE and AsyncE cases

src/mo_frontend/typing.ml
```diff
@@ -1026,8 +1026,8 @@ let rec is_explicit_exp e =
     true
   | LitE l -> is_explicit_lit !l
   | UnE (_, _, e1) | OptE e1 | DoOptE e1
-  | ProjE (e1, _) | DotE (e1, _) | BangE e1 | IdxE (e1, _) | CallE (e1, _, _)
-  | LabelE (_, _, e1) | AsyncE (_, _, e1) | AwaitE (_, e1) ->
+  | ProjE (e1, _) | DotE (e1, _) | BangE e1 | IdxE (e1, _) | CallE (_(*FIXME: correct?*), e1, _, _)
```
crusso, line 1029: I think you can remove this. IIRC is_explicit_exp returns an indication whether the inferred type of the expression is good enough to resolve operator  overloading. But not sure. You'd have to use git blame to discover the meaning of this from Andrea's original PR that introduced is_explicit.

src/mo_types/type.mli
```diff
@@ -133,6 +134,7 @@ val is_unit : typ -> bool
 val is_pair : typ -> bool
 val is_func : typ -> bool
 val is_async : typ -> bool
+val is_low_async_fut : typ -> bool
```
crusso, line 137: Is this code dead now? I can only find the definition not any use. Deletre?

src/prelude/internals.mo
```diff
@@ -309,6 +325,16 @@ func @getSystemRefund() : @Refund {
...
+/*<<<<<<< gabor/parentheticals
+func @coerce_and_cont(a : @Async<()>) :
+  (k : () -> (), r : @Cont<Error>) -> {
+                          #suspend;
+                          #schedule : () -> ()
+                      } =
+  func(k, r) = a(func() = k(), r);
+
+func @new_async<T <: Any>() : (@Async<T>, @Cont<T>, @Cont<Error>) {
+=======*/
```
crusso, line 337: Delete

src/mo_types/type.mli
ggreif: I believe I removed this 😬

ggreif: Yeah, dead, will zap!

ggreif: 8b98dc928

src/prelude/internals.mo
ggreif: Sure!

ggreif: 4cd902f4df657726b0b085990eec3aabf2801842

crusso (commented):
Added some interim comments. Still not done

src/lowering/desugar.ml
```diff
@@ -258,6 +263,23 @@ and exp' at note = function
       { it = I.LetD ({it = I.WildP; at = e.at; note = T.Any}, exp e);
         at = e.at; note = ()}], (unitE ()))
 
+and distill_meta = function
+  | None -> []
+  | Some par ->
+    let cycles, clean_cycles =
```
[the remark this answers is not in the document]
crusso: I guess another option would be to use the current value of the cycles register when the cycles field is absent. Then I guess code using old ExperimentalCycles.add and new timeouts would continue to work. 
One could also assert that the cycles register is zero when the cycles field is present.

src/codegen/compile_classical.ml
```diff
@@ -2519,10 +2519,13 @@ module Closure = struct
     Tagged.load_field env (Int32.add (header_size env) i)
 
   let store_data env i =
-    let (set_closure_data, get_closure_data) = new_local env "closure_data" in
-    set_closure_data ^^
-    Tagged.load_forwarding_pointer env ^^
-    get_closure_data ^^
+    (if G.to_instr_list (Tagged.load_forwarding_pointer env) <> [] then
```
crusso, line 2522:
```suggestion
    (if !Flags.gc_strategy = Flags.Incremental 
```
Is a bit more direct (the same predicate used by load_forwarding_pointer)

src/codegen/compile_enhanced.ml
```diff
@@ -2311,10 +2311,13 @@ module Closure = struct
...
-    set_closure_data ^^
-    Tagged.load_forwarding_pointer env ^^
-    get_closure_data ^^
+    (if G.to_instr_list (Tagged.load_forwarding_pointer env) <> [] then
+       let set_closure_data, get_closure_data = new_local env "closure_data" in
+       set_closure_data ^^
+       Tagged.load_forwarding_pointer env ^^
+       get_closure_data
+     else
+       G.nop) ^^
```
crusso, line 2320:
```suggestion
     let (set_closure_data, get_closure_data) = new_local env "closure_data" in
    set_closure_data ^^
    Tagged.load_forwarding_pointer env ^^
    get_closure_data ^^
```
Revert - compile_enhance is incremental_gc only so the forwarding is always required.

src/codegen/compile_enhanced.ml
```diff
@@ -12320,6 +12326,13 @@ and compile_prim_invocation (env : E.t) ae p es at =
   | SystemCyclesBurnPrim, [e1] ->
     SR.Vanilla, compile_exp_vanilla env ae e1 ^^ Cycles.burn env
 
+  | SystemTimeoutSetPrim, [e1] ->
+    SR.unit,
+    compile_exp_as env ae (SR.UnboxedWord64 Type.Nat32) e1 ^^
+    TaggedSmallWord.lsb_adjust Type.Nat32 ^^
```
crusso, line 12332:
```suggestion
    TaggedSmallWord.lsb_adjust Type.Nat32 ^^
```
This looks dubious to me.  Will need to review carefully.

src/mo_interpreter/interpret.ml
```diff
@@ -572,21 +572,22 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
...
-        let call_conv, f = V.as_func v1 in
-        check_call_conv exp1 call_conv;
-        check_call_conv_arg env exp v2 call_conv;
-        last_region := exp.at; (* in case the following throws *)
-        let c = context env in
-        f c v2 k
-      )
-    )
+  | CallE (par, exp1, typs, exp2) ->
+    interpret_exp env (Lib.Option.get par {exp with it = ObjE ([], [])})
```
crusso, line 576:
```suggestion
    (match par with 
     | Some exp0 -> interpret_exp env exp0 
     | None -> fun k -> k (V.obj Map.empty))     ...
```
(avoids creating and evaluating syntax).

src/prelude/internals.mo
```diff
@@ -9,7 +9,18 @@ code, and cannot be shadowed.
...
+// Function called by backend to add funds to call.
+// DO NOT RENAME without modifying compilation.
+func @pass_cycles(par : ?{ cycles : Nat }) {
+  @reset_cycles();
+  let ?{ cycles } = par else return;
+  if (cycles != 0) {
+    (prim "cyclesAdd" : Nat -> ()) cycles;
+  }
+};
+
```
crusso, line 21: Is this unused now? If so delete.

test/run-drun/par.mo
```diff
@@ -0,0 +1,152 @@
...
+    };
+
+    func localOuter() : async () { debugPrint ("localOuter: " # debug_show(Cycles.available())) };
+    public func test9() : async () {
+        debugPrint "test9()";
+        var env = 42;
+        func local() : async () { debugPrint ("local: " # debug_show(env + Cycles.available())) };
+
+        // give a (dynamically) bogus base
+        object base { /*public*/ func cycles() : Text = "bogus" };
```
crusso, line 123: Would `cycles` even be present in `base` if not declared public? I guess it depends on how we compile objects, but I suspect not.

src/mo_frontend/typing.ml
```diff
@@ -1528,6 +1459,12 @@ and infer_exp'' env exp : T.typ =
       | Some typ -> typ
       | None -> {it = TupT []; at = no_region; note = T.Pre}
     in
+    begin match exp1.it with
+    | AsyncE (Some par, _, _, _) when not env.pre && T.is_shared_sort shared_pat.it ->
+      local_error env par.at "M0213"
+        "parenthetical notes aren't allowed on public methods"
```
crusso, line 1465:
```suggestion
        "parenthetical notes aren't allowed on shared functions"
```

Hmm, are they allowed on local functions?
```
func f() : async Nat = (with cycles =...) async { .. }
```
I guess that might actually work but just flagging in case you meant to rule it out.

src/mo_frontend/typing.ml
```diff
@@ -2544,6 +2569,39 @@ and infer_obj env s exp_opt dec_fields at : T.typ =
...
+       | _ when T.is_shared_sort s -> ()
+       | [cod] when T.is_async cod -> ()
+       | _ -> warn env par.at "M0210" "unexpected parenthetical note on a non-send call"
+       end
+     | _ -> ()
+     end;
+     let checked = T.[ cycles_fld; timeout_fld ] in
+     let [@warning "-8"] par_infer env { it = ObjE (bases, fields); _ } =
+       infer_check_bases_fields env checked par.at bases fields in
+     let attrs = infer_exp_wrapper par_infer T.as_immut env par in
```
crusso, line 2589: The wrapper stuff is a bit ugly, but I guess you need to pass down checked to the infer_check_bases_fields. Hmm, ok.

I do wonder if it might just be simpler for the parser to insert a default base with zero cycles and timeout, and just check that the annotation is a subtype of {cycles : Nat; timeout:Nat32} and then warn about any additional field. But maybe that's not good enough to resolve the overloading of a timeout literal...(which would default to Nat). I can live with this...

crusso (changes requested):
Good stuff. Nearly there...

src/codegen/compile_enhanced.ml
ggreif: Look at `OtherPrim "setCandidLimits"` for a similar example. If the significant bits were in the lower portion, this would always send `0` to the RTS call. @luc-blaeser can you have a look too?

src/ir_def/construct.mli
```diff
@@ -91,6 +91,7 @@ val breakE: id -> exp -> exp
 val retE: exp -> exp
 val immuteE: exp -> exp
 val assignE : var -> exp -> exp
+val assignVarE : id -> exp -> exp
```
ggreif, line 94: maybe this is not justified? 3 uses

src/mo_frontend/typing.ml
ggreif: `(with timeout = 0)` has a very different meaning than not specifying a `timeout`. The former is a best-effort call, the latter and unbounded-wait one.

src/mo_frontend/typing.ml
ggreif: This is actually intended. A minor feature if you will. It shields the send from a caller-side parenthetical, and can impose modalities locally instead. Normally people will use the `func() : async () { ... }` syntax which just accepts the modalities of the sender side.

src/prelude/internals.mo
ggreif: Done: a39fd8120

test/run-drun/par.mo
ggreif: It would be present dynamically in `base`, but it won't enter into the parenthetical record, as it does only copy fields that are around in the type.

ggreif: With the new approach (imperative assignments to backend vars) we always follow the type, and never scan fields dynamically. So it doesn't really matter.

src/codegen/compile_enhanced.ml
ggreif: Oh, yeah, missed that!

ggreif: Done in 8821ac888e6e87445e9fcbf5ea6c65dd4bbeaaa8

src/lowering/desugar.ml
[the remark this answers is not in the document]
ggreif: I have made the naming of vars to reflect the meaning, and added comments. Please re-read.

src/ir_interpreter/interpret_ir.ml
ggreif: Not an issue any more.

src/ir_interpreter/interpret_ir.ml
ggreif: Moot.

src/mo_frontend/typing.ml
crusso: Ok, but you could use `timeout = ?n` instead, which would be nice if you want to explicitly set no timeout.

However, a quick experiment in the playground reveals that we don't currently implement checking mode for `object combinations`, so one would need to extend checking mode to pass down any expected field types to the field extensions for this to work with timout : (?)Nat32. (since inference defaults to Nat). I guess you would be stuck doing inference on the bases if there is more than one though... For  single or no base, I think you could just remove the overridden fields from the expected type.

src/codegen/compile_enhanced.ml
crusso: Ok, you are right it is fine.
I was confused because StackRep.adjust calls TaggedSmallWord.untag, but that only clears the tag bits, it doesn't shift down. Unlike BitTagged.untag, which shifts....

src/mo_frontend/typing.ml
ggreif: Yeah one could do that, at the expense of complicating things. The current way is "timeout is mentioned in the parenthetical" then best effort, otherwise unbounded wait. The feature is marked as experimental, so we can change it if users really want a change.

ggreif: Also users could have have `(waiting with cycles = 333)` where `waiting` is just an empty record.

ggreif:
> I can live with what we have, since it works, just wanted to write down what was bugging me.

Yeah, it feels somewhat ad-hoc. But we still don't know where this is ultimately heading to, and whether parentheticals can be encompassed by a record type as we move forward. We can still have `timeout = -1` for no timeout if we _really_ want a common type, so I am concerned too much. And breaking the current schema is also an option if we figure all out.

doc/md/reference/language-manual.md
```diff
@@ -2206,6 +2208,10 @@ Otherwise, `exp2` is evaluated to a result `r2`. If `r2` is `trap`, the expressi
 
 Otherwise, `r1` is a function value, `<shared-pat>? func <X0 <: V0, …​, n <: Vn> <pat1> { <exp> }` (for some implicit environment), and `r2` is a value `v2`. If `<shared-pat>` is present and of the form `shared <query>? <pat>` then evaluation continues by matching the record value `{caller = p}` against `<pat>`, where `p` is the [`Principal`](../base/Principal.md) invoking the function, typically a user or canister. Matching continues by matching `v1` against `<pat1>`. If pattern matching succeeds with some bindings, then evaluation returns the result of `<exp>` in the environment of the function value not shown extended with those bindings. Otherwise, some pattern match has failed and the call results in `trap`.
 
+A `<parenthetical>`, when present, changes the modality of the message send (provided that the return type `T` is of form `async T₀`, i.e. a future). Attributes recognised are
```
crusso, line 2211:
```suggestion
A `<parenthetical>`, when present, modifies the semantics of the message send (provided that the return type `T` is of form `async U`, i.e. a future). Attributes recognised are
```

doc/md/writing-motoko/async-data.md
```diff
@@ -69,6 +70,26 @@ Unlike a local function call, which blocks the caller until the callee has retur
 
 Awaiting a future a second time will just produce the same result, including re-throwing any error stored in the future. Suspension occurs even if the future is already complete; this ensures state changes and message sends prior to every `await` are committed.
 
+## Using parentheticals to modify message send modalities
+
+In the above examples all messages sent to the `Counter` actor are of unbounded-wait kind and attach no cycles. Both of these
+aspects can be changed by syntactically adding a parenthetical and thus modifying the modality of the send.
```
crusso, line 76:
```suggestion
aspects can be configured by syntactically adding a parenthetical and thus modifying the dynamic attributes of the message send.
```

doc/md/writing-motoko/async-data.md
```diff
@@ -69,6 +70,26 @@ Unlike a local function call, which blocks the caller until the callee has retur
...
+
+To add cycles to the send one would write
+``` motoko no-repl
+let a = (with cycles = 42_000_000) Counter.bump();
+```
+Similarly, one can specify a cutoff for potentially long-running messaging by providing a timeout, and thus delivering a bounded-wait
+message to the destination actor
```
crusso, line 83: Similarly, one can specify an explicit timeout to apply when awaiting the result of the message. This is useful in applications that prefer best-effort  over guaranteed response message delivery:

doc/md/reference/language-manual.md
```diff
@@ -2490,6 +2496,10 @@ The implicit return type in `<block-or-exp>` is `T`. That is, the return express
 
 Evaluation of `async <block-or-exp>` queues a message to evaluate `<block-or-exp>` in the nearest enclosing or top-level actor. It immediately returns a future of type `async T` that can be used to `await` the result of the pending evaluation of `<exp>`.
 
+The presence of `<parenthetical>` changes the modality of the self-send to
```
crusso, line 2499:
```suggestion
The presence of `<parenthetical>` modifies the semantics of the async expression to
```

doc/md/reference/language-manual.md
```diff
@@ -2490,6 +2496,10 @@ The implicit return type in `<block-or-exp>` is `T`. That is, the return express
...
+- attach cycles with attribute `cycles : Nat`
+- change to bounded-wait by specifying an attribute `timeout : Nat32`.
```
crusso, line 2501:
```suggestion
- impose a timeout applied when await the result by specifying the attribute `timeout : Nat32`.
```

src/mo_frontend/typing.ml
```diff
@@ -2544,6 +2569,39 @@ and infer_obj env s exp_opt dec_fields at : T.typ =
...
+  | None -> ()
+  | Some par ->
+     let env = { env with async = C.NullCap } in
+     begin match typ_opt with
+     | Some fun_ty when T.is_func fun_ty ->
+       let s, _, _, _, ts2 = T.as_func fun_ty in
+       begin match ts2 with
+       | _ when T.is_shared_sort s -> ()
+       | [cod] when T.is_async cod -> ()
+       | _ -> warn env par.at "M0210" "unexpected parenthetical note on a non-send call"
```
crusso, line 2582:
```suggestion
       | _ -> warn env par.at "M0210" "misplaced parenthetical (this call does not send a message)"
```

src/mo_frontend/typing.ml
```diff
@@ -2544,6 +2569,39 @@ and infer_obj env s exp_opt dec_fields at : T.typ =
...
+         if not (sub env par.at typ want)
+         then local_error env par.at "M0214" "field %s in parenthetical is declared with type%a\ninstead of expected type%a" lab
+                display_typ typ
+                display_typ want in
+       match List.find_opt (fun { T.lab = l; _} -> l = lab) checked with
+       | Some { T.typ; _} -> check typ
+       | None -> () in
+     List.iter check_lab attrs_flds;
+     let unrecognised = List.(filter T.(fun {lab; _} -> lab <> cycles_lab && lab <> timeout_lab) attrs_flds |> map (fun {T.lab; _} -> lab)) in
+     if unrecognised <> [] then warn env par.at "M0212" "unrecognised attribute %s in parenthetical note" (List.hd unrecognised);
```
crusso, line 2603:
```suggestion
     if unrecognised <> [] then warn env par.at "M0212" "unrecognised attribute %s in parenthetical " (List.hd unrecognised);
```

test/fail/par-bad-base.mo
```diff
@@ -0,0 +1,19 @@
...
+    await (with cycles = "") async ();
+    func cycles() {};
+    await (with cycles) async ();
+    await (with timeout = "") async ();
+  };
+  func b() : async () {
+    object base {
+      public func cycles() {}
+    };
+    await (base with) async ();
```
crusso, line 17: Hmm, this is kinda ugly.
I wonder if
(with object-combination) would work in the parser...
(with base)
(with cycles = 100l, ...)
(with base with cycles = 100).
Probably not because of punning...

doc/md/canister-maintenance/cycles.md
```diff
@@ -13,11 +13,13 @@ In Motoko programs deployed on ICP, each actor represents a canister and has an
 
 Callees can accept all, some, or none of the available cycles up to limit determined by their actor’s current balance. Any remaining cycles are refunded to the caller. If a call traps, all its accompanying cycles are automatically refunded to the caller without loss.
 
-In future, we may see Motoko adopt dedicated syntax and types to support safer programming with cycles. For now, we provide a temporary way to manage cycles through a low-level imperative API provided by the [ExperimentalCycles](../base/ExperimentalCycles.md) library in package `base`.
+In near future, we shall transition Motoko to dedicated syntax and types to support safer programming with cycles. As a first step we employ parenthetical notes on message sends where the user attaches `(where cycles = <amount>)` as a prefix to message sends. This supplants the use of `ExperimentalCycles.add<system>` in below examples.
```
crusso, line 16:
```suggestion
Motoko is adopting dedicated syntax and types to support safer programming with cycles.  Users can now attach `(where cycles = <amount>)` as a prefix to a message sends and async expressions. 
This new syntax will eventually obsolete the use of `ExperimentalCycles.add<system>(cycles)` in the examples that follow.
```

crusso: Why not also give the piggy bank example with the new syntax?

doc/md/writing-motoko/async-data.md
crusso, line 83:
```suggestion
Similarly, one can specify a cutoff for potentially long-running messaging by providing a timeout, and thus delivering a bounded-wait
message to the destination actor
```

crusso, line 83:
```suggestion
Similarly, one can specify an explicit timeout to apply when awaiting the result of the message. This is useful in applications that prefer best-effort  over guaranteed response message delivery:
```

ggreif, line 83:
```suggestion
Similarly, one can specify an explicit timeout to apply when awaiting the result of the message. This is useful in applications that prefer best-effort over guaranteed response message delivery:
```

src/ir_def/check_ir.ml
crusso: Do you actually need to?

doc/md/writing-motoko/async-data.md
ggreif: I have committed this already.

src/mo_def/arrange.ml
crusso: The idea was to drop the parenthetical completely when outputting syntax for vscode.

src/lowering/desugar.ml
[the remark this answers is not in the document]
crusso: It wasn't so much the naming but the structure of the code, but you can leave it.

src/mo_frontend/typing.ml
ggreif: 4f4e93e03

src/lowering/desugar.ml
```diff
@@ -258,6 +263,28 @@ and exp' at note = function
...
         at = e.at; note = ()}], (unitE ()))
 
+and parenthetical = function
+  | None -> []
+  | Some par ->
+    (* fishing for relevant attributes in the parenthetical based on its static type *)
+    let cycles, clean_cycles =
+      if T.(sub par.note.note_typ (Obj (Object, [T.cycles_fld])))
+      then [fun parV -> dotE parV T.cycles_lab T.nat |> assignVarE "@cycles" |> expD], []
+      else [], [natE Mo_values.Numerics.Nat.zero |> assignVarE "@cycles" |> expD] in
```
crusso, line 273:
```suggestion
      else [], [] in
```
If you did this, wouldn't the old mechanism still work if you only apply at timeout but don't mention cycles?
Then the Changelog doesn't have to say the approaches can't be mixed.

src/mo_def/arrange.ml
crusso: Of course the alternative is to check the plug in still works and adapt if necessary....

Changelog.md
```diff
@@ -1,9 +1,26 @@
...
 
+## 0.14.2 (FUTURE)
+
 * motoko (`moc`)
 
-  * bugfix: `mo-doc` will now generate documentation for `actor`s and `actor class`es (#4905).
+  * Added support for sending `cycles` and setting a `timeout` in parentheticals.
+    This is an **experimental feature**, with new syntax, and now also allowing best-effort
+    message sends. The legacy call `Cycles.add<system>` is still supported, but the
+    two approaches cannot be mixed (#4608).
```
crusso, line 10: See comment in desugar.ml and how to make them coexist.

src/lowering/desugar.ml
ggreif: As you say, a `Cycles.add` would continue to work. Ah, you are worried about codebase migration, where folks want to add `(with timeout) but are hesitant to check for the presence of `Cycles.add` in front of the call?

Changelog.md
ggreif: Why would coexistence be important? If the user attaches a parenthetical, why not go all the way?

crusso: I suppose its weird if someone specifies just a timout but that clears any cycles they added using Cycles.add. Probably rare but we don't give them any compiler warnings about mixing both, so maybe it would just be better to make it work.

src/lowering/desugar.ml
crusso: yep

ggreif: Hmm, I am not married to the idea, but I think we should deprecate the old way sooner than later. I'll make this change then.

src/mo_def/arrange.ml
ggreif: I went the `include_parenthetical` route!

Changelog.md
ggreif: Hmm, Okay, 0189496f513ed7786df3eb75c0bc4cf54cd961f5 changed the default. Test fa768703d3f0b48d8f215e8d25aa01d6dc8f2f84 shows that `Cycles.add` now indeed punches the parenthetical.

src/lowering/desugar.ml
```diff
@@ -258,6 +263,28 @@ and exp' at note = function
...
         at = e.at; note = ()}], (unitE ()))
 
+and parenthetical = function
+  | None -> []
+  | Some par ->
+    (* fishing for relevant attributes in the parenthetical based on its static type *)
+    let cycles, clean_cycles =
+      if T.(sub par.note.note_typ (Obj (Object, [T.cycles_fld])))
+      then [fun parV -> dotE parV T.cycles_lab T.nat |> assignVarE "@cycles" |> expD], []
+      else [], [] in
```
crusso, line 273: Thanks!

crusso (approved):
Fire the hole!

PR #4611, opened 2024-07-12
IR: don't create unnecessary variable renaming for the sake of `NewObjE`

ggreif:
`NewObjE` wants variable names for field values. When the expression is _already a variable_ then just reuse its name!

Also, if we managed to eliminate all necessary variable (re-)bindings, then the enclosing new scope (`BlockE`) is redundant.

By generating less IR we speed up compilation, and potentially create opportunities for further simplifications.

Note: To make this work I had to allow initialisation of `NewObjE` immutable fields from mutable variables. This necessitated changes in both the `interpret_ir.ml` and `check_ir.ml`. Now the interpreters and the backend behave the same.

TODO:
- [x] tests
- [ ] `var field = immutVar` (not possible — unless we intro a _varenv_ that teaches us that the source `immutVar` is a `LetD`, and even then the backend is tricky*)

If we don't want a varenv in `desugar.ml` then we must also admit `var field = mutVar` without the fresh `VarD` in the enclosing block, furthermore:
* Why is the backend tricky? `get_aliased_box` when called with an _rvalue_ would have to allocate a mutbox on the fly, fill it, and return the pointer.
* the IR-interpreter would have to examine the `{T.Mut/non-Mut}`⨯`{V.Mut/non-Mut}` matrix and do something like
``` OCaml
   let v = match f.note, Lib.Promise.value (find f.it.var env.vals) with
     | T.Mut _, V.Mut v -> V.Mut (ref !v) (* fresh mutable cell *)
     | T.Mut _, _ -> V.Mut reg
     | _, V.Mut v -> !v
     | _, v -> v in
```
`refD` is probably a problem, as it would allocate a fresh `V.Mut` with a reused `ref` in it, and then the first line would destroy the aliasing.

I don't want to think this through. (Examining the experimental flag (`= false`) would probably help.) But `var` fields are seldom, so just don't overcomplicate it.


src/lowering/desugar.ml
```diff
@@ -634,7 +636,8 @@ and obj obj_typ efs bases =
   let ds, fs = map (exp_field obj_typ) efs |> split in
   let ds', fs' = concat_map gap (T.as_obj obj_typ |> snd) |> split in
   let obj_e = newObjE T.Object (append fs fs') obj_typ in
-  I.BlockE(append base_decs (append ds ds'), obj_e)
+  let decs = append base_decs (append (filter_map (fun o -> o) ds) ds') in
+  if decs = [] then obj_e.it else I.BlockE(decs, obj_e)
```
ggreif, line 640: this could be `blockE`

ggreif: Done: 75143f131.

src/ir_def/check_ir.ml
```diff
@@ -1067,6 +1067,7 @@ and type_exp_field env s f : T.field =
     with Not_found -> error env f.at "field typing for %s not found" name
   in
   assert (t <> T.Pre);
+  let t = T.(match f.note with Mut _ -> t | _ -> as_immut t) in
```
ggreif, line 1070: This covers the subsumption that an immutable field can be initialised from a mutable variable by just dereferencing the mutcell.

src/lowering/desugar.ml
```diff
@@ -592,15 +592,17 @@ and exp_field obj_typ ef =
...
     in
     assert (not (T.is_mut typ));
-    let id' = fresh_var id.it typ in
-    let d = letD id' (exp e) in
+    let e = exp e in
+    let id', d = match e.it with
+    | I.VarE v -> var v typ, None
+    | _ -> let id' = fresh_var id.it typ in id', Some (letD id' e) in
     let f = { it = I.{ name = id.it; var = id_of_var id' }; at = no_region; note = typ } in
     (d, f)
```
crusso, line 607:
```suggestion
    let id', d_opt = match e.it with
    | I.VarE v -> var v typ, None
    | _ -> let id' = fresh_var id.it typ in id', Some (letD id' e) in
    let f = { it = I.{ name = id.it; var = id_of_var id' }; at = no_region; note = typ } in
    (d_opt, f)
```

test/run-drun/4611.mo
```diff
@@ -0,0 +1,10 @@
+import Prim "mo:⛔";
+
+actor {
+    var bool = true;
+    let text = "Hello World!";
+    stable let rec = { bool; text };
+    stable let mrec = { var bool; var text };
```
crusso, line 7:
```suggestion
    stable let mrec = { var bool; var text };
    bool := false;
    assert mrec.bool;  // check no aliasing between bool and mrec.bool
```

crusso (approved):
LGTM, but not sure we should do too much optimization in desugaring...

ggreif:
> LGTM, but not sure we should do too much optimization in desugaring...

The good thing with doing some simplification in `desugar.ml` is that all following passes have less to chew on. Of course the simplifications must be safe.

PR #4614, opened 2024-07-12
IR: `DotPrim` elim

ggreif:
`DotPrim` on a literal object can be reduced.

Records/objects are expensive to create (heap- and cycles-wise), so doing a compile-time garbage-collection seems reasonable.

TODO:
- [ ] use `FileCheck` on `//MOC-FLAG -dl` output?
- [ ] other types of definition junk? (must not trap to qualify!)
  - [x] `DotPrim` from `VarE`?
  - [x] `NewObjE`?


luc-blaeser:

PR #4615, opened 2024-07-14
improve n-ary continuation coercion

ggreif:


src/ir_passes/async.ml
```diff
@@ -78,16 +78,9 @@ let new_nary_async_reply ts =
...
       let k = fresh_var "k" (contT u T.unit) in
-      let r = fresh_var "r" (err_contT T.unit) in
-      [k; r] -->* (
-        varE unary_async -*-
-          (tupE [
-             [v] -->* (varE k -*- varE v);
-             varE r
-          ])
-      )
+      varE (var "@coerce_and_cont" (unary_async --> ([k; fail] -->* (varE unary_async -*- tupE [varE unary_fulfill; varE fail])) |> typ))
```
ggreif, line 82: actually this is for 0-ary continuations (impedance matching to unary, with unit)

Can we improve below?

ggreif: Nope, we really get pairs etc., not only unit. Fixing...

crusso: Is there something wrong with the original code? The replacement actually seems way more convoluted. I would revert, tbh.

src/ir_passes/async.ml
```diff
@@ -91,16 +108,22 @@ let new_nary_async_reply ts =
...
     | [t1] ->
-      begin
-      match T.normalize t1 with
-      | T.Tup _ ->
-        (* TODO(#3740): find a better fix than PR #3741 *)
-        (* HACK *)
-        coerce t1
+      begin match T.normalize t1 with
+      | T.Tup [] ->
+        varE (var "@coerce_and_cont00" coerce_and_cont0T) -*- varE unary_async
```
ggreif, line 113: seems like this case never gets hit

test/run-drun/await.mo
```diff
@@ -66,6 +66,9 @@ actor a {
        Prim.debugPrint a;
        Prim.debugPrint b;
     };
+    Prim.debugPrint (debug_show (await p()));
+    Prim.debugPrint (debug_show (await async (1, "two", '3')));
```
ggreif, line 70: I want to exercise the higher tuples too

luc-blaeser:

PR #4617, opened 2024-07-15
fix: promote parameters that are bounded with function types

### LINKED ISSUE #4616: parameter bounds not promoted for execution
> $ cat contra-bound.mo
``` Motoko
func foo<T, K <: T -> ()>(k : K, v : T) = k v;

foo<Nat, Int->()>(func (i : Int){}, 1);
```
In the interpreter this dies with:
> $ moc -i contra-bound.mo 
> Motoko compiler (source 0.11.2-14-g2d6fe4a6b-dirty)
> contra-bound.mo:1.45-1.46: internal error, Invalid_argument("call_conv_of_typ K")

In the compiler too:
> Fatal error: exception Invalid_argument(“[Type.as](http://Type.as)_func”)
> Raised at Mo_types__Type.invalid in file “mo_types/type.ml”, line 587, characters 16-54
> Called from Codegen__Compile.compile_prim_invocation in file “codegen/compile.ml”, line 10580, characters 45-74

ggreif:
so that execution can observe

fixes #4616

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,11 @@
 # Motoko compiler changelog
 
+## 0.11.3 (2024-07-XX)
+
+* motoko (`moc`)
+
+  * bugfix: Corrects the interpreter (and compiler) to recognise certain type parameters as callable function types (#4617).
+
 ## 0.11.2 (2024-07-06)
 
 * motoko (`moc`)


src/codegen/compile.ml
```diff
@@ -10577,7 +10577,7 @@ and compile_prim_invocation (env : E.t) ae p es at =
   begin match p, es with
   (* Calls *)
   | CallPrim _, [e1; e2] ->
-    let sort, control, _, arg_tys, ret_tys = Type.as_func e1.note.Note.typ in
+    let sort, control, _, arg_tys, ret_tys = Type.(as_func (promote e1.note.Note.typ)) in
```
ggreif, line 10580: see why I don't use `as_func_sub` in 67f0b8744ac620f2f46edffd99a9c81c0dbf0442.

crusso (approved):
LGTM! Nice catch and fix

PR #4618, opened 2024-07-15
chore: update docs

ggreif:


jessiemongeon1:
It looks like the revised URL leads to a broken page, while the current URL leads to the expected page on Candid serialization.

ggreif:
@jessiemongeon1 so this should better not go in, right?

ggreif:
I wonder where the stale link comes from, then.

jessiemongeon1:
@ggreif Correct

dfx-json (approved):

PR #4622, opened 2024-07-18
experiment: fix bug in try-finally checking

crusso:
follow on to #4507 

The code was checking the effect of the finally clause in only one of the cases (inference or checking, but not both). Plus it's more consistent to just restrict the capability in the environment rather than checking the effect.

The error message is slightly worse, but then this is a rather obscure feature that I hope won't get used much....


test/fail/ok/try-finally.tc.ok
```diff
@@ -1,5 +1,6 @@
-try-finally.mo:10.17-10.31: type error [M0200], a cleanup clause must not send messages
-try-finally.mo:16.17-16.39: type error [M0200], a cleanup clause must not send messages
+try-finally.mo:10.26-10.29: type error [M0047], send capability required, but not available
```
ggreif, line 1: One could add an `.md` explanation file to give users a hint how to fix this (`M0047`).

ggreif: I'd be happy to do that if you think that's a good idea.

ggreif (approved):
LGTM! Rejecting `throw` a priori is certainly better than observing the effect a posteriori. Thanks for the fix!

crusso:
> LGTM! Rejecting `throw` a priori is certainly better than observing the effect a posteriori. Thanks for the fix!

Also, at one point in the past, sending a message didn't actually have a T.Await effect, IIRC, just await and other cps requiring operations did. Now they do, because we detect message queue full errors and report them as catchable errors.

test/fail/ok/try-finally.tc.ok
crusso: I've merged into your branch. Up to you if you want to add an .md file... If you do, there's some sections in the doc about where sends are allowed that you could re-use... That could also be a completely separate PR of course.

PR #4623, opened 2024-07-19
bug: `NewObjE` initialisation short-circuiting with mutable variables is changing behaviour

ggreif:
This PR extends testcase `run-drun/4611.mo` to demonstrate evaluation order problem in the presence of side-effects.

For reading mutable variables it is relevant when the read happens. If there is an assignment to the mutable cell in-between two reads (also via some call or aliased cell) then the read result will differ.

In #4611 we change the point in time when a mutable variable was read, so a test can be fabricated to create a discrepancy between the reference interpreter and the other executions.

This PR backs out the flawed logic of #4611, but keep the test.

To properly fix such an optimisation one needs a hint whether the variable we read from is mutable. If so, the variable rebinding needs to be kept.

One could (contextually) lower mutable reads to a primitive (see C compiler's `volatile` lowering) or have `VarE` with an additional type field (`ref typ`) that gets filled by the type checker.

### DIFF
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -75,10 +75,10 @@
         "homepage": null,
         "owner": "dfinity",
         "repo": "motoko-base",
-        "rev": "732b86ef5bf37ddba65eb48fc750d677ad243c96",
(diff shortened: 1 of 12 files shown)


crusso (approved):
LTGM

(might have been easier to literally, not manually, revert the commit)

PR #4624, opened 2024-07-22
experiment: perf and allocation counter based idl decoder limiting

crusso:
periodically check the instruction counter during IDL decoding to prevent abuse

- [x] Set limit relative to initial perf_counter and some instruction limit based allowance
- [x] Disable for destabilization (iff Registers.get_rel_buf_opt is zero)
- [x] Port new candid spacebomb test suite to drun-tests, to test against real perf counter provided by drun. 
- [x] Add pseudo idl instruction counter for running under wasmtime.
- [x] Bump candid dependency to most recent
- [x] Pass new spacebomb tests, both in candid test suite on wasmtime using pseudo counter


src/codegen/compile.ml
```diff
@@ -5131,7 +5131,7 @@ module IC = struct
     )
 
   (* For debugging *)
-  let _compile_static_print env s =
+  let compile_static_print env s =
```
crusso, line 5134: Revert before merge

rts/motoko-rts/src/idl.rs
```diff
@@ -314,7 +319,9 @@ unsafe fn skip_any_vec(buf: *mut Buf, typtbl: *mut *mut u8, t: i32, count: u32)
         // makes no progress. No point in calling it over and over again.
         // (This is easier to detect this way than by analyzing the type table,
         // where we’d have to chase single-field-records.)
-        return;
+        for _ in 1..count {
+            idl_limit_check();
+        }
```
crusso, line 324:
```suggestion
        }
        return;
```
Oops

rts/motoko-rts/src/idl.rs
```diff
@@ -314,6 +319,9 @@ unsafe fn skip_any_vec(buf: *mut Buf, typtbl: *mut *mut u8, t: i32, count: u32)
         // makes no progress. No point in calling it over and over again.
         // (This is easier to detect this way than by analyzing the type table,
         // where we’d have to chase single-field-records.)
+        for _ in 1..count {
```
crusso, line 322: Better might be to measure the cost of the previous skip and lower the limits based on a multiplier of that cost and return (without this faux for_loop). Then we would exit even earlier!

PR #4628, opened 2024-07-24
experiment: fix for issue in try-finally translation

crusso:
Fixes an issue in #4507.

The `await`-transformation in #4507, when translating an await, should be able to assume that a cleanup continuation is in scope, rather than insert the default `@cleanup` when none can be found. 
This is just as it can assume a `throw`-continuation is in scope.

Unfortunately, some corner cases (top-level programs) were not declaring the cleanup continuation, only the throw continuation, resulting in obscure compiler crashes in `await.ml`.
The bug was particularly subtle because our test programs in run-drun, if they have additional interpreter declarations following the main actor, are compiled as toplevel programs `ProgU` (not `ActorU`), for which one needs to explicitly define the cleanup continuation because there is an implicit `await` around the asynchronous actor declaration. That implicit `await` needs both a `throw`- and a cleanup-cont.

 All very ugly.

### DIFF
--- a/src/ir_passes/await.ml
+++ b/src/ir_passes/await.ml
@@ -341,18 +341,17 @@ and c_exp' context exp k =
                note = Note.{ exp.note with typ = typ' } }))
     end)
   | TryE (exp1, cases, finally_opt) ->
-     let pre k =
-       match finally_opt with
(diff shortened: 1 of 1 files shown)


ggreif (approved):
Thanks for figuring this out! I can follow the description.

PR #4632, opened 2024-07-26
experiment: refactor try-finally to avoid potential code duplication

crusso:
Simplify the await transformation.

For the await transformation, we need to precompose any continuations that could exit a finally block with a call to the finally thunk.

Previously, we would do that by mapping over the continuation context, replacing named continations by new metaconts, and metaconts by composed metaconts, leading to potential code duplication.

An alternative is to simply enclose the try finally block with additional named continuations that pre-compose the finally handler,  avoding the introduction of more meta-continuations in the context.

Indeed, simple observation reveals that meta-continations  never needed to be stored in the context anyway, leading to further simplification in follow on PR #4633.

### DIFF
--- a/src/ir_passes/await.ml
+++ b/src/ir_passes/await.ml
@@ -100,7 +100,7 @@ let preconts context vthunk scope =
      context
      ([], LabelEnv.empty)
   in
-  blockE ds (scope context)
+  blockE ds (scope ctxt)
 
 let typ_cases cases = List.fold_left (fun t case -> T.lub t (typ case.it.exp)) T.Non cases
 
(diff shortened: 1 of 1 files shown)


src/ir_passes/await.ml
```diff
@@ -39,17 +39,7 @@ let letcont k scope =
...
-let precont k vthunk =
-  let finally e = blockE [expD (varE vthunk -*- unitE ())] e in
-  match k with
-  | ContVar k' ->
-     let typ = match typ_of_var k' with
-       | T.(Func (Local, Returns, [], ts1, _)) -> T.seq ts1
-       | _ -> assert false in
-    MetaCont (typ, fun v -> finally (varE k' -*- varE v))
-  | MetaCont (typ, cont) ->
-    MetaCont (typ, fun v -> finally (cont v))
```
crusso, line 52: This is no longer necessary.

crusso:
Merge with https://github.com/dfinity/motoko/pull/4633.

PR #4633, opened 2024-07-26
experiment: simplify await.ml

crusso:
Simple inspection reveals that we never add MetaCont to the context, which means we can simplify the code to enforce that invariant and consider fewer cases. This is an experiment to that effect, building on #4632.


ggreif (approved):
LGTM! Thanks for cleaning up!

PR #4637, opened 2024-07-27
optimisation: annotations to track (reads from) mutable declarations and adapt `Ir.VarE`

ggreif:
By making reads from mutable variables distinguishable, we restore the referential transparency of `VarE (Const, id)`. This was the reason why #4611 failed (and had to be reverted in #4623).

This opens up a bunch of optimisation possibilities
- `VarE` path compression
- dead `LetD` elimination
- `BlockE`/`NewObjE` simplification
- nested `BlockE` merging (given the defs are disjoint and outer uses don't reference inner ones; use `freevars.ml`)
- etc.

In the long run this may allow us to extend the IR effect system with {`Load`, `Store`} effects, and thus pave the way for transactional memory.

We now have a note on `Syntax.VarE`'s identifier, being immutable by default. When performing type checking, this field can change to signify a mutable variable source after looking up the identifier in the environment. Then desugaring uses mutability info to create either `IR.VarE (Const, _)` (for immutable) or a`IR.VarE (Var, _) (for mutable). The IR type checker then asserts that the right kind of access is used.

The Wasm output should be unchanged.

_NOTE_: with this PR `git revert 13d9d5748700662d30d54d6ccfcf2787deac482e` can be done and it passes. So it seems to work 😀


src/mo_def/syntax.ml
```diff
@@ -22,6 +22,7 @@ type resolved_import =
 (* Identifiers *)
 
 type id = string Source.phrase
+type idm = (string, bool) Source.annotated_phrase
```
ggreif, line 25: Instead of being boolean blind we could have `type var_source = Mutable | Immutable` here. WDYT?

dfx-json (approved):

src/ir_interpreter/interpret_ir.ml
```diff
@@ -316,6 +316,11 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
         check_call_conv_arg env exp v2 call_conv;
         last_region := exp.at; (* in case the following throws *)
         f (context env) v2 k
+      | MutReadPrim id, [] ->
```
crusso, line 319: It's odd that this operation doesn't actually do the read, it just returns the binding of the location, exactly like the VarE case.

I guess the reading is actually done elsewhere.

To me this suggest it might be best to just add a mutability field to VarE itself, instead of a misleadingly named MutReadPrim that also needs to be special cased everywhere because it has an identifier payload that contributes, eg. to freevar calc.

src/mo_def/syntax.ml
crusso: Sure. there might even be a type you can reuse for that: Syntax.mut'.

crusso (changes requested):
See comments - happy to discuss on zoom

src/ir_def/construct.ml
```diff
@@ -76,7 +76,9 @@ let wildP =
 (* Primitives *)
 
 let varE (id, typ) =
```
crusso, line 78: If we just add a flag to VarE, then varE would always construct a VarE, not one of two diffferent things.

src/mo_def/syntax.ml
ggreif: Yep, that's a great suggestion!

src/ir_interpreter/interpret_ir.ml
ggreif: Hmm, I wonder how many places I have touch in the IR-related source files. Having `VarE (Const, id)` is a bit clumsy and wild-carding as `VarE _` (or `VarE (_, id)`) still carries the risk of messing up with mutable code.
I am not opposed to this at all, though. And I appreciate its charm. Let's see how it works out.

src/ir_def/construct.ml
ggreif: Correct, there is some charm to it!

src/ir_interpreter/interpret_ir.ml
ggreif: Turns out, I didn't touch _that_ many places, see 1273d3ea8.

src/ir_passes/await.ml
```diff
@@ -48,7 +48,7 @@ let ( -@- ) k exp2 =
     varE v -*- exp2
   | MetaCont (typ0, k) ->
     match exp2.it with
-    | VarE v -> k (var v (typ exp2))
+    | VarE (Const, v) -> k (var v (typ exp2))
```
ggreif, line 51: Hmmm, this could have been a real bug. But I think this is a very fringe case and the time when a mutable variable is recorded is not really specified.

src/ir_passes/erase_typ_field.ml
```diff
@@ -101,7 +101,7 @@ let transform prog =
     let exp' = exp.it in
     match exp' with
     | LitE _ -> exp'
-    | VarE id -> exp'
+    | VarE (_, id) -> exp'
```
ggreif, line 104: join with prev line

src/ir_passes/tailcall.ml
```diff
@@ -92,12 +92,12 @@ and assignEs vars exp : dec list =
     List.mapi (fun i v -> expD (assignE v (projE (varE v) i))) vars
 
 and exp' env e  : exp' = match e.it with
-  | (VarE _ | LitE _) as it -> it
+  | (VarE (Const, _) | LitE _) as it -> it
```
ggreif, line 95: We can probably extend the pattern to mutable refs too.

ggreif: happened in 9584b52cf

src/mo_def/syntax.ml
ggreif: Done: 53f1ec738930cfff761af7cc57599ab33f300e17

src/ir_def/check_ir.ml
```diff
@@ -387,11 +387,19 @@ let rec check_exp env (exp:Ir.exp) : unit =
...
-  | VarE id ->
-    let { typ; loc_known; const } =
+  | VarE (Const, id) ->
+    let { typ; _ } =
       try T.Env.find id env.vals
       with Not_found -> error env exp.at "unbound variable %s" id
     in
+    assert (not (T.is_mut typ));
+    typ <: t
+  | VarE (Var, id) ->
```
ggreif, line 397: consolidate with above?

src/ir_passes/erase_typ_field.ml
ggreif: Done

src/ir_def/check_ir.ml
ggreif: Done in bcb93304e.

src/ir_def/check_ir.ml
```diff
@@ -859,7 +858,7 @@ let rec check_exp env (exp:Ir.exp) : unit =
   if exp.note.Note.const
   then begin
     match exp.it with
-    | VarE id -> check_var "VarE" id
+    | VarE (Const, id) -> check_var "VarE" id
```
crusso, line 861: So if the flag isn't Const, we now fall through? I guess that's ok

src/ir_def/construct.ml
```diff
@@ -77,8 +77,8 @@ let wildP =
...
   if T.is_mut typ
-  then { it = PrimE (MutReadPrim id, []); at = no_region; note = Note.{ def with typ = T.as_immut typ } }
-  else { it = VarE id; at = no_region; note = Note.{ def with typ } }
+  then { it = VarE (Var, id); at = no_region; note = Note.{ def with typ = T.as_immut typ } }
+  else { it = VarE (Const, id); at = no_region; note = Note.{ def with typ } }
```
crusso, line 81: I think you could push the outer if into then VarE of the then branch, and drop the else branch.

src/ir_passes/await.ml
crusso: Interesting. Definitely better to introduce the let when is mutable by falling through. I wonder if  the case that v is mutable ever arose?

src/mo_def/syntax.ml
```diff
@@ -33,6 +33,7 @@ type func_sort = Type.func_sort Source.phrase
 
 type mut = mut' Source.phrase
 and mut' = Const | Var
+and idm = (string, mut') Source.annotated_phrase
```
crusso, line 36: Maybe a better name would be id_occ for identifier occurence, or id_ref for identifier reference?

crusso (approved):
LGTM (see minor suggestions)

src/ir_passes/await.ml
ggreif: I doubt it very much as the `v` carries the result of the previous computation, and CPS never allocates mutable bindings.

src/ir_def/check_ir.ml
ggreif: Yes, we'd get "unexpected constant expression" below. Never happens.

src/mo_def/syntax.ml
ggreif: Sure! c007a8a3ebc640fd9a35acc32c0666450dfd3da7

src/ir_def/construct.ml
ggreif: Indeed: 31cf9cc8a

PR #4640, opened 2024-07-30
docs: Add Motoko footer image

jessiemongeon1:
Adds a footer image to all Motoko docs (aside from Base reference pages) 

Screenshot 2024-07-30 at 3 48 11 PM


ggreif (approved):
This looks good to me. We should not fall in the trap of making technical content _too_ flashy.

PR #4645, opened 2024-08-02
IR: don't create unnecessary variable renaming for the sake of `NewObjE`

ggreif:
This reapplies #4611 without changes (that were previously) necessary to the interpreter and IR checker. PR #4611 had to be backed out due to changed behaviour w.r.t. mutable variables: #4623.

### From #4611:

`NewObjE` wants variable names for field values. When the expression is _already a variable_ then just reuse its name!

Also, if we managed to eliminate all necessary variable (re-)bindings, then the enclosing new scope (`BlockE`) is redundant.

By generating less IR we speed up compilation, and potentially create opportunities for further simplifications.

Another positive is that reading the IR (`-dl` flag) becomes less convoluted.

_Note_: Since the IR recently has been taught to track whether the read is from an immutable or from a mutable variable, we can now observe renaming of immutable variables and eliminate only those.

### DIFF
--- a/src/ir_def/arrange_ir.ml
+++ b/src/ir_def/arrange_ir.ml
@@ -13,7 +13,7 @@ let prim_ty p = typ (Type.Prim p)
 let kind k = Atom (Type.string_of_kind k)
 
 let rec exp e = match e.it with
-  | VarE (_, i)         -> "VarE"    $$ [id i] (* FIXME: EXPOSE *)
+  | VarE (m, i)         -> (if m = Var then "VarE!" else "VarE") $$ [id i]
   | LitE l              -> "LitE"    $$ [lit l]
   | PrimE (p, es)       -> "PrimE"   $$ [prim p] @ List.map exp es
   | AssignE (le1, e2)   -> "AssignE" $$ [lexp le1; exp e2]
--- a/src/ir_def/construct.ml
+++ b/src/ir_def/construct.ml
@@ -670,7 +670,7 @@ let (-->*) xs exp =
   nary_funcE "$lambda" fun_ty xs exp
 
 let close_typ_binds cs tbs =
-  List.map (fun {it = {con; sort; bound}; _} -> {T.var = Cons.name con; sort=sort; bound = T.close cs bound}) tbs
+  List.map (fun {it = {con; sort; bound}; _} -> {T.var = Cons.name con; sort; bound = T.close cs bound}) tbs
 
 (* polymorphic, n-ary local lambda *)
 let forall tbs e =
@@ -776,7 +776,7 @@ let objE sort typ_flds flds =
                @ fld_tys)))
     | (lab, exp)::flds ->
       let v, addv = match exp.it with
-        | VarE v -> var v (typ exp), fun _ ds -> ds
+        | VarE (Const, v) -> var v (typ exp), fun _ ds -> ds
         | _ ->
           let v = fresh_var lab (typ exp) in
           v, fun exp ds -> letD v exp :: ds in
--- a/src/ir_def/construct.ml
+++ b/src/ir_def/construct.ml
@@ -775,17 +775,17 @@ let objE sort typ_flds flds =
               ((List.map (fun (id,c) -> (id, T.Typ c)) typ_flds)
                @ fld_tys)))
     | (lab, exp)::flds ->
-      let v, addv = match exp.it with
-        | VarE (Const, v) -> var v (typ exp), fun _ ds -> ds
+      let v, ds1 = match exp.it with
+        | VarE (Const, v) -> var v (typ exp), ds
         | _ ->
           let v = fresh_var lab (typ exp) in
-          v, fun exp ds -> letD v exp :: ds in
+          v, letD v exp :: ds in
       let field = {
         it = {name = lab; var = id_of_var v};
         at = no_region;
         note = typ exp
       } in
-      go (addv exp ds) (field::fields) ((lab, typ exp)::fld_tys) flds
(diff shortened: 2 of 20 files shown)


src/ir_def/arrange_ir.ml
```diff
@@ -13,7 +13,7 @@ let prim_ty p = typ (Type.Prim p)
 let kind k = Atom (Type.string_of_kind k)
 
 let rec exp e = match e.it with
-  | VarE (_, i)         -> "VarE"    $$ [id i] (* FIXME: EXPOSE *)
+  | VarE (m, i)         -> (if m = Var then "VarE!" else "VarE") $$ [id i]
```
ggreif, line 16: This was forgotten in #4637.

src/ir_def/construct.ml
```diff
@@ -776,13 +775,17 @@ let objE sort typ_flds flds =
               ((List.map (fun (id,c) -> (id, T.Typ c)) typ_flds)
                @ fld_tys)))
     | (lab, exp)::flds ->
-      let v = fresh_var lab (typ exp) in
+      let v, addv = match exp.it with
+        | VarE (Const, v) -> var v (typ exp), fun _ ds -> ds
+        | _ ->
+          let v = fresh_var lab (typ exp) in
+          v, fun exp ds -> letD v exp :: ds in
```
crusso, line 782:
```suggestion
      let v, ds1 = match exp.it with
        | VarE (Const, v) -> var v (typ exp), ds
        | _ ->
          let v = fresh_var lab (typ exp) in
          v, letD v exp :: ds in
```

src/ir_def/construct.ml
```diff
@@ -776,13 +775,17 @@ let objE sort typ_flds flds =
...
       let field = {
         it = {name = lab; var = id_of_var v};
         at = no_region;
         note = typ exp
       } in
-      go ((letD v exp)::ds) (field::fields) ((lab, typ exp)::fld_tys) flds
+      go (addv exp ds) (field::fields) ((lab, typ exp)::fld_tys) flds
```
crusso, line 788:
```suggestion
      go ds1 (field::fields) ((lab, typ exp)::fld_tys) flds
```

src/lowering/desugar.ml
```diff
@@ -661,17 +661,19 @@ and exp_field obj_typ ef =
     let id' = fresh_var id.it typ in
     let d = varD id' (exp e) in
     let f = { it = I.{ name = id.it; var = id_of_var id' }; at = no_region; note = typ } in
-    (d, f)
+    (Some d, f)
```
crusso, line 664:
```suggestion
    ([d], f)
```

src/lowering/desugar.ml
```diff
@@ -661,17 +661,19 @@ and exp_field obj_typ ef =
...
       | Some typ -> typ
       | None -> e.note.S.note_typ
     in
     assert (not (T.is_mut typ));
-    let id' = fresh_var id.it typ in
-    let d = letD id' (exp e) in
+    let e = exp e in
+    let id', d_opt = match e.it with
+    | I.(VarE (Const, v)) -> var v typ, None
+    | _ -> let id' = fresh_var id.it typ in id', Some (letD id' e) in
```
crusso, line 674:
```suggestion
    | I.(VarE (Const, v)) -> var v typ, []
    | _ -> let id' = fresh_var id.it typ in id', [letD id' e] in
```

src/lowering/desugar.ml
```diff
@@ -703,7 +705,8 @@ and obj obj_typ efs bases =
   let ds, fs = map (exp_field obj_typ) efs |> split in
   let ds', fs' = concat_map gap (T.as_obj obj_typ |> snd) |> split in
   let obj_e = newObjE T.Object (append fs fs') obj_typ in
-  I.BlockE(append base_decs (append ds ds'), obj_e)
+  let decs = append base_decs (append (filter_map Fun.id ds) ds') in
+  (blockE decs obj_e).it
```
crusso, line 709:
```suggestion
  let decs = append base_decs (append (flatten ds) ds') in
  (blockE decs obj_e).it
```
(and rename this ds to dss since its now a nested list?)

src/lowering/desugar.ml
```diff
@@ -661,17 +661,19 @@ and exp_field obj_typ ef =
...
   | S.Const ->
     let typ = match T.lookup_val_field_opt id.it fts with
       | Some typ -> typ
       | None -> e.note.S.note_typ
     in
     assert (not (T.is_mut typ));
-    let id' = fresh_var id.it typ in
-    let d = letD id' (exp e) in
+    let e = exp e in
+    let id', d_opt = match e.it with
```
crusso, line 672:
```suggestion
    let id', ds = match e.it with
```

src/lowering/desugar.ml
```diff
@@ -661,17 +661,19 @@ and exp_field obj_typ ef =
...
+    | I.(VarE (Const, v)) -> var v typ, None
+    | _ -> let id' = fresh_var id.it typ in id', Some (letD id' e) in
     let f = { it = I.{ name = id.it; var = id_of_var id' }; at = no_region; note = typ } in
-    (d, f)
+    (d_opt, f)
```
crusso, line 676:
```suggestion
    (ds, f)
```

crusso (approved):
Minor suggestions, otherwise looks good.

Did you lose the test for evaluation order preservation? Might be good to keep it.

ggreif:
> Did you lose the test for evaluation order preservation? Might be good to keep it.

I never reverted them. They are in the testsuite.

PR #4648, opened 2024-08-05
EOP: Rust Upgrade for RTS

luc-blaeser:
Upgrading integrated EOP (https://github.com/dfinity/motoko/pull/4488) to recent Rust and LLVM dependencies:
* LLVM 18.1.8
* Rust 1.82.0

Special aspects:
* Support for both `wasm32` and `wasm64`.
* Support new `dylink.0` instead of `dylink`.
* Support `GOT.mem` for RTS in the Motoko linker.
* LLVM LLD (used for `wasm-ld`) needs a bug fix: https://github.com/llvm/llvm-project/pull/104043. Therefore, patching `wasm-ld`. 
* Float formatting is now based on Rust and no longer on musl. There are format changes to output of `NaN` and `infinity`. Moreover, one particular output format (hexadecimal float) is no longer supported. Updating the documentation and base library is needed.
* `musl` is still used in 32-bit but not in 64-bit. In the future, we should get rid of `musl` entirely. The `musl` import is rather old, upgrading is complicated as it involves a lot of unwanted dependencies.
* No longer using `emscripten` for 64-bit Wasm compilation (as there is a nix build issue with latest `emscripten`). Instead, the `clang` compiler is directly invoked.
* Adjusting the RTS code in line with the latest Rust features/rules.

Note:
* If your `nix-shell` crashes with a segmentation fault with this PR, you need to upgrade `nix`. For this, you need to manually uninstall and re-install `nix`.


src/linking/linkModule.ml
```diff
@@ -58,37 +58,50 @@ Similarly exposing a data pointer generates a GOT.mem import. All GOT.mem
...
 
+- After merging the sections, we replace the GOT imports with globals in the 
+  module section.
+
+  For each `GOT.func`, we create an element in the table with the function index
+  determined in the first step. The `GOT.func` import is replaced by a global
+  that refers to the new table element.
+
+  For each `GOT.mem`, we patch all accesses to this global in the module's AST to 
+  refer to the global that impplements the data pointer as determined in the first
```
crusso, line 90:
```suggestion
  refer to the global that implements the data pointer as determined in the first 
```

src/linking/linkModule.ml
```diff
@@ -58,37 +58,50 @@ Similarly exposing a data pointer generates a GOT.mem import. All GOT.mem
...
+  module section.
+
+  For each `GOT.func`, we create an element in the table with the function index
+  determined in the first step. The `GOT.func` import is replaced by a global
+  that refers to the new table element.
+
+  For each `GOT.mem`, we patch all accesses to this global in the module's AST to 
+  refer to the global that impplements the data pointer as determined in the first 
+  step. The `GOT.mem` import is replaced by a dummy global that only serves to
+  maintain the the numbering of the globals.
```
crusso, line 92:
```suggestion
  maintain the numbering of the globals.
```

src/linking/linkModule.ml
luc-blaeser: Thank you!

src/linking/linkModule.ml
luc-blaeser: Thank you!

rts/Makefile
```diff
@@ -136,9 +138,21 @@ _build/wasm32:
...
-RTS_EMSCRIPTEN_64_DEBUG=$(RTS_EMSCRIPTEN_64)
-RTS_EMSCRIPTEN_64_RELEASE=$(RTS_EMSCRIPTEN_64) -O2
+# Arguments obtained by Emscripten 3.1.64:
+# `emcc -c --target=wasm64-unknown-emscripten -fpic --verbose`
+# while skipping unneeded arguments.
+RTS_LL_64=$(CLANG) -cc1 -triple wasm64-unknown-emscripten -emit-obj \
+	-mrelocation-model pic -pic-level 1 -mframe-pointer=none \
+	-ffp-contract=on -fno-rounding-math -mconstructor-aliases \
+	-target-feature +mutable-globals -fvisibility=hidden \
+	-resource-dir $(WASM_CLANG_LIB)/lib/clang/19/ \
```
luc-blaeser, line 148: This should be `clang/18`. Probably the option is not needed and can be removed...

luc-blaeser:
Closing in favor of https://github.com/dfinity/motoko/pull/4677 to merge with latest `master`.

PR #4654, opened 2024-08-08
chore: consolidate `try`/`finally` tests

ggreif:
The error message for `finally { ... async ... }` is a bit bizarre, but for now I can live with it.


crusso (approved):
LGTM

PR #4657, opened 2024-08-09
experiment: value quota based idl decoder limiting

crusso:
Simplifies #4624 to a simple linear limit on the number of decoded values as a function of decoded payload size,
instead of using two linear functions on perfcounter (simulated or real) and allocation counter.

The function is:

value_quota(blob) : Nat64 = blob.size() * (numerator/denominator) + bias

where blob is the candid blob to be decoded, and `numerator` (default 1), `denominator` (default 1) and `bias` (default 1024) are `Nat32s`.

Much simpler than #4624 and doesn't depend on vagaries of instruction metering and byte allocation which varies with gc and compiler options, but is it good enough?

The constants can be (globally) modified/inspected using prims (Prim.getCandidLimits/Prim.setCandidLimits) which will need to get exposed in base eventually.

The quota is decremented on every call to deserialise or skip a value in vanilla candid mode (destabilization is not metered).
The quota is eagerly checked before deserializing or skipping arrays.

One possible refinement would be to combine the value quota with a memory quota (though the latter would still vary with gc flavour and perhaps word-size unless we count logical words)

- [x] Disable for destabilization (iff Registers.get_rel_buf_opt is zero)
- [x] Port new candid spacebomb test suite to drun-tests, to test against real perf counter provided by drun. 
- [x] Bump candid dependency to most recent
- [x] Pass new spacebomb tests, both in candid test suite on wasmtime using value counter.


src/codegen/compile.ml
```diff
@@ -6531,14 +6536,6 @@ module MakeSerialization (Strm : Stream) = struct
     G.i (GlobalGet (nr (E.get_global env "__typtbl_idltyps")))
 
   module Registers = struct
-    let register_globals env =
```
luc-blaeser, line 6534: Just an idea (of lowest importance, please feel free to ignore): In light of merging (and structure), we could maybe keep and extend this function at this place and pass the initial metering limit parameters as (structured) argument. The lower metering initialization logic could then call this register declaration function (e.g. with a name like `initialize`), while the module initialization calls again this more general initializer.

src/codegen/compile.ml
```diff
@@ -6568,6 +6565,87 @@ module MakeSerialization (Strm : Stream) = struct
...
+    let set_value_denominator env =
+      G.i (GlobalSet (nr (E.get_global env "@@value_denominator")))
+
+    let get_value_bias env =
+      G.i (GlobalGet (nr (E.get_global env "@@value_bias")))
+    let set_value_bias env =
+      G.i (GlobalSet (nr (E.get_global env "@@value_bias")))
+
+    (* interval for checking instruction counter *)
+    let idl_value_numerator = 1l
```
luc-blaeser, line 6590: Tuning these parameters will be interesting :)
In any case, I like this simple solution compared to the former experiment.

rts/motoko-rts/src/idl.rs
```diff
@@ -11,6 +11,11 @@ use core::cmp::min;
 
 use motoko_rts_macros::ic_mem_fn;
 
+extern "C" {
+    // check instruction decoding limit, exported by moc
+    pub fn idl_limit_check(value_count: u32);
```
luc-blaeser, line 16: Should we maybe consider using `u64` just in case to avoid/defer overflows (in the compiler backend limit calculation)? Maybe not a concern.

src/codegen/compile.ml
```diff
@@ -6568,6 +6565,87 @@ module MakeSerialization (Strm : Stream) = struct
...
+      get_rel_buf_opt ^^
+      G.if0
+      begin (* Candid deserialization *)
+        (* set instruction limit *)
+        get_blob ^^
+        Blob.len env ^^
+        G.i (Convert (Wasm.Values.I64 I64Op.ExtendUI32)) ^^
+        get_value_numerator env ^^
+        G.i (Convert (Wasm.Values.I64 I64Op.ExtendUI32)) ^^
+        G.i (Binary (Wasm.Values.I64 I64Op.Mul)) ^^
```
luc-blaeser, line 6604: Detail: Maybe a comment would be helpful, saying that this can't overflow because the parameters are `Nat32`. 
If the user sets max numerator and max bias (as sentinels to skip metering), perhaps it could overflow. Probably not relevant?

src/codegen/compile.ml
```diff
@@ -11730,6 +11814,26 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
     const_sr (SR.UnboxedWord64 Type.Int64) (Word64.btst_kernel env)
 
+  | OtherPrim "setCandidLimits", [e1; e2; e3] ->
+    SR.unit,
+    compile_exp_as env ae (SR.UnboxedWord32 Type.Nat32) e1 ^^
+    Serialization.Registers.set_value_numerator env ^^
+    compile_exp_as env ae (SR.UnboxedWord32 Type.Nat32) e2 ^^
+    Serialization.Registers.set_value_denominator env ^^
+    Serialization.Registers.get_value_denominator env ^^
+    E.else_trap_with env "Candid limit denominator cannot be zero" ^^
```
luc-blaeser, line 11824: Good!

src/codegen/compile.ml
```diff
@@ -11730,6 +11814,26 @@ and compile_prim_invocation (env : E.t) ae p es at =
   | OtherPrim "btstInt64", [_;_] ->
     const_sr (SR.UnboxedWord64 Type.Int64) (Word64.btst_kernel env)
 
+  | OtherPrim "setCandidLimits", [e1; e2; e3] ->
+    SR.unit,
+    compile_exp_as env ae (SR.UnboxedWord32 Type.Nat32) e1 ^^
+    Serialization.Registers.set_value_numerator env ^^
+    compile_exp_as env ae (SR.UnboxedWord32 Type.Nat32) e2 ^^
```
luc-blaeser, line 11821: Should we maybe limit the numerator below `u32::MAX` to avoid overflow when adding max bias (pls. see comment above)?

src/mo_values/prim.ml
```diff
@@ -382,4 +382,12 @@ let prim trap =
   | "canister_version" ->
       fun _ v k -> as_unit v; k (Nat64 (Numerics.Nat64.of_int 42))
 
+  (* fake *)
+  | "setCandidLimits" ->
+      fun _ v k -> k unit
+  | "getCandidLimits" ->
+      fun _ v k -> k (Tup [
+        Nat64 Numerics.Nat64.zero; Nat64 Numerics.Nat64.zero;
+        Nat64 Numerics.Nat64.zero; Nat64 Numerics.Nat64.zero])
```
luc-blaeser, line 391: I wonder what the fourth item is. And the types are `Nat64` (while the other returned values are `Nat32`). Probably not relevant because it is a fake function.

luc-blaeser:
One thing that I am unsure or I have probably missed (sorry for the potentially stupid question): Is the metering limit also called on ordinary decoding of Candid when it is not skipped?

luc-blaeser (approved):
Very nice PR! 
I have only some minor comments and one question (probably not relevant).

src/mo_values/prim.ml
crusso: Good catch - I forgot to adjust this...

src/codegen/compile.ml
crusso: Oh, I now saturate the bound u64::MAX instead....

src/codegen/compile.ml
crusso: Added a comment and now saturate the bound to u64::MAX on overflow

rts/motoko-rts/src/idl.rs
crusso: Done

crusso:
> One thing that I am unsure or I have probably missed (sorry for the potentially stupid question): Does the metering limit also need to be checked on ordinary decoding of Candid when it is not skipped and not recursively called?

Actually, I'm not sure I understand the question (which worries me). I think all calls, initial, skip and  recursive should be doing the check, but I maybe you've seen something I haven't.

src/codegen/compile.ml
crusso: Ok, I did something slight different that avoids moving register_globals elsewhere...

crusso:
@luc-blaeser PTAL (and thanks for the review!)

chenyan-dfinity (commented):
Can we document the cost model somewhere, so that people can roughly know how to tune the parameters?

luc-blaeser:
> @luc-blaeser PTAL (and thanks for the review!)

Not that I have any concrete concern, I was just curious whether I got all aspects.
If I understood correctly, the metering covers skipping and recursive calls. Also, the recursive call is done for every decoded Candid value. 
I was wondering about the following cases of whether metering could be relevant there or would already be implicitly handled (by the recursion metering). 
* If a value would have dynamic encoded length such as an array of listed elements, I guess the recursive function would be called for every sub-element. This would be fine.
* Would there be a need for size-dependent metering of blobs? I guess there is no recursive call involved per blob byte. 
* IIRC, some constant Candid values can inflate to variable-sized objects (arrays, maybe also blob). This would be a non-linear effort (initialization) that is maybe not metered.
These are just thoughts of mine, probably not relevant.

crusso:
> > @luc-blaeser PTAL (and thanks for the review!)
> 
> Not that I have any concrete concern, I was just curious whether I got all aspects. If I understood correctly, the metering covers skipping and recursive calls. Also, the recursive call is done for each decoded Candid value. I was wondering about the following cases of whether metering could be relevant there or would already be implicitly handled (by the recursion metering).
> 
> * If a value would have dynamic encoded length such as an array of listed elements, I guess the recursive function would be called for each sub-element. This would be fine.
Yes, that's the idea.

> * Would there be a need for size-dependent metering of blobs? I guess there is no recursive call involved per blob byte.

Fortunately, the blob decoder first checks that the blob does not exceed the length of the candid payload, which is limited to 10MB by the IC (in the worst case). The Nat and Int decoders also check they don't decode past the candid payload, so I was hoping that would limit the allocation.

> * IIRC, some constant Candid values can inflate to variable-sized objects (arrays, maybe also blob). This would be a non-linear effort (initialization) that is maybe not metered.

Hmm, maybe one could decrement the quota by number of elements before allocation, and then increment the quota before
deseralizing each element (to compensate for the earlier bulk check). Not sure.


>   These are just thoughts of mine, probably not relevant.

rts/motoko-rts/src/idl.rs
```diff
@@ -314,6 +319,7 @@ unsafe fn skip_any_vec(buf: *mut Buf, typtbl: *mut *mut u8, t: i32, count: u32)
         // makes no progress. No point in calling it over and over again.
         // (This is easier to detect this way than by analyzing the type table,
         // where we’d have to chase single-field-records.)
+        idl_limit_check((count - 1) as u64);
```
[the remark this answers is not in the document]
crusso: No, because we return on previous line 311 when count == 0

crusso:
> Can we document the cost model somewhere, so that people can roughly know how to tune the parameters?
Done in Changelog for now. Will add to base eventually...

PR #4658, opened 2024-08-14
update: docs for Mops packages

jessiemongeon1:


ggreif (approved):
LGTM, but I suggest @rvanasa to have another look...

rvanasa (approved):
Also LGTM!

Issue #4625, opened 2024-07-22
It doesn't understand that the type in canister is the same


vporton:
The bug is triggered by https://github.com/vporton/bug-canister-type ([84b7e50](https://github.com/vporton/bug-canister-type/commit/84b7e5054c43fd1e20b4622dc94265ae49cfa572)):

```
$ dfx deploy 
...
/home/porton/Projects/bug/mytest/src/mytest_backend/main.mo:9.14-9.18: type error [M0096], expression of type
  actor {type X = Null; f : shared X__1 -> async ()}
cannot produce expected type
  actor {f : shared X -> async ()}
```

`f` has the same type in both and should compile without an error.

For completeness, here is the code:

`lib.mo`:

```motoko
module {
    public type X = ();
}
```

`call.mo`:

```motoko
import lib "lib";

actor {
    public shared func f(x: lib.X): async () {
        ();
    };
}
```

`main.mo`:

```motoko
import lib "lib";
import call "canister:call";

actor {
  type T = actor {
    f(x: lib.X): async ();
  };

  let z: T = call;
};
````

`dfx.json`:

```json
{
  "canisters": {
    "mytest_backend": {
      "main": "src/mytest_backend/main.mo",
      "type": "motoko",
      "dependencies": ["call"]
    },
    "call": {
      "main": "src/mytest_backend/call.mo",
      "type": "motoko"
    }
  },
  "defaults": {
    "build": {
      "args": "",
      "packtool": ""
    }
  },
  "output_env_file": ".env",
  "version": 1
}
```

vporton:
```
$ moc --version
Motoko compiler 0.11.2+ (source 3w73qd3r-5pqc5q4p-crdjd8hk-1q2s4w4s)
```

ggreif:
I can reproduce the error. If I have
``` Motoko
module {
    public type X = Int;
}
```
in `lib.mo` then everything deploys just fine. My hunch is that the Candid conversion of `()` to `Null` is the problem.

ggreif:
[...]

PR #4659, opened 2024-08-15
bug: issue-4625 repro

crusso:
Non-dfx repro for issue #4625


test/run-drun/issue-4625/call.mo
```diff
@@ -0,0 +1,7 @@
+import lib "lib";
+
+actor {
+    public shared func f(x: lib.X): async () {
```
ggreif, line 4:
```suggestion
    public shared func f(x : lib.X) : async () {
```

test/run-drun/issue-4625.mo
```diff
@@ -0,0 +1,17 @@
...
+//MOC-FLAG --actor-alias call lg264-qjkae
+import lib "issue-4625/lib";
+import call "canister:call";
+
+actor {
+  type T = actor {
+    f(x: lib.X): async ();
+  };
+
+  let z: T = call;
```
ggreif, line 11:
```suggestion
    f(x : lib.X) : async ();
  };

  let z : T = call;
```

ggreif (dismissed):
LGTM, maybe you can add the 3 missing newlines too.

test/run-drun/issue-4625.mo
```diff
@@ -0,0 +1,17 @@
...
+    f(x : lib.X) : async ();
+  };
+
+  let z : T = call;
+};
+
+//SKIP run
+//SKIP run-ir
+//SKIP run-low
+//SKIP ic-ref
```
crusso, line 17:
```suggestion
//SKIP ic-ref

```

test/run-drun/issue-4625/call.mo
```diff
@@ -0,0 +1,7 @@
+import lib "lib";
+
+actor {
+    public shared func f(x : lib.X) : async () {
+        ();
+    };
+}
```
crusso, line 7:
```suggestion
}

```

ggreif:
Maybe this should be under `test/fail`, then?

crusso:
I'm not sure how fixable this is.  I wonder if design/IDL-Motoko.md is underspecified on how to translate Motoko/Candid references to type definitions, i.e. translate without or after expansion and that affects the translation of argument sequences.



In addition, both Motoko Null and () map to the Candid null.

so type X = () translates to type X = Null
but
func f(x : X) : async X has Candid type () -> ()
 (I think), not, null -> null.

PR #4661, opened 2024-08-16
Experiment: RTS Size Optimizations

luc-blaeser:
Trying to reduce the RTS size regression incurred by the Rust version upgrade.


luc-blaeser:
Applied to https://github.com/dfinity/motoko/pull/4677

PR #4676, opened 2024-08-28
Record Upgrade Instructions for Classical Persistence

luc-blaeser:
Also support `Prim.rts_upgrade_instructions()` on classical persistence, to inspect the instructions consumed by the last upgrade (including stabilization plus destabilization) for performance measurements. This functionality is already supported with enhanced orthogonal persistence.

## Mechanism
1. The pre-upgrade records its consumed instructions, in particular for stabilization, in the stable memory metadata. The location depends on the stable memory version.
2. The post-upgrade loads this information, if present, and adds its consumed instructions, in particular for destabilization.
3. `Prim.rts_upgrade_instructions()` returns the sum of those costs.

## Backwards Compatibility
The record of upgrade instructions is optional in the stable memory. This is to ensure backwards compatibility to older Motoko programs that do not record this information. When upgrading from such an older Motoko program,  `Prim.rts_upgrade_instructions()` returns `Nat64.maximumValue`.

## Notes
* `Prim.rts_upgrade_instructions()` currently returns >0 after installation because some upgrade check needs to be performed there too.
* This function is to be exposed to the Motoko base library together with the other diagnostic runtime information, since `Prim` is only intended for internal use.


src/codegen/compile_classical.ml
```diff
@@ -8660,6 +8690,19 @@ module Stabilization = struct
   let destabilize env ty save_version =
     match E.mode env with
     | Flags.ICMode | Flags.RefMode ->
+      let (set_instructions, get_instructions) = new_local64 env "instructions" in
+      let handle_missing_instructions = 
+        get_instructions ^^
+        compile_eq64_const 0L ^^
+        (G.if0
+        begin
+          (* Default to -1 if if no upgrade instructions were recorded, i.e.
```
crusso, line 8699:
```suggestion
          (* Default to -1 if no upgrade instructions were recorded, i.e. 
```

test/run-drun/ok/migrate-upgrade-instr-stable-mem.drun.ok
```diff
@@ -0,0 +1,7 @@
+ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
+debug.print: Stable memory size: 10
+ingress Completed: Reply: 0x4449444c0000
+debug.print: Stable memory size: 20
+ingress Completed: Reply: 0x4449444c0000
+debug.print: Upgrade instructions: 18_446_744_073_709_551_615
```
crusso, line 6: Why is the instruction count so large? Is it really doing that much work?

crusso (commented):
This looks ok to me but I don't quite understand why the number of instructions is so large?
Or is that just because its undefined and equal to UInt64.max = 18,446,744,073,709,551,615?

test/run-drun/upgrade-instructions.mo
```diff
@@ -11,6 +10,8 @@ actor {
     lastInstructions := Prim.rts_upgrade_instructions();
     assert(lastInstructions > 0);
   };
+  // Assert that the upgrade instructions are defined, i.e. no backwards compatibility to old Motoko programs.
```
crusso, line 13: Ah, ok, I understand now.

crusso (approved):
Ok, I get it now. LGTM. Nice work

src/codegen/compile_classical.ml
luc-blaeser: Thank you!

test/run-drun/ok/migrate-upgrade-instr-stable-mem.drun.ok
luc-blaeser: It is a sentinel (u64::MAX) if the upgrade instructions are unknown when upgrading from an old canister that did not record the pre-upgrade costs.

luc-blaeser: (sorry, just saw your comment below :) )

luc-blaeser:
The `drun` Heisenbug hit us during merge-to-master...

PR #4679, opened 2024-08-30
Test Canister Snapshots

luc-blaeser:
Test canister snapshot with classical and enhanced orthogonal persistence.

Requires a more recent drun version with snapshot fix: https://github.com/luc-blaeser/ic/commit/58b86e65fab2ca931f1734a38a706a7fd9d30990


crusso (approved):
Nice. Might make a good example with a little rewriting to avoid Prim...

luc-blaeser:
> Nice. Might make a good example with a little rewriting to avoid Prim...

Thank you. Indeed, we could add this to the example set (with proper base-lib imports).

PR #4685, opened 2024-09-06
Fix: Start Wasm table for the RTS at offset >= 1

luc-blaeser:
Rust requires a table offset of at least 1 as the element index 0 is considered invalid and causes a debug null check to panic when called. On the other hand, `elem[0]` can be used by the Motoko backend code, as correct Rust-generated Wasm code does not call `elem[0]`.

This bug is independent of `wasm32` and `wasm64`.

This issue has been observed in https://github.com/dfinity/motoko/actions/runs/10703077671/job/29672766216?pr=4683 and happened only on Linux and only under `nix-build` (not `nix-shell`).

### DIFF
--- a/src/linking/linkModule.ml
+++ b/src/linking/linkModule.ml
@@ -846,7 +846,7 @@ let link (em1 : extended_module) libname (em2 : extended_module) =
     )
   end else ();
 
-  let max x y = if x >= y then x else y in
+  let max x y = if x >= y then x else y in (* use `Int.max` when bumping to 4.13 *)
 
   (* Rust requires a table offset of at least 1 as elem[0] is considered invalid. 
      There are debug checks panicking if the element index is zero.
--- a/src/linking/linkModule.ml
+++ b/src/linking/linkModule.ml
@@ -846,7 +846,13 @@ let link (em1 : extended_module) libname (em2 : extended_module) =
     )
   end else ();
 
-  let old_table_size = read_table_size em1.module_ in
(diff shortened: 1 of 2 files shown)


src/linking/linkModule.ml
```diff
@@ -846,7 +846,13 @@ let link (em1 : extended_module) libname (em2 : extended_module) =
     )
   end else ();
 
-  let old_table_size = read_table_size em1.module_ in
+  let max x y = if x >= y then x else y in
```
ggreif, line 849:
```suggestion
  let max x y = if x >= y then x else y in (* use `Int.max` when bumping to 4.13 *)
```

ggreif (commented):
Why didn't the bytecode-verifier in the replica catch this? Do we have Motoko-compiled canisters out there with this bug baked-in?

src/linking/linkModule.ml
luc-blaeser: Thank you. I also noticed it would be in Ocaml 4.13. Good to have a todo comment in the code.

luc-blaeser:
> Why didn't the bytecode-verifier in the replica catch this? Do we have Motoko-compiled canisters out there with this bug baked-in?

It is actually only a requirement by Rust, not Wasm. It only affects debug Rust compilation AFAIK. I believe people probably do not release debug Motoko code. So, probably it is okay.

crusso (approved):
LGTM

ggreif (approved):
Fine by me!

PR #4690, opened 2024-09-12
feat: burning cycles programmatically

ggreif:
This adds a new primitive `(prim "cyclesBurn")` and a `prim.mo` function `cyclesBurn : <system> Nat -> Nat`. It can be used to burn (some of the) canister's cycles programmatically using the system interface `ic0.cycles_burn128`.

### DIFF
--- a/src/codegen/compile_enhanced.ml
+++ b/src/codegen/compile_enhanced.ml
@@ -5387,7 +5387,7 @@ module Cycles = struct
 
   let burn env =
     Func.share_code1 Func.Always env "cycle_burn" ("cycles", I64Type) [I64Type] (fun env get_x ->
-      Stack.with_words env "dst" 4L (fun get_dst ->
+      Stack.with_words env "dst" 2L (fun get_dst ->
         get_x ^^
         to_two_word64 env ^^
         get_dst ^^
--- a/src/codegen/compile_enhanced.ml
+++ b/src/codegen/compile_enhanced.ml
@@ -4731,7 +4731,7 @@ module IC = struct
       E.add_func_import env "ic0" "msg_cycles_available128" [I64Type] [];
       E.add_func_import env "ic0" "msg_cycles_refunded128" [I64Type] [];
       E.add_func_import env "ic0" "msg_cycles_accept128" (i64s 3) [];
-      E.add_func_import env "ic0" "cycles_burn128" [I64Type; I64Type; I32Type] [];
+      E.add_func_import env "ic0" "cycles_burn128" (i64s 3) [];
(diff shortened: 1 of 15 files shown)


luc-blaeser (approved):
Thank you! This looks good!

src/codegen/compile_enhanced.ml
```diff
@@ -5385,6 +5385,18 @@ module Cycles = struct
       )
     )
 
+  let burn env =
+    Func.share_code1 Func.Always env "cycle_burn" ("cycles", I64Type) [I64Type] (fun env get_x ->
+      Stack.with_words env "dst" 4L (fun get_dst ->
```
luc-blaeser, line 5390: We could use `2L` as word size is 64-bit and the result is a `u128` cycles value. 
Same for `msg_cycles_accept128` and `msg_cycles_refunded128` (my mistake).

```suggestion
      Stack.with_words env "dst" 2L (fun get_dst ->
```

test/run-drun/basic-cycles.mo
```diff
@@ -91,6 +92,7 @@ actor a {
   public func go() : async (){
     await overflow();
     await iter();
+    assert 1000 == burn<system> 1000
```
luc-blaeser, line 95: Detail: Maybe we could test burn 0 and a very large `u128` just to check the double `u64` ordering?

ggreif: I can test 0, but the very large will clip to available...

luc-blaeser: Yes, I thought so. Just an idea, not important.

src/codegen/compile_enhanced.ml
ggreif: Nice find, please check eed383fe7c8bb30be52f15fb70ae7c536e03e53e too @luc-blaeser

test/run-drun/basic-cycles.mo
ggreif: I think we are fine

crusso (commented):
What about the interpreters? Did we just crash on other cycles prims or do we need to adjust?

src/codegen/compile_enhanced.ml
```diff
@@ -5338,7 +5338,7 @@ module Cycles = struct
 
   let balance env =
     Func.share_code0 Func.Always env "cycle_balance" [I64Type] (fun env ->
-      Stack.with_words env "dst" 4L (fun get_dst ->
```
crusso, line 5341: Good catch (here and below). I guess we were over stack allocating here ...

crusso (approved):
Looks good to me. Thanks for the quick implementation!

PR #4712, opened 2024-09-26
fix: reproduce and fix `motoko-base/issues/653`

ggreif:
Cf. fixes dfinity/motoko-base#653

It appears that `a <> Big_int.zero_big_int` blows up when
`a` is zero. Weird. Eliminating the comparison doesn't hurt
correctness, and is probably more efficient too, as negating
zero should be cheap!

@timohanke feel free to review too!


test/run/issue-base653.mo
```diff
@@ -0,0 +1,3 @@
+import { floatToInt64 } = "mo:prim";
+
+floatToInt64 (-0.5);
```
luc-blaeser, line 3: Should we maybe also test `floatToInt64(floatCopySign(0.0, -1.0)))`?

luc-blaeser (approved):
Thank you. Looks good to me. I tested it also with `-0.0` (using copySign).

test/run/issue-base653.mo
ggreif: Done: 5f9f320a0

PR #4714, opened 2024-09-27
feat: check restriction on main actor before type checking

crusso:
Improve the error message for compiled code that contains illegal top-level declarations in the main program.

c.f. https://forum.dfinity.org/t/misplaced-await-and-misplaced-async/34479


test/run-drun/bad-actor.mo
```diff
@@ -0,0 +1,12 @@
...
+type T1 = {};
+type T2 = {};
+
+actor Self {
+
+};
+
+// bad trailing decs
+type U1 = {};
+type U2 = {};
```
crusso, line 12:
```suggestion
type U2 = {};

```

test/run-drun/ok/bad-actor.comp-ref.ok
```diff
@@ -0,0 +1,4 @@
+bad-actor.mo:3.1-4.13: type error [M0141], move these declarations into the body of the main actor or actor class
+bad-actor.mo:11.1-12.13: type error [M0141], move these declarations into the body of the main actor or actor class
+bad-actor.mo:6.1-8.2: type error [M0141], an actor or actor class must be the only non-imported declaration in a program
+  (This is a limitation of the current version and flag -ref-system-api.)
```
luc-blaeser, line 4: Just a detail, not worth adjusting, maybe something for backlog: Is it really a limitation or maybe it is the clean way of how we encourage it for the future?

luc-blaeser (approved):
Very nice! Thanks for implementing.

test/run-drun/bad-actor.mo
```diff
@@ -0,0 +1,12 @@
+
+// bad leading decs
+type T1 = {};
+type T2 = {};
+
+actor Self {
```
ggreif, line 6: Perhaps move to `test/fail`? Or could we later define a semantics to consider such outside declarations to be in the actor scope? Many people tend to think this to be natural style.

ggreif (approved):
Very nice rationalisation of error messages!

test/run-drun/ok/bad-actor.comp-ref.ok
crusso: Yeah, it seems weird to allow declaration of state outside the actor - maybe we could relax in future and only allow static declarations in the prelude.

PR #4715, opened 2024-09-27
docs: Recursive types

jessiemongeon1:


doc/md/writing-motoko/local-objects-classes.md
```diff
@@ -162,7 +162,32 @@ The `class` keyword syntax shown above is a shorthand for these two definitions
...
+
+``` motoko no-repl
+import Nat "mo:base/Nat";
+
+actor {
+  class Counter(init : Nat, flag : Bool) {
+    var c = init;
+    var f = flag;
+    public func inc() : Nat {
+      if (f) {
```
ggreif, line 183: the parens are superfluous.

doc/md/writing-motoko/recursive-types.md
```diff
@@ -0,0 +1,43 @@
+---
+sidebar_position: 29
+---
+
+# Recursive types
+
+## Overview
+
+A recursive type is a type that contains the values of the same type. Recursive types enable you to create complex recursive data structures, such as linked lists or trees.
```
ggreif, line 9: Or of a type derived from it, like
``` Motoko
type Inception<T> = { #deeper : Inception<Inception<T>>; #level : Inception<T>; #shallower : T }
```
But this is currently disallowed ("is expansive") bu `moc`, so maybe my this comment is just moot :-(

doc/md/writing-motoko/recursive-types.md
```diff
@@ -0,0 +1,43 @@
...
+    case (?(x, null)) { ?x };
+    case (?(_, t)) { last<T>(t) };
+  };
+};
+```
+
+This generic function `last<T>` takes one argument `l` of type `List<T>`, which refers to the head of a list. If this function returns the last element of a list, it returns an optional value `?T`. If there isn't a last element, it will return `null`. The body of the function uses a `switch` statement to determine if the list passed as an argument is a null list, the last element of a list, or if it is the tail of a list with a next value.
+
+In this switch statement, the `last<T>` function is used recursively, since it is called within itself with `t` as the argument. The function is called again each time the case statement is satisfied, and the function receives a list head that it can switch on until the last element is returned.
+
```
ggreif, line 40: Maybe add a note that one needs recursive functions to reach/access all data in a recursive type.

doc/md/writing-motoko/recursive-types.md
```diff
@@ -0,0 +1,43 @@
...
+sidebar_position: 29
+---
+
+# Recursive types
+
+## Overview
+
+A recursive type is a type that contains the values of the same type. Recursive types enable you to create complex recursive data structures, such as linked lists or trees.
+
+Motoko supports linked lists, one data structure that implements recursive types.
```
ggreif, line 11:
```suggestion
Motoko supports linked lists, a data structure that is an example of a recursive type.
```

doc/md/writing-motoko/recursive-types.md
```diff
@@ -0,0 +1,43 @@
...
+func last<T>(l : List<T>) : ?T {
+  switch l {
+    case null { null };
+    case (?(x, null)) { ?x };
+    case (?(_, t)) { last<T>(t) };
+  };
+};
+```
+
+This generic function `last<T>` takes one argument `l` of type `List<T>`, which refers to the head of a list. If this function returns the last element of a list, it returns an optional value `?T`. If there isn't a last element, it will return `null`. The body of the function uses a `switch` statement to determine if the list passed as an argument is a null list, the last element of a list, or if it is the tail of a list with a next value.
```
ggreif, line 37: "is a null list" -> "is an empty list"

ggreif (approved):
Thanks, LGTM!

PR #4719, opened 2024-09-30
feat: allow access to `Self` and its public methods from `actor` initialisers

### LINKED ISSUE #4731: bug: `actor` self reference oddity
See:

https://forum.dfinity.org/t/what-is-special-about-actor-wrapper-and-this-motoko/36049

ggreif:
This PR accomplishes following improvements:
- allow `actor Self { ... Self ... }` in the actor initialiser (similarly for `actor class`)
- ditto for accessing all public methods `Self.method` from the actor initialiser
- accessing these from under lambdas without making the respective methods undefined when using them

The latter should allow setting up timers: `addTimer(b, period, Self.method)` from actor initialiser.

Also fixes the hole in definedness analysis (#4731).

-------------
_NOTES_:
- calling `Self.method` from actor initialiser is prohibited by the capability checker
- accessing `Self.method` from actor initialiser is currently allowed when `method` is not defined yet, and results in a descriptor (pair)
- comparison (in terms of `shared`) of the internal methods with external methods is trapping in the interpreter, but should work in the deployed actor.

There is an idea inside to fix these problems.

----
TODO:
- adapt `renaming.ml`? — nope, I don't have business in IR
- [x] test `actor class` too
- [x] test clashing `Self` — there is already `self-shadow.mo`
- [x] `test/run/issue1464ok.mo` fails — fixed snafu
- [x] deal with `FIXME`s
- [x] `shared` method equality in the interpreters
- [x] `shared` call viability checking (in interpreters)
- [x] fix #4731
- [x] write issue about the mixed (pair/closure) `shared func` comparison problem — here: #4732
- [x] resolve the `Promise.is_fulfilled` problem — #4735
- [x] check the "For actors, this may be too permissive" comment in the code — irrelevant, removed!
- [ ] try out the "crazy idea" mentioned in the feedback


src/mo_interpreter/interpret.ml
```diff
@@ -785,7 +785,7 @@ and declare_pat_fields pfs ve : val_env =
 and define_id env id v =
   define_id' env id.it v
 
-and define_id' env id v =
+and define_id' env id v = if not(Lib.Promise.is_fulfilled (find id env.vals)) then
```
ggreif, line 788: This must be solved differently

src/mo_interpreter/interpret.ml
```diff
@@ -909,14 +915,20 @@ and declare_dec_fields dec_fields ve_ex ve_in : val_env * val_env =
     let ve_in' = V.Env.adjoin ve_in ve' in
     declare_dec_fields dec_fields' ve_ex' ve_in'
 
-and interpret_dec_fields env dec_fields ve (k : V.value V.cont) =
+and interpret_dec_fields' env upd dec_fields ve (k : V.value V.cont) =
```
ggreif, line 918: use `?(incremental=fun () -> ())`

src/mo_frontend/definedness.ml
```diff
@@ -206,10 +206,10 @@ and group msgs (grp : group) : f =
            defined after j *)
         if j < i
         then () (* all izz well *)
-        else
+        else (*assert eager;*)
           Diag.add_msg
             msgs
-            (Diag.error_message
+            (Diag.warning_message
```
ggreif, line 212: This needs to be solved!

src/mo_interpreter/interpret.ml
ggreif: Done.

src/mo_frontend/definedness.ml
```diff
@@ -101,9 +101,9 @@ let rec exp msgs e : f = match e.it with
   | FromCandidE e       -> exp msgs e
   | TupE es             -> exps msgs es
   | ProjE (e, i)        -> exp msgs e
-  | ObjBlockE (s, _, dfs) ->
+  | ObjBlockE (s, (self_id_opt, _), dfs) ->
     (* For actors, this may be too permissive; to be revised when we work on actors again *)
```
ggreif, line 105: Is this comment still relevant? What would be too permissive for `actor`?

test/fail/access-self-prematurely.mo
```diff
@@ -0,0 +1,6 @@
+actor /*class () =*/ Self {
+    func caller(_callee : shared () -> async ()) { };
+    caller(Self.method1); // FIXME: should be error too
```
ggreif, line 3: The AST interpreter crashes her with
> execution error, Actor "ID:1" has no method "method1"
This could be resolved by allowing `(<actor>, <method>)` as an `async`-callable in the interpreters and make `DotE` produce those when projected out of an `actor`. (Idea by @crusso).

src/ir_interpreter/interpret_ir.ml
```diff
@@ -800,10 +813,10 @@ and interpret_dec env dec k =
       k ()
     )
 
-and interpret_decs env decs (k : unit V.cont) =
+and interpret_decs env ?(increments=ignore) decs (k : unit V.cont) =
   match decs with
```
ggreif, line 817: call `increments` now?

test/fail/access-self-prematurely.mo
crusso: I think that would also eliminate the need for the `increment` function...

ggreif: I was thinking of stashing the pairs into the `env.actors_env[actor_id]` where the `is_fullfilled` would give back `false`, but you may well be right. That would mean complicating the `DotE` evaluation a bit, though.

test/fail/access-self-prematurely.mo
```diff
@@ -0,0 +1,5 @@
+actor /*class () =*/ Self {
```
ggreif, line 1: this could be deleted, pointless

src/ir_interpreter/interpret_ir.ml
ggreif: remove, rather!

src/mo_interpreter/interpret.ml
```diff
@@ -508,10 +508,11 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
...
       | V.Obj fs ->
         k (find id.it fs)
-      | V.Blob aid when T.sub exp1.note.note_typ (T.Obj (T.Actor, [])) ->
+      | V.Blob aid as actor when T.sub exp1.note.note_typ (T.Obj (T.Actor, [])) ->
         begin match V.Env.find_opt aid !(env.actor_env) with
-        (* not quite correct: On the platform, you can invoke and get a reject *)
-        | None -> trap exp.at "Unknown actor \"%s\"" aid
+        | None ->
+          (* method not defined yet, just pair them up *)
+          k V.(Tup [actor; Text id.it])
```
crusso, line 515: I would simplify the outer case on DotE into two cases, one on actor types, and one on other types.
The first just pairs the actor blob with the field name, the second just reads the field from the actor.
```
| DotE (exp1, id)  when T.sub exp1.note.note_typ (T.Obj (T.Actor, []))  ->
     interpret_exp env exp1 (fun v1 ->
     k (Tup [v1; Text id.it])
| DotE (exp1, id) ->
    interpret_exp env exp1 (fun v1 ->
      let fs = V.as_obj v1 in in
        k (find id.it fs)
```
Hopefully you can then just special case the application on actor fields accordingly (issuing rejects on undefined actor or reference)

crusso (commented):
Interim comments

src/ir_interpreter/interpret_ir.ml
```diff
@@ -307,10 +307,18 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
     )
   | LitE lit ->
     k (interpret_lit env lit)
+  | PrimE (ActorDotPrim n, [{ it = VarE (_, actor); _ }]) when not(Lib.Promise.is_fulfilled (find actor env.vals)) ->
+    (* actor not defined yet, just pair them up *)
+    k V.(Tup [Blob (env.self); Text n])
```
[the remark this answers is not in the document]
ggreif: I am pretty sure I need it (not that I like it!), but I'll double-check.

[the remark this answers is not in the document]
ggreif: using `env.self` by guessing is a bit sick, I agree...

src/ir_interpreter/interpret_ir.ml
```diff
@@ -337,8 +345,9 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
       | ActorDotPrim n, [v1] ->
         let id = V.as_blob v1 in
         begin match V.Env.find_opt id !(env.actor_env) with
-        (* not quite correct: On the platform, you can invoke and get a reject *)
-        | None -> trap exp.at "Unknown actor \"%s\"" id
+        | None ->
```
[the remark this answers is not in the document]
ggreif: Sure we can do that, I was just being cautious here.

src/mo_frontend/definedness.ml
```diff
@@ -84,55 +84,54 @@ let rec exp msgs e : f = match e.it with
   (* Or anything that is occurring in a call (as this may call a closure): *)
   | CallE (e1, ts, e2)  -> eagerify (exps msgs [e1; e2])
   (* And break, return, throw can be thought of as calling a continuation: *)
-  | BreakE (i, e)       -> eagerify (exp msgs e)
```
ggreif, line 87: no changed semantics in this function, just cleanups

src/mo_frontend/definedness.ml
```diff
@@ -192,9 +191,13 @@ and decs msgs decs : group =
     (d.at, defs, eager_vars f, delayed_vars f)
   ) decs
 
-and group msgs (grp : group) : f =
+and group msgs ?(extern=None) (grp : group) : f =
```
ggreif, line 194: this is the meat

src/ir_interpreter/interpret_ir.ml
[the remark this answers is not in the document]
ggreif: Done: 76d3d1a64

src/mo_interpreter/interpret.ml
ggreif: 54bb508e4

ggreif:
@infu Please note the trick in https://forum.dfinity.org/t/what-is-special-about-actor-wrapper-and-this-motoko/36049. It exploits a small hole in `moc`'s definedness checker. We'll only plug that hole _after_ this PR has been merged and give you a heads-up when you can use both methods. @crusso I hope I didn't promise too much :-)

src/mo_interpreter/interpret.ml
```diff
@@ -978,7 +988,7 @@ and interpret_dec env dec (k : V.value V.cont) =
               rets = Some k'';
               throws = Some r }
             in
-            interpret_obj env''' obj_sort.it dec_fields (fun v' ->
+            interpret_obj env''' obj_sort.it (Some id') dec_fields (fun v' ->
```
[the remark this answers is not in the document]
ggreif: Hmm, working around it in 58a5aae57.

src/mo_interpreter/interpret.ml
```diff
@@ -885,16 +886,31 @@ and match_shared_pat env shared_pat c =
...
   match obj_sort with
   | T.Actor ->
      let self = V.fresh_id() in
+     let self' = V.Blob self in
      let ve_ex, ve_in = declare_dec_fields dec_fields V.Env.empty V.Env.empty in
-     let env' = adjoin_vals { env with self = self } ve_in in
+     let env' = adjoin_vals { env with self } ve_in in
+     (* Define self_id eagerly *)
+     Option.iter (fun id -> define_id env' id self') self_id;
+     (* Define self_id in inner scope if there is a non-shadowing way to do it *)
```
ggreif, line 898: this can go

src/mo_interpreter/interpret.ml
```diff
@@ -786,7 +783,10 @@ and define_id env id v =
   define_id' env id.it v
 
 and define_id' env id v =
-  Lib.Promise.fulfill (find id env.vals) v
+  let b = find id env.vals in
+  if Lib.Promise.(is_fulfilled b && value b = v)
+  then ()
+  else Lib.Promise.fulfill (find id env.vals) v
```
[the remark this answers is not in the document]
crusso: Yeah, I considered something like that too. I guess the other option is to just rebind this (in the actor body) to a defined promise holding the self blob, and rely on the existing code to define the outer one, avoing the duplicate definition.

src/mo_interpreter/interpret.ml
```diff
@@ -885,16 +894,31 @@ and match_shared_pat env shared_pat c =
...
-and interpret_obj env obj_sort dec_fields (k : V.value V.cont) =
+and interpret_obj env obj_sort self_id dec_fields (k : V.value V.cont) =
   match obj_sort with
   | T.Actor ->
      let self = V.fresh_id() in
+     let self' = V.Blob self in
+     Option.iter (fun id -> define_id env id self') self_id;
      let ve_ex, ve_in = declare_dec_fields dec_fields V.Env.empty V.Env.empty in
-     let env' = adjoin_vals { env with self = self } ve_in in
+     let env' = adjoin_vals { env with self } ve_in in
```
crusso, line 904: I think if you declare_id self (if necessary before adding ve_in to env you might be ok.

src/mo_frontend/definedness.ml
```diff
@@ -84,55 +84,54 @@ let rec exp msgs e : f = match e.it with
...
+  | RetE e
   | ThrowE e            -> eagerify (exp msgs e)
   (* Uses are delayed by function expressions *)
   | FuncE (_, sp, tp, p, t, _, e) ->
-      delayify ((exp msgs e /// pat msgs p) /// shared_pat msgs sp)
+    delayify ((exp msgs e /// pat msgs p) /// shared_pat msgs sp)
+  | ObjBlockE (s, (self_id_opt, _), dfs) ->
+    group msgs ~extern:self_id_opt (dec_fields msgs dfs)
   (* The rest remaining cases just collect the uses of subexpressions: *)
   | LitE _ | ActorUrlE _
```
crusso, line 96: Fix this

test/run-drun/pass-class-self.mo
```diff
@@ -0,0 +1,13 @@
...
+
+actor class C() = Self {
+    public func method() : async () { debugPrint "YESS!"; ignore principalOfActor Self };
+    func caller(_callee : shared () -> async ()) { };
+
+    debugPrint "Before!";
+    ignore principalOfActor Self;
+    caller(Self.method);
+    caller(method);
+    //debugPrint (debug_show(principalOfActor Self));
```
ggreif, line 11: Bizarre: uncommenting this gives a forward declaration inference error... but we already use `debugPrint` above!

src/ir_interpreter/interpret_ir.ml
```diff
@@ -446,13 +455,21 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
         let reject = Option.get env.rejects in
         let e = V.Tup [V.Variant ("canister_reject", V.unit); v1] in
         Scheduler.queue (fun () -> reject e)
-      | ICCallPrim, [v1; v2; kv; rv; cv] ->
+      | ICCallPrim, V.[Tup [Blob aid; Text id]; v2; kv; rv; cv] ->
+        let v1 = V.Env.(find aid !(env.actor_env) |> V.as_obj |> find id) in
```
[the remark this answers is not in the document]
ggreif: Done!

src/ir_interpreter/interpret_ir.ml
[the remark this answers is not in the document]
ggreif: when doing `when false && not...` I get
```
incomplete-self-reference.mo:4.42-4.46: execution error, accessing identifier before its definition
```
in both `run-ir` and `run-low`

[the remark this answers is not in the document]
ggreif: I don't understand yet, `.run-ir` traces as
```
+Defining $getState/0
+Want to access Self
+Defining stored
+Defining method
+Defining go
+Defining Self
```
but here is the IR-dump:
```

```

src/mo_interpreter/interpret.ml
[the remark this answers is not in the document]
ggreif: Workaround not needed with `env'`-introduction.

src/mo_interpreter/interpret.ml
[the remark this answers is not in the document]
ggreif: It is gone.

src/mo_interpreter/interpret.ml
ggreif: Yeah!

src/mo_frontend/definedness.ml
ggreif: Done: 6abf6185a

test/run-drun/pass-class-self.mo
ggreif: @crusso any idea how to explain this?

src/ir_interpreter/interpret_ir.ml
```diff
@@ -446,6 +453,14 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
         let reject = Option.get env.rejects in
         let e = V.Tup [V.Variant ("canister_reject", V.unit); v1] in
         Scheduler.queue (fun () -> reject e)
+      | ICCallPrim, V.[Tup [Blob aid; Text id]; v2; kv; rv; cv] ->
```
crusso, line 456: Can you merge this into to the next case as you did on line 326 (for the before cps version). Most
of the code except function resolution is the same.

src/mo_frontend/definedness.ml
```diff
@@ -178,12 +178,15 @@ and dec msgs d = match d.it with
   | VarD (i, e) -> (M.empty, S.singleton i.it) +++ exp msgs e
   | TypD (i, tp, t) -> (M.empty, S.empty)
   | ClassD (csp, i, tp, p, t, s, i', dfs) ->
+    let extern = if s.it = Type.Actor then Some i' else None in
     (M.empty, S.singleton i.it) +++ delayify (
-      group msgs (dec_fields msgs dfs @ class_self d.at i') /// pat msgs p /// shared_pat msgs csp
+      group msgs ~extern (dec_fields msgs dfs @ class_self d.at i' s.it) /// pat msgs p /// shared_pat msgs csp
```
crusso, line 183: Instead of adding the ~extern optional parameter, why not just add (class_self ... ) to beginning or end of the decs argument, depending on whether this is an actor (at front) or at other object (at end).

crusso: (and then rename class_self to obj_self...)

src/mo_frontend/definedness.ml
```diff
@@ -192,9 +195,13 @@ and decs msgs decs : group =
...
   ) decs
 
-and group msgs (grp : group) : f =
+and group msgs ?(extern=None) (grp : group) : f =
   (* Create a map from declared variable to their definition point *)
   let defWhen = M.disjoint_unions (List.mapi (fun i (_, defs, _, _) -> map_of_set i defs) grp) in
+  (* Insert the externally defined binding in front if present, non-shadowing *)
+  let defWhen = match extern with
+    | Some b when M.find_opt b.it defWhen |> Option.is_none -> M.add b.it (-1) defWhen
+    | _ -> defWhen in
```
crusso, line 204: You might not need this if you just extend the decs appropriately,

src/mo_interpreter/interpret.ml
```diff
@@ -572,7 +573,11 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
     in k v'
   | CallE (exp1, typs, exp2) ->
     interpret_exp env exp1 (fun v1 ->
-      interpret_exp env exp2 (fun v2 ->
+       let v1 = begin match v1 with
+         | V.(Tup [Blob aid; Text id]) -> lookup_actor env exp1.at aid id
+         | _ -> v1
+        end in
+     interpret_exp env exp2 (fun v2 ->
```
crusso, line 580:
```suggestion
      interpret_exp env exp2 (fun v2 ->
```

test/run-drun/pass-class-self.mo
crusso: Is this resolved?

src/mo_interpreter/interpret.ml
```diff
@@ -885,16 +891,22 @@ and match_shared_pat env shared_pat c =
...
+     | Some id -> adjoin_vals env (declare_id id)
+     | None -> env in
+     Option.iter (fun id -> define_id env' id self') self_id;
      let ve_ex, ve_in = declare_dec_fields dec_fields V.Env.empty V.Env.empty in
-     let env' = adjoin_vals { env with self = self } ve_in in
-     interpret_dec_fields env' dec_fields ve_ex
+     let env'' = adjoin_vals { env' with self } ve_in in
+     interpret_dec_fields env'' dec_fields ve_ex
      (fun obj ->
         (env.actor_env := V.Env.add self obj !(env.actor_env);
```
crusso, line 908: Crazy idea: At the moment, the actor environment maps actor ids to Objects containing a map from field names to values.
Suppose we changed that, so an actor environment just contains a map from field names to promises (i.e. `ve_ex`).
Then we can update the actor map before we interpret the body. As the promises in `ve_ex` get defined, so do the ones accessible from the self-reference, a bit like your previous 'increments' but automatic.
I think that might let us implement the equality on mixtures of  symbolic and function references (provided we also give the equality operator) access to the environment. 

Best attempted in a separate PR I think. Might be a dead-end with too many pervasive changes.

src/mo_values/operator.ml
```diff
@@ -185,7 +185,7 @@ let structural_equality t =
           let v1 = as_array v1 in
           let v2 = as_array v2 in
           Bool (
-            Array.length v1 == Array.length v2 &&
+            Array.length v1 = Array.length v2 &&
```
crusso, line 188: Good catch!

src/mo_values/operator.ml
```diff
@@ -235,8 +235,13 @@ let structural_equality t =
           else
             go (List.find (fun f -> f.T.lab = l1) fs).T.typ v1 v2
     | T.Func (s, c, tbs, ts1, ts2) ->
-        assert (T.is_shared_sort s);
-        fun v1 v2 -> Bool (v1 == v2)  (* HACK *)
+       assert (T.is_shared_sort s);
+       fun v1 v2 -> match v1, v2 with
+       | Tup [Blob _; Text _], Tup [Blob _; Text _] -> Bool (v1 = v2) (* public methods *)
+       | Func _, Tup [Blob _; Text _]
+       | Tup [Blob _; Text _], Func _ -> assert false; (* mixed, cannot determine equality *)
```
crusso, line 242: If equality had access to the actor env then we could probably do better here and maybe even get it right (see above)

crusso (changes requested):

test/run-drun/pass-class-self.mo
crusso: I just played around with it. I expect it's because the call to debugPrint is the last declaration that is caussing the issue...

This also produces the error:
```
import { debugPrint; principalOfActor } = "mo:⛔";

actor class C() = Self {
    debugPrint "So far so good!" 
}
```

crusso: Filed an issue 
https://github.com/dfinity/motoko/issues/4733

src/mo_values/operator.ml
ggreif: Yeah, this one was a bit ugly.

src/ir_interpreter/interpret_ir.ml
ggreif: Yeah, this was an oversight on my side.

test/run-drun/pass-class-self.mo
ggreif: I suspected that it is unrelated. Thanks for checking!

src/ir_interpreter/interpret_ir.ml
ggreif: Done in de19de574.

src/mo_frontend/definedness.ml
ggreif: Well, that would break
```
$ cat test/run-drun/self-shadow.mo
actor foo {
  public func foo() {};

  flexible func go() : async () {
    let bar = actor bar { public func bar() {} }
  };
}
```
right? See also the discussion in #1104.

crusso: Ok, I had a go - you are right about the shadowing but that's possible to account for #4736

src/mo_interpreter/interpret.ml
ggreif: On first look this might work, the callback could simply define the method as `interpret_obj` is iterating over the decs.

src/mo_values/operator.ml
ggreif: Yes, but I wouldn't hold my breath.

src/mo_frontend/definedness.ml
ggreif: Merged, thanks!

src/mo_frontend/definedness.ml
ggreif: #4735 is merged, thanks!

src/mo_frontend/definedness.ml
ggreif: @crusso do you remember why this comment is here?

test/run-drun/pass-self.mo
```diff
@@ -0,0 +1,13 @@
...
+actor Self {
+    public func method() : async () { debugPrint "YESS!"; ignore principalOfActor Self };
+    func caller(_callee : shared () -> async ()) { };
+
+    debugPrint "Before!";
+    ignore principalOfActor Self;
+    caller(Self.method);
+    caller(method);
+    debugPrint (debug_show(principalOfActor Self));
+    debugPrint "So far so good!";
```
crusso, line 12: Strange, this one doesn't generate a a type error about trailing 
debugPrint, but pass-class-self does?

crusso (approved):
LGTM. Can you open an issue to improve function equality in the interpreters in future?

test/run-drun/pass-self.mo
ggreif: Yeah, very bizarre.

crusso:
Users should appreciate this...

ggreif:
> LGTM. Can you open an issue to improve function equality in the interpreters in future?

Yeah, already did: #4732.

PR #4720, opened 2024-09-30
feat: suggest corrections for unbound identifiers/fields based on distance metric 

crusso:
Add flag --ai-errors to tailor error reporting for human (default) or artificial users.

For humans, instead of reporting that a value or type identifier is not available in references or projections, we now suggest  likely candidates based on:

* whether the identifier is a prefix of the candidate identifiers.
* whether the identifier is some [edit distance](https://en.wikipedia.org/wiki/Levenshtein_distance) away from the candidate identifiers. 

and report identifiers in order of edit distance.

When the object/actor/module  projected from has few bindings (<16), we display them with their types, to avoid producing
huge error messages.

With --ai-errors, we don't give hints but display all available bindings with their types

(Thanks to Claude.ai for coughing up an implementation.)


test/fail/suggest-local-type.mo
```diff
@@ -0,0 +1,2 @@
+type Foo = Int;
+type Zap = [Fo];
```
crusso, line 2:
```suggestion
type Zap = [Fo];

```

test/fail/suggest-local.mo
```diff
@@ -0,0 +1,5 @@
+let xxx = 1;
+x;
+
+
+
```
crusso, line 5:
```suggestion
x;

```

test/fail/suggest-local-type.mo
crusso, line 2:
```suggestion
type Zap = [Fo];
```

test/fail/suggest-long.mo
```diff
@@ -0,0 +1,4 @@
+import Prim "mo:⛔";
+
+Prim.stableM;
+
```
crusso, line 4:
```suggestion
```

test/fail/suggest-short.mo
```diff
@@ -0,0 +1,4 @@
+import Prim "mo:⛔";
+
+Prim.s;
+
```
crusso, line 4:
```suggestion
Prim.s;
```

test/fail/suggest-type.mo
```diff
@@ -0,0 +1,4 @@
+import Prim "mo:⛔";
+
+type T = Prim.Types.Num;
+
```
crusso, line 4:
```suggestion
type T = Prim.Types.Num;
```

test/fail/suggest-short.mo
crusso, line 4:
```suggestion
Prim.s;
```

crusso, line 4:
```suggestion
```

test/fail/ok/suggest-label.tc.ok
```diff
@@ -0,0 +1,2 @@
+suggest-label.mo:2.9-2.11: type error [M0083], unbound label fo
+Did you mean foo?
```
chenyan-dfinity, line 2: If the suggestion was wrong, does it make the error message more confusing? Also, AI tends to listen to the suggestions. If it's wrong, we may lead it to a wrong direction.

test/fail/ok/suggest-long.tc.ok
```diff
@@ -0,0 +1,259 @@
+suggest-long.mo:3.1-3.5: type error [M0072], field stableM does not exist in type 
+  module {
```
chenyan-dfinity, line 2: Would be good to have a way to shorten this error message. In practice, nobody really looks at it, because it's too long.

chenyan-dfinity (commented):
If the suggestion is wrong, does it make the error message more confusing?

test/fail/ok/suggest-long.tc.ok
rvanasa: The long error message might be useful for LLMs...maybe instead of showing this by default, we could gate this behind a compiler flag intended for giving additional context for AI code completion? We could also include the full [error message explanations](https://github.com/dfinity/motoko/tree/master/src/lang_utils/error_codes) and other context which could help with automatically fixing errors.

test/fail/ok/suggest-label.tc.ok
rvanasa: Maybe we could show this by default and then include a compiler to flag for AI code generation (see [this comment](https://github.com/dfinity/motoko/pull/4720/files#r1783587695)). In general, this would make it so we can fine-tune what the AI sees without sacrificing readability for humans.

test/fail/ok/suggest-long.tc.ok
chenyan-dfinity: Yes, we have `--print-source-on-error` and `--error-detail` flags already. Maybe `print-source` flags can be absorbed by `error-detail`, and `error-detail` can also control whether we list the module bindings.

Another idea is to have two error messages, one for human, one for AI...

test/fail/ok/suggest-label.tc.ok
chenyan-dfinity: It's a bit strange for a strongly typed language to suggest an identifier based on distance metric instead of types...If there is an AI mode for error message, maybe just listing the current in scope env is enough for AI to make a decision?

crusso: Ocaml does the distance metric thing. That's where I stole the idea from.

I expect AI won't generate any mistyped identifiers, but I was hoping this would catch the error of using `Text.toLowercase`  instead of `Text.toLower`. (c.f. https://docs.google.com/spreadsheets/d/1osfLjjlqr43FtVzUkYrS_oBk5699jfMXB_XRZZ52sqM/edit?usp=sharing)

src/mo_frontend/typing.ml
```diff
@@ -114,6 +114,27 @@ let kind_of_field_pattern pf = match pf.it with
...
+    let d = distance id0 in
+    if Lib.String.starts_with id id0 || d <= limit then
+      Some (d, id0)
+    else None) ids in
+  let suggestions = List.sort compare weighted_ids |> List.map snd in
+  if suggestions = [] then ""
+  else
+    let ids, id = Lib.List.split_last suggestions in
+    "\nDid you mean " ^
+      (if ids <> [] then (String.concat ", " ids) ^ " or " else "") ^
```
luc-blaeser, line 135: Just a detail: Maybe a unit test for multiple suggestion matches would be interesting.

luc-blaeser (approved):
Very nice PR! I like the editing distance heuristics.

test/fail/ok/suggest-label.tc.ok
crusso: I looked at filtering the suggestion to those that match on type, but the type checker doesn't have the expected type available when checking identifiers or projections - these are always done in inference mode. 

Moreover, in the common cases, applications, the type of the function is actually inferred.
For resolving function, knowing the type of the expected result would help a little,
but we don't have that info readily available.

test/fail/ok/suggest-long.tc.ok
crusso: The error-detail flag is just for syntax errors. I've added a -ai-errors flag for now and used it to disable the suggestions.

test/fail/ok/suggest-label.tc.ok
chenyan-dfinity: That spreadsheet shows the errors the AI encountered, but it doesn't show whether it can fix itself after feeding back the error messages. I wonder if the current error message (displaying all the functions in the module) is enough for AI to fix it.

I'm fine that we explore this idea, but not sure if a wrong suggestion can lead AI to a wrong path. A few corner cases I can think of:

* The code uses `fold`, do we suggest `foldLeft` or `foldRight`
* The code uses `Array.exists` or `Array.forall`, which doesn't exist in the module. Then any suggestion seems to be wrong?

ggreif: Maybe later we could (also) generate a set of identifiers that are in scope and then filter them so that they fit into the context. But for now I think this is just fine.

crusso:
@chenyan-dfinity PTAL

chenyan-dfinity (approved):
LGTM. We also need to remember to turn on the ai flag for caffeine once it's released.

test/fail/ok/suggest-local-type-ai.tc.ok
```diff
@@ -0,0 +1,8 @@
+suggest-local-type-ai.mo:3.13-3.15: type error [M0029], unbound type Fo in type environment:
+  {
+    type Foo = Int;
+    type UpgradeOptions = {wasm_memory_persistence : ?WasmMemoryPersistence};
+    type WasmMemoryPersistence = {#Keep; #Replace};
```
crusso, line 5: Hmm, these should not be leaking out. Will prefix with @ like other internals.

crusso: Done https://github.com/dfinity/motoko/pull/4720/commits/6b753868efcd92f265c1129d34dcd8db7cefc5c1

src/mo_frontend/typing.ml
crusso: I think there are some (Prim.stableM ) produces several..

PR #4735, opened 2024-10-17
experiment: simpler version of claudio/self-pass

crusso:
There's no need to add `self_id` to `ActorE` after all, thus avoiding much complexity.

The internal self declaration, if present, is enough.

The real culprit was some bad lowering that wasn't passing down the self-binding in the 
`ObjBlockE` case (now fixed).

### DIFF
--- a/src/lowering/desugar.ml
+++ b/src/lowering/desugar.ml
@@ -91,7 +91,7 @@ and exp' at note = function
       (* case ? v : *)
       (varP v) (varE v) ty).it
   | S.ObjBlockE (s, (self_id_opt,_), dfs) ->
-     obj_block at s self_id_opt dfs note.Note.typ
+    obj_block at s self_id_opt dfs note.Note.typ
   | S.ObjE (bs, efs) ->
     obj note.Note.typ efs bs
   | S.TagE (c, e) -> (tagE c.it (exp e)).it
--- a/src/lowering/desugar.ml
+++ b/src/lowering/desugar.ml
@@ -90,8 +90,8 @@ and exp' at note = function
       (breakE "!" (nullE()))
       (* case ? v : *)
       (varP v) (varE v) ty).it
-  | S.ObjBlockE (s, _t, dfs) ->
-    obj_block at s None dfs note.Note.typ
+  | S.ObjBlockE (s, (self_id_opt,_), dfs) ->
+     obj_block at s self_id_opt dfs note.Note.typ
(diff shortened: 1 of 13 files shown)


src/lowering/desugar.ml
```diff
@@ -90,8 +90,8 @@ and exp' at note = function
       (breakE "!" (nullE()))
       (* case ? v : *)
       (varP v) (varE v) ty).it
-  | S.ObjBlockE (s, _t, dfs) ->
-    obj_block at s None dfs note.Note.typ
+  | S.ObjBlockE (s, (self_id_opt,_), dfs) ->
```
crusso, line 93: This was the bug, which was failing to introduce a the self declaration for some cases.

test/run-drun/ok/pass-class-self.diff-low.ok
```diff
@@ -1,4 +1,5 @@
 --- pass-class-self.run
 +++ pass-class-self.run-low
-@@ -0,0 +1 @@
+@@ -0,0 +1,2 @@
 +Before!
++ys6dh-5cjiq-5dc
```
ggreif, line 5: Why this diff? Didn't `run.ok` also change, did it?

crusso: Good question...

crusso: Hmm, why is the having any side-effect at all? The test is an actor class declaration and should do nothing until instantiated.

ggreif: Yeah, it is probably doing what `drun` does, auto-deploying.

crusso: Run doesn't instantiate the class, but run-ir etc do.

I tried this on master (removing the self references) and it seems to be existing behaviour. Somewhat odd though, but perhaps just an artifact of "compiled" code.

crusso: Ah, the ir_interpreter decides to execute the body of an actor class with zero arguments (to match semantics of installation on ic).

Nvm.

ggreif: It helps to see the side effects!

ggreif (approved):
LGTM! Nice improvement.

PR #4743, opened 2024-10-24
Make it clearer that pre/postupgrade hooks are discouraged

jessiemongeon1:


doc/md/reference/language-manual.md
```diff
@@ -1348,6 +1348,10 @@ The declaration `<dec>` of a `system` field must be a manifest `func` declaratio
 
 -   `postupgrade`: When declared, is called during an upgrade, immediately after the replacement actor body has initialized its fields, inheriting values of the retired actors' stable variables, and before its first message is processed. Its `<system>` type parameter is implicitly assumed and need not be declared.
 
+:::danger
+Using the pre- and post-upgrade system methods is discouraged. It is error-prone and can render a canister unusable. Per best practices, using these methods should be avoided if possible.
```
crusso, line 1352:
```suggestion
Using the pre- and post-upgrade system methods is discouraged. It is error-prone and can render a canister unusable. In particular, if a `preupgrade` method traps and cannot be prevented from trapping by other means, then your canister may be left in a state in which it can no longer be upgraded. 
Per best practices, using these methods should be avoided if possible.
```

crusso: I leave it to you whether to add that sentence here, and perhaps above too.

doc/md/canister-maintenance/upgrades.md
```diff
@@ -207,9 +205,11 @@ For more information, see the [example of explicit migration](compatibility.md#e
 
 The following aspects are retained for historical reasons and backwards compatibility:
 
-### Preupgrade and postupgrade system methods
+### Pre-upgrade and post-upgrade system methods
 
-This is an advanced functionality that is not recommended for standard cases, as it is error-prone and can render the canister unusable.
+:::danger
+Using the pre- and post-upgrade system methods is discouraged. It is error-prone and can render a canister unusable. Per best practices, using these methods should be avoided if possible.
```
crusso, line 211:
```suggestion
Using the pre- and post-upgrade system methods is discouraged. It is error-prone and can render a canister unusable. In particular, if a `preupgrade` method traps and cannot be prevented from trapping by other means, then your canister may be left in a state in which it can no longer be upgraded.  Per best practices, using these methods should be avoided if possible.
```

crusso (approved):
Approved. I'll leave it to you to decide whether to explain why using my suggestions (or similar).

PR #4756, opened 2024-11-07
chore: update `motoko-mode.el` to v29 apostrophe conventions

ggreif:
this removes `emacs` warnings when loading the mode

see: https://emacs.stackexchange.com/questions/73047/emacs-29-docstring-single-quote-escaping-rules-compiler-level-event

not sure whether elisp really qualify as docstrings, but this helps


ggreif:
Nope, the warnings are coming from other packages.

PR #4765, opened 2024-11-14
add back gc flags for moc.js

chenyan-dfinity:


crusso (approved):
I *think* this ok, but as Luc said, eop only supports one gc (incremental) and I'm not sure what the compiler will do if you specify copying and eop (perhaps it just ignores the gc setting?).

luc-blaeser (approved):
Thanks for adding the options!

PR #4766, opened 2024-11-15
experiment: add optional stability modifiers before body to control default

crusso:
Investigate with prefixing body of actor or actor class with optional, default stability modifier, supporting the opt-in inversion from flexible to stable, without changing the semantics of existing code.

(Prefixing the `actor` keyword introduces hundreds of shift/reduce conflicts, so I tried this instead.
Perhaps the conflicts can be avoid by some grammar refactoring...)

Note: we only allow the stable modifier on simple let patterns bind a single identier and (all) mutable variables, so care must be taken not to default to stable in other cases...

```motoko
actor Counter stable {

  var count : Nat = 0; // defaults to stable

  public func inc() : async Nat {
    count += 1;
    count
  };

  // let f = func(){}; // rejected as unstable

}
```
and 

```motoko
actor class Counter() stable {

  var count : Nat = 0;  // defaults to stable

  public func inc() : async Nat {
    count += 1;
    Prim.debugPrint (debug_show(count));
    count
  };

  // let f = func(){}; // rejected as unstable

}
```

Not sure I'm loving it, tbh.

### DIFF
--- a/doc/md/examples/grammar.txt
+++ b/doc/md/examples/grammar.txt
@@ -305,7 +305,7 @@
     'type' <id> ('<' <list(<typ_bind>, ',')> '>')? '=' <typ>
     <obj_sort> <id>? (':' <typ>)? '='? <stab> <obj_body>
     <shared_pat_opt> 'func' <id>? <typ_params_opt> <pat_plain> (':' <typ>)? <func_body>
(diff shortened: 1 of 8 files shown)


ggreif:
Wanna have some SQL feeling? `actor Counter stable` ➡ `actor Counter with stable var` 😜

crusso:
closing in favour of #4779

PR #4777, opened 2024-11-22
perf: use less cycles in `@timer_helper`

ggreif:
Since the observation of a `null` in the expiration array means there will be no more found, it is pointless to stay in the loop. I expect 1 expiration per callback in the frequent case.

### DIFF
--- a/src/prelude/internals.mo
+++ b/src/prelude/internals.mo
@@ -612,10 +612,10 @@ func @timer_helper() : async () {
   ignore (prim "global_timer_set" : Nat64 -> Nat64) exp;
   if (exp == 0) @timers := null;
 
-  for (o in thunks.vals()) {
+  label out for (o in thunks.vals()) {
     switch o {
-      case (?thunk) { ignore thunk() };
-      case _ { }
+      case (?thunk) ignore thunk();
+      case _ break out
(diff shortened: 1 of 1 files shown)


src/prelude/internals.mo
```diff
@@ -612,10 +612,10 @@ func @timer_helper() : async () {
   ignore (prim "global_timer_set" : Nat64 -> Nat64) exp;
   if (exp == 0) @timers := null;
 
-  for (o in thunks.vals()) {
+  label out for (o in thunks.vals()) {
     switch o {
-      case (?thunk) { ignore thunk() };
-      case _ { }
+      case (?thunk) ignore thunk();
+      case _ break out
```
crusso, line 618: I think you could also just return  without the label and break

crusso (approved):

src/prelude/internals.mo
crusso: I think the real overhead is the messaging cost (regardless of cycles)

ggreif: Right, but in the grand scheme of things this helps a bit. Why burn unnecessarily?

rvanasa (approved):

PR #4779, opened 2024-11-25
experiment: allow stable/flexible modifiers before top-level actor(class) declarations to flip defaults.

crusso:
The only way to fit this into the grammar is for top-level actor declarations (class libraries and programs), but that's all we support when compiling to the IC anyway...

Benefits:
* backward compatible
* opt-in
* adds persistent/transient as synonyms for stable/flexbile.

closing due to preference for simpler #4786


doc/md/examples/grammar.txt
```diff
@@ -268,6 +271,10 @@
     'flexible'
     'stable'
 
+<stab_mod> ::= 
+    'flexible'
```
crusso, line 275: If we don't like the option of

flexible actor (class...) 

we could just rule this production out...

crusso:
superseded by #4779

PR #4782, opened 2024-11-28
WIP: comma parentheticals

ggreif:
Prompted by https://github.com/dfinity/motoko/pull/4608#discussion_r1819459961


ggreif:
Consensus is that we'll keep the semicolon field separators.

PR #4789, opened 2024-11-30
Persistent Functions, Persistent Objects, and Persistent Classes

luc-blaeser:
# Persistent Functions, Persistent Objects, and Persistent Classes

Functions, objects, and classes can also be persisted with enhanced orthogonal persistence, i.e. they been used in Motoko's persistent actor variables (aka stable variables in legacy mode).

Note: This is not supported with classical persistence.

With this, the only non-persistent (non-stable) values remain to be lambdas, async handles (aka continuations, futures), local async function references, and errors.

## Example

With this, the following program can then used for orthogonal persistence. For this example, you need a specific core library version, which contains extensions of persistent object-oriented data structures.

```
import Map "mo:core/object-oriented/Map";
import Nat "mo:core/Nat";

persistent actor {
  let map = Map.Map<Nat, Text>(Nat.compare);

  public func main() : async () {
    assert map.isEmpty();
    map.add(1, "One");
    assert map.get(1) == ?"One";
    map.remove(1);
    assert map.get(1) == null;
  }
};
```

## Persistent Modifier

A `persistent` modifier can be applied to functions, classes, and function types, as prerequisite to save functions as part of enhanced orthogonal persistent.

`persistent` function:
```
persistent func tick() { ... }
```

`persistent` class:
```
persistent class Clock() {
  public func tick() { ... }; // implicitly `persistent`
};
```

Functions (methods) of a `persistent` class are implicitly declared `persistent`.

`persistent` function types:
```
type Comparison<T> = persistent (T, T) -> Order.Order`
```

A persistent function implicitly has a corresponding persistent function type.

## Persistent Functions and Persistent Scopes

A persistent function is a function that is declared `persistent` and occurs in a persistent scope.

A persistent scope is:
* the main actor,
* an actor class,
* an imported module,
* a nested module in a persistent scope,
* a `persistent` function in a persistent scope,
* a `persistent` class in a persistent persistent scope, 
* a method in a `persistent` class in a persistent scope, or,
* a named object in a persistent scope.

## Restrictions

Persistent functions anc classes cannot be anonymous.

Persistent functions capture only local variables of surrounding functions/classes/objects (closure) if they are of a stable type.
This restriction does not apply to immutable variables (`let`) and actor variables.

Generic type parameters of persistent functions and persistent classes are bounded to stable types.

## Persistent Function Types

A persistent function type is also a stable type. Thus, a persistent function has a stable type.

A persistent function type is a sub-type of a non-persistent function type with type-compatible signature, i.e. `persistent X' -> Y <: X -> Y'` for `X' <: X` and `Y' :< Y`.

## Upgrades of Persistent Functions

Persistent functions are upgraded as follows:
* All persistent functions that are reachable from persistent actor variables are considered alive.
* Each alive persistent function must have a matching declaration in the new program version.
* Persistent functions match between program versions if they have an equal fully qualified name.
* For matching functions, the function type of the new version must be compatible to the previous version (super-type).
* For matching functions, the closure type in the new version must be compatible with the previous version, see below.

All other functions, such as e.g. functions that are not declared `persistent`, functions in a unnamed objects, functions in a non-persistent class, or functions in a lambda are transient functions.

## Persistent Closures

The closures of persistent functions are represented in a portable format with the following clearly defined binding of captured variables:
* Local variables (`var`) are captured by name.
* Parameters are captured by position (also considering nested functions).

This offers flexibility for changes of persistent closures in new program versions:
* New program version can capture less variables/parameters in persistent functions.
* One can rename parameters even if they are captured by persistent functions.

On a persistent function upgrade, the closure type of the persistent function must remain compatible:
* The new version of the persistent function does not capture more variables/parameters than the previous version.
* The captured variable in the new version is a valid super-type of the previous version.

Specific aspects apply to generic types used in a persistent closure:
* Generic types used for captured variables must match the previous declaration order (e.g. one cannot swap generic types).
* The generic type bounds must remain compatible.
* However, generic types do not need to be reified, see the reasoning in [generics in persistent closures](GenericsInPersistentClosure.md).

## Runtime System Design

Function references are encoded by a function id in the following representation:
* Persistent function id, encoded as non-negative number:
  A persistent function reference that stays invariant across upgrades.
* Transient function id, encoded as negative number:
  A transient function reference that is invalidated on upgrade.

Each program version defines a set of named local functions that can be used as persistent function references. Each such function obtains a persistent function id on program initialization and upgrade. If the persistent function was already declared in the previous version, its function id is reused on upgrade. Thereby, the compatibility of the function type and closure type are checked. Otherwise, if it is a new persistent function, it obtains a new persistent function id, or a recycled id.

The runtime system supports persistent functions by two mechanisms:

1. **Persistent virtual table** for persistent function calls:
    
   The persistent virtual table maps persistent function ids to Wasm table indices, for supporting dynamic calls of persistent functions. Each entry also stores the hashed name of the persistent function to match and rebind the persistent function ids to the corresponding functions of the new Wasm binary on a program upgrade. Moreover, each entry also records the type of the closure, referring to the persistent type table. The table survives upgrades and is built and updated by the runtime system. To build and update the persistent virtual table, the compiler provides a **persistent function map**, mapping the hashed name of a potentially persistent function to the corresponding Wasm table index, plus its closure type pointing to the new type table. For performance, the persistent function map is sorted by the hashed names.
   
2. **Function literal table** for materializing persistent function literals:

   As the compiler does not yet know the function ids of persistent function literals/constants, this table maps a Wasm table index of the current program version to a persistent function id. The function literal table is re-built on program initialization and upgrade. When a persistent function literal is loaded, it serves for resolving the corresponding function id and thus the persistent function reference. The table is discarded on upgrades and (re-)constructed by the runtime system, based on the information of the **persistent function map**.

The runtime system distinguishes between transient and persistent function references by using a different encoding. This is to avoid complicated conversion logic been inserted by the compiler when a persistent function reference is assigned to transient reference, in particular in the presence of sharing (a function reference can be reached by both a persistent and transient function type) and composed types (function references can be deeply nested in a composed value that is assigned).

## Compatibility Check

A persistent function compatibility check is performed by the runtime system on upgrade.
* It checks for a matching function in the new version.
* The function type compatibility is implicitly covered by the upgrade memory compatibility check, since the persistent function in use needs to be reachable by the persistent actor type.
* The closure compatibility is additionally checked for each mapped persistent function. This covers all captured variables of the persistent function. This check is supported by the information of the persistent virtual table and the persistent function map.

Transient function references are represented as negative function ids determining the Wasm table index, specifically `-wasm_table_index - 1`.

## Garbage Collection

The runtime systems relies on a dedicated garbage collector of persistent functions:
* On pre-upgrade, the runtime systems determines which persistent functions are still alive, i.e. transitively reachable from persistent actor variables.
* Only those alive persistent functions need to exist in the new program version.
* All other persistent functions of the previous version are considered garbage and their slots in the virtual table can be recycled.
* For efficiency, the GC is type-directed such that it only selectively traverses fields that may lead to persistent function type. Note: Same objects may be revisited if appearing by a different static type.

Garbage collection is necessary to allow programs to use persistent classes and persistent functions in only transient contexts or not even using imported persistent classes or functions. Moreover, it allows programs to drop persistent functions and classes, if they are no longer used for persistence.

The runtime system reports the fully qualified name of missing persistent functions (not only the internal name hash).


src/mo_types/type.ml
```diff
@@ -68,7 +69,7 @@ and field = {lab : lab; typ : typ; src : src}
 and con = kind Cons.t
 and kind =
   | Def of bind list * typ
-  | Abs of bind list * typ
+  | Abs of bind list * typ * int option
```
crusso, line 72: What's the `int option` for?

src/mo_types/type.ml
```diff
@@ -489,10 +494,9 @@ let close cs t =
   let sigma = List.fold_right2 ConEnv.add cs ts ConEnv.empty in
   subst sigma t
 
-let close_binds cs tbs =
+let close_binds cs tbs is_stable =
```
crusso, line 497: is_stable appears unused?

luc-blaeser: Thanks. A relict...

src/mo_types/type.ml
luc-blaeser: Add comment

src/mo_types/type.ml
```diff
@@ -1049,6 +1059,11 @@ let rec rel_typ d rel eq t1 t2 =
   | _, _ -> false
   end
 
+and rel_sort s1 s2 =
+  match s1, s2 with
+  | Local Stable, Local Flexible -> true (* stable function can be assigned to flexible function *)
```
crusso, line 1064: This should probably require equality when checking is invariant (rel == eq)

src/mo_types/type.ml
```diff
@@ -1035,7 +1045,7 @@ let rec rel_typ d rel eq t1 t2 =
   | Tup ts1, Tup ts2 ->
     rel_list d rel_typ rel eq ts1 ts2
   | Func (s1, c1, tbs1, t11, t12), Func (s2, c2, tbs2, t21, t22) ->
-    s1 = s2 && c1 = c2 &&
+    rel_sort s1 s2 && c1 = c2 &&
```
crusso, line 1048:
```suggestion
    rel_sort rel eq s1 s2 && c1 = c2 &&
```

src/mo_types/type.ml
crusso, line 1064:
```suggestion
and rel_sort rel eq s1 s2 =
  match s1, s2 with
  | Local Stable, Local Flexible -> 
     not (rel == eq) (* stable function subtype flexible functions *)
```

src/mo_types/type.ml
```diff
@@ -1399,6 +1414,18 @@ let glb t1 t2 = let glbs = ref M.empty in combine glbs (ref M.empty) glbs t1 t2
...
+(* Stable function support *)
+
+type captured_variable = {
+  stable_name: string; (* positional id for parameters, named id for locals *)
+  variable_type: typ;
+}
+
+type stable_closure = {
+  function_path: string list; (* fully qualified function name *)
+  captured_variables: captured_variable Env.t; (* captured mutable variables *)
```
crusso, line 1426:
```suggestion
  captured_variables: captured_variable Env.t; (* captured mutable variables *)
```
Just the mutable ones?

src/mo_types/type.mli
```diff
@@ -259,6 +264,17 @@ val open_binds : bind list -> typ list
 
 module Env : Env.S with type key = string
 
+(* Stable function support *)
+
+type captured_variable = {
+  stable_name: string; (* positional id for parameters, named id for locals *)
```
crusso, line 270: Maybe a dedicated variant would be be than just a string.

src/mo_def/syntax.ml
```diff
@@ -172,7 +172,7 @@ and exp' =
   | AssignE of exp * exp                       (* assignment *)
   | ArrayE of mut * exp list                   (* array *)
   | IdxE of exp * exp                          (* array indexing *)
-  | FuncE of string * sort_pat * typ_bind list * pat * typ option * sugar * exp  (* function *)
+  | FuncE of string * sort_pat * typ_bind list * pat * typ option * sugar * function_context * exp  (* function *)
```
crusso, line 175: An alternative might have been to make the (mutable) function_context a custom annotation on the sort_pat (since we can update those already).

Just an idea, not saying it should be done.

src/ir_def/rename.ml
```diff
@@ -62,10 +62,10 @@ and exp' rho = function
   | DeclareE (i, t, e)  -> let i',rho' = id_bind rho i in
                            DeclareE (i', t, exp rho' e)
   | DefineE (i, m, e)   -> DefineE (id rho i, m, exp rho e)
-  | FuncE (x, s, c, tp, p, ts, e) ->
+  | FuncE (x, s, c, tp, p, ts, closure, e) ->
```
crusso, line 65: I closure records the set of free variables, don't those need to be renamed consistently too? But if they are, doesn't that break the names of stable variables?

src/ir_def/arrange_ir.ml
```diff
@@ -26,7 +26,7 @@ let rec exp e = match e.it with
   | AsyncE (Type.Cmp, tb, e, t) -> "AsyncE*"  $$ [typ_bind tb; exp e; typ t]
   | DeclareE (i, t, e1) -> "DeclareE" $$ [id i; exp e1]
   | DefineE (i, m, e1)  -> "DefineE" $$ [id i; mut m; exp e1]
-  | FuncE (x, s, c, tp, as_, ts, e) ->
+  | FuncE (x, s, c, tp, as_, ts, _, e) ->
```
crusso, line 29: Might be worth displaying the closure environment

src/ir_def/check_ir.ml
```diff
@@ -259,7 +259,7 @@ and check_con env c =
   else
   begin
     env.seen := T.ConSet.add c !(env.seen);
-    let T.Abs (binds,typ) | T.Def (binds, typ) = Cons.kind c in
+    let T.Abs (binds,typ,_) | T.Def (binds, typ) = Cons.kind c in
```
crusso, line 262:
```suggestion
    let T.Abs (binds, typ, _) | T.Def (binds, typ) = Cons.kind c in
```

src/ir_def/check_ir.ml
```diff
@@ -581,10 +584,10 @@ let rec check_exp env (exp:Ir.exp) : unit =
...
     | CPSAsync (s, t0), [exp] ->
       (match typ exp with
-       | T.Func (T.Local, T.Returns, [tb],
-                 T.[Func (Local, Returns, [], ts1, []);
-                    Func (Local, Returns, [], [t_error], []);
-                    Func (Local, Returns, [], [], [])],
+       | T.Func (T.Local _, T.Returns, [tb],
+                 T.[Func (Local _, Returns, [], ts1, []);
+                    Func (Local _, Returns, [], [t_error], []);
+                    Func (Local _, Returns, [], [], [])],
```
crusso, line 590:
```suggestion
       | T.Func (T.Local T.Flexible, T.Returns, [tb],
                 T.[Func (Local Flexible, Returns, [], ts1, []);
                    Func (Local Flexible, Returns, [], [t_error], []);
                    Func (Local Flexible, Returns, [], [], [])],
```
I guess those _ are all Flexible?

src/ir_def/construct.ml
```diff
@@ -177,7 +177,7 @@ let cps_asyncE s typ1 typ2 e =
 
 let cps_awaitE s cont_typ e1 e2 =
   match cont_typ with
-  | T.Func(T.Local, T.Returns, [], _, ts2) ->
+  | T.Func(T.Local _, T.Returns, [], _, ts2) ->
```
crusso, line 180:
```suggestion
  | T.Func(T.Local T.Flexible, T.Returns, [], _, ts2) ->
```

src/mo_types/type.ml
```diff
@@ -14,9 +14,10 @@ type obj_sort =
  | Module
  | Memory          (* (codegen only): stable memory serialization format *)
 
+type stable_sort = Flexible | Stable
 type async_sort = Fut | Cmp
 type shared_sort = Query | Write | Composite
-type 'a shared = Local | Shared of 'a
+type 'a shared = Local of stable_sort | Shared of 'a
```
crusso, line 20:
```suggestion
type ('a, 'b) shared = Local | Stable of 'b | Shared of 'a
```
I wonder if this might have reduced some code changes elsewhere... Don't do it now, just a note to self.
Then you could attach the free variable sets to stable functions only, using a suitable instantiations of 'b.

src/mo_frontend/typing.ml
```diff
@@ -280,15 +300,92 @@ let detect_unused env inner_identifiers =
...
+let generic_variable_index env =
+  let index = !(env.generic_variable_count) in
+  env.generic_variable_count := index + 1;
+  Some index
+
+let append_parameter parameters pattern =
+  match pattern.it with
+  | VarP id | AnnotP ({it = VarP id; _}, _) ->
+    let { count; positions } = parameters in
+    let positions = T.Env.remove id.it positions in
```
crusso, line 373:
```suggestion
    let positions = T.Env.remove id.it positions in
```
I don't think there's any need to remove, is there?

src/mo_frontend/typing.ml
```diff
@@ -847,13 +947,14 @@ and check_typ_bind_sorts env tbs =
   (* assert, don't error, since this should be a syntactic invariant of parsing *)
   List.iteri (fun i tb -> assert (i = 0 || (tb.T.sort = T.Type))) tbs;
 
-and check_typ_binds env typ_binds : T.con list * T.bind list * Scope.typ_env * Scope.con_env =
+and check_typ_binds env stable_scope typ_binds : T.con list * T.bind list * Scope.typ_env * Scope.con_env =
```
crusso, line 950: I don't think `stable_scope` is used to enforce any check in the body. Is it redundant or the intended check unimplemented?

src/mo_frontend/typing.ml
```diff
@@ -869,19 +970,21 @@ and check_typ_binds env typ_binds : T.con list * T.bind list * Scope.typ_env * S
...
     | k' -> assert (eq_kind env Source.no_region k k')
   ) cs ks;
   let env' = add_typs env xs cs in
   let _ = List.map (fun typ_bind -> check_typ env' typ_bind.it.bound) typ_binds in
   List.iter2 (fun typ_bind c -> typ_bind.note <- Some c) typ_binds cs;
   cs, tbs, te, T.ConSet.of_list cs
 
-and check_typ_bind env typ_bind : T.con * T.bind * Scope.typ_env * Scope.con_env =
-  match check_typ_binds env [typ_bind] with
+and check_typ_bind env stable_scope typ_bind : T.con * T.bind * Scope.typ_env * Scope.con_env =
```
crusso, line 986: Again, I don't think `stable_scope` enforces anything.

src/mo_types/type.ml
```diff
@@ -851,18 +858,20 @@ let serializable allow_mut t =
       | Mut t -> allow_mut && go t
       | Con (c, ts) ->
         (match Cons.kind c with
-        | Abs _ -> false
+        | Abs (bind_list, _, _) ->
+          allow_stable_functions
```
crusso, line 862: This means we consider all type parameters stable. Is that even sound? I wouldn't expect it to be.

src/mo_frontend/typing.ml
```diff
@@ -1468,25 +1580,51 @@ and infer_exp'' env exp : T.typ =
...
       check_exp_strong (adjoin_vals env'' ve2) codom exp1;
-      leave_scope env ve2 initial_usage;
+      leave_scope env'' ve2 initial_usage;
+      assert(!closure = None);
+      closure := stable_function_closure env'' named_scope;
+      env.captured := S.union !(env''.captured) !(env'.captured);
+      (match !closure with
+      | Some Type.{ captured_variables; _ } ->
+        T.Env.iter (fun id T.{ variable_type; _ } ->
+          if not (T.stable variable_type) then
```
crusso, line 1622: I just checked the definition of T.stable and it appears to consider type parameters as stable. Why is that ok?

src/mo_frontend/typing.ml
```diff
@@ -1514,13 +1652,14 @@ and infer_exp'' env exp : T.typ =
       end
     end;
     let ts1 = match pat.it with TupP _ -> T.seq_of_tup t1 | _ -> [t1] in
+    let sort = if is_flexible && sort = T.Local T.Stable then T.Local T.Flexible else sort in
```
crusso, line 1655: So this is implicitly weakening the sort to flexible if necessary.
What happens if the function was recursive, and you assumed it was stable for all recursive calls, but then decided it was flexible after all. Can that lead to problems?

src/mo_frontend/typing.ml
```diff
@@ -2105,6 +2249,13 @@ and infer_call env exp1 inst exp2 at t_expect_opt =
       (* explicit instantiation, check argument against instantiated domain *)
       let typs = match inst.it with None -> [] | Some (_, typs) -> typs in
       let ts = check_inst_bounds env sort tbs typs t_ret at in
+      (if sort = T.Local T.Stable then
```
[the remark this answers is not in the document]
crusso: Doesn't this mean that our stable data structures are effectively limited to stable instantiations. So `Map<Nat, () ->()>(Nat.compare)` is not allowed, if Map is  stable class, but `Map.empty<Nat, stable () -> ()>(Nat.compare)`) is allowed.

design/StableFunctions.md
```diff
@@ -0,0 +1,126 @@
...
+* For matching functions, the function type of the new version must be compatible to the previous version (super-type).
+* For matching functions, the closure type in the new version must be compatible with the previous version, see below.
+
+All other functions, such as lambdas, named functions in a lambda, async functions, or functions imported from a module without a unique import identifier, are flexible functions.
+
+## Stable Closures
+
+The closures of stable functions are represented in a portable format with the following clearly defined binding of captured variables:
+* Local variables are captured by name.
+* Parameters are captured by position (also considering nested functions).
```
crusso, line 70: I'm not sure if position is really well defined since we have argument binding by pattern matching.

design/StableFunctions.md
```diff
@@ -0,0 +1,126 @@
...
+
+This offers flexibility for changes of stable closures in new program versions:
+* New program version can capture less variables/parameters in stable functions.
+* One can rename parameters even if they are captured by stable functions.
```
crusso, line 74: By implication, I guess one cannot rename local variables. That seems a bit fragile since it violates what programmers may think is allowed by refactoring.

design/StableFunctions.md
```diff
@@ -0,0 +1,126 @@
...
+
+A stable scope is:
+* the main actor,
+* an actor class,
+* a module imported with a unique identifier from a stable scope,
+* a named non-async function in a stable scope,
+* a class in a stable scope, or,
+* a named object in a stable scope.
+
+Generic type parameters of stable functions and stable classes are bounded to stable types.
```
crusso, line 46:
```suggestion
Generic type parameters of stable functions and stable classes are bounded to stable types.
```
I'm not sure this is fully implemented given the counter-example in #5143. Maybe we need another sort of bound, beyond the current Scope|Type - Scope|Type|StableType. Then we can require type parameters of stable functions to be StableType, and type parameters of local functions to be either Type Or StableType. 
Parameters with bound StableType can only be instantiated at stable types, and only parameters with bound StableType are stable types (currently any parameter is a stable type).

design/StableFunctions.md
```diff
@@ -0,0 +1,126 @@
...
+
+A stable function is also a stable type.
+
+Syntactically, function types are prefixed by `stable` to denote a stable function, e.g.
+`stable X -> Y`. Stable functions implicitly have a corresponding stable reference type.
+
+A stable function type is a sub-type of a flexible function type with type-compatible signature, i.e. `stable X' -> Y <: X -> Y'` for `X' <: X` and `Y' :< Y`.
```
crusso, line 53: Generic functions require some more care. You probably want the type parameters to match (exactly), with equivalent sorts and bounds.

src/mo_types/type.ml
luc-blaeser: I changed this here to consider the generic type parameters.
Stable functions with stable type parameters cannot be promoted to flexible functions with implicitly flexible type parameters.

I thought `rel` would always be sub-typing but probably I missed a `RelArg` on equality?

luc-blaeser: This means that rel can also be equality check? (Did not find that but probably I oversaw) Thanks for noticing

src/mo_types/type.ml
luc-blaeser: It actually captures only mutable variables that then implicitly become stable. Immutable ones are not part of the closures, directly referenced or copied.

src/mo_types/type.mli
luc-blaeser: Yes, agree. Good suggestion

src/mo_def/syntax.ml
luc-blaeser: I find this a good idea. Thanks for raising

src/ir_def/rename.ml
luc-blaeser: Yes, consistent renaming would be needed, maybe by indirection or a symbol table in longer run. When renaming closure set would be consistently updated too.

src/ir_def/arrange_ir.ml
luc-blaeser: Agree.

src/ir_def/check_ir.ml
luc-blaeser: Will batch with another commit, to avoid long CI runs :)

src/ir_def/check_ir.ml
```diff
@@ -318,8 +318,11 @@ and check_typ_bounds env (tbs : T.bind list) typs at : unit =
   List.iter2
     (fun tb typ ->
       check env at (T.sub typ (T.open_ typs tb.T.bound))
-        "type argument does not match parameter bound")
-    tbs typs
+        "type argument does not match parameter bound";
+      (* TODO: Stable functions: Check this *)
```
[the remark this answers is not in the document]
luc-blaeser: Thanks for noticing. Should be enabled and checked

src/mo_types/type.ml
luc-blaeser: Can indeed be a simplification. Worth checking

src/mo_frontend/typing.ml
```diff
@@ -280,15 +300,92 @@ let detect_unused env inner_identifiers =
...
+   For nested generic classes and functions, the numbering is continued
+   in the sub-scopes. *)
+let generic_variable_index env =
+  let index = !(env.generic_variable_count) in
+  env.generic_variable_count := index + 1;
+  Some index
+
+let append_parameter parameters pattern =
+  match pattern.it with
+  | VarP id | AnnotP ({it = VarP id; _}, _) ->
```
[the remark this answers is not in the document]
luc-blaeser: IIRC positions should consider nesting. But certainly good to double check.
I need to check the record pattern on parameter. Thanks for noticing

src/mo_frontend/typing.ml
luc-blaeser: Yes, redundant, as `add` is insert or replace :)

src/mo_frontend/typing.ml
luc-blaeser: Yes, a relict, I first thought that bounds would need to be stable too, but this is not necessary as the type parameter is already required stable if needed. Can be removed.

src/mo_frontend/typing.ml
luc-blaeser: Yes, a relict. Thanks for noticing

src/mo_types/type.ml
luc-blaeser: Yes, for stable functions, type parameters are implicitly stable.
For flexible functions not.
The type arguments are also checked to be stable for stable functions only.

I initially thought this would be visible restriction but didn't notice any in the test code so far. My reasoning is that lambdas for type arguments are relatively rare and if needed, a flexible function can do.

src/mo_frontend/typing.ml
luc-blaeser: This check here iterates over the variables, not type parameters. The check for stable/flexible type parameters is in `infer_call`.

`T.stable` just allows type parameters in general when EOP is on. The rule could be refined but the main check (no flexible type argument for stable parameter) is already in place at that separate location.

src/mo_frontend/typing.ml
luc-blaeser: No, this is only because the parser (due LR) does not yet consider nested scopes and just gives an initial classification on stable/flexible. Later on, I consider the nested scopes, and if the surrounding scope is flexible, a parser-assumed stable keyword is weakened to flexible too. A LL-parser could handle that. Or, ideally, this classification could be done on symbol table only. And the parser only records explicit `stable` use, no implied annotations yet.

src/mo_frontend/typing.ml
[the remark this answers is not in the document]
luc-blaeser: Yes. I also thought about this but think this case is really rare that (imperatively-biased) classes get used in (functionally-biased) lambdas for type parameters. I did not find any relevant limitation case in existing tests. 

I also believe no double types are needed, one for stable and one for flexible. So far, my rationale.

design/StableFunctions.md
luc-blaeser: Yes, true. Maybe we should forbid pattern matching on parameters on stable functions. Or use names for pattern matching as done now IIRC.

design/StableFunctions.md
luc-blaeser: No, those would be captured by name. I believe reason is by the weak nature of closures (a encapsulation breach in the sense that it "silently" connects to intrinsics of its environment without a proper contract and listing in type).

I was thinking of alternative, but thought this is probably the simplest and still pragmatic approach. The binding of captured variables also happens by name. So, I am even not so sure of the positional binding tbh.

design/StableFunctions.md
luc-blaeser: I believe it is solved, pls. see latest commit.

design/StableFunctions.md
luc-blaeser: True. A specification of generics would be needed (as currently implemented):
* A stable function with type parameters cannot be promoted to a flexible functions. This is because stable functions have implicitly stable type parameters, while flexible functions have flexible type parameters.
* Bounds need to match. This should already be the case with classical implementation.

luc-blaeser:
> The current implementation isn't quite right since it breaks alpha-conversion of local names in the ir. This is too fragile.
> 
> See [#5143 (comment)](https://github.com/dfinity/motoko/pull/5143#issuecomment-2901458950) for a repro, possible explanation and fix.

I believe this is more compiler-internal thing, rather than any pressing conceptual thing. If we rename in a future compiler step, we have to rename all occurrences consistently, either with extra path or preparing renaming e.g. by indirecting variables, or on longer term, using a symbol table.

luc-blaeser:
I believe main aspects are addressed:
* The small type check issue on type parameters for stable->flexible promotion: Fixed
* Renaming support for future: Probably not a blocker, as it is compiler-internal, could be added even later, or maybe now if time is
* Backwards-compatibility to classical persistence: I would suggest to skip and just go forward with EOP.

luc-blaeser:
Ah, also, merging in latest changes would be again needed (I know this is usually quite painful)

crusso:
Stable class as possible anti-pattern: Adding a method to a stable class is a breaking change for any client that has an instance of the class in a stable variable. 

This means it will be difficult to extend the API of any stable class,unless we still separate the data fiels from the functions as we are currently forced to do.

Instance migration might still be possible but only using an explicit migration function that extends the instance with any new method.

crusso:
> > The current implementation isn't quite right since it breaks alpha-conversion of local names in the ir. This is too fragile.
> > 
> > See [#5143 (comment)](https://github.com/dfinity/motoko/pull/5143#issuecomment-2901458950) for a repro, possible explanation and fix.
> 
> I believe this is more compiler-internal thing, rather than any pressing conceptual thing. If we rename in a future compiler step, we have to rename all occurrences consistently, either with extra path or preparing renaming e.g. by indirecting variables, or on longer term, using a symbol table.

I will try to fix the issue before merging. Some passes already do a little bit of renaming (desugar.ml, maybe others) and maybe vulnerable to this issue.  However, I do think it's fixable.

crusso:
> Unfortunately, I have little time to work on smaller fixes on this PR due to the other project I have. Only if possible, and okay for you, I would be super grateful if you could maybe directly refine/adjust smaller pieces if needed?
> Would be great if we can get this merged. Thanks a lot. Always happy to discuss things about this PR

Sure, I'll have to once back

crusso:
> I believe main aspects are addressed:
> * The small type check issue on type parameters for stable->flexible promotion: Fixed
> * Renaming support for future: Probably not a blocker, as it is compiler-internal, could be added even later, or maybe now if time is
> * Backwards-compatibility to classical persistence: I would suggest to skip and just go forward with EOP.


Your fix appears to be to restrict subtyping to non-generic functions. Might be good enough but probably more restrictive than adding a sort to type parameters.

Other options:
* Restrict all type parameters to stable type. 
* Add a sort to type parameters and require equivalent sort when subtyping, but then users would need to be able to declare stable type parameters for non-stable functions/classes.

luc-blaeser:
Note for review: 
* Pls. double check that `block_size()` is not used in logic, as this is not supported when incremental GC is in marking phase. Otherwise, `block_size()` could be adjusted to cope with this.

crusso:
Found another problem - migration functions can induce use-before-define errors when a stable function is invoked before it's free variables are properly defined. 
Counter-example:
https://github.com/dfinity/motoko/pull/5382

crusso:
Found a bug while experimenting:

crash.mo:
```
crusso@crusso-Virtual-Machine:~/motoko/test/run-drun$ more crash.mo
persistent actor {
   persistent func g<T>(x : T) : persistent () -> T {
     persistent func h() : T { x };
     h;
   };
   stable let h = g<persistent () -> ()>(func () {});
}
```
produces:
```
Fatal error: exception File "mo_frontend/typing.ml", line 2311, characters 32-38: Assertion failed
```
(Stack trace below)

I think there's some missing logic in bi_match.ml when comparing function sorts. You might also want to check if Type.lub and Type.glb are handling the finer sorts correctly.

```
crusso@crusso-Virtual-Machine:~/motoko/test/run-drun$ moc crash.mo
OOPS! You've triggered a compiler bug.
Please report this at https://github.com/dfinity/motoko/issues/new with the following details:

Motoko (source 0.16.1-188-gc88e5ef01)

Fatal error: exception File "mo_frontend/typing.ml", line 2311, characters 32-38: Assertion failed
Raised at Mo_frontend__Typing.check_func_step in file "mo_frontend/typing.ml", line 2311, characters 32-44
Called from Mo_frontend__Typing.check_exp' in file "mo_frontend/typing.ml", line 2234, characters 26-108
Called from Mo_frontend__Typing.check_exp in file "mo_frontend/typing.ml", line 2089, characters 11-45
Called from Mo_frontend__Typing.infer_call in file "mo_frontend/typing.ml", line 2397, characters 26-58
Called from Mo_frontend__Typing.infer_exp'' in file "mo_frontend/typing.ml", line 1775, characters 12-53
Called from Mo_frontend__Typing.infer_exp_wrapper in file "mo_frontend/typing.ml", line 1430, characters 10-21
Called from Mo_frontend__Typing.infer_exp in file "mo_frontend/typing.ml" (inlined), line 1414, characters 2-31
Called from Mo_frontend__Typing.infer_dec in file "mo_frontend/typing.ml", line 3489, characters 12-29
Called from Mo_frontend__Typing.infer_block in file "mo_frontend/typing.ml", line 3454, characters 10-55
Called from Mo_frontend__Typing.infer_obj in file "mo_frontend/typing.ml", line 3113, characters 17-53
Called from Mo_frontend__Typing.infer_exp'' in file "mo_frontend/typing.ml", line 1600, characters 12-61
Called from Mo_frontend__Typing.infer_exp_wrapper in file "mo_frontend/typing.ml", line 1430, characters 10-21
Called from Mo_frontend__Typing.infer_exp in file "mo_frontend/typing.ml" (inlined), line 1414, characters 2-31
Called from Mo_frontend__Typing.infer_exp'' in file "mo_frontend/typing.ml", line 1918, characters 12-31
Called from Mo_frontend__Typing.infer_exp_wrapper in file "mo_frontend/typing.ml", line 1430, characters 10-21
Called from Mo_frontend__Typing.infer_exp in file "mo_frontend/typing.ml" (inlined), line 1414, characters 2-31
Called from Mo_frontend__Typing.infer_exp_promote in file "mo_frontend/typing.ml", line 1420, characters 10-27
Called from Mo_frontend__Typing.infer_exp'' in file "mo_frontend/typing.ml", line 1926, characters 13-39
Called from Mo_frontend__Typing.infer_exp_wrapper in file "mo_frontend/typing.ml", line 1430, characters 10-21
Called from Mo_frontend__Typing.infer_exp in file "mo_frontend/typing.ml" (inlined), line 1414, characters 2-31
Called from Mo_frontend__Typing.infer_dec in file "mo_frontend/typing.ml", line 3482, characters 16-33
Called from Mo_frontend__Typing.infer_block in file "mo_frontend/typing.ml", line 3454, characters 10-55
Called from Mo_frontend__Typing.infer_prog.(fun) in file "mo_frontend/typing.ml", line 3957, characters 26-66
Called from Mo_frontend__Typing.recover_opt.(fun) in file "mo_frontend/typing.ml", line 152, characters 55-60
Called from Mo_frontend__Typing.recover_with in file "mo_frontend/typing.ml", line 151, characters 56-59
Called from Diag.with_message_store in file "lang_utils/diag.ml", line 89, characters 10-13
Called from Pipeline.infer_prog.(fun) in file "pipeline/pipeline.ml", line 191, characters 12-69
Called from Mo_types__Cons.session in file "mo_types/cons.ml", line 35, characters 19-23
Re-raised at Mo_types__Cons.session in file "mo_types/cons.ml", line 40, characters 5-12
Called from Mo_types__Cons.session in file "mo_types/cons.ml", line 35, characters 19-23
Re-raised at Mo_types__Cons.session in file "mo_types/cons.ml", line 40, characters 5-12
Called from Pipeline.check_progs.go in file "pipeline/pipeline.ml", line 214, characters 8-106
Called from Pipeline.load_progs_cached in file "pipeline/pipeline.ml", line 548, characters 23-57
Called from Diag.bind in file "lang_utils/diag.ml", line 32, characters 27-30
Called from Diag.bind in file "lang_utils/diag.ml", line 32, characters 27-30
Called from Diag.bind in file "lang_utils/diag.ml", line 32, characters 27-30
Called from Diag.bind in file "lang_utils/diag.ml", line 32, characters 27-30
Called from Pipeline.load_progs in file "pipeline/pipeline.ml", line 563, characters 4-78
Called from Pipeline.compile_files in file "pipeline/pipeline.ml", line 912, characters 27-90
Called from Dune__exe__Moc.process_files in file "exes/moc.ml", line 298, characters 49-94
Called from Dune__exe__Moc in file "exes/moc.ml", line 395, characters 10-29
```

crusso:
Is this behaviour intentional:

```
crusso@crusso-Virtual-Machine:~/motoko/test/run-drun$ more crusso.mo
persistent actor {
   func g<T>(x : T) : persistent () -> T {
     persistent func h() : T { x };
   };
}
```
crusso@crusso-Virtual-Machine:~/motoko/test/run-drun$ moc --enhanced-orthogonal-persistence crusso.mo
crusso.mo:3.6-3.35: type error [M0096], expression of type
  () -> T
cannot produce expected type
  persistent () -> T
```
It looks like declaring a persistent function in a non-persistent scope just weakens its type rather than rejecting the declaration.

File: src/mo_types/type.ml [hunk 27]
```diff
@@ -16,10 +16,11 @@ type obj_sort =
  | Module
```
crusso, line 17: I think you also need to adjust the lub and glb computation on types to account for the new subtyping between stable and local functions (to match your changes to subtyping)

for example
```
func l() {};
persistent func p() {};
let g = if b l else p;
```
Typing g requires you to take the lub of the local and persistent function types of  `l` and `p`. You code will probably produce Any for g rather that (*local*) () -> ()

crusso:
#5483 demonstrates an unsoundness (allowing a persistent definition to evolve to a supertype is not sound, since it might break aliases of that definition used at another type (that may no longer be related to the supertype, even if it was related to the original type).

I think the safe direction is that the definition can evolve to a subtype, it if evolves at all.

rts/motoko-rts/src/persistence/compatibility.rs
```diff
@@ -136,37 +138,72 @@ unsafe fn create_type_check_cache<M: Memory>(
...
+    /// Test whether the new stable type complies with the existing old stable type.
+    /// This uses the existing IDL subtype test.
+    pub unsafe fn is_compatible(&self, old_type_index: i32, new_type_index: i32) -> bool {
+        debug_assert!(
+            old_type_index != MAIN_ACTOR_TYPE_INDEX || new_type_index == MAIN_ACTOR_TYPE_INDEX
+        );
+        self.cache.init(); // Needs to be-reinitialized on every check.
+        crate::idl::memory_compatible(
+            &self.cache,
+            TypeVariance::Covariance,
```
crusso, line 193: Assuming this is checking the closure signatures, not environment, I think you should be checking that the old signature is a supertype of the new signature 
(c.f. #5483).

rts/motoko-rts/src/persistence/compatibility.rs
```diff
@@ -136,37 +138,72 @@ unsafe fn create_type_check_cache<M: Memory>(
...
+            self.new_table_end,
+            old_type_index,
+            new_type_index,
+            new_type_index == MAIN_ACTOR_TYPE_INDEX,
+        )
+    }
+
+    // Test whether the new stable type of the main actor complies with
+    // the existing old stable type of the main actor.
+    pub unsafe fn compatible_stable_actor(&self) -> bool {
```
crusso, line 206: Is this just comparing the stable actor type to itself? Is this just a sanity check?

design/PersistentFunctions.md
```diff
@@ -0,0 +1,169 @@
...
+
+A persistent function type is a sub-type of a non-persistent function type with type-compatible signature, i.e. `persistent X' -> Y <: X -> Y'` for `X' <: X` and `Y' :< Y`.
+
+## Upgrades of Persistent Functions
+
+Persistent functions are upgraded as follows:
+* All persistent functions that are reachable from persistent actor variables are considered alive.
+* Each alive persistent function must have a matching declaration in the new program version.
+* Persistent functions match between program versions if they have an equal fully qualified name.
+* For matching functions, the function type of the new version must be compatible to the previous version (super-type).
```
crusso, line 94:
```suggestion
* For matching functions, the function type of the new version must be compatible to the previous version (subtype).
```
I think this was the wrong way around, also in the implementation.

design/PersistentFunctions.md
```diff
@@ -0,0 +1,169 @@
...
+* For matching functions, the closure type in the new version must be compatible with the previous version, see below.
+
+All other functions, such as e.g. functions that are not declared `persistent`, functions in a unnamed objects, functions in a non-persistent class, or functions in a lambda are transient functions.
+
+## Persistent Closures
+
+The closures of persistent functions are represented in a portable format with the following clearly defined binding of captured variables:
+* Local variables (`var`) are captured by name.
```
crusso, line 102:
```suggestion
* Local variables (`var`) are captured by name.
```
Don't you have to capture all local variable, not just mutable ones?

Say a class method closes over a local let in the constructor. That value needs to be captured too. Something like:
```
persistent Class Box<T>(v :T) {
    let w = v;
    func get() : T { w };
}
```
I'm assuming the closure for get has a slot for w, eventhough w is immutable.

design/PersistentFunctions.md
```diff
@@ -0,0 +1,169 @@
...
+* Parameters are captured by position (also considering nested functions).
```
crusso, line 103: > * Parameters are captured by position (also considering nested functions).

I think this only makes sense if persistent functions only take tuples as arguments - otherwise the positions of identifiers bound within patterns are not well-defined.

design/PersistentFunctions.md
crusso: > The function type compatibility is implicitly covered by the upgrade memory compatibility check, since the persistent function in use needs to be reachable by the persistent actor type.

Maybe this is the source of error. Checking that potential aliases only are used a supertype doesn't guarantee the signature of the definition evolves to a subtype. Perhaps an explicit check on reachable definition signatures is required (if missing)

design/PersistentFunctions.md
crusso: Or by `var` do you just mean identifiers in general, mutable (declared with `var`) or not?

rts/motoko-rts/src/idl.rs
```diff
@@ -799,33 +808,64 @@ pub(crate) unsafe fn memory_compatible(
...
+                        false,
+                    ) {
+                        return false;
+                    }
                 }
             }
-            a11 == a21 && a12 == a22 && a13 == a23
+            true
+        }
+        (IDL_EXT_type_variable, IDL_EXT_type_variable) => {
```
crusso, line 860: Probably just  a minor issue but with Motoko subtyping, you are able to promote a parameter to its bound, and this code doesn't seem to allow that.

For example:

```motoko
crusso@crusso-Virtual-Machine:~/sf/motoko/src$ rlwrap moc
Motoko compiler (source 0.16.1-58-g33e114588)
> func f(g :<T <: Nat> T -> Nat){};
let f : (g : <T <: Nat>T -> Nat) -> () = <func>
> f(func <T <: Nat >(x:T) : T {x});
```
I'm not sure how to reconcile these two. Either we add a  context carrying the bounds of type parameters and variables or add the bound to each occurrence in the type descriptor,  or we restrict Motoko subtyping so we ignore the bounds when checking comparing the domain and codomain of a function. The latter would mean that the implementation of a function type could exploit the bounds in the code, but subtyping between functions would not.

rts/motoko-rts/src/idl.rs
```diff
@@ -799,33 +808,64 @@ pub(crate) unsafe fn memory_compatible(
...
-                    _ => {}
+            // There is at most one annotation per function in our persistent type table.
+            assert_eq!(annotation_count1, 1);
+            assert_eq!(annotation_count2, 1);
+
+            let annotation1 = read_byte(&mut tb1);
+
+            let annotation2 = read_byte(&mut tb2);
+
+            if annotation1 != annotation2 {
```
crusso, line 827: Can local, but not persistent, function signatures appear in the type descriptor  (say for a persistent map function). Then I would expect the local descriptor to have its own annotation (separate from stable_local_function) and allow some subtyping between stable and local.

Are we missing a type descriptor?

src/codegen/compile_enhanced.ml
```diff
@@ -6920,22 +7043,25 @@ module Serialization = struct
...
             add_leb128 0; (* no annotation *)
           | Shared Query, _ ->
             add_leb128 1; add_u8 1; (* query *)
           | Shared Composite, _ ->
             add_leb128 1; add_u8 3; (* composite *)
-          | _ -> assert false
+          | Local _, _ -> (* local *)
+            (* Note: These are generally stable functions, however with one exception:
+               In function parameters, flexible local functions can even occur. *)
+            add_leb128 1; add_u8 255; (* local *)
```
crusso, line 7141: If we allow both flavours of local functions then we should have two (not just one) type descriptors (254 and 255) and subtyping between them I think.

PR #4791, opened 2024-12-02
chore: use `macos-13` as `macos-12` is deprecated

ggreif:
See: https://github.com/actions/runner-images/issues/10721

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,7 @@ jobs:
   tests:
     strategy:
       matrix:
-        os: [ ubuntu-latest, macos-12 ]
+        os: [ ubuntu-latest, macos-13 ]
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -7,7 +7,7 @@ jobs:
   tests:
     strategy:
       matrix:
-        os: [ ubuntu-latest, macos-12 ]
+        os: [ ubuntu-latest, macos-12, macos-13 ]
```
ggreif, line 10: `macos-12` will be removed in a follow-up PR, as removing it now renders the CI stuck. Maybe there is a better way, I just don't know...

luc-blaeser (approved):
Thank you for keeping CI running!

.github/workflows/test.yml
ggreif: Okay, this works now, IDX changed the required check.

luc-blaeser (approved):
Thanks!

ggreif:
@Mergifyio refresh

PR #4798, opened 2024-12-04
bugfix: bound subtyping depth to avoid reliance on unpredictable stack overflow

### LINKED ISSUE #3057: Segmentation fault: 11 when trying to define a method which returns a new “instance” of a class

crusso:
The idea is to fail fast and deterministically, especially in VSCode extension. OCaml stack overflow detection is unreliable and sometimes issues OOM, sometimes stack overflow.

Fixes #3057.

- [x] turn failure into compiler error


src/lang_utils/error_codes/M0200.md
```diff
@@ -0,0 +1,44 @@
...
+Another workaround is to define the problematic method as a separate
+function, outside of the class:
+
+``` motoko
+class Box<T>(v : T) {
+  public let value = v
+};
+
+func map<T, R>(b : Box<T>, f : T -> R) : Box<R> {
+  Box<R>(f (b.value))
```
crusso, line 42:
```suggestion
  Box<R>(f(b.value))
```

rvanasa (dismissed):

ggreif (dismissed):
LGTM, but can you check whether it catches #3057?

crusso:
> LGTM, but can you check whether it catches #3057?

Yeah, it does. Added test.

test/fail/issue-3057.mo
```diff
@@ -0,0 +1,7 @@
+module {
+  public class Id<A>(x : A) {
+    public func map<B>(f : A -> B) : Id<B> {
+      Id(f(x));
+    };
+  };
+};
```
crusso, line 7:
```suggestion
};

```

ggreif (approved):
LGTM!

PR #4800, opened 2024-12-05
ci(Mergify): configuration update

ggreif:
This change has been made by @ggreif from the Mergify Queue Rule Configurator.

### DIFF
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -75,10 +75,10 @@
         "homepage": "",
         "owner": "dfinity",
         "repo": "motoko-base",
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -38,7 +37,8 @@ pull_request_rules:
       - label=autoclose
     actions:
       close:
-        message: CI looks good, this dependency bump would not cause problems, hence closing this trial PR.
+        message: CI looks good, this dependency bump would not cause problems, hence
+          closing this trial PR.
```
ggreif, line 41:
```suggestion
        message: CI looks good, this dependency bump would not cause problems, hence closing this trial PR.
```

PR #4811, opened 2024-12-09
experiment: mo-doc add option to strip comments

crusso:
To ease overview of API changes 

TODO:

- [ ] Use a global flag instead? Passing down the flag is a pain.


crusso:
Doesn't seem to required so closing for now

PR #4812, opened 2024-12-10
feat: attach optional migration expression to `actor` (`class`) using a parenthetical `migration` field

crusso:
Support the specification of an optional migration function on the definition of an actor or actor class.

The function must consume and produce records of stable types.

Roughly:

* Fields in the domain of the migration function override any stable variables of the same name in the regular stable signature.
  This determines the pre stable signature of the actor. 
* The post stable signature is determined by the actors stable field declarations (as previously).
* Fields in the codomain/range of the migration function that also occur in the regular stable signature of the actor must be consumable (subtypes, ignoring top-level var/non-var distinctions) by the stable variables. Irrelevant, new  fields produce errors.
* Fields neither in the domain nor codomain are treated as usual and transferred from the old actor, if present and a subtype, or initialized when absent.

Dynamically, on upgrade, the (implicit or explicit) post signature of the  old actor must be compatible with the pre-signature of the new actor. This is either checked offline, statically in the classical compiler (and by Candid deserialization), and dynamically  in the EOP compiler, using a pre signature type descriptor. All stable variables in the domain of the migration function must be non-null in the deserialized or transferred stable record. The migration function is applied to the subset of these fields to produce the codomain record.  A new stable record is constructed using the values from the codomain and the values that were not overridden by the domain. Values in the domain are set to null (and will be initialized if present in the post signature).
(It would also be possible to transfer values from the domain, if not in the codomain, but  required and compatible with the post-signature, but this design choice is not implemented)

A fresh install ignores the migration code; an upgrade applies it before entering the constructor.
All variables in the codomain must be present for the migration to succeed (since there is no way to provide an initial value).

You will need to remove the migration code to do a self-upgrade.

Follow on PRs implement support for EOP (#4829) and syntactic stable signatures (#4833) (both merged here but might simplify reviewing)

# Extended stable signatures

A stable signature (stored in metadata) is now either a single interface of stable fields (as before) or (the new bit)  a dual interface recording the pre and post upgrade interface when performing migration.

The pre and post fields are implicitly identical for a singleton interface and provide for backwards compatibility and ordinary upgrades (sans explicit migration).

Stability compatibility now checks the post-signature of the old actor is compatible with the pre-signature of the new actor.

````
<stab_sig> ::=
  <typ_dec>;* actor { <stab_field>;* }
  <typ_dec>;* actor ( { <stab_field>;* },  <stab_field>;* } )
````

Done this way, there should be no need to modify dfx, which defers to moc for the compatibility check anyway.

# Example: adding record fields

To upgrade from:

``` motoko no-repl
persistent actor {
  type Card = {
    title : Text;
  };
  var map : [(Nat32, Card)] = [];
  var log : []
};
```
to *incompatible* stable type interface:

``` motoko no-repl
persistent actor {
  type Card = {
    title : Text;
    description : Text; // new field!
  };
  var map : [(Nat32, Card)] = [];
};
```

1. Define a migration module and function that transform the old stable variable, at its current type, into the new stable variable at its new type.

``` motoko no-repl
// CardMigration.mo
import Array "mo:base/Array";

module CardMigration {
  type OldCard = {
    title : Text;
  };

  type NewCard = {
    title : Text;
    description : Text;
  };

  // our migration function
  public func migration(old : {
      var map : [(Nat32, OldCard)] // old type
    }) :
    {
      var map : [(Nat32, NewCard)] // new type
    } {
    { var map : [(Nat32, NewCard)] =
        Array.map<(Nat32, OldCard), (Nat32, NewCard)>(
          old.map,
          func(key, { title }) { (key, { title; description = "<empty>" }) }) }
  }

}
```

2. Specify the migration function as the migration expression of your actor declaration:

``` motoko no-repl
import { migration } "CardMigration";

(with migration) // declare the migration function
persistent actor
  {
  type Card = {
    title : Text;
    description : Text;
  };

  var map : [(Nat32, Card)] = []; // initialized by migration on upgrade
  var log : [Text] = []; // migrated implicitly as usual

};
```

**After** we have successfully upgraded to this new version, we can also upgrade once more to a version that drops the migration code.

``` motoko no-repl
persistent actor {
  type Card = {
    title : Text;
    description : Text;
  };

  var map : [(Nat32, Card)] = [];
  var log  : [Text] = []
};
```

FUTURE:
- [ ]  Devise some mechanism to detect repeated migration with the same migration code, or a coding pattern to prevent it? 
       Some, but not all migrations might be reject as type-incompatible, but others may not be:


ggreif:
Did you consider `persistent actor A (with migration = mf) { ... };`?

test/run-drun/upgrade-migration/Migration2.mo
```diff
@@ -0,0 +1,15 @@
...
+      { var four : [var (Nat, Text)] } {
+    let new =
+    {
+      var four = old.three;
+    };
+    Prim.debugPrint(debug_show {migration = {old; new}});
+    new
+  }
+
+}
```
crusso, line 15:
```suggestion
}

```

crusso:
> Did you consider `persistent actor A (with migration = mf) { ... };`?

Not  yet, but I also don't want `A` to be in scope in mf.

Maybe

```
persistent actor (with migration = mf) A { ... };`
```
Or
```
(with migration = mf)
persistent actor  { ... };`
```
But I expect parsing hell from the last one.

Are parentheticals only mean to attach to expressions?

ggreif:
> Are parentheticals only meant to attach to expressions?

Not specifically. Any AST node can be it.

test/run-drun/upgrade-migration/Migration1.mo
```diff
@@ -0,0 +1,27 @@
...
+    { var three : [var (Nat, Text)] } {
+    let new =
+    {
+      var three = zip(old.one, old.two);
+    };
+    Prim.debugPrint(debug_show{migration={old;new}});
+    new
+  }
+
+}
```
crusso, line 27:
```suggestion
}

```

src/mo_frontend/typing.ml
```diff
@@ -2587,7 +2588,40 @@ and stable_pat pat =
...
+       match T.promote typ with
+       | T.Obj(T.Object, tfs) ->
+          if not (T.stable typ) then
+            local_error env exp.at "M0131"
+              "expected stable type, but migration expression %s non-stable type %a"
+              desc
+              display_typ_expand typ;
+          tfs
+       | _ ->
+          local_error env exp.at "M0093"
```
crusso, line 2605: Why a different error-code from above?

src/mo_types/type.ml
```diff
@@ -1860,16 +1889,19 @@ let rec match_stab_sig tfs1 tfs2 =
...
     )
 
-let string_of_stab_sig fields : string =
+let string_of_stab_sig stab_sig : string =
   let module Pretty = MakePretty(ParseableStamps) in
-  "// Version: 1.0.0\n" ^
-  Format.asprintf "@[<v 0>%a@]@\n" (fun ppf -> Pretty.pp_stab_sig ppf) fields
+  (match stab_sig with
+  | Single _ -> "// Version: 1.0.0\n"
+  | PrePost _-> "// Version: 2.0.0\n") ^
```
crusso, line 1905:
```suggestion
  | PrePost _ -> "// Version: 2.0.0\n") ^
```

src/mo_frontend/typing.ml
```diff
@@ -2580,6 +2582,92 @@ and stable_pat pat =
...
+  match exp_opt with
+  | None -> ()
+  | Some exp ->
+    Static.exp env.msgs exp; (* preclude side effects *)
+    let check_fields desc typ =
+      match typ with
+      | T.Obj(T.Object, tfs) ->
+         if not (T.stable typ) then
+           local_error env exp.at "M0201"
+             "expected stable type, but migration expression %s non-stable type%a"
```
ggreif, line 2601:
```suggestion
             "expected stable type, but migration expression %s has non-stable type%a"
```

src/mo_frontend/typing.ml
```diff
@@ -2580,6 +2582,92 @@ and stable_pat pat =
...
+             desc
+             display_typ_expand typ;
+         tfs
+      | _ ->
+         local_error env exp.at "M0202"
+           "expected object type, but migration expression %s non-object type%a"
```
ggreif, line 2607:
```suggestion
           "expected object type, but migration expression %s has non-object type%a"
```

src/mo_frontend/typing.ml
crusso: actually it was fine, reverted

src/mo_frontend/typing.ml
crusso: actually, it was fine. reverted

src/codegen/compile_classical.ml
```diff
@@ -11628,6 +11628,15 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
     StableMem.get_mem_size env ^^ BigNum.from_word64 env
 
+  | OtherPrim "rts_in_install", [] -> (* classical specific *)
+    assert (not !Flags.enhanced_orthogonal_persistence);
+    SR.Vanilla,
+    StableMem.stable64_size env ^^
+    G.i (Test (Wasm_exts.Values.I64 I64Op.Eqz)) ^^
+    G.if1 I32Type
+      (Bool.lit true)
+      (Bool.lit false)
```
ggreif, line 11638: See `Bool.vanilla_lit`
```suggestion
    G.i (Test (Wasm_exts.Values.I32 I32Op.Eqz))
```

src/mo_def/syntax.ml
```diff
@@ -235,7 +235,11 @@ and prog' = dec list
 (* Signatures (stable variables) *)
 
 type stab_sig = (stab_sig', prog_note) Source.annotated_phrase
-and stab_sig' = (dec list * typ_field list)      (* type declarations & stable actor fields *)
+and stab_sig' = (dec list * stab_body)      (* type declarations & stable actor fields *)
+and stab_body = stab_body' Source.phrase    (* type declarations & stable actor fields *)
+and stab_body' =
```
ggreif, line 240: one could have only `PrePost` and when both equal unparse it as version 1.0.0. Similarly parse that and set `pre` = `post`.

src/mo_frontend/parser.mly
```diff
@@ -993,7 +997,22 @@ stab_field :
 parse_stab_sig :
   | start ds=seplist(typ_dec, semicolon) ACTOR LCURLY sfs=seplist(stab_field, semicolon) RCURLY
     { let trivia = !triv_table in
-      fun filename -> { it = (ds, sfs); at = at $sloc; note = { filename; trivia }}
+      let sigs = Single sfs in
+      fun filename -> {
+          it = (ds, {it = sigs; at = at $sloc; note = ()});
+          at = at $sloc;
+          note =
+          { filename; trivia }}
```
ggreif, line 1005:
```suggestion
          note = { filename; trivia } }
```

src/mo_frontend/typing.ml
```diff
@@ -2580,6 +2582,92 @@ and stable_pat pat =
   | AnnotP (pat', _) -> stable_pat pat'
   | _ -> false
 
+and infer_migration env exp_opt =
+  match exp_opt with
+  | Some exp ->
+    Some (infer_exp_promote { env with async = C.NullCap; rets = None; labs = T.Env.empty } exp)
+  | None -> None
```
ggreif, line 2589: this is an `Option.map`

src/mo_frontend/typing.ml
```diff
@@ -2580,6 +2582,92 @@ and stable_pat pat =
...
+     try
+      let sort, tbs, t_dom, t_rng = T.as_func_sub T.Local 0 typ in
+      if sort <> T.Local || tbs <> [] then raise (Invalid_argument "");
+      (check_fields "consumes" (T.normalize t_dom),
+       check_fields "produces" (T.promote t_rng))
+     with Invalid_argument _ ->
+       local_error env exp.at "M0203"
+         "expected non-generic, local function type, but migration expression produces type%a"
+         display_typ_expand typ;
+       ([], [])
```
ggreif, line 2623: redundant parens 3x

src/mo_frontend/typing.ml
```diff
@@ -1035,7 +1035,7 @@ let rec is_explicit_exp e =
   | ObjE (bases, efs) ->
     List.(for_all is_explicit_exp bases
           && for_all (fun (ef : exp_field) -> is_explicit_exp ef.it.exp) efs)
-  | ObjBlockE (_, _, dfs) ->
+  | ObjBlockE (_, _e_opt, _, dfs) ->
```
ggreif, line 1038:
```suggestion
  | ObjBlockE (_, _exp_opt, _, dfs) ->
```

src/codegen/compile_enhanced.ml
```diff
@@ -11690,6 +11697,15 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
     StableMem.get_mem_size env ^^ BigNum.from_word64 env
 
+  | OtherPrim "rts_in_install", [] -> (* EOP specific *)
+    assert (!Flags.enhanced_orthogonal_persistence);
+    SR.Vanilla,
+    EnhancedOrthogonalPersistence.load_stable_actor env ^^
+    compile_test I64Op.Eqz ^^
+    E.if1 I64Type
+      (Bool.lit true)
+      (Bool.lit false)
```
ggreif, line 11707: This is basically
``` Wat
𝗂𝟨𝟦.𝖾𝗑𝗍𝖾𝗇𝖽_𝗎/𝗂𝟥𝟤 | 𝟶𝚡𝙰𝙳 | [𝗂𝟥𝟤]→[𝗂𝟨𝟦]
```
under the assumption that no runtime tagging is happening.

rts/motoko-rts/src/persistence.rs
```diff
@@ -206,6 +206,22 @@ pub unsafe fn register_stable_type<M: Memory>(
     (*metadata).stable_type.assign(mem, &new_type);
 }
 
+/// Register the stable actor type on canister initialization and upgrade.
+/// The type is stored in the persistent metadata memory for later retrieval on canister upgrades.
+/// The `new_type` value points to a blob encoding the new stable actor type.
+#[ic_mem_fn]
+pub unsafe fn assign_stable_type<M: Memory>(
```
luc-blaeser, line 213: I wonder whether it could make sense to split `register_stable_type` in two parts, checking compatibility and assigning/storing the new stable type. The logic of `assign_stable_type` is like a subset of `register_stable_type`. Just to reduce some code duplication. Not a big deal, just an idea.

src/ir_passes/async.ml
```diff
@@ -452,7 +452,7 @@ let transform prog =
          inspect = t_exp inspect;
          low_memory = t_exp low_memory;
          stable_record = t_exp stable_record;
-         stable_type = t_typ stable_type;
+         stable_type = {pre = t_typ stable_type.pre; post = t_typ stable_type.post};
```
luc-blaeser, line 455: I like this symmetry :)

src/lowering/desugar.ml
```diff
@@ -480,7 +481,7 @@ and export_runtime_information self_id =
   let scope_con2 = Cons.fresh "T2" (Abs ([], Any)) in
   let bind1  = typ_arg scope_con1 Scope scope_bound in
   let bind2 = typ_arg scope_con2 Scope scope_bound in
-  let gc_strategy =
```
luc-blaeser, line 483: Oh, my trailing white space...

src/mo_frontend/typing.ml
```diff
@@ -2580,6 +2582,92 @@ and stable_pat pat =
...
+      if sort <> T.Local || tbs <> [] then raise (Invalid_argument "");
+      (check_fields "consumes" (T.normalize t_dom),
+       check_fields "produces" (T.promote t_rng))
+     with Invalid_argument _ ->
+       local_error env exp.at "M0203"
+         "expected non-generic, local function type, but migration expression produces type%a"
+         display_typ_expand typ;
+       ([], [])
+   in
+   List.iter
```
luc-blaeser, line 2625: One thing that slightly worries me is that the co-domain/range of the migration function can be subset of the new actor's stable type which might cause accidental data loss in the following case:

Initial version:
```
persistent actor {
   var data1 : [Nat] = ...;
   var data2 : Text = ...;
}
```

Next version:
```
persistent actor [Migration.run] {
   var data1 : [Text] = ...;
   var data2 : Text = ...;
}
```
with Migration logic:
```
module {
   type OldActor = {
      data1 : [Nat];
      data2 : Text; // specifies all stable variables
   };
   type NewActor = {
      data1 : [Text];
      data2 : Text; // specifies all stable variables
   };
   type NewActor = {
      data1 : [Text];
      // accidentally forgot to mention `data2`
   }
   public void run(old: OldActor) : NewActor {
      ...
   }
}
```

With this, `data2` is accidentally lost, i.e. re-initialized on migration. This is because the variable is mentioned in domain of migration function but not in range of the migration function.

Maybe we should still require that migration function returns the full stable type of the actor, even if it would be less convenient for large actors?

luc-blaeser (dismissed):
Great work! Many thanks for this big effort.
I tested a few cases, all worked perfectly.

The only thing that I am thinking of is the accidental data loss that can happen if migration function skips a stable field in its return type, but that field is part of the input of the migration function (I left a more detailed comment above).

doc/md/examples/grammar.txt
```diff
@@ -14,7 +14,7 @@
 
 <obj_sort> ::= 
     'object'
-    'persistent'? 'actor'
+    'persistent'? 'actor' ('[' <exp> ']')?
```
luc-blaeser, line 17: I like the syntax although it might be confused with array syntax. An alternative could be `('migrate' <exp>)?` or, not sure, if we want to keep existing reserved keywords maybe`('with' <exp>)?`.

src/codegen/compile_enhanced.ml
```diff
@@ -11690,6 +11697,15 @@ and compile_prim_invocation (env : E.t) ae p es at =
     SR.Vanilla,
     StableMem.get_mem_size env ^^ BigNum.from_word64 env
 
+  | OtherPrim "rts_in_install", [] -> (* EOP specific *)
+    assert (!Flags.enhanced_orthogonal_persistence);
+    SR.Vanilla,
+    EnhancedOrthogonalPersistence.load_stable_actor env ^^
```
luc-blaeser, line 11703: This will trigger a trap because the memory is default-initialized in the runtime system on first initialization. If you run with sanity checks, it will trigger `(*metadata).stable_type.assert_initialized();` in RTS persistence. (I believe this is also the reason for CI currently failing).

test/run-drun/upgrade-migration.drun
```diff
@@ -0,0 +1,13 @@
+# SKIP ic-ref-run
```
luc-blaeser, line 1: I believe it would be good to also test all these with graph copy.

src/codegen/compile_classical.ml
crusso: But only on 32-bit, right? I'd have to extend to 64-bit  in the eop compiler. Will look.

doc/md/examples/grammar.txt
ggreif: It can be that simple:
``` Motoko
import { migration } = "Migration";

persistent actor (with migration) Frogger { ...
```

src/codegen/compile_classical.ml
ggreif: Correct, I didn't touch the 64-bit backend.

src/codegen/compile_enhanced.ml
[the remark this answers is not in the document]
crusso: I think    https://github.com/dfinity/motoko/pull/4812/commits/fad79eadfc8536b2e26b9c13588f77ddd7d83759 fixes it.

However, no I worry if this covers all the other migration paths properly or not. By the time this call is executed, will any previous data have been migrated and placed in the stable_actor field, e.g. when upgrading from 32-bit or with graph-copy?

doc/md/examples/grammar.txt
crusso: That's cute, I guess. Luc shall I try to reuse the parenthetical syntax for this too, as gabor is suggesting. It avoids introducing a new keyword since migration is just a field name (but the explanation will be more convoluted - the argument has to be record with field migration of type etc....)

src/codegen/compile_classical.ml
```diff
@@ -11628,6 +11628,15 @@ and compile_prim_invocation (env : E.t) ae p es at =
     SR.Vanilla,
     StableMem.get_mem_size env ^^ BigNum.from_word64 env
 
+  | OtherPrim "rts_in_install", [] -> (* classical specific *)
```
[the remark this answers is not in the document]
crusso: Maybe rts_has_stable_data? It's annoying there is no simple way to tell if the canister is being installed or upgraded...

src/mo_def/syntax.ml
crusso: I think its cleaner this way. PrePost means migration is involved, Single means just implicit. Also, it is not trivial to check equality.

src/mo_frontend/typing.ml
crusso: > One thing that slightly worries me is that the return type of the migration function can be super-type of the new actor's stable type which might cause accidental data loss in a case like the following

Actually, its not quite super-type: the fields of the return type must be declared stable and subtypes of the corresponding fields in the actor, but you don't have to mention every field.

doc/md/examples/grammar.txt
ggreif: Type checking is basically `par <: (Opt (Obj (Object, [{lab : "migration"; typ : Obj dom -> Obj cod }]))`, but you have to infer it first.

src/codegen/compile_classical.ml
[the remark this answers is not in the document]
crusso: renamed to rts_in_upgrade (negating result)

src/codegen/compile_classical.ml
crusso: obsolete due to refactoring.

rts/motoko-rts/src/persistence.rs
crusso: I refactor both to use a common helper instead.

src/mo_frontend/typing.ml
crusso: @luc-blaeser are you ok with just a warning?

src/codegen/compile_enhanced.ml
crusso: No longer relevant after a refactor.

src/mo_frontend/typing.ml
crusso: I actually dislike Ocamls tuple syntax, but ok...

test/run-drun/upgrade-migration.drun
crusso: Indeed. It seems to be crashing...

src/mo_frontend/typing.ml
crusso: Ok, I implemented the warning https://github.com/dfinity/motoko/pull/4812/commits/a367798cf8d2b4ced9fec42263c274da28a75612

test/run-drun/upgrade-migration.drun
crusso: Now fixed and tested

crusso:
@luc-blaeser PTAL at the migration tests and bug fix... (ie. stuff since last review).

I will now tackle the parenthetical syntax in another pr.

rts/motoko-rts/src/persistence.rs
```diff
@@ -127,6 +127,17 @@ unsafe fn use_enhanced_orthogonal_persistence() -> bool {
     }
 }
 
+/// Returns the availability of the stable actor record (false on (re-)install, true on upgrade)
+#[no_mangle]
+pub unsafe extern "C" fn has_stable_actor() -> Value {
+    let metadata = PersistentMetadata::get();
+    if (*metadata).stable_actor.forward_if_possible() == DEFAULT_VALUE {
+        Value::from_scalar(0)
```
luc-blaeser, line 135: Detail: Alternatively, one could also return a `bool`. I believe this would be more consistent to other RTS function exports, e.g. `contains_field`,

src/mo_frontend/typing.ml
luc-blaeser: Thanks a lot. Warning is good.

luc-blaeser (dismissed):
Great stuff. Let's get this merged! Thanks a lot

src/codegen/compile_enhanced.ml
```diff
@@ -10130,6 +10134,24 @@ module Persistence = struct
       end) ^^
     StableMem.region_init env
 
+  let in_upgrade env =
+    use_enhanced_orthogonal_persistence env ^^
+    (E.if1 I64Type
+      (EnhancedOrthogonalPersistence.has_stable_actor env)
```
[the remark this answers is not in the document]
crusso: Fixed

rts/motoko-rts/src/persistence.rs
crusso: Fixed

ggreif (dismissed):
Very nice! Thanks for putting in all the effort 👏

Changelog.md
```diff
@@ -1,5 +1,32 @@
...
+
+  ```
+  import Nat32 "mo:base/Nat32";
+
+  (with migration =
+    func (old : { var size : Nat32 }) : { var length : Nat } {
+      { var length = Nat32.toNat(old.size) }
+    }
+  )
+  persistent actor {
```
ggreif, line 20: Question: does this only work with `persistent actor` or also with `actor`?

Changelog.md
```diff
@@ -1,5 +1,32 @@
...
 
+
+* motoko (`moc`)
+
+* Support explicit, safe migration of persistent data allowing arbitrary
+  transformations on a selected subset of stable variables.
+  Additional static checks warn against possible data loss (#4812).
+
+  As a very simple example:
+
```
ggreif, line 11:
```suggestion
```

src/mo_frontend/parser.mly
```diff
@@ -633,6 +641,10 @@ exp_post(B) :
 exp_un(B) :
   | e=exp_post(B)
     { e }
+(*
```
ggreif, line 644: Okay to leave here, I'll deal with it when merging my stuff.

src/mo_frontend/parser.mly
```diff
@@ -691,7 +703,7 @@ exp_un(B) :
     { RetE(TupE([]) @? at $sloc) @? at $sloc }
   | RETURN e=exp(ob)
     { RetE(e) @? at $sloc }
-  | ASYNC e=exp_nest
+  | (* parenthetical_opt *) ASYNC e=exp_nest
```
ggreif, line 706: ditto

Changelog.md
ggreif: Looks like it works too.

Changelog.md
```diff
@@ -1,5 +1,32 @@
 # Motoko compiler changelog
 
+
+* motoko (`moc`)
```
ggreif, line 4:
```suggestion
## 0.14.0 (FUTURE)

* motoko (`moc`)
```

Changelog.md
crusso: With both

src/mo_frontend/parser.mly
crusso: That's why I added them, to get  a conflict ;->

rvanasa (approved):

ggreif (approved):

PR #4833, opened 2024-12-20
experiment: support migration via extended stable interfaces

crusso:
builds on #4829 and #4812, addressing syntactic stable signatures.

A stable signature is now either a single interface of stable fields (as before) or (the new bit)  a dual interface recording the pre and post upgrade interface when performing migration.

The pre and post fields are implicitly identical for a singleton interface and provide for backwards compatibility and ordinary upgrades (sans explicit migration).

Stability compatibility now checks the post interface of old module is compatible with the pre-interface of the new module.

````
<stab_sig> ::=
  <typ_dec>;* actor { <stab_field>;* }
  <typ_dec>;* actor ( { <stab_field>;* },  <stab_field>;* } )
````

Done this way, there should be no need to modify dfx, which defers to moc for the compatibility check anyway.

- [x] fig grammar.sed to delete new production
- [ ] add some tests to test/cmp

### DIFF
--- a/src/mo_types/type.ml
+++ b/src/mo_types/type.ml
@@ -1789,18 +1789,18 @@ and pp_stab_sig ppf sig_ =
     match sig_ with
     | Single tfs ->
       fprintf ppf "@[<v 2>%s{@;<0 0>%a@;<0 -2>}@]"
-       (string_of_obj_sort Actor)
-       (pp_print_list ~pp_sep:semi (pp_stab_field vs)) tfs
+        (string_of_obj_sort Actor)
+        (pp_print_list ~pp_sep:semi (pp_stab_field vs)) tfs
     | PrePost (pre, post) ->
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: src/mo_types/type.ml [hunk 1]
```diff
@@ -1860,16 +1889,17 @@ let rec match_stab_sig tfs1 tfs2 =
...
-       match_stab_sig tfs1 tfs2'
+       match_stab_fields tfs1 tfs2'
     )
 
-let string_of_stab_sig fields : string =
+let string_of_stab_sig stab_sig : string =
   let module Pretty = MakePretty(ParseableStamps) in
-  "// Version: 1.0.0\n" ^
-  Format.asprintf "@[<v 0>%a@]@\n" (fun ppf -> Pretty.pp_stab_sig ppf) fields
+  "// Version: 2.0\n" ^
```
crusso, line 1903: Refine to 2.0.0. Maybe emit 1.0.0 if just a legacy `Single` sig?

crusso: Done

PR #4841, opened 2025-01-08
chore: test recovery procedure when `finally` traps

### LINKED ISSUE #4840: Write an upgrade test for the "trap in `finally`" scenario
test the recovery recipe from https://forum.dfinity.org/t/try-finally-preventing-a-future-upgrade/39433/7

ggreif:
This PR adds a substitution variable `$PRINCIPAL` to `.drun` scripts, encoding a message argument `{ canister_id = $ID }`.
Using this information we can invoke the management canister's `start_canister`/`stop_canister` methods and with this we can implement a test for the recovery sequence detailed in #4840, to upgrade even if there are continuations in the futures table, e.g. due to trapping `finally` block.

Fixes #4840.


crusso (approved):
Nice. Thanks for testing this works!

PR #4846, opened 2025-01-14
bugfix: check that timer servicing worked

ggreif:
This fixes a disappearing timer situation described by Timo in https://dfinity.slack.com/archives/CPL67E7MX/p1736347600078339.

It turns out that under high message load the `async` timer servicing routine cannot be run. The fix is simple, check if the self-call succeeded (causes a `throw` already), and if not, set a very near global timer to retry ASAP (in the top-level `catch`).

TODO:
- [x] `catch` send errors for user workers (and mitigate) — see #4852
- [ ] document that the user thunk may be called more than once, and thus should have no side effects other than submitting the self-call — see https://github.com/dfinity/motoko-base/pull/682

### DIFF
--- a/src/ir_passes/await.ml
+++ b/src/ir_passes/await.ml
@@ -681,12 +681,13 @@ and t_on_throw context exp t_exp =
 
 and t_ignore_throw context exp = t_on_throw context exp (tupE[])
 
+(* if self-call queue full: expire global timer soon and retry *)
 and t_timer_throw context exp =
-  let check_timer_send_type = T.(Func (Local, Returns, [], [], [])) in
   t_on_throw context exp
-    (callE
-       (varE (var "@check_timer_send" check_timer_send_type)) []
-       (unitE()))
+    (blockE [expD (primE
+                (OtherPrim "global_timer_set")
+                [Mo_values.Numerics.Nat64.of_int 1 |> nat64E])]
+       (tupE[]))
 
 and t_prog (prog, flavor) =
   (t_comp_unit LabelEnv.empty prog, { flavor with has_await = false })
--- a/src/ir_passes/await.ml
+++ b/src/ir_passes/await.ml
@@ -684,7 +684,8 @@ and t_ignore_throw context exp = t_on_throw context exp (tupE[])
 (* if self-call queue full: expire global timer soon and retry *)
 and t_timer_throw context exp =
   t_on_throw context exp
-    (blockE [expD (primE
+    (blockE
+       [expD (primE
                 (OtherPrim "global_timer_set")
                 [Mo_values.Numerics.Nat64.of_int 1 |> nat64E])]
        (tupE[]))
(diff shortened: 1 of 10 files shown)


src/prelude/internals.mo
```diff
@@ -675,5 +675,13 @@ func @cancelTimer(id : Nat) {
   }
 };
 
-
 func @set_global_timer(time : Nat64) = ignore (prim "global_timer_set" : Nat64 -> Nat64) time;
+
+// Function called by backend to check if self-send worked.
+// DO NOT RENAME without modifying compilation.
+func @check_timer_send() {
+  if (not @call_succeeded()) {
```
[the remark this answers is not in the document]
ggreif: Yes: c9b442ad760e75803d0f5b9c7622c096d9d5e3b3

src/mo_frontend/typing.mli
```diff
@@ -13,3 +13,4 @@ val check_actors : ?viper_mode:bool -> ?check_actors:bool -> scope -> Syntax.pro
 val check_stab_sig : scope -> Syntax.stab_sig -> (field list) Diag.result
 
 val heartbeat_type : typ
+val global_timer_set_type : typ
```
crusso, line 16: Maybe these should live in Type.ml? I think there are other well-known types in there.

src/ir_passes/await.ml
```diff
@@ -671,14 +671,23 @@ and t_ignore_throw context exp =
...
           or timer won't be initialized in compile.ml *)
        with at = exp.at
      }
 
+and t_ignore_throw context exp = t_on_throw context exp (tupE[])
+
+(* if self-call queue full: expire global timer soon and retry *)
+and t_timer_throw context exp =
+  t_on_throw context exp
+    (blockE [expD (primE
```
crusso, line 687: Are we sure there aren't other reasons for the error than queue full?  If so, it might make make sense to test the error code of the error and only set the global timer when appropriate...

ggreif: Checking the Wasm, the only possibility is send failure.

src/mo_frontend/typing.mli
ggreif: I tend to agree...

crusso (commented):
LGTM so far. What will you do about failure to enqueue task?

ggreif:
> LGTM so far. What will you do about failure to enqueue tasks?

A PR on top of this one and with the logic to insert the failed expirations into the priority queue's head (in order).

src/mo_frontend/typing.mli
ggreif: Done: 980cd23c8

berestovskyy (dismissed):
Seems it fixes the issue, I can't reproduce it anymore on my branch https://github.com/dfinity/ic/commits/andriy/motoko-timers-repro/

crusso (approved):
LGMT

PR #4849, opened 2025-01-15
Support IC Low Wasm Memory Hook

luc-blaeser:
## Low memory hook

The IC allows to implement a low memory hook, which is a warning trigger when main memory is becoming scarce.

For this purpose, a Motoko actor or actor class instance can implement the system function `onLowMemory()`. This system function is scheduled when canister's free main memory space has fallen below the defined threshold `wasm_memory_threshold`, that is is part of the canister settings. In Motoko, `onLowMemory()` implements the `canister_on_low_wasm_memory` hook defined in the IC specification.

Reference: https://github.com/dfinity/portal/pull/3761

Example of implementing the low memory hook:
```
actor {
    system func onLowMemory() : async () {
        Debug.print("Low memory!");
    }
}
```

The following properties should be considered when using the low memory hook:
* The execution of `onLowMemory` happens with a certain delay, as it is scheduled as a separate asynchronous message that runs after the message in which the threshold was crossed.
* Once executed, `onLowMemory` will only be triggered again when the main memory free space grows above the threshold (e.g. by lowering the threshold or shrinking the main memory through canister reinstallation) and then again falls below the threshold.
* Traps or unhandled errors in `onLowMemory` are ignored. They only revert the changes done in `onLowMemory`.
* Due to its `async` return type, the `onLowMemory` function may send further messages and await results.

### DIFF
--- a/doc/md/canister-maintenance/memory.md
+++ b/doc/md/canister-maintenance/memory.md
@@ -2,7 +2,7 @@
 sidebar_position: 5
 ---
 
-# Memory management
+# Memory diagnostics
 
 ## Low memory hook
 
--- a/doc/md/canister-maintenance/memory.md
+++ b/doc/md/canister-maintenance/memory.md
@@ -1,0 +1,26 @@
+---
+sidebar_position: 5
+---
+
+# Memory management
+
+## Low memory hook
+
+The IC allows to implement a low memory hook, which is a warning trigger when main memory is becoming scarce.
+
+For this purpose, a Motoko actor or actor class instance can implement the system function `onLowMemory()`. This system function is scheduled when canister's free main memory space has fallen below the defined threshold `wasm_memory_threshold`, that is is part of the canister settings. In Motoko, `onLowMemory()` implements the `canister_on_low_wasm_memory` hook defined in the IC specification.
+
+Example of using the low memory hook:
+```
+actor {
+    system func onLowMemory() : async () {
+        Debug.print("Low memory!");
+    }
+}
+```
+
+A few aspects need to be considered with the low memory hook:
+* The execution of `onLowMemory` happens with a certain delay, as it is scheduled as separate asynchronous message that runs after the message in which the threshold was crossed.
+* Once executed, `onLowMemory` is only triggered again when the main memory free space went above the threshold (e.g. by lowering the threshold or shrinking the main memory through canister reinstallation) and when the free space again fell below the threshold.
+* Traps or unhandled errors in `onLowMemory` are ignored. They only revert the changes done in `onLowMemory`.
+* Due to its `async` return type, the `onLowMemory` function may send further messages and await results.
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -394,12 +394,16 @@ let timer_type =
     [Func (Local, Returns, [], [Prim Nat64], [])],
     [Async (Fut, Var (default_scope_var, 0), unit)]))
 
+let low_memory_type =
+  T.(Func (Local, Returns, [scope_bind], [], [Async (Fut, Var (default_scope_var, 0), unit)]))
+    
 let system_funcs tfs =
   [
     ("heartbeat", heartbeat_type);
     ("timer", timer_type);
     T.("preupgrade", Func (Local, Returns, [scope_bind], [], []));
     T.("postupgrade", Func (Local, Returns, [scope_bind], [], []));
+    ("onLowMemory", low_memory_type);
     ("inspect",
      (let msg_typ = T.decode_msg_typ tfs in
       let record_typ =
--- a/test/run-drun/low-memory.mo
+++ b/test/run-drun/low-memory.mo
@@ -1,0 +1,63 @@
+import Prim "mo:⛔";
+import LowMemoryActor "low-memory/low-memory-actor";
+import Cycles = "cycles/cycles";
+
+actor Self {
+  type canister_settings = {
+    wasm_memory_threshold: ?Nat;
+  };
+
+  func setMemoryThreshold(a : actor {}, threshold : Nat) : async () {
+    let ic00 = actor "aaaaa-aa" : actor {
+      update_settings : shared {
+        canister_id : Principal;
+        settings : canister_settings;
+      } -> async ();
+    };
+    let settings : canister_settings = {
+        wasm_memory_threshold = ?threshold;
+    };
+    await ic00.update_settings({
+      canister_id = Prim.principalOfActor(a);
+      settings });
+  };
+
+  let kb = 1024;
+  let mb = 1024 * kb;
+  let gb = 1024 * mb;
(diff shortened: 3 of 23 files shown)


src/mo_frontend/typing.ml
```diff
@@ -394,12 +394,16 @@ let timer_type =
     [Func (Local, Returns, [], [Prim Nat64], [])],
     [Async (Fut, Var (default_scope_var, 0), unit)]))
 
+let low_memory_type =
```
ggreif, line 397: Just a heads-up, these type definitions will live in `type.ml` after #4846. I'll take care of this in the migration.

test/fail/ok/M0129.tc.ok
```diff
@@ -1 +1 @@
-M0129.mo:2.15-2.21: type error [M0129], unexpected system method named foobar, expected heartbeat or timer or preupgrade or postupgrade or inspect
+M0129.mo:2.15-2.21: type error [M0129], unexpected system method named foobar, expected heartbeat or timer or preupgrade or postupgrade or onLowMemory or inspect
```
ggreif, line 1: `onLowMemory` is kind-of clumsy, what about `tightmemory`? Would be more in line with the other names...

luc-blaeser: Yes, I see. In IC it is called `canister_on_low_wasm_memory`... Just thought to use a similar name for resemblance, but I am not fully happy either.
Maybe @crusso has some suggestion on how to name this.

crusso: I think I prefer `onLowMemory` to `tightmemory`. I guess just `memory` if we want to keep it short and sweet like the others (https://internetcomputer.org/docs/current/motoko/main/reference/language-manual#system-fields), but onLowMemory is more meaningfull.

doc/md/canister-maintenance/memory.md
```diff
@@ -0,0 +1,26 @@
...
+actor {
+    system func onLowMemory() : async () {
+        Debug.print("Low memory!");
+    }
+}
+```
+
+The following properties apply to the low memory hook:
+* The execution of `onLowMemory` happens with a certain delay, as it is scheduled as a separate asynchronous message that runs after the message in which the threshold was crossed.
+* Once executed, `onLowMemory` is only triggered again when the main memory free space went above the threshold (e.g. by lowering the threshold or shrinking the main memory through canister reinstallation) and when the free space again fell below the threshold.
```
crusso, line 24:
```suggestion
* Once executed, `onLowMemory` is only triggered again when the main memory free space exceeds the threshold (e.g. by lowering the threshold or shrinking the main memory through canister reinstallation) and when the free space again falls below the threshold.
```
Not sure I understand this... should that be an or?

doc/md/canister-maintenance/memory.md
```diff
@@ -0,0 +1,26 @@
...
+* Traps or unhandled errors in `onLowMemory` are ignored. They only revert the changes done in `onLowMemory`.
```
crusso, line 25:
```suggestion
* Traps or unhandled errors in `onLowMemory` are ignored. Traps only revert the changes done in `onLowMemory`.
```

test/run-drun/low-memory.mo
```diff
@@ -0,0 +1,60 @@
...
+
+  public shared func run() : async () {
+    Cycles.add<system>(2_000_000_000_000);
+    let lowMemoryActor = await LowMemoryActor.LowMemoryActor(lowMemoryCallback);
+    
+    let threshold1 = maxMemory - (await lowMemoryActor.memorySize()) - mb : Nat;
+    await setMemoryThreshold(lowMemoryActor, threshold1);
+    await lowMemoryActor.allocateMemory();
+
+    // Not yet implemented on IC: Should retrigger when lowering threshold.
```
crusso, line 42: Oh dear.

crusso (approved):
I guess we could consider allowing the return type to be async* but the method would need to force the async* somehow. That would avoid another scheduling hop.

Do we want merge this now or wait for the IC implementation to be fixed?

test/fail/ok/M0129.tc.ok
ggreif: What about `lowmemory`?

doc/md/canister-maintenance/memory.md
luc-blaeser: I will clarify. The idea is that it first exceeds and then falls below.

test/fail/ok/M0129.tc.ok
luc-blaeser: Will chane to `lowmemory`

test/run-drun/low-memory.mo
luc-blaeser: I reported as these as issues

doc/md/canister-maintenance/memory.md
luc-blaeser: I adjusted it. Hopefully it is clearer now.

luc-blaeser:
> I guess we could consider allowing the return type to be async* but the method would need to force the async* somehow. That would avoid another scheduling hop.
> 
> Do we want merge this now or wait for the IC implementation to be fixed?

I changed it to `async*`. Thanks for the suggestion.

I think we can merge it as soon as it is ready from our side. The IC fix can come later and would be automatically become effective.

test/fail/ok/M0129.tc.ok
ggreif: Great. Please update the PR description too. (Also w.r.t. the delay problem.)

doc/md/canister-maintenance/memory.md
```diff
@@ -0,0 +1,26 @@
...
+        Debug.print("Low memory!");
+    }
+}
+```
+
+The following properties apply to the low memory hook:
+* The execution of `lowmemory` happens with a certain delay, as it is scheduled as a separate asynchronous message that runs after the message in which the threshold was crossed.
+* Once executed, `lowmemory` is only triggered again when the main memory free space first exceeds and then falls below the threshold.
+* Traps or unhandled errors in `lowmemory` are ignored. Traps only revert the changes done in `lowmemory`.
+* Due to its `async` return type, the `lowmemory` function may send further messages and await results.
```
ggreif, line 26:
```suggestion
* Due to its `async*` return type, the `lowmemory` function may send further messages and `await` results.
```

ggreif (dismissed):
Added one suggestion, otherwise LGTM!

doc/md/canister-maintenance/memory.md
luc-blaeser: Thank you!

luc-blaeser:
Thanks for reviewing!

ggreif (approved):

ggreif (approved):

ggreif (approved):

crusso (approved):
LGMT

PR #4851, opened 2025-01-15
bugfix: broken eop memory compatibility check

crusso:
The eop memory compatibility check was slightly broken and would try to compare `memory` fields with different tags.

Consider:

v0.mo
```
persistent actor {

  var four : [var (Nat, Text)] = [var];
  var zero = 0;

}
```
v1.mo
```
persistent actor {

  var three : [var (Nat, Text)] = [var];
  var zero = 0;

}
```
this upgrade should be allowed because `four` is dropped and `three` is added, while `zero` is preserved, but eop would fail with memory compatibility error.

The search for the next tag in the subtype would assume that the next tag found was less or equal when, in fact, it could be a higher tag in the case the field is absent on the left, which is allowed for memory compatibility.

Needs careful review. 

Maybe it would be safer to adjust the code to have two cases, one for main_actor, and another for !main_actor.


test/run-drun/tmp_upgrade/version1.mo
```diff
@@ -0,0 +1,6 @@
+persistent actor {
+
+  var three : [var (Nat, Text)] = [var];
+  var zero = 0;
+
+}
```
crusso, line 6:
```suggestion
}

```

luc-blaeser (approved):
Thanks a lot for having found this bug and for this fix!

crusso:
Superseded

PR #4852, opened 2025-01-16
Mitigate timer job submission failures

ggreif:
This deals with the (unlikely) possibility that the send queue is not full when the timer servicing action is submitted, but becomes full while submitting the user jobs. Now we catch the failure and re-add (single-expiration) jobs to the start of the priority queue.

This is the missing piece to #4846.

This is an incremental change, so that we don't have to touch the happy path. A rewrite would be justified to collapse gathering and self-sends.

There is an optimisation realised in `@prune`, the same could be done in `gatherExpired(n.post)` with a slight restructuring of the conditions:
``` Motoko
      if (n.expire[0] <= now and gathered < thunks.size()) {
        if (n.expire[0] > 0) {
          ...
        };
        gatherExpired(n.post)
      }
```

### DIFF
--- a/src/prelude/internals.mo
+++ b/src/prelude/internals.mo
@@ -614,7 +614,10 @@ func @timer_helper() : async () {
 
   var failed : Nat64 = 0;
   func reinsert(job : () -> async ()) {
-    if (failed == 0) @timers := @prune @timers;
+    if (failed == 0) {
+      @timers := @prune @timers;
+      ignore (prim "global_timer_set" : Nat64 -> Nat64) 1
+    };
     failed += 1;
     @timers := ?(switch @timers {
       case (?{ id = 0; pre; post; job = j; expire; delay })


src/prelude/internals.mo
```diff
@@ -612,9 +612,22 @@ func @timer_helper() : async () {
   ignore (prim "global_timer_set" : Nat64 -> Nat64) exp;
   if (exp == 0) @timers := null;
 
+  var failed : Nat64 = 0;
+  func reinsert(job : () -> async ()) {
+    if (failed == 0) @timers := @prune @timers;
```
crusso, line 617: is this ìf`supposed to fall through to next code?

src/prelude/internals.mo
```diff
@@ -612,9 +612,22 @@ func @timer_helper() : async () {
...
+    failed += 1;
+    @timers := ?(switch @timers {
+      case (?{ id = 0; pre; post; job = j; expire; delay })
```
crusso, line 620: why does this match on id =  0?

src/prelude/internals.mo
ggreif: The PQ needs to be pruned of expunged expirations only once.

But this is also the place where we should set the global timer to 1.

src/prelude/internals.mo
ggreif: Only already `reinsert`-ed nodes (special `id` = 0) should be delegated into the `pre` part of the PQ.

src/prelude/internals.mo
```diff
@@ -544,7 +544,7 @@ func @prune(n : ?@Node) : ?@Node = switch n {
     if (n.expire[0] == 0) {
       @prune(n.post) // by corollary
     } else {
-      ?{ n with pre = @prune(n.pre); post = @prune(n.post) }
+      ?{ n with pre = @prune(n.pre) }
```
ggreif, line 547: rationale: current node is not expunged, so the `post` portion won't contain any either

src/prelude/internals.mo
ggreif: This tree rotation maintains the original expiration order.

src/prelude/internals.mo
```diff
@@ -612,9 +612,25 @@ func @timer_helper() : async () {
   ignore (prim "global_timer_set" : Nat64 -> Nat64) exp;
   if (exp == 0) @timers := null;
 
+  var failed : Nat64 = 0;
+  func reinsert(job : () -> async ()) {
+    if (failed == 0) {
+      @timers := @prune @timers;
+      ignore (prim "global_timer_set" : Nat64 -> Nat64) 1
```
ggreif, line 619: At this point there are no `id` = 0 nodes in the PQ. They were all in the front and have been gathered.

src/prelude/internals.mo
ggreif: Done in 321786aee3ccfe910c42b16c02c7c48030c91e9c.

crusso (approved):
LGTM - thanks for the offline explanation

PR #4853, opened 2025-01-16
chore: refactor memory compatibility check for clarity

crusso:


test/run-drun/tmp_upgrade/version0.mo
```diff
@@ -0,0 +1,6 @@
+persistent actor {
+
+  var four : [var (Nat, Text)] = [var];
+  var zero = 0;
+
+}
```
crusso, line 6:
```suggestion
}

```

PR #4854, opened 2025-01-16
fix: Use lower case for Wasm persistence modes

luc-blaeser:
Related to https://github.com/dfinity/ic/pull/3479.

This changes the variant names for the Wasm memory persistence modes used for enhanced orthogonal persistence to lower case to align it with the IC specification, i.e. using `keep` and `replace` instead of `Keep` and `Replace`.

For installed Motoko programs using enhanced orthogonal persistence, the change only affects the programmatic upgrades of Motoko actor class instances. Existing such programs would need to be recompiled with a new Motoko compiler and upgraded. The IC detects mismatching variant names of Wasm memory persistence by raising an error while still preserving persistent memory.


crusso:
Might have been easier to just change the spec to match the implementation ;->

luc-blaeser:
> Might have been easier to just change the spec to match the implementation ;->

Yes, I was also tempted. But better going this more painful path to keep IC specification more consistent (reducing "technical debt").

crusso (approved):

crusso:
LGTM.
Should we mark it DO_NOT_MERGE for now?

luc-blaeser:
We can merge. It is supported in next IC release.

ggreif (approved):
LGTM!

PR #4855, opened 2025-01-16
experiment: eop compatibility check rewrite

crusso:
* Use dual of variant subtyping for (nested) records
* Rewrite memory subtyping to pointwise subtyping of common fields

Fixed. I hope


luc-blaeser (approved):
Nice refactoring. Thanks for implementing this!

PR #4868, opened 2025-01-27
feat: passing cycles to `call_raw`

ggreif:
This was not possible before, as the attached cycles would go to the self-call, unutilised. Now we use the same logic as with actor class instantiation and delegate them in `@call_raw`.


crusso (approved):
Good catch!

PR #4869, opened 2025-01-27
better grammar for parentheticals

ggreif:
This syntax won't allow repetition of parentheticals and only allow it on calls (sadly some duplication happens here) and `async` expressions. But I still think this is an improvement!


crusso (approved):
This looks much cleaner I think. Also avoid the issue of preserving any side-effects when attached to other nodes (the previous PR was just eliding them, IIRC)

PR #4874, opened 2025-01-31
experiment: migration expressions using parenthetical syntax.

crusso:
Bit ugly for classes, e.g.:

actor (with migration = Migration.run) class C()...

 Will experiment if this works out instead:

actor class 
  (with migration = Migration.run) 
  C()...


test/run-drun/upgrade-migration/version4.mo
```diff
@@ -2,7 +2,7 @@ import Prim "mo:prim";
 import Migration "Migration4";
```
ggreif, line 2:
```suggestion
import { run = migration } "Migration4";
```

test/run-drun/upgrade-migration/version4.mo
```diff
@@ -2,7 +2,7 @@ import Prim "mo:prim";
...
 
 // test adding a nested field, changing type
-actor [Migration.run] {
+actor (with migration = Migration.run) {
```
ggreif, line 5:
```suggestion
actor (with migration) {
```
exercise the punning way

crusso: I added a separate test for your idiom.

test/run-drun/upgrade-migration/version4.mo
crusso: I added a separate test for your idiom.

PR #4876, opened 2025-01-31
feat: add `.values()` alias for `.vals()`

rvanasa:
Makes it possible to write `x.values()` in place of `x.vals()` for an array or `Blob`. This is motivated by the upcoming base library changes in [this repository](https://github.com/dfinity/new-motoko-base).

`values()` is used in place of `vals()` for a large number of popular PLs (JS, Python, Java, C#, Rust, Swift, Kotlin, etc.), so this change would enable developers and LLMs to better transfer knowledge from other languages.

**Progress:**

* [x] Add alias in type checker
* [x] Update documentation and examples
* [x] Update unit tests to cover both `.vals()` and `.values()`


ggreif (dismissed):
LGTM, but make sure that we have at least one test exercising the original `vals` accessor... See also #4882.

Changelog.md
```diff
@@ -4,27 +4,29 @@
 
 * motoko (`moc`)
 
-* Support explicit, safe migration of persistent data allowing arbitrary
-  transformations on a selected subset of stable variables.
-  Additional static checks warn against possible data loss (#4812).
+  * Add `.values()` as an alias to `.vals()` for Arrays and Blobs.
```
ggreif, line 7:
```suggestion
  * Add `.values()` as an alias to `.vals()` for arrays and `Blob`s.
```

ggreif (approved):
Looks tike all mechanic changes, tests look good.

PR #4845, opened 2025-01-14
doc: migration expressions

crusso:
Builds on #4812.

Document migration expressions.

- [x] Extract examples to checkable files.
- [x] Review 

The best way to review this is to do `make preview` in motoko\doc\ so you can see the inlined examples...

PR #4879, opened 2025-02-01
doc: migration with prefixed parentheticals

crusso:
Revision of #4845 to adjust to new parenthetical syntax


luc-blaeser (approved):
Very nice. Thanks a lot.

PR #4890, opened 2025-02-05
chore: backend tweaks

ggreif:
Broken out of #4608. Mostly harmless cleanups.

### DIFF
--- a/src/codegen/compile_classical.ml
+++ b/src/codegen/compile_classical.ml
@@ -9197,10 +9197,10 @@ module VarEnv = struct
   let add_local_with_heap_ind env (ae : t) name typ =
       let i = E.add_anon_local env I32Type in
       E.add_local_name env i name;
-      ({ ae with vars = NameEnv.add name ((HeapInd i), typ) ae.vars }, i)
+      ({ ae with vars = NameEnv.add name (HeapInd i, typ) ae.vars }, i)
 
   let add_local_heap_static (ae : t) name ptr typ =
-      { ae with vars = NameEnv.add name ((HeapStatic ptr), typ) ae.vars }
+      { ae with vars = NameEnv.add name (HeapStatic ptr, typ) ae.vars }
 
   let add_local_public_method (ae : t) name (fi, exported_name) typ =
       { ae with vars = NameEnv.add name ((PublicMethod (fi, exported_name) : varloc), typ) ae.vars }
@@ -9209,7 +9209,7 @@ module VarEnv = struct
       { ae with vars = NameEnv.add name ((Const cv : varloc), typ) ae.vars }
 
   let add_local_local env (ae : t) name sr i typ =
-      { ae with vars = NameEnv.add name ((Local (sr, i)), typ) ae.vars }
+      { ae with vars = NameEnv.add name (Local (sr, i), typ) ae.vars }
 
   let add_direct_local env (ae : t) name sr typ =
       let i = E.add_anon_local env (SR.to_var_type sr) in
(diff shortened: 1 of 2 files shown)


src/codegen/compile_classical.ml
```diff
@@ -2540,7 +2540,7 @@ module Closure = struct
       I32Type :: Lib.List.make n_args I32Type,
       FakeMultiVal.ty (Lib.List.make n_res I32Type))) in
     (* get the table index *)
-    Tagged.load_forwarding_pointer env ^^
+    (*Tagged.load_forwarding_pointer env ^^ FIXME: NOT needed, accessing immut slots*)
```
crusso, line 2543: Did @luc-blaeser approve this? Seems a bit risky to me.

ggreif, line 2543: there can never be a mutation of the `funptr_field` between old space and newspace. Maybe we should have `Tagged.load_immutable_field` (= `Tagged.load_field`) to signify that `load_forwarding_pointer` is not needed.

src/codegen/compile_classical.ml
```diff
@@ -5520,7 +5520,7 @@ module IC = struct
     | Flags.(ICMode | RefMode) ->
       system_call env "call_cycles_add128"
     | _ ->
-      E.trap_with env "cannot accept cycles when running locally"
+      E.trap_with env "cannot add cycles when running locally"
```
ggreif, line 5523: copy&paste error

src/codegen/compile_classical.ml
```diff
@@ -10937,9 +10936,8 @@ and compile_prim_invocation (env : E.t) ae p es at =
 
          StackRep.of_arity return_arity,
          code1 ^^ StackRep.adjust env fun_sr SR.Vanilla ^^
+         Closure.prepare_closure_call env ^^ (* FIXME: move to front elsewhere too *)
```
ggreif, line 10939: Some fields (that could have mutable portions) could be in newspace, so let's do this eagerly. TODO: spot other similar uses...

N.B. Line 10943 potentially looked into oldspace, but probably only loaded an immutable field (function index?).

src/codegen/compile_classical.ml
crusso: Hmm, maybe that's ok because its the fun ptr field which is just an integer. Indeed, forward this would actually be wrong unless we represent those as heap object too now.

src/codegen/compile_classical.ml
crusso, line 10939: So this now storing the prepared (forwarded closure in clos and re_loading it later to pass to call_closure (already forwarded). I guess that's ok. @luc-blaeser ?

src/codegen/compile_classical.ml
ggreif: I don't know how Luc wants to implement named (shared) functions, maybe then the index changes? Better ask!

src/codegen/compile_enhanced.ml
```diff
@@ -2332,7 +2330,7 @@ module Closure = struct
       I64Type :: Lib.List.make n_args I64Type,
       FakeMultiVal.ty (Lib.List.make n_res I64Type))) in
     (* get the table index *)
-    Tagged.load_forwarding_pointer env ^^
+    (*Tagged.load_forwarding_pointer env ^^ FIXME: NOT needed, accessing immut slots*)
```
luc-blaeser, line 2333: Targets of immutable slots may also be moved by the GC.

src/codegen/compile_classical.ml
luc-blaeser: Please note that targets of immutable slots can also be moved by GC, so this  extra step is necessary I believe

PR #4891, opened 2025-02-05
doc: `make doc` without requiring nix magic

crusso:
Add a new make target that builds a documentation preview outside of nix.

Might require installation of `make`, `node` and `npm` if not available.

```
cd doc
make doc
```
 
or
 

`make -C doc doc`


rvanasa (dismissed):
LGTM

rvanasa (approved):

PR #4895, opened 2025-02-10
backend: lazy closure restore

ggreif:
This gives us a less intrusive way of closure variable restoration. Instead of
``` Wasm
# function prelude
  local.get $clos
  i32.load offs_a
  local.set $a
  ...
  local.get $a
  # use $a
```
one has
``` Wasm
# empty prelude
  ...
  local.get $clos
  i32.load offs_a
  # use it instead of $a
```
Some functions _only_ use closed-over variables to pass them on to other closures! So it makes total sense to avoid pulling them into locals. Such functions arise from (non-inlined) local-`async` functions.

_Note_: Right now this change is enabled globally, so I expect that the code size will increase due to closure variables used more than once. When we'll do this selectively, I expect a reduction in code size.


src/codegen/compile_classical.ml
```diff
@@ -9125,6 +9125,9 @@ module VarEnv = struct
        in the given stackrep (Vanilla, UnboxedWord32, …) so far
        Used for immutable and mutable, non-captured data *)
     | Local of SR.t * int32
+    (* An already captured binding from the caller's frame
+       Used for immutable and mutable, non-captured data *)
```
ggreif, line 9129:
```suggestion
       Used for immutable and mutable data *)
```

src/codegen/compile_enhanced.ml
```diff
@@ -8971,6 +8971,9 @@ module VarEnv = struct
        in the given stackrep (Vanilla, UnboxedWord64, …) so far
        Used for immutable and mutable, non-captured data *)
     | Local of SR.t * int32
+    (* An already captured binding from the caller's frame
+       Used for immutable and mutable, non-captured data *)
```
ggreif, line 8975:
```suggestion
       Used for immutable and mutable data *)
```

ggreif:
> This might also reduce stack pressure - but won't it require repeated forwarding in the incremental gc? @luc-blaeser

Definitely, it's a tradeoff. I expect it to be a total win when we just wrap an inner function (we can detect this in `freevars) and use all closure vars linearly. A good example is CPS-converted `func() : async () { ... }`, where the `async` block boils down to a function call.

PR #4900, opened 2025-02-14
refactor: parentheticals KISS

ggreif:
## KISS: keep it simple, stupid

This refactors the lowering of parentheticals for sends to simply assign to backend variables. So we piggy-back on the imperative mechanism instead of running the parenthetical through all the IR and backend hoops.

Maybe later we have to remove this shortcut, but for now it should work.

Thanks to @crusso for suggesting!


src/lowering/desugar.ml
```diff
@@ -258,6 +270,17 @@ and exp' at note = function
       { it = I.LetD ({it = I.WildP; at = e.at; note = T.Any}, exp e);
         at = e.at; note = ()}], (unitE ()))
 
+and distill_meta par =
+  let cycles =
+    if T.(sub par.note.note_typ (Obj (Object, [{ lab = "cycles"; typ = nat; src = empty_src }])))
+    then [dotE (exp par) "cycles" T.nat |> assignVarE "@cycles" |> expD]
```
ggreif, line 276: `par` is evaluated twice!

src/lowering/desugar.ml
```diff
@@ -1103,12 +1126,12 @@ and to_args typ po exp_opt p : Ir.arg list * Ir.exp option * (Ir.exp -> Ir.exp)
   let wrap_under_async e =
     if T.is_shared_sort sort
     then match control, e.it with
-      | (T.Promises, Ir.AsyncE (par, s, tb, e', t)) ->
-        { e with it = Ir.AsyncE (par, s, tb, wrap_exp_opt e', t) }
+      | (T.Promises, Ir.AsyncE ((*par, *)s, tb, e', t)) ->
```
ggreif, line 1129: clean these up!

src/mo_frontend/typing.ml
```diff
@@ -2584,8 +2584,8 @@ and validate_parenthetical env typ_opt = function
      | _ -> ()
      end;
      let [@warning "-8"] par_infer env { it = ObjE (bases, fields); _ } =
-       let checked = T.[ { lab = "cycles"; typ = nat; src = empty_src}
-                       ; { lab = "timeout"; typ = nat32; src = empty_src}
+       let checked = T.[ { lab = "cycles"; typ = nat; src = empty_src }
```
ggreif, line 2587: move to `type.ml`

src/lowering/desugar.ml
ggreif: Fixed!

src/lowering/desugar.ml
```diff
@@ -258,6 +270,17 @@ and exp' at note = function
...
 
+and distill_meta par =
+  let cycles =
+    if T.(sub par.note.note_typ (Obj (Object, [{ lab = "cycles"; typ = nat; src = empty_src }])))
+    then [dotE (exp par) "cycles" T.nat |> assignVarE "@cycles" |> expD]
+    else [natE Mo_values.Numerics.Nat.zero |> assignVarE "@cycles" |> expD] in
+  let timeout =
+    if T.(sub par.note.note_typ (Obj (Object, [{ lab = "timeout"; typ = nat32; src = empty_src }])))
+    then [dotE (exp par) "timeout" T.nat32 |> optE |> assignVarE "@timeout" |> expD]
+    else [] in
```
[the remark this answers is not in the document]
ggreif: Done!

src/ir_def/construct.ml
```diff
@@ -271,7 +268,7 @@ let blockE decs exp =
 let nat32E n =
   { it = LitE (Nat32Lit n);
     at = no_region;
-    note = Note.{ def with typ = T.nat32 }
+    note = Note.{ def with typ = T.(Prim Nat32) }
```
crusso, line 271:
```suggestion
    note = Note.{ def with typ = T.nat32) }
```

src/lowering/desugar.ml
ggreif: done: 5ef48945d

crusso (approved):
Skimmed and looks ok, but I'll leave the proper review for the merge.

Also, do we really want to reset the @cycles to zero in desugar.ml when the cycles field is absent. At least, do we want to this before we've actually removed ExperimentalCycles.add altogether?

ggreif:
> Also, do we really want to reset the @cycles to zero in desugar.ml when the cycles field is absent. At least, do we want to this before we've actually removed ExperimentalCycles.add altogether?

Yes, I think we should not have a mixture of both approaches. Either have a parenthetical _or_ `Cycles.add`. Makes it simpler to get rid of the latter when we deprecate it.

crusso:
Oh, ok, so the old mechanism still works on sends without parenthetical, right?

PR #4913, opened 2025-02-20
experiment: strip rts name section for reproducible builds

crusso:
Seems like most of the differences on mac and linux are just in the name section of the rts, apart from 2 functions that have swaped indices (to do with format! of floats, AFAICT)

### DIFF
--- a/src/pipeline/pipeline.ml
+++ b/src/pipeline/pipeline.ml
@@ -699,6 +699,7 @@ let load_as_rts () =
     | _ -> assert false
   in
   let rts = Wasm_exts.CustomModuleDecode.decode "rts.wasm" (Lazy.force rts) in
+  (* strip name section for reproducible build *)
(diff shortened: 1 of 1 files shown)


src/pipeline/pipeline.ml
```diff
@@ -698,7 +698,8 @@ let load_as_rts () =
     | (false, true, Flags.Incremental) -> Rts.wasm_incremental_debug
     | _ -> assert false
   in
-  Wasm_exts.CustomModuleDecode.decode "rts.wasm" (Lazy.force rts)
+  let rts = Wasm_exts.CustomModuleDecode.decode "rts.wasm" (Lazy.force rts) in
```
crusso, line 701:
```suggestion
  let rts = Wasm_exts.CustomModuleDecode.decode "rts.wasm" (Lazy.force rts) in
  (* strip name section for reproducible build *)
```

PR #4926, opened 2025-02-27
fix: change `M0061` and `M0062` from warnings to errors

rvanasa:
This PR promotes `M0061` ("comparing abstract type to itself at supertype") and `M0062` ("comparing incompatible types at common supertype") from warnings to errors. 

This is motivated by the unintuitive behavior of comparing at supertype `Any`, which always returns `true`. For instance, comparing `5` and `?4` returns `true` and therefore can result in buggy behavior or false positive test cases.

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -1279,12 +1279,12 @@ and infer_exp'' env exp : T.typ =
         error_bin_op env exp.at t1 t2;
       if not (eq env exp1.at t t1 || eq env exp2.at t t2) && not (sub env exp1.at T.nat t1 && sub env exp2.at T.nat t2) then
         if eq env exp.at t1 t2 then
-          warn env exp.at "M0061"
+          error env exp.at "M0061"
             "comparing abstract type%a\nto itself at supertype%a"
             display_typ_expand t1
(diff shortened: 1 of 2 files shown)


crusso:
I think this need some more thought. I think some use case might still be sensible but not sure.

rvanasa:
Here is a counterexample where a developer might want to be able to compare with a supertype:

```motoko
type A = { x : Int; y : Int};
type B = { x : Int; z : Int};

let a : A = { x = 0; y = 1 };
let b : B = { x = 1; z = 2 };

a == b // => false, supertype warning
```

Output:

```
false : Bool
mo:7.1-7.7: warning [M0062], comparing incompatible types
  {x : Int; y : Int}
and
  {x : Int; z : Int}
at common supertype
  {x : Int}
```

Maybe we could remove the warning or at least have the same behavior as comparing with a subtype. Here is the current behavior for a similar case:

```motoko
type A = { x : Int};
type B = { x : Int; z : Int};

let a : A = { x = 0 };
let b : B = { x = 1; z = 2 };

a == b // => false, no warning
```

A compromise could be to error for types such as `Any`, `()`, `{}`, etc.

PR #4931, opened 2025-03-04
LSP:  Implement a cache for typechecking (Serokell, Milestone-1)

GoPavel:
In order to improve the performance of the LSP server we need to be able to cache and pass cached the type checking context (`Scope.t`). It's helpful in two situations: 
* When the server initializes, it scans all files in the vscode workspace, and a library code (e.g. `mo:base`) is parsed and type-checked as many times as it's imported (transitively) in any other file. Caching fixes this by allowing dependency type information to be reused.
* Imagine that an IDE user is working on a file, so the content of the file is changed often while its dependencies are not. Scope caching then significantly reduces the compiler effort needed to get a new type information for the file.

**Changes:**
* `src/pipeline/`
  * Add `load_progs_cached`/`chase_imports_cached` that are like `load_progs`/`chase_imports` but also take the scope cache and populates it. Note, expiring logic is implemented in the TS part of the server, see adjoint MR: https://github.com/dfinity/vscode-motoko/pull/329). 
  * Expose these changes in parseMotokoTyped. Create parseMotokoWithDeps that is like parseMotoko but also returns immediate dependencies
* `src/js` (IDE API)
  * Replace `parseMotokoTyped paths` which returns typed AST with `parseMotokoTyped path scopeCache` which returns typed AST and updated `scopeCache`
  * Add new method `parseMotokoWithDeps` which is like `parseMotoko` but also returns a list of immediate imports 

See also interface changes in the `node-motoko`: https://github.com/dfinity/node-motoko/pull/113

### DIFF
--- a/src/js/common.ml
+++ b/src/js/common.ml
@@ -157,14 +157,14 @@ let js_parse_motoko s =
     end)
     in Js.some (js_of_sexpr (Arrange.prog prog)))
 
-let js_parse_motoko_with_deps s =
-  let main_file = "" in
+let js_parse_motoko_with_deps main_file s =
+  let main_file = Js.to_string main_file in
   let s = Js.to_string s in
   let prog_and_deps_result =
     let open Diag.Syntax in
     let* prog, _ = Pipeline.parse_string main_file s in
     let* deps =
-      Pipeline.ResolveImport.resolve (Pipeline.resolve_flags ()) prog s
+      Pipeline.ResolveImport.resolve (Pipeline.resolve_flags ()) prog main_file
     in
     Diag.return (prog, deps)
   in
--- a/src/js/common.ml
+++ b/src/js/common.ml
@@ -157,25 +157,116 @@ let js_parse_motoko s =
     end)
     in Js.some (js_of_sexpr (Arrange.prog prog)))
 
-let js_parse_motoko_typed paths =
-  let paths = paths |> Js.to_array |> Array.to_list in
-  let load_result = Mo_types.Cons.session (fun _ ->
-    Pipeline.load_progs Pipeline.parse_file (paths |> List.map Js.to_string) Pipeline.initial_stat_env)
+let js_parse_motoko_with_deps s =
+  let main_file = "" in
+  let s = Js.to_string s in
+  let prog_and_deps_result =
+    let open Diag.Syntax in
+    let* prog, _ = Pipeline.parse_string main_file s in
+    let* deps =
+      Pipeline.ResolveImport.resolve (Pipeline.resolve_flags ()) prog s
+    in
+    Diag.return (prog, deps)
   in
-  js_result load_result (fun (libs, progs, senv) ->
-  progs |> List.map (fun prog ->
+  js_result prog_and_deps_result (fun (prog, deps) ->
     let open Mo_def in
-    let module Arrange_sources_types = Arrange.Make (struct
+    let module Arrange = Arrange.Make (struct
       let include_sources = true
-      let include_types = true
+      let include_types = false
       let include_docs = Some prog.note.Syntax.trivia
       let include_parenthetical = false
-      let main_file = Some prog.at.left.file
-    end)
-    in object%js
-      val ast = js_of_sexpr (Arrange_sources_types.prog prog)
-      (* val typ = js_of_sexpr (Arrange_sources_types.typ typ) *)
-    end) |> Array.of_list |> Js.array |> Js.some)
+      let main_file = Some main_file
+    end) in
+    Js.some (
+      object%js
+        val ast = js_of_sexpr (Arrange.prog prog)
+        val immediateImports =
+          deps
+          |> List.map (fun dep -> Js.string (Pipeline.resolved_import_name dep))
+          |> Array.of_list
+          |> Js.array
+      end))
+
+module Map_conversion (Map : Map.S) = struct
+  let from_js
+    (type data)
+    (map : _ Js.t)
+    (from_key : 'k Js.t -> Map.key)
+    (from_data : 'd Js.t -> data)
+    : data Map.t
+  =
+  let result = ref Map.empty in
+  let callback =
+    (* [forEach] in JS gives value, key, and map, in this order. *)
+    Js.wrap_callback (fun v k _m ->
+      let k = from_key k in
+      let v = from_data v in
+      result := Map.add k v !result)
+  in
+  ignore (Js.Unsafe.meth_call map "forEach" [|Js.Unsafe.inject callback|]);
+  !result
+
+  let to_js
+    (type data)
+    (map : data Map.t)
+    (from_key : Map.key -> Js.Unsafe.top Js.t)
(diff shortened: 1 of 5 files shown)


rvanasa (commented):
LGTM on the `moc.js` side, although I'll wait for @ggreif or @crusso to take a look as well since they're more familiar with this codebase.

src/js/common.ml
```diff
@@ -157,25 +157,116 @@ let js_parse_motoko s =
...
+  let result = ref Map.empty in
+  let callback =
+    (* [forEach] in JS gives value, key, and map, in this order. *)
+    Js.wrap_callback (fun v k _m ->
+      let k = from_key k in
+      let v = from_data v in
+      result := Map.add k v !result)
+  in
+  ignore (Js.Unsafe.meth_call map "forEach" [|Js.Unsafe.inject callback|]);
+  !result
```
crusso, line 207: indent?

src/js/common.ml
```diff
@@ -157,25 +157,116 @@ let js_parse_motoko s =
...
+  =
+  let js_map = Js.Unsafe.new_obj (Js.Unsafe.pure_js_expr "Map") [||] in
+  Map.iter
+    (fun k v ->
+      ignore (Js.Unsafe.meth_call js_map "set" [| from_key k; from_data v |]))
+    map;
+  js_map
+
+  let from_ocaml = to_js
+  let to_ocaml = from_js
```
crusso, line 224: indent?

src/js/common.ml
```diff
@@ -157,25 +157,116 @@ let js_parse_motoko s =
...
+    (from_key : Map.key -> Js.Unsafe.top Js.t)
+    (from_data : data -> Js.Unsafe.top Js.t)
+    : _ Js.t
+  =
+  let js_map = Js.Unsafe.new_obj (Js.Unsafe.pure_js_expr "Map") [||] in
+  Map.iter
+    (fun k v ->
+      ignore (Js.Unsafe.meth_call js_map "set" [| from_key k; from_data v |]))
+    map;
+  js_map
```
crusso, line 221: indent?

src/pipeline/pipeline.ml
```diff
@@ -340,14 +339,31 @@ When we load a declaration (i.e from the REPL), we also care about the type
...
 *)
 
+type scope_cache = Scope.t Type.Env.t
+
+type load_result_cached =
+    ( Syntax.lib list
+    * (Syntax.prog * string list) list
+    * Scope.t
+    * scope_cache )
+  Diag.result
```
crusso, line 350:
```suggestion
    * scope_cache ) Diag.result

```

crusso (approved):
Wow, this made my head hurt. I think it's ok - see minor formatting suggestions.

src/js/common.ml
```diff
@@ -157,25 +157,116 @@ let js_parse_motoko s =
...
+
+  let from_ocaml = to_js
+  let to_ocaml = from_js
+end
+
+let js_parse_motoko_typed paths scope_cache =
```
rvanasa, line 227: Perhaps `scope_cache` could be optional for backwards-compatibility? We could do this on the `node-motoko` side by passing an empty map, but it would be ideal to preserve the original behavior when directly calling the `moc.js` function.

GoPavel: Sure. Could you clarify the rules/intentions regarding backward compatibility? So, having `scope_cache` optional allows us to run the older VSC extension with a newer compiler version.

Should we check some other cases?

src/js/common.ml
GoPavel: Can you give an idea how it should look like? I have tried to run `ocamlformat` and it changes it to

```ocaml
module Map_conversion (Map : Map.S) = struct
  let from_js (type data) (map : _ Js.t) (from_key : 'k Js.t -> Map.key)
      (from_data : 'd Js.t -> data) : data Map.t =
```
(the rest is the same)

src/js/common.ml
rvanasa: What I'm thinking is that when passing only one argument to `parseMotokoTyped()` in `node-motoko`, we should preserve the original behavior (including return type) to minimize the chance of breaking downstream applications. This may not be possible with `moc.js`, so we could perhaps accept the breaking change here and then maintain the original behavior on the `node-motoko` side. 

I also opened https://github.com/serokell/motoko/pull/44 as a starting point for testing the `parseMotokoTyped()` behavior in `moc.js`. Feel free to expand on this, although the tests in `node-motoko` are a higher priority imo.

src/js/common.ml
crusso: Oh, I just preferred the body of the function to be indented (lines 198-207), but don't worry about it.

src/js/common.ml
crusso: Would the other option be to just add some new entry point and keep the old ones?

src/js/common.ml
GoPavel: Right, I didn't notice it

src/pipeline/pipeline.ml
GoPavel: done

src/js/common.ml
GoPavel: done

src/js/common.ml
GoPavel: done

src/pipeline/pipeline.ml
GoPavel: done

src/js/common.ml
rvanasa: If we eventually wanted to support switching between different versions in the extension, it might be simpler to check whether the method uses the new function interface if it used a different name. Either way should be ok, although we would err on the side of using a new function for this.

GoPavel: So, should we do this? I mean make a new API function instead ?

rvanasa: Yes, that would be best as long as this doesn't introduce too much extra complexity to the codebase.

GoPavel: Done, you can check it out

GoPavel:
@crusso @rvanasa 
We have finished fixes after the first round of review.
Should we merge it, or will there be another round of review?

GoPavel:
Also, did you take a look at https://github.com/dfinity/vscode-motoko/pull/329 ?
Can I assume it's okay since we don't have any discussion there?

rvanasa:
> Also, did you take a look at https://github.com/dfinity/vscode-motoko/pull/329?

Apologies for the delay in commenting on this. I initially looked through the PR last week and did a more in-depth review today. LGTM aside from one suggestion for maintainability.

rvanasa (approved):
Let's ship it! Thanks again for your work on this! Having a separate `moc.js` function will be good because we can make further breaking changes as needed without impacting any existing applications.

crusso (approved):
Sorry, got way to distracted by other things...

crusso (approved):
LGTM

PR #4935, opened 2025-03-05
feat: support `import` values from `file:` URI scheme

ggreif:


test/fail/ok/file-import.tc.ok
```diff
@@ -0,0 +1 @@
+file-import.mo:1.9-1.10: warning [M0194], unused identifier f (delete or rename to wildcard `_` or `_f`)
```
ggreif, line 1: This should be a type error

crusso:
I think it  might be easier to have differnt uris

impor bloby = "blob:Bla.wasm";
import texty = "text:Bla.md";
import principally = "principal:Bla.txt";
import envvar= "env:USER";
import foreignclass= "class:foo.wasm";

letmejustputthishere:
hey @ggreif i see you made good progress here, any ETA 👀

ggreif:
I picked it up again, and swapping back in my brain how all the parts are interacting. It is a bit tricky how to type check it when the type is ultimately guided by the pattern, but I have an idea how to make it smooth (also for the users) and expandable in the future.

letmejustputthishere:
> but I have an idea how to make it smooth (also for the users) and expandable in the future.

awesome, looking forward to it 🦾

test/fail/non-utf8/sample
```diff
@@ -0,0 +1 @@
+import { debugPrint; encodeUtf8 } "mo:�
```
ggreif, line 1: this is intended

src/mo_frontend/typing.ml
```diff
@@ -2016,7 +2017,12 @@ and check_exp' env0 t exp : T.typ =
     t'
   | TagE (id, exp1), T.Variant fs when List.exists (fun T.{lab; _} -> lab = id.it) fs ->
     let {T.typ; _} = List.find (fun T.{lab; typ;_} -> lab = id.it) fs in
-    check_exp env typ exp1 ;
+    check_exp env typ exp1;
+    t
+  | ImportE (_, {contents = ImportedValuePath path}), (T.(Prim Text) as t) ->
+    Lib.FilePath.contents path |> check_text env exp.at |> ignore;
```
ggreif, line 2023: maybe stash the contents (or some AST) into a `ref` and use that in `desugar.ml` and the interpreeters?

src/lang_utils/diag.ml
```diff
@@ -89,7 +89,7 @@ let with_message_store ?(allow_errors = false) f =
   let r = f s in
   let msgs = get_msgs s in
   match r with
-  | Some x when not (has_errors msgs) || allow_errors -> Ok (x, msgs)
+  | Some x when allow_errors || not (has_errors msgs) -> Ok (x, msgs)
```
ggreif, line 92: innocent optimisation

src/lang_utils/diag.ml
```diff
@@ -105,3 +105,5 @@ let flush_messages : 'a result -> 'a option = function
 let run r = match flush_messages r with
   | None -> exit 1
   | Some x -> x
+
+let slienced : msg_store = ref []
```
ggreif, line 109: typo

src/lib/lib.ml
```diff
@@ -689,6 +689,8 @@ struct
           ic, [message]
       end
     else ic, []
+
+    let contents file = In_channel.with_open_bin file In_channel.input_all
```
ggreif, line 693: check how `.mo` files are read...

src/lowering/desugar.ml
```diff
@@ -1233,17 +1233,23 @@ let link_declarations imports (cu, flavor) =
...
       varE (var (id_of_full_path fp) t)
     | S.PrimPath ->
       varE (var (id_of_full_path "@prim") t)
     | S.IDLPath (fp, canister_id) ->
       primE (I.ActorOfIdBlob t) [blobE canister_id]
+    | S.ImportedValuePath path ->
+       let contents = Lib.FilePath.contents path in
+       T.(match t with
+          | Prim Text -> textE
+          | Non | Prim Blob -> blobE
```
ggreif, line 1251: remove `Non`

src/mo_def/compUnit.ml
```diff
@@ -73,6 +73,11 @@ let comp_unit_of_prog as_lib (prog : prog) : comp_unit =
   in
   go [] prog.it
 
+(* Lib as a single value *)
+let comp_unit_of_value full_path : comp_unit =
+  let at = Source.no_region in
+  let prog_typ_note = { empty_typ_note with note_typ = Type.Non (* common subtype of Text and Blob *) } in
```
ggreif, line 79: Can we get rid of this `Non`?

src/pipeline/pipeline.ml
```diff
@@ -450,6 +455,12 @@ let chase_imports_cached parsefn senv0 imports scopes_map
         pending := remove it !pending;
         Diag.return ()
       end
+    | Syntax.ImportedValuePath full_path ->
+      let lib = lib_of_value full_path in
+      libs := lib :: !libs; (* NB: Conceptually an append *)
+      let sscope = Scope.lib full_path Type.Non (* common subtype of Text and Blob *) in
```
ggreif, line 461: Can we survive with just `Blob` here?

src/mo_frontend/typing.ml
```diff
@@ -3254,6 +3273,19 @@ and infer_dec_valdecs env dec : Scope.t =
...
     let _ve = check_pat env obj_typ pat in
     Scope.{empty with val_env = singleton id obj_typ}
+  | LetD (pat, exp, None) when is_value_import dec ->
+    let typ, val_env = match recover_opt (infer_pat false {env with msgs = Diag.silenced}) pat with
+      | None -> T.blob, check_pat env T.blob pat
+      | Some tv -> tv
+    in
+    (match T.normalize typ with
+     | T.(Prim (Text | Blob)) -> ()
+     | T.Pre -> failwith "Pre"
```
ggreif, line 3283: clean up

src/mo_interpreter/interpret.ml
```diff
@@ -1122,7 +1128,7 @@ let interpret_prog flags scope p : (V.value * scope) option =
 (* Libraries *)
 
 (* Import a module unchanged, and a class constructor as an asynchronous function.
-   The conversion will be unnecessary once we declare classes as asynchronous. *)
+   The conversion will be unnecessary once we declare classes as asynchronous. FIXME: is this correct? *)
```
ggreif, line 1131: @crusso aren't they already?

src/mo_frontend/typing.ml
```diff
@@ -3357,6 +3372,17 @@ and infer_dec_valdecs env dec : Scope.t =
     let obj_typ = object_of_scope env obj_sort.it dec_fields obj_scope' at in
     let _ve = check_pat env obj_typ pat in
     Scope.{empty with val_env = singleton id obj_typ}
+  | LetD (pat, exp, None) when is_value_import dec ->
+    let typ, val_env = match recover_opt (infer_pat false {env with msgs = Diag.silenced}) pat with
+      | None -> T.blob, check_pat env T.blob pat
+      | Some tv -> tv
+    in
```
ggreif, line 3379:
```suggestion
      | Some tv -> tv in
```

test/fail/ok/file-import.tc.ok
ggreif: it is now

src/lib/lib.ml
ggreif: It is done in `Pipeline.parse_file'`

src/lowering/desugar.ml
```diff
@@ -1288,6 +1294,8 @@ let transform_unit_body (u : S.comp_unit_body) : Ir.comp_unit =
       I.ActorU (None, ds, fs, u, t)
     | _ -> assert false
     end
+  | S.FileU str ->
```
Kamirus, line 1297: `str` is meant to be unused?

ggreif: Good point, maybe this code is dead. Will check.

src/ic/url.ml
```diff
@@ -79,10 +81,12 @@ let parse (f: string) : (parsed, string) result =
       match Lib.String.chop_prefix "canister:" f with
       | Some suffix -> Ok (IcAlias suffix)
       | None ->
-        begin match Stdlib.String.index_opt f ':' with
-        | Some _ -> Error "Unrecognized URL"
-        | None -> Ok (Relative (Lib.FilePath.normalise f))
-        end
+        match Lib.String.chop_prefix "file:" f with (* FIXME: check official format for `file:` URI scheme *)
```
Kamirus: Do we need an extra URI validation? Can it be abused when malicious?
Maybe we should have a test with a bad URI

src/lowering/desugar.ml
```diff
@@ -1233,17 +1233,23 @@ let link_declarations imports (cu, flavor) =
...
     | S.PrimPath ->
       varE (var (id_of_full_path "@prim") t)
     | S.IDLPath (fp, canister_id) ->
       primE (I.ActorOfIdBlob t) [blobE canister_id]
+    | S.ImportedValuePath path ->
+       let contents = Lib.FilePath.contents path in
+       T.(match t with
+          | Prim Text -> textE
+          | Prim Blob -> blobE
+          | t -> (*Printf.eprintf "Cannot type: %s\n" (Wasm.Sexpr.to_string 80 (Arrange_type.typ t));*) assert false) contents
```
Kamirus: This is an unreachable case because there is validation somewhere else?
If so then I'd change it to:
```suggestion
          | _ -> assert false) contents
```

src/mo_def/compUnit.ml
```diff
@@ -73,6 +73,11 @@ let comp_unit_of_prog as_lib (prog : prog) : comp_unit =
   in
   go [] prog.it
 
+(* Lib as a single value *)
+let comp_unit_of_value full_path : comp_unit =
+  let at = Source.no_region in
+  let prog_typ_note = { empty_typ_note with note_typ = Type.blob } in
```
Kamirus: Always `Type.blob`? What about the `Text` type?

src/mo_frontend/typing.ml
```diff
@@ -3336,6 +3346,11 @@ and is_import d =
   | LetD (_, {it = ImportE _; _}, None) -> true
   | _ -> false
 
+and is_value_import d =
+  match d.it with
+  | LetD (_, {it = ImportE (_, {contents = ImportedValuePath _}); _}, None) -> true
```
Kamirus: I get it that the import statements are desugared to let declarations?

src/mo_frontend/typing.ml
```diff
@@ -3357,6 +3372,16 @@ and infer_dec_valdecs env dec : Scope.t =
     let obj_typ = object_of_scope env obj_sort.it dec_fields obj_scope' at in
     let _ve = check_pat env obj_typ pat in
     Scope.{empty with val_env = singleton id obj_typ}
+  | LetD (pat, exp, None) when is_value_import dec ->
+    let typ, val_env = match recover_opt (infer_pat false {env with msgs = Diag.silenced}) pat with
```
Kamirus: I'd avoid backtracking here, also this Diag.silenced would not be necessary.
We basically support two kinds of patterns for these value imports: annotated (Blob or Text) or not which means defaults to Text.
I'd just check if there is an annotation, then call check_pat_exhaustive.
Otherwise call infer_pat_exhaustive.

Kamirus: Or even better, differentiate between blob and text by the import URI

src/mo_interpreter/interpret.ml
```diff
@@ -456,6 +456,11 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
     | Unresolved -> assert false
     | LibPath {path; _} ->
       k (find path env.libs)
+    | ImportedValuePath path ->
+      let contents = find path env.libs |> V.as_blob in
+      k T.(match normalize exp.note.note_typ with | Prim Blob -> V.Blob contents | Prim Text -> V.Text contents
```
crusso:
```suggestion
      k T.(match normalize exp.note.note_typ with | Prim Blob -> V.Blob contents | Prim Text -> V.Text contents
```
Does V.Text validate this is utf8? Probably not and we should.

src/mo_interpreter/interpret.ml
```diff
@@ -1140,9 +1145,10 @@ let import_lib env lib =
             if tag = "new" && V.Env.find "settings" o = V.Null
             then k v
             else trap cub.at "actor class configuration unsupported in interpreter")))) ])
+  | Syntax.FileU str -> fun _ -> V.Blob str
+    (* always create a Blob here, but the ImportedValuePath will apply the coercion if Text needed *)
```
crusso: This would be unnecessay if the Uri indicted the type, right?

src/mo_frontend/typing.ml
crusso: That's a bit historical - Joachim originally allowed full import expressions anywhere, but then we restricted them to leading imports only, without changing the AST (FBOW)

src/mo_frontend/typing.ml
crusso: Yeah, I strongly agree.

src/ic/url.ml
crusso:
```suggestion
        match Lib.String.chop_prefix "file:" f with (* FIXME: check official format for `file:` URI scheme *)
```
At this point, couldn't we look for `blob` and `text` and act occordingly, adding the appropriate type to `FileValue (T.type * string)`

src/ic/url.ml
```diff
@@ -42,13 +42,15 @@ type parsed =
   | Relative of string
   | Ic of string
   | IcAlias of string
+  | FileValue of string
```
crusso, line 45:
```suggestion
  | FileValue of T.typ * string
```
Might make this easier and avoid the inference hacks.

src/mo_def/syntax.ml
```diff
@@ -17,6 +17,7 @@ type resolved_import =
   | Unresolved
   | LibPath of lib_path
   | IDLPath of (string * string) (* filepath * bytes *)
+  | ImportedValuePath of string
```
crusso, line 20:
```suggestion
  | ImportedValuePath of Type.typ * string
```
Might help

src/mo_frontend/typing.ml
crusso: Then all this code can go away and check_import just deals with it.

src/mo_def/syntax.ml
```diff
@@ -254,6 +255,7 @@ and comp_unit_body' =
  | ModuleU of id option * dec_field list     (* module library *)
  | ActorClassU of                            (* IC actor class, main or library *)
      exp option * sort_pat * typ_id * typ_bind list * pat * typ option * id * dec_field list
+ | FileU of string                           (* literal file value *)
```
crusso: Do we really need this? Couldn't ImportedValuePath carry the file contents, just as IDLPath carries the bytes?

crusso: You don't actually need this and could simplify the code by embedding the bytes in ImportValuePath (similar to the way IDLPath carries the bytes of the canister id. Then lowering can just inline the literal (the backend will share them if necessary).

crusso (changes requested):
I think this code can be simplified significantly and probably worth another take.
Sorry the existing stuff is quite confusing.

src/lowering/desugar.ml
```diff
@@ -1233,17 +1233,23 @@ let link_declarations imports (cu, flavor) =
...
     | S.LibPath {path = fp; _} ->
       varE (var (id_of_full_path fp) t)
     | S.PrimPath ->
       varE (var (id_of_full_path "@prim") t)
     | S.IDLPath (fp, canister_id) ->
       primE (I.ActorOfIdBlob t) [blobE canister_id]
+    | S.ImportedValuePath path ->
+       let contents = Lib.FilePath.contents path in
+       T.(match t with
+          | Prim Text -> textE
```
ggreif: Ugh, this should be reverted too.

src/mo_def/compUnit.ml
ggreif: Yes, it is `Blob` for now.

src/mo_def/syntax.ml
ggreif: Maybe. It would simplify the jungle.

src/lowering/desugar.ml
Kamirus:
```suggestion
       assert T.(t = Prim Blob);
       blobE contents
```

src/mo_interpreter/interpret.ml
```diff
@@ -1141,9 +1146,10 @@ let import_lib env lib =
...
```
Kamirus:
```suggestion
```
Can be removed, yes?

src/mo_interpreter/interpret.ml
```diff
@@ -457,6 +457,11 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
     | Unresolved -> assert false
     | LibPath {path; _} ->
       k (find path env.libs)
+    | ImportedValuePath path ->
+      let contents = find path env.libs |> V.as_blob in
+      k T.(match normalize exp.note.note_typ with | Prim Blob -> V.Blob contents | Prim Text -> V.Text contents
+                                                  | Non -> failwith "Non?"
+                                                  | _ -> assert false)
```
Kamirus: Aren't `note_typ`s always normalized? So no need for that. Also `Text` cases can be removed now:
```suggestion
      assert T.(exp.note.note_typ = Prim Blob);
      k (V.Blob contents)
```

src/mo_frontend/typing.ml
```diff
@@ -3052,9 +3051,13 @@ and infer_dec env dec : T.typ =
   let t =
   match dec.it with
   | ExpD exp -> infer_exp env exp
+  | LetD (pat, exp, None) when is_value_import dec ->
+    let typ = pat.note in
+    check_exp env typ exp;
+    typ
```
```suggestion
```
Not necessary, yes? `infer_exp` below should cover it. `is_value_import` can also be removed

src/mo_frontend/typing.ml
```diff
@@ -3395,6 +3403,9 @@ and infer_dec_valdecs env dec : Scope.t =
     let obj_typ = object_of_scope env obj_sort.it dec_fields obj_scope' at in
     let _ve = check_pat env obj_typ pat in
     Scope.{empty with val_env = singleton id obj_typ}
+  | LetD (pat, exp, None) when is_value_import dec ->
+    let val_env = check_pat_exhaustive local_error env T.blob pat in
+    Scope.{empty with val_env}
```
```suggestion
```
should be covered by the case below, `is_value_import` can be removed

src/mo_frontend/typing.ml
```diff
@@ -3374,6 +3377,11 @@ and is_import d =
   | LetD (_, {it = ImportE _; _}, None) -> true
   | _ -> false
 
+and is_value_import d =
+  match d.it with
+  | LetD (_, {it = ImportE (_, {contents = ImportedValuePath _}); _}, None) -> true
+  | _ -> false
+
```
```suggestion
```

src/ic/url.ml
ggreif: Well, I'd suggest that reaching out of the project directory should be disallowed. I _think_ `mo:` scheme already checks this, I'll have a look.

ggreif: I see where you are coming from, but it sounds like a hack to me. The URI schema's first word is a protocol (i.e. how to get the data), not a type specification for the contents. So I am rather hesitant to interpret it that way. When we have a blob store and we access blobs from it by hash (e.g. `import pic = "blob:7ad8227f-some-irrelevant-human-added-suffix"`), then _that_ would be a protocol.

ggreif: I have to correct that. We don't check even for `Relative` URLs that they are not absolute paths. So let's not complicate this. My FIXME comment above was about the format of the `file://` URI scheme's specification, not about the usage here. But I think those extra slashes are just adding noise for the users, so I prefer to _not_ adhere to the URI spec.

ggreif:
```suggestion
        match Lib.String.chop_prefix "file:" f with
```

Kamirus: @ggreif is the PR ready for final reviews? Or you plan more changes like this suggestion above?

ggreif: Yes, I consider it done, but will try to get rid of `FileU` (maybe in a follow-up PR) if I figure it out how to do it. (Haven't yet, but also didn't sink a lot of time in it.)

crusso: Isn't the type telling you how to interpret (get) the data?
another alternative might be:

"resource://blob/<path>"

src/ic/url.ml
```diff
@@ -79,10 +81,12 @@ let parse (f: string) : (parsed, string) result =
       match Lib.String.chop_prefix "canister:" f with
       | Some suffix -> Ok (IcAlias suffix)
       | None ->
-        begin match Stdlib.String.index_opt f ':' with
-        | Some _ -> Error "Unrecognized URL"
-        | None -> Ok (Relative (Lib.FilePath.normalise f))
-        end
+        match Lib.String.chop_prefix "blob:file:" f with
```
crusso, line 84: Hmm, if this is meant to be a URI, then I don't think you can have a colon within the scheme."blob.file:" would be ok as is "mo.blob:"

ggreif: They stack. See http://www.w3.org/TR/FileAPI/#url, which is referenced by the link in the PR description.

crusso: Oh ok, I was looking in the wrong place perhaps: https://datatracker.ietf.org/doc/html/rfc3986#section-3.1

doc/md/16-language-manual.md
```diff
@@ -361,9 +361,10 @@ The syntax of an **import** `<imp>` is as follows:
   "mo:<package-name>/<filepath>"    Import module from package
   "ic:<canisterid>"                 Import external actor by <canisterid>
   "canister:<name>"                 Import external actor by <name>
+  "blob:file:<filepath>"            Import literal `Blob` value from <filepath>
 ```
 
-An import introduces a resource referring to a local source module, module from a package of modules, or canister imported as an actor. The contents of the resource are bound to `<pat>`.
+An import introduces a resource referring to a local source module, module from a package of modules, a canister imported as an actor, or a literal [`Blob`](https://internetcomputer.org/docs/motoko/core/Blob.md) value. The contents of the resource are bound to `<pat>`.
```
ggreif: Or should we link against the anchor for the `Type Blob` (around line 719)?

ggreif: Groq suggests `## Section 1 {#section1}` as the anchor and `[link](#section1)` as the reference.

crusso: Yeah, would just link to the local section (I hate the fact that we know have hard-coded links in the doc because of some limitation of docusaurus)

crusso: Could you also add a short section and example in the file corresponding to this 
https://internetcomputer.org/docs/motoko/fundamentals/modules-imports

doc/md/fundamentals/7-modules-imports.md
```diff
@@ -185,3 +185,20 @@ persistent actor CountToTen {
...
+actor {
+    func checkSig(key : Blob, cyphertext : Text) { ... };
+
+    public func verify(cyphertext : Text) : async () {
+        checkSig(pub, cyphertext);
+    };
+};
+```
+
+This usecase also caters for the import of externally-built Wasm `Blob`s for low-level installation and upgrade by means of the management canister.
```
crusso:
```suggestion
This usecase also caters for the import of externally-built Wasm `Blob`s for low-level installation and upgrade by means of the management canister.
```
```suggestion
This use case also caters for the import of externally-built Wasm `Blob`s for low-level installation and upgrade by means of the management canister.
```

PR #4952, opened 2025-03-17
chore: provide a tarballs for `Darwin-arm64` and `Linux-aarch64`

ggreif:
Most of this is straightforward, however there are three gotchas:
- When building statically-linked executables on `aarch64-linux`, the toolset leaves a loader-interpreter path in the ELF.
  This doesn't get corrected in the fixup phase, so we do it in the install phase earlier. This is in order to function in the sandbox. But we only edit for non-release, as the release exe will go into tarball (we check in the GH action for runnability it before putting it in).
- The `moc` flag `-measure-rts-stack` behaves differently on Arm64, so we skip such tests there. It is experimental anyway.
- Every touched page costs 1000 cycles, so an Arm64 Mac (with 16 vs. 4 kB pages) spends less cycles. Since most Motoko devs have M1 (or better) Macs, we only run `tests.bench` in CI for `aarch64-darwin`. x86_64-based devs should be careful not to accept in `test/bench`.


.github/workflows/release.yml
```diff
@@ -11,6 +11,7 @@ on:
     - '*'
     branches:
     - 'master'
+    - 'gabor/Arm64'
```
ggreif, line 14: remove this, it was just for testing

.github/workflows/release.yml
```diff
@@ -63,7 +64,7 @@ jobs:
     - name: "nix-build"
       # these are the dependencies listed in release-files. Sorry for the duplication
       run: |
-        nix-build --max-jobs 1 --arg officialRelease true -A moc -A mo-ide -A mo-doc -A js.moc -A js.moc_interpreter
+        nix-build --max-jobs 1 --arg officialRelease true -A moc -A mo-ide -A mo-doc -A js.moc
```
ggreif, line 67: undo!

default.nix
```diff
@@ -95,6 +95,8 @@ let commonBuildInputs = pkgs:
     pkgs.obelisk
     pkgs.perl
     pkgs.removeReferencesTo
+    pkgs.which
```
ggreif, line 98: these are for figuring out why aarch64-linux doesn't want to build static executables.

default.nix
```diff
@@ -107,7 +109,7 @@ let ocaml_exe = name: bin: rts:
     staticpkgs.stdenv.mkDerivation {
       inherit name;
 
-      allowedRequisites = [];
+      allowedRequisites = if is_static then staticpkgs.musl else [];
```
ggreif, line 112:
```suggestion
      allowedRequisites = if system == "aarch64-linux" then staticpkgs.musl else [];
```
but even this doesn't help in a.t.m.

default.nix
```diff
@@ -140,13 +144,31 @@ let ocaml_exe = name: bin: rts:
...
+        which patchelf
+        ldd $(which patchelf)
+        ls -l $out/bin/*
+        chmod +w $out/bin/*
+        patchelf --set-interpreter "${staticpkgs.musl}/lib/ld-musl-aarch64.so.1" $out/bin/* || true
+        ldd $out/bin/* || true
+        file $out/bin/*
+        ls -l $out/bin/*
+        chmod a-w $out/bin/*
+      '' else "" + ''
```
ggreif, line 167: remove above debugging code

default.nix
```diff
@@ -573,7 +595,7 @@ rec {
         doInstallCheck = true;
         test = ./test + "/test-${n}.js";
         installCheckPhase = ''
-          NODE_PATH=$out/bin node --experimental-wasm-memory64 $test
+          # NODE_PATH=$out/bin node --experimental-wasm-memory64 $test
```
ggreif, line 598: undo: this _should_ work now

test/run.sh
```diff
@@ -323,6 +323,14 @@ do
         continue
       fi
     fi
+    if grep -q "//SUPPORTED-PLATFORM-ONLY" $base.mo
```
ggreif, line 326: maybe check for `MOC-FLAG=-measure-rts-stack`

default.nix
```diff
@@ -140,13 +143,21 @@ let ocaml_exe = name: bin: rts:
...
+        # also, there is a reference to /nix/store/…/share/menhir/standard.mly.
         # Let's remove that, too
         remove-references-to \
           -t ${staticpkgs.ocamlPackages.menhir} \
           $out/bin/*
-        # sanity check
-        $out/bin/* --help >/dev/null
+      '' + nixpkgs.lib.optionalString (!officialRelease && is_dyn_static) ''
+        # these systems need a fixup to the loader interpreter
+        chmod +w $out/bin/*
```
ggreif, line 153: try removing this

test/run.sh
ggreif: @luc-blaeser would that make sense or should we leave as-is?

.github/workflows/release.yml
```diff
@@ -43,7 +43,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ ubuntu-latest, macos-13 ]
+        os: [ ubuntu-latest, macos-13, ubuntu-24.04-arm, macos-latest ]
```
ggreif, line 46: `macos-latest` are M1-M4 based Mac builders

.github/workflows/release.yml
```diff
@@ -64,6 +64,11 @@ jobs:
       # these are the dependencies listed in release-files. Sorry for the duplication
       run: |
         nix-build --max-jobs 1 --arg officialRelease true -A moc -A mo-ide -A mo-doc -A js.moc -A js.moc_interpreter
+        echo "## checks"
+        uname -s -m
+        $(nix-build --max-jobs 1 --arg officialRelease true -A moc)/bin/moc --version
+        file $(nix-build --max-jobs 1 --arg officialRelease true -A moc)/bin/moc
+        ldd $(nix-build --max-jobs 1 --arg officialRelease true -A moc)/bin/moc || true
```
ggreif, line 71: Macs don't have `ldd`, thus `|| true`

ggreif, line 71: This is just some smoke test to see the process worked before putting stuff into the tarballs.

default.nix
ggreif: Nope, I'd get
```
       > Running phase: installPhase
       > 'deser' -> '/nix/store/pk3i33rczhz61rwrg8l743si1yv3krbq-deser/bin/deser'
       > chmod +w /nix/store/pk3i33rczhz61rwrg8l743si1yv3krbq-deser/bin/deser
       > -r-xr-xr-x 1 nixbld nixbld 4724712 Mar 28 10:34 /nix/store/pk3i33rczhz61rwrg8l743si1yv3krbq-deser/bin/deser
       > patchelf: open: Permission denied
```

test/run.sh
ggreif: It works: dd2516743

default.nix
```diff
@@ -103,11 +103,14 @@ let ocaml_exe = name: bin: rts:
       if is_static
       then "release-static"
       else "release";
+    is_dyn_static =
```
ggreif, line 106: see PR description

.github/workflows/release.yml
vincent-dfinity: For now, `macos-latest` is `macOS 14 Arm64`, not sure if it's tied to a specific M* chipset, but it's Arm64-based.

vincent-dfinity (approved):
Thanks!

alexandru-uta (approved):
looks good, thanks!

.github/workflows/release.yml
ggreif: Yeah, sure one of those is it :-)

PR #4955, opened 2025-03-18
Docs restructure ICP features

Demali-876:
This PR restructures the documentation for the **ICP features** section, including **randomness, timers, caller identification, Candid serialization, stable memory/regions, system functions** (`heartbeat`, `timer`, `preupgrade`, `postupgrade`, `inspect`). and sections for **Motoko style guidelines, grammar rules, and compiler references**. 

There are some key changes to the caller identification section with emphasis on access control, a complete overhaul of stable memory, the introduction of `fuzz` and `idempotency-keys` libraries for randomness. The introduction of `mocv` to allow developers to use the moc compiler outside of dfx. And other changes to the documentation.


doc/md/icp-features/caller-identification.md
```diff
@@ -2,4 +2,399 @@
...
+      throw Error.reject("Access denied: Caller is not whitelisted.");
+    };
+    "Access granted to restricted action.";
+  };
+}
+```
+
+### Role based access control
+
+Role-Based Access Control (RBAC) assigns permissions based on **roles** rather than individual users. This allows for structured authorization, where users can be grouped into predefined roles with different access levels. In this example, the canister assigns roles dynamically, allowing the controller to manage administrators and members, each with different permissions:
```
jessiemongeon1, line 113:
```suggestion
Role-based access control (RBAC) assigns permissions based on roles rather than individual users. This allows for structured authorization, where users can be grouped into predefined roles with different access levels. In this example, the canister assigns roles dynamically, allowing its controller to manage administrators and members, each with different permissions:
```

doc/md/icp-features/caller-identification.md
```diff
@@ -2,4 +2,399 @@
...
+      case (?#Member) "Member action executed.";
+      case _ throw Error.reject("Access denied: Members only.");
+    };
+  };
+}
+```
+
+### Mandatory access control
+
+Mandatory Access Control (MAC) enforces **strict, system-defined rules** that users cannot modify. Unlike **Discretionary Access Control (DAC)**, where the owner can grant and revoke permissions, MAC establishes fixed access policies that remain immutable after deployment. This model is useful for high-security applications, where access rules must be enforced without user intervention.
```
jessiemongeon1, line 172:
```suggestion
Mandatory access control (MAC) enforces strict, system-defined rules that users cannot modify. Unlike discretionary access control (DAC), where the owner can grant and revoke permissions, MAC establishes fixed access policies that remain immutable after deployment. This model is useful for high-security applications, where access rules must be enforced without user intervention.
```

doc/md/icp-features/caller-identification.md
```diff
@@ -2,4 +2,399 @@
...
+import Principal "mo:base/Principal";
+import OrderedMap "mo:base/OrderedMap";
+import Error "mo:base/Error";
+
+shared(msg) persistent actor class MandatoryAccess() {
+  let controller = msg.caller;
+
+  type AccessLevel = { #ReadOnly; #WriteOnly; #ReadWrite };
+
+  // Create OrderedMap instance for storing fixed access rules
```
jessiemongeon1, line 184:
```suggestion
  // Create an OrderedMap instance for storing fixed access rules
```

doc/md/icp-features/caller-identification.md
```diff
@@ -2,4 +2,399 @@
...
+      case (?#ReadWrite) "Data written: " # newData;
+      case _ throw Error.reject("Access denied: No write permissions.");
+    };
+  };
+}
+```
+
+### Access control list
+
+An **Access Control List (ACL)** defines explicit per-user permissions for specific actions, allowing fine-grained control over function access. Unlike **Role-Based Access Control (RBAC)**, which groups users into roles, ACLs require manual management of user access for each function. While ACLs provide flexibility, they can become difficult to scale in larger applications. The following example implements separate read and write allow lists, where the controller manages permissions by adding or removing users:
```
jessiemongeon1, line 223:
```suggestion
An **access control list (ACL)** defines explicit per-user permissions for specific actions, allowing fine-grained control over function access. Unlike role-based access control (RBAC) which groups users into roles, ACLs require manual management of user access for each function. ACLs are flexible but can become difficult to scale in larger applications. The following example implements separate read and write allow lists, where the controller manages permissions by adding or removing users:
```

doc/md/icp-features/caller-identification.md
```diff
@@ -2,4 +2,399 @@
...
+      throw Error.reject("Access denied: Write access required.");
+    };
+    "Data written: " # newData;
+  };
+}
+```
+
+### Attribute-based access control
+
+**Attribute-Based Access Control (ABAC)** grants permissions based on **user attributes** rather than predefined roles or explicit allow lists. In decentralized applications, attributes can include **token balances, reputation scores, or other on-chain data**, making ABAC ideal for Web3 environments. This example implements a **token-gated access model**, where users must hold a minimum number of tokens to **access restricted functions**.
```
jessiemongeon1, line 300:
```suggestion
**Attribute-based access control (ABAC)** grants permissions based on user attributes rather than predefined roles or explicit allow lists. In decentralized applications, attributes can include token balances, reputation scores, or other onchain data, making ABAC ideal for Web3 environments. This example implements a token-gated access model where users must hold a minimum number of tokens to access restricted functions.
```

doc/md/icp-features/caller-identification.md
```diff
@@ -2,4 +2,399 @@
...
+      throw Error.reject("Access denied: Insufficient tokens.");
+    };
+    "Access granted: You meet the token requirement.";
+  };
+}
+```
+
+### Policy-based access control
+
+Policy-Based Access Control (PBAC) enforces access rules dynamically based on contextual conditions rather than predefined roles or attributes. Policies can include time-based restrictions, geographical constraints, or  multi-factor conditions. This model is useful for governance, scheduled access, and automated compliance rules. The following example implements a time-based policy, where access is only granted within a specific time window (9AM- 5PM UTC):
```
jessiemongeon1, line 343:
```suggestion
Policy-based access control (PBAC) enforces access rules dynamically based on contextual conditions rather than predefined roles or attributes. Policies can include time-based restrictions, geographical constraints, or  multi-factor conditions. This model is useful for governance, scheduled access, and automated compliance rules. The following example implements a time-based policy, where access is only granted within a specific time window (9AM- 5PM UTC):
```

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,410 @@
 sidebar_position: 5
 ---
 
-# Stable memory / regions
\ No newline at end of file
+# Stable memory and Regions
```
jessiemongeon1, line 5:
```suggestion
# Stable memory and regions
```

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,410 @@
...
+
+<!---The small code blocks in this these snippets build upon each other, it might be useful to have a single program run in the backend to eliminate unbound errors-->
+
+<!---This section could also benefit from illustrations but i'm not sure what md will allow-->
```
jessiemongeon1, line 9: We can use mermaid diagrams(https://mermaid.js.org/) for simple flow-chart-like graphs, or we can ask the design team to create us some visuals.

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,410 @@
...
+
+When a canister is upgraded, all heap memory (i.e., variables and data structures in normal memory) is reset. Motoko provides stable memory as a mechanism for preserving data across canister upgrades. Stable regions extend this functionality by allowing more structured and flexible memory management.
```
jessiemongeon1, line 11:
```suggestion
When a canister is upgraded, all [heap memory](https://internetcomputer.org/docs/building-apps/canister-management/storage#heap-memory) is reset. Motoko utilizes [stable memory](https://internetcomputer.org/docs/building-apps/canister-management/storage#stable-memory) through the [stable storage feature](https://internetcomputer.org/docs/building-apps/canister-management/storage#motoko-storage-handling) for preserving data across canister upgrades. Stable regions extend this functionality by allowing more structured and flexible memory management.
```

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,410 @@
...
+
+## What is a `Region`?
+
+A **region** is an **isolated chunk of stable (virtual) memory** that can be allocated, grown, and managed independently. It functions like a dedicated section of a storage system, ensuring that its contents remain separate and inaccessible to other parts of the program.
```
jessiemongeon1, line 15:
```suggestion
A **region** is an isolated chunk of stable (virtual) memory that can be allocated, grown, and managed independently. It functions like a dedicated section of a storage system, ensuring that its contents remain separate and inaccessible to other parts of the program.
```

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,410 @@
...
+
+## `Region` structure
```
jessiemongeon1, line 17:
```suggestion
### `Region` structure
```

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,410 @@
...
+
+A region in stable memory consists of a **contiguous block of memory** that is allocated and managed independently. Each region has a starting address, a current size, and a capacity that can grow dynamically. One could think of a region similarly to an array, where each position corresponds to an index and data is accessed through byte offsets. However, unlike a traditional array, memory management within a region is manual, requiring explicit tracking of space usage and current position to ensure accurate data access and storage.
```
jessiemongeon1, line 19:
```suggestion
A region in stable memory consists of a contiguous block of memory that is allocated and managed independently. Each region has a starting address, a current size, and a capacity that can grow dynamically. One could think of a region similarly to an array, where each position corresponds to an index and data is accessed through byte offsets. However, unlike a traditional array, memory management within a region is manual, requiring explicit tracking of space usage and current position to ensure accurate data access and storage.
```

jessiemongeon1 (commented):
Unfinished review; will finish when I return from PTO on Monday.

doc/md/icp-features/stable-memory.md
Demali-876: I think for presentation it would be great to have the design team bring the memory layouts to life. I like the work that the design team does, I'm sure they'll can do a much better job than i can. I hope they go all of for the motoko domain design.

jessiemongeon1: I can submit a request to design, but their return time is usually 3-4 weeks.

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,410 @@
...
+
+```motoko no-repl
+// Store a floating-point value at offset 200
+Region.storeFloat(myRegion, 200, 3.14159);
+
+// Read the floating-point value from offset 200
+let pi = Region.loadFloat(myRegion, 200);  // Returns 3.14159
+```
+
+Floating-point numbers can also be stored in a region. Since these values occupy multiple bytes, offsets should be spaced accordingly to avoid overwriting adjacent data.
```
jessiemongeon1, line 110:
```suggestion

```

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,410 @@
...
+
+Regions also allows storing and retrieving binary data as blobs. This is useful for handling arbitrary sequences of bytes, such as serialized objects or encoded information.
```
jessiemongeon1, line 112:
```suggestion
`Regions` also allows storing and retrieving binary data as blobs. This is useful for handling arbitrary sequences of bytes, such as serialized objects or encoded information.
```

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,410 @@
...
+let myData = Blob.fromArray([1, 2, 3, 4, 5]);
+
+// Store the blob at offset 300
+Region.storeBlob(myRegion, 300, myData);
+
+// Read 5 bytes from offset 300
+let retrievedData = Region.loadBlob(myRegion, 300, 5);  // Returns the same blob
+```
+
+Blobs can be stored at any offset, but their size must be considered when choosing a starting position. Retrieving the correct number of bytes ensures that data integrity is maintained when working with binary storage in a region. Proper offset management and ensuring adequate space for each data type are critical for efficient memory handling in stable regions.
```
jessiemongeon1, line 125:
```suggestion
Blobs can be stored at any offset, but their size must be considered when choosing a starting position. Retrieving the correct number of bytes ensures that data integrity is maintained when working with binary storage in a `Region`. 
```

jessiemongeon1: Last sentence is redundant

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,410 @@
...
+
+## Memory layout strategies
+
+While the basic operations above demonstrate how to store and retrieve individual values in a region, real-world applications typically require more organized approaches to memory management. The following memory layout strategies build upon these fundamental operations to create structured patterns for data organization. These strategies provide systematic ways to **allocate**, **track**, and **access** data within regions, allowing developers to implement more complex data structures while maintaining efficient memory usage. Each strategy offers different trade-offs between simplicity, flexibility, and performance depending on your application's specific requirements.
```
jessiemongeon1, line 129:
```suggestion
While the basic operations above demonstrate how to store and retrieve individual values in a `Region`, real-world applications typically require more organized approaches to memory management. The following memory layout strategies build upon these fundamental operations to create structured patterns for data organization. These strategies provide systematic ways to allocate, track, and access data within `Region`s, allowing developers to implement more complex data structures while maintaining efficient memory usage. Each strategy offers different trade-offs between simplicity, flexibility, and performance depending on your application's specific requirements.
```

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,410 @@
...
+
+### Sequential
+
+In sequential storage, data is written **one after the other**, tracking the next available position to prevent overwriting. This approach appends new data to the **end of the region**, recording its size before the actual content. Since data is stored continuously, fragmentation is minimized, but tracking **nextPosition** is essential to avoid overwriting.
```
jessiemongeon1, line 133:
```suggestion
In sequential storage, data is written one after the other, tracking the next available position to prevent overwriting. This approach appends new data to the end of the `Region`, recording its size before the actual content. Since data is stored continuously, fragmentation is minimized, but tracking the next position is essential to avoid overwriting.
```

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,410 @@
...
+  nextPosition := position + 8 + size;
+  
+  // Return the position where we stored the data
+  position
+}
+```
+
+### Fixed-size entry
+
+Fixed-size storage divides memory into **uniformly sized slots**, making it easier to access entries at predictable positions. Each entry is assigned a **fixed slot**, making lookups straightforward using `index * ENTRY_SIZE`. This method simplifies **retrieval** but may **waste space** if entries contain less data than the allocated size.
```
jessiemongeon1, line 157:
```suggestion
Fixed-size storage divides memory into uniformly sized slots, making it easier to access entries at predictable positions. Each entry is assigned a fixed slot, making lookups straightforward using `index * ENTRY_SIZE`. This method simplifies retrieval but may waste space if entries contain less data than the allocated size.
```

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,410 @@
...
+// Read from a specific entry slot
+func readEntry(index : Nat64, size : Nat) : Blob {
+  let position = index * ENTRY_SIZE;
+  Region.loadBlob(myRegion, position, size)
+}
+```
+
+### Index-based
+
+For variable-sized data, an **index region** is used to track where each entry is stored within a separate **data region**. This method maintains **two separate memory regions**: one for **indexes** that store positions and sizes, and another for **actual data**. Using an index allows efficient access to **variable-length** data while avoiding fragmentation issues.
```
jessiemongeon1, line 179:
```suggestion
For variable-sized data, an index `Region` is used to track where each entry is stored within a separate data `Region`. This method maintains two separate memory `Regions`: one for **indexes** that store positions and sizes and one for **actual data**. Using an index allows efficient access to variable-length data while avoiding fragmentation issues.
```

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,410 @@
...
+  let position = Region.loadNat64(indexRegion, index * 16);
+  let size = Region.loadNat64(indexRegion, index * 16 + 8);
+  
+  Region.loadBlob(dataRegion, position, Nat64.toNat(size))
+}
+```
+
+## Best practices
+
+Managing stable memory effectively requires structuring data in a way that ensures **compatibility, scalability, and reliability**. Using **versioning, metadata headers, and transaction logs** can help maintain a well-organized memory layout.
```
jessiemongeon1, line 213:
```suggestion
Managing stable memory effectively requires structuring data in a way that ensures compatibility, scalability, and reliability. It is best practice to use versioning, metadata headers, and transaction logs to help maintain a well-organized memory layout.
```

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,410 @@
...
+
+### Versioning data
+
+Including a version marker at the beginning of a region ensures compatibility across different data formats and future upgrades. This approach ensures that **older versions** of the stored data remain accessible while allowing **new formats** to be introduced. The version marker, typically stored at offset `0`, is checked before performing any operations.
```
jessiemongeon1, line 217:
```suggestion
Including a version marker at the beginning of a `Region` ensures compatibility across different data formats and future upgrades. This approach ensures that older versions of the stored data remain accessible while allowing new formats to be introduced. The version marker, typically stored at offset `0`, is checked before performing any operations.
```

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,410 @@
...
+if (version == 1) {
+  // Handle version 1 data format
+} else if (version == 2) {
+  // Handle version 2 data format
+}
+```
+
+### Metadata header
+
+A metadata header at the beginning of a region allows efficient tracking of **data version, entry count, and available space**. This structure provides a **centralized reference** for managing stored data, preventing unnecessary scanning of the entire memory region to determine its state.
```
jessiemongeon1, line 238:
```suggestion
A metadata header at the beginning of a `Region` allows efficient tracking of data version, entry count, and available space. This structure provides a centralized reference for managing stored data, preventing unnecessary scanning of the entire `Region` to determine its state.
```

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,410 @@
...
+    version = Region.loadNat64(myRegion, 0);
+    count = Region.loadNat64(myRegion, 8);
+    position = Region.loadNat64(myRegion, 16);
+  }
+}
+```
+
+### Transaction log pattern
+
+A transaction log pattern enables **efficient appends** while maintaining a structured, reliable state. This approach ensures that **new log entries are appended sequentially**, preventing data corruption and supporting an **event-driven architecture** where changes can be tracked over time.
```
jessiemongeon1, line 266:
```suggestion
A transaction log pattern enables efficient appends while maintaining a structured, reliable state. This approach ensures that new log entries are appended sequentially, preventing data corruption and supporting an event-driven architecture where changes can be tracked over time.
```

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,410 @@
...
+    meta.version, 
+    meta.count + 1, 
+    position + 8 + size
+  );
+}
+```
+
+## Example
+
+This example illustrates the simultaneous use of stable variables and stable memory. It uses a single stable variable, `state`, to keep track of the two regions and their size in bytes, but stores the contents of the log directly in stable memory.
```
jessiemongeon1, line 290:
```suggestion
This example illustrates the simultaneous use of stable variables and stable memory. It uses a single stable variable, `state`, to keep track of the two `Region`s and their size in bytes, but stores the contents of the log directly in stable memory.
```

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,410 @@
...
+    Region.storeNat64(state.elems, elem_i * elem_size + 0, elem_pos);
+    Region.storeNat64(state.elems, elem_i * elem_size + 8, Nat64.fromNat(blob.size()));
+    elem_i
+  }
+};
+```
+
+## Performance considerations
+
+Efficient use of stable memory requires optimizing **read and write operations** to reduce overhead. **Batching** multiple small writes into fewer, larger operations improves performance by minimizing the number of memory transactions. Large data structures should be **aligned to page boundaries** whenever possible, as crossing page boundaries can lead to additional processing overhead. Frequently accessed data should be **cached in native memory** to avoid repeated stable memory reads, which are slower than standard heap operations. Since stable memory is persistent, **temporary or short-lived data** should be stored in native memory instead, reducing unnecessary stable memory usage.
```
jessiemongeon1, line 361:
```suggestion
Efficient use of stable memory requires optimizing read and write operations to reduce overhead. Batching multiple small writes into fewer, larger operations improves performance by minimizing the number of memory transactions. Large data structures should be aligned to page boundaries whenever possible, as crossing page boundaries can lead to additional processing overhead. Frequently accessed data should be cached in native memory to avoid repeated stable memory reads, which are slower than standard heap operations. Since stable memory is persistent, temporary or short-lived data should be stored in native memory instead, reducing unnecessary stable memory usage.
```

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+    ``` motoko no-repl
+    let mask : Nat32 = 0xfc03_ff00;
+    let offset : Nat32 = size + 1;
+    ```
+
+### Picking types
+
+- Use [`Nat`](../old-docs/base/Nat.md) for any integral value that cannot be negative.
+
+- Use fixed-width `NatN` or `IntN` only when storing many values and space usage matters, when bit-fiddling requires the low-level interpretation of a number as a vector of bits or when matching types imposed by external requirements, such as other canisters.
```
jessiemongeon1, line 723:
```suggestion
- Use fixed-width `NatN` or `IntN` only when storing many values and space usage matters, when bit-fiddling requires the low-level interpretation of a number as a vector of bits, or when matching types imposed by external requirements such as other canisters.
```

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+
+- Avoid proliferation of option types, and therefore `null`.
+
+    Limit their use to as small a scope as possible. Rule out the `null` case and use non-option types wherever possible.
+
+- Consider using records instead of tuples when there are more than 2 or 3 components. Records are just simple objects with named fields.
+
+    Note that record types need not be declared but can be used in place.
```
jessiemongeon1, line 731:
```suggestion
    Note that record types can be used in place and do not need to be declared.
```

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+
+    ``` motoko no-repl
+      func nodeInfo(node : Node) : {parent : Node; left : Node; right : Node} { ... }
+    ```
+
+- Consider using variants instead of [`Bool`](../old-docs/base/Bool.md) to represent binary choices.
+
+    Note that variant types need not be declared but can be used in place.
```
jessiemongeon1, line 739:
```suggestion
    Note that variant types can be used in place and do not need to be declared.
```

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+    };
+
+    process{config = Config(); data = read(); delta = 0.01; seed = 1.0};
+    ```
+
+    Rationale: This expresses named parameters. This way, arguments can be freely reordered at the call site and callers are prevented from accidentally passing them in the wrong order.
+
+- Higher-order functions, such as functions that take a callback argument, should put the function parameter last.
+
+    Rationale: Makes call sites more readable, and in the absence of currying, there is no point in putting the function first, like you often would in functional languages.
```
jessiemongeon1, line 769:
```suggestion
    Rationale: This makes call sites more readable, and in the absence of currying, there is no point in putting the function first, like you often would in functional languages.
```

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+- Use `for` loops instead of `while` loops for iterating over a numeric range or a container.
+
+    ``` motoko no-repl
+    for (i in Iter.range(1, 10)) { ... };
+    for (x in array.values()) { ... };
+    ```
+
+    Rationale: For loops are less error-prone and easier to read.
+
+- Use `if` or `switch` as expressions where appropriate.
```
jessiemongeon1, line 798: This is too broad; what is considered "appropriate" in this context?

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+
+    ``` motoko no-repl
+    func abs(i : Int) : Int { if (i < 0) -i else i };
+
+    let delta = switch mode { case (#up) +1; case (#dn) -1 };
+    ```
+
+- Motoko requires that all expressions in a block have type `()`, in order to prevent accidentally dropped results.
```
jessiemongeon1, line 806:
```suggestion
- Motoko requires that all expressions in a block have type `()` to prevent accidentally dropped results.
```

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+
+    Use `ignore` to explicitly drop results. Do *not* use `ignore` when it’s not needed.
+
+    ``` motoko no-repl
+    ignore async f();  // fire of a computation
+    ```
+
+- Motoko allows to omit the `return` at the end of a function, because a block evaluates to its last expression.
```
jessiemongeon1, line 814:
```suggestion
- Motoko allows omitting the `return` at the end of a function because a block evaluates to its last expression.
```

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+
+    Use this when a function is short and in "functional" style, that is, the function does not contain complex control flow or side effects.
```
jessiemongeon1, line 816:
```suggestion
    Omit `return` when a function is short and in "functional" style, that is, the function does not contain complex control flow or side effects.
```

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+
+    ```motoko no-repl
+    let buf = Buffer<Nat>(10);
+    buf.add(42);
+    let arr = Buffer.toArray<Nat>(buf);
+    ```
+
+## Motoko grammar
+
+This section describes the concrete syntax, or grammar, of Motoko. The specification is auto-generated with a tool. This grammar is is in the [Extended Backus–Naur form(EBNF)](https://tomassetti.me/ebnf/).
```
jessiemongeon1, line 962:
```suggestion
This section describes the concrete syntax (grammar) of Motoko. The specification is auto-generated with a tool. This grammar is in the [Extended Backus–Naur form(EBNF)](https://tomassetti.me/ebnf/).
```

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+      case (#possiblity1) { processPossibility1() };
+      case (#possiblity2) { processPossibility2() }
+      case (#possiblity2) { processPossibility3() };
+      case (_) { handleExecption(err) };;
+    };
+    ```
+
+### Data structures
+
+- Prefer `Buffer` over dynamic array resizing
```
jessiemongeon1, line 951:
```suggestion
- Prefer `Buffer` over dynamic array resizing.
```

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+
+    Rationale: Only records can be sent as message parameters or results and can be stored in stable variables. Objects with methods are also more expensive to create and represent in memory.
+
+- Use full objects only as a means for encapsulating state or behavior.
+
+### Classes
+
+- Use `class` to create multiple objects of the same shape.
+
+- Name classes after their conceptual functionality, not their implementation, except when having to distinguish multiple different implementations of the same concept. For example, `OrderedMap` vs `HashMap`).
```
jessiemongeon1, line 860:
```suggestion
- Name classes after their conceptual functionality, not their implementation, except when having to distinguish multiple different implementations of the same concept. For example, `OrderedMap` vs `HashMap`.
```

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+
+    Do not use classes unless both these roles are intended; use plain type aliases or functions returning an object in other cases.
+
+- Do not overuse classes.
+
+    Use a module defining a plain type and functions on it where appropriate. Use classes only as a means for encapsulating state or behavior.
+
+    Rationale: Objects with methods have disadvantages over simple record types with separate functions (see above).
+
+- If values of a class are meant to be sendable (shared), the class needs to provide a pair of `share`/`unshare` methods that convert to/from a sharable representation, for example, as a record.
```
jessiemongeon1, line 872:
```suggestion
- If values of a class are meant to be shared, the class needs to provide a pair of `share`/`unshare` methods that convert to/from a sharable representation, for example, as a record.
```

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+
+    :::note
+
+    For immutable classes it may seem more natural to make `unshare` a kind of static function. However, even for immutable ones it may depend on constructor arguments (such as an ordering function), so that the a pattern like `Map(compareInt).unshare(x)` seems appropriate.
+
+    :::
+
+- For the time being, avoid overloading classes with too many methods, since that is currently expensive.
```
jessiemongeon1, line 880:
```suggestion
- Avoid overloading classes with too many methods, since that is expensive.
```

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+
+    Restrict to a sufficiently small set of canonical methods and make less essential ones that can be implemented on top of those into functions in the enclosing module.
```
jessiemongeon1, line 882:
```suggestion
    Restrict to a sufficiently small set of canonical methods. Make less essential methods that can be implemented on top of the canonical methods into functions in the enclosing module.
```

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+
+- Use `module` to group definitions, including types, and create a name space for them.
+
+- Where applicable, name modules after the main type or class they implement or provide functions for.
+
+- Limit each module to a single main concept/type/class or closely entangled family of concepts/types/classes.
+
+### Error handling
+
+- Use `Result` for error handling where applicable
```
jessiemongeon1, line 896:
```suggestion
- Use `Result` for error handling where applicable.
```

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+    ```motoko no-repl
+    type MyError = { #NotFound; #InvalidInput; #PermissionDenied };
+
+    func safeDivide(a : Int, b : Int) : Result<Int, MyError> {
+      if (b == 0) { return #err(#InvalidInput) };
+      return #ok(a / b);
+    };
+    ```
+
+- Use `assert` only for programming errors
```
jessiemongeon1, line 909:
```suggestion
- Use `assert` only for programming errors.
```

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+  - `assert` should be reserved for conditions that should never fail if the program is correct.
+
+    ```motoko no-repl
+    assert(x > 0);  // Only use if x should never be <= 0
+    ```
+
+### Concurrency
+
+- Use `await` judiciously
```
jessiemongeon1, line 918:
```suggestion
- Use `await` judiciously.
```

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+  - Minimize the number of `await` calls in a function to avoid unnecessary suspensions.
+
+    ```motoko no-repl
+    let user = await db.getUser(id); // Good ~4s
+    let orders = await db.getOrders(user); // Bad: sequential calls introduce high latency ~6s
+    ```
+
+  - Instead, parallelize calls when possible:
```
jessiemongeon1, line 926:
```suggestion
  - Instead, use parallel calls when possible:
```

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+
+    ```motoko no-repl
+    let (user, orders) = await (db.getUser(id), db.getOrders(id));
+    ```
+
+- Use composite queries where applicable.
```
jessiemongeon1, line 932: This is too broad and needs more context

doc/md/motoko-style-guide.md
```diff
@@ -0,0 +1,1302 @@
...
+
+### Pattern matching
+
+- Use `switch` instead of manual conditionals for variants
```
jessiemongeon1, line 936:
```suggestion
- Use `switch` instead of manual conditionals for variants.
```

doc/md/motoko-compiler-reference.md
```diff
@@ -0,0 +1,94 @@
...
+| `--stable-types`                          | Compile binary and emit signature of stable types to `.most` file.                                                                                    |
+| `--stable-compatible <pre> <post>`        | Test upgrade compatibility between stable-type signatures `<pre>` and `<post>`.                                                                       |
+| `--rts-stack-pages <n>`                   | Set maximum number of pages available for runtime system stack (only supported with classical persistence, default 32).                               |
+| `--trap-on-call-error`                    | Trap, don't throw an [`Error`](../base/Error.md), when an IC call fails due to destination queue full or freezing threshold is crossed.               |
+|                                           | Emulates behaviour of moc versions < 0.8.0.                                                                                                           |
```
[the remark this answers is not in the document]
Demali-876: I will remove this because there seems to be no command listed associated with it. I double checked the compiler and it outputs the same thing with this out of place. Maybe this is something we can ask the language team.

doc/md/icp-features/stable-memory.md
Demali-876: Oh well scratch that lol. A few fancy diagrams is not worth a 3-4 week wait. Onward.

doc/md/icp-features/stable-memory.md
```diff
@@ -2,4 +2,411 @@
...
+};
+```
+
+## Performance considerations
+
+Efficient use of stable memory requires optimizing read and write operations to reduce overhead. Batching multiple small writes into fewer, larger operations improves performance by minimizing the number of memory transactions. Large data structures should be aligned to page boundaries whenever possible, as crossing page boundaries can lead to additional processing overhead. Frequently accessed data should be cached in native memory to avoid repeated stable memory reads, which are slower than standard heap operations. Since stable memory is persistent, temporary or short-lived data should be stored in native memory instead, reducing unnecessary stable memory usage.
+
+### Memory usage estimation
+
+Estimating the amount of memory needed requires calculating the expected storage requirements based on entry size and quantity. The total memory requirement is determined by multiplying the average entry size by the number of expected entries, then adding an overhead buffer for metadata and alignment, typically between 10–20%. The resulting size is then divided by 65)536 bytes (the size of a page) to determine the number of pages required.
```
jessiemongeon1, line 368:
```suggestion
Estimating the amount of memory needed requires calculating the expected storage requirements based on entry size and quantity. The total memory requirement is determined by multiplying the average entry size by the number of expected entries, then adding an overhead buffer for metadata and alignment, typically between 10–20%. The resulting size is then divided by 65_536 bytes (the size of a page) to determine the number of pages required.
```

doc/md/icp-features/system-functions.md
```diff
@@ -4,8 +4,177 @@ sidebar_position: 6
...
-preupgrade
-postupgrade
-inspect
+ICP supports five system functions that canisters can call to interact with the ICP runtime environment:
+
+- `heartbeat`
+- `timer`
+- `preupgrade`
+- `postupgrade`
+- `inspect`
```
jessiemongeon1, line 13:
```suggestion
- [`heartbeat`](https://internetcomputer.org/docs/references/ic-interface-spec#heartbeat)
- [`timer`](https://internetcomputer.org/docs/references/ic-interface-spec#global-timer)
- [`preupgrade`](#preupgrade)
- [`postupgrade`](#postupgrade)
- [`inspect`](https://internetcomputer.org/docs/references/ic-interface-spec#system-api-inspect-message)
```

doc/md/icp-features/system-functions.md
```diff
@@ -4,8 +4,177 @@ sidebar_position: 6
...
+
+Declaring any other system function will result in an error. Canisters can use these functions to efficiently manage state transitions, automate tasks, or handle system-level operations. 
+
+## `heartbeat()`
+
+Canisters can opt to receive heartbeat messages by exposing a `canister_heartbeat` function. In Motoko, this is achieved by declaring the system function `heartbeat`, which takes no arguments and returns an asynchronous unit type (`async ()`). Since `heartbeat()` is async, it can invoke other asynchronous functions and await their results. This function executes on every **subnet heartbeat**, enabling periodic task execution without requiring external triggers.
```
jessiemongeon1, line 19:
```suggestion
Canisters can opt to receive [heartbeat messages](https://internetcomputer.org/docs/building-apps/network-features/periodic-tasks-timers#heartbeats) by exposing a `canister_heartbeat` function. In Motoko, this is achieved by declaring the system function `heartbeat`, which takes no arguments and returns an asynchronous unit type (`async ()`). Since `heartbeat()` is async, it can invoke other asynchronous functions and await their results. This function executes on every **subnet heartbeat**, enabling periodic task execution without requiring external triggers.
```

doc/md/icp-features/system-functions.md
```diff
@@ -4,8 +4,177 @@ sidebar_position: 6
...
+
+If a canister exports a function named `canister_heartbeat`, it must have the type `() -> ()`, ensuring it adheres to the expected [system function signature](https://internetcomputer.org/docs/references/ic-interface-spec#heartbeat).
+
+One example use case for heartbeats is a recurring alarm, where the canister performs an action at regular intervals. In the example below, the canister triggers an event every fifth heartbeat by maintaining a counter:
```
jessiemongeon1, line 23:
```suggestion
One example use case for [heartbeats](https://internetcomputer.org/docs/building-apps/network-features/periodic-tasks-timers#heartbeats) is a recurring alarm, where the canister performs an action at regular intervals. In the example below, the canister triggers an event every fifth heartbeat by maintaining a counter:
```

doc/md/icp-features/system-functions.md
```diff
@@ -4,8 +4,177 @@ sidebar_position: 6
...
+}
+```
+
+The heartbeat function executes asynchronously, meaning it can issue inter-canister calls and await responses. However, because every async call in Motoko introduces a context switch, the actual execution of the heartbeat function may be delayed relative to when the subnet issues the heartbeat. The function's result is ignored, so any errors or traps do not affect subsequent executions.
+
+Since subnet heartbeats operate at the protocol level, their timing is not precise and depends on network conditions and execution load. As a result, using heartbeats for **high-frequency or time-sensitive operations should be done cautiously**, as there is no guarantee of real-time execution.
+
+## `timer(setGlobalTimer : Nat64 -> ())`
+
+The `timer()` system function allows canisters to schedule a single execution of a task after a specified delay. To make it repeat, the function must explicitly reset the timer using `setGlobalTimer()` within its body. It takes a single argument for resetting the global timer and returns `async ()`.
```
jessiemongeon1, line 52:
```suggestion
The [`timer()` system function](https://internetcomputer.org/docs/building-apps/network-features/periodic-tasks-timers#timers) allows canisters to schedule a single execution of a task after a specified delay. To make it repeat, the function must explicitly reset the timer using `setGlobalTimer()` within its body. It takes a single argument for resetting the global timer and returns `async ()`. 
```

doc/md/icp-features/system-functions.md
```diff
@@ -4,8 +4,177 @@ sidebar_position: 6
...
+
+Since subnet heartbeats operate at the protocol level, their timing is not precise and depends on network conditions and execution load. As a result, using heartbeats for **high-frequency or time-sensitive operations should be done cautiously**, as there is no guarantee of real-time execution.
+
+## `timer(setGlobalTimer : Nat64 -> ())`
+
+The `timer()` system function allows canisters to schedule a single execution of a task after a specified delay. To make it repeat, the function must explicitly reset the timer using `setGlobalTimer()` within its body. It takes a single argument for resetting the global timer and returns `async ()`. 
+
+Unlike `heartbeat()`, which runs automatically on every subnet tick, `timer()` requires **manual rescheduling** after execution. This gives canisters precise control over whether the timer runs once or continuously, depending on when (or if) `setGlobalTimer()` is called again.
```
jessiemongeon1, line 54:
```suggestion
Unlike `heartbeat()`, which runs automatically on every subnet tick, `timer()` requires **manual rescheduling** after execution. This gives canisters precise control over whether the [timer](https://internetcomputer.org/docs/building-apps/network-features/periodic-tasks-timers#timers) runs once or continuously, depending on when (or if) `setGlobalTimer()` is called again.
```

doc/md/icp-features/system-functions.md
```diff
@@ -4,8 +4,177 @@ sidebar_position: 6
...
+}
+```
+
+The **use of upgrade hooks is not recommended** as they can fail and cause the program to enter an unrecoverable state. With advancements in orthogonal persistence, these hooks are expected to be deprecated.
+
+`preupgrade()` can make a canister trap and fail to upgrade. In many cases, stable variables or actor initialization expressions can replace `postupgrade()`. Complex transformations increase the risk of errors and failures.
+
+## `inspect()`
+
+The `inspect()` system function allows a canister to **inspect ingress messages** before execution, determining whether to accept or reject them. The function receives a record of message attributes, including:
```
jessiemongeon1, line 133:
```suggestion
The [`inspect()` system function](https://internetcomputer.org/docs/references/ic-interface-spec#system-api-inspect-message) allows a canister to **inspect ingress messages** before execution, determining whether to accept or reject them. The function receives a record of message attributes, including:
```

doc/md/icp-features/system-functions.md
```diff
@@ -4,8 +4,177 @@ sidebar_position: 6
...
+
+ - The **caller’s principal**
+ - The **raw argument blob**
+ - A **variant identifying the target function**
+ 
+It returns a `Bool`, where `true` permits execution and `false` rejects the message. Similar to a **query**, any side effects are discarded. If `inspect()` traps, it is equivalent to returning `false`. Unlike other system functions, the argument type of `inspect()` depends on the actor's exposed interface, meaning it can selectively handle different methods or ignore unnecessary fields. 
+
+However, **`inspect()` should not be used for definitive access control** because it runs on a **single replica without going through consensus**, making it susceptible to boundary node spoofing. Additionally, `inspect()` only applies to **ingress messages**, not inter-canister calls, meaning secure access control must still be enforced within shared functions.
```
jessiemongeon1, line 141:
```suggestion
However, `inspect()` should not be used for definitive access control because it runs on a single replica without going through consensus, making it susceptible to boundary node spoofing. Additionally, `inspect()` only applies to [ingress messages](https://internetcomputer.org/docs/building-apps/essentials/message-execution), not [inter-canister calls](https://internetcomputer.org/docs/references/async-code), meaning secure access control must still be enforced within shared functions.
```

doc/md/icp-features/system-functions.md
```diff
@@ -4,8 +4,177 @@ sidebar_position: 6
...
+
+## `inspect()`
+
+The `inspect()` system function allows a canister to **inspect ingress messages** before execution, determining whether to accept or reject them. The function receives a record of message attributes, including:
+
+ - The **caller’s principal**
+ - The **raw argument blob**
+ - A **variant identifying the target function**
+ 
+It returns a `Bool`, where `true` permits execution and `false` rejects the message. Similar to a **query**, any side effects are discarded. If `inspect()` traps, it is equivalent to returning `false`. Unlike other system functions, the argument type of `inspect()` depends on the actor's exposed interface, meaning it can selectively handle different methods or ignore unnecessary fields.
```
jessiemongeon1, line 139:
```suggestion
It returns a `Bool`, where `true` permits execution and `false` rejects the message. Similar to a [query](https://internetcomputer.org/docs/building-apps/essentials/message-execution), any side effects are discarded. If `inspect()` traps, it is equivalent to returning `false`. Unlike other system functions, the argument type of `inspect()` depends on the actor's exposed interface, meaning it can selectively handle different methods or ignore unnecessary fields. 
```

doc/md/icp-features/timers.md
```diff
@@ -2,4 +2,86 @@
 sidebar_position: 2
 ---
 
-# Timers
\ No newline at end of file
+# Timers
+
+Canisters deployed on ICP can schedule code execution at specific intervals using [timers](https://internetcomputer.org/docs/motoko/main/base/Timer). Timers are used to execute periodic tasks such as recurring state mutation, automating maintenance, or enforcing timed access control. 
+
+Motoko provides the `Timer` module, which can be used to create recurring timers to execute code at regular intervals. These timers are managed at the application level.
```
jessiemongeon1, line 9:
```suggestion
Motoko provides the [`Timer`](https://internetcomputer.org/docs/motoko/main/base/Timer) module, which can be used to create recurring timers to execute code at regular intervals. These timers are managed at the application level.
```

doc/md/icp-features/timers.md
```diff
@@ -2,4 +2,86 @@
 sidebar_position: 2
 ---
 
-# Timers
\ No newline at end of file
+# Timers
+
+Canisters deployed on ICP can schedule code execution at specific intervals using [timers](https://internetcomputer.org/docs/motoko/main/base/Timer). Timers are used to execute periodic tasks such as recurring state mutation, automating maintenance, or enforcing timed access control.
```
jessiemongeon1, line 7:
```suggestion
Canisters deployed on ICP can schedule code execution at specific intervals using [timers](https://internetcomputer.org/docs/building-apps/network-features/periodic-tasks-timers). Timers are used to execute periodic tasks such as recurring state mutation, automating maintenance, or enforcing timed access control. 
```

Demali-876:
This PR will be replaced by PR 5031

PR #4959, opened 2025-03-20
LSP: Add parser error recovery (Serokell, Milestone-2)

GoPavel:
In order to be able to work with files that are editing and report multiple syntax errors in the LSP server we have added error recovery to the parser via [serokell/ocaml-recovery-parser](https://github.com/serokell/ocaml-recovery-parser).

Also, the js API has been updated to enable recovery and get recovered AST or fast fail (the argument is optional for backward compatibility):
```js

moc.parseMotoko(file, /* enableRecovery=*/true)
```

This PR follows #4931

### DIFF
--- a/src/js/moc_js.ml
+++ b/src/js/moc_js.ml
@@ -34,6 +34,7 @@ let () =
       method compileWasm mode s = Flags.compiled := true; js_compile_wasm mode s
       method parseCandid s = js_parse_candid s
       method parseMotoko s = js_parse_motoko s
-      method parseMotokoTyped paths = js_parse_motoko_typed paths
+      method parseMotokoWithDeps s = js_parse_motoko_with_deps s
+      method parseMotokoTyped paths scopeCache = js_parse_motoko_typed paths scopeCache
       method printDeps file = print_deps file
      end);
--- a/src/js/moc_js.ml
+++ b/src/js/moc_js.ml
@@ -34,7 +34,7 @@ let () =
       method compileWasm mode s = Flags.compiled := true; js_compile_wasm mode s
       method parseCandid s = js_parse_candid s
       method parseMotoko s = js_parse_motoko s
-      method parseMotokoWithDeps s = js_parse_motoko_with_deps s
+      method parseMotokoWithDeps path s = js_parse_motoko_with_deps path s
       method parseMotokoTyped paths scopeCache = js_parse_motoko_typed paths scopeCache
       method printDeps file = print_deps file
      end);
--- a/src/js/moc_js.ml
+++ b/src/js/moc_js.ml
@@ -33,8 +33,8 @@ let () =
       method stableCompatible pre post = js_stable_compatible pre post
       method compileWasm mode s = Flags.compiled := true; js_compile_wasm mode s
       method parseCandid s = js_parse_candid s
-      method parseMotoko s = js_parse_motoko s
-      method parseMotokoWithDeps path s = js_parse_motoko_with_deps path s
-      method parseMotokoTyped paths scopeCache = js_parse_motoko_typed paths scopeCache
+      method parseMotoko enableRecovery s = js_parse_motoko enableRecovery s
+      method parseMotokoWithDeps enableRecovery path s = js_parse_motoko_with_deps enableRecovery path s
+      method parseMotokoTyped enableRecovery paths scopeCache = js_parse_motoko_typed enableRecovery paths scopeCache
       method printDeps file = print_deps file
      end);
--- a/src/mo_frontend/dune
+++ b/src/mo_frontend/dune
@@ -1,13 +1,33 @@
 (library
   (name mo_frontend)
-  (libraries menhirLib lib lang_utils mo_config mo_def mo_types mo_values wasm_exts)
+  (libraries menhirLib lib lang_utils mo_config mo_def mo_types mo_values wasm_exts ocaml-recovery-parser.menhirRecoveryLib)
+  (modules (:standard \ test_recovery))
   (instrumentation (backend bisect_ppx --bisect-silent yes))
 )
+
 (menhir
   (modules parser assertions)
   (merge_into parser)
-  (flags --table --inspection -v --strict)
+  (flags --table --inspection -v --strict --cmly)
   (infer true)
 )
 
+(rule
+  (targets recover_parser.ml)
+  (deps parser.cmly)
+  (action
+    (with-stdout-to recover_parser.ml
+      (run ocaml-recovery-parser.menhir-recovery-generator parser.cmly)
+    )
+  )
+)
+
 (ocamllex source_lexer)
+
+(library
+  (name test_recovery)
+  (inline_tests)
+  (modules test_recovery)
+  (libraries mo_frontend menhirLib lib lang_utils mo_config mo_def mo_types mo_values wasm_exts ocaml-recovery-parser.menhirRecoveryLib )
+  (flags (:standard -w -40))
+  (preprocess (pps ppx_inline_test ppx_expect)))
--- a/src/mo_frontend/menhir_error_reporting.ml
+++ b/src/mo_frontend/menhir_error_reporting.ml
@@ -150,27 +150,6 @@ module Make
       ) []
     )
 
-  (* We drive the parser in the usual way, but records the last [InputNeeded]
-     checkpoint. If a syntax error is detected, we go back to this checkpoint
-     and analyze it in order to produce a meaningful diagnostic. *)
-
-  (* exception Error of (Lexing.position * Lexing.position) * explanation list *)
-
-  (* TODO: remove to give control of parsing outside *)
-  (* let entry (start : 'a I.checkpoint) lexer = *)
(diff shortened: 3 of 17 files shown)


GoPavel:
I have rebased all branches over the last version of Milestone-1

src/mo_frontend/dune
```diff
@@ -1,13 +1,33 @@
...
+  (libraries menhirLib lib lang_utils mo_config mo_def mo_types mo_values wasm_exts ocaml-recovery-parser.menhirRecoveryLib)
+  (modules (:standard \ test_recovery))
   (instrumentation (backend bisect_ppx --bisect-silent yes))
 )
+
 (menhir
   (modules parser assertions)
   (merge_into parser)
-  (flags --table --inspection -v --strict)
+  (flags --table --inspection -v --strict --cmly)
```
crusso, line 11: what does --cmly do?

GoPavel: generates cmly file: https://gallium.inria.fr/~fpottier/menhir/manual.html#sec%3Asdk

src/mo_frontend/parsing.ml
```diff
@@ -70,39 +70,95 @@ let slice_lexeme lexbuf i1 i2 =
...
+  let default_value (loc: Custom_compiler_libs.Location.t) sym =
+      let open Custom_compiler_libs.Location in
+      let open Lexing in
+      let open Source in
+      let file = loc.loc_start.pos_fname in
+      let region_loc : region = {
+        left  : pos = {file; line = loc.loc_start.pos_lnum; column = loc.loc_start.pos_bol};
+        right : pos = {file; line = loc.loc_end.pos_lnum; column = loc.loc_end.pos_bol};
+      } in
+      default_value region_loc sym
```
crusso, line 100: Is this `default_value` obtained from one of the open modules above? Otherwise I looks non-terminating...

src/mo_frontend/test_recovery.ml
```diff
@@ -0,0 +1,457 @@
...
+let parse_from_lexbuf lexbuf : Mo_def.Syntax.prog Diag.result =
+  let open Mo_frontend in
+  let open Diag.Syntax in
+  let name = "test.mo" in
+  let lexer_mode = Lexer.mode in
+  let lexer, triv_table = Lexer.tokenizer lexer_mode lexbuf in
+  let () = Parser_lib.triv_table := triv_table in
+  let start =  Parser.Incremental.parse_prog lexbuf.Lexing.lex_start_p in
+  let error_details = !Mo_config.Flags.error_detail in
+  let* mk_syntax = Mo_frontend.Parsing.parse ~recovery:true lexer_mode error_details start lexer lexbuf
```
crusso, line 11: These test might be a bit unmaintable and break every time the grammar changes. 
Maybe it would be better to dial down the `error_detail` so that only the unexpected token is reported and you don't get all the extra info in the expected test output.

Just a suggestion, but I think this might be more maintainable (unless you are actually testing the error output rather than the recovered AST).

src/mo_frontend/test_recovery.ml
```diff
@@ -0,0 +1,457 @@
...
+      <relop> <exp_bin(ob)>
+      else <exp_nest>
+      . <id>
+      : <typ_nobin>
+      <binop> <exp_bin(ob)>
+      <binassign> <exp(ob)>
+      and <exp_bin(ob)>
+      <unop> <exp_bin(ob)>
+      <inst> <exp_nullary(ob)>
+      [ <exp(ob)> ]
```
crusso, line 104: I think you could condense this by lowering the error_detail!

test/test-moc.js
```diff
@@ -143,6 +143,18 @@ assert.deepStrictEqual(Motoko.check("bad.mo"), {
...
+    {
+        source: 'bad.mo',
+        severity: 1,
+        range: {
+          start: { line: 0, character: 0 },
+          end: { line: 0, character: 0 }
+        },
+        code: 'M0057',
+        category: 'type',
+        message: 'unbound variable __error_recovery_var__'
```
crusso, line 155: Maybe we should plug holes with values of bottom type None instead? 
Maybe `loop{}` or a special prim of type None?

src/mo_frontend/parser.mly
```diff
@@ -607,7 +614,8 @@ exp_plain :
   | LPAR es=seplist(exp(ob), COMMA) RPAR
     { match es with [e] -> e | _ -> TupE(es) @? at $sloc }
 
-exp_nullary(B) :
+(* recovery comment: force to emit special variable instead of "_" to filter spurious errors *)
+exp_nullary [@recover.expr VarE ("__error_recovery_var__" @~ loc) @? loc] (B) :
```
crusso, line 618: Instead of an unbound variable, that produces a spurious type error, we could insert the ast for 
  ``(primE "trap" : Text -> None) "..."``
or ``loop {}``, both of which have type `None`

src/js/moc_js.ml
```diff
@@ -33,11 +33,11 @@ let () =
...
-         for backwards compability. They're mostly used by the language server. *)
-      method parseMotoko s = js_parse_motoko s
-      method parseMotokoWithDeps path s = js_parse_motoko_with_deps path s
+      method parseMotoko enableRecovery s = js_parse_motoko enableRecovery s
+      method parseMotokoWithDeps enableRecovery path s = js_parse_motoko_with_deps enableRecovery path s
+      (* The method [parseMotokoTyped] is saved for backward compability.
+         New versions of tooling should use [parseMotokoTypedWithScopeCache]. *)
       method parseMotokoTyped paths = js_parse_motoko_typed paths
-      method parseMotokoTypedWithScopeCache paths scopeCache = js_parse_motoko_typed_with_scope_cache paths scopeCache
+      method parseMotokoTypedWithScopeCache enableRecovery paths scopeCache = js_parse_motoko_typed_with_scope_cache enableRecovery paths scopeCache
```
crusso, line 41: Will these extra arguments break older LSP clients? Would it be better to just have new entry points?
Just checking. Maybe this works.

crusso: Or are they already optional when viewed from JS somehow?

src/mo_frontend/menhir_error_reporting.ml
```diff
@@ -150,26 +150,6 @@ module Make
       ) []
     )
 
-  (* We drive the parser in the usual way, but records the last [InputNeeded]
```
crusso, line 153: I don't quite grok why this code has disappeared, was it unused?

crusso (changes requested):
Very nice - no real changes to the grammar!

Made some minor suggestions

src/mo_frontend/test_recovery.ml
GoPavel: I was using `dune runtest --auto-promote`. Does it solve the maintenance problem?

src/mo_frontend/menhir_error_reporting.ml
GoPavel: See https://github.com/dfinity/motoko/pull/4959/commits/95e29ec137cfef6a1b8e7b1315f84d17dda97893
I have inlined it into `parsing.ml`

src/mo_frontend/parsing.ml
GoPavel: This is a bit awkward part of the recovery API where we tried to allow to have any location type.

So `default_value` is generated in `recover_parser.ml` and doesn't restrict location type, so you can use any location type in the `[@recover.expr ..]` attributes of the grammar. This `default_value` will have `Source.region` -> 'a I.symbol -> 'a` type. 

But to fit the signature of `MenhirRecoveryLib.Make`, you have to provide `default_value` of type `Custom_compiler_libs.Location.t -> 'a I.symbol -> 'a`. So we define `default_value` with the right type via the generated converting location type and calling the generated `default_value :: Source.region -> ..`

GoPavel: So those `default_value` are actually different functions, no recursion

src/mo_frontend/test_recovery.ml
crusso: Oh, I doubt anyone is aware of that. Maybe just add a comment to use that at the top of the file?

src/mo_frontend/parsing.ml
crusso, line 100:
```suggestion
      default_value region_loc sym  (* default_value is included from Recover_parser *)
```

crusso (dismissed):
All good then!

src/mo_frontend/parser.mly
GoPavel: `prim "trap"` requires `import Prim "mo:⛔";` at the beginning of the file, doesn't it?

GoPavel: The loop generates another spurious error: 
image

src/mo_frontend/parsing.ml
GoPavel: done

src/mo_frontend/test_recovery.ml
GoPavel: done

crusso (approved):

PR #4963, opened 2025-03-21
feat: extract `let` bindings with manifest function type as honest-to-goodness `func`s

ggreif:
A few become
``` diff
-let toText : Float -> Text
+func toText(_ : Float) : Text
```
which is semantially correct, but with time we can make the argument types to named ones!

Improvement for record types could be
``` diff
-let foo : {a : Nat; …} -> Text
+func foo({a : Nat; …}) : Text
```
(punning trick) but we have very few of these...


src/docs/extract.ml
```diff
@@ -128,13 +128,40 @@ struct
...
     | _ -> []
 
+  let extract_ty_args = function
+    | { it = Syntax.ParT arg; _ } ->
+        [ { name = "_"; typ = Some arg; doc = None } ]
+    | { it = Syntax.NamedT ({ it = name; _ }, arg); _ } ->
+        [ { name; typ = Some arg; doc = None } ]
+    | { it = Syntax.TupT args; _ } ->
+        List.map
+          (fun (id_opt, typ) ->
```
ggreif, line 138: factor this out

rvanasa (approved):
LGTM!

PR #4967, opened 2025-03-24
perf: don't needlessly traverse unexpired nodes

ggreif:
or when we already gathered enough expirations

### DIFF
--- a/src/prelude/internals.mo
+++ b/src/prelude/internals.mo
@@ -589,30 +589,33 @@ func @timer_helper() : async () {
     case null ();
     case (?n) {
       gatherExpired(n.pre);
-      if (n.expire[0] > 0 and n.expire[0] <= now and gathered < thunks.size()) {
-        thunks[gathered] := ?(n.job);
-        switch (n.delay) {
-          case (null or ?0) ();
-          case (?delay) {
-            // re-add the node, skipping past expirations
-            let expire = n.expire[0] + delay * (1 + (now - n.expire[0]) / delay);
-            n.expire[0] := 0;
-            // N.B. reinsert only works on pruned nodes
-            func reinsert(m : ?@Node) : @Node = switch m {
-              case null ({ n with expire = [var expire]; pre = null; post = null });
-              case (?m) {
-                assert m.expire[0] != 0;
-                if (expire < m.expire[0]) ({ m with pre = ?reinsert(m.pre) })
-                else ({ m with post = ?reinsert(m.post) })
-              }
-            };
-            @timers := ?reinsert(@prune(@timers));
+      if (n.expire[0] <= now and gathered < thunks.size()) {
+        if (n.expire[0] > 0) {
+          // not expunged yet
+          thunks[gathered] := ?(n.job);
+          switch (n.delay) {
+            case (null or ?0) ();
+            case (?delay) {
+              // re-add the node, skipping past expirations
+              let expire = n.expire[0] + delay * (1 + (now - n.expire[0]) / delay);
+              n.expire[0] := 0;
+              // N.B. reinsert only works on pruned nodes
+              func reinsert(m : ?@Node) : @Node = switch m {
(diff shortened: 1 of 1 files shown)


src/prelude/internals.mo
```diff
@@ -589,30 +589,33 @@ func @timer_helper() : async () {
...
+            }
           };
+          n.expire[0] := 0;
+          gathered += 1;
         };
-        n.expire[0] := 0;
-        gathered += 1;
-      };
-      gatherExpired(n.post);
+        gatherExpired(n.post)
```
ggreif, line 617: In contrast to prior practice (where we traversed the `post` part unconditionally), we now only attempt to gather from the`post` nodes when
- there are still slots to fill, and
- expiration of the currently visited node is _not_ in the future (i.e. either past/present or expunged).

src/prelude/internals.mo
```diff
@@ -588,31 +588,33 @@ func @timer_helper() : async () {
...
-            func reinsert(m : ?@Node) : @Node = switch m {
-              case null ({ n with expire = [var expire]; pre = null; post = null });
-              case (?m) {
-                assert m.expire[0] != 0;
-                if (expire < m.expire[0]) ({ m with pre = ?reinsert(m.pre) })
-                else ({ m with post = ?reinsert(m.post) })
-              }
-            };
-            @timers := ?reinsert(@prune(@timers));
+      if (n.expire[0] > 0) gatherExpired(n.pre); // by corollary
```
ggreif, line 591: see "Focus node is expunged" in the description

src/prelude/internals.mo
```diff
@@ -588,31 +588,34 @@ func @timer_helper() : async () {
...
+      if (pivot > 0) gatherExpired(n.pre); // by corollary
+      if (pivot <= now and gathered < thunks.size()) {
+        if (pivot > 0) {
+          // not expunged yet
+          thunks[gathered] := ?(n.job);
+          switch (n.delay) {
+            case (null or ?0) n.expire[0] := 0;
+            case (?delay) {
+              // re-add the node, skipping past expirations
+              let expire = pivot + delay * (1 + (now - pivot) / delay);
```
Kamirus, line 601: What if this new `expire` is `<= now`? Should it be gathered again now?

ggreif: This PR doesn't (should't) change the semantics.

src/prelude/internals.mo
```diff
@@ -588,31 +588,34 @@ func @timer_helper() : async () {
...
+              // N.B. reinsert only works on pruned nodes
+              func reinsert(m : ?@Node) : @Node = switch m {
+                case null ({ n with expire = [var expire]; pre = null; post = null });
+                case (?m) {
+                  assert m.expire[0] != 0;
+                  if (expire < m.expire[0]) ({ m with pre = ?reinsert(m.pre) })
+                  else ({ m with post = ?reinsert(m.post) })
+                }
+              };
+              @timers := ?reinsert(@prune(@timers));
```
Kamirus, line 612: This does not have any effect on the ongoing `gatherExpired`, right? `reinsert` is not mutating anything.
So could we avoid calling `prune` and `reinsert` for every gathered job and simply keep nodes that need to be re-added later in a max 10-element array and add them after `gatherExpired` finished? After sorting the array it should be possible to traverse `timers` just once.
Note: this is assuming that re-inserted nodes are not to be gathered immediately.

Kamirus (approved):

src/prelude/internals.mo
ggreif: Right, the current traversal is not touched by reinsertion. And yes we could optimise more, but let's do it in small increments!

ggreif: Also, I expect a very small number of recurrent timers to expire at any point (< 2, I guess). So the benefit is not sure.

PR #4969, opened 2025-03-24
experiment: binary search for regression in `jsoo`

ggreif:
- [x] 5.0.1 ✅
- [x] 5.1.0 🚩
- [x] 5.2.0 🚩
- [x] 5.4.0 🚩
- [x] 5.8.2 🚩

Looks like the problem started with 5.1.0. Below patch fixes the stack overflow (just like #4965)
``` diff
$ git diff
diff --git a/test/test-moc_interpreter.js b/test/test-moc_interpreter.js
index 55ba4697b..7ffe421a7 100644
--- a/test/test-moc_interpreter.js
+++ b/test/test-moc_interpreter.js
@@ -10,7 +10,7 @@ moc.Motoko.saveFile('empty.mo', '');
 moc.Motoko.saveFile('ok.mo', '1');
 moc.Motoko.saveFile('warn.mo', '2 - 1');
 moc.Motoko.saveFile('bad.mo', '1+');
-moc.Motoko.saveFile('limit.mo', 'var i = 0; while (i < 10000) { i += 1 }; i');
+moc.Motoko.saveFile('limit.mo', 'var i = 0; while (i < 7600) { i += 1 }; i');
 moc.Motoko.saveFile('text.mo', `let s = "${'⛔|'.repeat(10000)}"; s.size()`); // #3822
 
 try {

```

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -46,10 +46,10 @@ let
 
               # upgrade `js_of_ocaml(-compiler)` until we have figured out the bug related to 4.1.0 (which is in nixpkgs)
               js_of_ocaml-compiler = super.ocamlPackages.js_of_ocaml-compiler.overrideAttrs rec {
-                version = "5.0.1";
+                version = "5.4.0";
                 src = self.fetchurl {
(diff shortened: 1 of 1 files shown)


ggreif:
I have seen enough. Bye!

PR #4974, opened 2025-03-26
chore: bump OCaml to 4.14

ggreif:


src/lib/lib.ml
```diff
@@ -524,13 +524,6 @@ struct
   end
 end
 
-module Seq =
```
ggreif, line 527: This was dead code to begin with, but now OCaml 4.14 has it.

rvanasa (approved):

PR #4978, opened 2025-03-28
feat: detect lost fields in records (extensions)

### LINKED ISSUE #4976: Unexpected Behavior in with-statement when misspelling field names
I recently encountered an interesting behavior in Motoko that isn't necessarily a bug, but it might be unexpected for an average developer.

When using the with-statement to create a modified version of a record while changing only one property, I discovered that misspelling a field name does not cause a compiler error or warning. Instead, the compiler silently ignores the intended change, which did lead to a debugging roller coaster...

Minimal Reproduction Example:

```
import Text "mo:base/Text";
import Nat "mo:base/Nat";
import Array "mo:base/Array";
import Debug "mo:base/Debug";

persistent actor {
    public type Song = {
        title: Text;
        artist: Text;
        releaseYear: Nat;
        nostalgiaRating: Nat;
        memory: Text;
    };

    var songLibrary : [Song] = [];
    
    public func updateRating(title: Text, newRating: Nat) : async () {
        var exists = false;
        songLibrary := Array.map<Song, Song>(
            songLibrary,
            func song {
                if(Text.toLowercase(song.title) == Text.toLowercase(title)) {
                    exists := true;
                    {song with rating = newRating};  // "rating" does not exist in `Song`
                } else {
                    song;
                };
            },
        );
        if (not exists) {
            Debug.trap("Song does not exist");
        };
    };
};
```

Issue Explanation
- The Song record does not have a field named _rating_, but instead has _nostalgiaRating_.
- The with-statement does not produce a compile-time warning when an invalid field name (rating) is used.
As a result, this update silently fails without modifying the record.

ggreif:
This now warns when a user constructs a record with a field that will be dropped because the result is used as a supertype.

It cross-checks user-provided explicit fields (i.e. ignoring bases) and warns for those that don't appear in the expected type (in checking mode).

Resolves #4976.

TODOs:
- Do we want this for `object`s and `class`es too? (see #4981)

### DIFF
--- a/src/lang_utils/error_codes.ml
+++ b/src/lang_utils/error_codes.ml
@@ -218,4 +218,5 @@ let error_codes : (string * string option) list =
     "M0212", Some([%blob "lang_utils/error_codes/M0212.md"]); (* Unrecognised attribute in parenthetical note *)
     "M0213", None; (* Parenthetical note on shared functions is disallowed *)
     "M0214", None; (* Expected type of field in parenthetical note differs from inferred *)
+    "M0215", None; (* Field is lost in record used at supertype *)
   ]
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -1868,7 +1868,7 @@ and check_exp' env0 t exp : T.typ =
   | TupE exps, T.Tup ts when List.length exps = List.length ts ->
     List.iter2 (check_exp env) ts exps;
     t
-  | ObjE ([], exp_fields), T.Obj(T.Object, fts) -> (* TODO: infer bases? Default does a decent job. *)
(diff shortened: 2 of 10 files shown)


Kamirus:
Does it warn only for these fields that are newly added/updated with the 'with' syntax? Or for all the fields?
Does is still warn when used indirectly via a let-binding?

ggreif:
> Does it warn only for these fields that are newly added/updated with the 'with' syntax? Or for all the fields? Does is still warn when used indirectly via a let-binding?

I improved the description above. Also see the test (`.mo`) and new warnings (`.tc.ok`) which highlight old "mistakes" in the test suite.

Kamirus (approved):
Good extra safety!

PR #4981, opened 2025-03-29
feat: detect ignored public fields

ggreif:


src/mo_frontend/typing.ml
```diff
@@ -2052,6 +2053,20 @@ and detect_lost_fields env t = function
...
+    let pub_fields = pub_fields dec_fields |> snd in
+    let pub_ids = T.Env.keys pub_fields in
+    let [@warning "-8"] T.Obj (_, fts) = t in
+    List.iter
+      (fun id ->
+        match List.find_opt (fun ft -> ft.T.lab = id) fts with
+        | Some _ -> ()
+        | None ->
+           warn env ((T.Env.find id pub_fields).id_region) "M0215"
+             "Public field `%s` is ignored in object of type%a"
```
ggreif, line 2066: we could just pass in the verbal context, here `"object"` or `"class"`, above `"record"` or `"record extension".

src/mo_frontend/typing.ml
```diff
@@ -2048,10 +2049,24 @@ and detect_lost_fields env t = function
          | Some _ -> ()
          | None ->
             warn env fld.at "M0215"
-              "Field `%s` is ignored in record of type%a"
-              id
+              "Field `%s` is ignored in record%s of type%a"
```
crusso, line 2052:
```suggestion
              "field `%s` is provided but not expected in record%s of type%a"
```
Won't this also warn on regular subtyping? That might be annoying.
One thing you could do is detect if there are similar fields in the record type using the Suggest functions, in case this is a typo.

crusso: Maybe it's ok, since you only warn on record/object literals, not at every use of subtyping on records.

src/mo_frontend/typing.ml
```diff
@@ -2048,10 +2049,24 @@ and detect_lost_fields env t = function
...
+    let pub_fields = pub_fields dec_fields |> snd in
+    let pub_ids = T.Env.keys pub_fields in
+    let [@warning "-8"] T.Obj (_, fts) = t in
+    List.iter
+      (fun id ->
+        match List.find_opt (fun ft -> ft.T.lab = id) fts with
+        | Some _ -> ()
+        | None ->
+           warn env ((T.Env.find id pub_fields).id_region) "M0215"
+             "Public field `%s` is ignored in object of type%a"
```
crusso, line 2066:
```suggestion
             "public field `%s` is provided but not expected in object of type%a"
```

src/mo_frontend/typing.ml
```diff
@@ -2048,10 +2049,24 @@ and detect_lost_fields env t = function
...
+              id (if bs = [] then "" else " extension")
               display_typ t)
       flds
+  | ObjBlockE (_exp_opt, { it = Type.Object; _}, _typ_opt, dec_fields) ->
+    let pub_fields = pub_fields dec_fields |> snd in
+    let pub_ids = T.Env.keys pub_fields in
+    let [@warning "-8"] T.Obj (_, fts) = t in
+    List.iter
+      (fun id ->
+        match List.find_opt (fun ft -> ft.T.lab = id) fts with
```
crusso, line 2062:
```suggestion
        match List.find_opt (fun ft -> ft.T.lab = id) fts with
```
This might confuse typ_fields with value fields. Use Type.find_val_field_opt or what ever it's called.

src/mo_frontend/typing.ml
```diff
@@ -2048,10 +2049,24 @@ and detect_lost_fields env t = function
          | Some _ -> ()
```
crusso, line 2049: Don't use List.find_opt on previous line (see below)

src/mo_frontend/typing.ml
ggreif: Yeah, it only warns for those where a field is given by the user (either in the record (extension) and it is suppressed bin the typing context, making it unreachable. (Effects survive obviously).

For this PR a `public` in `object`/`class` is given by the user and _immediately shunned by a user-provided type annotation_. This is an indication that the field is either redundant (like above) or if recursively used in the object it could be made `private`.

src/mo_frontend/typing.ml
ggreif: Oh this is in the type, I'll correct and add tests. Maybe we can also warn for ignored public types?

src/mo_frontend/typing.ml
ggreif: One thing is still missing, namely `let { field } = { field = 4; moot = "MOOT" };` where we don't warn yet. That's because we don't (always) have a pattern type at hand. But we could just reduce to field labels.

src/mo_frontend/typing.ml
ggreif: Good point, thanks!

src/mo_frontend/typing.ml
ggreif: Weird, this test doesn't trigger the warning
``` Motoko
let bar_type_no_warn : { foo : Text } = object { public let foo = "FOO"; public type Quux = Text };
```
is it this that you had in mind?

ggreif: Gotcha: this should warn:
``` Motoko
let bar_type_warn : { type Quux = Text; foo : Text} = object { public let foo = "FOO"; public let Quux = "QUUX"; public type Quux = Text };
```
but doesn't a.t.m.

src/mo_frontend/typing.ml
ggreif: Well thinking about this if the type has type fields, the with record (extension) it will always cause a type error, consider:
``` Motoko
let foo_type_warn : { type Quux = Text; foo : Text} = { foo = "FOO"; Quux = "QUUX" };
```
since we can't define type fields with records...

src/mo_frontend/typing.ml
ggreif: 0e00886b5 fixes this.

crusso (approved):
LGTM!

src/mo_frontend/typing.ml
ggreif: Fixed the code anyway!

PR #4995, opened 2025-04-02
Fix RTS memory check

luc-blaeser:
* Incremental GC: Should not clear tag on evacuation, as tag is looked up without forwarding (e.g. during object size query, which goes over object field hash blob).
* Generational GC: Adjustments for new Rust version


crusso (commented):
Is this the fix for the bug Alex found running benchmarks (see stacktrace below)?


I diagnosed this as a corrupted blob in an object's field blob (the hash_blob). Was the tag just cleared?


2025-04-02 13:30:37.575385701 UTC: [Canister bkyz2-fmaaa-aaaaa-qaaaq-cai] panic occurred in file 'src/types.rs' at line 433: assertion failed: self.is_blob()
Error: Failed update call.
Caused by: The replica returned a rejection error: reject code CanisterError, reject message Error from Canister bkyz2-fmaaa-aaaaa-qaaaq-cai: Canister called `ic0.trap` with message: RTS error: RTS panicked
Canister Backtrace:
ic_trap
rts_trap
motoko_rts::trap_with_prefix::h2b8ee811cfd2f7f3
motoko_rts::rts_trap_with::h3e13edb82f788860
rust_begin_unwind
core::panicking::panic_fmt::he5f324e480c5e012
core::panicking::panic::h735b4e41bffa16f0
motoko_rts::types::Value::as_blob::ha4fe7a3b98c2820c
motoko_rts::types::Object::size::hf97f1c4f555d524e
motoko_rts::types::block_size::h96121eb5cec5e8f1
motoko_rts::gc::incremental::phases::evacuation_increment::EvacuationIncrement$LT$M$GT$::evacuate_object::hc0927fc76975ce84
motoko_rts::gc::incremental::phases::evacuation_increment::EvacuationIncrement$LT$M$GT$::evacuate_partition::h7a6091e74dae9627
motoko_rts::gc::incremental::phases::evacuation_increment::EvacuationIncrement$LT$M$GT$::run::h1647d152aaf9b08b
motoko_rts::gc::incremental::IncrementalGC$LT$M$GT$::empty_call_stack_increment::h9de138ae17614b0d
motoko_rts::gc::incremental::incremental_gc::hcbdd7420c14db6e1
motoko_rts::gc::incremental::schedule_incremental_gc::h75aabba5dc793d68
schedule_incremental_gc
@callback<u>

crusso:
Labellig DO-NOT-MERGE until https://github.com/dfinity/motoko/pull/4931 merges

crusso:
I wonder if the bench numbers are different because of this https://github.com/dfinity/motoko/pull/4952.
Did you regenerate the numbers on a linux machine or x86 emulation?

luc-blaeser:
Yes, fixing the issue detected when testing Alex's PR. But it is unrelated to that one. I believe the issue became apparent, when we removed length field in object but indirect via hash blob.

luc-blaeser:
I generate on MacOS x64, but numbers should be platform-independent (IC instruction counting).

luc-blaeser:
Maybe something is off in our test benchmark, as binary seem unchanged, and instruction numbers change...

luc-blaeser:
Removing "DO-NOT-MERGED", as https://github.com/dfinity/motoko/pull/4931 has been merged.

ggreif:
> I wonder if the bench numbers are different because of this #4952. Did you regenerate the numbers on a linux machine or x86 emulation?

THe bench numbers are only checked on `arm64-darwin` and thus should be accepted there too. I have uploaded them.

luc-blaeser:
> > I wonder if the bench numbers are different because of this #4952. Did you regenerate the numbers on a linux machine or x86 emulation?
> 
> The bench numbers are only checked on `arm64-darwin` and thus should be accepted there too. I have uploaded them. See [#4952 (comment)](https://github.com/dfinity/motoko/pull/4952#issue-2925153463)

Uh, that's a problem for me, because I don't have a arm64-darwin, would need to reinstall my machine without Rosetta due to nix. (I also tried to reinstall nix without x64, but does not work, as dependencies are too deep, if once installed with Rosetta). Maybe we should use linux for bench numbers, to not tie ourselves to Mac.

luc-blaeser:
Also, I think we should revisit, as until now, we did not have platform-specific Wasm builds (at least I never noticed bench differences across Mac and Linux).

crusso (approved):

crusso:
> > > I wonder if the bench numbers are different because of this #4952. Did you regenerate the numbers on a linux machine or x86 emulation?
> > 
> > 
> > The bench numbers are only checked on `arm64-darwin` and thus should be accepted there too. I have uploaded them. See [#4952 (comment)](https://github.com/dfinity/motoko/pull/4952#issue-2925153463)
> 
> Uh, that's a problem for me, because I don't have a arm64-darwin, would need to reinstall my machine without Rosetta due to nix. (I also tried to reinstall nix without x64, but does not work, as dependencies are too deep, if once installed with Rosetta). Maybe we should use linux for bench numbers, to not tie ourselves to Mac.

Yeah, it would be a problem for me too. Maybe we can just disable the bench tests on ci and add modify the workflow  task that publishes perf regressions as a comment on the PR to also include bench (if it doesn't already).

ggreif:
Yeah, if this is inconvenient, then we can certainly change the modus operandi. E.g. I could write a GH action that automatically updates those numbers based on x86_64-linux runs. Then it would appear as a change in the PR done just like automatic hash fixes. Does that sound like a way forward?

PR #4999, opened 2025-04-03
Test low-memory hook edge cases

luc-blaeser:
Enable edge cases for low memory hook that was not working in initial version. Now fixed with recent IC/drun.


luc-blaeser:
Noticed that the cases are still broken. Maybe we have a too old drun version, not including the IC fixes on Wasm low memory hook

PR #5003, opened 2025-04-03
update: Base library doc generation

jessiemongeon1:


flake.lock
```diff
@@ -134,6 +134,22 @@
...
+        "lastModified": 1749116131,
+        "narHash": "sha256-e9tj7Fh63qpfNhii1ySSvld9mDwG0HooEYCOU04RCj0=",
+        "owner": "dfinity",
+        "repo": "new-motoko-base",
+        "rev": "130756f7368562dba24dfabe66046f165410b889",
+        "type": "github"
+      },
+      "original": {
+        "owner": "dfinity",
+        "repo": "new-motoko-base",
```
rvanasa: Renamed the repository:

```suggestion
        "repo": "motoko-core",
        "rev": "130756f7368562dba24dfabe66046f165410b889",
        "type": "github"
      },
      "original": {
        "owner": "dfinity",
        "repo": "motoko-core",
```

flake.nix
```diff
@@ -41,6 +41,10 @@
       url = "github:dfinity/motoko-base/next-moc";
       flake = false;
     };
+    motoko-core-src = {
+      url = "github:dfinity/new-motoko-base";
```
rvanasa: Renamed the repository:

```suggestion
      url = "github:dfinity/motoko-core";
```

flake.nix
```diff
@@ -41,6 +41,10 @@
       url = "github:dfinity/motoko-base/next-moc";
       flake = false;
     };
+    motoko-core-src = {
+      url = "github:dfinity/motoko-core";
```
crusso, line 45: Do we want to do the /next-moc thing with motoko-core too?  I suspect we'll need to when we add new prims

crusso, line 45: Do we want to do the /next-moc thing with motoko-core too?  I suspect we'll need to when we add new prims

rvanasa: We may want to do this. Corresponding PR: https://github.com/dfinity/motoko-core/pull/335

It will be more challenging to get the tests to run with the in-development `moc` version using `mops toolchain`, but we should be able to find a solution for this.

PR #5013, opened 2025-04-05
fix: issue 4309

ggreif:


src/mo_idl/mo_to_idl.ml
```diff
@@ -44,74 +45,88 @@ module MakeState() = struct
...
+    | Principal -> PrincipalT
     | Region
     | Error -> assert false
 
-  let rec typ t =
-    (match t with
-    | Any -> I.PrimT I.Reserved
-    | Non -> I.PrimT I.Empty
+  let rec typ t0 =
+    (match t0 with
```
ggreif, line 83: `function`

src/mo_idl/mo_to_idl.ml
```diff
@@ -44,74 +45,88 @@ module MakeState() = struct
...
         | Def (_, t) ->
            (match (open_ ts t) with
             | Prim p -> prim p
-            | Any -> I.PrimT I.Reserved
-            | Non -> I.PrimT I.Empty
+            | Any -> I.(PrimT Reserved)
+            | Non -> I.(PrimT Empty)
+            | Con (c1, ts1) when ts = [] && Cons.eq c c1 -> failwith "CON EQ"
+            | Con (c1, ts1) when false && List.length ts = 0 && List.length ts1 = 1 && Cons.name c = Cons.name c1 -> failwith (Cons.name c ^ " = Innerer.Credit<Nat>")
+            | Con (c1, ts1) when false && List.length ts = 1 && compare ts ts1 = 0 && Cons.name c = Cons.name c1 -> failwith "CON"
```
ggreif, line 97: just exploring... remove this

src/mo_idl/mo_to_idl.ml
```diff
@@ -44,74 +45,88 @@ module MakeState() = struct
...
+
+
+            | Con (c1, ts1) when compare ts ts1 = 0 && Cons.name c = Cons.name c1 ->
```
ggreif, line 100:
```suggestion
            | Con (c1, ts1) when ts = ts1 && Cons.name c = Cons.name c1 ->
```
would this work?

crusso: I don't think so. Perhaps a better way would be to check the def ( not the expansion) is 
<T0, ... TN> d<T0,..., Tn>
ie
(vs, Con(c1, [Var n,.... Var 0])  where len vs = n, or maybe
(vs, Con(c1, [Var 0,.... Var n])  where len vs = n
depending on how we number our deBruin variables.

Then you would know that c1 is an eta-expansion of c2, probably arising from
type c<T1,....Tn> = c1<T1,..., Tn>

If you checked the names (not stamps) of the constructors are eqaul then you can probably just shorted
type Result<T,E> = Result.Result<T,E>
but not
type ErrorMonad<T, E> = Resut.Result<T,E>

Stupid question, but does Yan's code monomorphize 0-ary constructor applications. That might be silly...

src/mo_idl/mo_to_idl.ml
ggreif: no

src/mo_idl/mo_to_idl.ml
ggreif: it's gone

src/mo_idl/mo_to_idl.ml
ggreif: Abandoned this approach, see PR's description.

ggreif: Btw. I think also 0-ary `Con's need to be "monomorphised", as there already may be a homonymous perviously discovered n-ary `Con` in the `env`.

src/mo_idl/mo_to_idl.ml
```diff
@@ -11,6 +11,9 @@ module I = Idllib.Syntax
 module MakeState() = struct
 
   let env = ref Env.empty
+  let hide = ref Env.empty
```
crusso, line 14: Could `hide` just be a set?

src/mo_idl/mo_to_idl.ml
```diff
@@ -44,74 +48,86 @@ module MakeState() = struct
...
-                   env := Env.add id t !env
-                 end;
-               I.VarT (id @@ no_region))
+              let id = monomorphize_con ts c in
+              if Env.mem id !env then
+                match Env.find id !env with
+                | I.PreT -> I.VarT (id @@ no_region)
+                | I.VarT _ as seen -> seen
+                | seen -> I.VarT (RevMap.find (Cons.name c, seen) !rev @@ no_region)
+              else begin
```
crusso, line 105:
```suggestion
              match Env.find_opt id !env with
              | Some t -> ( match t with
                | I.PreT -> I.VarT (id @@ no_region)
                | I.VarT _ as seen -> seen
                | seen -> I.VarT (RevMap.find (Cons.name c, seen) !rev @@ no_region))
             | None ->
```

ggreif: Yeah, further removed the nested `match`: 385b419f682a39547099189c6666f85721088b8b

crusso (commented):
I still don't quite follow what is going on, but ok..

For point of reference, I tried using the eta-contraction idea here #5029 but didn't drive it to conclusion. There were still some strange names being generated for unrelated reasons, but I couldn't quite see why. Your results look better.

crusso (approved):

src/mo_idl/mo_to_idl.ml
```diff
@@ -44,74 +48,85 @@ module MakeState() = struct
...
-                   env := Env.add id (I.PreT @@ no_region) !env;
-                   let t = typ (normalize t) in
-                   env := Env.add id t !env
-                 end;
-               I.VarT (id @@ no_region))
+              let id = monomorphize_con ts c in
+              match Env.find_opt id !env with
+              | Some PreT -> VarT (id @@ no_region)
+              | Some (VarT _ as seen) -> seen
+              | Some seen -> VarT (RevMap.find (Cons.name c, seen) !rev @@ no_region)
```
crusso, line 103: Why can't this rev lookup fail?

crusso: Ah, because lines 101 and 102 cover the unreversed entries of 105 and 113.

src/mo_idl/mo_to_idl.ml
```diff
@@ -44,74 +48,89 @@ module MakeState() = struct
...
-               if not (Env.mem id !env) then
-                 begin
-                   env := Env.add id (I.PreT @@ no_region) !env;
-                   let t = typ (normalize t) in
-                   env := Env.add id t !env
-                 end;
-               I.VarT (id @@ no_region))
+              let id = monomorphize_con ts c in
+              match Env.find_opt id !env with
+              | Some PreT -> VarT (id @@ no_region)
```
ggreif, line 101: type recursion: we are about to add `id` to the `env` below

src/mo_idl/mo_to_idl.ml
```diff
@@ -44,74 +48,89 @@ module MakeState() = struct
...
+              | Some (VarT _ as seen) ->
+                assert (Set.mem id !hide);
+                seen
+              | Some seen ->
+                assert (Set.mem id !hide |> not);
```
ggreif, line 106: This is a "first seen" typedef (i.e. it has a `rev` entry), so it is safe to look up the canonical monomorphised name.

src/mo_idl/mo_to_idl.ml
ggreif: Added a few asserts here, which clarify the path we have taken below after the recursive `typ` call.

PR #5014, opened 2025-04-06
chore: automatically push accepted `test/bench` numbers on `x86_64-linux`

ggreif:
Looks like our local build environments are too diverse to single out one platform for local testing. So I went with CI-side updating of the numbers, creating commits.

This will require folks to do either a `git pull` at times or a `git pull --rebase` if you added commits in the meantime. Trying to push without that will be rejected. I hope that is acceptable!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,7 +59,7 @@ jobs:
         echo "Branch name: ${{ github.event.pull_request.head.ref }}"
         echo "Branch namehere: ${{ github.head_ref || github.ref_name }}"
         git diff --no-index $(nix-build -A tests.bench)/share test/bench/ok | patch -p1 -R
-        # git switch ${github.event.pull_request.head.ref}
+        git switch ${{ github.event.pull_request.head.ref }}
 
     - name: Commit `test/bench` changes
       if: github.actor != 'dependabot[bot]' && runner.os == 'Linux' && runner.arch == 'X64' && github.event_name == 'pull_request'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,7 +55,7 @@ jobs:
 
     - name: Discover `test/bench` changes
       if: github.actor != 'dependabot[bot]' && runner.os == 'Linux' && runner.arch == 'X64' && github.event_name == 'pull_request'
-      run: git diff --no-index $(nix-build -A tests.bench) test/bench/ok
+      run: git diff --no-index $(nix-build -A tests.bench)/share test/bench/ok | patch -p1 -R
 
     - name: Commit `test/bench` changes
       if: github.actor != 'dependabot[bot]' && runner.os == 'Linux' && runner.arch == 'X64' && github.event_name == 'pull_request'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,6 +17,8 @@ jobs:
         fetch-depth: 0
         # fetch PR commit, not predicted merge commit
         ref: ${{ github.event.pull_request.head.sha }}
+        # This is needed to be able to push and trigger CI with that push
+        token: ${{ secrets.NIV_UPDATER_TOKEN }}
     - uses: cachix/install-nix-action@v31
 
     # We are using the ic-hs-test cachix cache that is also used by
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,7 +51,7 @@ jobs:
     - run: nix-env -iA nix-build-uncached -f nix/
 
     - name: "nix-build"
-      run: nix-build-uncached --max-jobs 4 -A all-systems-go -build-flags -L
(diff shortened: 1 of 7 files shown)


.github/workflows/test.yml
```diff
@@ -17,6 +17,8 @@ jobs:
         fetch-depth: 0
         # fetch PR commit, not predicted merge commit
         ref: ${{ github.event.pull_request.head.sha }}
+        # This is needed to be able to push and trigger CI with that push
+        token: ${{ secrets.NIV_UPDATER_TOKEN }}
```
ggreif, line 21: I'll try without this. But my hunch is that this activates the checkout over `ssh`, which can push securely.

.github/workflows/test.yml
```diff
@@ -51,15 +53,36 @@ jobs:
...
 
     - name: "nix-build"
-      run: nix-build-uncached --max-jobs 4 -A all-systems-go -build-flags -L
+      run: nix-build-uncached --max-jobs auto -A all-systems-go -build-flags -L
+
+    - name: Discover `test/bench` changes
+      if: github.actor != 'dependabot[bot]' && github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository && runner.os == 'Linux' && runner.arch == 'X64' && github.event_name == 'pull_request'
+      run: |
+        echo "Branch name: ${{ github.event.pull_request.head.ref }}"
+        echo "Branch here: ${{ github.head_ref || github.ref_name }}"
```
ggreif, line 62: Debugging. Remove.

.github/workflows/test.yml
```diff
@@ -51,15 +53,36 @@ jobs:
...
+        git diff --no-index $(nix-build -A tests.bench)/share test/bench/ok | patch -p1 -R
+        git switch ${{ github.event.pull_request.head.ref }}
+        git branch --show-current
```
ggreif, line 65: check that we are on the right one (can be removed)

.github/workflows/test.yml
```diff
@@ -51,15 +53,36 @@ jobs:
...
+      if: github.actor != 'dependabot[bot]' && runner.os == 'Linux' && runner.arch == 'X64' && github.event_name == 'pull_request'
+      uses: EndBug/add-and-commit@v9
+      with:
+        default_author: github_actions
+        author_name: Cycle and memory benchmark updater
+        github_token: ${{ secrets.NIV_UPDATER_TOKEN }}
+        committer_name: GitHub Actions
+        committer_email: actions@github.com
+        message: "Updating `test/bench` numbers"
+        commit: -- test/bench/ok
```
ggreif, line 77: this is trying to be selective, not sure if it works as expected with a prior `git add`

default.nix
```diff
@@ -535,14 +543,14 @@ rec {
...
-      perf       = perf_subdir "perf"       [ moc nixpkgs.drun ];
+      perf       = perf_subdir false "perf" [ moc nixpkgs.drun ];
       viper      = test_subdir "viper"      [ moc nixpkgs.which nixpkgs.openjdk nixpkgs.z3_4_12 ];
       # TODO: profiling-graph is excluded because the underlying parity_wasm is deprecated and does not support passive data segments and memory64.
       inherit qc lsp unit candid coverage;
     }
     // nixpkgs.lib.optionalAttrs
-         (system == "aarch64-darwin")
-         (fix_names { bench = perf_subdir "bench" [ moc nixpkgs.drun ic-wasm ];})
+         (system == accept-bench)
```
ggreif, line 552: This way we can test on any arch:
```
nix-build -A tests.bench --arg accept-bench '"aarch64-darwin"'
```

test/bench/ok/region-mem.drun-run.ok
```diff
@@ -1,4 +1,4 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
-debug.print: {heap_diff = 0; instr_diff = 5_286_359_401}
+debug.print: {heap_diff = 0; instr_diff = 5_290_967_401}
```
ggreif, line 3: Background why these number changed: https://github.com/dfinity/motoko/pull/4952

luc-blaeser (approved):
Thank you!

christoph-dfinity (approved):
My knee-jerk reaction is that I'd prefer to not get pushes to my PR branches from automation.

As I haven't actually had to update this number yet, maybe this happens so rarely that I wouldn't care. If it solves the immediate problem we can turn it on and look for a more ergonomic fix if it proves to be irritating.

.github/workflows/test.yml
ggreif: It worked! Maybe zap it from `.github/workflows/update-hash.yml` too?

ggreif (commented):
Okay, so let's merge this for now, if it gets into the way of folks, we'll reconsider...

PR #5032, opened 2025-04-10
Use nixbuild.net for CI

basvandijk:


.github/workflows/test.yml
```diff
@@ -45,12 +57,27 @@ jobs:
         docker-images: true
         swap-storage: false
 
+    - if: matrix.os == 'ubuntu-latest' || matrix.os == 'ubuntu-24.04-arm'
+      name: "nix build"
+      run: |
+        nix build \
+          --print-build-logs \
+          --builders "" \
+          --max-jobs 2 \
```
ggreif, line 66:
```suggestion
          --max-jobs auto \
```
maybe better?

basvandijk: Note that I'm using the remote-store building instructions:
https://github.com/nixbuild/nixbuild-action?tab=readme-ov-file#remote-store-building

.github/workflows/test.yml
```diff
@@ -18,7 +18,18 @@ jobs:
         fetch-depth: 0
         # fetch PR commit, not predicted merge commit
         ref: ${{ github.event.pull_request.head.sha }}
-    - uses: cachix/install-nix-action@v31
+
+    - if: matrix.os == 'ubuntu-latest' || matrix.os == 'ubuntu-24.04-arm'
```
ggreif, line 22:
```suggestion
    - if: ${startsWith(matrix.os, 'ubuntu-')}
```
Similarly in other places too!

.github/workflows/test.yml
```diff
@@ -19,16 +19,16 @@ jobs:
         # fetch PR commit, not predicted merge commit
         ref: ${{ github.event.pull_request.head.sha }}
 
-    - if: matrix.os == 'ubuntu-latest' || matrix.os == 'ubuntu-24.04-arm'
+    - if: ${startsWith(matrix.os, 'ubuntu-')}
```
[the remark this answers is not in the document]
ggreif: Fixed: 5e9004f0f357b340fd7d1aa7c9b4b990aa76f69a.

basvandijk:
> Looks nice, but WDYT of changing these too: https://github.com/dfinity/motoko/actions/runs/19700658265/workflow

Unfortunately, nixbuild.net does not support MacOS.

ggreif:
> Unfortunately, nixbuild.net does not support MacOS.

And what is this? https://github.com/nixbuild/nixbuild-action/blob/master/README.md

basvandijk:
I really wish nixbuild.net would support MacOS. Then it would just be a drop in replacement and you wouldn't need these ugly if-conditions in your workflows.

basvandijk:
> I manually started a release action, but [it failed](https://github.com/dfinity/motoko/actions/runs/19701034578/).

Thanks for checking. I'll push a fix after lunch.

PR #5033, opened 2025-04-10
chore: support for splitting tests between `release` and `debug` builds

alexandru-uta:
This PR splits the CI testing pipeline between release and debug builds, adding more fine-grained control and parallelism for running the tests.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -808,62 +808,68 @@ rec {
       # Get the actual derivations for the selected names
       builtins.map (name: tests.${name}) selected_names;
 
-  # Release version - excludes debug tests.
-  let release-systems-go = nixpkgs.releaseTools.aggregate {
-    name = "release-systems-go";
-    constituents = [
-      moc
-      mo-ide
-      mo-doc
-      didc
-      deser
-      samples
-      rts
-      base-src
-      base-tests
-      base-doc
-      docs
-      report-site
-      shell
-      check-formatting
-      check-rts-formatting
-      check-generated
-      check-grammar
-      check-error-codes
-    ] ++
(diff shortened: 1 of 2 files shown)


default.nix
```diff
@@ -787,6 +787,83 @@ rec {
...
+      # Select which set of names to use
+      selected_names = if type == "debug" then debug_tests else release_tests;
+    in
+      # Get the actual derivations for the selected names
+      builtins.map (name: tests.${name}) selected_names;
+
+  # Release version - excludes debug tests
+  release-systems-go = nixpkgs.releaseTools.aggregate {
+    name = "release-systems-go";
+    constituents = [
```
ggreif, line 814: please use a `let` to use a common list (below too).

default.nix
```diff
@@ -787,6 +787,83 @@ rec {
...
+      check-generated
+      check-grammar
+      check-error-codes
+    ] ++
+    filter_tests "debug" tests  # Only include debug tests.
+    ++ builtins.attrValues js
+    ;
+  };
+
   all-systems-go = nixpkgs.releaseTools.aggregate {
```
ggreif, line 867: we probably don't need this one any more

alexandru-uta: yup, deleted now.

default.nix
alexandru-uta: like that? I am not sure I properly understood the ask.

alexandru-uta: and I am also not sure that what I did now actually works

ggreif: Just factoring it out. I have a commit here, can I push?

alexandru-uta: thanks a lot!

PR #5040, opened 2025-04-14
experiment: Turn Motoko into a Nix Flake (based on nixbuild.net)

basvandijk:
* Switch to Nix Flakes which is the modern way to setup Nix projects.

* Switch from the impure [mozilla-nixpkgs](https://github.com/mozilla/nixpkgs-mozilla) overlay to the pure [oxalica/rust-overlay](https://github.com/oxalica/rust-overlay).

* Upgrade the latest stable Rust toolchain from 1.85.0 to 1.86.0.

* Upgrade nixpkgs to release-24.11.


.github/workflows/niv-updater-rare.yml
basvandijk, line 1: Replace `.github/workflows/niv-*.yml` with https://github.com/DeterminateSystems/update-flake-lock.

PR #5041, opened 2025-04-14
chore: get rid of `ic-ref-run`

ggreif:
Don't compile to `.ref.wasm` and don't try to run the comparison.

Remaining vestiges:
- `test/run-drun/actor-import/lg264-qjkae.did` (and users of it)
- `git grep ic-ref`


alexandru-uta (approved):
nice, thanks for the simplification

test/run-drun/actor-import/rwlgt-iiaaa-aaaaa-aaaaa-cai.did
```diff
@@ -0,0 +1,3 @@
+service : {
```
ggreif, line 1: Deal with `test/run-drun/actor-import/lg264-qjkae.did`? Looks like a fossil!

PR #5042, opened 2025-04-15
docs: updates comment on parse_idl_header

christoph-dfinity:
Just something I spotted while reading the Candid subtyping check in the RTS. This function doesn't return a pointer anymore, it updates the passed `buf` instead


crusso (approved):
Indeed!

PR #5046, opened 2025-04-15
chore: updates mergify rules for new CI job names

christoph-dfinity:


crusso (approved):

crusso (commented):
Oh hang on - do we want to require all of those?

christoph-dfinity:
Yes, I think so. They're marked required in Github Actions. (can still change later if I'm wrong)

PR #5049, opened 2025-04-16
chore: modularize and parallelize CI tests

alexandru-uta:
Add more parallelism and modularity by splitting the tests into multiple categories:
1. Common tests (run on all OS versions)
2. GC tests (run on all OS versions)
3. Regular tests (run on all OS versions x {debug, release} modes.

The PR also adds two "dummy" jobs:
A. First waits for (1) & (2) above
B. Waits for (3) in {ubuntu-latest, macos-latest} x {release, debug} targets.

The two dummy jobs will serve as branch protection targets.

### DIFF
--- a/nix/shell.nix
+++ b/nix/shell.nix
@@ -48,6 +48,7 @@ pkgs.mkShell {
         pkgs.openjdk
         pkgs.z3_4_12 # for viper dev
         pkgs.difftastic
+        pkgs.drun
       ] ++ pkgs.lib.optional pkgs.stdenv.isDarwin pkgs.darwin.apple_sdk.frameworks.Security
     ));
 


flake.nix
```diff
@@ -198,14 +207,36 @@
...
 
+        # Common tests version - includes non-GC, non-release/debug specific tests.
+        common-systems-go = pkgs.releaseTools.aggregate {
+          name = "common-systems-go";
+          constituents =
+            pkgs.lib.attrValues common-constituents ++
+            pkgs.lib.attrValues checks ++
+            pkgs.lib.attrValues buildableReleaseMoPackages ++
+            filter_tests "common"  # Only include common tests
+            ++ builtins.attrValues js;
```
basvandijk, line 218: Why do these aggregate jobs need to build all of those common packages? Wouldn't it be better to just have:
```nix

        # Common tests version - includes non-GC, non-release/debug specific tests.
        common-systems-go = pkgs.releaseTools.aggregate {
          name = "common-systems-go";
          constituents = filter_tests "common";  # Only include common tests
        };
```

basvandijk (commented):
Just making some mental notes:

Note that if you parallelise things on the job level it means you'll have duplicate builds since both the `common-tests`, `gc-tests` and `tests` all depend on common targets like the `rts`, `moc`, etc. So these things all start building at the same time.

Of course there's a cache but if the same things start building at the same time a cache won't help that much.

This will increase parallelism of course so it might be the way to go but it will increase total compute. But that might not matter that much.

The easiest way to increase parallelism is to scale vertically by using bigger GitHub runners with more CPUs.

Then you could scale horizontally by letting Nix distribute the builds over  multiple builders. This is what we get with nixbuild.net. Unfortunately they only support Linux.

flake.nix
alexandru-uta: Ah, yes, that's correct, I just overlooked that. Will fix.

alexandru-uta:
> The easiest way to increase parallelism is to scale vertically by using bigger GitHub runners with more CPUs.
> 
> Then you could scale horizontally by letting Nix distribute the builds over multiple builders. This is what we get with nixbuild.net. Unfortunately they only support Linux.

Yup, I know, and I would do that but then macos would still be an issue (and it's still the slowest target, especially on x86.)

flake.nix
```diff
@@ -198,25 +207,37 @@
 
         release-files = import ./nix/release-files.nix { inherit self pkgs; };
 
-        # Release version - excludes debug tests
+        # Common tests version - includes non-GC, non-release/debug specific tests.
+        common-systems-go = pkgs.releaseTools.aggregate {
```
basvandijk, line 211: What about using the following name which also matches its corresponding job:
```suggestion
        common-tests = pkgs.releaseTools.aggregate {
```

flake.nix
```diff
@@ -198,25 +207,37 @@
...
+          name = "common-systems-go";
+          constituents = filter_tests "common";  # Only include common tests.
+        };
+
+        # GC tests version - only includes GC tests.
+        gc-systems-go = pkgs.releaseTools.aggregate {
```
basvandijk, line 217: What about using the following name which also matches its corresponding job:
```suggestion
        gc-tests = pkgs.releaseTools.aggregate {
```

basvandijk:
> but then macos would still be an issue (and it's still the slowest target, especially on x86.)

Yeah that's a bummer. Even switching to https://namespace.so/ (which we use for dfinity/ic) wouldn't help here since for macos they only support aarch64.

BTW have you considered [macos-latest-large](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/about-larger-runners#about-macos-larger-runners) GitHub hosted runners?

alexandru-uta:
> BTW have you considered [macos-latest-large](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/about-larger-runners#about-macos-larger-runners) GitHub hosted runners?

I have, let's see where we get with this -- I also have some future work ideas of running a merge queue, perhaps we can put there the longer/heavier stuff and on just regular CI we run fewer tests and not debug for mac.

flake.nix
alexandru-uta: yup, fixed.

flake.nix
alexandru-uta: thanks, fixed.

nix/tests.nix
```diff
@@ -214,12 +214,11 @@ let
...
+    run-release = test_subdir "run" [ moc ];
     run-debug = snty_subdir "run" [ moc ];
     run-eop-release = enhanced_orthogonal_persistence_subdir "run" [ moc ];
     run-eop-debug = snty_enhanced_orthogonal_persistence_subdir "run" [ moc ];
-    # ic-ref-run = test_subdir "run-drun"   [ moc ic-ref-run ];
-    drun = test_subdir "run-drun" [ moc pkgs.drun ];
+    drun-release = test_subdir "run-drun" [ moc pkgs.drun ];
     drun-debug = snty_subdir "run-drun" [ moc pkgs.drun ];
     drun-compacting-gc = snty_compacting_gc_subdir "run-drun" [ moc pkgs.drun ];
     drun-generational-gc = snty_generational_gc_subdir "run-drun" [ moc pkgs.drun ];
```
ggreif, line 224: `snty_` is debug too

alexandru-uta: I know, but they go in this separate gc section now. Before they were in release mode anyhow.

ggreif:
> > but then macos would still be an issue (and it's still the slowest target, especially on x86.)
> 
> Yeah that's a bummer. Even switching to https://namespace.so/ (which we use for dfinity/ic) wouldn't help here since for macos they only support aarch64.
> 
> BTW have you considered [macos-latest-large](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/about-larger-runners#about-macos-larger-runners) GitHub hosted runners?

`-latest` is probably Arm64. There exist no fast x64 builders.

We should nudge `dfx` to finally ship with Arm in Mac, and then we only need to use `Darwin-x64` builders for PRs that are on a `release-<version>` branch (or some highly job).

ggreif (approved):
LGTM, we had a zoom review!

PR #5054, opened 2025-04-16
experiment: elide stamps when printing, hash stamps when needed (refinement of #5408)

crusso:
version of #5048 that 

* elides type constructor stamps from error messages entirely 
* uses the OCaml hash of stamps (optional filename + localized stamp) to distinguish type constructors when it really matters (e.g. producing stable signatures)

A future PR could lazily allocate distinct display names to constructors as they are required, for error messages only, to show distinguishing logical stamps that are textually small (not including larger hashes).

* With this approach, there's a chance that the hashes will collide (e.g. in stable signature files) and fail to distinguish some types, but it's probably very unlikely.

- [ ] Once this passes CI, introduce a `show_scopes` flag, separate from `show_stamps`, to print scope parameters on async types (some error messages are now confusing because the scopes no  longer appear when needed)


test/cmp/files/version0.drun.most
```diff
@@ -0,0 +1,6 @@
+// Version: 1.0.0
```
crusso, line 1: Delete these files!

test/test-moc.js
```diff
@@ -149,8 +149,8 @@ assert.deepStrictEqual(Motoko.check("bad.mo"), {
 
 // Run interpreter
 assert.deepStrictEqual(Motoko.run([], "actor.mo"), {
-  stdout:
-    "`ys6dh-5cjiq-5dc` : actor {main : shared query () -> async A__9<Text>}\n",
+  stdout: "`ys6dh-5cjiq-5dc` :\n" +
+    "  actor {main : shared query () -> async A<Text>}\n",
```
crusso, line 153:
```suggestion
  stdout: "`ys6dh-5cjiq-5dc` :  actor {main : shared query () -> async A<Text>}\n",
```

test/run-drun/ok/scope-example-implicit.tc.ok
```diff
@@ -1,4 +1,4 @@
 scope-example-implicit.mo:10.12-10.19: type error [M0096], expression of type
```
crusso, line 1: Ok, this error is now a little mysterious...

crusso: I wonder if the show_stamps flag is used to suppress scopes...

crusso: (yep, it is). Added a TODO

crusso:
Superseded by #5055

PR #5055, opened 2025-04-16
feat: type constructors: use per-file type constructor ids, elide type stamps when printing, hash stamps when needed

crusso:
Instead of using a global counter to generate type constructor ids, use one counter per source file, storing the source and relative id in the stamp.

* add flag show_scopes
* precompute hash of name and stamp for printing and faster constructor comparison

Improves error messages, makes VSCode dependency caching work better.

builds on #5054 and #5048 

(detection of other `src/prelude` modules than "prelude" look unnecessary so omitted for now.)


crusso:
CI keeps failing. Merging with master for improved CI.

christoph-dfinity (approved):
Looks reasonable. Only thing that's slightly worrying is if there are any spots we missed calling `Cons.session`...

Maybe I'll see if I kind find an incremental path towards making `stamps` an argument we pass around...

PR #5056, opened 2025-04-17
feat: enable unqualified import of types

christoph-dfinity:
## Overview

Allows pattern matching on type fields of modules and objects.

```motoko
import { type Result } "mo:base/Result";

func parse(input : Text) : Result<Nat, Text> {
  ...
}

// More contrived example

func example() {
  class Example() {
    public type T = Text;
  };
  let ex : Example = Example();
  let { type T } = ex;
  let x : T = "Hello";
};
```

## Design considerations

The current implementation does not require specifying type parameters on the type field pattern:

```motoko
import { type Result } "...";
// As opposed to:
import { type Result<Ok, Err> } "...";
```

The more conservative choice would be to require to specify the parameters and their bounds, but so far I haven't come up with an example where this is a problem. We might want to require it anyway, as it would be a breaking change to require them at a future time.

At the moment the implementation only resolves these type fields patterns on `let`'s (and imports), and requires a restricted "value path" on the rhs of the let. We should be able to relax these restrictions in the future, but for now this lets us cover the important use-cases while keeping implementation complexity down. Relaxing these restrictions would be a non-breaking change.

## Testing

I added a few test cases to the test suite and made sure this continues to compile the "big package set".

Ran a couple manual tests for candid generation, and everything seemed to be fine.

## Reviewer notes

There are a few places with NOTE or TODO comments left, I'd like to discuss those during the review.


src/mo_frontend/typing.ml
```diff
@@ -1201,20 +1207,25 @@ let error_bin_op env at t1 t2 =
...
 
 let rec combine_pat_fields_srcs env t tfs (pfs : pat_field list) : unit =
   match tfs, pfs with
   | _, [] | [], _ -> ()
   | T.{lab; typ = Typ _; _}::tfs', _ ->  (* TODO: remove the namespace hack *)
     combine_pat_fields_srcs env t tfs' pfs
-  | T.{lab; typ; src}::tfs', pf::pfs' ->
-    match compare pf.it.id.it lab with
+  | T.{lab; typ; src}::tfs', {it = TypPF(_); _}::pfs' ->
+    combine_pat_fields_srcs env t tfs' pfs'
```
christoph-dfinity: Note to self: This seems wrong

src/mo_frontend/coverage.ml
```diff
@@ -300,16 +302,19 @@ and match_tup ctxt descs_r descs pats ts sets =
...
+
 and match_obj ctxt ldescs (pat_fields : pat_field list) tfs sets =
   match pat_fields with
   | [] -> succeed ctxt (Obj ldescs) sets
-  | pat_field::pat_fields' ->
-    let l = pat_field.it.id.it in
+  | {it = VarPF(id, p); _}::pat_fields' ->
+    let l = id.it in
+    (* TODO: This seems wrong even before the addition of typed pattern fields? *)
     let tf = List.find (fun tf -> tf.T.lab = l) tfs in
```
crusso: Yeah, I think this should be using Type.lookup_val_field or similar, otherwise it might find a type field of the same name looking for a value field.

src/mo_frontend/typing.ml
```diff
@@ -3221,27 +3318,42 @@ and gather_dec env scope dec : Scope.t =
...
     error_duplicate env "" id;
   T.Env.add id.it (T.Pre, id.at, val_kind) ve
 
+and gather_typ_id env scope id : Scope.t =
+  let open Scope in
+  if T.Env.mem id.it scope.typ_env then
+    error_duplicate env "type " id;
+  (* NOTE: If we decide to require specifying the arity and bounds of
+  type constructors on type pattern fields we need to record them here *)
+  let pre_k = T.Abs ([], T.Pre) in
```
```suggestion
  let pre_k = T.Def ([], T.Pre) in
```
Just for consistency, I would expect these to be Def, not Abs (which we use for type parameters). Don't think it matters though.

I'm a little suprised you can get away with the incorrect arity [] though...

crusso: (Maybe we should add a placeholder kind PreK for unkown...)

src/mo_frontend/typing.ml
```diff
@@ -3280,7 +3392,14 @@ and infer_dec_typdecs env dec : Scope.t =
...
     )
-  | LetD _ | ExpD _ | VarD _ ->
+  | LetD (pat, exp, _) ->
+       begin match infer_val_path env exp with
+       | Some t ->
+          let te = check_pat_typ_dec {env with pre = true} t pat in
+          Scope.{empty with typ_env = te}
+       | None -> Scope.empty
+       end
+  | ExpD _ | VarD _ ->
```
crusso, line 3461: Hmm, I wonder if you could also do the same as for LetD for VarD? Might be more consistent, though odd in its own way.

src/mo_def/syntax.ml
```diff
@@ -116,8 +116,17 @@ and pat' =
 *)
 
 and pat_field = pat_field' Source.phrase
-and pat_field' = {id : id; pat : pat}
+and pat_field' =
+  | VarPF of id * pat
```
crusso:
```suggestion
  | ValPF of id * pat
```
I think I'd call this ValPF for value field vs type field. Var is for mutable, and we don't allow pattern matching on that (though we could actually allow naming the location, for which we could reserver VarPF for the future)

src/mo_frontend/typing.ml
```diff
@@ -2528,30 +2543,110 @@ and check_pats env ts pats ve at : Scope.val_env =
...
       display_typ_expand t
   | T.{lab; typ = Typ _; _}::tfs', _ ->  (* TODO: remove the namespace hack *)
     check_pat_fields env t tfs' pfs ve at
-  | T.{lab; typ; src}::tfs', pf::pfs' ->
-    match compare pf.it.id.it lab with
-    | -1 -> check_pat_fields env t [] pfs ve at
-    | +1 -> check_pat_fields env t tfs' pfs ve at
+  | T.{lab; typ; src}::tfs', ({it = VarPF(id, pat); _} as pf)::pfs' ->
+    match compare id.it lab with
+    | -1 -> check_pat_fields env t [] pfs ve pf.at
```
crusso: (yuck, this delays the error to the recursive call by passing [])

src/mo_frontend/typing.ml
```diff
@@ -2528,30 +2543,110 @@ and check_pats env ts pats ve at : Scope.val_env =
 and check_pat_fields env t tfs pfs ve at : Scope.val_env =
   match tfs, pfs with
   | _, [] -> ve
-  | [], pf::_ ->
-    error env pf.at "M0119"
+  | _, {it = TypPF(id); at; _}::pfs' ->
+    if Option.is_none id.note then
```
crusso, line 2566:
```suggestion
    if Option.is_none id.note then
```
This feels a bit dirty. Would be possible to pass in the local type environment (like ve) and check against that?

crusso: Should this code be checking that TypPF(id) has a corresponding type field in tfs? I'm not sure it's actually doing that. Maybe you should follow the pattern for VarPF (assuming both are sorted by field id)

src/mo_frontend/typing.ml
```diff
@@ -1201,20 +1207,25 @@ let error_bin_op env at t1 t2 =
     display_typ_expand t1
     display_typ_expand t2
 
-let compare_pat_field (pf1 : pat_field) (pf2 : pat_field) =
-  compare pf1.it.id.it pf2.it.id.it
+let compare_pat_field pf1 pf2 = match pf1.it, pf2.it with
```
crusso, line 1211: This has to agree with the order of Type.compare_field for the typechecker to work, right? Maybe check it does and add a comment.

src/mo_frontend/typing.ml
```diff
@@ -2431,7 +2446,7 @@ and check_pat_aux' env t pat val_kind : Scope.val_env =
   | ObjP pfs ->
     let pfs' = List.stable_sort compare_pat_field pfs in
```
crusso, line 2461: Any idea why we use List.stable sort here but List.sort with T.compare_field above? Just so we report errors in textual order?

src/mo_frontend/typing.ml
```diff
@@ -2528,30 +2543,110 @@ and check_pats env ts pats ve at : Scope.val_env =
...
+      let te' = T.Env.add id.it t' te in
+      match pfs' with
+      | {it = TypPF(id'); at = at';_}::_ when id'.it = lab ->
+        error env at' "M0121" "duplicate type field %s in object pattern" lab
+      | _ -> check_pat_fields_typ_dec env t tfs' pfs' te' at
+    ) else
+      check_pat_fields_typ_dec env t tfs' pfs te at
+  | T.{lab; typ; _}::tfs', { it = TypPF(_); _ }::_ ->
+    check_pat_fields_typ_dec env t tfs' pfs te at
+  | T.{lab; typ; _}::tfs', { it = VarPF(id, p); _ }::pfs' ->
```
```suggestion
  | T.{lab; typ; _}::tfs', { it = VarPF(id, p); _ }::pfs' ->
```
What if `typ` is a type field `Typ t'`. Is this code still doing  the right thing?

test/fail/M0216.mo
```diff
@@ -0,0 +1,14 @@
+func _unableToDetermineType() {
+  // Only able to infer type fields for val paths at the moment
+  let { type K } = module { type K = Nat };
```
crusso: Oh, that's a but unfortunate, but I guess not likely to come up for real.

test/fail/M0216.mo
```diff
@@ -0,0 +1,14 @@
...
+};
+
+func _requiresAnnotation() {
+  class HashMap() {
+    public type Hash = Nat64;
+  };
+  // let map : HashMap = HashMap(); // Compiles fine
+  let map = HashMap();
+  let { type Hash } = map;
```
crusso: Ditto also unfortunate.
I wonder if this actually works when map is defined in another scope, instead of as part of the same recursive block.

src/mo_def/syntax.ml
christoph-dfinity: Fixed in a7d5716ff554c229f2fdd66d3434e5435c08b4cf

src/mo_frontend/typing.ml
christoph-dfinity: It does, yeah. I added a comment.

test/fail/M0216.mo
Kamirus: It could be used like a local "import", to open the module only in limited scope, but this is a bit confusing to see.
Maybe it is simpler to only allow `type` patterns in `import`s on toplevel?

test/fail/objpat-type-fields.mo
```diff
@@ -0,0 +1,74 @@
...
+  let (#left{ type N }) = x else {
+    return ()
+  };
+
+  let _n : N = 10;
+};
+
+func _f1_1(x : V1) {
+  // Error with mismatched types
+  let (#left{ type N } or #right { type N }) = x else {
```
Kamirus, line 32: Interesting and confusing 😄 In the original issue I thought we just wanted unqualified imports only in `import`s on the toplevel.
This is complicating the compiler code, right? You need to traverse the pattern to add potential types to the scope. Complex logic that will probably be unused by most users, but would need to be maintained.

src/mo_frontend/typing.ml
christoph-dfinity: `var` doesn't allow you to bind a pattern, so I don't think that works.

test/fail/objpat-type-fields.mo
christoph-dfinity: It sounds like you're expecting there to be a separate codepath for checking import patterns, but those are checked using the same code as any other pattern. I've cut down on the complexity by restricting the RHS of these to be `val_path`, but I don't see any other clear boundaries I could draw.

src/mo_frontend/typing.ml
christoph-dfinity: I'm not quite sure. Maybe at some point this code only compared labels and needed to make sure to not swap around type and value fields with the same label.
`compare_pat_field` and `T.compare_field` both produce a total ordering, so I think any sort will do.

src/mo_frontend/coverage.ml
christoph-dfinity: Fixed in 40a52e8de7cedb57667fe7d786040c21588b634d

src/mo_frontend/typing.ml
christoph-dfinity: Refactored the "sorted walk" in ba0d71ffbdd2321536c64a93b82811e1cfa6b4b6. LMK what you think

test/fail/objpat-type-fields.mo
Kamirus: Ok since imports are desugared to lets I guess it's the easiest to allow type patterns in every object pattern

src/mo_frontend/typing.ml
```diff
@@ -2552,28 +2558,30 @@ and check_pat_fields env t tfs pfs ve at : Scope.val_env =
...
+       id.it
+       display_typ_expand t
+  | tf::tfs', pf::pfs' ->
+     match compare_pat_typ_field tf pf with
+     | -1 -> check_pat_fields env t tfs' pfs ve at
+     | 1 -> (match pf.it with
+            | ValPF(_) -> check_pat_fields env t [] pfs ve at
+            | _ -> check_pat_fields env t tfs pfs' ve at)
+     | _ -> match tf, pf with
+        | T.{lab; typ; src}, {it = ValPF(id, pat); _} ->
```
christoph-dfinity, line 2581: Note to self: Think about type fields on both sides?

src/mo_frontend/typing.ml
christoph-dfinity: Chatted about this and determined it's fine for now.

src/mo_frontend/typing.ml
crusso: Is this ok now?

src/mo_frontend/typing.ml
```diff
@@ -2434,7 +2457,7 @@ and check_pat_aux' env t pat val_kind : Scope.val_env =
   | ObjP pfs ->
     let pfs' = List.stable_sort compare_pat_field pfs in
     let s, tfs =
-      try T.as_obj_sub (List.map (fun (pf : pat_field) -> pf.it.id.it) pfs') t
+      try T.as_obj_sub (List.map (fun pf -> (pf_id pf).it) pfs') t
```
crusso: I think there may be a tiny bug here since you wind up conflating type and value fields.
```
let {type t ; t} = loop {} 
```
might be a good example to try.

crusso: T.as_obj_sub [x,y,...,z] None returns { x : None, y : None , ... z : None} and is used to promote a None to a record type of the correct shape. Not sure how to generalize this to type fields since they are invariant unless you've got an expected type around...
Might be worth checking the other uses of T.as_obj_sub...

crusso: May you should just enter the catch clause if there are any type fields patterns in pfs..

crusso (commented):
LPGTM. I guess the language-manual.md and doc might need some extending too, though. Nice work!

src/mo_frontend/typing.ml
christoph-dfinity: It was not. Thank you for the reminder. Fixed in [f9c1fd4](https://github.com/dfinity/motoko/pull/5056/commits/f9c1fd4d2ebf4be1147b5ad6f7adc0e8b1ac8087)

src/mo_frontend/typing.ml
christoph-dfinity: [89b7895](https://github.com/dfinity/motoko/pull/5056/commits/89b7895eddae7f0b35eaaf5cc77fb006a9aaa482) is good enough.

christoph-dfinity:
I added a bit of documentation and a CHANGELOG entry. PTAL

doc/md/fundamentals/7-modules-imports.md
```diff
@@ -26,10 +26,10 @@ Debug.print("Hello, world!");
 
 The `mo:` prefix identifies a Motoko module. The declaration does not include the `.mo` file extension.
 
-You can also selectively import and rename a subset of named values from a module by using the object pattern syntax:
+You can also selectively import and rename a subset of named values and types from a module by using the object pattern syntax:
 
 ``` motoko
-import { map; find; foldLeft = fold } = "mo:base/Array";
+import { type Map; get; contains; foldLeft = fold } = "mo:base/OrderedMap";
```
crusso: This example isn't correct actually (try cd doc; make preview). I tried fixing it using List instead but then uncovered a bug. Perhaps the interpreters are still buggy?

See https://github.com/dfinity/motoko/pull/5362 for a repro.

crusso:
```suggestion
import { type List; get; foldLeft = fold } = "mo:base/List";
```
might actually type check after bug fix.

src/mo_interpreter/interpret.ml
```diff
@@ -824,8 +824,12 @@ and declare_pat_fields pfs ve : val_env =
...
   | [] -> ve
   | pf::pfs' ->
-    let ve' = declare_pat pf.it.pat in
-    declare_pat_fields pfs' (V.Env.adjoin ve ve')
+     match pf_pattern pf with
+     | Some p ->
+        let ve' = declare_pat p in
+        declare_pat_fields pfs' (V.Env.adjoin ve ve')
+     | None ->
+        ve
```
crusso: Maybe the bug is here: you need to recurse on pfs in the None case?

crusso: And you need to adjust the interpret_ir.ml file too.

christoph-dfinity: I don't think the `interpret_ir.ml` file needs adjusting, as I remove the type fields from object patterns during lowering.

PR #5067, opened 2025-04-21
chore: turn Motoko into a `nix` flake

basvandijk:
* Switch to Nix Flakes which is the modern way to setup Nix projects and gives access to the more user-friendly `nix <subcommand>` CLI. Make sure that the following is present in your Nix configuration (`/etc/nix/nix.conf`):
```
extra-experimental-features = nix-command flakes
```

* Introduce a `.envrc` containing `use flake` so that you can use [`direnv`](https://direnv.net) with [`nix-direnv`](https://github.com/nix-community/nix-direnv) instead of using `nix-shell` / `nix develop`, i.e. a `cd` to the motoko directory is enough to load all dependencies.

* Moved most derivations to their own dedicated files under `nix/` instead of inlining them all like was done previously in `default.nix` which was over 900 lines. Now `flake.nix` is only 235 lines which is easier to comprehend.

* `nix/gitSource.nix` is removed since Nix Flakes automatically only use files added to Git.

* Switch from the impure [mozilla-nixpkgs](https://github.com/mozilla/nixpkgs-mozilla) overlay to the pure [oxalica/rust-overlay](https://github.com/oxalica/rust-overlay).

* Upgrade the latest stable Rust toolchain from 1.85.0 to 1.86.0.

* Upgrade nixpkgs to release-24.11.


Building.md
```diff
@@ -166,16 +188,3 @@ build system.)
...
-When the `.cabal` file of a Haskell package is changed you need to make sure the
-corresponding `.nix` file (stored in `nix/generated/`) is kept in sync with it. These files are automatically generated; run
-```
-nix-shell nix/generate.nix
-```
-to update.
-
-Don't worry if you forget to update the `default.nix` file, the CI job
-`check-generated` checks if these files are in sync and fail with a diff if
-they aren't.
```
basvandijk, line 181: Note to reviewers: we now use the nixpkgs function `callCabal2nix` to generate the derivations for our Haskell packages which means we no longer need to pre-generate the .nix files. This does use IFD (Import From Derivation) but since we no longer use Hydra to build things we don't care about that.

.github/workflows/niv-updater-rare.yml
basvandijk, line 1: I forgot I removed these `niv-updater*` workflows. I will replace them with ones that work with the Nix Flake.

.gitignore
```diff
@@ -32,3 +32,5 @@ Session.vim
 
 # Mac OS
 *DS_STORE*
+
+.direnv
```
ggreif, line 36:
```suggestion
.direnv

```

.github/workflows/niv-updater-rare.yml
basvandijk, line 1: Note to reviewers: these `.github/workflows/niv-updater*.yml` workflows have been replaced by the `.github/workflows/update-flake-lock*.yml` workflows below.

christoph-dfinity (approved):
I'm not the resident nix export so definitely wait for a review from Gabor, but these are the things I did to maybe help a bit:

- I reviewed the formatting changes caused by the `ocamlformat` update to the language server and docs code and they all look good 
- I read through the new `.nix` files and having them separated out into individual tasks underneath `nix/` seems like a nice improvement
- `direnv` integration worked nicely, I was able to build the compiler without having to explicitly enter `nix develop`
- I checked that making a small change to a type error message and running `nix build .#tests.fail` produced the appropriate test failure
- The cachix integration looks to work, as I only started building OCaml files once I'd made a change that deviated from `master`

.github/workflows/niv-updater.yml
```diff
@@ -1,32 +0,0 @@
-name: Update niv dependencies (daily)
```
ggreif, line 1: Wait a sec. There is a job that updates the `base` library in here. We shouldn't kill that! I thought you only wanted to remove the rare one and stop `ic-hs` from being built. (I think `ic-ref` is zapped since last week).

ggreif: Maybe go through closed PRs or updater branches and see which are still relevant.

ggreif: I see. Just read your note!

CI.md
```diff
@@ -23,16 +23,11 @@ Knowing if it is broken
...
 All pushes to any branch are built by a Github Action job, on Linux and Darwin.
 
-**Implementation (internal):**
-All pushes to `master`, as well as to all branches with open PRs cause hydra to
-build the jobs described in `ci.nix` resp. `ci-pr.nix`.
-
 The build status is visible via the Github status (coarsly: only evaluation,
-`all-systems-go` and `all-jobs`) and via hydra directly (full per-job detail
-including logs).
+`all-systems-go` and `all-jobs`).
```
ggreif, line 30: this is now `debug-`/`release-systems-go`

CI.md
```diff
@@ -23,16 +23,11 @@ Knowing if it is broken
...
 
-**Implementation (external):**
+**Implementation:**
 All pushes to any branch are built by a Github Action job, on Linux and Darwin.
 
-**Implementation (internal):**
-All pushes to `master`, as well as to all branches with open PRs cause hydra to
-build the jobs described in `ci.nix` resp. `ci-pr.nix`.
-
 The build status is visible via the Github status (coarsly: only evaluation,
```
ggreif, line 29:
```suggestion
The build status is visible via the Github status (coarsely: only evaluation,
```

nix/drun.nix
```diff
@@ -1,83 +1,81 @@
...
-      # To automate this, .github/workflows/update-hash.yml has been
-      # installed. You will normally not be bothered to perform
-      # the command therein manually.
+  # update this after bumping the dfinity/ic pin.
+  # 1. change the hash to something arbitrary (e.g. flip one digit to 0 or use `pkgs.lib.fakeSha256`)
+  # 2. run nix-build -A drun nix/
+  # 3. copy the “expected” hash from the output into this file
+  # 4. commit and push
+  #
+  # To automate this, .github/workflows/update-hash.yml has been
```
ggreif, line 13: IIRC, above you say `update-hash.yml` is obsolete. I think this comment can go.

ggreif: Or maybe not?

basvandijk: No update-hash.yml is still used to update the hashes in drun.

nix/tests.nix
```diff
@@ -0,0 +1,248 @@
+{ pkgs, llvmEnv, esm, viper-server, commonBuildInputs, debugMoPackages }:
+with debugMoPackages;
+let
+  # The following were previously arguments to default.nix but flakes don't accept options yet.
```
ggreif, line 4: Okay, I think that was manly needed for testing. I'll watch this space.

nix/tests.nix
```diff
@@ -0,0 +1,248 @@
...
+    [ filecheck pkgs.wasmtime ];
+
+  filecheck = pkgs.runCommandNoCC "FileCheck" { } ''
+    mkdir -p $out/bin
+    cp ${pkgs.llvm}/bin/FileCheck $out/bin
+  '';
+
+  # extra deps for test/ld
+  ldTestDeps =
+    with pkgs; [ llvmPackages_18.lld llvmPackages_18.clang ];
```
ggreif, line 29:
```suggestion
    with pkgs.llvmPackages_18; [ lld clang ];
```

ggreif (approved):
Heroic reorg! Thanks for doing this. Having a flake-savvy dev around always pays off!

(some minor nits only)

nix/shell.nix
```diff
@@ -0,0 +1,82 @@
...
+      #builtins.concatMap (d: d.buildInputs or [ ]) (builtins.attrValues tests) ++
+      [
+        nix-update
+        pkgs.ncurses
+        pkgs.ocamlPackages.merlin
+        pkgs.ocamlPackages.utop
+        pkgs.ocamlformat
+        pkgs.ocamlPackages.ocaml-lsp
+        pkgs.fswatch
+        pkgs.niv
```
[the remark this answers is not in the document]
basvandijk: Yes, good catch. Will remove ...

alexandru-uta (approved):
Very cool, Bas! Thanks for the effort, I also had a look and it seems good to me.

PR #5070, opened 2025-04-22
fix: doc preview

rvanasa:
Fixes runnable code snippets in `make -C doc preview` (previously had an error related to importing `Wasm.Lib`).


src/Makefile
```diff
@@ -13,7 +13,7 @@ MOC_JS_TARGET = _build/default/js/moc_js.bc.js
 MOC_INTERPRETER_TARGET = _build/default/js/moc_interpreter.bc.js
 DIDC_JS_TARGET = _build/default/js/didc_js.bc.js
 
-DUNE_OPTS ?=
+DUNE_OPTS ?= --profile=release
```
rvanasa, line 16: @crusso, I think we need this here because the error occurs whenever using `moc.js` / `moc_interpreter.js` from the Makefile (beyond the doc preview).

crusso (approved):

crusso (approved):

PR #5075, opened 2025-04-23
chore: deletes `mo-ide`

### LINKED ISSUE #5072: Remove `mo-ide`
The OCaml based language server in here has languished and is replaced by a JS based one in `vscode-motoko`. With Serokell's grant work focused on the JS implementation, let's remove what is essentially dead code.

We also need to make sure to update some downstream locations:

- `dfx _language-service` needs to be removed
- Remove code-paths referencing `dfx _language-service` in `vscode-motoko`
- Check if any of the tools downloading the compiler binaries expect the `mo-ide` binary to be around (dfx, vessel, mops)

christoph-dfinity:
Closes #5072


rvanasa (approved):
LGTM so far!

PR #5076, opened 2025-04-23
LSP: syntax error examples, serialize ID locations, keep all origins in typechecker (Serokell, Milestone-3)

GoPavel:
This PR contains following changes:
* Enhance syntax error messages with examples:
```
syntax error [M0001], unexpected token 'let', expected one of token or <phrase> sequence:
   } (e.g. '}')
   ; seplist(<dec_field>,<semicolon>) (e.g. '; public let x : Int = 0') |}] 
```
* Serialize the location of ID to precisely locate the user cursor in the AST and fix corner cases with the "go-to-definition" functionality in the LSP server
* Also to support "find-references" and "go-to-definition"  functionality for fields in the LSP server:
  * Enhance serialization of `typ_field`
   * Fixup location of `VarP` in `LetD`
   * Track possible origins of object fields (enabled via new API function `setTypecheckerCombineSrcs`)

Related PRs:
* in `node-motoko`: https://github.com/dfinity/node-motoko/pull/128
* in `vscode-motoko`: https://github.com/dfinity/vscode-motoko/pull/347


src/mo_frontend/test_recovery.ml
```diff
@@ -256,9 +397,102 @@ actor Main {
...
+      and <exp_bin(ob)> (e.g. 'and 42')
+      <unop> <exp_bin(ob)> (e.g. '+ 42')
+      <inst> <exp_nullary(ob)> (e.g. '<Int> 42')
+      [ <exp(ob)> ] (e.g. '[ 42 ]')
+
+    (unknown location): syntax error [M0001], unexpected token '(', expected one of token or <phrase> sequence:
+      func <pat_plain> <annot_opt> <func_body> (e.g. 'func x : Int {}')
+      class <pat_plain> <annot_opt> <class_body> (e.g. 'class x : Int = {}')
+      object class <pat_plain> <annot_opt> <class_body> (e.g. 'object class x : Int = {}')
+      module class <pat_plain> <annot_opt> <class_body> (e.g. 'module class x : Int = {}')
```
crusso, line 463: These examples are syntactically ok, but would never type checked because the pattern are untyped.
I wonder if we can improve this by introducing
<arg_pat> := <pat_plain>
in the grammar and using it at appropriat places,
and then using a different example for <arg_pat> (such as ( x : Int)).
Maybe too hacky.

crusso (commented):
Interim commnents. Need to reboot.

crusso:
I did a little experiment here #5108 to see if refactoring the grammar improves the error message. Unfortunately, I don't recall how to update the dune test output... so just added a log in the PR description

src/mo_frontend/test_recovery.ml
```diff
@@ -84,9 +96,123 @@ let%expect_test "test1" =
...
+      .<nat> (e.g. '.<nat>')
+      ! (e.g. '!')
+      <exp_nullary(ob)> (e.g. '42')
+      <unop> <exp_bin(ob)> (e.g. '^ 42')
+      <unassign> <exp(ob)> (e.g. '^= 42')
+      <binop> <exp_bin(ob)> (e.g. '-% 42')
+      <binassign> <exp(ob)> (e.g. '-%= 42')
+      <binop> <exp_bin(ob)> (e.g. '**% 42')
+      <binassign> <exp(ob)> (e.g. '**%= 42')
+      <binop> <exp_bin(ob)> (e.g. '*% 42')
```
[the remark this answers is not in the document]
crusso: If we could condense all the cases that have the same item, but different examples, the errors might actually be less overwhelming and more useful. Would that be easy, either in a principled or hacky way? Maybe just create a set of errors base on the text of the item and then only explain this smaller set with examples?

src/mo_frontend/test_recovery.ml
crusso: That seems to work nicely in #5108. Not saying you should merge that it was just an experiment.

src/mo_frontend/test_recovery.ml
crusso, line 110: For example:
```
      <binop> <exp_bin(ob)> (e.g. '-% 42')
      <binassign> <exp(ob)> (e.g. '-%= 42')
      <binop> <exp_bin(ob)> (e.g. '**% 42')
      <binassign> <exp(ob)> (e.g. '**%= 42')
      <binop> <exp_bin(ob)> (e.g. '*% 42')
```
could be condensed to
```
      <binop> <exp_bin(ob)> (e.g. '+ 42')
      <binassign> <exp(ob)> (e.g. '+= 42')
```
Since the lhs are repeated but the rhs are just minor variations.

src/mo_frontend/printers.ml
```diff
@@ -68,184 +91,188 @@ let string_of_symbol = function
...
-  | X (T T_ASYNC) -> "async"
-  | X (T T_ASYNCSTAR) -> "async*"
+  | X (T T_CASE) -> simple_token "case"
+  | X (T T_BREAK) -> simple_token "break"
+  | X (T T_BOOL) -> simple_token "<bool>"
+  | X (T T_AWAIT) -> simple_token "await"
+  | X (T T_AWAITSTAR) -> simple_token "await*"
+  | X (T T_ASYNC) -> simple_token "async"
+  | X (T T_ASYNCSTAR) -> simple_token "async*"
   | X (T T_ASSIGN) -> binassign "assign"
```
crusso, line 157:
```suggestion
  | X (T T_ASSIGN) -> binassign ":="
```

src/mo_frontend/printers.ml
```diff
@@ -3,58 +3,81 @@ open Parser.MenhirInterpreter
...
-  | X (T T_error) -> "error"
+(* private helper *)
+let repr_of_symbol : xsymbol -> (string * string) =
+  let simple_token con = (con, con) in
+  let binop = abstract "<binop>" in
+  let relop = abstract "<relop>" in
+  let binassign = abstract "<binassign>" in
+  (* all unary operators are also binary operators, so keep them unary *)
+  let unop = abstract "<unop>" in
+  let unassign = abstract "<unassign>" in
```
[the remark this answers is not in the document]
crusso: Ah, I just tried that here (forgot I hadn't done this) and it seems to work quite well:
https://github.com/dfinity/motoko/pull/5108/commits/a109fb3171a4ff8a9984f59d86e94985b96bb329

src/mo_frontend/printers.ml
```diff
@@ -68,184 +91,188 @@ let string_of_symbol = function
...
+  | X (N N_annot_opt) -> "<annot_opt>", eg_annot
+  | X (N N_seplist_case_semicolon_) -> seplist ("<case>", eg_case) semi
+  | X (N N_seplist_dec_SEMICOLON_) -> seplist ("<dec>", eg_dec) semi2
+  | X (N N_seplist_dec_semicolon_) -> seplist ("<dec>", eg_dec) semi
+  | X (N N_seplist_typ_dec_semicolon_) -> seplist ("<typ_dec>", eg_typ_dec) semi
+  | X (N N_seplist_dec_field_semicolon_) -> seplist ("<dec_field>", eg_dec_field) semi
+  | X (N N_seplist_exp_ob__COMMA_) -> seplist ("<exp(ob)>", eg_exp) comma
+  | X (N N_seplist_exp_field_semicolon_) -> seplist ("<exp_field>", eg_exp_field) semi
+  | X (N N_seplist1_exp_field_semicolon_) -> "seplist1(<exp_field>,<semicolon>)", eg_exp_field
+  | X (N N_separated_nonempty_list_AND_exp_post_ob__) -> "seplist+(<exp_post(ob)>,and)", eg_exp ^ "and"
```
crusso, line 230:
```suggestion
  | X (N N_separated_nonempty_list_AND_exp_post_ob__) -> "seplist+(<exp_post(ob)>,and)", eg_exp ^ " and"
```

crusso: (I got weird output with "{;" in the interpreter with a trailing "and"

src/mo_def/arrange.ml
```diff
@@ -54,10 +54,17 @@ module Make (Cfg : Config) = struct
   | Mo_types.Type.Triv -> Atom "Triv"
   | Mo_types.Type.Await -> Atom "Await"
 
-  let annot_typ t it = if Cfg.include_types then ":" $$ [it; typ t] else it
+  let annot_typ t it =
+    if Cfg.include_types
+    then
+      if Cfg.include_type_rep
+      then ":" $$ [it; typ t; Arrange_type.typ t]
```
crusso, line 61: Can't these type annotations get huge? There's no sharing at all so every subexpression will receive its own type annotation that will duplicate lots of information that could be shared with other type annotations.

src/mo_types/arrange_type.ml
```diff
@@ -43,6 +43,19 @@ let prim = function
 
 let con c = Atom (Type.string_of_con c)
 
+let pos p =
+  "Pos" $$
+    [ Atom p.Source.file
```
crusso, line 48: This will lead to lots of duplicated file names in the s-expression (twice for every region, and then many regions might share the same filename)

src/mo_types/type.ml
```diff
@@ -41,6 +41,21 @@ type prim =
...
+
+  let compare l r =
+    let open Source in
+    let join_compare l r = if l = 0 then r else l in
+    let compare_pos l r =
+      join_compare
+        (compare l.file r.file)
+        (join_compare (compare l.line r.line) (compare l.column r.column))
+    in
+    join_compare (compare_pos l.left r.left) (compare_pos l.right r.right)
```
crusso, line 55: Use nested if or match to avoid doing redundant comparisons.

src/mo_types/type.ml
```diff
@@ -64,15 +79,15 @@ and scope = typ
 and bind_sort = Scope | Type
 
 and bind = {var : var; sort: bind_sort; bound : typ}
-and src = {depr : string option; region : Source.region}
+and src = {depr : string option; region : Source.region; mutable srcs : Region_set.t}
```
crusso, line 82: Yikes! do we really need to make the type mutable for this?

src/mo_types/type.ml
```diff
@@ -1059,8 +1074,17 @@ and rel_fields d rel eq tfs1 tfs2 =
...
-      rel_fields d rel eq tfs1' tfs2'
+      let is_rel =
+        rel_typ d rel eq tf1.typ tf2.typ &&
+        rel_fields d rel eq tfs1' tfs2'
+      in
+      if !Mo_config.Flags.typechecker_combine_srcs && is_rel then begin
+        let srcs = Region_set.union tf1.src.srcs tf2.src.srcs in
+        if rel == eq then
+          tf1.src.srcs <- srcs;
+        tf2.src.srcs <- srcs;
```
crusso, line 1085: This mutation scares me. What if the subtyping relation fails to hold after all, won't the destructive merges remain?

src/mo_types/type.ml
```diff
@@ -1375,7 +1416,12 @@ and combine_fields rel lubs glbs fs1 fs2 =
     | _ ->
       match combine rel lubs glbs f1.typ f2.typ with
       | typ ->
-       {lab = f1.lab; typ; src = empty_src} :: combine_fields rel lubs glbs fs1' fs2'
+        let src =
```
crusso, line 1419: This seems safer to me since it's actually functional, we are constructing a new type without updating f1.src or f2.src.

Can we do the analysis without the destructive subtyping?

src/mo_frontend/typing.ml
```diff
@@ -1196,6 +1209,61 @@ let error_bin_op env at t1 t2 =
...
+    | +1 -> combine_pat_fields_srcs env t tfs' pfs
+    | _ ->
+      combine_pat_srcs env typ pf.it.pat;
+      combine_pat_fields_srcs env t tfs' pfs';
+
+and combine_id_srcs env t id : unit =
+  match T.Env.find_opt id.it env.vals with
+  | None -> ()
+  | Some (t', _, _, _) ->
+    (* Use [sub] to merge the fields of sources, if one type is indeed a subtype
```
crusso, line 1232: Even if they aren't subtype, the merging will  persist because of the destructive update, will it not?

src/mo_frontend/typing.ml
```diff
@@ -23,7 +23,7 @@ type ret_env = T.typ option
 type val_env  = (T.typ * Source.region * Scope.val_kind * avl) T.Env.t
 
 (* separate maps for values and types; entries only for _public_ elements *)
-type visibility_src = {depr : string option; id_region : Source.region; field_region : Source.region}
+type visibility_src = {depr : string option; id_region : Source.region; field_region : Source.region; srcs : T.Region_set.t}
```
crusso, line 26: what is srcs? The locations of bindings referencing this identifier? Does that include transitive references somehow?

src/mo_types/type.ml
crusso: Maybe a safer approach would be to accumulate the updates in delayed form and then only commit them when the outer call to subtyping succeeds and discard otherwise. Something like a mutable list of thunks that get forced only on success.

crusso (changes requested):
I think I need a meeting and more comments to understand the occurrence tracking.

I'm also a bit worried about the size of the typed terms communicate back to LSP. I expect these could get huge due to lack of sharing.

I wonder if it might be feasilble to pass an object back to JS, storing the Ocaml AST and providing appropriate query methods to resolve references etc. That would avoid serializing the AST as potentially very large string. Especially when type annotations are included.

src/mo_frontend/test_recovery.ml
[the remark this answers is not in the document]
GoPavel: Yeah I think collapsing this set based on this "abstract" representation makes total sense. So basically, we already have this implicit grouping: "**%=", "*%" become "<binop>".
Then we can just factorize the list by the "abstract" representation and pick any example. It's not so hacky. Parser says all possibilities to complete the grammar rule, and we group them by higher-level terms "binop" or "binassign".

The set of errors is another approach that is closer to `.message` files. I am not sure about its maintainability :shrug:

I can try collapsing. It's gonna be real quick, and the set of errors based on text could be the next step.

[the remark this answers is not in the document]
GoPavel: Oh, I see you already did it, I will check you changes and will add it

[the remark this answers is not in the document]
GoPavel: Oh, I see you already did it, I will check you changes and will add it

crusso:
BTW, we recently changed out nix setup to use nix-flakes. So if you merge with master, re-read Building.md. There's a bit of setup (not much) and you now enter a nix shell with `nix develop` not `nix-shell` as before.

src/mo_def/arrange.ml
heitor-lassarote: The current implementation introduces `typ t` to every node, which is the type AST in string form, so we thought it would be fine to also introduce the AST in its structural form, provided that both `include_types` and `include_type_rep` are enabled. But after investigating a bit, it's indeed a problem that can be observed in benchmarks.

Perhaps one simple way to mitigate this might be to arrange these types only where the language server currently requires it, I'll better investigate this point and get back to you.

src/mo_types/arrange_type.ml
heitor-lassarote: I've changed it so the `region` has the filename from the left position, and removed it from `pos`. Note that in `Arrange`, the filename is introduced in both `pos`itions, but if it matches the `main_file` from the configuration, then it outputs the empty string instead. I don't do the `main_file` check here, but if you'd like, I can change it. Since those are only output when there are source fields, I think it's fine.

(Note: I'm addressing this in https://github.com/serokell/motoko/pull/57)

src/mo_types/type.ml
```diff
@@ -41,6 +41,21 @@ type prim =
   | Principal
   | Region
 
+module Region_ord = struct
+  type t = Source.region
+
+  let compare l r =
+    let open Source in
+    let join_compare l r = if l = 0 then r else l in
```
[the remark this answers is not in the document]
heitor-lassarote: I've changed it so it uses `match ... with` instead. Although I wonder if annotating `join_compare` with `[@inline]` would have helped. :thinking: ([Link to changes](https://github.com/serokell/motoko/pull/57))

src/mo_types/type.ml
heitor-lassarote: Done. ([Link to changes](https://github.com/serokell/motoko/pull/57))

src/mo_types/type.ml
heitor-lassarote: This seemed like the least intrusive way to propagate occurrences between fields when we are running `eq` or `sub`. Having a very complicated solution with a bigger maintenance burden for the language server seemed suboptimal.

src/mo_types/type.ml
heitor-lassarote: Good catch, the subtyping relation may fail at a later point, which I didn't think about. I've applied your idea. ([Link to changes](https://github.com/serokell/motoko/pull/57))

src/mo_frontend/typing.ml
heitor-lassarote: > what is srcs? The locations of bindings referencing this identifier?

We are specifically trying to track possible origins of an identifier rather than references with those changes. For example, if we have two classes `C1` and `C2`, and `C1` is a subtype of `C2`, we might have something like `let c = if b { C1() } else { C2() }; let x = c.x` and we want to understand from where `x` originated. Since this is decided at runtime, we can overapproximate this by collecting the `id_region` of both classes and merge them during some operations such as `lub`, `glb`, `eq`, `sub`, etc.

Most of the heavy-lifting for finding references is done in `vscode-motoko` and not in the typechecker.

> Does that include transitive references somehow?

I'm not sure what you mean by transitive references, but if it's, for example, like `let a = f.x; let b = a;` where `b` is transitively referencing `f.x` through `a`, then no, we are trying to collect direct references in this sense. But they may be "transitive" in the sense that a field can be referenced through a subtyping relation, i.e., if `sub f g` holds, then `f.x` and `g.x` will be referenced with `x`.
The current implementation, due to the mutation in the supertype's field, will be "tainted" with the references from the subtype, but for finding references, this is a reasonable approximation.

src/mo_frontend/typing.ml
```diff
@@ -1342,7 +1410,7 @@ and infer_exp'' env exp : T.typ =
     end
   | TagE (id, exp1) ->
     let t1 = infer_exp env exp1 in
-    T.Variant [T.{lab = id.it; typ = t1; src = empty_src}]
+    T.Variant [T.{lab = id.it; typ = t1; src = {empty_src with srcs = singleton_srcs id.at}}]
```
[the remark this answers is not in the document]
heitor-lassarote: I left a reply to your previous comment explaining how we're changing the typechecker to track possible sources of a field, not its references, which is done by `vscode-motoko`. Hence, there are no changes for `DotE` here.

heitor-lassarote:
> I wonder if it might be feasilble to pass an object back to JS, storing the Ocaml AST and providing appropriate query methods to resolve references etc. That would avoid serializing the AST as potentially very large string. Especially when type annotations are included.

My understanding is that, although when converting the AST we transform most data as strings since S-expressions expect strings in `Atom`s, the entire AST won't be serialized into one big string. After this conversion, it's traversed one more time (in `Js.Common.js_of_sexpr`) to convert from the Js_of_ocaml representation to a more orthodox JS representation, but it still retains its structure.

However, some investigation is needed on better handling the type annotations. I'll check it.

But I think changing how the AST is given to `node-motoko` (which `vscode-motoko` depends on) would be a breaking change, and also outside the scope for this grant.

> Just as a straw man suggestion. What is wrong with the naive approach of:
> 
> Add a global imperative map from occurences to definitions, one per filescope scope.
> Store (immutable) defining region of identifier in environment and types.
> On identifier lookup, add map entry from occurrence of identifier to definining region, found in environment or object > type.
> Invert the map(s) to find occurrences for particular definitions.
> I guess I'm assuming that queries are just done based on regions and expect regions as results.

I believe it could work, but I don't see how much it would improve the current solution, except perhaps when there is the mutation of `srcs`, where perhaps it could help a little with the approximation. It seems that we'd still have to add fields to the map in the same places where we currently add to `srcs`.

One point of discussion is how to expose these fields after that, we'd need to change the pipeline to either (1) serialize this table with the return result, or (2) bring it to `Arrange_type` to keep the fields, or (3) add new methods to `moc.js` to manage this table (which I believe is your idea with queries).

src/mo_frontend/parsing.ml
```diff
@@ -26,9 +26,20 @@ let abstract_symbols explanations =
...
+  let ss = List.map Printers.string_of_symbol future in
   String.concat " " ss
 
+let abstract_future_with_example future =
+  let ss = List.map Printers.string_of_symbol future in
+  let f x xs =
+    let example = Printers.example_of_symbol x in
+    if String.compare example "" = 0
+    then xs
+    else example :: xs
```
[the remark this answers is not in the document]
GoPavel: done

src/mo_frontend/test_recovery.ml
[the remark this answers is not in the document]
GoPavel: done

src/mo_frontend/printers.ml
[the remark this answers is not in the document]
GoPavel: done

src/mo_frontend/test_recovery.ml
GoPavel: fixed

src/mo_frontend/printers.ml
GoPavel: fixed

src/mo_frontend/printers.ml
GoPavel: yeah, it was wrong, fixed

GoPavel:
I've fixed the syntax error examples there: https://github.com/serokell/motoko/pull/58
I merge it when build pass

src/mo_types/type.ml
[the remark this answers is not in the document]
crusso: I hope @inline would not help here otherwise @inline would be changing semantics, making something strict lazy ;->

crusso:
> > I wonder if it might be feasilble to pass an object back to JS, storing the Ocaml AST and providing appropriate query methods to resolve references etc. That would avoid serializing the AST as potentially very large string. Especially when type annotations are included.
> 
> My understanding is that, although when converting the AST we transform most data as strings since S-expressions expect strings in `Atom`s, the entire AST won't be serialized into one big string. After this conversion, it's traversed one more time (in `Js.Common.js_of_sexpr`) to convert from the Js_of_ocaml representation to a more orthodox JS representation, but it still retains its structure.

Ok. Thanks for the explanation

> However, some investigation is needed on better handling the type annotations. I'll check it.
> 
> But I think changing how the AST is given to `node-motoko` (which `vscode-motoko` depends on) would be a breaking change, and also outside the scope for this grant.

Fair enough.

> > Just as a straw man suggestion. What is wrong with the naive approach of:
> > Add a global imperative map from occurences to definitions, one per filescope scope.
> > Store (immutable) defining region of identifier in environment and types.
> > On identifier lookup, add map entry from occurrence of identifier to definining region, found in environment or object > type.
> > Invert the map(s) to find occurrences for particular definitions.
> > I guess I'm assuming that queries are just done based on regions and expect regions as results.
> 

> believe it could work, but I don't see how much it would improve the current solution, except perhaps when there is the mutation of `srcs`, where perhaps it could help a little with the approximation. It seems that we'd still have to add fields to the map in the same places where we currently add to `srcs`.


So the way I think of goto-definition and goto-references is perhaps a bit naive, but I'm think  that for identifiers it basically maps each occurrence to its binding site, and a binding site to all its occurrences.
If we store the definition location with each identifier and field, then variable and field lookup will naturally resolve a variable or field reference to its definition and all we need to do is store that resolution somewhere, either in a separate map, or just in the AST itself, i.e. instead of having a global map, we could to just store the region of the definition with each occurrence as an extended note on identifiers.

The conversion to JS could then extend the maps from definitions to occurrences and from occurrences to definitions.
You probably don't need to transmit the type of every subexpression with the AST now, as long the identifier and projection nodes store the defining region.

crusso:
I guess another mild objection I have to the mutable src fields in the types is that we will wind up mutating things that are in the scope cache, which seems a bit dangerous. Unless I've missed something.

heitor-lassarote:
> > > Just as a straw man suggestion. What is wrong with the naive approach of:
> > > Add a global imperative map from occurences to definitions, one per filescope scope.
> > > Store (immutable) defining region of identifier in environment and types.
> > > On identifier lookup, add map entry from occurrence of identifier to definining region, found in environment or object > type.
> > > Invert the map(s) to find occurrences for particular definitions.
> > > I guess I'm assuming that queries are just done based on regions and expect regions as results.
> 
> > believe it could work, but I don't see how much it would improve the current solution, except perhaps when there is the mutation of `srcs`, where perhaps it could help a little with the approximation. It seems that we'd still have to add fields to the map in the same places where we currently add to `srcs`.
> 
> So the way I think of goto-definition and goto-references is perhaps a bit naive, but I'm think that for identifiers it basically maps each occurrence to its binding site, and a binding site to all its occurrences. If we store the definition location with each identifier and field, then variable and field lookup will naturally resolve a variable or field reference to its definition and all we need to do is store that resolution somewhere, either in a separate map, or just in the AST itself, i.e. instead of having a global map, we could to just store the region of the definition with each occurrence as an extended note on identifiers.
> 
> The conversion to JS could then extend the maps from definitions to occurrences and from occurrences to definitions. You probably don't need to transmit the type of every subexpression with the AST now, as longas the identifier and projection nodes store the defining region.

@crusso, I've changed the code based on your suggestions and pushed them to this PR, but I did so with a couple of differences:

* I kept the source tracking so as not to include an occurrence, but just possible definitions, as with the previous solution. This is so we can keep reusing the code in vscode-motoko that already existed for definitions and to avoid including even more complexity in the typechecker.
* I changed the arrangement to, for now, only serialize the type AST for nodes that the language server cares about. The reason is that it's generally useful to have types there for more capabilities if implemented (I can see it being used for type definition and semantic tokens, for example). This will make things smoother for future contributions, although one thing to consider will be how to share these types. However, although I originally mentioned a performance impact, after some consideration, it might still be reasonable, given that the language server only requests the typed AST with some requests and not on every keystroke. Nonetheless, it's now only done for some nodes to decrease this impact.
* Although suggested to have one map per filename scope, I decided to keep everything in a single map. This is because the map's keys are `region`s, which include the file name. However, I introduced this map to `Scope.scope`, and so the cache will create a filename scope for libs, regardless.

> I guess another mild objection I have to the mutable src fields in the types is that we will wind up mutating things that are in the scope cache, which seems a bit dangerous. Unless I've missed something.

You are right, this could be a problem. I use a mutable hash table inside the typechecker, but I later turn it into an immutable map. Likewise, I give the initial table from the scope to the type checker as an immutable map and turn it into a mutable map. This is to avoid mutations in the cache and also to avoid having to perform very intrusive changes in `Mo_types.Type` (or we'd need to change many functions to take in a table, although perhaps we could turn it into a functor taking the table as a configuration) and `Mo_frontend.Typing` (otherwise most places where we call `sub`/`eq`/`lub`/`glb` would possibly need to be changed).

Hopefully, everything will be fine now; if not, let me know, and I'll address it.

crusso:
> > > > Just as a straw man suggestion. What is wrong with the naive approach of:
> > > > Add a global imperative map from occurences to definitions, one per filescope scope.
> > > > Store (immutable) defining region of identifier in environment and types.
> > > > On identifier lookup, add map entry from occurrence of identifier to definining region, found in environment or object > type.
> > > > Invert the map(s) to find occurrences for particular definitions.
> > > > I guess I'm assuming that queries are just done based on regions and expect regions as results.
> > 
> > 
> > > believe it could work, but I don't see how much it would improve the current solution, except perhaps when there is the mutation of `srcs`, where perhaps it could help a little with the approximation. It seems that we'd still have to add fields to the map in the same places where we currently add to `srcs`.
> > 
> > 
> > So the way I think of goto-definition and goto-references is perhaps a bit naive, but I'm think that for identifiers it basically maps each occurrence to its binding site, and a binding site to all its occurrences. If we store the definition location with each identifier and field, then variable and field lookup will naturally resolve a variable or field reference to its definition and all we need to do is store that resolution somewhere, either in a separate map, or just in the AST itself, i.e. instead of having a global map, we could to just store the region of the definition with each occurrence as an extended note on identifiers.
> > The conversion to JS could then extend the maps from definitions to occurrences and from occurrences to definitions. You probably don't need to transmit the type of every subexpression with the AST now, as longas the identifier and projection nodes store the defining region.
> 
> @crusso, I've changed the code based on your suggestions and pushed them to this PR, but I did so with a couple of differences:
> 
> * I kept the source tracking so as not to include an occurrence, but just possible definitions, as with the previous solution. This is so we can keep reusing the code in vscode-motoko that already existed for definitions and to avoid including even more complexity in the typechecker.
> * I changed the arrangement to, for now, only serialize the type AST for nodes that the language server cares about. The reason is that it's generally useful to have types there for more capabilities if implemented (I can see it being used for type definition and semantic tokens, for example). This will make things smoother for future contributions, although one thing to consider will be how to share these types. However, although I originally mentioned a performance impact, after some consideration, it might still be reasonable, given that the language server only requests the typed AST with some requests and not on every keystroke. Nonetheless, it's now only done for some nodes to decrease this impact.
> * Although suggested to have one map per filename scope, I decided to keep everything in a single map. This is because the map's keys are `region`s, which include the file name. However, I introduced this map to `Scope.scope`, and so the cache will create a filename scope for libs, regardless.
> 
> > I guess another mild objection I have to the mutable src fields in the types is that we will wind up mutating things that are in the scope cache, which seems a bit dangerous. Unless I've missed something.
> 
> You are right, this could be a problem. I use a mutable hash table inside the typechecker, but I later turn it into an immutable map. Likewise, I give the initial table from the scope to the type checker as an immutable map and turn it into a mutable map. This is to avoid mutations in the cache and also to avoid having to perform very intrusive changes in `Mo_types.Type` (or we'd need to change many functions to take in a table, although perhaps we could turn it into a functor taking the table as a configuration) and `Mo_frontend.Typing` (otherwise most places where we call `sub`/`eq`/`lub`/`glb` would possibly need to be changed).
> 
> Hopefully, everything will be fine now; if not, let me know, and I'll address it.

Thanks for the changes! I'm busy with another review but will take a look soon!

src/mo_def/arrange.ml
```diff
@@ -54,20 +58,44 @@ module Make (Cfg : Config) = struct
...
+     currently. As a compromise, we currently annotate only the nodes that
+     currently matter for the language server, i.e. [DotE] nodes and the left
+     expression of a [DotE] node (the latter is handled in [exp]). *)
+  let should_arrange_exp_typ ?arrange_typ e =
+    match e.it with
+    | DotE _ -> true
+    | _ -> Option.value ~default:false arrange_typ
+
+  let rec exp ?arrange_typ e = source e.at (annot ~arrange_typ:(should_arrange_exp_typ ?arrange_typ e) e.note (match e.it with
     | VarE x              -> "VarE"      $$ [id x]
```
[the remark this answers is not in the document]
heitor-lassarote: We are only concerned with the definitions of field sources in this PR (since prior to our work we could not find the definition of `f` from `o.f`). For ordinary variables, the existing code in the VSCode extension traverses the AST to find its definition. No type resolution is done for this case.

src/pipeline/test_field_srcs.ml
```diff
@@ -60,11 +60,11 @@ let gather_let_srcs sexpr =
   in
   List.of_seq @@ Sexpr_set.to_seq @@ go Sexpr_set.empty sexpr
 
-let arrange filename : (module Mo_def.Arrange.S) =
+let arrange filename srcs_tbl : (module Mo_def.Arrange.S) =
```
crusso, line 63: Ha, I finally encounter a first-class module in the wild...

crusso (approved):
Ok, thanks for making the changes!

I wonder if this is still a little complicated and users won't actually understand the info it provides, but I guess we can just try it and simplify later if necessary.

src/pipeline/test_field_srcs.ml
```diff
@@ -136,44 +148,53 @@ let%expect_test "" =
...
-      (@@ (Pos test-field-srcs.mo 3 16) (Pos test-field-srcs.mo 3 20))
-      (@@ (Pos test-field-srcs.mo 9 16) (Pos test-field-srcs.mo 9 20))
-    ) |}]
+      (@@ test-field-srcs.mo (Pos 3 16) (Pos 3 20))
+      (@@ test-field-srcs.mo (Pos 9 16) (Pos 9 20))
+    )
+    Sources table:
+    test-field-srcs.mo:3.17-3.21: test-field-srcs.mo:3.17-3.21
+    test-field-srcs.mo:9.17-9.21: test-field-srcs.mo:3.17-3.21 test-field-srcs.mo:9.17-9.21
+    test-field-srcs.mo:14.15-14.19: test-field-srcs.mo:14.15-14.19 |}]
```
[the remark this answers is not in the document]
heitor-lassarote: > Odd, should that last one 14.5 (c1.meth) point to 9.17 (C2.meth)? Or does it only point to itself because it can't determine where the method came from?

The source for 14.15 is for `test`. But now that you mention it, we should also have `Class1` and `Class2`, which weren't added since we are currently only adding public fields. I've patched it to also include private fields.

> Does 9.17 (Class2().meth) point to 3.17 only because of the subtyping in let c1 : Class2 = Class1 ()

Yes.

> I wonder if asking for field definitions even make sense in the context of Motoko, and all we should expect goto-definition and goto references to do is map from an identifier to its binding and back though I guess that woudn't help with goto-definition for a function defined in (the field) of a library.

I think it's generally useful for users to be able to ask the editor where the field came from for better code discovery. A lot of the changes done here were mostly to try to handle corner cases, but in practice, I don't think users will have very complicated subtyping relations in most code. It's much better than `grep`ing for those.

heitor-lassarote:
Force-pushed to rebase with `master`.

heitor-lassarote:
@crusso, may I ask you to take another look at the latest changes, please? :slightly_smiling_face: I hope all points should be addressed now, and the PR is rebased onto `master`.

crusso:
Hmm, I think we might need anothe rebase onto master. Sorry about that.

src/mo_frontend/test_recovery.ml
```diff
@@ -256,9 +397,102 @@ actor Main {
     )
 
      with errors:
-    (unknown location): syntax error [M0001], unexpected token 'public'
-
-    (unknown location): syntax error [M0001], unexpected token '(' |}]
+    (unknown location): syntax error [M0001], unexpected token 'public', expected one of token or <phrase> sequence:
+      } (e.g. '}')
```
crusso, line 401: Something has gone wrong - did you some revert the changes that @GoPavel  adopted from https://github.com/dfinity/motoko/pull/5108? We had managed to condense all the suggestions in a much shorter list.

heitor-lassarote: I discussed the matter with Pavel, it looks like those changes weren't merged into this PR. Instead, this was being developed here: https://github.com/serokell/motoko/pull/58

I've just merged them into this PR. Could you check whether it's fine now?

crusso:
> > @heitor-lassarote @GoPavel I think something has regressed on this branch - the syntax error suggestions have suddenly become much longer again?
> 
> I've merged [serokell#58](https://github.com/serokell/motoko/pull/58) (the PR where Pavel was improving the syntax error messages) into this PR, hopefully, it should be fine now.

Great, thanks! Perhaps I got confused by the separate PRs. Can you merge with out master? I'm hoping that fixed the CI failures (not sure why it is failing now).

heitor-lassarote:
> > > @heitor-lassarote @GoPavel I think something has regressed on this branch - the syntax error suggestions have suddenly become much longer again?
> > 
> > 
> > I've merged [serokell#58](https://github.com/serokell/motoko/pull/58) (the PR where Pavel was improving the syntax error messages) into this PR, hopefully, it should be fine now.
> 
> Great, thanks! Perhaps I got confused by the separate PRs. Can you merge with out master? I'm hoping that fixed the CI failures (not sure why it is failing now).

Done. Hopefully CI becomes green now.

crusso:
Hmm, I wonder why this is still failing. @basvandijk is there a problem with running CI on forked PRs?

basvandijk:
@crusso 
> is there a problem with running CI on forked PRs?

Yes, GitHub Actions triggered by a pull request coming from a fork don't have access to secrets. It would be easy for an attacker to steal to secrets otherwise.

So what to do? I see two options:

* We could make the workflow not fail when it doesn't have access to the `CACHIX_AUTH_TOKEN` by not enabling the cache in that case.

* Create a copy of the PR in your name from an internal branch. Then CI will run on it. This will make it harder for Serokell to make iterative changes of course since you'll need to push them to the internal branch every time.

crusso:
How strange, for some reason it is running now. Maybe because I re-ran the workflows?

heitor-lassarote:
> How strange, for some reason it is running now. Maybe because I re-ran the workflows?

Looks like the running checks here are the same as your branch. Even the "Successful in %d+m" matches. I don't know the reason, but perhaps it's because the commit SHAs are the same?

crusso:
> > How strange, for some reason it is running now. Maybe because I re-ran the workflows?
> 
> Looks like the running checks here are the same as your branch. Even the "Successful in \d+m" matches. I don't know the reason, but perhaps it's because the commit SHAs are the same?

Hurrah, it's in!

PR #5077, opened 2025-04-23
chore: only support flakes for from and to in `perf-delta.nix`

basvandijk:
Now that https://github.com/dfinity/motoko/pull/5067 is merged it means the base of any PR is a flake so we no longer have to support the case where it isn't. So this simplifies `perf-delta.nix` to only support flakes.

### DIFF
--- a/perf-delta.nix
+++ b/perf-delta.nix
@@ -54,46 +54,22 @@ let
 
   checkout = rev: builtins.fetchGit { url = ./.; ref = ref; inherit rev; };
 
-  baseCheckout = checkout from;
-
-  isBaseFlake = builtins.hasAttr "flake.nix" (builtins.readDir baseCheckout);
-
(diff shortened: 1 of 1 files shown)


ggreif (approved):
Rubber-stamping, but what about long-running PRs that started before the flake era? Should they merge from `master`? Is that enough?

basvandijk:
> what about long-running PRs that started before the flake era? Should they merge from master? Is that enough?
That wouldn't hurt but I think it's not necessary since the `from` is defined as [the merge-base of the HEAD of the PR with the base branch](https://github.com/dfinity/motoko/blob/master/.github/workflows/test.yml#L77). Which means that if the PR branch was branched before the move to flakes the code would still use `default.nix` and `from` and `to` would both point to commits before the move to flakes.

PR #5085, opened 2025-04-24
chore: add missing tools to get `make -C test` working again

ggreif:
Not sure about `direnv`'s needs, though...


christoph-dfinity (approved):
`direnv` should just be able to pick these up. Thanks!

PR #5089, opened 2025-04-25
chore: allow `drun` to run as app subnet

alexandru-uta:
`run-drun` tests can now run as an application subnet, which is important for certain kinds of tests where subnet types differ.

To run under this mode, one needs to add
```
# APPLICATION-SUBNET
```

in their `.drun` test file. The default is still running as a system subnet.


Kamirus (approved):
LGTM!

PR #5091, opened 2025-04-25
docs restructure: chapter 3

jessiemongeon1:


doc/md/3-declarations/1-variable-declarations.md
```diff
@@ -0,0 +1,49 @@
...
+## Immutable variables
+
+A `let` declaration names the value of an expression so that the expression can be evaluated once and its result, a value, referenced by that name. The compiler ensures that any attempt to modify an immutable variable results in an error.
+
+```motoko no-repl
+let x = 10;
+x := 20; // Error: Cannot assign to immutable variable
+```
+
+## Mutable variables
```
jessiemongeon1, line 27: Perhaps mention that mutable variables do not support pattern matching. i.e. var (a, b) = (1, 2) is not supported. For comparison, maybe also provide an example let binding with pattern matching.

doc/md/3-declarations/2-function-declarations.md
```diff
@@ -0,0 +1,127 @@
...
+applyFunction(double, 10);
+```
+
+`applyFunction` takes a function `f` and applies it to `value`.  `double` is passed as an parameter, allowing to construct higher-order functions.
+
+## Anonymous functions
+
+Motoko supports functions without names (anonymous). They can be used in expressions or passed as arguments.
+
+```motoko no-repl
```
jessiemongeon1, line 51: A better example might be func twice(f) : Int -> Int = func (x) { f(f(x)) };

This take a function as an argument and returns an anonymous function as its result.
twice(func x = x+1)(2) returns 4. or something.

doc/md/3-declarations/3-object-declaration.md
```diff
@@ -0,0 +1,73 @@
...
+| Private fields | No | Yes |
+| Recursively defined fields | No | Yes |
+| Mutability | Supports [`let` (immutable) and `var` (mutable)](https://internetcomputer.org/docs/motoko/fundamentals/declarations/variable-declarations) | [Supports `let` (immutable) and `var` (mutable)](https://internetcomputer.org/docs/motoko/fundamentals/declarations/variable-declarations) |
+| Supports field visibility (`public`, `private`) | No | Yes |
+| Combination (`and`, `with`) | Yes | Yes |
+| Supports async behavior | No | Yes (supports `async` functions) |
+
+### Record example
+
+Records are typically used to store structured data. Fields are always accessible using [dot notation](https://en.wikipedia.org/wiki/Object-oriented_programming). Records do not support visibility modifiers (`public`, `private`). However, they do support `and` and `with` for combination and modification.
```
jessiemongeon1, line 23: maybe give examples for and and with?

doc/md/3-declarations/3-object-declaration.md
```diff
@@ -0,0 +1,73 @@
...
+
+Debug.print(ghost.name); // "Motoko"
+ghost.age += 1; // Updates age
+```
+
+### Object example
+
+Objects extend records by supporting functions. Fields can have explicit visibility modifiers, such as `public` or `private`. Objects support async functions and structured behavior.
+
+```motoko no-repl
```
jessiemongeon1, line 39: Maybe use a better example that needs encapsulate state (and increasing counter, though boring, is good for this, or a log.

Demali-876:
```motoko no-repl
import List "mo:base/List";
import Debug "mo:base/Debug";

let logger = object {
    private var logs = List.empty<Text>()
    
    public func add(message : Text) : () {
        List.add(logs, message);
        Debug.print("New entry: " # message);
    };
    
    public func clear() : () {
        logs.clear();
        Debug.print("Logs cleared");
    };
    
    public func size() : Nat {
        logs.size();
    };
};
```

doc/md/3-declarations/1-variable-declarations.md
Demali-876: :::info
Mutable variables declared with `var` do not support pattern matching. For example, the following is invalid:

```motoko no-repl
var (a, b) = (1, 2); // Not supported
```

In contrast, `let` bindings do support pattern matching:

```motoko
let (a, b) = (1, 2); // Supported
```
:::

doc/md/3-declarations/3-object-declaration.md
Demali-876: This is covered in section 2 types 5-Records. We can link it or we make other examples. Might be easier to link because it is covered comprehensively on that page.

Demali-876 (commented):
I've completed a review of this section

Remaining comments:

1. Review of related outline 

Segment: **Declarations**
- Variable declarations: let/var 
    - Applied suggestions from the team(re: mutable variables)
- Function definitions 
   - Applied suggestions for anonymous functions.
- Object declarations
   - New object example implemented
- Classes 
   - Updated example for class
- Type declarations
   - Update variant example
- Expression declarations

2. All instances of `actor` should be identified and replaced with the `persistent actor` syntax.
3. Any instance of an early `return` should be omitted.

4. In contrast to local-objects-classes on the master branch this section breaks various concepts related to objects and classes into small sections. It seems to be missing actor classes after reviewing this section.

doc/md/3-declarations/4-class-declarations.md
```diff
@@ -0,0 +1,55 @@
+---
+sidebar_position: 4
+---
+
+# Class declarations
```
Kamirus, line 5: We've got some feedback on the forum about the explanation of Motoko classes. It might be a good idea to incorporate some of these ideas. I see that the current description matches quite well the suggestions in the post. But is it still lacking the info about mimicking `this` / `self`. Maybe another section on that?

> The docs are good, but they could use some simplification. They start with inner-workings and only make sense if you read them completely (a sign that they were written by a deeply technical person). For example, “Objects Classes” section starts with In Motoko, an object encapsulates state, and an object "class" is a package of two entities that share a common name. A package? Two entities? What?
When you read the section carefully and completely you get that the author is trying to say, that there are no “classes” or OOP in the usual meaning and Motoko’s classes are just a sprinkle of sugar on top of objects and functions, but this is not what I want at my first steps with Motoko. I tried laguages already, I know that stuff can be implemented in f-d up ways under-the-hood, this is not useful to me yet. Now I want to know, if this abstraction would allow me to achieve what I want to achieve or I should use something else.
And ideal version of this section for me would be something like: “Classes in Motoko are not real classes like in other OOP languages, but they serve the same purpose. You use classes when you want to be able to create multiple instances of objects of the same shape. For example, here is a human class: … There are no “this” or “self” keywords in Motoko, because of the inner-workings, but here is how you can hack yourself one to make it feel more familiar: … --  https://forum.dfinity.org/t/motoko-vs-rust-trade-offs-now-and-into-the-future/45093/7

PR #5092, opened 2025-04-25
docs restructure: chapter 4

jessiemongeon1:


doc/md/fundamentals/4-control-flow/1-basic-control-flow.md
```diff
@@ -0,0 +1,184 @@
...
+
+``` motoko no-repl
+let numbers = [1, 2, 3, 4, 5];
+for (num in numbers.vals()) {
+    Debug.print(Nat.toText(num));
+}
+```
+
+## `let-else`
+
```
jessiemongeon1, line 122: Need to update this example to show the let...else syntax, e.g. let ?a = ?5 else assert false

how about something like this?

let ?{ name; age } = person else return;  
OR

let ?(x, y) = maybePair else return;  
would need to double check the code, I just handwrote it here.

doc/md/fundamentals/4-control-flow/3-conditionals.md
```diff
@@ -0,0 +1,46 @@
...
+
+## `if` expression
+
+```motoko no-repl
+let identity : Text = if (x == 1) "x is 1" else "x is not 1"; // Produces a value
+```
+
+The result of the `if` block is assigned to `identity`. Both branches must return a matching type ([`Text`](https://internetcomputer.org/docs/motoko/base/Text) in this case).
+
+## `if` statement
```
jessiemongeon1, line 35: Might be worth moving this above the "if expression" section since it's likely to be more familiar to a larger percentage of developers.

doc/md/fundamentals/4-control-flow/1-basic-control-flow.md
```diff
@@ -0,0 +1,184 @@
...
+    };
+}
+```
+
+## `switch`
+
+A `switch` matches a value against multiple cases and executes the corresponding block of code.
+
+```motoko no-repl
+type HttpRequestStatus = {
```
jessiemongeon1, line 38: This example needs to be fixed

doc/md/fundamentals/4-control-flow/1-basic-control-flow.md
```diff
@@ -0,0 +1,184 @@
...
+| `option block` | Evaluates an expression and wraps the result in an option type, allowing scoped handling of `null` values. |
+| `label/break` | Allows exiting loops early. |
+| `while` | Runs while a condition is `true`. |
+| `for` | Iterates over elements in a collection, terminating when no elements remain. |
+| `let-else` | Conditionally binds values. |
+
+A return statement immediately exits a function and provides a result. Unlike `break` or `continue`, `return` stops execution entirely and returns a value to the caller.
+
+```motoko no-repl
+func add(a : Nat, b : Nat) : Nat {
```
jessiemongeon1, line 22: this example needs to be fixed

doc/md/fundamentals/4-control-flow/3-conditionals.md
```diff
@@ -0,0 +1,46 @@
...
+# Conditionals
+
+| If-expression                                        | If-statement                              |
+|------------------------------------------------------|-------------------------------------------|
+| Type `T` (must be consistent across branches)              | `()` (does not return a value)           |
+| Produces a result that can be used or assigned. | Executes a block of code but does not return a value. |
+| Can be assigned to a variable.               | Executes for side effects only.       |
+| Used when a decision determines a value.       | Used when a decision triggers an action.  |
+
+An `if-else` expression allows a program to make decisions based on a condition. It evaluates a boolean expression and selects one of two possible execution paths.
```
jessiemongeon1, line 14: Maybe explain first what an if-expression and an if-statement are? or give an early example, otherwise the "type T discussion can be pretty abrupt.

doc/md/fundamentals/4-control-flow/3-conditionals.md
```diff
@@ -0,0 +1,46 @@
...
+
+```motoko no-repl
+var age = 21;
```
jessiemongeon1, line 17: I find this example a bit confusing for this place in the text. either make explicit prints for the messages ("you are a minor"), or make it an expression (like below)?

PR #5096, opened 2025-04-25
docs restructure: chapter 7

jessiemongeon1:


doc/md/5-actors-async.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+Another customer asks immediately after:
+
+```motoko no-repl
+let status2 = await pizzaParlor.getDeliveryStatus("margherita");
+// "Your margherita will arrive in 25 minutes."
+```
+
+## `await` vs `await*`
```
jessiemongeon1, line 108: We might want some concrete examples of this, since it's both a commonly-used and unintuitive part of the language.

doc/md/8-modules-imports.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+A prefix is not required for local imports and the `.mo` file extension is omitted. The imported modules must be in the same directory as `main.mo`.
+
+```motoko no-repl
+import Types "types";
+import Utils "utils";
+```
+
+## Importing from another package
```
jessiemongeon1, line 62: Maybe we could split this into two sections: "Importing from another directory" and "Importing from a package" (perhaps combining with the section below) with 1-2 examples for each?

doc/md/10-pattern-matching.md
```diff
@@ -0,0 +1,91 @@
...
+
+# Pattern matching
+
+Pattern matching allows concise and expressive handling of different data structures in Motoko. It can be used in `switch` expressions to destructure and process values.
+
+Motoko supports several types of patterns:
+
+| Pattern type | Description | Example |
+|-------------|-------------|---------|
+| Wildcard (`_`) | Matches any value without binding it to a variable. | `switch (x) { case (_) { ... } }` |
```
jessiemongeon1, line 13: Maybe it's better to avoid the parenthesis around _?
Also with literals?

doc/md/10-pattern-matching.md
```diff
@@ -0,0 +1,91 @@
...
+func getValue(opt : ?Nat) : Text {
+    switch opt {
+        // Parentheses required around `?n` 
+        case (?n) { "Value: " # Nat.toText(n) };
+        case null { "No value" };
+    };
+};
+```
+
+## Object
```
jessiemongeon1, line 62: Mention that the records have the same pattern.
Missing section on tuples.
Any other patterns missing?

doc/md/10-pattern-matching.md
```diff
@@ -0,0 +1,91 @@
...
+
+Objects with named fields can be matched to extract specific properties.
+
+```motoko no-repl
+type Person = { name : Text; age : Nat };
+
+func describePerson(person : Person) : Text {
+    switch (person) {
```
jessiemongeon1, line 70: I understand the need to show case that such patterns exist, but this also showcases an anti-pattern. In this case you'd always deconstruct the person in the method signature, or using a single let to avoid unnecessary nesting with switch

doc/md/11-error-handling.md
```diff
@@ -0,0 +1,99 @@
+---
+sidebar_position: 11
+---
+
+# Error handling
```
jessiemongeon1, line 5: Mention that exception handling works only in async context.
We already had some confusion around the fact that the assert inside the try block will not be caught -- this is worth documenting somewhere in this chapter.

doc/md/11-error-handling.md
```diff
@@ -0,0 +1,99 @@
...
+
+func divide(a : Nat, b : Nat) : Nat {
+    if (b == 0) {
+        Debug.trap("Cannot divide by zero");
+    };
+    return a / b;
+};
+```
+
+## Assertions
```
jessiemongeon1, line 35: And here add all the info about debug assert and --release moc flag and Release option from dfx.
See the comment in the style guide -- add a link to this section there

doc/md/11-error-handling.md
```diff
@@ -0,0 +1,99 @@
...
+
+Assertions enforce expected conditions. If the condition is false, they introduce a trap but are not errors themselves.
+
+```motoko no-repl
+func validateAge(age : Nat) : () {
+    assert(age >= 18);  // Traps if age is below 18
```
jessiemongeon1, line 41:
```suggestion
    assert age >= 18;  // Traps if age is below 18  
```

doc/md/11-error-handling.md
```diff
@@ -0,0 +1,99 @@
...
+};
+```
+
+## Error codes
+
+| Error Code | Description | Example | Explanation |
+|------------|-------------------|---------|-------------|
+| M0001 | Parsing errors. | `let x = 5 + ;` | Syntax error - missing operand after the `+` operator. |
+| M0003 | Module tried to import itself. | `import Self "self.mo"` | A module cannot import itself, creating a circular dependency. |
+| M0009 | File not found for import. | `import Lib "wrong/path_to_file.mo"` | The specified file path in the import statement doesn't exist. |
```
jessiemongeon1, line 51: Do we really want that info? Here? These are a bit different errors 😄 Could be nice to have all the moc errors documented. These could be linked in the IDE so that users could learn more about the error. But the table below should be autogenerated, do we have any guarantees that the error codes will not change?
Could be ok to include it for now, but maybe in a separate section, separate file?

doc/md/12-garbage-collection.md
```diff
@@ -0,0 +1,46 @@
+---
+sidebar_position: 12
+---
+
+# Garbage collection
```
jessiemongeon1, line 5: The incremental GC is not beta afaik and is the default GC for Enhanced OP. We need to mention this somewhere here as well.

doc/md/13-orthogonal-persistence.md
```diff
@@ -0,0 +1,105 @@
...
+- Removing object fields.
+- Adding variant fields.
+- Changing [`Nat`](https://internetcomputer.org/docs/motoko/base/Nat) to [`Int`](https://internetcomputer.org/docs/motoko/base/Int).
+- Adjustments to shared function parameters and return types.
+
+If an upgrade is incompatible, the system **automatically rolls it back** to prevent data loss. More complex changes require explicit migration.
+
+When switching to enhanced persistence, the system performs one final deserialization from stable memory before transitioning fully to memory retention. Once a canister is using enhanced persistence, it cannot be downgraded back to classical persistence. Legacy stable memory remains accessible as secondary storage, allowing developers to combine stable regions with orthogonal persistence if needed.
+
+ICP provides options for Wasm memory retention during upgrades:
```
jessiemongeon1, line 58: I think we either need to get into a bit more details here or not get into this at all; This wasm_memory_persistence is a canister setting, but the notion of canister settings is never even mentioned here; [@crusso](https://crispy-robot-ggpxwq95g543wwgq.github.dev/) , any thoughts on this?

doc/md/5-actors-async.md
```diff
@@ -0,0 +1,161 @@
...
+
+Motoko adopts the actor model to enable scalable, non-blocking execution on the Internet Computer. Motoko actors run in a single-threaded environment but achieve high concurrency by handling requests [asynchronously](https://internetcomputer.org/docs/motoko/fundamentals/actors-async#async--await).
+
+## Async actors
+
+To demonstrate how asynchronous actors work, consider the following example.
+
+Customers place orders at a pizza restaurant, but the chef can only make one pizza at a time. Orders are taken **[asynchronously](https://internetcomputer.org/docs/motoko/fundamentals/actors-async#async--await)**, meaning customers do not have to wait for previous orders to be completed before placing their own. However, each pizza is prepared sequentially. This is representative of an asynchronous actor.
+
+```motoko no-repl
```
jessiemongeon1, line 24: implicitly shared right? So maybe we should avoid that? Perhaps making a note in the comment that public actor functions are shared by default

Demali-8762 days ago
This is stated elsewhere but i guess we can put an admonition here, I wonder, is it encouraged to include the shared keyword or because its public hence shared not necessarily required also doesn’t shared imply additional functionalities like being able to be called by external canisters/frontends or do i have that wrong?

Kamirus2 days ago
AFAIK shared [is just redundant here](https://crispy-robot-ggpxwq95g543wwgq.github.dev/) because this is a public function of an actor. Just like we omit private for functions that are private by default. We could be explicit here ofc, but I'd write these examples with the idiomatic code. Also, in that link the example below uses a persistent actor with a count, omitting shared and stable

crusso1 day ago
Public actor methods are always implicitly shared (indeed the parser just inserts the shared modifier). The only reason you'd want to explicitly mark it shared is for clarity or to access the caller (which follows the shared keyword).

doc/md/5-actors-async.md
```diff
@@ -0,0 +1,161 @@
...
+await PizzaParlor.placeOrder("BBQ"); // Order received: BBQ
+```
+
+This request is handled [asynchronously](https://internetcomputer.org/docs/motoko/fundamentals/actors-async#async--await), meaning other orders can be placed without waiting for a response. The actor remains responsive, even while executing other tasks.
+
+## `async*` / `await*`
+
+Customers can place multiple orders at the same time. However, when they ask for an update on their delivery status, responses must be generated sequentially based on the preparation time. Instead of constantly tracking delivery updates, `async*` can be used to calculate the estimated delivery time only when requested.
```
jessiemongeon1, line 71: After reading that I still don't understand what async* / await* do 😄 The example below does not help either. checkDelivery could be pure? right? So this is probably not a good example. Maybe sth that uses the await* inside an async* function?

Demali-8762 days ago
In other languages async gives a future(same as motoko) and async* gives a stream that can yield values as defined by the stream type (motoko does not have this) instead its supposed to be used for internal processes that may be async, in this example its calculating the wait time for each customer(which could be synchronous or async).

The await* expression(as i understand it) works by evaluating its expression to a result, and if that result is a delayed computation, it executes that computation synchronously at first. It only switches to asynchronous execution requiring the messaging system when (and if) it encounters a regular await inside that computation. This means await* doesn't automatically commit state changes or suspend execution like regular await does. The * is actually from regular expressions, so its saying zero or more async computations. Maybe i should add that too.

As Claudio explained it to me, its intended to be an optimization for on demand computation without suspending execution unless there an await inside it. This example is intended to explain that without overcomplicating it. Though I do realise the * is missing from status1 and status2 so I will fix that, Maybe [@crusso](https://crispy-robot-ggpxwq95g543wwgq.github.dev/) can share some thoughts on this.

Kamirus1 day ago
This is a great explanation! Maybe we could incorporate that here. For now there is not much intuition described behind await*. This could prove useful for user's understanding

doc/md/5-actors-async.md
```diff
@@ -0,0 +1,161 @@
...
+
+## `await` vs `await*`
+
+| Feature         | `await`                      | `await*`                        |
+|---------------|-----------------------------|--------------------------------|
+| Execution     | Suspends execution until the result is available. | Does not suspend execution unless an `await` exists inside `async*`. |
+| State commit | Commits all [state](https://internetcomputer.org/docs/motoko/fundamentals/state) changes before resuming. | Does not commit [state](https://internetcomputer.org/docs/motoko/fundamentals/state) unless a proper `await` is inside. |
+| Re-evaluation | Runs once, returning the same value if awaited multiple times. | Each `await*` triggers a fresh computation of `async*`. |
+
+## `try/finally`
```
jessiemongeon1, line 116: what about catch? I'd describe when the exceptions are exactly caught and when not

doc/md/6-state.md
```diff
@@ -0,0 +1,45 @@
...
+
+Motoko enforces strict control over mutable state, ensuring that concurrent execution remains predictable and error-free.
+
+The following actor maintains a private mutable counter that can only be modified through its own functions:
+
+```motoko no-repl
+actor Counter {
+    stable var count : Nat = 0; // Private mutable state
+
+    public shared func increment() : async Nat {
```
jessiemongeon1, line 31: I'd keep the examples with the current standard which is using persistent actor and avoiding stable and shared when possible, like in [the example below here](https://crispy-robot-ggpxwq95g543wwgq.github.dev/)

Suggested change:

persistent actor Counter {  
    var count : Nat = 0; // Private mutable state  

    public func increment() : async Nat {

doc/md/8-modules-imports.md
```diff
@@ -0,0 +1,184 @@
...
+ ├── main.mo   // Contains the main actor
+ ├── types.mo  // Stores type definitions
+ ├── utils.mo  // Contains helper functions
+```
+
+A prefix is not required for local imports and the `.mo` file extension is omitted. The imported modules must be in the same directory as `main.mo`.
+
+```motoko no-repl
+import Types "types";
+import Utils "utils";
```
jessiemongeon1, line 59: The convention is to start filenames with an upper-case letter, just like types. So that modules, types and files would have the same name

doc/md/9-data-persistence.md
```diff
@@ -0,0 +1,259 @@
...
+- Changing the mutability of a field.
+- Removing fields from a [record](https://internetcomputer.org/docs/motoko/fundamentals/types/records) type.
+- Adding new fields to a [variant](https://internetcomputer.org/docs/motoko/fundamentals/types/variants) type.
+- Changing [`Nat`](https://internetcomputer.org/docs/motoko/base/Nat) to [`Int`](https://internetcomputer.org/docs/motoko/base/Int) (since [`Nat`](https://internetcomputer.org/docs/motoko/base/Nat) is a subtype of [`Int`](https://internetcomputer.org/docs/motoko/base/Int).
+- Adjustments in shared function parameters and return types.
+- Any transformation allowed by Motoko’s subtyping rules.
+
+If a change falls within these categories, no manual migration is required. The system ensures that stable variables persist correctly across upgrades.
+
+### Explicit migration
```
jessiemongeon1, line 191: We should probably mention the new migration syntax (with migration = ...

doc/md/8-modules-imports.md
```diff
@@ -0,0 +1,184 @@
...
+It can be imported into another file:
+
+```motoko no-repl
+import Counters "Counters";
+import Debug "mo:base/Debug";
+import Nat "mo:base/Nat";
+
+persistent actor CountToTen {
+  public func countToTen() : async () {
+    let counterActor = await Counters.Counter(1);
```
Demali-876, line 128: let counterActor = await (with cycles = 1_000_000_000) Counters.Counter(1);

doc/md/8-modules-imports.md
Demali-876: ├── Main.mo   // Contains the main actor
 ├── Types.mo  // Stores type definitions
 ├── Utils.mo  // Contains helper functions

doc/md/10-pattern-matching.md
ggreif: and in `switch (x)` too

ggreif: In Motoko "that single thing" rarely needs parentheses. I know it is _muscle memory_ when coming from C/JS etc.

doc/md/12-garbage-collection.md
```diff
@@ -0,0 +1,46 @@
...
+
+Garbage collection is an automatic process that manages memory usage by removing unreferenced or dead objects and freeing up allocated heap memory.
```
luc-blaeser, line 7:
```suggestion
Garbage collection is an automatic process that manages memory usage by detecting unreachable objects and freeing up their allocated heap memory.
```

doc/md/12-garbage-collection.md
```diff
@@ -0,0 +1,46 @@
...
+
+Motoko’s default garbage collection method uses a copying approach that depends on the amount of heap memory currently in use. An alternative garbage collector uses a marking approach, which instead considers the amount of free heap memory. These garbage collectors are triggered when enough changes have been made to heap memory since the last collection cycle. Garbage collection can be forced to run after every message using the `--force-gc` flag in the project's `dfx.json` file:
```
luc-blaeser, line 9: The alternative GC would be the incremental one?

doc/md/5-actors-async.md
ggreif: Also there is no messaging happening unless one forces it (with `await*`).

doc/md/12-garbage-collection.md
luc-blaeser, line 9: I think we should not mention `--force-gc`. This is only for testing purposes. Maybe we should instead mention last-resort GC calls `__motoko_gc_trigger()` as an expert option.

doc/md/12-garbage-collection.md
```diff
@@ -0,0 +1,46 @@
...
+```json
+"defaults": {
+  "build": {
+    "packtool": "",
+    "args": "--force-gc"
+  }
+}
+```
+
+Both garbage collectors are limited by the ICP instruction limit per message, preventing them from collecting the entire heap memory pool in a single execution. As a result, [canisters](https://internetcomputer.org/docs/building-apps/essentials/canisters) cannot fully utilize the entire 4GiB memory pool, as some free space must be reserved for the garbage collector to operate.
```
luc-blaeser, line 20: Both GCs? Which would be the other one? The incremental GC is not affected by this.
I think we should simply recommend using incremental GC if people have larger memory. Otherwise having this note looks quite bad for Motoko especially as we have a solution.

luc-blaeser, line 20:
```suggestion
Both garbage collectors are limited by the ICP instruction limit per message, preventing them from collecting the entire heap memory pool in a single execution. As a result, [canisters](https://internetcomputer.org/docs/building-apps/essentials/canisters) cannot fully utilize the entire 4GiB memory pool. The limit depends on the heap constellation, amount of live objects etc.
```
The note on free space for GC is not correct reason for limitation.

doc/md/12-garbage-collection.md
```diff
@@ -0,0 +1,46 @@
...
+
+A beta incremental garbage collection process is available, which distributes the workload across multiple messages as needed. This approach allows canisters to allocate up to three times more heap space after collection while consuming fewer cycles on average. With incremental garbage collection, canisters can take advantage of the entire 4GiB heap memory pool.
```
luc-blaeser, line 22:
```suggestion
An incremental garbage collection process is available and recommended to be used for larger heaps, which distributes the workload across multiple messages as needed. This approach allows canisters to use the entire heap space (with some reserve maintained for collection). With incremental garbage collection, canisters can take advantage of the entire heap memory (4 GB for 32-bit and more with 64-bit).
```

doc/md/12-garbage-collection.md
```diff
@@ -0,0 +1,46 @@
...
+
+The incremental garbage collector can be enabled by specifying the `--incremental-gc` compiler flag in the project's `dfx.json` file.
```
luc-blaeser, line 24:
```suggestion
The incremental garbage collector is enabled by specifying the `--incremental-gc` compiler flag in the project's `dfx.json` file.
```

doc/md/12-garbage-collection.md
luc-blaeser: No longer beta

doc/md/12-garbage-collection.md
```diff
@@ -0,0 +1,46 @@
...
+
+:::caution
+
+This garbage collector is still in beta and should be used with caution.
```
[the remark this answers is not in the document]
luc-blaeser: No longer beta

doc/md/12-garbage-collection.md
```diff
@@ -0,0 +1,46 @@
...
+
+:::
+
+
+```json
```
luc-blaeser, line 33: A good source for more background knowledge: https://dl.acm.org/doi/10.1145/3623507.3627672
Maybe we should even reference this.

doc/md/11-error-handling.md
jessiemongeon1: Including for now, will revisit in future

doc/md/13-orthogonal-persistence.md
jessiemongeon1: Removed

doc/md/5-actors-async.md
jessiemongeon1: removed shared

PR #5097, opened 2025-04-26
remove weird rule

christoph-dfinity:
Just want to see if CI passes


src/mo_frontend/parser.mly
```diff
@@ -786,8 +786,6 @@ exp_nonvar(B) :
 exp [@recover.expr mk_stub_expr loc] (B) :
   | e=exp_nonvar(B)
     { e }
-  | d=dec_var
-    { match d.it with ExpD e -> e | _ -> BlockE([d]) @? at $sloc }
```
crusso, line 790: The match is even redundant because dec_var always return VarD AFAICT

christoph-dfinity: Yeah, that's what initially made me go "Wat?" here...

christoph-dfinity:
I understand why this is here now

PR #5119, opened 2025-05-05
chore: add nightly macos intel x86 job

alexandru-uta:


crusso (approved):
More of a rubberstamp than a review ;->

PR #5125, opened 2025-05-07
chore: relax `dune` profile for local compilation

ggreif:
A recent change to `src/Makefile` had the unintended effect of `make -C src` erroring out on harmless warnings, like `assert false; ...`, etc.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -18,7 +18,7 @@ NATIVE_TARGETS = $(MO_LD_TARGET) $(MO_DOC_TARGET) $(MOC_TARGET) $(DIDC_TARGET) $
 JS_TARGETS = $(MOC_JS_TARGET) $(DIDC_JS_TARGET) $(MOC_INTERPRETER_TARGET)
 ALL_TARGETS = $(NATIVE_TARGETS) $(JS_TARGETS)
 
-.PHONY: all clean exes grammar candid-tests unit-tests test-quick test-parallel test format
+.PHONY: all clean exes grammar unit-tests test-quick test-parallel test format
 
 # let make update check this file every time
 SOURCE_ID = source_id/generated.ml
@@ -73,6 +73,7 @@ deser: $(SOURCE_ID)
 
 candid-tests: $(SOURCE_ID)
 	dune build $(DUNE_OPTS) $(CANDID_TESTS_TARGET)
+	@ln -fs $(CANDID_TESTS_TARGET) $@
 
 .PHONY: $(SOURCE_ID)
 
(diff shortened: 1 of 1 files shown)


src/Makefile
```diff
@@ -49,17 +49,17 @@ mo-doc: $(SOURCE_ID)
 	@ln -fs $(MO_DOC_TARGET) mo-doc
 
 moc.js: $(SOURCE_ID)
-	dune build $(DUNE_OPTS) $(MOC_JS_TARGET)
+	dune build $(DUNE_OPTS) --profile=release $(MOC_JS_TARGET)
 	@ln -fs $(MOC_JS_TARGET) moc.js
```
ggreif, line 53:
```suggestion
	@ln -fs $(MOC_JS_TARGET) $@
```

src/Makefile
```diff
@@ -49,17 +49,17 @@ mo-doc: $(SOURCE_ID)
...
 	ls -al $(MOC_JS_TARGET)
 
 moc_interpreter.js: $(SOURCE_ID)
-	dune build $(DUNE_OPTS) $(MOC_INTERPRETER_TARGET)
+	dune build $(DUNE_OPTS) --profile=release $(MOC_INTERPRETER_TARGET)
 	@ln -fs $(MOC_INTERPRETER_TARGET) moc_interpreter.js
```
ggreif, line 58:
```suggestion
	@ln -fs $(MOC_INTERPRETER_TARGET) $@
```

src/Makefile
```diff
@@ -49,17 +49,17 @@ mo-doc: $(SOURCE_ID)
...
 	ls -al $(MOC_INTERPRETER_TARGET)
 
 didc.js: $(SOURCE_ID)
-	dune build $(DUNE_OPTS) $(DIDC_JS_TARGET)
+	dune build $(DUNE_OPTS) --profile=release $(DIDC_JS_TARGET)
 	@ln -fs $(DIDC_JS_TARGET) didc.js
```
ggreif, line 63:
```suggestion
	@ln -fs $(DIDC_JS_TARGET) $@
```

src/Makefile
```diff
@@ -49,17 +49,17 @@ mo-doc: $(SOURCE_ID)
 	@ln -fs $(MO_DOC_TARGET) mo-doc
 
 moc.js: $(SOURCE_ID)
-	dune build $(DUNE_OPTS) $(MOC_JS_TARGET)
+	dune build $(DUNE_OPTS) --profile=release $(MOC_JS_TARGET)
```
alexandru-uta, line 52: shall we add a new var that has `--profile=release` and then we link that for the .js targets? Maybe something like `JS_TARGET_OPTS` or so.

ggreif (commented):
Also these are manifest targets, and have no place in the `.PHONY` list

src/Makefile
ggreif: I don't think that is necessary. Nobody builds these targets locally.

Issue #5127, opened 2025-05-08
`mo-doc`: Lost `{}` (record) function arguments


vporton:
Create a valid Motoko file `a/x.mo`:

```motoko
actor {
  public shared func f({a : Int}) {};
}
```

Run command:
```
$ `dfx cache show`/mo-doc --source a --output x --format html
```

Now in  `x/x.html` we have `public func f()` instead of `public shared func f({a: Int})`.

ggreif:
This is probably due to these "self-typing" patterns. I'll have a look, as a similar issue is bugging me.

vporton:
@ggreif `mo-doc` installed together with `moc` with compiling from sources, still creates wrong docs for `packageInstall()` at https://github.com/vporton/icp-package-manager (`make docs`).

vporton:
... and the example `x.mo` above also produced `()`.

ggreif:
Did you rebuild `mo-doc` on `master`? I.e. `make -C src`?

vporton:
I compiled `master` as specified in `Building.md`.

I also tried to download the last release `motoko-Linux-x86_64-0.14.10.tar.gz` and tried to use it for building - same result (nothing inside the `()`).

vporton:
I don't understand when and in which directory to run `which mo-doc`.

ggreif:
In the build directory, which is `git` repo top-level (for me `/Users/ggreif/motoko`).

vporton:
As I found, following `Building.md` is not enough for `mo-doc` to be built. Also:

```
$ make -C src
make: Entering directory '/home/porton/t/motoko/src'
source_id/gen.sh
dune build  _build/default/exes/mo_ld.exe _build/default/exes/mo_doc.exe _build/default/exes/moc.exe _build/default/exes/didc.exe _build/default/exes/deser.exe _build/default/exes/candid_tests.exe
make: dune: No such file or directory
make: *** [Makefile:31: exes] Error 127
make: Leaving directory '/home/porton/t/motoko/src'
```

ggreif:
`dune` should be included in the `nix develop` workflow!

vporton:
> `dune` should be included in the `nix develop` workflow!

Doe `mo-doc` build only in development mode? I don't know how to install it.

vporton:
I  guessed that the right command is `nix profile install .#release.mo-doc`. Please, add this to docs.

ggreif:
There is an official 0.14.11 now!

[...]

PR #5128, opened 2025-05-08
fix: extend extractor to also cover #5127

ggreif:


src/docs/extract.ml
```diff
@@ -87,6 +87,32 @@ type extracted = {
   docs : doc list;
 }
 
+let typed_pat_field : Syntax.pat_field -> bool = function
+  | { it = Syntax.{ pat = { it = AnnotP _; _ }; _ }; _ }
+  | { it = Syntax.{ pat = { it = TupP []; _ }; _ }; _ } ->
```
ggreif, line 92: use or-pattern

src/docs/extract.ml
```diff
@@ -87,6 +87,32 @@ type extracted = {
...
+      true
+  | _ -> false
+
+let pf2tf (pat_field : Syntax.pat_field) =
+  let ann = function
+    | { it = Syntax.AnnotP (_, typ); _ } -> typ.it
+    | { it = Syntax.TupP []; _ } -> Syntax.TupT []
```
ggreif, line 99: because this is a singleton-typed value

src/docs/extract.ml
ggreif, line 92:
```suggestion
  | { it = Syntax.{ pat = { it = AnnotP _ | TupP []; _ }; _ }; _ }
```

christoph-dfinity (commented):
I would expect `function_arg_doc` to turn into a sum type, to support multiple types of arguments here. Let's not add "unforced" hacks if we can...

Why are you special casing the `()` pattern? We're not performing "type inference" for anything else in the docs generation. Maybe I'm missing something but wouldn't you just omit a `x = ()` pattern for a parameter object anyway?

I was also surprised to see we're not rendering documentation comments on arguments yet, given that we collect them (probably my oversight or I couldn't figure out the HTML to generate). But once we do we'll need to also collect the docs on the object pattern fields, so that examples like the following work.

```motoko
module {
  public func myRangeArgumentObject({
    /// Upper bound, inclusive
    upper : Nat;
    /// Upper bound, exclusive
    lower : Nat;
  }) : Iter.Iter<Nat> {
    // Impl
  };

  public func myRangePositional(
    /// Upper bound, inclusive
    upper : Nat,
    /// Upper bound, exclusive
    lower : Nat,
  ) : Iter.Iter<Nat> {
    // Impl
  };
}
```

ggreif:
> I would expect `function_arg_doc` to turn into a sum type, to support multiple types of arguments here. Let's not add "unforced" hacks if we can avoid it.

I am happy to do it in a follow up, but that requires a slightly deeper dive. This PR is about satisfying an immediate need, no architectural work included.

> 
> Why are you special casing the `()` pattern? We're not performing type inference for anything else in the docs generation. Maybe I'm missing something but wouldn't you just omit a `x = ()` pattern for a parameter object anyway?

The answer is "because we can". Due to it being a singleton type not much inference is involved. I can take it out if you insist.

christoph-dfinity:
> I am happy to do it in a follow up, but that requires a slightly deeper dive.

Didn't think it was that deep a dive. See what you think about https://github.com/dfinity/motoko/pull/5134

> The answer is "because we can". Due to it being a singleton type not much inference is involved. I can take it out if you insist.

Not really convinced. `{}`, `((), ())` would qualify just as well, right? Better to just take it out IMO.

christoph-dfinity (approved):

christoph-dfinity:
Thanks for making improvements to `mo-doc`!

PR #5139, opened 2025-05-14
fix: don't collect types during `infer_dec_valdecs`

christoph-dfinity:
If my understanding is correct we will have collected and added all types to the `env` by this point, so we can get rid of some redundant code.

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -3338,7 +3338,7 @@ and is_import d =
 
 and infer_dec_valdecs env dec : Scope.t =
   match dec.it with
-  | ExpD _ ->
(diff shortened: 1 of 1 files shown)


src/mo_frontend/typing.ml
```diff
@@ -3338,7 +3338,7 @@ and is_import d =
 
 and infer_dec_valdecs env dec : Scope.t =
```
crusso, line 3339: In that case, could this just return a val_env?

christoph-dfinity: You're right!

crusso (approved):
I agree this looks benign. But who knows for sure....

PR #5143, opened 2025-05-15
experiment: stable functions

crusso:
additional tests 

- stable-class-bug.mo: demonstrates unsoundness of promoting generic stable functions to generic local functions.
- stable-box-bug.mo: a simplified counter-example.

### DIFF
--- a/src/ir_def/rename.ml
+++ b/src/ir_def/rename.ml
@@ -12,6 +12,7 @@ let id rho i =
   with Not_found -> i
 
 let id_bind rho i =
+  if i.[0] = '@' then (i, rho) else
   let i' = fresh_id i in
   (i', Renaming.add i i' rho)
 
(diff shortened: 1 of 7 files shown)


crusso:
I added a renaming pass to ir_passes that renames all bound variables apart from @-prefixed ones that the code generator relies on. As expected, this appears to break eop, presumably because the free variable annotations on stable closures are not safe for renaming/alpha conversion. Broken, but probably fixable with some work.

The eop test that fail are

[nix-shell:~/motoko/test/run-drun]$ EXTRA_MOC_ARGS=--enhanced-orthogonal-persistence make
Some tests failed:
for-await.mo gc-random-test-force-gc.mo gc-random-test-no-force-gc.mo nested-generic-classes.mo nested-stable-functions.mo stable-box-bug.mo stable-captures-stable.mo stable-class-bug.mo stable-class.mo generic-stable-function.drun incompatible-stable-closure.drun life.drun life-mut.drun migrate-stable-function.drun reachable-stable-functions.drun

I'm pretty sure the tests all succeed when run without eop...

PR #5154, opened 2025-05-21
docs: 1-variable-declarations.md

jessiemongeon1:


doc/md/fundamentals/3-declarations/1-variable-declarations.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Immutable variables
+
+A `let` declaration names the value of an expression so that the expression can be evaluated once and its result, a value, referenced by that name. The compiler ensures that any attempt to modify an immutable variable results in an error.
+
+```motoko no-repl
+let x = 10;
+x := 20; // Error: Cannot assign to immutable variable
+```
+
```
crusso, line 21: It might be better do say let bindings support pattern matching here, with an example
````suggestion

The left hand side of a `let` can be also be a more general pattern, naming the components of a value by matching its structure:

For example, the declaration 

``` motoko
let (fst, snd) = (1, 2); 
```
uses the pattern `(fst, snd)` to name the components of the pair `(1,2)`. The value of `fst` is `1` and the value of `snd` is 2. Both `fst` and `snd` are immutable.
````

And then later say that `var` doesn't not support pattern matching.

File: doc/md/fundamentals/3-declarations/1-variable-declarations.md [hunk 2]
```diff
@@ -0,0 +1,51 @@
...
+```motoko no-repl
+var (a, b) = (1, 2); // Not supported
+```
+
+In contrast, `let` bindings do support pattern matching:
+
+```motoko no-repl
+let (a, b) = (1, 2); // Supported
+```
```
crusso:
```suggestion
In contrast, `let` bindings do support pattern matching:

```motoko no-repl
let (a, b) = (1, 2); // Supported
```
```
You can remove this if you apply the suggestion above.

File: doc/md/fundamentals/3-declarations/1-variable-declarations.md [hunk 3]
```diff
@@ -0,0 +1,44 @@
...
+
+## Mutable variables
+
+A `var` declaration in Motoko defines a mutable variable whose value can be updated after it is initially assigned. Unlike `let`, which creates an immutable binding, `var` allows reassignment using the `:=` operator. This means that the variable can hold different values over time, making it suitable for scenarios where state changes are required.
+
+```motoko no-repl
+var y = 10;
+y := 20; // Allowed, updates the value of y
+```
```
crusso, line 44: I would move the discussion on var patterns to the end:

```suggestion

## Mutable variables

A `var` declaration in Motoko defines a mutable variable whose value can be updated after it is initially assigned. 
Unlike `let`, which creates an immutable binding, `var` allows reassignment using the `:=` operator. This means that the variable can hold different values over time, making it suitable for scenarios where state changes are required.

```motoko no-repl
var y = 10;
y := 20; // Allowed, updates the value of y
```

Unlike, `let` declarations, `var` declarations do not support pattern matching. 
For example, the following is a syntax error:

```motoko no-repl
var (a, b) = (1, 2); // Not supported
```
```

File: doc/md/fundamentals/3-declarations/1-variable-declarations.md [hunk 4]
```diff
@@ -0,0 +1,44 @@
...
+
+In Motoko, variables are declared using:
+
+- `let` for immutable bindings: Once assigned, the value cannot be changed.
+
+- `var` for mutable bindings: The value can be updated later.
+
+## Immutable variables
+
+The `let` keyword is used to bind a name to the result of an expression, creating an immutable variable. When a let declaration is made, the expression on the right-hand side is evaluated once, and its resulting value is associated with the given name. This value cannot be changed after the initial assignment. The Motoko compiler enforces this immutability by generating a compile-time error if there is any attempt to modify the variable later in the code. This behavior helps ensure safety and predictability, especially in concurrent or distributed applications running on ICP.
```
crusso, line 15:
```suggestion
The `let` keyword is used to bind a name to the result of an expression, creating an immutable variable. In a `let` declaration, the expression on the right-hand side is evaluated once, and its resulting value is associated with the given name. This value cannot be changed after the initial assignment. The Motoko compiler enforces this immutability by generating a compile-time error if there is any attempt to modify the variable later in the code. 

In concurrent code, immutable declarations should be preferred over mutable declarations, since they prevent concurrent modification.
```

doc/md/fundamentals/3-declarations/1-variable-declarations.md
```diff
@@ -0,0 +1,49 @@
...
+
+The left hand side of a `let` can be also be a more general pattern, naming the components of a value by matching its structure:
+
+For example, the declaration:
+
+``` motoko
+let (fst, snd) = (1, 2); 
+```
+
+uses the pattern `(fst, snd)` to name the components of the pair `(1,2)`. The value of `fst` is `1` and the value of `snd` is 2. Both `fst` and `snd` are immutable.
```
crusso, line 32:
```suggestion
uses the pattern `(fst, snd)` to name the components of the pair `(1,2)`. The value of `fst` is `1` and the value of `snd` is `2`. Both `fst` and `snd` are immutable.
```

PR #5161, opened 2025-05-22
`pocket-ic` integration first steps

alexandru-uta:
This PR aims to build a local `test-runner` that will completely replace our dependency on `drun`, which will be decommissioned soon. 

This PR does the following:
* Adds nix flake for checking out and building from source `pocket-ic` both in CI and local development. NB: the nix flake patches pocket-ic to not emit backtraces such that our string matching test framework doesn't complain about it.
* Small rust project (`test-runner`) that depends on `pocket-ic` and its rust library crate.
* CI pipeline is slightly optimized (some tests were running unnecessarily in multiple derivations).
* The instruction performance counter is now coming from the canisters reporting (through [#5401](https://github.com/dfinity/motoko/pull/5401) instead of the Prometheus interface of drun); The `run.sh` script is modified accordingly.
* currently all tests run, with the exception of `upgrades.drun` (remains to be seen why).

NB: Developers need to be in a nix shell (`nix develop`) when running tests locally.

What needs to be done later is:
* figure out why `upgrades.drun` fails
* figure out a way to update pocket-ic related packages without too much manual intervention
* get rid of DRUN


christoph-dfinity (commented):
The Rust part looks great! I read through the `nix` and `shell` changes and nothing jumped out at me, but it would probably good if Gabor/Bas took a look as well.

Any chance we could upstream our patches to `pocket-ic`? (Behind a flag of course)

alexandru-uta:
> Any chance we could upstream our patches to `pocket-ic`? (Behind a flag of course)

Yup, I'm going to do that after consulting with @mraszyk .

nix/pocket-ic.nix
```diff
@@ -0,0 +1,91 @@
...
+      echo '+++ b/rs/pocket_ic_server/src/pocket_ic.rs' >> pocket_ic_server.patch
+      echo '@@ -512,6 +512,12 @@ impl PocketIcSubnets {' >> pocket_ic_server.patch
+      echo '             .embedders_config' >> pocket_ic_server.patch
+      echo '             .feature_flags' >> pocket_ic_server.patch
+      echo '             .rate_limiting_of_debug_prints = FlagStatus::Disabled;' >> pocket_ic_server.patch
+      echo '+' >> pocket_ic_server.patch
+      echo '+        hypervisor_config' >> pocket_ic_server.patch
+      echo '+            .embedders_config' >> pocket_ic_server.patch
+      echo '+            .feature_flags' >> pocket_ic_server.patch
+      echo '+            .canister_backtrace = FlagStatus::Disabled;' >> pocket_ic_server.patch
```
mraszyk, line 44: A pragmatic approach to include this upstream would be to place it [here](https://github.com/dfinity/ic/blob/12f93cd3f38337d6a4b98272d360990f808221cf/rs/pocket_ic_server/src/pocket_ic.rs#L550-L569) if your tests don't exercise instruction limit exceeded (since that config also raises the instruction limit to 100T instructions for the sake of canbench - this is the main use case enabled by that config).

nix/pocket-ic.nix
```diff
@@ -0,0 +1,91 @@
...
+      echo '+' >> pocket_ic_server.patch
+      echo '         let state_machine_config = StateMachineConfig::new(subnet_config, hypervisor_config);' >> pocket_ic_server.patch
+      echo '         let t = time' >> pocket_ic_server.patch
+      echo '             .duration_since(SystemTime::UNIX_EPOCH)' >> pocket_ic_server.patch
+
+      echo "Applying patch..."
+      patch -p1 < pocket_ic_server.patch
```
basvandijk, line 51: You could also consider saving this patch as a file so you get nice syntax highlighting and which would be easier to maintain. Then simply add the attribute:
```
patches = [ ./pocket_ic_server.patch ];
```
You then probably have to rename `patchPhase` to `postPatch`.

test-runner/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "test-runner"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+pocket-ic = { path = "/nix/store/wqdx4nfb1ja6nsar14cixyvcn35frxj1-source/packages/pocket-ic" }  # This will be replaced by the shell hook
```
basvandijk: Having `/nix/store` paths in source code is a bit of a code smell. Although atm I don't know exactly how to avoid it.

One idea I have is to fetch `pocket-ic` from a different registry like:

```
pocket-ic = { version = "9.0.2", registry = "ic-registry" }
```

and defining this registry with Nix and pointing to it by setting the following environment variable in the nix-shell: `CARGO_REGISTRIES_IC_REGISTRY_INDEX` (see [docs](https://doc.rust-lang.org/cargo/reference/registries.html#using-an-alternate-registry)).

But we can look into how to set this up in a follow up PR.

nix/shell.nix
```diff
@@ -76,6 +85,11 @@ pkgs.mkShell {
     # some cleanup of environment variables otherwise set by nix-shell
     # that would be confusing in interactive use
     unset XDG_DATA_DIRS
+
+    # Update Cargo.toml with the correct pocket-ic path if it exists
+    if [ -f "test-runner/Cargo.toml" ]; then
+      sed -i "s|pocket-ic = \".*\"|pocket-ic = { path = \"$POCKET_IC_LIBRARY\" }|" test-runner/Cargo.toml
+    fi
```
basvandijk: Like also [mentioned](https://github.com/dfinity/motoko/pull/5161/files#r2297490772) in the `test-runner/Cargo.toml`, it's a bit of a code smell to commit `/nix/store` paths to source code. But approving anyway since we can fix this later.

flake.nix
```diff
@@ -29,6 +29,10 @@
       url = "github:dfinity/ic";
       flake = false;
     };
+    pocket-ic-src = {
+      url = "github:dfinity/ic/master";
+      flake = false;
+    };
```
basvandijk, line 35: Why can't you take the `ic-src` for this? Or was that used for something else and you want to keep the `pocket-ic-src` separate from that?

alexandru-uta: thanks Bas for noticing it! I wanted to keep pocket-ic stuff separate from drun and initially it seemed to make sense, now perhaps not; let me think about it some more.

basvandijk: Makes sense. It's also easy to unify this later.

alexandru-uta: We'll also remove drun for good, so it's easy to kick it out then.

test-runner/Cargo.toml
alexandru-uta: shall we just for now replace that path with some %PLACEHOLDER% that will anyhow be replaced by the nix shell (and CI)?

basvandijk: That would be worse I think since that will cause a `git diff` every time you enter the nix shell.

alexandru-uta: talked to @nmattia and seems like we can simply rely on crates.io; we just need to be sure that the pocket-ic library from crates.io is compatible with the pocket-ic server built locally from source.

nix/shell.nix
alexandru-uta: fixed, see above

christoph-dfinity (approved):

PR #5163, opened 2025-05-22
docs: basic control flow

jessiemongeon1:


doc/md/fundamentals/4-control-flow/1-basic-control-flow.md
```diff
@@ -0,0 +1,219 @@
...
+let x : ?Nat = foo(); // x might be null or have a Nat value
+let y = ?Nat = foo(); // y might also be null or have a Nat value
+let z = do ? { x! + 3 / y! }; // If x or y is null, z is also null; otherwise, z = x + 3 / y
+```
+
+Instead of having to switch `x` and `y` in a verbose manner the use of the postfix operator `!` makes it simple to unwrap.
+
+## `label`
+
+A `label` assigns a name to a block of code that executes like any other block, but its result can be accessed directly. Labels provide more control over execution, allowing clear exit points and helping to structure complex logic effectively.
```
crusso, line 89:
```suggestion
A `label` assigns a name to a block of code that executes like any other block, but its result can be produced early using a `break` to the label.
If the block produces a non-`(0)` result, the `break` can include a value.
Labels provide more control over execution, allowing clear exit points and helping to structure complex logic effectively.
```

crusso (commented):
Interim comments. More to come. Do not merge yet.

doc/md/fundamentals/4-control-flow/1-basic-control-flow.md
```diff
@@ -0,0 +1,225 @@
...
+let y : ?Nat = foo(); // y might also be null or have a Nat value
+let z = do ? { x! + 3 / y! }; // If x or y is null, z is also null; otherwise, z = ?(x + 3 / y)
+```
+<!-- TODO(future): better, complete example. Perhaps there's one already the option section -->
+Instead of having to switch on the options `x` and `y` in a verbose manner the use of the postfix operator `!` makes it easy to unwrap their values but exit the block with `null` when either is `null`.
+
+## `label`
+
+A `label` assigns a name to a block of code that executes like any other block, but its result can be produced early using a `break` to the label.
+If the block produces a non-`(0)` result, the `break` can include a value.
```
crusso, line 94:
```suggestion
If the block produces a non-`()` result, the `break` can include a value.
```

doc/md/fundamentals/4-control-flow/1-basic-control-flow.md
```diff
@@ -0,0 +1,225 @@
...
+
+let numbers = [1, 2, 3, 4, 5];
+for (num in numbers.vals()) {
+    Debug.print(Nat.toText(num));
+}
+```
+
+## Program flows
+
+In Motoko, code executes sequentially, running one statement after another. However, certain constructs can alter this flow, such as exiting a block early, skipping iterations in a loop, returning a value from a function, or invoking another function.
```
crusso, line 167:
```suggestion
In Motoko, code executes sequentially, evaluating expressions and declarations in order. However, certain constructs can alter this flow, such as exiting a block early, skipping iterations in a loop, returning a value from a function, or invoking another function.
```

doc/md/fundamentals/4-control-flow/1-basic-control-flow.md
```diff
@@ -0,0 +1,225 @@
...
+
+### `continue`
+
+A `continue` statement skips the remainder of the current iteration in a loop and immediately proceeds to the next iteration. Like `break`, `continue` must reference a label and only works within a labeled loop.
```
crusso, line 171:
```suggestion
A `continue` expression skips the remainder of the current iteration in a loop and immediately proceeds to the next iteration. Like `break`, `continue` must reference a label and only works within a labeled `while`, `for` or `loop` expression.
```

doc/md/fundamentals/4-control-flow/1-basic-control-flow.md
```diff
@@ -0,0 +1,225 @@
...
+|--------------|---------------|
+| `return` | Exits a function and returns a value. |
+| `if` | Executes a block if the condition is `true`. |
+| `if/else` | Executes different blocks based on a condition. |
+| `switch` | [Pattern matching](https://internetcomputer.org/docs/motoko/fundamentals/pattern-matching) for variants, options, results, etc. |
+| `let-else` | Destructures a pattern and handle the failure case inline. |
+| `option block` | Evaluates an expression and wraps the result in an option type, allowing scoped handling of `null` values. |
+| `label/break` | Allows exiting loops early. |
+| `while` | Runs while a condition is `true`. |
+| `for` | Iterates over elements in a collection, terminating when no elements remain. |
```
crusso, line 17: Are try-catch and throw discussed elsewhere?

doc/md/fundamentals/4-control-flow/1-basic-control-flow.md
```diff
@@ -0,0 +1,225 @@
+---
+sidebar_position: 1
+---
+
```
crusso, line 4:
```suggestion
<!-- TODO(Future): the examples using  `sum` are ok, but artificial. Rewriting them to use  `product` would be more meaningful since you can exit early when a factor is 0 and skip the `*=` when factors are 1 -->
```

doc/md/fundamentals/4-control-flow/1-basic-control-flow.md
```diff
@@ -0,0 +1,225 @@
...
+var i = 0;
+while (i < 5) {
+  Debug.print(Nat.toText(i));
+  i += 1;
+}
+```
+
+## `for`
+
+A `for` loop iterates over elements in a collection or range, executing a block of code for each element.
```
crusso, line 153:
```suggestion
A `for` loop iterates over the elements of an iterator, executing a block of code for each element.
```

crusso (changes requested):
See suggestions

doc/md/fundamentals/4-control-flow/1-basic-control-flow.md
```diff
@@ -0,0 +1,280 @@
...
+    let numbers : [Int] = [3, 1, 5, -1, -2, 7];
+    return processNumbers(numbers); // Function call
+  };
+
+}
+```
+
+Execution begins in `functionCallControlFlow()`, where the function `processNumbers()` is invoked, transferring control to its logic. Inside `processNumbers()`, the numbers are processed one by one. If a negative number is encountered, a `return` statement immediately halts the function and returns the current sum. Control then flows back to `functionCallControlFlow()`, which receives the result and returns it.
+
+Function calls temporarily interrupt the normal sequential flow by shifting execution to a separate block of logic. Once the called function completes, control resumes at the point where the call was made.
```
crusso, line 279:
```suggestion
Function calls temporarily interrupt the normal sequential flow by shifting execution to a separate block of logic. Once the called function completes, control resumes at the point where the call was made, continuing with its result.
```

crusso (approved):
LGTM

(I replaced calculateZ with addOpt)

PR #5167, opened 2025-05-22
docs: control flow switch

jessiemongeon1:


doc/md/fundamentals/4-control-flow/5-switch.md
```diff
@@ -0,0 +1,31 @@
+---
+sidebar_position: 5
+hide_table_of_contents: true
+---
+
+# Switch
+
```
crusso, line 7:
```suggestion

<!-- TODO(future): add more examples, unless covered elsewhere -->

```

doc/md/fundamentals/4-control-flow/5-switch.md
```diff
@@ -0,0 +1,33 @@
...
+<!-- TODO(future): add more examples, unless covered elsewhere -->
+
+A `switch` expression is a selection control structure used for [pattern matching](https://internetcomputer.org/docs/motoko/fundamentals/pattern-matching). It evaluates a value and executes the code block corresponding to the first matching pattern.
```
crusso, line 10:
```suggestion
A `switch` expression is a selection control structure used for [pattern matching](https://internetcomputer.org/docs/motoko/fundamentals/pattern-matching). It evaluates a value and executes the branch, an expression or block, corresponding to the first case whose pattern matches the value. Any identifiers bound by the pattern are available in the branch, allowing you to access case specific data in the branch. 
```

doc/md/fundamentals/4-control-flow/5-switch.md
```diff
@@ -0,0 +1,33 @@
+---
+sidebar_position: 5
+hide_table_of_contents: true
+---
+
+# Switch
+
+<!-- TODO(future): add more examples, unless covered elsewhere -->
+
+A `switch` expression is a selection control structure used for [pattern matching](https://internetcomputer.org/docs/motoko/fundamentals/pattern-matching). It evaluates a value and executes the branch, an expression or block, corresponding to the first case whose pattern matches the value. Any identifiers bound by the pattern are available in the branch, allowing you to access case specific data in the branch.
```
crusso, line 10:
```suggestion

A `switch` expression is a control flow construct that, given a value, selects a control flow path based on the
pattern - or shape - of the value.

 A switch is constructed from an expression and a sequence  of cases. Each case consists of a [pattern](https://internetcomputer.org/docs/motoko/fundamentals/pattern-matching) guarding an expression or block that defines a possible branch of execution.

Switch evaluates its expression and based on its value, selects the first case whose pattern matches the value.
Any identifiers bound by the pattern are available in the selected branch, allowing you to access case specific data in the branch. 

Only one branch is ever executed in a switch, the first one that matches. If no case matches the value, execution traps.

Motoko will warn you if any of the cases are redundant and can never be executed. 
It will also warn you if the cases don't cover all possible values meaning  that the switch could fail with a trap at runtime.
```

doc/md/fundamentals/4-control-flow/5-switch.md
```diff
@@ -0,0 +1,44 @@
...
+Only one branch is ever executed in a switch, the first one that matches. If no case matches the value, execution traps.
+
+Motoko will warn you if any of the cases are redundant and can never matched due to previous cases. 
+
+Motoko will also warn you if the cases don't cover all possible values, meaning that the switch could fail with a trap at runtime.
+
+A `switch` is useful for replacing long `if-else` chains, improving both readability and structure. It lets you match specific values or patterns and handle each case individually.
+
+Only the first case of a `switch` expression that matches will execute. The wildcard pattern `_` matches any value and should be placed last, because if `_` appears earlier, it will prevent later cases from being checked. Once a match is found, evaluation continues with the code of the matching case.
+
```
crusso, line 26:
````suggestion

The simplest use of switch is emulate an `if-else` expression:

``` motoko
func toText(b : Bool) : Text {
   switch b {
      case true "true";
      case false "false";
   }
}
```

If we add second case for `true`, Motoko issues warning that it will never be matched and is unreachable or dead code:

``` motoko
func toText(b : Bool) : Text {
   switch b {
      case true "true";
      case false "false";
      case true "dead code";
   }
}
```

If we forget the case for `false`, Motoko will also issue a warning that `false is not covered by any case:

``` motoko
func toText(b : Bool) : Text {
   switch b {
      case true "true";
   }
}
```

Motoko is able to issue these warnings for much more complicated patterns, where it might be hard to see that you've made a mistake, helping you construct correct code and catch errors at compile-time.
````

crusso (approved):
I'll approve but please review and apply the suggestion @jessiemongeon1 (at your leisure of course)

PR #5170, opened 2025-05-22
docs: ICP features: caller ID

jessiemongeon1:


doc/md/icp-features/3-caller-identification.md
```diff
@@ -0,0 +1,388 @@
...
+}
+```
+
+### Role based access control
+
+**Role-based access control (RBAC)** assigns permissions based on roles rather than individual users. This approach enables structured and scalable authorization by grouping users into predefined roles, each with specific access levels and responsibilities.
+
+In this example, the canister assigns roles dynamically, allowing its controller to manage administrators and members, each with different permissions.
+
+```motoko no-repl
```
jessiemongeon1, line 97: This example does not compile, returns several errors

doc/md/icp-features/3-caller-identification.md
```diff
@@ -0,0 +1,388 @@
+---
+sidebar_position: 3
+---
+
+# Caller identification
```
jessiemongeon1, line 5: IMO the examples on this page should be wrapped in actors since they're rather long and showcase entire access control examples.

Demali-876: Agreed! Will do

doc/md/icp-features/3-caller-identification.md
Demali-876: I stand corrected. It seems you can deploy actor class on the playground now.

PR #5180, opened 2025-05-26
feat: Infer lambda body instead of defaulting to unit `()`

Kamirus:


src/mo_types/type.mli
```diff
@@ -197,7 +198,7 @@ val compare_field : field -> field -> int
 val set_kind : con -> kind -> unit
 
 module ConEnv : Env.S with type key = con
-module ConSet : Dom.S with type elt = con
+module ConSet = ConEnv.Dom
```
Kamirus, line 201: without it `ConSet` functions cannot be used on `ConEnv.dom`

test/fail/lambdas-core.mo
```diff
@@ -0,0 +1,488 @@
...
+let _ = Array.mapEntries(ar, func(x, i) = i * x);
+let _ = Array.flatMap(ar, func x = [x, -x]);
+let _ = Array.foldRight(ar, "", func(x, acc) = natToText(x) # acc);
+let _ = Array.all(ar, func x = x > 0);
+let _ = Array.any(ar, func x = x > 3);
+
+// VarArray module explicit type instantiation tests
+let va1 = VarArray.tabulate<Int>(4, func i = i * 2);
+let va2 : [var Int] = VarArray.tabulate(4, func i = i * 2);
+let _ = check(va1, va2);
```
Kamirus, line 309: Invariant; needs an annotation or explicit instantiation

src/mo_frontend/typing.ml
```diff
@@ -2183,31 +2188,149 @@ and infer_call env exp1 inst exp2 at t_expect_opt =
...
+          (* Future work: more cases *)
+          | _ -> Some (infer_exp env exp)
+        in
+        let infer_or_defer env (subs, deferred, to_fix) exp target_type =
+          match try_infer_exp env exp with
+          | Some t -> t, ((t, target_type) :: subs, deferred, to_fix) (* subtype problem for bi_match *)
+          | None -> target_type, (subs, (exp, target_type) :: deferred, to_fix) (* deferred *)
+        in
+        let rec decompose env exp target_type acc =
+          match exp.it, T.promote target_type with
```
crusso: promoting the target_type will replace a type parameter with its bound, are you sure this is what you want here?

crusso: Just because exp meets the bound of T doesn't mean exp has type T (is my concern).

Kamirus: I just need to unwrap the `Named`, I'll add a separate method that does only that

Kamirus: Good catch! Thanks. Done

src/mo_frontend/typing.ml
```diff
@@ -2204,31 +2209,153 @@ and infer_call env exp1 inst exp2 at t_expect_opt =
...
+          (* Future work: more cases *)
+          | _ -> Some (infer_exp env exp)
+        in
+        let infer_or_defer env (subs, deferred, to_fix) exp target_type =
+          match try_infer_exp env exp with
+          | Some t -> t, ((t, target_type) :: subs, deferred, to_fix) (* subtype problem for bi_match *)
+          | None -> target_type, (subs, (exp, target_type) :: deferred, to_fix) (* deferred *)
+        in
+        let rec decompose env exp target_type acc =
+          match exp.it, T.unwrap_named target_type with
```
```suggestion
          match exp.it, T.unwrap_named target_type with
```
Can T.unwrap_named not be T.normalize (which unfolds definitions but doesn't promote types to bounds)? This is the place you had T.promote before, right?

test/fail/lambdas-open-types-in-deferred.mo
```diff
@@ -0,0 +1,5 @@
+func foo<T, I>(t : T, f : I -> T) {};
+func useFoo() {
+  foo(1, func x = -1);
+  // `1` is used to infer `T`, however `I` cannot be inferred, error
+};
```
crusso:
```suggestion
};

```

src/mo_frontend/typing.ml
```diff
@@ -2137,6 +2112,36 @@ and check_exp_field env (ef : exp_field) fts =
   | None ->
     ignore (infer_exp env exp)
 
+and check_func_step ?(sub=sub) in_actor env (shared_pat, pat, typ_opt, exp) (s, c, ts1, ts2) : (env * T.typ) =
```
crusso: redefining sub at the call-sites feels a bit hacky? Is there a better way to factor this?

src/mo_frontend/typing.ml
```diff
@@ -2204,31 +2209,153 @@ and infer_call env exp1 inst exp2 at t_expect_opt =
...
+          | [], [] -> List.rev acc_ts, acc
+          | _ -> assert false
+        in
+        decompose env exp target_type ([], [], [])
+      in
+
+      (* Infer the argument as much as possible, defer sub-expressions that cannot be inferred *)
+      let (t2, (subs, deferred, to_fix)) = infer_subargs_for_bimatch_or_defer env exp2 t_arg in
+      if Bi_match.debug then begin
+        print_endline (Printf.sprintf "exp2 : %s" (Source.read_region_with_markers exp2.at |> Option.value ~default:""));
```
crusso: The debug spew is probably helpful, but can you abstract the code into functions to make the real logic easier to see?

src/mo_frontend/typing.ml
```diff
@@ -2204,31 +2209,153 @@ and infer_call env exp1 inst exp2 at t_expect_opt =
...
+            | FuncE (_, shared_pat, [], pat, typ_opt, _, body), T.Func (s, c, [], ts1, ts2) ->
+              (* Check that all type variables in the function input type are fixed, fail otherwise *)
+              Bi_match.fail_when_types_are_not_closed remaining ts1;
+              (* Check the function input type and prepare for inferring the body *)
+              let sub _ _ t t' =
+                (* Checking `typ_opt <: ts2` will fail when ts2 is not closed.
+                 * Add it to the subtype problems instead.
+                 *)
+                extra_subs := (t, t') :: !extra_subs;
+                true
```
crusso: This is a bit ugly - overrdiding the sub function. Maybe check_func_step could return the stuff to check and its done here and at the other call site?

src/mo_frontend/typing.ml
Kamirus: You are right. I'll use `normalize`. I guess without it some type annotations might not get unfolded

src/lang_utils/env.ml
```diff
@@ -25,6 +26,9 @@ struct
   exception Clash of key
 
   let dom env = List.fold_left (fun s (x, _) -> Dom.add x s) Dom.empty (bindings env)
+  let filterDom dom env = Dom.fold (fun k acc -> match find_opt k env with
+    | Some v -> add k v acc
+    | None -> acc) dom empty
```
christoph-dfinity:
```suggestion
  let filter_dom dom env = Dom.fold (fun k acc -> match find_opt k env with
    | Some v -> add k v acc
    | None -> acc) dom empty
```

Naming is `snake_case`.

I'm having a bit of a hard time to parse what this function is doing... What would be a good intuition?

src/mo_frontend/typing.ml
Kamirus: I left them because otherwise one would have to reintroduce them every time on debugging this flow.
It's a good idea to extract as much of this debug logic. I just did that, let me know if it's better now.
On call site I left it as `if debug then Debug.print_XYZ ...` which hopefully makes it obvious that the call is actually not executed usually.

src/lang_utils/env.ml
Kamirus: Good catch. I'm missing a linter that would catch these 😭 
`filter_dom` should perform a `filter` that returns only elements from the `dom` (domain).
So I'm getting an `env` with a filtered domain.
In key/value terms: it should filter these (key, value) pairs that appear in the keys called `dom`

Kamirus: Should be equivalent to `filter (fun k _ -> Dom.mem k dom) env`, but it iterates over `dom` instead of whole `env`

christoph-dfinity: Ahh I see. Maybe `intersect` would be a bit clearer than `filter` here?

src/mo_frontend/typing.ml
```diff
@@ -2204,31 +2216,144 @@ and infer_call env exp1 inst exp2 at t_expect_opt =
...
+        Partial Argument Inference:
+        We need to infer the type of the argument and find the best instantiation for the call expression.
+        However, some expressions cannot be inferred, e.g. unannotated lambdas like `func x = x + 1`.
+        Idea:
+        - Decompose the argument into sub-expressions and defer inference for those that would fail.
+        - Find a partial instantiation first using the inferred sub-expressions.
+        - Substitute and proceed with the remaining sub-expressions to get the full instantiation.
+       *)
+      let infer_subargs_for_bimatch_or_defer env exp target_type =
+        let rec cannot_infer_pat pat =
```
christoph-dfinity: Could we maybe use the `is_explicit_exp`/`is_explicit_pat` functions defined earlier in this file to do this classification?

src/mo_frontend/typing.ml
```diff
@@ -2204,31 +2216,144 @@ and infer_call env exp1 inst exp2 at t_expect_opt =
...
+          (* Future work: more cases *)
+          | _ -> infer_or_defer env acc exp target_type
+        and decompose_list env exps ts acc_ts acc =
+          match exps, ts with
+          | exp::exps, t::ts ->
+            let t', acc' = decompose env exp t acc in
+            decompose_list env exps ts (t' :: acc_ts) acc'
+          | [], [] -> List.rev acc_ts, acc
+          | _ -> assert false
+        in
```
christoph-dfinity: Can we write this as a `List.fold_left2`?

Kamirus: Is that better? Do you think maybe it's better to avoid pushing the `acc` parameter through all these functions here and have simply 3 `ref`s?

src/mo_frontend/typing.ml
Kamirus: Perfect suggestion! I was not aware of that function.
`is_explicit_exp` I cannot use because I just need to check the parameters in a func, not the return type annotation.

src/mo_frontend/typing.ml
Kamirus: I've tried it and the code looks clearer to me

src/mo_frontend/typing.ml
```diff
@@ -2204,31 +2216,131 @@ and infer_call env exp1 inst exp2 at t_expect_opt =
...
+          match exp.it, T.normalize target_type with
+          | TupE exps, T.Tup ts when List.length exps = List.length ts ->
+            let ts' = List.map2 decompose exps ts in
+            let target_type' = T.Tup ts' in
+            (* exp.note would need to be fixed later after the substitution *)
+            to_fix := (exp, target_type') :: !to_fix;
+            target_type'
+          (* Future work: more cases to decompose, e.g. T.Opt, T.Obj, T.Variant... *)
+          | _ -> infer_or_defer exp target_type
+        in
```
christoph-dfinity: I think this does look nicer with the refs... Can we simplify even further by collapsing the two functions into one? We should be able to just add the `FuncE` branch to the pattern match of `decompose` and bottom out in the `infer_exp` branch.

src/lang_utils/env.ml
Kamirus: Ok so you'd prefer `intersect_dom` instead of `filter_dom`, sounds good! Done

src/mo_frontend/bi_match.ml
```diff
@@ -244,94 +331,178 @@ let bi_match_subs scope_opt tbs subs typ_opt =
...
+
+(** Solves the given constraints [ts1, ts2] in the given context [ctx].
+    Unused type variables can be deferred to the next round.
+    [deferred_typs] are types to appear in the constraints of the next round. Used to determine which type variables to defer.
+ *)
+let solve ctx (ts1, ts2) deferred_typs =
+  if debug then Debug.print_solve ctx (ts1, ts2);
+
+  (* Defer solving type variables that can be solved later. More constraints appear in the next round, let them influence as many variables as possible *)
+  let to_defer, defer_verify = if deferred_typs = [] then (ConSet.empty, false) else
```
Kamirus, line 345: In the 1st round we should solve as few type variables as possible.
1. This avoids restricting the solution too early (more constraints are added in the next round)
2. and avoids inferring default bounds (like `Non`) for inputs to deferred functions (see [lambdas-open-types-in-deferred.mo](https://github.com/dfinity/motoko/pull/5180/files?diff=split&w=0#diff-828afbefe38428c6942a903c5467260bd51a28db56184c181f17417f7596281f)) that lead to confusing errors

src/mo_frontend/typing.ml
Kamirus: done, thanks! It turned out better than before!

src/mo_frontend/typing.ml
Kamirus: done!

src/mo_frontend/bi_match.ml
```diff
@@ -244,94 +332,173 @@ let bi_match_subs scope_opt tbs subs typ_opt =
...
+
+(** Solves the given constraints [ts1, ts2] in the given context [ctx].
+    Unused type variables can be deferred to the next round.
+    [deferred_typs] are types to appear in the constraints of the next round. Used to determine which type variables to defer.
+ *)
+let solve ctx (ts1, ts2) must_solve =
+  if debug then Debug.print_solve ctx (ts1, ts2) must_solve;
+
+  (* Defer solving type variables that can be solved later. More constraints appear in the next round, let them influence as many variables as possible *)
+  let to_defer, defer_verify = if must_solve = [] then (ConSet.empty, false) else
```
Kamirus, line 346: In the 1st round we should solve as few type variables as possible.
1. This avoids restricting the solution too early (more constraints are added in the next round)
2. and avoids inferring default bounds (like `Non`) for inputs to deferred functions (see [lambdas-open-types-in-deferred.mo](https://github.com/dfinity/motoko/pull/5180/files?diff=split&w=0#diff-828afbefe38428c6942a903c5467260bd51a28db56184c181f17417f7596281f)) that lead to confusing errors

Kamirus:
> Minor: odd error message:
> 
> ```
> > func app<T>(f: T -> T, x : T) : T { f(x)};
> let app : <T>(f : T -> T, x : T) -> T = <func>
> > let r : Int = app(func j = -1, 1);
> stdin:17.15-17.34: type error [M0098], cannot implicitly instantiate function of type
>   <T>(f : T -> T, x : T) -> T
> to argument of type
>   (f : Nat -> Nat, Nat)
> to produce result of type
>   Int
> because no instantiation of  makes
>   Int  <:  Nat
> > 
> ```
> 
> because no instantiation of ??? makes
> 
> Maybe this is because you applied the partial instantiation before reporting the error?

This is a good point! I've addressed that now, see this example in the tests

src/mo_frontend/typing.ml
```diff
@@ -2326,17 +2326,18 @@ and infer_call env exp1 inst exp2 at t_expect_opt =
...
-            if not closed && body_typ = codom then begin
-              (* When we just have [codom] and it is open : we need to infer the body *)
-              let actual_t = infer_exp env' body in
-              to_fix2 := (exp, T.Func (s, c, [], ts1, T.as_seq actual_t)) :: !to_fix2;
-              subs := (actual_t, body_typ) :: !subs;
-            end
+
+            (* When [codom] is open, we need to solve it *)
+            if not closed then
+              if body_typ <> codom then
```
crusso, line 2314: Lemme check this again

src/lang_utils/env.ml
```diff
@@ -7,7 +7,7 @@ sig
   exception Clash of key
 
   val dom : 'a t -> Dom.t
-  val filter_dom : Dom.t -> 'a t -> 'a t
+  val intersect_dom : Dom.t -> 'a t -> 'a t
```
crusso: I guess another name might be `restrict`

val restrict : 'a t  -> Dom.t -> 'a t

https://en.wikipedia.org/wiki/Restriction_(mathematics)

crusso: (no need to change, just being a geek)

crusso: (no need to change, just being a geek)

test/fail/lambdas-ok.mo
```diff
@@ -58,6 +58,25 @@ func foo<T, A>(t : T, _ : A, _ : A -> T) : T = t;
 // In the 2nd round: Int <: T  (solve T := Int)
 let _ = foo(1, "abc", func _ = -1);
 
+module ClosedBody1 {
+  func f1<A>(_ : A, _ : A -> Int) {};
+  func f2<A, B>(_ : A, _ : A -> Int) {};
+  // It should defer the func, solve A=Nat in the 1st round, but (1 : Int) should be checked and leave nothing to solve in the 2nd round
+  func _main() {
+    let _ = f1(1, func _ = 1);
+    let _ = f2(1, func _ = 1); // extra unused type variable B
```
crusso, line 67: These both work because in the first case it doesn't matter what we chose for B and the second B is determined by expected return type, right?

src/mo_frontend/bi_match.ml
```diff
@@ -341,38 +341,26 @@ let bi_match_typs ctx =
...
+    let to_solve = List.fold_left (fun acc t ->
       match normalize t with
-      | Func (_, _, _, t1, t2) ->
-        let inputs = ConSet.union inputs (cons_typs t1) in
-        let outputs = ConSet.union outputs (cons_typs t2) in
-        (inputs, outputs)
-      | _ ->
-        (ConSet.union inputs (cons t), outputs)
-      ) (ConSet.empty, ConSet.empty) deferred_typs
+      | Func (_, _, _, t1, _) -> ConSet.union acc (cons_typs t1)
```
crusso: Does it matter that t1 might contain Var's from the Func (it might have non-empty bounds). Does cons_typs expect to see a freestanding Var i ? Might be worth checking.

src/mo_frontend/typing.ml
```diff
@@ -2204,31 +2216,131 @@ and infer_call env exp1 inst exp2 at t_expect_opt =
...
+            Bi_match.fail_when_types_are_not_closed remaining ts1;
+            (* Check the function input type and prepare for inferring the body *)
+            let env', body_typ, codom = check_func_step false env (shared_pat, pat, typ_opt, body) (s, c, ts1, ts2) in
+            (* [codom] comes from [ts2] which might contain unsolved type variables. *)
+            let closed = Bi_match.is_closed remaining codom in
+            if not env.pre && (closed || body_typ <> codom) then
+              (* Closed [codom] implies closed [body_typ].
+               * [body_typ] is closed when it comes from [typ_opt] (which is when it is different from [codom])
+               * Since [body_typ] is closed, no need to infer.
+               *)
```
crusso, line 2306: Maybe its worth adding an assert that `is_closed remaining body_typ`

src/mo_frontend/typing.ml
```diff
@@ -2204,31 +2216,131 @@ and infer_call env exp1 inst exp2 at t_expect_opt =
...
+            (* When [codom] is open, we need to solve it *)
+            if not closed then
+              if body_typ <> codom then
+                (* [body_typ] is closed, body is already checked above, we just need to solve the subtype problem *)
+                subs := (body_typ, codom) :: !subs
+              else begin
+                (* We just have open [codom], we need to infer the body *)
+                let actual_t = infer_exp env' body in
+                to_fix2 := (exp, T.Func (s, c, [], ts1, T.as_seq actual_t)) :: !to_fix2;
+                subs := (actual_t, body_typ) :: !subs;
```
crusso, line 2321: This is also checking (actual_t, codom), since codom = body_typ?

crusso (approved):
Nice refactoring and improvements!

The complexity makes it hard to review with confidence, but I'd say let's merge and let it bake for a while. 

@christoph-dfinity is there any way we can test this on the mops suite?

test/fail/lambdas-ok.mo
Kamirus: - `ClosedBody1.f2` can infer `B` to `Non` just like before because it is totally unused
- `ClosedBody2.f2` infers `B` during the 2nd round:
  - but since we have the return type annotation in `func _ : Int = 1` we `check_exp` the body `1` instead of calling `infer_exp`
  - But we still need to verify that (Int <: B), we cannot call `sub` now, because there are open type variables. So we add this as a subtype problem and solve it in the 2nd round

Kamirus: Without the return type `: Int` we would:
- infer that `1 : Nat` and add `Nat <: B` so solve in the 2nd round

src/mo_frontend/typing.ml
Kamirus: Yes.
Well, this is unfortunately quite complex because we have `body_typ`, `codom` and `actual_t`.
`body_typ = codom` when `typ_opt = None` (when there is no return type annotation). In this case we need to infer the body (we cannot `check_exp` because of open types) and add `actual_t <: body_typ` to solve in the 2nd round (we cannot `sub` because of open types)

src/mo_frontend/bi_match.ml
Kamirus: It does not contain `Var`s, they are substituted with `open_` at the end of `bi_match_subs` function

test/fail/lambdas-ok.mo
crusso: Ok -  I no longer understand my original comment ;->

src/mo_frontend/bi_match.ml
crusso: I think what I was worried about was:

| Func (_, _, _tbs, t1, _) -> ConSet.union acc (cons_typs t1)

What if tbs <> [] (there are parameters) and t1 mentions one of them. But perhaps this case can't arise and tbs = [] is an invariant.

crusso: Or maybe you should strengthen the case pattern to
| Func (_, _, [], t1, _) -> ConSet.union acc (cons_typs t1)

crusso: Hmm, I think this code no longer exists anyway. Resolving.

Changelog.md
```diff
@@ -1,5 +1,39 @@
...
+       With this partial instantiation, it knows that `x : Nat`.
+    2. In the second phase, it completes the instantiation by inferring the bodies of the `func` arguments; assuming that all parameters were inferred in the first phase.
+       In the example above, it knows that `x : Nat`, so inferring the body `x * 2` will infer the type `O` to be `Nat`.
+       With this, the full instantiation `Array.map<Nat, Nat>` is inferred, and the type arguments can be omitted.
+
+    Limitations:
+    - Invariant type parameters must be explicitly provided in most cases.
+      e.g. `VarArray.map` must have the return type annotation:
+      ```motoko
+      let result : [var Int] = VarArray.map(varAr, func x = x * 2);
```
```suggestion
      let result = VarArray.map<Int, Int>(varAr, func x = x * 2);
```
Otherwise, in your example it is actually inferring from the context, so not provided!

crusso: You could show both variants (explicit arguments) or return type annotation.

crusso (approved):

Changelog.md
```diff
@@ -1,5 +1,44 @@
...
 
+* motoko (`moc`)
+
+  * Improved type inference for calling generic functions (#5180).
+    This means that type arguments can be omitted when calling generic functions in _most common cases_.
+    For example:
+
+    ```motoko
+    let ar = [1, 2, 3];
+    Array.map(ar, func x = x * 2);  // Needs no explicit type arguments!
```
P4P5T123: NIT: for clarity

```suggestion
    Array.map(ar, func x = x * 2);  // Needs no explicit type arguments anymore!
```

P4P5T123 (approved):

PR #5185, opened 2025-05-27
doc: rework section on subtyping

crusso:
@jessiemongeon1 I did a fairly extensive rewrite of subtyping. Hope it's an improvement.

All examples were checked in make preview, then changed to no-repl and _include.

### DIFF
--- a/doc/md/fundamentals/2-types/14-subtyping.md
+++ b/doc/md/fundamentals/2-types/14-subtyping.md
@@ -49,7 +49,7 @@ For example,
 * The mutable array type `[var T]` constructs a mutable array type from `T`.
 * The object type `{… f : T; …}` constructs an object type from a field type `T`.
 * The variant type `{… #f : T; …}` constructs a variant type from a field type `T`.
-* The function type `T1 -> T2` constructs a function type from the argument type `T1` and result type `T2`.
+* The function type `T -> U` constructs a function type from the argument type `T` and result type `U`.
 
 If replacing an argument to a type constructor by a supertype produces a supertype of the constructed type, then the constructor is
 _covariant_ in that argument.
@@ -64,13 +64,13 @@ Note that a type constructor that takes several arguments can have different var
 
 | Variance type | Description | Relationship | Motoko example |
 |---------------|-------------|--------------|----------------|
-| Covariant | Preserves the direction of subtyping. | If `T1 <: T2`, then `F[T1] <: F[T2]` | If `Nat <: Int`, then `?Nat <: ?Int`. You can use an optional natural number where an optional integer is expected. |
-| Contravariant | Reverses the direction of subtyping. | If `T1 <: T2`, then `F[T2] <: F[T1]` | If `Nat <: Int`, then `(Int -> ()) <: (Nat -> ())`. You can use a function that takes an integer as a function that takes a natural number |
-| Invariant | No subtyping relationship is preserved. | `F[T1]` <: `F[T2]` only when `T1 <: T2` and `T2 <: T1`  | Since `Nat  <: Int` but not `Int <: Nat`, `[var Nat]` and `[var Int]` are not related by subtyping. You cannot use a mutable array of natural numbers where a mutable array of integers is expected |
+| Covariant | Preserves the direction of subtyping. | If `T <: U`, then `F[T] <: F[U]` | If `Nat <: Int`, then `?Nat <: ?Int`. You can use an optional natural number where an optional integer is expected. |
+| Contravariant | Reverses the direction of subtyping. | If `T <: U`, then `F[U] <: F[T]` | If `Nat <: Int`, then `(Int -> ()) <: (Nat -> ())`. You can use a function that takes an integer as a function that takes a natural number |
+| Invariant | No subtyping relationship is preserved. | `F[T]` <: `F[U]` only when `T <: U` and `U <: T`  | Since `Nat  <: Int` but not `Int <: Nat`, `[var Nat]` and `[var Int]` are not related by subtyping. You cannot use a mutable array of natural numbers where a mutable array of integers is expected |
 
 In this table:
 
-- `T1` and `T2` are types.
+- `T` and `U` are types.
 
 - `F[T]` represents a type constructor applied to type `T` (e.g. `?T` for option type).
 
@@ -133,7 +133,7 @@ Every set `A` is a subset of the universal set, `A ⊆ U`.
 
 ## Options
 
-If `T1 <: T2`, then `?T1 <: ?T2` (option subtyping is covariant). This means an [optional value](https://internetcomputer.org/docs/motoko/fundamentals/types/options-results) of a subtype can be used as an optional value of a supertype.
+If `T <: U`, then `?T <: ?U` (option subtyping is covariant). This means an [optional value](https://internetcomputer.org/docs/motoko/fundamentals/types/options-results) of a subtype can be used as an optional value of a supertype.
 
 ```motoko norepl
 let a : ?Nat = ?5;
@@ -220,25 +220,26 @@ let rok : Result<Int, Text> = ok; // Allowed, since `{#ok : Nat} <: {#ok : Int; 
 
 ## Immutable arrays
 
-[Immutable arrays (`[T]`)](https://internetcomputer.org/docs/motoko/fundamentals/types/immutable-arrays) support covariant subtyping, meaning if `T1 <: T2`, then `[T1] <: [T2]`.
-If `T1 <: T2`, then an immutable array of type `[T1]` can be used as an immutable array of type `[T2]`.
+[Immutable arrays (`[T]`)](https://internetcomputer.org/docs/motoko/fundamentals/types/immutable-arrays) support covariant subtyping, meaning if `T <: U`, then `[T] <: [U]`.
+
+If `T <: U`, then an (immutable) array of type `[T]` can be used as an array of type `[U]`.
 
 ```motoko norepl
-let nums : [Nat] = [1, 2, 3];
-let numsAsInts : [Int] = nums;  // Allowed, since `Nat <: Int` we also have `[Nat] <: [Int]`.
+let nats : [Nat] = [1, 2, 3];
+let ints : [Int] = nats;  // Allowed, since `Nat <: Int` we also have `[Nat] <: [Int]`.
 ```
 
 ## Mutable arrays
 
 [Mutable arrays]](https://internetcomputer.org/docs/motoko/fundamentals/types/mutable-arrays) of the form  `[var T]` do not support interesting subtyping.
 The mutable array constructor `[var T]` is invariant in `T`.
-This means that `[var T1] <: [var T2]` only when `T1` and `T2` are equivalent.
+This means that `[var T] <: [var U]` only when `T` and `U` are equivalent.
 
-Allowing `[var T1] <: [var T2]` whenever `T1 <: T2`, without requiring `T2 <: T1` would not be safe.
+Allowing `[var T] <: [var U]` whenever `T <: U`, without requiring `U <: T` would not be safe.
 
 ```motoko norepl
 let nats : [var Nat] = [var 1, 2, 3];
-let ints : [var Int] = nats;  // Not allowed because `[var Nat] </: [var Int]`.
+let ints : [var Int] = nats;  // Not allowed, because `[var Nat] </: [var Int]`.
 ```
 
 If subtyping between mutable arrays were allowed, then the legal assignment
--- a/doc/md/fundamentals/2-types/14-subtyping.md
+++ b/doc/md/fundamentals/2-types/14-subtyping.md
@@ -85,7 +85,7 @@ let n : Nat = i;        // Not allowed, since `Int </: Nat`
 
 ## None: the least type
 
-In Motoko, `None` is an **empty type**. It contains no values at all. By definition, `None` is a subtype of every other type: for any type `T`,  there is `None <: T`.
+In Motoko, `None` is an **empty type**. It contains no values at all. By definition, `None` is a subtype of every other type: for any type `T`, `None <: T`.
 
 This makes `None` the least type in the subtype hierarchy.
 
@@ -107,7 +107,7 @@ Because `None` is the least type, it can be used wherever any other type is expe
 
 ## Any: the greatest type
 
-In Motoko, `Any` contains all possible values. By definition, every type is a subtype of `Any`: for any type `T`, there is `T <: Any`.
+In Motoko, `Any` contains all possible values. By definition, every type is a subtype of `Any`: for any type `T`, `T <: Any`.
 
 This makes `Any` the greatest type in the subtype hierarchy.
 
@@ -132,7 +132,7 @@ let a : ?Nat = ?5;
 let b : ?Int = a;     // Allowed, since `Nat <: Int` implies `?Nat <: ?Int`
 ```
 
-In Motoko, the literal `null` has the type `Null`, which is a subtype of any optional type. For any type `T`, there is `Null <: ?T`.
+In Motoko, the literal `null` has the type `Null`, which is a subtype of any optional type. For any type `T`, `Null <: ?T`.
 
 This means `null` can be used as the absent value for any optional type.
 
--- a/doc/md/fundamentals/2-types/14-subtyping.md
+++ b/doc/md/fundamentals/2-types/14-subtyping.md
@@ -262,7 +262,7 @@ would have the side-effect of also setting `nats[0]` to `-1`, but `nats` must on
 ## Functions
 
 [Functions](https://internetcomputer.org/docs/motoko/fundamentals/types/functions) also support subtyping.
-A function type `T1 -> T2` is a subtype of another function type `U1 -> U2` provided
+A function type `T1 -> T2` is a subtype of another function type `U1 -> U2` provided that:
 
 1. The argument types are related in the opposite direction (`U1 <: T1` ).
 2. The return types are related in the same direction (`T2 <: U2`).
--- a/doc/md/fundamentals/2-types/14-subtyping.md
+++ b/doc/md/fundamentals/2-types/14-subtyping.md
@@ -9,7 +9,7 @@ the first type is a subtype of the second.
 
 In Motoko, the simplest subtyping relation is between `Nat` and `Int`. `Nat` is a subtype of `Int`, written `Nat <: Int`.
 If you have a value of type `Nat`, you can supply it to a function taking the more general supertype `Int`.
-Motoko applies subtyping wherever it needs to, to prove that your program is type correct.
+Motoko applies subtyping when necessary to prove that a program is type correct.
 
 
 :::info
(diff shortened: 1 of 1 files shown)


doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
+impossible() # impossible();  // Allowed, since `None <: Text`
+impossible() + impossible();  // Allowed, since `None <: Nat`
+
+if (false == true) impossible(); // Allowed, since `None <: ()`
 ```
 
+Call the function `impossible()` will force your program to enter an infinite loop, returning no value at all.
+Because `None` is the least type, it can be used wherever any other type is expected.
+
+`None` is similar to the empty set: the empty set has no elements and is a subset of every other set `{} ⊆ A`.
```
jessiemongeon1, line 112:
```suggestion
Calling the function `impossible()` will force the program to enter an infinite loop and return no value.
Because `None` is the least type, it can be used wherever any other type is expected.

`None` is similar to an empty set, as it has no elements and is a subset of every other set `{} ⊆ A`.
```

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
+
+## Any: the greatest type
+
+Motoko's `Any` type contains all values. By definition, every type is a subset of `Any`, `T <: Any`.
+`Any` is the greatest (as in largest) type in Motoko.
+
+A function that accepts an `Any` argument can be applied to "any" type of argument.
```
jessiemongeon1, line 119:
```suggestion
In Motoko, `Any` contains all possible values. By definition, every type is a subtype of `Any`: for any type `T`, there is `T <: Any`.

This makes `Any` the greatest type in the subtype hierarchy.

A function that takes an argument of type `Any` can be applied to a value of any type, since all types are compatible with `Any`.
```

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
+``` motoko no-repl
+func discard(a : Any) {};
+
+discard(0); // Allowed, since `Nat <: Any`
+discard(true); // Allowed, since `Bool <: Any`
+discard("abc"); // Allowed, since `Text <: Any`
+```
+
+`Any` is similar to the universal set `U`, the set that contains all possible elements.
+Every set `A` is a subset of the universal set, `A ⊆ U`.
```
jessiemongeon1, line 130:
```suggestion
`Any` is similar to the universal set `U` that contains all possible elements. Every set `A` is a subset of the universal set `A ⊆ U`.
```

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
+
 ## Options
 
-If `T1 <: T2`, then `?T1 <: ?T2` (option subtyping is covariant). This means an [optional value](https://internetcomputer.org/docs/motoko/fundamentals/types/options-results) of a subtype can be used as an optional value of a supertype.
+If `T <: U`, then `?T <: ?U` (option subtyping is covariant). This means an [optional value](https://internetcomputer.org/docs/motoko/fundamentals/types/options-results) of a subtype can be used as an optional value of a supertype.
```
jessiemongeon1, line 134:
```suggestion
If `T <: U`, then `?T <: ?U` because option subtyping is covariant. This means an [optional value](https://internetcomputer.org/docs/motoko/fundamentals/types/options-results) of a subtype can be used as an optional value of a supertype.
```

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
+
+```motoko no-repl
+let a : ?Nat = ?5;
+let b : ?Int = a;     // Allowed, since `Nat <: Int` implies `?Nat <: ?Int`
+```
+
+In Motoko, the literal `null` has type `Null` which subtypes any option type, that is `Null <: ?T` (for any `T`).
+This means you can use `null` as the absent value of any optional type.
```
jessiemongeon1, line 142:
```suggestion
In Motoko, the literal `null` has the type `Null`, which is a subtype of any optional type. For any type `T`, there is `Null <: ?T`.

This means `null` can be used as the absent value for any optional type.
```

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
+let n : Null = null;
+let oi : ?Int = n;      // Allowed, since `Null <: ?Int`
+let ot : ?Text = n;     // Allowed, since `Null <: ?Text`
 ```
 
-## Records
+## Records and objects
 
-[Records](https://internetcomputer.org/docs/motoko/fundamentals/types/records) support structural subtyping, meaning a record with more fields can be used where a record with fewer fields is expected. A record type is a subtype of another if it has at least the same fields with matching types.
+[Records](https://internetcomputer.org/docs/motoko/fundamentals/types/records) and, more generally,  [objects](https://internetcomputer.org/docs/motoko/fundamentals/types/objects-classes)
```
jessiemongeon1, line 152:
```suggestion
[Records](https://internetcomputer.org/docs/motoko/fundamentals/types/records) and, more generally, [objects](https://internetcomputer.org/docs/motoko/fundamentals/types/objects-classes)
```

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
+let b : B = { name = "Other"; age = -1 };
+let c : C = { name = "Motoko" };
+
+let b1 : B = a; // Allowed since `age : Nat` implies `age : Int` (`Nat <: Int`).
+let c1 : C = b; // Allowed, since `B` has all fields of `C`
+let a1 : A = b; // Not allowed, since `age : Int` does not imply `age : Nat` (`Int </: Nat`).
+let b2 : B = c; // Not allowed, because `C` is missing field `age`.
+```
+
+If the field of an object is mutable (e.g. `var age : Nat`), then any field in the supertype must also be mutable, with equivalent content.
```
jessiemongeon1, line 174:
```suggestion
If the field of an object is mutable (e.g., `var age : Nat`), then any field in the supertype must also be mutable with equivalent content.
```

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
 
-let a : A = { name = "Motoko" };
-let b : B = { name = "Ghost"; age = 25 };
+For example, consider these object types with mutable `age` fields:
```
jessiemongeon1, line 176:
```suggestion
For example, consider the following object types with mutable `age` fields:
```

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
 
-let a2 : A = b;  // Allowed, since B has all fields of A
-let b2 : B = a; // Not allowed because A lacks age.
+```motoko no-repl
+type A = { name : Text; var age : Nat };
+type B = { name : Text; var age : Int };
 ```
 
+Now neither `A <: B` nor `B <: A` since the contents of the mutable `age` field, `Nat` and `Int`, are not equivalent.
+However, we still have `A <: C` and `B <: C` (dropping the `age` field entirely).
```
jessiemongeon1, line 184:
```suggestion
Now, neither `A <: B` nor `B <: A` holds, because the `age` fields differ in type because `Nat` and `Int` are not equivalent.

However, both `A` and `B` are still subtypes of `C`, since `C` lacks the `age` field entirely and subtyping allows dropping fields.
```

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
-let nums : [Nat] = [1, 2, 3];
-let numsAsInts : [Int] = nums;  // Allowed, since `Nat <: Int`
+let nats : [Nat] = [1, 2, 3];
+let ints : [Int] = nats;  // Allowed, since `Nat <: Int` we also have `[Nat] <: [Int]`.
 ```
 
 ## Mutable arrays
 
-[Mutable arrays (`var [T]`)](https://internetcomputer.org/docs/motoko/fundamentals/types/mutable-arrays) do not support subtyping due to type safety risks. Allowing `var [T1] <: var [T2]` could lead to runtime type mismatches.
+[Mutable arrays]](https://internetcomputer.org/docs/motoko/fundamentals/types/mutable-arrays) of the form  `[var T]` do not support interesting subtyping.
```
jessiemongeon1, line 245:
```suggestion
[Mutable arrays](https://internetcomputer.org/docs/motoko/fundamentals/types/mutable-arrays) of the form `[var T]` do not support interesting subtyping.
```

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
+The mutable array constructor `[var T]` is invariant in `T`.
+This means that `[var T] <: [var U]` only when `T` and `U` are equivalent.
+
+Allowing `[var T] <: [var U]` whenever `T <: U`, without requiring `U <: T` would not be safe.
```
jessiemongeon1, line 249:
```suggestion
The mutable array constructor `[var T]` is invariant in `T`.

This means that `[var T]` is a subtype of `[var U]` only if*`T` and `U` are equivalent types, that is, both `T <: U` and `U <: T` must hold.

Allowing `[var T] <: [var U]` whenever `T <: U` (without requiring `U <: T`) would be unsafe, as it could allow writing values of type `U` into an array that only expects values of type `T`.
```

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
+let ints : [var Int] = nats;  // Not allowed, because `[var Nat] </: [var Int]`.
+```
+
+If subtyping between mutable arrays were allowed, then the legal assignment
+
+``` motoko no-repl
+ints[0] := -1;
 ```
 
+would have the side-effect of also setting `nats[0]` to `-1`, but `nats` must only contain `Nat`s!
```
jessiemongeon1, line 262:
```suggestion

```

jessiemongeon1: Since this is not allowed, this is not necessary to call out.

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
+
+
 ## Functions
 
-[Function](https://internetcomputer.org/docs/motoko/fundamentals/types/functions) subtyping follows two rules:
+[Function](https://internetcomputer.org/docs/motoko/fundamentals/types/functions) also support subtyping.
```
jessiemongeon1, line 267:
```suggestion
[Functions](https://internetcomputer.org/docs/motoko/fundamentals/types/functions) also support subtyping.
```

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
+A function type `T1 -> T2` is a subtype of another function type `U1 -> U2` provided
```
jessiemongeon1, line 268:
```suggestion
A function type `T1 -> T2` is a subtype of another function type `U1 -> U2` provided that:
```

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
 
-1. Parameter types must be contravariant: `T2 <: T1` means `T1 -> R <: T2 -> R`
-2. Return types must be covariant: `R1 <: R2` means `T -> R1 <: T -> R2`
+1. `U1 <: T1` : the argument types are related in the opposite direction.
+2. `T2 <: U2` :  the return types are related in the same direction.
```
jessiemongeon1, line 271:
```suggestion
1. The argument types are related in the opposite direction (`U1 <: T1` ).
2. The return types are related in the same direction (`T2 <: U2`).
```

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
-func parent(x : Int) : Nat { 10 };
-func child(x : Nat) : Int { 20 };
+This means that function subtyping is contravariant in the arguments of the function and covariant in the results.
+
+This might seems confusing at first, but the intuition is quite simple.
+A function can be used at a supertype provided that type requires a more specific argument and produces a more general result.
+This ensures that the orginal function, when used at the supertype,
+will only receive arguments it can consume (since `U1 <: T1`) and produce results than are expected (since `T2 <: U2`).
+
+As a simple example, consider the `magnitude` function that returns the absolute value, or magnitude, of an integer.
```
jessiemongeon1, line 280:
```suggestion
Function subtyping is contravariant in the argument types and covariant in the return type.

A function of type `T1 → T2` can be used where a function of type `U1 → U2` is expected only if `U1 <: T1` (the expected argument is more specific) and `T2 <: U2` (the actual result is more generic).

This ensures safety because the function will never be passed arguments it can’t handle, and it will always produce results that meet the expectations of the surrounding context.

As a simple example, consider the `magnitude` function that returns the absolute value, or magnitude, of an integer.
```

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
+let i2i : Int-> Int = magnitude;
+let n2i : Nat-> Int = magnitude;
+```
 
-```motoko no-repl
-type Parent = { getName : () -> Text };
-type Child = { getName : () -> Text; getAge : () -> Nat };
+You may be wondering what happens if the one or both of the function types is generic and has type parameters.
+Motoko uses the simple rule that two function types with type parameters are subtypes only when both types have the same parameters
+(up to renaming), and the arguments and results are related according to 1. and 2. above.
```
jessiemongeon1, line 299:
```suggestion
You might wonder what happens when one or both of the function types are generic and have type parameters.

Motoko applies a straightforward rule: two generic function types are considered subtypes only if they have the same type parameters (up to renaming), and their parameter and result types follow the usual subtyping rules.
```

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
 
-let obj : Child = { getName = func() { "Motoko" }; getAge = func() { 25 } };
-let p : Parent = obj;  // Allowed, since `Child` has all methods of `Parent`
+## Modules and actors
+
+Modules and actors support subtyping just like objects, meaning a module(actor) is a subtype of a similar module(actor) with some fields removed,
+provided the types of the fields in common are related by subtyping.
+
+For actors, which can only contain shared functions as fields, this means that you can remove some functions or replace the types of common
+functions with supertypes.
```
jessiemongeon1, line 307:
```suggestion
Modules and actors in Motoko support subtyping in the same way as objects.

Specifically, a module (or actor) is a subtype of another module (or actor) with some fields removed, as long as the remaining shared fields have types that are related by subtyping.

In the case of actors, which can only contain shared functions as fields, this means:

* You can remove some functions.
* You can replace the types of shared functions with supertypes (according to function subtyping rules).
```

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
+
+## Recursive and generic types
+
+[Recursive and generic types](https://internetcomputer.org/docs/motoko/fundamentals/types/advanced-types) can be subtypes of each other when their definitions allow it.
+
+Consider these two recursive point types, the second one adding a `color` field:
```
jessiemongeon1, line 313:
```suggestion
Consider the following recursive point types, where the second extends the first by adding a color field:
```

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,356 @@ sidebar_position: 14
...
+  x = 0;
+  move = func (dx : Int) { { cp with x = cp.x + dx } }
+};
+let c : Point = cp;
+```
 
-```motoko no-repl
-type ListA = ?{ head : Nat; tail : ListA };
-type ListB = ?{ head : Nat; tail : ListB };
+This also works for recursive variants and even generic types:
```
jessiemongeon1, line 339:
```suggestion
This also works for recursive variants and even generic types.
```

doc/md/fundamentals/2-types/14-subtyping.md
crusso: (Reapplied manually - somehow your suggestion was lost).

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,347 @@ sidebar_position: 14
...
 
-```motoko no-repl
-func parent(x : Int) : Nat { 10 };
-func child(x : Nat) : Int { 20 };
+This means that function subtyping is contravariant in the arguments of the function and covariant in the results.
+
+This might seems confusing at first, but the intuition is quite simple.
+A function can be used at a supertype provided that type requires a more specific argument and produces a more general result.
+This ensures that the orginal function, when used at the supertype,
+will only receive arguments it can consume (since `U1 <: T1`) and produce results than are expected (since `T2 <: U2`).
```
crusso, line 266:
```suggestion
Function subtyping is contravariant in the argument types and covariant in the return type.

A function of type `T1 → T2` can be used where a function of type `U1 → U2` is expected only if `U1 <: T1` (the expected argument is more specific) and `T2 <: U2` (the actual result is more generic).

This ensures safety because the function will never be passed arguments it can’t handle, and it will always produce results that meet the expectations of the surrounding context.
```

doc/md/fundamentals/2-types/14-subtyping.md
crusso: (this suggestion was lost, I've reapplied manually)

doc/md/fundamentals/2-types/14-subtyping.md
```diff
@@ -4,121 +4,346 @@ sidebar_position: 14
...
 
-If `T1 <: T2`, then `?T1 <: ?T2` (option subtyping is covariant). This means an [optional value](https://internetcomputer.org/docs/motoko/fundamentals/types/options-results) of a subtype can be used as an optional value of a supertype.
+If `T <: U`, then `?T <: ?U` because option subtyping is covariant. This means an [optional value](https://internetcomputer.org/docs/motoko/fundamentals/types/options-results) of a subtype can be used as an optional value of a supertype.
+
+```motoko no-repl
+let a : ?Nat = ?5;
+let b : ?Int = a;     // Allowed, since `Nat <: Int` implies `?Nat <: ?Int`
+```
+
+In Motoko, the literal `null` has the type `Null`, which is a subtype of any optional type. For any type `T`, there is `Null <: ?T`.
```
crusso, line 135:
```suggestion
In Motoko, the literal `null` has the type `Null`, which is a subtype of any optional type. For any type `T`, `Null <: ?T`.
```

PR #5196, opened 2025-05-28
docs: Async data

jessiemongeon1:


doc/md/5-actors-async.md
```diff
@@ -0,0 +1,316 @@
...
+
+## Traps and commit points
+
+A trap is a non-recoverable runtime failure in Motoko, caused by errors such as:
+
+* Division by zero.
+* Out-of-bounds array access.
+* Numeric overflow.
+* Exceeding cycle limits.
+* Failing an assertion.
```
crusso, line 50:
```suggestion
* Failing an assertion.
* An explicit call to `Debug.trap()`. <!-- FUTURE: `Runtime.trap() in new base -->
```

doc/md/5-actors-async.md
```diff
@@ -0,0 +1,316 @@
...
+
+When a shared function executes without encountering an `await` expression, it does not suspend and runs atomically, meaning its execution cannot be interleaved or interrupted. Such functions are considered syntactically atomic because their structure guarantees uninterrupted execution from start to finish.
```
crusso:
```suggestion
When a shared function executes without evaluating an `await` expression, it never suspend and thus runs atomically, meaning its execution cannot be interleaved with that of another message. Functions that don't contain any `await` expressions are syntactically atomic and guarantee interference-free execution from start to finish.
```

doc/md/5-actors-async.md
```diff
@@ -0,0 +1,316 @@
...
+
+-   Function `read()` asynchronously reads the counter value and returns a future of type `async Nat` containing its value.
+
+-   Function `bump()` asynchronously increments and reads the counter.
+
+The only way to read or modify the state (`count`) of the `Counter` actor is through its shared functions.
+
+## Using `await` to consume async futures
+
+The caller of a shared function typically receives a future, a value of type `async T` for some T.
```
crusso:
```suggestion
The caller of a shared function typically receives a future, a value of type `async T` for some `T`.
```

doc/md/5-actors-async.md
```diff
@@ -0,0 +1,316 @@
...
+
+Typically, one rolls the two steps into one and just awaits an asynchronous call directly:
+
+``` motoko no-repl
+let n : Nat = await Counter.read();
+```
+
+Unlike a local function call, which waits for the result before continuing, a shared function call returns a future immediately without blocking. Later, calling `await` on that future pauses the current task until the future is finished. When the future completes, `await` either returns the result or throws an error if the future ended with one.
+
+If you `await` the same future again, it just returns the same result or error. Even if the future is already done, `await` still pauses briefly to make sure all previous state changes and messages are saved before continuing.
```
crusso:
```suggestion
If you `await` the same future again, it just returns the same result or error. Even if the future is already done, `await` will briefly suspend all pending state changes and outgoing messages. This means that you can rely on every `await` to commit state, whether its future is still in progress or already completed.
```

doc/md/5-actors-async.md
```diff
@@ -0,0 +1,316 @@
...
+In other languages without these features, developers often need to use advanced patterns like callbacks and event handlers. This low-level, systems-style programming can be powerful but is error-prone, as it breaks high-level logic into scattered events and shared state.
+
+### Example
+
+To demonstrate how asynchronous actors work, consider the following example.
+
+Customers place orders at a pizza restaurant, but the chef can only make one pizza at a time. Orders are taken **[asynchronously](https://internetcomputer.org/docs/motoko/fundamentals/actors-async#async--await)**, meaning customers do not have to wait for previous orders to be completed before placing their own. However, each pizza is prepared sequentially. This is representative of an asynchronous actor.
+
+```motoko no-repl
+import Array "mo:base/Array";
```
crusso, line 202: It would be better use a Deque (Queue) pushing new order to the ends and popping the next order to make from the front.

doc/md/5-actors-async.md
```diff
@@ -0,0 +1,316 @@
...
+Even with just a single async call, Motoko’s abstractions help keep the code clear. By using `await`, the programmer tells the compiler where interleaving can happen, avoiding the need to restructure logic to fit the system’s message-passing loop.
+
+In other languages without these features, developers often need to use advanced patterns like callbacks and event handlers. This low-level, systems-style programming can be powerful but is error-prone, as it breaks high-level logic into scattered events and shared state.
+
+### Example
+
+To demonstrate how asynchronous actors work, consider the following example.
+
+Customers place orders at a pizza restaurant, but the chef can only make one pizza at a time. Orders are taken **[asynchronously](https://internetcomputer.org/docs/motoko/fundamentals/actors-async#async--await)**, meaning customers do not have to wait for previous orders to be completed before placing their own. However, each pizza is prepared sequentially. This is representative of an asynchronous actor.
+
```
crusso:
```suggestion
<!-- TODO(FUTURE): It would be cleaner to use a Deque or Queue pushing new order to the end and popping the next order to make from the front. -->
```

doc/md/5-actors-async.md
```diff
@@ -0,0 +1,316 @@
...
+        let estimatedTime = 15 + (orders.size() * 5); // Base time + 5 mins per pending order
+        return "Your " # order # " will arrive in " # Nat.toText(estimatedTime) # " minutes.";
+    };
+
+    // Multiple users can check their delivery status asynchronously
+    public shared func getDeliveryStatus(order : Text) : async Text {
+        let status = await* checkDelivery(order);
+        return status;
+    };
+```
```
crusso:
````suggestion
<!--
### Example

Customers can place multiple orders at the same time. However, when they ask for an update on their delivery status, responses must be generated sequentially based on the preparation time. Instead of constantly tracking delivery updates, `async*` can be used to calculate the estimated delivery time only when requested.

```motoko no-repl
// async* function to calculate delivery time on demand
func checkDelivery(order : Text) : async* Text {
        let estimatedTime = 15 + (orders.size() * 5); // Base time + 5 mins per pending order
        return "Your " # order # " will arrive in " # Nat.toText(estimatedTime) # " minutes.";
    };

    // Multiple users can check their delivery status asynchronously
    public shared func getDeliveryStatus(order : Text) : async Text {
        let status = await* checkDelivery(order);
        return status;
    };
```
-->
````

This is *not* a good example of using async*/await* since the checkDelivery function could just as well be written as a synchronous, local function - it doesn't do anything async in its body.
I'm pretty sure the original doc had a better example of this (compare with) https://internetcomputer.org/docs/motoko/main/writing-motoko/async-star)

Either delete or comment out for now.

Do we still have https://internetcomputer.org/docs/motoko/main/writing-motoko/async-star in this version of the doc? If not, we should restore it.

File: doc/md/5-actors-async.md [hunk 8]
```diff
@@ -0,0 +1,316 @@
...
+
+## `try/finally`
```
crusso, line 288: Are try-catch and try-catch-finally explained anywhere? There's used to be section on this in error handling but it seems to have disappeared.

jessiemongeon1: It does appear to have vanished; I'll add it back here.

doc/md/5-actors-async.md
jessiemongeon1: Yes the async* info is in the section above (copied from that page and revised editorially, will use the old example here instead).

PR #5215, opened 2025-06-04
feat: introduce `await?` to short-circuit fulfilled `async` futures

### LINKED ISSUE #5212: Order of arrival of message responses can delay execution
@timohanke provided this feedback
``` Motoko
let fut1 = Actor1.call(); // the slow (e.g. inter-subnet) call
let fut2 = Actor2.call(); // the fast one

await fut1;
// state changes
await fut2;
```
in this scenario if the `fut2` response arrives before `fut1`'s, it gets saved while `await fut1` is still blocking. When we come to `await fut2` the continuation is not invoked immediately, because of the mandatory commit semantics enforcing a context switch.

Timo proposes an opt-in mechanism to avoid the commit and immediately go on executing the continuation, thus avoiding the delay.

I propose a parenthetical like
``` Motoko
(with commit = false) await fut2;
```
to opt out of `await`'s mandatory blocking.

The other possibility is a parallel `await`
``` Motoko
let vals = [await] [fut1, fut2];
```

@Kamirus suggests
``` Motoko
let val = switch (fulfilled fut2) {
  case (?val) val;
  case null await fut2
}
```

ggreif:
Fixes #5212.

Previously one had to use `await` (without the `*`) to force an `async` future to a value. The extended semantics implemented for `await?` will opt out of the the mandatory state commit coming with `await` and return the value from the future instantly if already fulfilled.

The intuition for the syntax is that this causes 0 or 1 commit points (the former when the future is already fulfilled in an earlier `await` or `await?`).

FYI: @mraszyk

TODO:
- [x] IR-interpreter
- [x] docs

### DIFF
--- a/test/run-drun/blockless.mo
+++ b/test/run-drun/blockless.mo
@@ -1,4 +1,4 @@
-import { debugPrint } = "mo:⛔";
+import { debugPrint; error } = "mo:⛔";
 
 actor A {
     public func foo() : async Int { 42 };
@@ -33,12 +33,32 @@ actor A {
       assert (if shortCircuit not changed else changed);
     };
 
+    public func throwing(shortCircuit : Bool) : async () {
+      debugPrint ("throw test: " # debug_show shortCircuit);
+      var changed = false;
+      try await async { 
+        let a = async { throw error "Bummer!"; /* issue */ 4578 };
+        try ignore await a catch _ debugPrint "caught it!";
+        changed := true;
+        try {
+          ignore if shortCircuit
+            await? a // fast await
+          else
+            await a; // proper await
+        } catch _ debugPrint "caught it again!";
+        assert false;
+      } catch _ {};
+      assert (if shortCircuit not changed else changed);
+    };
+
     public func go() : async () {
       ignore await? A.foo();
       await A.queue true;
       await A.queue false;
       await A.state false;
       await A.state true;
+      await A.throwing false;
+      await A.throwing true;
     }
 };
 
--- a/test/run-drun/ok/blockless.drun-run.ok
+++ b/test/run-drun/ok/blockless.drun-run.ok
@@ -5,4 +5,10 @@ debug.print: queue test: false
 debug.print: Peek-a-boo!
 debug.print: state test: false
 debug.print: state test: true
+debug.print: throw test: false
+debug.print: caught it!
+debug.print: caught it again!
+debug.print: throw test: true
+debug.print: caught it!
+debug.print: caught it again!
 ingress Completed: Reply: 0x4449444c0000


src/ir_def/arrange_ir.ml
```diff
@@ -108,8 +108,8 @@ and prim = function
   | SetCertifiedData  -> Atom "SetCertifiedData"
   | GetCertificate    -> Atom "GetCertificate"
   | OtherPrim s       -> Atom s
-  | CPSAwait (Type.Fut, t) -> "CPSAwait" $$ [typ t]
-  | CPSAwait (Type.Cmp, t) -> "CPSAwait*" $$ [typ t]
+  | CPSAwait (AwaitFut _, t) -> "CPSAwait" $$ [typ t]
```
ggreif: fix!

src/ir_def/ir.ml
```diff
@@ -186,6 +186,8 @@ and prim =
   | ICStableRead of Type.typ           (* deserialize value of stable type from stable memory *)
   | ICStableSize of Type.typ
 
+and await_sort = AwaitFut of bool | AwaitCmp
```
ggreif: comment?

src/ir_passes/async.ml
```diff
@@ -49,7 +49,8 @@ let t_async_fut as_seq t =
   Func (Local, Returns, [], [fulfillT as_seq t; failT; bailT],
         [sum [
              ("suspend", unit);
-             ("schedule", Func(Local, Returns, [], [], []))]])
+             ("schedule", Func(Local, Returns, [], [], []));
+             ("resume", Func(Local, Returns, [], [], []))]])
```
ggreif: remove

test/run-drun/blockless.mo
```diff
@@ -0,0 +1,6 @@
+actor A {
+    public func foo() : async Int { 42 };
+    public func go() : async Int { await* A.foo() }
```
ggreif: We need some cunning test for this!!! Maybe counting cycles (perf counter `0`)?

src/ir_def/ir.ml
ggreif: Ideally we'd be folding this into the `Type.async_sort` (as a GADT).

src/ir_def/arrange_ir.ml
ggreif:
```suggestion
  | CPSAwait (AwaitFut true, t) -> "CPSAwait?" $$ [typ t]
  | CPSAwait (AwaitFut _, t) -> "CPSAwait" $$ [typ t]
```
The `?` means 0 or 1 context switches. WDYT?

src/ir_passes/async.ml
ggreif: GH has a bug. This line is not in the tree any more!

src/lowering/desugar.ml
```diff
@@ -454,7 +456,7 @@ and call_system_func_opt name es obj_typ =
                 (primE (Ir.OtherPrim "trap")
                   [textE "canister_inspect_message explicitly refused message"]))
         | "lowmemory" ->
-          awaitE T.Cmp 
+          awaitE
```
ggreif: unused argument elided

src/prelude/internals.mo
```diff
@@ -364,10 +364,6 @@ func @new_async<T <: Any>() : (@Async<T>, @Cont<T>, @Cont<Error>, @CleanCont) {
 
   var cleanup : @BailCont = @cleanup;
 
-  func clean() {
```
ggreif, line 367: See the lambda in line 399. This reduces the closure size a bit.

test/run-drun/await.mo
```diff
@@ -62,7 +62,7 @@ actor a {
 
     func p():async (Text,Text) { ("fst","snd"); };
     let h = async {
-       let (a,b) = ("a","b"); /* await p(a,b);*/
+       let (a,b) = await p();
```
ggreif, line 65: Maybe this was commented out in a debug session and then snuck into a commit?

src/ir_def/check_ir.ml
```diff
@@ -566,8 +566,8 @@ let rec check_exp env (exp:Ir.exp) : unit =
       typ exp1 <: T.blob;
       T.Opt (T.seq ots) <: t
     | CPSAwait (s, cont_typ), [a; krb] ->
-      let (_, t1) =
-        try T.as_async_sub s T.Non (T.normalize (typ a))
+       let (_, t1) = (*assert (s = AwaitFut false);*)
```
ggreif:
```suggestion
      let (_, t1) = (*assert (s = AwaitFut false);*)
```

src/ir_def/check_ir.ml
```diff
@@ -566,8 +566,8 @@ let rec check_exp env (exp:Ir.exp) : unit =
       typ exp1 <: T.blob;
       T.Opt (T.seq ots) <: t
     | CPSAwait (s, cont_typ), [a; krb] ->
-      let (_, t1) =
-        try T.as_async_sub s T.Non (T.normalize (typ a))
+      let (_, t1) = (*assert (s = AwaitFut false);*)
```
ggreif:
```suggestion
      let (_, t1) =
```

src/ir_def/arrange_ir.ml
```diff
@@ -80,8 +80,8 @@ and prim = function
   | GetLastArrayOffset -> Atom "GetLastArrayOffset"
   | BreakPrim i       -> "BreakPrim"  $$ [id i]
   | RetPrim           -> Atom "RetPrim"
-  | AwaitPrim Type.Fut -> Atom "AwaitPrim"
-  | AwaitPrim Type.Cmp -> Atom "AwaitPrim*"
+  | AwaitPrim (AwaitFut _) -> Atom "AwaitPrim"
```
ggreif:
```suggestion
  | AwaitPrim (AwaitFut true) -> Atom "AwaitPrim?"
  | AwaitPrim (AwaitFut _) -> Atom "AwaitPrim"
```

ggreif: See below...

src/ir_def/construct.ml
```diff
@@ -162,8 +162,9 @@ let asyncE s typ_bind e typ1 =
                       eff = T.(if s = Fut then Await else Triv) }
   }
 
-let awaitE s e =
+let awaitE e =
```
ggreif: dead argument

src/prelude/internals.mo
```diff
@@ -364,10 +364,6 @@ func @new_async<T <: Any>() : (@Async<T>, @Cont<T>, @Cont<Error>, @CleanCont) {
 
   var cleanup : @BailCont = @cleanup;
```
ggreif, line 365: Obviously, otherwise it would stick to the `trivial cleanup action`.

ggreif: Check for
``` Motoko
    cleanup := b;
```

src/mo_frontend/typing.ml
```diff
@@ -1738,7 +1738,10 @@ and infer_exp'' env exp : T.typ =
     let t0 = check_AwaitCap env "await" exp.at in
     let t1 = infer_exp_promote env exp1 in
     (try
-       let (t2, t3) = T.as_async_sub s t0 t1 in
+       let s1 = match s, t1 with
+         | T.(Cmp, Async(Fut, _, _)) -> T.Fut (* we can await* an async *)
+         | _ -> s in
+       let (t2, t3) = T.as_async_sub s1 t0 t1 in
```
crusso, line 1744: Maybe this would be cleaner if T.as_async_sub returned the sort of t1  and then you act accordingly:

let (s1,t2,t3) = T.as_async _sub s t0 t1 in
...
if (s = T.Fut && s1 == T.Cmp)  then error....;

src/mo_interpreter/interpret.ml
```diff
@@ -736,7 +736,9 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
       (fun v1 -> await env exp.at (V.as_async v1) k)
   | AwaitE (T.Cmp, exp1) ->
     interpret_exp env exp1
-      (fun v1 -> (V.as_comp v1) k (Option.get env.throws))
+      (fun v1 -> match v1 with
+         | V.Async _ -> await env exp.at (V.as_async v1) k
```
crusso: Doesn't line 740 still do the extra scheduling when the async is complete?

src/ir_interpreter/interpret_ir.ml
```diff
@@ -369,10 +369,10 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
       | BreakPrim id, [v1] -> find id env.labs v1
       | RetPrim, [v1] -> Option.get env.rets v1
       | ThrowPrim, [v1] -> Option.get env.throws v1
-      | AwaitPrim Type.Fut, [v1] ->
+      | AwaitPrim (AwaitFut _), [v1] ->
         assert env.flavor.has_await;
         await env exp.at (V.as_async v1) k (Option.get env.throws)
-      | AwaitPrim Type.Cmp, [v1] ->
+      | AwaitPrim AwaitCmp, [v1] ->
         assert env.flavor.has_await;
```
crusso, line 384: Won't this still crash when the argument is a Future (see `V.as_comp`)?

src/ir_def/ir.ml
```diff
@@ -134,7 +134,7 @@ and prim =
   | IdxBlobPrim                       (* blob indexing *)
   | BreakPrim of id                   (* break *)
   | RetPrim                           (* return *)
-  | AwaitPrim of Type.async_sort       (* await/await* *)
+  | AwaitPrim of await_sort           (* await/await* *)
```
crusso:
```suggestion
  | AwaitPrim of Type.async_sort * Type.async_sort           (* await/await* *)
```
Might be cleaner where the second sort is the sort of the async. And then you rule out the case where the pair is (T.Fut, T.Cmp) in check_ir and `short-circuit` when the argument is (T.Cmp, T.Fut).

test/run-drun/blockless.mo
crusso: What about

```
var changed = false;
await async { 
   let a = async ();
   await a;
  changed := true;
  await* a; // fast await
  assert false;
};
assert (not changed);
```
vs:

```
var changed = false;
await async { 
   let a = async ();
   await a;
  changed := true;
  await a; // proper await
  assert false;
};
assert changed;
```

crusso (changes requested):
Looks promising. See suggestions.
I think the intepreters aren't quite right yet... (still do extra hops/crash in IR interpreter)

src/ir_interpreter/interpret_ir.ml
ggreif: No, it won't crash, because desugaring maps the `await* async 42` construct to `AwaitFut true`. So `AwaitCmp` always deals with a computation.

src/mo_frontend/typing.ml
ggreif: My GADT suggestion would eliminate that:
``` Haskell
data kind Mode = Intro | Elim
data AsyncSort Mode where
  Fut :: AsyncSort both
  FutNoBlock :: AsyncSort Elim
  Cmp :: AsyncSort both
```
and then `AsyncE (AsyncSort Intro, ...)` and `AwaitPrim (AsyncSort Elim, ...)`

src/mo_interpreter/interpret.ml
ggreif: Yes, it does, but the interpreter doesn't really care about it, or does it? Does the interpreter roll back state?

test/run-drun/blockless.mo
ggreif: Ooooh, I see, so you don't measure the operational cost benefit, but the diff in semantics... So the interpreter actually needs to follow it! Good point. Let me see if the `drun` executes accordingly.

ggreif: Another way would be to observe that the send queue gets reset (`async{print}; await* fulfilled; trap`) vs. flushed (`async{print}; await fulfilled; trap`).

ggreif: I implemented both ideas in 7dd2f453e

src/ir_def/ir.ml
ggreif: I'll do a follow-up PR for this cleanup.

src/ir_def/ir.ml
ggreif: will clean up in another PR

crusso:
Thinking about this some more, maybe we should *not* overload 'await*', but instead add 'await?' as a variant of await on futures.

Inspired by your pretty printing of IE and the fact that we are actually tweaking await, not await*.

The your current await_sort also makes more sense I think.

src/ir_def/ir.ml
crusso: Maybe you don't need to actually, if we change the syntax to 'await?' then this your current defs are sensible

ggreif:
> Thinking about this some more, maybe we should _not_ overload 'await*', but instead add 'await?' as a variant of await on futures.
> 
> Inspired by your pretty printing of IR and the fact that we are actually tweaking await, not await*.
> 
> Then your current await_sort also makes more sense I think.
> 
> And users could grep for 'await?' If needed.

The problem is, `?` is a prefix operator already.

```
$ rlwrap moc
Motoko compiler (source 0.14.11-36-g8491185d1-dirty)
> await?777;
stdin:2.6-2.10: type error [M0088], expected async type, but expression has type
  ?Nat
```

src/ir_def/construct.ml
crusso: I would actually pass in the `await_sort` rather than infer it, so adjust the old code to take Type.await_sort.

src/mo_interpreter/interpret.ml
crusso: No, but it would still do a context switch which you might observe. Can you just pass the sort to await to do the right thing?

test/run-drun/await.mo
crusso: git blame?

ggreif: git show af6e0ff1d6 -- test/run-drun/await.mo

from 2019! Initial commit already contains the comment. Slipped through the review...

src/mo_interpreter/interpret.ml
ggreif: Sure I can!

Changelog.md
```diff
@@ -1,12 +1,12 @@
...
-* motoko (`moc`)
-
 ## 0.14.12 (FUTURE)
 
+* motoko (`moc`)
+
   * Added the `rootKey` primitive (#4994).
 
-  * Allow `await*` to force `async` values, avoiding the commit point when the future is already fulfilled (#5215).
+  * Introduce `await?` to force `async` values, avoiding the commit point when the future is already fulfilled (#5215).
```
```suggestion
  * Introduce `await?` to synchronize `async` futures, avoiding the commit point when already fulfilled (#5215).
```

src/ir_def/construct.ml
ggreif: Done!

src/ir_interpreter/interpret_ir.ml
```diff
@@ -401,13 +409,13 @@ and interpret_exp_mut env exp (k : V.value V.cont) =
...
         end
-      | CPSAwait _, [v1; v2] ->
+      | CPSAwait (sort, _), [v1; v2] ->
         assert (not env.flavor.has_await && env.flavor.has_async_typ);
         begin match V.as_tup v2 with
          | [vf; vr] ->
            let (_, f) = V.as_func vf in
            let (_, r) = V.as_func vr in
-           await env exp.at (V.as_async v1)
+           await env exp.at (sort <> T.AwaitFut false)(*FIXME?*) (V.as_async v1)
```
Kamirus: FIXME?

ggreif: Good point. It _was_ a FIXME, but now it should be fine. I was unsure whether `AwaitCmp` is possible here. It probably isn't, but we don't behave differently for that as before, even then.

ggreif:
```suggestion
           await env exp.at (sort <> T.AwaitFut false) (V.as_async v1)
```

Kamirus: So maybe an assert just in case?

ggreif: No need, this code is working just fine for ages.

doc/md/fundamentals/5-actors-async.md
```diff
@@ -124,6 +124,28 @@ Unlike a local function call, which waits for the result before continuing, a sh
 
 If you `await` the same future again, it just returns the same result or error. Even if the future is already done, `await` will briefly suspend all pending state changes and outgoing messages. This means that you can rely on every `await` to commit state, whether its future is still in progress or already completed.
 
+## Using `await?` to efficiently await concurrent futures
+
+An `await` will always suspend execution and commit state, even if its future is already complete.
+
+When several futures are issued in parallel and racing to complete, it can be more efficient to opt out of the unconditional behavior of `await`
+and immediately continue with a result when it is available:
```
crusso:
```suggestion
When several futures are issued in parallel and racing to complete, it can be more efficient to opt out of the unconditional behavior of `await` and immediately continue with a result when it is available:
```

test/run-drun/blockless.mo
```diff
@@ -0,0 +1,45 @@
...
+        ignore async debugPrint "Peek-a-boo!";
+        if shortCircuit
+          await? a // fast await
+        else
+          await a; // proper await
+        assert false;
+      } catch _ {}
+    };
+
+    public func state(shortCircuit : Bool) : async () {
```
crusso, line 20: It might be worth adding a version of this test (or entire file) where `let a = async { throw Prim.error(..) }` (exiting via `throw`, not `return`) and we check the results accordingly.

crusso (approved):
LGTM, but maybe add one more test for await? on async that exits via throw

test/run-drun/blockless.mo
ggreif: See 087c2c721!

PR #5216, opened 2025-06-05
bugfix: fix stack overflow due to List.map in linker

crusso:
* Our linker now use a non-tail-rec List.map that stack overflows on large.

I'm not sure which instance is causing the failure so I've just replaced all by a tail-recursive safe_map.

- [ ] TODO: maybe review the latest linker changes for more lurking issues (List.mapi?)


crusso:
Subsumed by #5218

PR #5233, opened 2025-06-06
bugfix: deduplicate static constants to reduce size of root array in EOP

crusso:
fixes #5226

Debugging a compilation bug with CycleOps revealed that `moc` EOP creates a much, much larger object pool compared with the classical compilers static heap allocations. The problem is that it was only sharing constants allocated at the same IR location, not different locations, leading to many duplicated allocations (e.g. for inlined and duplicated backend trap messages).

Shaves 2MB off CycleOps binary (9.5-7.5 MB) and reduces number of root entries by 60% and function count (a proxy for constant allocations) by half.

Main culprit was not structured constants as much as blob literals that were rarely shared (if at all)

- [x] cleanup
- [x] implement Const eq interms of new compare
- [x] refactor lookups somehow
- [x] add some -v verbose mode diagnostics
- [x] fixes quadratic extension of object pool using reversed list with maintained length 

The PR now shares all blob lits with same content, including object field descriptors, and structured constants with the same value but allocated in different source locations.

This should no only reduce code size and initialization time, but also pressure on the gc with a much smaller root array.

FUTURE: we could probably load the (many) static strings corresponding to backend traps on demand, just before trapping, rather than preallocate them, or reduce their number somehow (don't allow compile-time computed ones, for example)

Sample -v diagnostics (to std errror)
```
shared constants = 4280
pool size = 4692
pool report = codegen/compile_enhanced.ml, line 9082[307]
codegen/compile_enhanced.ml, line 4481[24]
codegen/compile_enhanced.ml, line 4125[112]
codegen/compile_enhanced.ml, line 3797[2889]
codegen/compile_enhanced.ml, line 3668[5]
codegen/compile_enhanced.ml, line 2830[2]
codegen/compile_enhanced.ml, line 2424[939]
codegen/compile_enhanced.ml, line 2304[4]
codegen/compile_enhanced.ml, line 2242[198]
codegen/compile_enhanced.ml, line 2210[212]
```
The line numbers report counts of unique allocations attributed to the line in the backend. We used to have 1000s of blob all

File sizes (before/after):

```
crusso@crusso-Virtual-Machine:~/clean/motoko$ ls -l ../../CycleOps/.dfxmaster-eop/local/canisters/cycleops/cycleops.wasm
-rw-rw-r-- 1 crusso crusso 9554954 Jun  7 09:23 ../../CycleOps/.dfxmaster-eop/local/canisters/cycleops/cycleops.wasm
[...]

PR #5242, opened 2025-06-09
optimization: remove (most) quadratic list operations in compile_enhanced.ml

crusso:
Builds on #5233:

- removes quadratic behaviour in backend due to list based table representation used allocate indices to Wasm functions, datasegments, locals, shared objects  etc...

- shaves 12% off wall-clock compilation time for CycleOps (eop)  

(a similar optimization could apply to the classical backend)

### DIFF
--- a/src/codegen/compile_enhanced.ml
+++ b/src/codegen/compile_enhanced.ml
@@ -623,7 +623,7 @@ module E = struct
   let add_local_name (env : t) li name =
     let _ = reg env.local_names (li, name) in ()
 
-  let get_locals (env : t) = !(env.locals)
+  let get_locals (env : t) = Table.to_list !(env.locals)
   let get_local_names (env : t) : (int32 * string) list = Table.to_list !(env.local_names)
 
   let _add_other_import (env : t) m =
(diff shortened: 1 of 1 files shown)


src/codegen/compile_enhanced.ml
```diff
@@ -831,17 +849,18 @@ module E = struct
...
-      Printf.eprintf "\nshared constants = %i" (ConstEnv.cardinal (!(env.constant_pool)));
+      Printf.eprintf "\ndata segments = %i" (Table.length !(env.data_segments));
+      Printf.eprintf "\nshared constants = %i" (ConstEnv.cardinal !(env.constant_pool));
       Printf.eprintf "\npool size = %i" (object_pool_size env);
       Printf.eprintf "\npool report = %s" profile
     end
 
   let iterate_object_pool (env : t) f =
-    G.concat_mapi f (List.map (fun (l,a) -> a) (List.rev !(env.object_pool.objects)))
+    G.concat_mapi f (List.map (fun (l, a) -> a) (Table.to_list !(env.object_pool.objects)))
```
Kamirus, line 863: The order does not matter here? It's reversed compared to the old one, right?

src/codegen/compile_enhanced.ml
```diff
@@ -815,13 +834,12 @@ module E = struct
 
   let object_pool_add (env : t) line (allocation : t -> G.t) : int64 =
     if !(env.object_pool.frozen) then raise (Invalid_argument "Object pool frozen");
-    let index = !(env.object_pool.length) in
-    env.object_pool.objects := (line, allocation) :: !(env.object_pool.objects);
-    env.object_pool.length := index + 1;
+    let index, t = Table.add !(env.object_pool.objects) (line, allocation) in
```
Kamirus, line 837: For the record: this is adding to the back not the front, but order does not matter here, right?

Kamirus (approved):
with comments

src/codegen/compile_enhanced.ml
crusso: It should be producing the same order, since Table.to_list does a reverse.

Kamirus: But the `Table` stores objects in reverse order compared to the old changes.
So to get the same order as before you'd need to `List.rev`

PR #5250, opened 2025-06-11
intoduce and use `await?`

ggreif:


doc/md/examples/grammar.txt
```diff
@@ -217,6 +217,7 @@
     <parenthetical>? 'async' <exp_nest>
     'async*' <exp_nest>
     'await' <exp_nest>
+    'await''?' <exp_nest>
```
crusso: Something wrong here? Maybe you need to escape the `?` in the sed pattern

ggreif: Nope!
```
$ sed -e s/AWAITQUEST/\'await?\'/g
AWAITQUEST
'await?'
```
but maybe there is a later `s/\?/'?'/g`?

ggreif: I.e.
```
$ sed -e s/AWAITQUEST/\'await?\'/g -e "s/\?/'?'/g"
AWAITQUEST
'await'?''
```

ggreif: Nope: https://github.com/dfinity/motoko/blob/8c59176cfe67b27dcab13d3d96e5ab9b31950f0d/src/gen-grammar/grammar.sed#L75

crusso (approved):
LGTM!

I think this is nicer - agree?

doc/md/examples/grammar.txt
ggreif: Fixed by reordering

PR #5251, opened 2025-06-11
test: add `only.mk` and `accept.mk`

rvanasa:
This PR adds `*.mk` files written by @ggreif which simplify running and accepting changes for individual tests.

```sh
make -C test/run example.only # Run `example.mo`
make -C test/run example.accept # Run and accept output for `example.mo`
```

An additional property of the final `accept.mk`: the workflow

1. introduce a type error in `test.mo`
2. `make -s ... test.accept`  ← stages error output, unstages passing output
3. fix the error
4. `make -s ... test.accept`  ← restores passing output, unstages error output

leaves the repo index in exactly its original state — provided the `.ok` files are tracked and `test.mo` is restored to its original content. Step 4 re-stages `test.mo` and `test.drun-run.ok` (both now matching HEAD) and removes the ephemeral `test.tc.ok` / `test.tc.ret.ok` entries via the `git status --porcelain | awk | xargs git rm --cached` cleanup.

### DIFF
--- a/test/accept.mk
+++ b/test/accept.mk
@@ -1,0 +1,9 @@
+.PHONY: %.accept
+%.accept: %.mo
+	../run.sh -a $(RUNFLAGS) $<
+	@ git status -s $< ok
+	@ echo STAGING: $< $(wildcard ok/$(basename $<).*)
+	@ if [ -n "$(wildcard ok/$(basename $<).*)" ] \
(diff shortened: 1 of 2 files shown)


christoph-dfinity:
Wouldn't it be easier to directly use `test/run.sh` in those cases?
```sh
test/run.sh test/run/example.mo # Run `example.mo`
test/run.sh test/run/example.mo -a # Run and accept output for `example.mo`
```

PR #5255, opened 2025-06-12
chore: update release workflows to use GitHub App token for authentication

Kamirus:
- Added step to create a GitHub App token in the release PR workflow.
- Updated checkout step to use the generated token for accessing the repository.
- Removed the trigger release workflow step from the release workflow, simplifying the process.

### DIFF
--- a/.github/workflows/release-pr.yml
+++ b/.github/workflows/release-pr.yml
@@ -51,10 +51,17 @@ jobs:
     # Only when the PR is merged!
     if: github.event.pull_request.merged
     steps:
+    - name: Create GitHub App Token
+      uses: actions/create-github-app-token@v2
+      id: app-token
+      with:
+        app-id: ${{ vars.PR_AUTOMATION_BOT_PUBLIC_APP_ID }}
+        private-key: ${{ secrets.PR_AUTOMATION_BOT_PUBLIC_PRIVATE_KEY }}
     - uses: actions/checkout@v4
       with:
         # Checkout the merge commit so we can tag it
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -28,9 +22,8 @@ jobs:
     outputs:
       version: ${{ inputs.version || github.ref_name }}
     steps:
-      - name: Ensure tag is from master
-        if: github.event_name == 'push' && github.event.base_ref != 'refs/heads/master'
```
Kamirus, line 32: this was broken: `github.event.base_ref` was `null`.
I wanted this extra check, but for now I have no idea how to fix it and make it simple -- remove it for now

.github/workflows/release-pr.yml
```diff
@@ -51,10 +51,17 @@ jobs:
...
+      uses: actions/create-github-app-token@v2
+      id: app-token
+      with:
+        app-id: ${{ vars.PR_AUTOMATION_BOT_PUBLIC_APP_ID }}
+        private-key: ${{ secrets.PR_AUTOMATION_BOT_PUBLIC_PRIVATE_KEY }}
     - uses: actions/checkout@v4
       with:
         # Checkout the merge commit so we can tag it
         ref: ${{ github.event.pull_request.merge_commit_sha }}
+        token: ${{ steps.app-token.outputs.token }}
```
Kamirus, line 64: let's see if this will trigger the release so that we would not need to do that manually

alexandru-uta (approved):

PR #5256, opened 2025-06-12
feat: adds a primitive `Array_tabulateVar` function

christoph-dfinity:
This gives me a nice 2-3% speedup on the Hamt benchmarks.

Claudio wasn't sure if we needed extra write barriers compared to the non-mutable version for EOP, so I'm adding Luc to the reviewers.

### DIFF
--- a/src/mo_values/prim.ml
+++ b/src/mo_values/prim.ml
@@ -321,6 +321,18 @@ let prim trap =
       in go (fun xs -> xs) k 0
     | _ -> assert false
     )
+  | "Array.tabulateVar" -> fun c v k ->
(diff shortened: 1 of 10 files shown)


src/mo_values/prim.ml
```diff
@@ -321,6 +321,18 @@ let prim trap =
       in go (fun xs -> xs) k 0
     | _ -> assert false
     )
+  | "Array.tabulateVar" -> fun c v k ->
```
crusso, line 325: Err, let's add a todo to optimize both of these - they are way more expensive than they should be... As is representing an array by an array of refs..

PR #5271, opened 2025-06-17
chore: improve stable compatibility error message

### LINKED ISSUE #5265: enhancement: improve stable compatibility error message
Just noticed tiny detail of error message, e.g.
Compatibility error [M0170], stable variable z of previous type
  Text
cannot be consumed at new type
  Bool without data loss. Use an explicit migration function.
saying "without data loss" although type is completely non-compatible (unrelated to dropping information for lossy super-type)
Just noticed that LLM get driven into wrong fixing direction, assuming fields were missing during migration instead of mismatching types, e.g. Int -> Nat migration...

crusso:
Fixes #5265

### DIFF
--- a/Building.md
+++ b/Building.md
@@ -1,3 +1,18 @@
+- [Nix setup](#nix-setup)
+- [Installation using Nix](#installation-using-nix)
+  - [Other tools](#other-tools)
+- [Development using Nix](#development-using-nix)
+- [Replicating CI locally](#replicating-ci-locally)
+- [Making releases](#making-releases)
+  - [1. Update Changelog](#1-update-changelog)
+  - [2. Open a release PR](#2-open-a-release-pr)
+  - [3. Wait for the release to complete, and verify it](#3-wait-for-the-release-to-complete-and-verify-it)
+  - [4. Update `motoko-base`](#4-update-motoko-base)
+  - [Downstream](#downstream)
+- [Coverage report](#coverage-report)
+- [Profile the compiler](#profile-the-compiler)
+- [Benchmarking the RTS](#benchmarking-the-rts)
+
 ## Nix setup
 
 The Motoko build system relies on [Nix](https://nixos.org/) to manage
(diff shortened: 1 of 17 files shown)


src/mo_frontend/stability.ml
```diff
@@ -65,8 +73,11 @@ let match_stab_sig sig1 sig2 : unit Diag.result =
       | tf1::tfs1', (is_required, tf2)::tfs2' ->
         (match Type.compare_field tf1 tf2 with
          | 0 ->
-            if not (Type.stable_sub (as_immut tf1.typ) (as_immut tf2.typ)) then
-              error_sub s tf1 tf2;
+            begin
+              if not (Type.sub (as_immut tf1.typ) (as_immut tf2.typ)) then
+                error_sub s tf1 tf2 else
+                if not (Type.stable_sub (as_immut tf1.typ) (as_immut tf2.typ)) then                error_stable_sub s tf1 tf2;
```
christoph-dfinity: weird formatting

src/mo_frontend/stability.ml
```diff
@@ -24,7 +24,15 @@ let error_discard s tf =
...
     (Diag.error_message Source.no_region "M0170" cat
-      (Format.asprintf "stable variable %s of previous type%a\ncannot be consumed at new type%a without data loss. Use an explicit migration function."
+      (Format.asprintf "stable variable %s of previous type%a\ncannot be consumed at unrelated type%a. Use an explicit migration function."
+        tf1.lab
+        display_typ_expand tf1.typ
+        display_typ_expand tf2.typ))
+
+let error_stable_sub s tf1 tf2 =
+  Diag.add_msg s
+    (Diag.error_message Source.no_region "M0216" cat
```
christoph-dfinity, line 34: Not really a problem for this PR, but I'm surprised we don't have a location to report here.

test/cmp/ok/loss0-1.cmp.ok
```diff
@@ -0,0 +1,9 @@
+(unknown location): Compatibility error [M0216], stable variable one of previous type
+  var {field : Nat}
+cannot be consumed at supertype
+  var {} without data loss. Use an explicit migration function.
```
christoph-dfinity:
```suggestion
(unknown location): Compatibility error [M0216], stable variable one of previous type
  var {field : Nat}
cannot be consumed at supertype
  var {} 
without data loss. Use an explicit migration function.
```

How would you feel about this formatting? Makes the mismatched types pop out a bit more.

src/mo_frontend/stability.ml
crusso: I looked into this a bit but the locations only make sense for these particular tests. Generally, in dfx the .most files will be extracted from the wasm or custom section and won't be relevant sources for the user.

PR #5276, opened 2025-06-20
feat: lazy wasm imports

### LINKED ISSUE #5284: 0.14.13 fails with  Wasm module is not valid
0.14.9 works fine. I've attached the wasms.

Note: the error is not for the wasm that was installed. The wasm itself installs a wasm an imports a number of different canister classes including an NFT, a fungible token and a wallet.

Since I'm not getting any messages in my dfx window I'm guessing it is my first module that it tries to deploy:

ExperimentalCycles.add<system>(1_000_000_000_000);
    let icrc8 = await ICRC8IntentMarketplace.ICRC8IntentMarketplaceCanister(null); // Deploy marketplace

    ExperimentalCycles.add<system>(1_000_000_000_000);
    let fungible = await Fungible.Token(null); // Deploy fungible token
    ExperimentalCycles.add<system>(1_000_000_000_000);

    let nft = await NFT.Example({ icrc37_args = null; icrc3_args = null; icrc7_args = null }); // Deploy NFT canister
    ExperimentalCycles.add<system>(1_000_000_000_000);

Reinstalled code for canister list_test, with canister ID uxrrr-q7777-77774-qaaaq-cai
Deployed canisters.
URLs:
  Backend canister via Candid interface:
    list_test: http://127.0.0.1:8080/?canisterId=u6s2n-gx777-77774-qaaba-cai&id=uxrrr-q7777-77774-qaaaq-cai
afat@virusbook ICRC8IntentMarketplace-mo % dfx ledger fabricate-cycles --canister list_test
Fabricating 10000000000000 cycles onto list_test
Fabricated 10000000000000 cycles, updated balance: 28_500_000_000_000 cycles
afat@virusbook ICRC8IntentMarketplace-mo % dfx canister call list_test runTests            
Error: Failed update call.
Caused by: The replica returned a rejection error: reject code CanisterReject, reject message Error from Canister vizcg-th777-77774-qaaea-cai: Canister's Wasm module is not valid: Wasm module has an invalid import section. Module imports function 'root_key_copy' from 'ic0' that is not exported by the runtime..
This is likely an error with the compiler/CDK toolchain being used to build the canister. Please report the error to IC devs on the forum: https://forum.dfinity.org and include which language/CDK was used to create the canister., error code Some("IC0406")

[Archive.zip](https://github.com/user-attachments/files/20873940/Archive.zip)

Kamirus:
Remove unused wasm imports.

Resolves #5284 with problems like:
```
Canister's Wasm module is not valid: Wasm module has an invalid import section.
Module imports function 'root_key_copy' from 'ic0' that is not exported by the runtime
```
When the import is actually unused.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -175,11 +175,11 @@
         "treefmt-nix": "treefmt-nix"
       },
       "locked": {
-        "lastModified": 1750933557,
-        "narHash": "sha256-xm2zXjvsQxtr3ipRmXDfetd65VMorICq7F9mtiOaVFw=",
+        "lastModified": 1751242003,
+        "narHash": "sha256-4O4jfNCbPl6Nb6qxrWAzBFN/2I0mDDfGuuIuv2wyn58=",
         "owner": "Mic92",
         "repo": "nix-update",
-        "rev": "f612fb07436603c68f7d90659d4d6ef82b03a5e6",
+        "rev": "0e366c09ae7769a4e334e7fb98acdfcd78dc1f66",
         "type": "github"
       },
       "original": {


test/run-drun/empty-actor-enhanced.mo
```diff
@@ -1,7 +1,7 @@
 //ENHANCED-ORTHOGONAL-PERSISTENCE-ONLY
 actor {};
 
-// CHECK:  (func $@motoko_async_destabilization (type 0)
```
Kamirus, line 4: `linkModule.ml` reorders used types based on the imports, now it was `(type 8)`.
Type and import locations changed, but each import got the same type, just reordered.
Removing the `type` location from the test to not make it unstable again.

test/run-drun/remove-unused-imports.mo
```diff
@@ -0,0 +1,13 @@
...
+  public func print() : async () {
+    Prim.debugPrint("print");
+  };
+};
+
+// CHECK-NOT: import "ic0" "root_key_copy"
+// CHECK-NOT: import "ic0" "root_key_size"
+// CHECK: import "ic0" "debug_print"
+// CHECK-NOT: import "ic0" "root_key_copy"
+// CHECK-NOT: import "ic0" "root_key_size"
```
Kamirus, line 13: there should be no root key import when it's not used!

src/wasm-exts/ast.ml
```diff
@@ -324,3 +324,42 @@ let is_dwarf_like = function
   | Meta m -> is_dwarf_like' m
   | _ -> false
 
+
+(* AST traversals *)
```
Kamirus, line 328: moved from `linkModule.ml`

src/lib/lib.ml
```diff
@@ -424,6 +424,10 @@ struct
       (match list with
        | [] -> false
        | hd' :: tl' -> equal hd hd' && is_prefix equal tl tl')
+  
+  (* tail-recursive map *)
+  (* TODO: can be replaced by [@tail_mod_cons] map (OCaml 0.4.14)/List.map once we upgrade OCaml *)
+  let safe_map f l = List.rev (List.rev_map f l)
```
Kamirus, line 430: moved from `linkModule.ml`

src/codegen/compile_common.ml
```diff
@@ -0,0 +1,179 @@
...
+  let add_func_import (env : t) modname funcname arg_tys ret_tys =
+    if Table.length !(env.funcs) <> 0 then
+      raise (CodegenError "Add all imports before all functions!");
+
+    let i = {
+      module_name = Lib.Utf8.decode modname;
+      item_name = Lib.Utf8.decode funcname;
+      idesc = nr (FuncImport (nr (func_type env (FuncType (arg_tys, ret_tys)))))
+    } in
+    let used = ref false in
```
Kamirus, line 124: added a `ref` to track used imports

src/codegen/compile_common.ml
```diff
@@ -0,0 +1,179 @@
...
+    ref := t;
+    (i32, Lib.Promise.fulfill p)
+
+
+  module NameEnv = Env.Make(String)
+  
+  type local_names = (int32 * string) list (* For the debug section: Names of locals *)
+  type func_with_names = func * local_names
+  type lazy_function = (int32, func_with_names) Lib.AllocOnUse.t
+end
```
Kamirus, line 68: common code, taken from `compile_enhanced.ml`.
`Table` is not used in the `compile_classical.ml` apart from the `Imports` part.

crusso (approved):
Looks ok to me. I guess the only risk is that we might store in a funcref to an import in the function table and I'm not sure this code regards that as  a "use" but I don't think we ever do that anyway.

Nice refactoring! We should actually use the table abstraction in the classical backend too, but I thought I'd stop investing in that if we are planning to ditch it...

PR #5278, opened 2025-06-21
Stable Functions: Minor Fix (Promoting Stable Functions with Generic Type Parameters)

luc-blaeser:
Fixing https://github.com/dfinity/motoko/pull/5143

Essence of fix:
* Can only promote stable function type to flexible type, when there are no generic type parameters.

This is because:
* Stable functions imply stable type parameters
* Flexible functions imply flexible type parameters
* Promotion requires generic invariance on stability

### DIFF
--- a/src/lowering/desugar.ml
+++ b/src/lowering/desugar.ml
@@ -122,14 +122,7 @@ and exp' at note = function
     let tbs' = typ_binds tbs in
     let vars = List.map (fun (tb : I.typ_bind) -> T.Con (tb.it.I.con, [])) tbs' in
     let tys = List.map (T.open_ vars) res_tys in
-    let is_async = match typ_opt with
-      | Some { it = AsyncT _; _ } -> true
-      | _ -> false
-    in
-    let s = match s, is_async with
-    | T.Local T.Stable, true -> T.Local T.Flexible
-    | _ -> s
-    in
+    let (s, _, _, _, _) = T.as_func note.Note.typ in
     I.FuncE (name, s, control, tbs', args, tys, !closure, wrap (exp e))
   (* Primitive functions in the prelude have particular shapes *)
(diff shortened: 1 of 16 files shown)


src/mo_types/type.ml
```diff
@@ -1059,9 +1059,13 @@ let rec rel_typ d rel eq t1 t2 =
   | _, _ -> false
   end
 
-and rel_sort s1 s2 =
+and rel_sort s1 s2 tbs1 =
   match s1, s2 with
-  | Local Stable, Local Flexible -> true (* stable function can be assigned to flexible function *)
+  | Local Stable, Local Flexible -> 
+    not (List.exists (fun b -> b.sort = Type) tbs1)
```
luc-blaeser, line 1065: This is the fix

src/lowering/desugar.ml
```diff
@@ -122,6 +122,14 @@ and exp' at note = function
     let tbs' = typ_binds tbs in
     let vars = List.map (fun (tb : I.typ_bind) -> T.Con (tb.it.I.con, [])) tbs' in
     let tys = List.map (T.open_ vars) res_tys in
+    let is_async = match typ_opt with
```
luc-blaeser: Another detail: 
* Desugar needs the same rule as `typing.ml` (uncritical)

src/prelude/internals.mo
```diff
@@ -41,35 +41,35 @@ func @reset_cycles() {
 // Note that these return functions!
 // (Some optimizations in the backend might be feasible.)
 
-func @immut_array_get<A>(xs : [A]) : Nat -> A =
+let @immut_array_get = func<A>(xs : [A]) : Nat -> A =
```
luc-blaeser, line 44: Making these flexible

PR #5285, opened 2025-06-24
doc: Fix title

jessiemongeon1:


crusso (commented):
This is a generated file, we should change mo_doc instead.


But I'll rubberstamp anyway

crusso (approved):

PR #5288, opened 2025-06-25
fix(mo-doc): rewords heading for package index

christoph-dfinity:
Makes it so we produce the desired documentation output from #5285

### DIFF
--- a/src/docs/html.ml
+++ b/src/docs/html.ml
@@ -385,7 +385,7 @@ let make_index : string option -> render_input list -> string =
          (string
             (match package_opt with
             | None -> "Motoko package"
-            | Some s -> "Motoko " ^ s ^ " library"))
+            | Some s -> "Motoko " ^ s ^ " package"))
       ++ ul ~cls:"index-listing" ~licls:"index-item" (List.map make_link inputs)
       )
   in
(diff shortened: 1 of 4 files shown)


src/docs/html.ml
```diff
@@ -384,8 +384,8 @@ let make_index : string option -> render_input list -> string =
       (h1 ~cls:"index-header"
          (string
             (match package_opt with
-            | None -> "Index of modules"
-            | Some s -> "Index of module in package " ^ s))
+            | None -> "Motoko library"
```
jessiemongeon1:
```suggestion
            | None -> "Motoko package"
```

src/docs/html.ml
```diff
@@ -384,8 +384,8 @@ let make_index : string option -> render_input list -> string =
...
+            | Some s -> "Motoko " ^ s ^ " library"))
```
jessiemongeon1:
```suggestion
            | Some s -> "Motoko " ^ s ^ " package"))
```

src/docs/plain.ml
```diff
@@ -308,8 +308,8 @@ let render_docs : Common.render_input -> string =
 let make_index : string option -> Common.render_input list -> string =
  fun package_opt inputs ->
   let buf = Buffer.create 1024 in
-  bprintf buf "# Index%s\n\n"
-    (match package_opt with None -> "" | Some s -> " of package " ^ s);
+  bprintf buf "# Motoko %slibrary\n\n"
```
jessiemongeon1:
```suggestion
  bprintf buf "# Motoko %spackage\n\n"
```

crusso (approved):

PR #5290, opened 2025-06-25
docs: Remove Motoko ghost

jessiemongeon1:


rvanasa (commented):
Is there a reason to remove the logo aside from it being unnecessary? 

I don't really have an opinion about this but wanted to mention that the logo adds personality and subtly distinguishes the Motoko docs from other parts of the website.

rvanasa (approved):

PR #5303, opened 2025-07-01
feat: `moc` defaults to `enhanced orthogonal persistence`

alexandru-uta:
This PR brings a breaking change. From now on, the `moc` compiler defaults to enhanced orthogonal persistence. The old behavior can still be triggered by using the `--legacy-persistence` flag, which also enables the garbage collectors available for it.

This PR makes the following changes:
* adjusts the compiler flags
* adjusts the test runner logic
* adjusts the nix build flags
* adjusts some of the tests, which ran in mixed persistence modes (classical + eop)


src/exes/moc.ml
```diff
@@ -193,9 +193,9 @@ let argspec = [
   " Trap, don't throw an `Error`, when an IC call fails due to destination queue full or freezing threshold is crossed. Emulates behaviour of moc versions < 0.8.0.";
 
   (* persistence *)
-  "--enhanced-orthogonal-persistence",
```
crusso, line 196: removing the old flag might just break code that specifies it. Maybe its best to keep the flag and add a new one to turn the feature off.

crusso (commented):
Trying something a little simpler in #5305.

PR #5308, opened 2025-07-02
feat: guard against implicit migration from 32-bit to 64-bit

crusso:
Implement a safeguard to prevent unintentional migration from 32 to 64 bit of existing projects recompiled with new defaults.

Migration will trap and rollback unless explicitly compiled with --enhanced-orthogonal-persistence.


src/exes/moc.ml
```diff
@@ -194,14 +194,19 @@ let argspec = [
...
+  Arg.Unit (fun () -> Flags.enhanced_orthogonal_persistence := true;
+                      Flags.explicit_enhanced_orthogonal_persistence := true),
+  " use enhanced orthogonal persistence (default): Scalable and fast upgrades using a persistent 64-bit main memory. Also, enable upgrade from classical to enhanced orthogonal persistence";
 
   (* persistence *)
   "--legacy-persistence",
   Arg.Unit (fun () -> Flags.enhanced_orthogonal_persistence := false),
   " use legacy (classical) persistence. This also enables the usage of --copying-gc, --compacting-gc, and --generational-gc. Deprecated in favor of the new enhanced orthogonal persistence, which is default. Legacy persistence will be removed in the future.";
 
+  "-unguarded-enhanced-orthogonal-persistence",
```
Kamirus, line 206: why do we need that? a test without any flag is not enough?

crusso: No, because all our tests are run with --legacy-persistence for now to avoid rewriting them all. This is a hack to restore implicit eop, just for the purpose of the test. Ugly, I know, but I would like to test the safeguard somehow... Any suggestions?

Changelog.md
```diff
@@ -21,6 +24,7 @@
     * The garbage collector is fixed to incremental GC and cannot be chosen.
     * `Float.format(#hex prec, x)` is no longer supported (expected to be very rarely used in practice).
     * The debug print format of `NaN` changes (originally `nan`).
+
```
```suggestion
```

Changelog.md
```diff
@@ -21,6 +24,7 @@
     * The garbage collector is fixed to incremental GC and cannot be chosen.
     * `Float.format(#hex prec, x)` is no longer supported (expected to be very rarely used in practice).
     * The debug print format of `NaN` changes (originally `nan`).
```
ggreif: is this intended?

src/exes/moc.ml
ggreif: Use an environment variable, like `//MOC-ENV MOC_UNLOCK_PRIM=yesplease`

ggreif (commented):
Looks good, but you can eliminate that ugly flag, I think.

Changelog.md
alexandru-uta: hey @ggreif , does your question link to the line you actually intended? that doesn't seem to have anything in common with the current PR.

src/exes/moc.ml
crusso: Tried, but turns ugly fast, so sticking to this solution

PR #5320, opened 2025-07-08
feat: migration diagnostic to enable switch to default persistence (all fields stable by default, regardless of `persistent` keyword)

crusso:
Builds on #5298 

This PR makes migration to default persistent actor safer by encouraging actors to migrate their non-persistent actors to unambiguous persistent actors. Later, the persistent keyword can be removed:

* with --require-persistent-actors  (current default flag)
   * in non-`persistent` actor:
      * error on implicitly transient declarations, requiring explicit `transient`
      * when no implicit declarations remains, i.e. all are declared `transient` or `stable`, error on abscence of `persistent` keyword.
   * in `persistent` actor, warn on use of redundant `stable` keyword

* with --default-persistent-actors (future default flag)
  * in every actor, warn on use of redundant `stable` keyword
  * in `persistent` actor, warn about redundant `persistent` keyword.
  
* with --legacy-actors:
  * allow persistent and non-persistent actors with legacy (dual) semantics.
 
* all flags are deliberately undocumented and should not be used expect by tests and caffeine

- [x] get it working
- [x] rename flags --legacy-actors, --require-persistent-actors, --default-persistent-actors and use just one (no diagnostics) 
- [x] clean up some mess in syntax.ml
- [x] undocument *all* flags. Caffeine can enable the undocumented flag until we enable it by default.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,6 +14,9 @@
     Users not willing or able to migrate their code can opt in to the behavior of moc prior to this release with the new flag `--legacy-persistence`.
     Flag `--legacy-persistence` is required to select the legacy `--copying-gc` (the previous default), `--compacting-gc`,  or `generational-gc`.
 
+    As a safeguard, to protect users from unwittingly, and irreversibly, upgrading from legacy to enhanced orthogonal persistence, such upgrades will fail unless the new code is compiled with flag `--enhanced-orthogonal-persistence` explicitly set.
+    New projects should not require the flag at all (#5308) and will simply adopt enhanced mode.
+
     To recap, enhanced orthogonal persistence implements scalable and efficient orthogonal persistence (stable variables) for Motoko:
     * The Wasm main memory (heap) is retained on upgrade with new program versions directly picking up this state.
     * The Wasm main memory has been extended to 64-bit to scale as large as stable memory in the future.
--- a/.github/workflows/nightly-macos-test.yml
+++ b/.github/workflows/nightly-macos-test.yml
@@ -47,8 +47,49 @@ jobs:
           test-name: macos-13-${{ matrix.build_type }}-tests
           cachix-auth-token: ${{ secrets.CACHIX_AUTH_TOKEN }}
 
+  motoko-core-tests:
+    needs: systems-go-tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get `motoko` Repo
+        uses: actions/checkout@v4
+      - name: Populate `moc`
+        uses: ./.github/actions/test-blueprint
+        with:
+          os: ubuntu-latest
+          test-target: moc
+          test-name: fetch-moc
+          cachix-auth-token: ${{ secrets.CACHIX_AUTH_TOKEN }}
+      - name: Build `moc`
+        run: |
+          nix build .#moc --print-out-paths
+      - name: Save the `test-blueprint` action
+        # This is needed by the "Post Populate `moc`" step (cleanup)
+        # but since the checkout of `motoko-core` destroys the `pwd` contents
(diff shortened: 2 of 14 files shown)


Changelog.md
```diff
@@ -2,6 +2,20 @@
...
+
+    In the near future, the `persistent` keyword will be made optional again, and `let` and `var` declarations within actor and actor classes will be `stable` (by default) unless declared `transient`, inverting the previous default for non-`persistent` actors.
+    The goal of this song and dance is to *always* default actor declarations to stable unless declared `transient` and make the `persistent` keyword redundant.
+    New flags:
+    * `--non-persistent`: rejects code with non-`persistent` actors emitting
+       diagnostics to force transition to `persistent` actors. Current default.
+    * `--persistent`: allows non-`persistent` actors but defaults to `stable`, not `transient`.
+       Future default.
+    * `--legacy-non-persistent`: compiles code with legacy semantics and no diagnostics.
+       For internal use and those that can't adjust their code just now.
```
christoph-dfinity: I don't think we should mention this flag, and ideally hide it from the help menu.

crusso: What if someone wants to use a library that hasn't been updated and they can't access?

crusso: I guess its just the MSFT in me...

christoph-dfinity: > What if someone wants to use a library that hasn't been updated and they can't access?

They'd need to fork and use their fork until they get the fix upstreamed. With the flag set they'd be stuck as well, as they couldn't upgrade their own code, right?

This change also doesn't "really" apply to libraries, right? (I haven't seen many libraries exporting actors)

crusso: Well, they could have an actor class, but yeah, not seen in the wild fer sure

crusso:
TFFT

Changelog.md
crusso: Done - if you leave the doc string empty, Ocaml's Arg module won't display the option at all.

test/run-drun/persistent-default/version3.mo
```diff
@@ -8,3 +8,4 @@ actor {
   Prim.debugPrint(debug_show {x;y;z});
 
 }
+
```
crusso:
```suggestion
```

src/pipeline/test_field_srcs.ml
```diff
@@ -147,18 +147,8 @@ let%expect_test "" =
...
-      (@@ test-field-srcs.mo (Pos 9 16) (Pos 9 20))
-    )
-    Sources table:
-    test-field-srcs.mo:2.9-2.15: test-field-srcs.mo:2.9-2.15
-    test-field-srcs.mo:3.17-3.21: test-field-srcs.mo:3.17-3.21
-    test-field-srcs.mo:8.9-8.15: test-field-srcs.mo:8.9-8.15
-    test-field-srcs.mo:9.17-9.21: test-field-srcs.mo:3.17-3.21 test-field-srcs.mo:9.17-9.21
-    test-field-srcs.mo:14.15-14.19: test-field-srcs.mo:14.15-14.19 |}]
+    Diagnostics:
+    test-field-srcs.mo:1.1-1.6: type error [M0220], this actor or actor class should be declared `persistent` |}]
```
christoph-dfinity: This doesn't look intentional?

christoph-dfinity (changes requested):
Very nice, thank you! Just spotted one test diff that looks suspicious.

src/pipeline/test_field_srcs.ml
crusso: Fixed I think. Missing persistent keyword

crusso: Done

PR #5322, opened 2025-07-09
docs: Motoko docs follow-up tasks

jessiemongeon1:
Completes tickets in Jira epic: https://dfinity.atlassian.net/browse/DD-673


doc/md/fundamentals/1-basic-syntax/1-defining-an-actor.md
```diff
@@ -73,6 +73,7 @@ service : {
 }
 ```
 
+
```
christoph-dfinity, line 76: Was this intentional? From what I can tell we usually only have one newline before new sections

doc/md/fundamentals/4-declarations/7-module-declarations.md
```diff
@@ -139,5 +139,24 @@ import {init; identity} "Matrix";
...
+    };
+};
+
+// Creating different currency converters
+let usdConverter = ExchangeRate(1.1);
+let eurConverter = ExchangeRate(0.9);
+
+Debug.print(debug_show(Float.toText(usdConverter.convert(100))));  // "110.0"
+Debug.print(debug_show(Float.toText(eurConverter.convert(100))));  // "90.0"
+```
```
christoph-dfinity: I'm very confused about this one. I don't think a "module class" is a thing. The grammar allows it, but isn't this just the same as a `class` without the `module`? @crusso

christoph-dfinity (commented):
Changes look good! Just had one question and then I'm happy to approve.

doc/md/fundamentals/4-declarations/7-module-declarations.md
crusso: No it's a thing. A plain class is short for an object class that produces objects. Objects and modules are very slightly different.  The main difference is that you can import a module and the contents must be static. Tbh I'm not sure the static restriction applies to non-imported modules would need to check. Another difference is that modules are never shareable, but objects can be depending on the types of their fields.

doc/md/fundamentals/3-types/12-advanced-types.md
```diff
@@ -201,3 +201,32 @@ But the function can also be used to return the maximum of two `Nat`s and still
 max<Nat>(5, 10); // returns 10 : Nat
 ```
 
+## Actor type annotation
```
crusso:
````suggestion
## Actor reference expression


The *actor reference* expression `actor <exp>` compute a reference to an actor from a text argument `<exp>`, the textual encoding of a canister id. The expression is typically combined with an (actor) type annotation,  `actor <exp> : <typ>`  that declares the expected type of the actor reference. The argument  can either be a text literal, identifier, or, when enclosed in parentheses, a more complicated expression that computes a textual id.

A simple example of using actor references is to access the management canister with textual address `"aaaaa-aa"`. Amongst other things, it has a method `raw-rand` for generating cryptographically random bytes as a `Blob`.

```motoko
persistent actor Coin {
  public func flip() : async Bool {
    let managementCanister = actor "aaaaa-aa" : actor { raw_rand : () -> async Blob };
    let entropy = await managementCanister.raw_rand();
    (entropy[0] & 1) == 1;
  }; 
}
```

A variation computes the textual canister identifier from a given principal. A call to `flipWith(p)` will succeed is called with `Principal.fromBlob("aaaaa-aa")`, but may fail with another argument, if the canister does not exist or does not have a `raw_rand` function:

```motoko
import Principal "mo:base/Principal";

persistent actor Coin {
  public func flipWith(principal : Principal) : async Bool {
    let canister = actor (Principal.toText(principal)) : actor { raw_rand : () -> async Blob };
    let entropy = await canister.raw_rand();
    (entropy[0] & 1) == 1;
  }; 
} 
```

:::warn

There is currently no way to verify the actual type of the actor is compatible with the declared type. The validity of the actor reference and any type incompatibility will be detected later, on interaction with the actor.
For this reason, you should only use actor references sparingly. It's typically safer to use explicitly imported canisters with their given actor types and avoid using `Text` or `Principal` to represent canisters in your methods (just use actor types instead).

For example, a safer variant of `flipWith` is:

```motoko no-repl
persistent actor Coin {
   public func flipWith(canister : actor { raw_rand : () -> async Blob }) : async Bool { 
       let entropy = await canister.raw_rand();
       (entropy[0] & 1) == 1;
  }; 
} 
```

This `flipWith` takes a strongly-typed actor, not a weakly-typed `Principal`.

````

I don't think this section should be here but under expressions (not advanced types).

I think with these simpler examples, the following Publisher example can be removed but we can also keep it if you prefer

File: doc/md/fundamentals/4-declarations/7-module-declarations.md (hunk 2 above)
Comment (crusso) [reply doc/md/fundamentals/4-declarations/7-module-declarations.md]: Example:
```
persistent actor {
  module class F(){ public let x = 1};
  transient let M : module {} = F();
} 
```
and it looks like the body of F() must be static

File: doc/md/fundamentals/5-control-flow/1-basic-control-flow.md [hunk 4]
```diff
@@ -139,58 +154,68 @@ If the block produces a non-`()` result, the `break` can include a value. Labels
 When a labeled block runs, it evaluates the block to produce a result. Labels don’t change how the block executes but enable early exits from the block using a `break` to that label. If the type is not `()` those breaks must have an argument, to use as the result of the labelled expression. Just as `return` exits a function early with a result, `break` exits its label early with a result.
 
 ```motoko no-repl
```
crusso, line 151: These examples are a bit weird. Rather than suggest the changes, I'll open a branched PR to make my edits for you to approve and merge into this branch

crusso:
@jessiemongeon1 are you ok with #5331? If so, I can merge it into this PR and then merge this PR into master

crusso:
I'll take another look tomorrow (on mobile now)

doc/md/fundamentals/3-types/12-advanced-types.md
```diff
@@ -201,3 +201,80 @@ But the function can also be used to return the maximum of two `Nat`s and still
 max<Nat>(5, 10); // returns 10 : Nat
 ```
 
+## Actor reference expression
+
+
+The *actor reference* expression `actor <exp>` compute a reference to an actor from a text argument `<exp>`, the textual encoding of a canister id. The expression is typically combined with an (actor) type annotation,  `actor <exp> : <typ>`  that declares the expected type of the actor reference. The argument  can either be a text literal, identifier, or, when enclosed in parentheses, a more complicated expression that computes a textual id.
+
+A simple example of using actor references is to access the management canister with textual address `"aaaaa-aa"`. Amongst other things, it has a method `raw-rand` for generating cryptographically random bytes as a `Blob`.
```
crusso:
```suggestion
A simple example of using actor references is to access the management canister with textual address `"aaaaa-aa"`. Amongst other things, it has a method `raw_rand` for generating cryptographically random bytes as a `Blob`.
```

doc/md/fundamentals/3-types/12-advanced-types.md
```diff
@@ -201,3 +201,80 @@ But the function can also be used to return the maximum of two `Nat`s and still
...
+       (entropy[0] & 1) == 1;
+  }; 
+} 
+```
+
+This `flipWith` takes a strongly-typed actor, not a weakly-typed `Principal`.
+
+
+Actor type annotations offer flexibility when working with external canisters, but there’s no guarantee that the function signatures will match at runtime. If the signatures don’t align, the calls will fail.
+
```
crusso, line 255:
```suggestion

As another example of using actor reference expressions, we present a simple `Publisher` actor that tracks sets of subscribers by their `Principal` and uses an actor reference expression access the `notify` of each subscriber when a message is `published`:

<!-- TODO: improve or remove sample (e.g. use a set of subscribers and oneway notify that returns `()` not `async ()` -->

```

doc/md/fundamentals/4-declarations/7-module-declarations.md
```diff
@@ -139,5 +139,24 @@ import {init; identity} "Matrix";
 let zero = init(3, 3, 0);
 let id4 = identity(4);
 ```
-<!-- TODO (future): short section on module classes, perhaps for square (n*n) matrices? -->
 
+### Module classes
+
+A module class can be used to produce multiple modules with different configurations.
+
```
crusso, line 146:
```suggestion

<!-- TODO: a better example would have type fields, e.g. a simple vector lib -->

```

crusso (approved):

PR #5339, opened 2025-07-17
experiment: replay Claudio's merge fixes on top of a proper merge commit between master and stable-functions

christoph-dfinity:


christoph-dfinity:
@crusso I think I incorporated your changes and gotten this to a point where it passes CI. The accepted changes in d7b3409ab0c3c2b10e9b6bf04ccd73d3f8590bbf look like regressions, maybe we can take a look at them together and decide if they're introduced by the merge or already exist in the original PR.

PR #5354, opened 2025-07-25
WIP: stable function upgrade examples

christoph-dfinity:


test/run-drun/brittle-stable-function.mo
```diff
@@ -0,0 +1,14 @@
+//ENHANCED-ORTHOGONAL-PERSISTENCE-ONLY
+//SKIP ic-ref-run
+import Clamper "Clamper";
```
crusso: Is this file missing from the PR?

christoph-dfinity: This was a leftover file. I removed it in favor of `brittle-stable-functions.drun`

test/run-drun/brittle-stable-functions/RenamedHack.mo
```diff
@@ -0,0 +1,18 @@
+persistent actor {
+  class Clamper(l : Nat, h : Nat) {
+    public func clamp(x : Nat) : Nat {
+      // When trying to evolve this class to hardcode the upper limit,
```
crusso, line 4: That's odd, I'd expect the loss of a free variable to be tolerated. Maybe this a bug rather than design flaw

From the PR description:

```
On a stable function upgrade, the closure type of the stable function must remain compatible:

The new version of the stable function does not capture more variables/parameters than the previous version.
The captured variable in the new version is a valid super-type of the previous version.
```

test/run-drun/brittle-stable-functions/Renamed.mo
```diff
@@ -0,0 +1,15 @@
+persistent actor {
+  class Clamper(low : Nat, high : Nat) {
```
crusso, line 2: I can't tell from the log if the upgrade from Initial to Renamed works or not. I suspect it should from the PR description, but does it fail?

The closures of stable functions are represented in a portable format with the following clearly defined binding of captured variables:

```
Local variables are captured by name.
Parameters are captured by position (also considering nested functions).
```
I guess some printfs might help

PR #5359, opened 2025-07-26
kamil/invariant bounds choice

Kamirus:
- **try infer lambda body**
- **exclude subtypes of unit**
- **working tests with problematic cases ok too**
- **use Diag.with_message_store**
- **problematic breaking change with let**
- **plan**
- **infer_subargs_for_bimatch_or_defer**
- **full cannot infer pat**
- **use, desugar problem with lingering Pre**
- **fix exp.note of decomposed expressions to avoid Pre types there**
- **missing bi match changes**
- **fix cannot_infer_pat**
- **update lambdas.mo**
- **unwrap named type**
- **refactor**
- **refactor**
- **ok and problematic map examples**
- **Track unused tvars in bi_match and avoid checking deferred when they contain the unused**
- **2nd round of bimatch to infer bodies of deferred funcs**
- **fix**
- **comments**
- **style lambdas-map.mo**
- **comment out all prints**
- **fix unused identifier bug**
- **refactor**
- **cleanup and cos in bi_match**
- **cleanups and docs**
- **fix**
- **fix open types in deferred lambda input**
- **decomposition changes**
- **tests**
- **tests**
- **fw**
- **lambdas-core and dirty hacks to make it work**
- **fix test**
- **wrong assert for async scope case, just remove**
- **err_subst**
- **refactor the `ctx` in bi_match**
- **refactor splitting the ctx**
- **refactor**
- **cleanups**
- **fix the test, preserve the order of variables in the error message**
- **refactor + preserve named types in errors**
- **cleanups**
- **refactor**
- **found bug or slight problem**
- **fix: use must_fix and can_defer**
- **refactor and debug prints**
- **reenable verification**
- **refactor**
- **fix filterDom, mapi2 tailrec, typesafe no 2nd round when not needed**
- **experiment: pick non-trivial invariant bound**


src/mo_frontend/bi_match.ml
```diff
@@ -107,12 +107,33 @@ let fail_open_bound c bd =
...
+let try_choose_invariant lb ub =
+  (* When there is exactly one non-trivial bound, choose it as the solution.
+   * Note that this is not a principal solution! A user might need to fix it by type annotation or explicit instantiation.
+   * However, without this strategy the user ALWAYS needs to supply a type annotation or explicit instantiation.
+   *)
+  (* TODO: more tests: try to push it to create test cases that create confusing errors when this strategy is used,
+   e.g. it causes errors later when coding a function, when the user forgot that the compiler made some not optimal choices
+  in these cases in might be better without this strategy, ask about type annotation early to avoid confusion later. *)
+  (* TODO: Can we restrict the types when this strategy can be applied?
+    Is it reasonable to pick a solution for some types and not for others?
```
crusso: My hunch was to only do this when the type has no non-trivial lower and upper bounds. So for (unbound T), Principal, Blob, Bool, Char, Text, NatXX, IntXX, [var T], etc but *not* bound T,  [T], null, ?T,  Int, Nat, records or variants since these have non-trivial super/subtypes. Bit of a hack though.

Kamirus: Ah now I see what you've meant. The first group of types is reasonable to accept, but why not to include `[T]` and `?T` as well? Is there a practical example where it would choose a non principal solution? All of {Principal, Blob, Bool, Char, Text, NatXX, IntXX, [var T]} are compatible only with itself, yes? `X <: Y` implies `X=Y` (when X or Y comes from the set) right? If so then `[T]` could also be included. 
I know that `Null <: ?T` so it does not apply to `?T`, but honestly I'd ignore that `Null` type even exists... I don't think there is a practical reason where user wanted to infer `Null` instead of `?T`...

I'd honestly include `Nat`, `Int` and records because:
- We already default to `Nat` in definitions like `let x = [var 1]`, so users should be used to fixing such definitions with an explicit type annotation, e.g. `[var Int]`
- We already have warnings in place that warn when record fields are implicitly dropped, so when considering a solution for `{x;y;z} <: invariant X <: Any` I'd not consider dropping any of the fields. IMO it is unlikely that user actually meant a subset like `{x;z}` as the solution.

crusso: Sorry, I was lost in your other PR ;->.

[T] is actually an immutable array, not mutable, so actually variant and  [T] < [U] does not imply T = U, e.g. [Nat] <: [Int].

Similarly ?T is variant so ?T <: ?U does not imply T = U. It makes a difference whether you infer [var ?Int] or [var ?Nat] and  the user might no appreciate it when we get it wrong.

If you think of the use of record to encode objects, the class ColoredPoint <: Point  then I think its not unreasonable that user might expect a mutable variable to store Points, not ColoredPoint. But, yeah I agree we've already messed up there with mutable arrays and unannotated vars.

Kamirus: > All of {Principal, Blob, Bool, Char, Text, NatXX, IntXX, [var T]} are compatible only with itself, yes? `X <: Y` implies `X=Y` (when X or Y comes from the set) right? If so then `[T]` could also be included.

Sorry I think I did not explain that precise enough 😄 Let me try again:
I claim that types restricted to `{Principal, Blob, Bool, Char, Text, NatXX, IntXX, [var T]}` are not variant. We cannot implicitly coerce other arbitrary type `X` to any of these restricted ones, nor coerce to `X`. Because, e.g. `Nat32` **cannot** be used where `Int32` is needed, and so on. `Nat` and `Int` are **not** in this set btw.
And actually adding immutable `[T]` to this restricted set does not change the property.
Sure `[T1] <: [T2]` is when `T1 <: T2` but actually this only happens when `T1 = T2` because these types are restricted to this set above.
So it is safe to add immutable arrays.

It is not safe to add `?T`, because the following coercion is possible `Null <: ?T`. But personally I would ignore existence of the `Null` type 😄 This should be always ok in practice, because when would you actually want to infer `Null` instead of `?T`

Sure, records are even less safe. Yes you are right that this `ColoredPoint <: Point` might be desired.

But I think we can enable this PR's mechanism for the restricted set of types built from the `{Principal, Blob, Bool, Char, Text, NatXX, IntXX, [var T], [T]}` and we won't loose principal solution, because these types "are invariant".

And if we include `?T` too, we might loose the principal property but only for impractical cases.

crusso:
At first glance this looks very nice! How do you feel about it?

Kamirus:
> At first glance this looks very nice! How do you feel about it?

I feel good about it, I think this is a good change 😄

src/mo_types/type.ml
```diff
@@ -1479,6 +1479,37 @@ and singleton_field co tf = singleton_typ co tf.typ
 
 and singleton t : bool = singleton_typ (ref S.empty) t
 
+(** A type is isolated if it has no proper supertypes nor proper subtypes (ignoring top [Any] and bottom [Non]). *)
+let rec isolated t =
+  match normalize t with
```
crusso: Actually this is not complete, I think you will need to unfold definitions and assume isolated unless proven otherwise.



For type parameters (Abs) parameters with bound Any are isolated, otherwise note.

On mobile so hard to check code

I think the is_isolated check might fail to terminate on, e.g.

type Bot  = [Bot]

src/mo_types/type.ml
```diff
@@ -1479,6 +1479,42 @@ and singleton_field co tf = singleton_typ co tf.typ
...
+    | Principal
+    | Region
+    (* All except Nat, Int, Null as they have proper super/subtypes: Nat <: Int, ?T <: Null *)
+    )
+  | Mut _ -> true
+  | Array t
+  | Async (_, _, t)
+  | Weak t -> isolated_typ co t
+  | Tup ts -> List.for_all (isolated_typ co) ts
+  | Func (_, _, _, ts1, ts2) ->
```
crusso, line 1521: I think you still want a case for  a type parameter (Con(t, ts) with kind c = Abs). This can be isolated iff the bound is Any (which is common) (I think).

I guess that means that

func box<T>( v : T) {
   var r = Box(v);
}
would work.
But

func box<T <: Nat>( v : T) {
   var r = Box(v);
}
would not.

Kamirus: Ok I've added unbounded type parameters as isolated types

christoph-dfinity (approved):
Code changes look good. I think I can follow the reasoning and motivation as well, so LGTM!

PR #5362, opened 2025-07-28
experiment: type field bug

crusso:


christoph-dfinity:
Added this test to the PR. Thanks

PR #5363, opened 2025-07-28
feat: Weak References in Motoko

alexandru-uta:
First implementation of Weak References in Motoko with GC and type system support. The main functionality so far is to dereference a weak reference and check if the object pointed at by it is still live (i.e., wasn't garbage collected).

So far, the API exposes the following:
1. Create a weak reference to an object, via `weakRef = Prim.allocWeakRef(object)`.
2. Dereference a weak reference, via `object = Prim.getWeak()`.
3. Check if a target of a weak reference is still alive, via `Prim.isLive(weakRef)`.

This all works by adding a new type in the compiler and a new tag type in the RTS for tracking everything. The most non-intrusive change that allowed for achieving the aforementioned functionality was to add: (1) a new data structure that collects all the Weak References during marking; and (2) a new GC step immediately after marking that goes through this Weak Reference data structure and checks whether objects they point to are still live. If not, the weak reference pointer is updated to be NULL.

The functionality of the PR covers the difficult case of upgrades that may fall within a GC cycle by ensuring that the metadata layout is compatible even with adding a new data structure.


alexandru-uta:
@crusso , this kind of code doesn't compile:

```
persistent actor {

  var ref : Weak[var Nat64] = Prim.allocWeakRef(Prim.Array_init<Nat64>(3, 1));

  public func test() : async () {

  };
};
```

rts/motoko-rts/src/types.rs
```diff
@@ -529,6 +529,9 @@ pub const TAG_ONE_WORD_FILLER: Tag = 41;
 #[enhanced_orthogonal_persistence]
 pub const TAG_FREE_SPACE: Tag = 43;
 
+#[enhanced_orthogonal_persistence]
+pub const TAG_WEAK_REF: Tag = 45;
```
luc-blaeser, line 533: The `TAG_ARRAY_SLICE_MIN` below would need to be increased to 46.

rts/motoko-rts/src/gc/incremental/phases/mark_increment.rs
```diff
@@ -49,6 +57,8 @@ impl<'a, M: Memory + 'a> MarkIncrement<'a, M> {
     pub unsafe fn mark_completed(state: &State) -> bool {
         let mark_state = state.mark_state.as_ref().unwrap();
         debug_assert!(!mark_state.complete || mark_state.mark_stack.is_empty());
+        #[cfg(feature = "enhanced_orthogonal_persistence")]
+        debug_assert!(mark_state.weak_ref_registry.is_empty());
```
luc-blaeser: I think it should be:
```suggestion
        debug_assert!(!mark_state.complete || mark_state.weak_ref_registry.is_empty());
```
Weak registry can be non-empty during mark phase.

rts/motoko-rts/src/gc/incremental/phases/mark_increment.rs
```diff
@@ -145,6 +171,27 @@ impl<'a, M: Memory + 'a> MarkIncrement<'a, M> {
...
+            // Process all weak references collected during marking.
+            // If the target object is not marked, clear the weak reference.
+            loop {
+                let weak_ref_value = self.weak_ref_registry.pop();
+                if weak_ref_value == STACK_EMPTY {
+                    break;
+                }
+                let weak_ref_obj = weak_ref_value.get_ptr() as *mut crate::types::WeakRef;
+                let target = (*weak_ref_obj).field;
+                if target.is_non_null_ptr() && target.get_ptr() >= self.heap.base_address() {
```
luc-blaeser: I believe `&& target.get_ptr() >= self.heap.base_address()` is not needed

rts/motoko-rts/src/gc/incremental/phases/update_increment.rs
```diff
@@ -102,7 +102,12 @@ impl<'a> UpdateIncrement<'a> {
             // Keep mark bit and later resume updating more slices of this array.
             return;
         }
+        #[enhanced_orthogonal_persistence]
+        debug_assert!(object.tag() < TAG_ARRAY_SLICE_MIN || object.tag() == TAG_WEAK_REF);
+
+        #[classical_persistence]
```
luc-blaeser: `TAG_ARRAY_SLICE_MIN` would need to be adjusted to cover `TAG_WEAK_REF` (left a comment below).
```suggestion
```

rts/motoko-rts/src/gc/incremental/phases/update_increment.rs
```diff
@@ -124,8 +129,20 @@ impl<'a> UpdateIncrement<'a> {
                     length
                 },
             );
-            if object.tag() < TAG_ARRAY_SLICE_MIN || self.time.is_over() {
-                return;
+            #[cfg(feature = "classical_persistence")]
```
luc-blaeser: Same thing here: I believe original check is fine. Just `TAG_ARRAY_SLICE_MIN` needs to be increased.

rts/motoko-rts/src/memory.rs
```diff
@@ -83,3 +83,42 @@ pub unsafe fn alloc_array<M: Memory>(mem: &mut M, tag: Tag, len: usize) -> Value
...
+#[ic_mem_fn]
+pub unsafe fn alloc_weak_ref<M: Memory>(mem: &mut M, target: Value) -> Value {
+    use crate::barriers::allocation_barrier;
+
+    let weak_ref = mem.alloc_words(size_of::<WeakRef>() + Words(16));
+
+    let weak_ref_obj = weak_ref.get_ptr() as *mut WeakRef;
+    (*weak_ref_obj).header.tag = TAG_WEAK_REF;
+    (*weak_ref_obj).header.init_forward(weak_ref);
+    (*weak_ref_obj).field = target;
```
luc-blaeser: Here a barrier is needed (doing fwd resolution of target)

```suggestion
    init_with_barrier(mem, &mut (*weak_ref_obj).field, target);
```

rts/motoko-rts/src/memory.rs
```diff
@@ -83,3 +83,42 @@ pub unsafe fn alloc_array<M: Memory>(mem: &mut M, tag: Tag, len: usize) -> Value
...
     skewed_ptr
 }
+
+/// Allocate a new weak reference.
+#[enhanced_orthogonal_persistence]
+#[ic_mem_fn]
+pub unsafe fn alloc_weak_ref<M: Memory>(mem: &mut M, target: Value) -> Value {
+    use crate::barriers::allocation_barrier;
+
+    let weak_ref = mem.alloc_words(size_of::<WeakRef>() + Words(16));
```
luc-blaeser: Not sure about the Words(16)?
I think `motoko_rts::types::size_of::<WeakRef>()` is the right approach. This counts the words. 
(I know the naming is suboptimal here, a relict).

rts/motoko-rts/src/memory.rs
```diff
@@ -83,3 +83,42 @@ pub unsafe fn alloc_array<M: Memory>(mem: &mut M, tag: Tag, len: usize) -> Value
...
+    // TODO: double check this!!!
+    allocation_barrier(weak_ref)
+}
+
+/// Check if a weak reference is still live.
+#[enhanced_orthogonal_persistence]
+#[ic_mem_fn]
+pub unsafe fn weak_ref_is_live<M: Memory>(_mem: &mut M, weak_ref: Value) -> bool {
+    if !weak_ref.is_non_null_ptr() {
+        return false; // Invalid WeakRef pointer.
```
luc-blaeser: This would be bug? If so, I would suggest a trap/panic. or non-debug assert

rts/motoko-rts/src/memory.rs
```diff
@@ -83,3 +83,42 @@ pub unsafe fn alloc_array<M: Memory>(mem: &mut M, tag: Tag, len: usize) -> Value
...
+    let field_tag = (*weak_ref_obj).header.tag;
+
+    match field_tag {
+        TAG_WEAK_REF => {
+            // Get the target object and check if it is null.
+            let target = (*weak_ref_obj).field;
+            return target.is_non_null_ptr();
+        }
+        _ => {
+            return false;
```
luc-blaeser: Also this would be probably be a bug (case for trap)?

luc-blaeser: Nit: Could be maybe put in a impl function of the WeakRef struct

rts/motoko-rts/src/visitor/enhanced.rs
```diff
@@ -42,6 +42,12 @@ pub unsafe fn visit_pointer_fields<C, F, G>(
             }
         }
 
+        TAG_WEAK_REF => {
+            // WeakRef objects themselves are visited (so they don't get collected)
+            // but we deliberately do NOT visit the 'field' to avoid marking
+            // the weakly referenced object as live
```
luc-blaeser: Visiting the pointer field is important to update its target (when alive) to new relocated location during GC update phase.

luc-blaeser: In marking, it needs be skipped though

rts/motoko-rts/src/gc/incremental/phases/mark_increment.rs
```diff
@@ -111,12 +123,26 @@ impl<'a, M: Memory + 'a> MarkIncrement<'a, M> {
...
+                use crate::types::is_weak_ref_tag;
+                let tag = value.as_obj().tag();
+                if is_weak_ref_tag(tag) {
+                    // Collecting the weak references here ensures that
+                    // no weak reference is collected twice.
+                    // That is because the mark_object() primitive above
+                    // ensures that we do not mark the same object twice.
+                    self.weak_ref_registry.push(self.mem, value);
+                }
+            }
```
luc-blaeser, line 146: Comment for line below:
The mark check should exempt TAG_WEAK_REF:
```
let tag = (*array).header.tag;
if tag >= TAG_ARRAY_SLICE_MIN && tag != TAG_WEAK_REF {
   ... // mark
```

luc-blaeser:
Impressive work in such short amount of time, 5 days!
Just a few tweaks and I believe this PR is good to go

rts/motoko-rts/src/memory.rs
alexandru-uta: Ah, yes, the Words(16) is a debugging leftover, thanks for catching it!

rts/motoko-rts/src/visitor/enhanced.rs
alexandru-uta: I think it should be fixed now, please check.

rts/motoko-rts/src/gc/incremental/phases/mark_increment.rs
alexandru-uta: Not sure I understand this, can you please explain?

rts/motoko-rts/src/memory.rs
alexandru-uta: yup, done.

rts/motoko-rts/src/gc/incremental/partitioned_heap.rs
```diff
@@ -690,7 +690,7 @@ impl PartitionedHeap {
         true
     }
 
-    #[cfg(debug_assertions)]
+    //#[cfg(debug_assertions)]
```
luc-blaeser:
```suggestion
```

rts/motoko-rts/src/gc/incremental/phases/mark_increment.rs
luc-blaeser: Already addressed :)

rts/motoko-rts/src/memory.rs
```diff
@@ -83,3 +86,30 @@ pub unsafe fn alloc_array<M: Memory>(mem: &mut M, tag: Tag, len: usize) -> Value
...
+    allocation_barrier(weak_ref)
+}
+
+/// Check if a weak reference is still live.
+#[enhanced_orthogonal_persistence]
+#[ic_mem_fn]
+pub unsafe fn weak_ref_is_live<M: Memory>(_mem: &mut M, weak_ref: Value) -> bool {
+    if !weak_ref.is_non_null_ptr() {
+        crate::rts_trap_with(
+            "weak_ref_is_live: Invalid WeakRef pointer. This is a bug, report tothe Motoko team.",
```
luc-blaeser:
```suggestion
            "weak_ref_is_live: Invalid WeakRef pointer. This is a bug, report to the Motoko team.",
```

src/mo_types/arrange_type.ml
```diff
@@ -92,7 +92,8 @@ module Make (Cfg : Config) = struct
     | Non                    -> Atom "Non"
     | Pre                    -> Atom "Pre"
     | Typ c                  -> "Typ" $$ [con c]
-    | Named (n, t)            -> "Name" $$ [Atom n; typ t]
+    | Named (n, t)           -> "Name" $$ [Atom n; typ t]
+    | Weak t                 -> "weak" $$ [ typ t]
```
crusso:
```suggestion
    | Weak t                 -> "Weak" $$ [ typ t]
```

(not important, just for consistency. this is just debug spew)

src/mo_values/prim.ml
```diff
@@ -442,4 +442,22 @@ let prim trap =
       fun _ v k -> k (Tup [
         Nat32 Numerics.Nat32.zero; Nat32 Numerics.Nat32.zero; Nat32 Numerics.Nat32.zero])
 
+  | "alloc_weak_ref" ->
+     fun _ v k ->
+       let w = Weak.create 1 in
```
crusso, line 448:
```suggestion
       (* TODO: model trapping on non-scalars *)
       let w = Weak.create 1 in
```

src/codegen/compile_classical.ml
```diff
@@ -11783,6 +11783,26 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
+    let msg = Diag.error_message Source.no_region "weak_get" "classical" 
+    "Weak references are not supported in classical mode." in
+    Diag.print_messages [msg];
+    exit 0
+
+  | OtherPrim "weak_ref_is_live", [weak_ref] ->
+    let msg = Diag.error_message Source.no_region "weak_ref_is_live" "classical" 
+    "Weak references are not supported in classical mode." in
+    Diag.print_messages [msg];
+    exit 0
```
crusso:
```suggestion
  (* Weak refs are disallowed in classical mode *)
  (* The compiler will exit with an error if it encounters a related call *)
  | OtherPrim "alloc_weak_ref", [target] ->
    let msg = Diag.error_message Source.no_region "alloc_weak_ref" "classical" 
      "Weak references are not supported in classical mode."
    in
    Diag.print_messages [msg];
    exit 0

  | OtherPrim "weak_get", [weak_ref] ->
    let msg = Diag.error_message Source.no_region "weak_get" "classical" 
      "Weak references are not supported in classical mode." 
    in
    Diag.print_messages [msg];
    exit 0

  | OtherPrim "weak_ref_is_live", [weak_ref] ->
    let msg = Diag.error_message Source.no_region "weak_ref_is_live" "classical" 
      "Weak references are not supported in classical mode." 
    in
    Diag.print_messages [msg];
    exit 0
```

Does it work as expected?

alexandru-uta: yup, all good. thanks for the formatting.

rts/motoko-rts/src/persistence.rs
```diff
@@ -269,3 +281,53 @@ pub unsafe extern "C" fn buffer_in_32_bit_range() -> usize {
...
+    // No barrier needed, the lifetime of the whole weak reference registry
+    // is just during the GC marking phase. After this it can be garbage collected.
+    // Every GC phase will reinitialize the weak reference registry.
+    let markstack_blob = alloc_blob(mem, TAG_BLOB_B, Bytes(size_of::<MarkStack>()));
+    let markstack_ptr = markstack_blob.as_blob_mut() as *mut MarkStack;
+    // Initialize the MarkStack object inside this pointer.
+    *markstack_ptr = MarkStack::new(mem);
+
+    let metadata = PersistentMetadata::get();
+    (*metadata).weak_ref_registry = Value::from_ptr(markstack_ptr as usize);
```
luc-blaeser, line 321: Maybe just a quick comment that write barrier is not needed because it is transitional (no marking needed) and not moved (no forwarding needed).

luc-blaeser: (Or, for consistency, use `initialize_with_barrier`)

rts/motoko-rts/src/persistence.rs
```diff
@@ -269,3 +281,53 @@ pub unsafe extern "C" fn buffer_in_32_bit_range() -> usize {
...
+}
+
+/// Clear the weak reference registry in persistent metadata.
+/// This is done only after the marking phase is finished.
+pub(crate) unsafe fn clear_weak_ref_registry() {
+    let metadata = PersistentMetadata::get();
+    (*metadata).weak_ref_registry = NULL_POINTER;
```
luc-blaeser, line 328: Similar comment here: Maybe a comment saying that no write barrier is needed (as object is transitional and thus marking of previous weak ref object is not needed)

src/mo_types/arrange_type.ml
alexandru-uta: yup, fixed now

rts/motoko-rts/src/persistence.rs
alexandru-uta: added comment

rts/motoko-rts/src/persistence.rs
alexandru-uta: added comment

src/codegen/compile_enhanced.ml
```diff
@@ -1949,6 +1952,7 @@ module Tagged = struct
     | OneWordFiller -> 41L
     | FreeSpace -> 43L
     | ArraySliceMinimum -> 44L
```
luc-blaeser: This should be 46L
```suggestion
    | ArraySliceMinimum -> 46L
```

src/codegen/compile_enhanced.ml
```diff
@@ -6863,7 +6899,10 @@ module Serialization = struct
           add_idx f.typ
         ) fs
       | Mut t ->
-        add_sleb128 idl_alias; add_idx t
+         add_sleb128 idl_alias; add_idx t
+      | Weak t -> (* TODO  - probably want another idl_type, idl_weak, for Weak, otherwise we can
+         deserialize one as the other *)
+         add_sleb128 idl_alias; add_idx (Weak t)
```
luc-blaeser: Do we need to serialize the weak ref? It is not used in cross-canister messages and not in classical persistence IIUC

src/codegen/compile_enhanced.ml
```diff
@@ -6966,6 +7005,8 @@ module Serialization = struct
         get_x ^^ Tagged.load_tag env ^^ clear_array_slicing ^^ set_tag ^^
         (* Sanity check *)
         get_tag ^^ compile_eq_const Tagged.(int_of_tag StableSeen) ^^
+        get_tag ^^ compile_eq_const Tagged.(int_of_tag WeakRef) ^^
+        G.i (Binary (Wasm_exts.Values.I64 I64Op.Or)) ^^
```
luc-blaeser, line 7039: Not sure we need this, maybe a check that `WeakRef` does not occur here

src/codegen/compile_enhanced.ml
```diff
@@ -7061,7 +7102,9 @@ module Serialization = struct
           inc_data_size (compile_unboxed_const 12L) ^^ (* |id| + |page_count| = 8 + 4 *)
           get_x ^^ Region.vec_pages env ^^ size env blob)
       | Mut t ->
-        size_alias (fun () -> get_x ^^ MutBox.load_field env ^^ size env t)
+         size_alias (fun () -> get_x ^^ MutBox.load_field env ^^ size env t)
+      | Weak t ->
+         size_alias (fun () -> get_x ^^ WeakRef.load_field env ^^ size env t)
```
luc-blaeser: Dito

src/codegen/compile_enhanced.ml
alexandru-uta: fixed

src/codegen/compile_enhanced.ml
```diff
@@ -1948,7 +1951,8 @@ module Tagged = struct
     | Region -> 39L
     | OneWordFiller -> 41L
     | FreeSpace -> 43L
-    | ArraySliceMinimum -> 44L
+    | ArraySliceMinimum -> 46L
+    | WeakRef -> 45L
```
luc-blaeser: Nit: 
```suggestion
    | WeakRef -> 45L
    | ArraySliceMinimum -> 46L
```

src/codegen/compile_enhanced.ml
```diff
@@ -2158,6 +2162,35 @@ module MutBox = struct
     E.object_pool_add env __LINE__ alloc
 end
 
+module WeakRef = struct
+  (*
+      Weak references
+
+       ┌──────┬─────┬─────────┐
+       │ obj header │ payload │
```
luc-blaeser:
```suggestion
       │ obj header │ field │
```

src/codegen/compile_enhanced.ml
```diff
@@ -2158,6 +2162,35 @@ module MutBox = struct
...
+       └──────┴─────┴─────────┘
+
+     The object header includes the obj tag (Weak) and the forwarding pointer.
+  *)
+
+  let field = Tagged.header_size
+
+  let alloc env =
+    Tagged.obj env Tagged.WeakRef [ compile_unboxed_zero ]
+
```
luc-blaeser: I believe this is not used
```suggestion
```

src/codegen/compile_enhanced.ml
```diff
@@ -2158,6 +2162,35 @@ module MutBox = struct
...
+    Tagged.load_forwarding_pointer env ^^
+    Tagged.load_field env field
+
+  let store_field env =
+    let (set_weak_value, get_weak_value) = new_local env "weak_value" in
+    set_weak_value ^^
+    Tagged.load_forwarding_pointer env ^^
+    get_weak_value ^^
+    Tagged.store_field env field
+
```
luc-blaeser: Also redundant IIUC
```suggestion
```

src/codegen/compile_enhanced.ml
crusso: Probably not. I was working under the assumption that we would add them to the classic compiler too, eventually, in which case we would need destabilization logic. But safe to remove for now.

src/codegen/compile_enhanced.ml
crusso: Yes, we can delete this if unused.

src/codegen/compile_enhanced.ml
crusso: yes we can delete if unused.

src/codegen/compile_enhanced.ml
crusso: Hmm, so weak ref's are stable and we always generate deserialization code, even if we don't use it. Providing no support for deserialization in the Weak t case just crashes the compiler. So we need some support.

crusso: I've decided to keep all the support but trap in relevant places. The alloc/set_field functions are now needed to
implemenent deserialization (which was actually causing a compiler message to std-err before).

See #5395

doc/md/examples/grammar.txt
```diff
@@ -48,6 +48,7 @@
 <typ_un> ::= 
     <typ_nullary>
     '?' <typ_un>
+    WEAK <typ_un>
```
crusso: I need to fix this.

crusso: (needs a change to the sed script in src/grammar)

rts/motoko-rts/src/types.rs
```diff
@@ -540,7 +543,7 @@ pub const TAG_FREE_SPACE: Tag = 43;
 // a lower boundary to distinguish slice information from
 // the actual tag values.
 #[enhanced_orthogonal_persistence]
-pub const TAG_ARRAY_SLICE_MIN: Tag = 44;
+pub const TAG_ARRAY_SLICE_MIN: Tag = 46;
```
crusso, line 546: @luc-blaeser is it safe to renumber this? Is there any possibility of TAG_ARRAY_SLICE_MIN being on the heap during an upgrade? If so, I think this is dangerous - it's essentially changing the interpretation of tags which needs to stay the same across upgrades.

But maybe I'm forgetting something.

src/prelude/prim.mo
```diff
@@ -89,11 +89,24 @@ func rts_upgrade_instructions() : Nat {
...
+//TODO: just call this `weak`?
+func allocWeakRef<T>(obj : T) : weak T {
+  (prim "alloc_weak_ref" : T -> weak T)(obj);
+};
+
+func weakGet<T>(w : weak T) : ?T {
+  (prim "weak_get" : weak T -> ?T)(w);
+};
+
+func isLive(weak_ref : weak Any) : Bool {
```
crusso, line 108: This typing for weak_ref only works if the weak is covariant (weak T <: weak U when T <: U), as it is at the moment since we can only read, but not write to a `weak T`.
However, if we later intend to add weakSet : <T> (weak T , T) -> (), then we would need a more specific `isLive<T>(weak_ref : weak T) : Bool` since this isLive won't be general enough. 

We would also need to make `weak T` invariant....

src/mo_values/prim.ml
```diff
@@ -442,4 +442,23 @@ let prim trap =
       fun _ v k -> k (Tup [
         Nat32 Numerics.Nat32.zero; Nat32 Numerics.Nat32.zero; Nat32 Numerics.Nat32.zero])
 
+  | "alloc_weak_ref" ->
```
crusso, line 445: There's a lot of whitespace changes in this file - did you use an auto-formatter (or vscode?) Doesn't hurt I guess but might be nicer to revert and just add your changes.

test/run-drun/weak-ref-upgrade.drun
```diff
@@ -0,0 +1,7 @@
+# ENHANCED-ORTHOGONAL-PERSISTENCE-ONLY
+install $ID weak-ref-upgrade/v1.mo ""
+ingress $ID test1 "DIDL\x00\x00"
+upgrade $ID weak-ref-upgrade/v2.mo ""
+ingress $ID test2 "DIDL\x00\x00"
+upgrade $ID weak-ref-upgrade/v3.mo ""
+ingress $ID test3 "DIDL\x00\x00"
```
crusso:
```suggestion
ingress $ID test3 "DIDL\x00\x00"

```

rts/motoko-rts/src/types.rs
alexandru-uta: I don't think anything is tagged as TAG_ARRAY_SLICE_MIN, it's more like a catch-all tag to be used in code to check for all tags. So it should be safe. @luc-blaeser please confirm.

src/prelude/prim.mo
alexandru-uta: up to you, I wouldn't complicate things, I find the current interface sufficient without the `weakSet()`;

src/mo_values/prim.ml
alexandru-uta: I'm not sure I see the whitespace change; For .ml files I haven't used any formatter.

src/prelude/prim.mo
alexandru-uta: and I'm guessing that these changes can also be done later if needed?

doc/md/examples/grammar.txt
alexandru-uta: I guess that's done?

src/mo_values/prim.ml
crusso: Sorry, I think I comment on the wrong file. It was a *.mo file (maybe prim.mo)?

src/prelude/prim.mo
crusso: I think it would be a breaking change to make it invariant after releasing as covariant, but perhaps not a biggie. I just thought it might be wise to keep our options more open.

My point of reference is https://ocaml.org/manual/5.3/api/Weak.html which seems to support both read and writes for some reason. @luc-blaeser do you anticipate us adding a user operation to set the contents of a weak reference?

src/prelude/prim.mo
```diff
@@ -89,11 +89,24 @@ func rts_upgrade_instructions() : Nat {
 };
```
crusso, line 90: This is the file with all the spurious whitespace changes. VSCode motoko package perhaps?

src/mo_values/prim.ml
alexandru-uta: ah, yes, prim.mo is formatted by our vs code extension..

src/prelude/prim.mo
alexandru-uta: exactly. Do we not like what our own formatter does?

src/codegen/compile_enhanced.ml
```diff
@@ -11955,6 +12019,22 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
 
+  | OtherPrim "alloc_weak_ref", [target] ->
+    SR.Vanilla,
+    compile_exp_vanilla env ae target ^^
+    E.call_import env "rts" "alloc_weak_ref"
+
+  | OtherPrim "weak_get", [weak_ref] ->
+    SR.Vanilla,
+    compile_exp_vanilla env ae weak_ref ^^
+    WeakRef.load_field env
```
crusso: I think we have another lurking bug with weak ref. The alloc_weak_ref takes a T and stores it as T in the field without any further wrapping option injection. weak_get reads the same field value and returns an ?T. Seem wrong to me and I suspect this will crash if T is itself an option type.

crusso: counterexample in https://github.com/dfinity/motoko/pull/5363/commits/5282c9fde915ba6c8225873b7670175c5f42f5e6
fix in
https://github.com/dfinity/motoko/pull/5363/commits/d354a6b42d33f4ed58b35245eb194939934cce91

rts/motoko-rts/src/types.rs
crusso: Form slack
Luc Bläser
  7:50 PM
I think it safe. TAG_ARRAY_MIN_SLICE is not stored on heap, just sanity check for array slicing, which should start at 128 II

src/prelude/prim.mo
crusso: No, I just don't like unrelated changes much

src/prelude/prim.mo
luc-blaeser: I believe read-only is powerful enough.

PR #5374, opened 2025-07-31
doc: reference `core` package

rvanasa:
Manually checked all of the changes in this PR (no AI or find/replace due to lots of potential edge cases). It's possible that I missed a few spots, but this should get us 99% of the way to using `core` across the docs and code examples.


doc/md/fundamentals/2-actors/5-messaging.md
```diff
@@ -50,7 +50,7 @@ In this example, a publisher canister maintains a list of subscribers and sends
 
 ```motoko no-repl
 import Subscriber "canister:subscriber";
-import Array "mo:base/Array";
+import Array "mo:core/Array";
 
 actor Publisher {
     stable var subscribers : [Subscriber.Subscriber] = [];
```
rvanasa, line 56: Making a note that we may want to revisit these previously incomplete pub/sub examples.

doc/md/14-language-manual.md
```diff
@@ -0,0 +1,2860 @@
+---
+sidebar_position: 14
+---
+
+# Language reference
```
Kamirus: Is this file meant to be here? There is `16-language-manual.md` as well

doc/md/16-language-manual.md
```diff
@@ -690,46 +690,46 @@ As bitwise types, these types support bitwise operations and (`&`), or (`|`) and
...
 
-### Type [`Principal`](https://internetcomputer.org/docs/motoko/base/Principal.md)
+### Type [`Principal`](https://internetcomputer.org/docs/motoko/core/Principal.md)
 
-The type [`Principal`](https://internetcomputer.org/docs/motoko/base/Principal.md) of category O (Ordered) represents opaque principals such as canisters and users that can be used to identify callers of shared functions and used for simple authentication. Although opaque, principals may be converted to binary [`Blob`](https://internetcomputer.org/docs/motoko/base/Blob.md) values for more efficient hashing and other applications.
+The type [`Principal`](https://internetcomputer.org/docs/motoko/core/Principal.md) of category O (Ordered) represents opaque principals such as canisters and users that can be used to identify callers of shared functions and used for simple authentication. Although opaque, principals may be converted to binary [`Blob`](https://internetcomputer.org/docs/motoko/core/Blob.md) values for more efficient hashing and other applications.
```
Kamirus, line 725: Do we have a way of automatically checking that every link is correct (does not return 404 or another error)?

crusso:
I think we may still need to update the examples in md/examples. There's a ./check.sh script to run them. Maybe I should do that bit as a stacked PR on this one.

doc/md/14-language-manual.md
rvanasa: Thanks for catching this! Seems like this was some sort of merge issue when renumbering the files. Removed the original file.

doc/md/16-language-manual.md
rvanasa: My understanding is that this is covered for relative links through Docusaurus in the `dfinity/portal` CI. We may eventually want to convert these URLs to relative links or see there is a prebuilt way to verify these as well.

doc/md/14-language-manual.md
crusso: I just double checked and you've kept the correct one (mentions Christoph's recent type field patterns - the other did not)

crusso (approved):
I did a quick make preview to look for build errors. Looks fine. 

Feel free to merge at your leisure.

PR #5391, opened 2025-08-11
Explanatory Upgrade Error Messages

luc-blaeser:
Trying to make upgrade compatibility messages (for DFX and Caffeine) more self-explanatory:
* Removing the unreadable hash suffixes from displayed type names
* Determining the actual inner-most type incompatibility cause
* Providing a path from stable variable down to the cause of mismatch
* Revised error messages, less verbose type displays, plus documentation link


crusso:
I've got a private branch that refactors some of this to use the pretty printer and remove some ineffectiency 

The build time on cycleops goes up from 24s to 28s which I can live with (it used to be 1m30 soo...)

crusso:
WIP PR is here #5393

PR #5393, opened 2025-08-12
experiment: refactor and optimize subtype explanations

crusso:
Refactor #5391 

* refactor to reduce diff
* separate formatting from subtype check
* to re-use the pretty printer rather than a custom one
* refactor context argument into similar RelArg, saving additional arguments
* Tweaks to error message (including more context for type mismatches)
* Remove perf regressions


Kamirus (commented):
I just browsed through and it looks really good!

Kamirus (approved):
I've reviewed and the code looks good! There was one copy-paste bug in `incompatible_func_sorts` but I've pushed the fix.

PR #5398, opened 2025-08-14
chore: add `motoko-core` artifacts to GitHub releases

rvanasa:


crusso (approved):

nix/release-files-ubuntu-latest.nix
```diff
@@ -9,4 +9,5 @@ pkgs.runCommandNoCC "motoko-release-${common.releaseVersion}" { } ''
   cp ${common.as_js "moc" packages.js.moc} $out/moc-${common.releaseVersion}.js
   cp ${common.as_js "moc-interpreter" packages.js.moc_interpreter} $out/moc-interpreter-${common.releaseVersion}.js
   tar --exclude=.github -C ${pkgs.sources.motoko-base-src} -czvf $out/motoko-base-library.tar.gz .
+  tar --exclude=.github -C ${pkgs.sources.motoko-core-src} -czvf $out/motoko-core.tar.gz .
```
Kamirus, line 12: @rvanasa I think we just need it here.
The other os targets check that moc compiles on these systems.
But the final release step just works on ubuntu latest and there we need to have base and core

Kamirus (changes requested):

nix/release-files-ubuntu-latest.nix
Kamirus: You can test that artifacts are correctly included by running the [draft release step here](https://github.com/dfinity/motoko/actions/workflows/release.yml)
Just click "Run workflow" and pick this branch, it will create a [draft release here](https://github.com/dfinity/motoko/releases)

rvanasa: Ok, I'll remove these from the other files. Thanks for the tip about the draft release. Seems to be working as expected!

Kamirus (approved):

PR #5400, opened 2025-08-15
experiment: make weak refs invariant and add IDL_EXT_weak for safe compatibility check

crusso:


test/run-drun/weak-ref-upgrade.drun
```diff
@@ -5,3 +5,8 @@ upgrade $ID weak-ref-upgrade/v2.mo ""
 ingress $ID test2 "DIDL\x00\x00"
 upgrade $ID weak-ref-upgrade/v3.mo ""
 ingress $ID test3 "DIDL\x00\x00"
+# the rest should fail
+upgrade $ID weak-ref-upgrade/v4.mo ""
```
crusso, line 9: Note these two new tests did not fail on master!

rts/motoko-rts/src/idl.rs
```diff
@@ -741,6 +743,11 @@ pub(crate) unsafe fn memory_compatible(
         (_, IDL_PRIM_reserved) => false, // information lost
         (IDL_PRIM_empty, _) | (IDL_PRIM_nat, IDL_PRIM_int) => variance != TypeVariance::Invariance,
         (_, IDL_CON_alias) | (IDL_CON_alias, _) => false,
+        (IDL_EXT_weak, IDL_EXT_weak) => {
+            let t11 = sleb128_decode(&mut tb1);
+            let t21 = sleb128_decode(&mut tb2);
+            memory_compatible(rel, TypeVariance::Invariance, typtbl1, typtbl2, end1, end2, t11, t21, false)
```
luc-blaeser: Does it need invariance because it is an immutable type?

src/codegen/compile_enhanced.ml
```diff
@@ -2280,6 +2252,63 @@ module Opt = struct
 
 end (* Opt *)
 
+
+module WeakRef = struct
+  (*
+      Weak references
+
+       ┌──────┬─────┬─────────┐
+       │ obj header │ field │
```
alexandru-uta, line 2261: add a few spaces after field so that the "|" gets aligned with the other ascii art?

rts/motoko-rts/src/idl.rs
```diff
@@ -741,12 +743,17 @@ pub(crate) unsafe fn memory_compatible(
...
+            let t21 = sleb128_decode(&mut tb2);
+            memory_compatible(rel, variance, typtbl1, typtbl2, end1, end2, t11, t21, false)
+        }
         (IDL_CON_opt, IDL_CON_opt) => {
             let t11 = sleb128_decode(&mut tb1);
             let t21 = sleb128_decode(&mut tb2);
             memory_compatible(rel, variance, typtbl1, typtbl2, end1, end2, t11, t21, false)
         }
-        (_, IDL_CON_opt) => false,
+        (_, IDL_CON_opt) => false, // BUG?: what about Null <: ?T
```
crusso, line 756: I'll investigate this in a separate PR.
I suspect it means we can't upgrade from
stable let x = null
to 
stable let x = ? 0
which seems wrong (as in stricter than the static check)

PR #5401, opened 2025-08-15
feat: add program lifetime instruction counter

alexandru-uta:
This PR adds a new monotonic program/canister instruction counter; Will wrap around u64. Will be used by the CI pipeline for performance tests; Currently, drun reports instruction counters, but once we migrate to pocket-ic we don't have this feature. The counter introduced here can be used in a similar way.


crusso (commented):
LGTM.
I you are worried about overflow, I think we have some support for emulating 128 bit values for other purposes.

We could also maybe report this stat in the rts stats method that Luc add (possibly in lowering/desugar.ml)

crusso (approved):

PR #5409, opened 2025-08-19
feat: report sub type failure explanations in typing

crusso:
Enable explanations for subtype checks


test/fail/ok/f-bounds-fail.tc.ok
```diff
@@ -18,14 +18,29 @@ f-bounds-fail.mo:9.53-9.54: type error [M0096], expression of type
   A
 cannot produce expected type
   C
+because: The type 
+  Any
```
Kamirus, line 30: this is interesting, why `Any` 😄

Kamirus: Ah it coerced `A` to its bound, which is `Any` in this case

test/fail/ok/bad-type-comp.tc.ok
```diff
@@ -2,10 +2,18 @@ bad-type-comp.mo:5.10-5.19: type error [M0096], expression of type
...
 cannot produce expected type
   module {type T = Null}
+because: Unsupported additional field `T` in type 
+  module {type T = Null}
+ of top level
 bad-type-comp.mo:6.10-6.39: type error [M0096], expression of type
   module {type T = Int}
 cannot produce expected type
   module {type T = Null}
+because: The type
```
crusso, line 12: This one looks a bit weird. Maybe we need a better printer for standalone Typ c cases

test/fail/ok/bounds-mismatch.tc.ok
```diff
@@ -2,26 +2,56 @@ bounds-mismatch.mo:13.5-13.14: type error [M0096], expression of type
   () -> ()
 cannot produce expected type
   <A>() -> ()
+because: Incompatible function types:
```
crusso: This still looks redundant.

test/fail/ok/bounds-mismatch.tc.ok
```diff
@@ -35,34 +65,74 @@ bounds-mismatch.mo:75.21-75.22: type error [M0096], expression of type
...
+because: Incompatible function types: 
+  <X>X -> X
+ does not match 
+  <Y <: Nat>Y -> Y
+ in top level
 bounds-mismatch.mo:81.21-81.22: type error [M0096], expression of type
   <X <: Int, Y <: {}>X -> Y
 cannot produce expected type
   <X <: Int, Y <: {x : Nat}>X -> Y
+because: Incompatible function types:
```
crusso: would be nice if we could pinpoint the incompatible bound.

test/fail/ok/bounds-mismatch.tc.ok
```diff
@@ -35,34 +65,74 @@ bounds-mismatch.mo:75.21-75.22: type error [M0096], expression of type
...
+because: Incompatible function types: 
+  <X <: Nat, Y>X -> Y
+ does not match 
+  <Y <: Nat, X <: Y>Y -> X
+ in top level
 bounds-mismatch.mo:93.21-93.22: type error [M0096], expression of type
   <X, Y>X -> X
 cannot produce expected type
   <X <: Y, Y>X -> X
+because: Incompatible function types:
```
crusso: ditto

test/fail/ok/bounds-mismatch.tc.ok
Kamirus: true! I'll add func types to the redundant check

test/fail/ok/check-record.tc.ok
```diff
@@ -16,3 +16,8 @@ check-record.mo:24.9-24.27: type error [M0096], expression of type
   {a : Nat}
 cannot produce expected type
   {b : Nat}
+because: Unsupported additional field `b` in type
```
crusso:
```suggestion
because: Unsupported additional field `b` in type 
```
because: Missing exected field `b`.

Might be more succinct...

test/fail/ok/f-bounds-fail.tc.ok
```diff
@@ -18,14 +18,53 @@ f-bounds-fail.mo:9.53-9.54: type error [M0096], expression of type
   A
 cannot produce expected type
   C
+because: Type variable 
+  A
+ was upcast to its bound
```
crusso:
```suggestion
 was promoted to its bound 
```

test/fail/ok/bad-type-comp.tc.ok
```diff
@@ -2,10 +2,18 @@ bad-type-comp.mo:5.10-5.19: type error [M0096], expression of type
...
+ of top level
 bad-type-comp.mo:6.10-6.39: type error [M0096], expression of type
   module {type T = Int}
 cannot produce expected type
   module {type T = Null}
+because: The type 
+  = (type = Int)
+ is not compatible with type 
+  = (type = Null)
+ of top level.T
```
Kamirus: `top level.T` ? That is also weird

test/fail/ok/f-bounds-fail.tc.ok
crusso: Promoted might be better than cast, since we don't have casts...

test/fail/ok/objpat-infer.tc.ok
```diff
@@ -4,19 +4,44 @@ objpat-infer.mo:23.13-23.21: type error [M0096], expression of type
...
+ in top level
 objpat-infer.mo:26.13-26.14: type error [M0096], expression of type
   actor {bar : shared {} -> async Nat}
 cannot produce expected type
   {}
+because: Incompatible object sorts: 
+  actor {bar : shared {} -> async Nat}
+ does not match 
+  {}
+ in top level
```
crusso: Maybe we can just omit "in top level" ?

test/fail/ok/objpat-infer.tc.ok
```diff
@@ -4,19 +4,44 @@ objpat-infer.mo:23.13-23.21: type error [M0096], expression of type
...
+because: Incompatible object sorts: 
+  {}
+ does not match 
+  actor {}
+ in top level
 objpat-infer.mo:30.13-30.22: type error [M0096], expression of type
   {}
 cannot produce expected type
   actor {}
+because: Incompatible object sorts:
```
crusso: Maybe we should just report the sorts, rather than the entire type?

test/fail/ok/pipe-ill-typed.tc.ok
```diff
@@ -2,10 +2,16 @@ pipe-ill-typed.mo:7.33-7.34: type error [M0096], expression of type
   {#B}
 cannot produce expected type
   {#A}
+because: Missing tag `#B` in type
```
crusso:
```suggestion
because:  Case `#B` missing from type 
```

test/run-drun/ok/scope-example-implicit.tc.ok
```diff
@@ -2,3 +2,8 @@ scope-example-implicit.mo:10.12-10.19: type error [M0096], expression of type
   async<$@anon-async-8.12> Int
 cannot produce expected type
   async<$f> Int
+because: The type
```
crusso:
```suggestion
because: The scope
```

test/run-drun/ok/scope-example-implicit.tc.ok
```diff
@@ -2,3 +2,8 @@ scope-example-implicit.mo:10.12-10.19: type error [M0096], expression of type
   async<$@anon-async-8.12> Int
 cannot produce expected type
   async<$f> Int
+because: The type 
+  $@anon-async-8.12
+ is not compatible with type
```
crusso:
```suggestion
 is not compatible with the scope 
```

test/fail/ok/objpat-infer.tc.ok
Kamirus: I assume Luc wanted it for the AI, it does not hurt to leave it for now

test/fail/ok/issue-3318.tc.ok
```diff
@@ -2,23 +2,53 @@ issue-3318.mo:6.24-6.25: type error [M0096], expression of type
   (x : (Nat, Nat)) -> (Int, Int)
 cannot produce expected type
   ((Nat, Nat)) -> ((Int, Int))
+because: The type 
+  Int
+ is not compatible with type 
+  (Int, Int)
+ of top level
```
crusso, line 9: Maybe we could distinguish dom/range (or argument/result) and tuple contexts.
`in tuple of return type`

test/run-drun/ok/scope-example-implicit.tc.ok
```diff
@@ -2,3 +2,8 @@ scope-example-implicit.mo:10.12-10.19: type error [M0096], expression of type
   async<$@anon-async-8.12> Int
 cannot produce expected type
   async<$f> Int
+because: Incompatible async scopes: 
+  $@anon-async-8.12
+ does not match 
+  $f
+ in top level
```
crusso, line 9: Maybe the "in top level" are a bit redundant.
Maybe "in the expected type" would be better?

test/fail/ok/f-bounds-fail.tc.ok
```diff
@@ -126,51 +165,200 @@ f-bounds-fail.mo:70.27-70.28: type error [M0096], expression of type
...
+ and: The type 
+  Any
+ is not compatible with type 
+  C
+ of top level
 f-bounds-fail.mo:82.24-82.25: type error [M0096], expression of type
   A
 cannot produce expected type
   C
+because: Type variable
```
crusso, line 339: These are a bit long-winded, but also very rare, so perhaps no worth improving much.

src/mo_frontend/typing.ml
```diff
@@ -377,6 +377,27 @@ let sub env at t1 t2 =
       display_typ_expand t1
       display_typ_expand t2
 
+let sub_explained env at t1 t2 =
```
Kamirus, line 380: Let's make sure we address these kind of errors
```
type error [M0204], migration expression produces field `userProfiles` of type
  var Map<Principal, UserProfile>
, not the expected type
  var Map<Principal, UserProfile>"
```

test/fail/ok/f-bounds-fail.tc.ok
Kamirus: not worth worrying about these cases, they are rare like you said

Kamirus (approved):
That's largely an improvement, we could spend even more time polishing this, but I'd rather merge it and improve in another PR than have this unmerged for months

crusso:
(Continued in and replaced by #5643)

Issue #5283, opened 2025-06-23
Sometimes breaks or continues with awaits cause a compiler error


skilesare:
Something funky with awaits and breaks. Sure enough, when I took out all of my breaks I could compile again:

https://chatgpt.com/share/6859e5c2-067c-8007-914d-4c40d22e28bb

It turned out not to be Motoko, nor your library, but the way `await.ml` is threading its label-environment around your async–CPS bits that was tripping you up.  In particular, you were accidentally “leaking” continuation entries (the `Cont k` bindings for your implicit return/throw/cleanup labels) into places where you really only wanted the user-defined “Label” bindings.  That meant that by the time you got to

```ocaml
| PrimE (BreakPrim id, [exp1]) ->
  begin
    match LabelEnv.find_opt (Named id) context with
    | Some (Cont k) ->
        (* …wrongly taken, because we had a Cont entry for id… *)
        (retE (varE k -*- t_exp context exp1)).it
    | Some Label ->
        (breakE id (t_exp context exp1)).it
    | None -> assert false
  end
```

you sometimes really *did* have a `Cont k` for that same `id` in your `context` (from a previous async/block), so you would mistakenly lower `break` into a `retE` and then end up with an IR that had no matching label, blowing up later.

---

### The fix

1. **Isolate your `context`** per transform.  Right now, you do

   ```ocaml
   let t_exp context exp = …
   and c_exp  context exp k = …
   ```

   everywhere and you keep adding to the very same `context` when you descend into an `AsyncE`.  Instead, make sure that the `LabelEnv` you pass to the *rest* of your translate after finishing a CPS‐transform is a *fresh* copy (i.e. just the labels you really want).  That way a user‐defined label never collides with one of your “return/throw/cleanup” conts.

2. **Swap the match-arms** in the *pure* case so that you always handle the true *syntax*‐label first:

   ```diff
   | PrimE (BreakPrim id, [exp1]) ->
   -  (match LabelEnv.find_opt (Named id) context with
   -   | Some (Cont k) -> …(* never really want this in pure code *)…
   -   | Some Label    -> breakE id (…)
   -   | None          -> assert false)
   +  (match LabelEnv.find_opt (Named id) context with
   +   | Some Label    ->
   +       (breakE id (t_exp context exp1)).it
   +   | Some (Cont k) ->
   +       (* this really only makes sense in a CPS‐context *)
   +       (retE (varE k -*- t_exp context exp1)).it
   +   | None ->
   +       assert false)
[...]

PR #5412, opened 2025-08-20
bugfix: issue-5283 (repro and fix)

crusso:
Fix for issue #5283.

The await translation would crash in obscure cases. Turns out its due to a bug in ir/rename.ml that wasn't properly distinguishing labels from identifiers when renaming them.

The counter-example is this code from the repro

```
                       label satset for (satset in [1,2,4].vals()) {
                          if (4==5) {
                            
                            break satset;
                          }
                        };
```

This uses a label and identifier with the same name, and unless kept distinct, after broken renaming,  would result in the inner break referencing the new identifier for (inner) satset rather than the new label for (outer) satset.

The fix is for renaming to use a map with a domain that distinguishes identifers from labels. 
Even better would be if we didn't use the same type (string, yuck) for identifiers and labels, but that is a bigger refactoring for later.


src/ir_def/rename.ml
```diff
@@ -23,10 +46,10 @@ let rec ids_bind rho = function
 
 let arg_bind rho a =
   let i' = fresh_id a.it in
-  ({a with it = i'}, Renaming.add a.it i' rho)
+  ({a with it = i'}, Renaming.add (Binder.Id a.it) i' rho)
```
christoph-dfinity:
```suggestion
let arg_bind rho a =
  let i', rho' = id_bind rho a.it in
  ({a with it = i'}, rho')
```

Couldn't this use the `id_bind` function?

crusso: Yep, think so!

PR #5379, opened 2025-08-05
Claudio/simple stable functions dynamics

crusso:

PR #5413, opened 2025-08-20
experiment: simpler stable functions, eop edition (WIP)

crusso:
Implements the simpler stable function proposal  for both classical and eop.

Basic idea:

- We maintain a global record of stable functions, with mutable fields storing `Local` functions
- A stable function declaration updates the table with the local version of the function, and returns a stable proxy.
- The proxy is a closure with function index -1 and a single field storing the label/name of the stable function.
- On closure application, we detect proxies and indirect through the table.
- On deserialization, we allocate a proxy using just the expect function name. Could be shared but aren't for now.
- For eop, proxies have a completely stable representation and can be re-used across upgrades without change.

TODO:
- [x] in desugar.ml register initial implementations that signal use-before-define instead of looping in lookup
- [ ] optimize Closure.prepare_call and Closure.call to avoid the redundant dynamic test in the second (easiest to modify call sites)
- [ ] the  write barriers when updating function table are ugly and should be refactored.
- [ ] stable_func lookup would be much faster (possibly 0(1)) if the backend knew the type of the stable-record (optimizing and/or avoiding field search). Make it so.
- [ ] the original classic implementation in #5379 is actually faster (no funky Closure.prepare) - maybe there's a compromise.
     Our indirect function calls are now more expensive. 
- [ ] Need to test upgrades from classical - eop.
- [ ] frontend: Consider adding paired type definitions (as we do for classes) so its easier to write the type of shared function `compare` as type `compare` (= `stable compare (Nat, Nat) -> Order.Order`) (but beware of existing class punning, sigh)
- [ ] replace current type inference hack with something more principled.
- [ ] test recursive stable functions don't suck (and work)
- [ ] optimize calls to known stable function to direct calls.
- [x] add tests for use-before-define
- [ ] Add support for generic stable functions - needed fancier type descriptors c.f. Luc's stable functions.
- [x ] ild.rs for skipping stable functions needs to do nothing (not advance)
- [ ]  type descriptor for stable functions (and maybe idl) (`4`) should also include name hash for safe comparison (and generics)
- [ ] check graph stabilization works


src/codegen/compile_enhanced.ml
```diff
@@ -11666,6 +11743,24 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
+  | OtherPrim "set_stable_funcs", [e] ->
+    SR.unit,
+    Heap.get_static_variable env (E.get_stable_funcs __LINE__ env) ^^
+    Tagged.load_forwarding_pointer env ^^
+    compile_add_const ptr_unskew ^^
+    compile_add_const (Int64.mul MutBox.field Heap.word_size) ^^
+    compile_exp_vanilla env ae e ^^
+    Tagged.write_with_barrier env
+
+  | OtherPrim "get_stable_funcs", [] ->
```
crusso: Delete me?

src/codegen/compile_classical.ml
```diff
@@ -11491,6 +11577,19 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
 
+  (* stable function record *)
+
+  | OtherPrim "set_stable_funcs", [e] ->
+    SR.unit,
+    compile_unboxed_const (E.get_stable_funcs env) ^^
+    compile_exp_vanilla env ae e ^^
+    MutBox.store_field env
+
+  | OtherPrim "get_stable_funcs", [] ->
```
crusso: Delete me?

src/mo_values/prim.ml
```diff
@@ -440,6 +440,10 @@ let prim trap =
       fun _ v k -> k unit
   | "getCandidLimits" ->
       fun _ v k -> k (Tup [
-        Nat32 Numerics.Nat32.zero; Nat32 Numerics.Nat32.zero; Nat32 Numerics.Nat32.zero])
+      Nat32 Numerics.Nat32.zero; Nat32 Numerics.Nat32.zero; Nat32 Numerics.Nat32.zero])
```
crusso:
```suggestion
        Nat32 Numerics.Nat32.zero; Nat32 Numerics.Nat32.zero; Nat32 Numerics.Nat32.zero])
```

PR #5415, opened 2025-08-21
fix: warn (and ignore) parentheticals on `async*` calls

ggreif:
- **WIP**
- **try building `motoko-core`**
- **chore: fix test**
- **fix: warn (and ignore) parentheticals on `async*` calls**


test/run-drun/ok/par.drun-run.ok
```diff
@@ -9,11 +9,11 @@ debug.print: test3()
 debug.print: oneshot deadline set: false
 debug.print: oneshot cycles: 0
 debug.print: oneshot deadline set: false
-debug.print: oneshot cycles: 3_456
-debug.print: oneshot deadline set: true
 debug.print: oneshot cycles: 0
-debug.print: oneshot deadline set: true
-debug.print: oneshot cycles: 4_567
+debug.print: oneshot deadline set: false
+debug.print: oneshot cycles: 0
```
ggreif: this is not good

ggreif: fixed!

test/run-drun/ok/par.drun-run.ok
```diff
@@ -38,3 +38,8 @@ ingress Completed: Reply: 0x4449444c0000
 debug.print: test10()
 debug.print: anon: 34_567
 ingress Completed: Reply: 0x4449444c0000
+debug.print: test11()
+debug.print: star: 0
+debug.print: oneshot deadline set: false
```
ggreif, line 44: previously this would show up with a deadline and cycles :-(

src/lowering/desugar.ml
```diff
@@ -214,7 +214,8 @@ and exp' at note = function
     I.PrimE (I.OtherPrim "blob_size", [exp e1])
   (* Normal call *)
   | S.CallE (par_opt, e1, inst, e2) ->
```
crusso: Isn't it weird to just drop it? It might have a side-effect?

test/run-drun/ok/par.tc.ok
```diff
@@ -3,4 +3,5 @@ par.mo:51.11-51.55: warning [M0212], unrecognised attribute yeah in parenthetica
 par.mo:79.11-79.41: warning [M0212], unrecognised attribute moot in parenthetical
 par.mo:126.15-126.26: warning [M0211], redundant empty parenthetical note
 par.mo:128.15-128.26: warning [M0211], redundant empty parenthetical note
+par.mo:144.16-144.48: warning [M0210], misplaced parenthetical (`async*` calls cannot be modified)
```
crusso: Why not just error for now? Then dropping the parenthetical isn't an issue either

src/mo_frontend/typing.ml
```diff
@@ -2970,7 +2970,8 @@ and check_parenthetical env typ_opt = function
        let s, _, _, _, ts2 = T.as_func fun_ty in
        begin match ts2 with
        | _ when T.is_shared_sort s -> ()
-       | [cod] when T.is_async cod -> ()
+       | [cod] when T.is_fut cod -> ()
+       | [cod] when T.is_async cod -> warn env par.at "M0210" "misplaced parenthetical (`async*` calls cannot be modified)"
```
crusso, line 2974:
```suggestion
       | [cod] when T.is_cmp cod -> warn env par.at "M0210" "misplaced parenthetical (`async*` calls cannot be modified)"
```
Do we have that?

src/lowering/desugar.ml
ggreif: Yeah, we shouldn't just drop it. Hmmm, I need to pimp it a bit.

ggreif: 76b40cfb4dbac5bb6b2bf291915bad5c92ddfed0 fixes the effects issue.

src/lowering/desugar.ml
```diff
@@ -214,7 +214,10 @@ and exp' at note = function
...
-    let ds = parenthetical par_opt in
+    let send e1_typ = T.(is_func e1_typ &&
+                           (let s, _, _, _, _ = as_func e1_typ in
+                            is_shared_sort s || is_fut note.Note.typ)) in
+    let ds = parenthetical (send e1.note.S.note_typ) par_opt in
     (blockE ds { at; note; it = I.PrimE (I.CallPrim inst.note, [exp e1; exp e2]) }).it
```
ggreif: As @crusso mentioned in Slack, both `exp e1` and `exp e2` may in turn contain parentheticals/sends, which may mess up the right order to fill in the registers. So we'll need a bit more logic to `let`-bind those and only set the registers immediately for `CallPrim`. deb29d0c2 demonstrates the issue.

ggreif: This is fixed by 2900d358d94039b679af851211abd4de615b553f.

src/lowering/desugar.ml
```diff
@@ -213,9 +213,18 @@ and exp' at note = function
...
+  (* Call with parenthetical *)
+  | S.CallE (Some _ as par_opt, e1, inst, e2) ->
+    let send e1_typ = T.(is_func e1_typ &&
+                           (let s, _, _, _, _ = as_func e1_typ in
+                            is_shared_sort s || is_fut note.Note.typ)) in
+    let ds, rs = parenthetical (send e1.note.S.note_typ) par_opt in
+    let v1, v2 = fresh_var "e1" e1.note.S.note_typ, fresh_var "e2" e2.note.S.note_typ in
+    (blockE
+       (ds @ [letD v1 (exp e1); letD v2 (exp e2)])
+       (blockE rs I.{ at; note; it = PrimE (CallPrim inst.note, [varE v1; varE v2]) })).it
```
crusso:
```suggestion
    (blockE
       (ds @ 
        (letD v1 (exp e1) :: 
         letD v2 (exp e2) :: 
         rs)
       (primE (CallPrim inst.note)  [varE v1; varE v2]).it
```
Might work too (unchecked)

src/lowering/desugar.ml
```diff
@@ -247,12 +256,12 @@ and exp' at note = function
   | S.RetE e -> (retE (exp e)).it
   | S.ThrowE e -> I.PrimE (I.ThrowPrim, [exp e])
   | S.AsyncE (par_opt, s, tb, e) ->
-    let ds = parenthetical par_opt in
+    let ds, rs = parenthetical true par_opt in
```
crusso:
```suggestion
    let ds, rs = parenthetical (s.it = Fut) par_opt in
```
I guess you only want to do something real on async, not async*?

src/mo_frontend/typing.ml
crusso: Would it not make sense to just make this an error?

src/lowering/desugar.ml
ggreif: Yes, but IIRC the parser ensures this
```
> (with bla = 9) async* {};
stdin:1.16-1.22: syntax error [M0001], unexpected token 'async*', expected one of token or <phrase> sequence:
  <obj_or_class_dec>
  async <exp_nest>
  <exp_post(ob)> <inst> <exp_nullary(ob)>
```
Still, makes sense to change.

src/mo_frontend/typing.ml
ggreif: There is no `is_cmp`.

Also there is nothing inherently wrong with parentheticals, like they could be `(with inline = true) ...` or such in the future.

src/lowering/desugar.ml
ggreif: Adopted!

ggreif: Alas, it doesn't work, because `primE` wants the resulting type of it, but `CallPrim` doesn't have it!

src/mo_frontend/typing.ml
crusso: Ok, but users are more likely to ignore warnings and can't ignore errors...But fine, let's keep it like this.

src/mo_frontend/typing.ml
```diff
@@ -2970,7 +2970,8 @@ and check_parenthetical env typ_opt = function
        let s, _, _, _, ts2 = T.as_func fun_ty in
```
crusso, line 2970: check_parenthetical seems to only reject application that return an async*. Does it complain when the parenthetical is attached to an async* expression ?

e.g.
```
(with cycles = 100) async* { };
```

crusso: Odd, because I'm sure I commented on this earlier but perhaps I didn't submit the comment properly. Or GH lost it.

ggreif: My response also disappeared:
```
(with cycles = 100) async* { };
stdin:2.21-2.27: syntax error [M0001], unexpected token 'async*', expected one of token or <phrase> sequence:
  <obj_or_class_dec>
  async <exp_nest>
  <exp_post(ob)> <inst> <exp_nullary(ob)>
```
These simply don't parse at all.

src/mo_frontend/typing.ml
ggreif: Still, nothing bad will happen.

src/lowering/desugar.ml
ggreif: Here is the response @crusso.

src/mo_frontend/typing.ml
ggreif: With #5419 there is a `T.is_cmp` now.

PR #5419, opened 2025-08-22
feat: warn when `ignore`-ing an `async*`

ggreif:
There are examples out there and this helps avoiding that mistake. Any `async*` value is inert until an `await*` forces it. `ignore`-ing it may drop the only reference to it, which means it has been built in vain. What people probably want is `ignore`-ing the `await*`-ed value.


Kamirus (approved):
LGTM!

PR #5442, opened 2025-09-03
chore: add test-runner tests

alexandru-uta:
This PR adds initial tests for the test-runner package and refines the nix setup to remove duplication and add a derivation for the test-runner tests.

### DIFF
--- a/test-runner/src/main.rs
+++ b/test-runner/src/main.rs
@@ -248,6 +248,7 @@ impl TestCommand {
     }
 
     fn read_wasm_file(wasm_path: &PathBuf) -> Result<Vec<u8>, std::io::Error> {
+        println!("Reading wasm file: {}", wasm_path.display());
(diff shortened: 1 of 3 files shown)


test-runner/src/main.rs
```diff
@@ -706,3 +706,36 @@ fn main() {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // TODO: Add more tests.
+
```
christoph-dfinity:
```suggestion
```

Let's either be more concrete about what specific things need testing, or just drop this. In general I think TODO's should be actionable

christoph-dfinity (approved):
Thanks! Just one nit

test-runner/src/main.rs
alexandru-uta: agreed!

PR #5445, opened 2025-09-04
chore: add `List.at()` in the docs

alexandru-uta:


christoph-dfinity:
Did you make this change "by hand" or use some kind of automation?

alexandru-uta:
> Did you make this change "by hand" or use some kind of automation?
> 
> EDIT: Ahh found it:
> 
> https://github.com/dfinity/motoko/blob/13e2ee067bde2feaf560c1e6ead46d18c0d729c9/doc/Makefile#L20-L23
> 
> Did you use that?

Errr, just by hand; had no idea you could do it differently

christoph-dfinity:
I'm a bit surprised CI isn't choking then... we should run this on the library versions we reference via `nix` and verify on CI...

alexandru-uta:
> I'm a bit surprised CI isn't choking then... we should run this on the library versions we reference via `nix` and verify on CI...

yup, doing it now, rooking mistake :(

PR #5447, opened 2025-09-04
fix: fix update-flake-lock nigthly action

alexandru-uta:
The update-flake-lock nightly action was broken due to nix packages incompatibilities. Pinning a version of nix-update should fix it for now.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -2,7 +2,7 @@
   description = "The Motoko compiler";
 
   inputs = {
-    nixpkgs.url = "github:NixOS/nixpkgs/nixos-24.11";
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-25.05";
 
     flake-utils.url = "github:numtide/flake-utils";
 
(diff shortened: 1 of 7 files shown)


flake.nix
```diff
@@ -6,7 +6,7 @@
 
     flake-utils.url = "github:numtide/flake-utils";
 
-    nix-update-flake.url = "github:Mic92/nix-update";
+    nix-update-flake.url = "github:Mic92/nix-update/09aadb5d6d9e1fc57df0b61def4bdd8b43ea47a1";
```
ggreif: The weekly updates seem to work. How is this recipe different?

ggreif: Okay, I triggered the daily workflow manually, and it seems like it worked: https://github.com/dfinity/motoko/actions/runs/17547142050/job/49831108455

alexandru-uta: the problem appears when the PR is pushed, see:
https://github.com/dfinity/motoko/pull/5457

alexandru-uta: I managed to fix something, but now something else breaks...

alexandru-uta: looks like it's fixed now

Issue #2537, opened 2021-05-21
[Feature Idea] Contextual method modules


rossberg:
## Contextual method modules

[This is a half-baked idea meant as food for discussion, not a finalised proposal. Let me know what you think, or if you see ways to make it more powerful.]

### Motivation

* There is a schism in Motoko between object-oriented (using classes and methods) vs procedural/functional (using types and functions) patterns for defining abstractions. Both have pros and cons that cannot be reconciled in the language as is.

* We have special rules for treating arrays, blobs, and text as pseudo objects when used with dot notation. Besides being super ad-hoc, the special status of these functions has repeatedly confused users who tried to find them, e.g., in the Array module.

* Functional notation, as required for most of our library functions, has a well-known UX disadvantage in that e.g. code completion does not apply.

* We have repeatedly had the desire for a type-class like mechanism, e.g., for printing or serialisation. (This proposal does not have comparable expressiveness, but works for basic cases.)

### Idea

This is a moderate proposal to essentially generalise the special mechanism for arrays and text and make pseudo methods similar to theirs user-definable.

It does so without any syntactic extension and with only a local change to the type system (though making it syntactically more explicit would be an option).

The basic idea is to generalise the typing rule for the dot operator. Given the expression `e.x`:

1. If `e` does not have object type, or it is an object with no field `x`, then the field is searched for in the environment.

2. The search looks for a _module_ that (1) has a (possibly parameteterised) type member named `Self` that (can be instantiated such that it) is a supertype of `e`'s type, and (2) has a value member `x`.

3. In case there are multiple modules in the context satisfying these two conditions, the innermost (most recent) binding takes precedence. (This is to maintain compositionality.)

4. If a suitable module, say `M`, is found, then the expression is typed as if it was the application `M.x(e)` (including possible inference for omitted type arguments).

Note: If the type of `e` is `None`, then by subsumption it has any field `x`, so this fallback does not apply and it is typed the same as always.

### Examples

The existing magic array, blob, and text methods could be defined by their respective modules Array, Blob, and Text as follows (assuming we include appropriate sugar for curried function definitions):
```
module Array {
  public type Self<T> = [T];

  public func get<T>(a : [T])(i : Nat) : T { a[i] };
  public func size<T>(a : [T])() : Nat { ... };
  public func keys<T>(a : [T])() : Iter<Nat> { ... };
  public func vals<T>(a : [T])() : Iter<T> { ... };
  ...
}
// ...and something analogue for mutable arrays

[...]

PR #5458, opened 2025-09-09
feat: implements contextual dot resolution

christoph-dfinity:
At a high level this allows writing a function call like `List.concat(list1, list2)` as `list1.concat(list2)`, by finding the `List.concat` function in the context using the inferred type of `list1`. This allows for a more fluent style of using code defined with data types and functions as opposed to classes and methods.

This implements the design in https://github.com/dfinity/motoko/issues/2537 with its first two variations. It extends it slightly by not only taking the `Self` type of the module into account, but also checking compatibility with the first argument of the to be picked overload.

TODO:
- [ ] Decide if the `Self` module type is the way we want to go. I went with it over searching through every function in scope with the hope of keeping compile times in check.
- [ ] Write up proper spec, by adjusting the design in https://github.com/dfinity/motoko/issues/2537 to the current implementation
- [ ] Make a branch of `core` that uses this throughout to see if we like the impact


design/ContextualDot.md
```diff
@@ -0,0 +1,23 @@
...
+
+Functional notation, as required for most of our library functions, has a well-known UX disadvantage in that e.g. code completion does not apply.
+
+
+## Design
+The basic idea is to generalise the typing rule for the dot operator. Given the expression `e.x(args)`:
+
+If e does not have object type, or it is an object with no field `x`, then the field is searched for in the environment.
+
+The search looks for a module that (1) has a (possibly parameteterised) type member named `Self` that (can be instantiated such that it) is a supertype of `e`'s type, and (2) has a value member `x` with a function type `x : (t_arg, ...) -> t_res` where `t_arg` (can be instantiated such that it) is a supertype of `e`'s type. Notably these two instantiations do _not_ need to be the same, so a module can export functions with more specialized arguments than its `Self` type.
```
crusso:
```suggestion
The search looks for a module that (1) has a (possibly parameterized) type member named `Self` that (can be instantiated such that it) is a supertype of `e`'s type, and (2) has a value member `x` with a function type `x : (t_arg, ...) -> t_res` where `t_arg` (can be instantiated such that it) is a supertype of `e`'s type. Notably these two instantiations do _not_ need to be the same, so a module can export functions with more specialized arguments than its `Self` type.
```

crusso (approved):
Opening an issue to improve doc

PR #5459, opened 2025-09-09
feat: mixins

christoph-dfinity:
## Mixins

Mixins allow defining parts of an actor in their own file and then _including_ them into another actor.

## Mixed Counter example

``` motoko
// File mixins/CounterMixin.mo
mixin(incAmount : Nat) {
  var counter : Nat = 0;
  public func inc() : async () {
    counter += incAmount;
  };
};
```

``` motoko
import CounterMixin "mo:mixins/CounterMixin";

persistent actor {
  include CounterMixin(2);
  public func incAndGet() : async Nat {
    await inc();
    counter
  };
}
```

When an actor includes a mixin it adds its declarations to its body. `public` definitions of the mixin also become part of the interface of the including actor. The actor has access to the `private` fields, types and methods of the mixin. Access to private fields is required, so that a migration expression on the main actor can migrate the composition of actor and mixin.

Mixins can be parameterized over arguments that are bound as transient lets (with renaming so they don't conflict with existing definitions). This means that if you want the mixin to have access to state of the including actor you can't just pass a variable reference, but need to wrap it up in an object. Passing already mutable objects like `core/Map`, `core/List`, `[var A]` works just fine.

## Todo
- [x] Check interaction with `inspect` method
- [x] Fix bug involving named actor and mixin
- Documentation
  - [ ] Language manual
  - [ ] Guides
- [x] Enable actor checks in `infer_obj` for mixins (needs adjusting)
- [ ] Interpreter implementation


src/pipeline/pipeline.ml
```diff
@@ -177,7 +177,8 @@ let async_cap_of_prog prog =
   match (CompUnit.comp_unit_of_prog false prog).it.body.it with
   | ActorClassU _ -> Async_cap.NullCap
   | ActorU _ -> Async_cap.initial_cap()
-  | ModuleU _ -> assert false
+  | MixinU _ -> Async_cap.initial_cap()
+  | ModuleU _
```
crusso: Why not assert false any more?

src/ir_def/subst_var.ml
```diff
@@ -0,0 +1,175 @@
+open Source
+open Ir
+
+module Binder = Rename.Binder
+module Renaming = Rename.Renaming
+
+(* One traversal for each syntactic category, named by that category *)
+
+let fresh_id id = Construct.fresh_id id ()
```
crusso: Unused?

christoph-dfinity: Fixed in 7b15c82

src/pipeline/pipeline.ml
christoph-dfinity: Not changed intentionally, thanks for spotting.

src/lowering/desugar.ml
```diff
@@ -579,12 +579,26 @@ and export_runtime_information self_id =
   )],
   [{ it = I.{ name = lab; var = v }; at = no_region; note = typ }])
 
+and build_stabs (df : S.dec_field) : stab option list = match df.it.S.dec.it with
+  | S.TypD _ | S.MixinD _ -> []
```
crusso: I guess the MixinD case could even be an assert.

src/lowering/desugar.ml
```diff
@@ -1302,10 +1333,12 @@ let transform_unit_body (u : S.comp_unit_body) : Ir.comp_unit =
     end
   | S.ActorU (persistence, exp_opt, self_id, fields) ->
     let eo = Option.map exp exp_opt in
-    let actor_expression = build_actor u.at [] eo self_id fields u.note.S.note_typ in
+    let ty = u.note.S.note_typ in
+    (* Printf.printf "COMPILING AT TYPE: %s\n" (T.string_of_typ ty); *)
```
crusso:
```suggestion
    let ty = u.note.S.note_typ in
    let actor_expression = build_actor u.at [] eo self_id fields ty in
```

src/mo_frontend/parser.mly
```diff
@@ -928,6 +928,11 @@ dec_nonvar :
       let is_sugar, e = desugar_func_body sp x t fb in
       let_or_exp named x (func_exp x.it sp tps p t is_sugar e) (at $sloc) }
   | eo=parenthetical_opt mk_d=obj_or_class_dec  { mk_d eo }
+  | MIXIN p=pat_plain dfs=obj_body {
+     let dfs = List.map (share_dec_field (Stable @@ no_region)) dfs in
```
crusso, line 932: I wonder if the stability should actually depends on the default persistence mode (see below in obj_or_class_dec).

crusso: Actually, I think this is sensible as as (default to stable) and we just document that. We can't really make it sensibly dependent on the call-site.

crusso: I guess we could rule out mixin in non-persistent actors at the call-site, if we are worried about that.

design/Mixins.md
```diff
@@ -0,0 +1,31 @@
...
+  include CounterMixin(2);
+  public func incAndGet() : async Nat {
+    await inc();
+    counter
+  };
+}
+```
+
+When an actor includes a mixin it adds its declarations to its body. `public` definitions of the mixin also become part of the interface of the including actor. The actor has access to the `private` fields, types and methods of the mixin. Access to private fields is required, so that a migration expression on the main actor can migrate the composition of actor and mixin.
+
```
crusso, line 30:
```suggestion

The fields of an actor are `stable` by default, even when included in legacy, non-persistent actors.

```

src/mo_frontend/static.ml
```diff
@@ -99,7 +99,10 @@ and exp_fields m efs = List.iter (fun (ef : exp_field) ->
   exp m ef.it.exp) efs
 
 and dec m d = match d.it with
-  | TypD _ | ClassD _ -> ()
+  | TypD _ | ClassD _ | MixinD _ -> ()
+  | IncludeD _ ->
+    (* TODO require static on args?*)
```
crusso: Probably should check the args, but these could even be asserts since I guess its illegal to include a mixin in a module? but fine to be more permissive

src/mo_interpreter/interpret.ml
```diff
@@ -1033,6 +1036,15 @@ and interpret_dec env dec (k : V.value V.cont) =
     )
   | TypD _ ->
     k V.unit
+  | MixinD _ -> k V.unit
+  | IncludeD (_, _arg, _note) ->
+     (* TODO
```
crusso, line 1042: assert/fail, don't continue

src/pipeline/pipeline.ml
```diff
@@ -177,6 +177,7 @@ let async_cap_of_prog prog =
   match (CompUnit.comp_unit_of_prog false prog).it.body.it with
   | ActorClassU _ -> Async_cap.NullCap
   | ActorU _ -> Async_cap.initial_cap()
+  | MixinU _ -> Async_cap.initial_cap()
```
crusso, line 180: Maybe we should add a test, that you can't await from a mixin but call a system method.

src/pipeline/pipeline.ml
```diff
@@ -714,7 +715,12 @@ let run_files_and_stdin files =
 
 (* Desugaring *)
 
-let desugar_unit imports u name : Ir.prog =
+let desugar_unit imports u name : Ir.prog Diag.result =
+  match u.Source.it.Syntax.body.Source.it with
+  | Syntax.MixinU _ ->
+    let at = u.Source.it.Syntax.body.Source.at in
+    Diag.error at "M0250" "compile" "A mixin cannot be used as an entry point. It needs to be included in an actor (class)"
```
crusso: Can a mixin be included in a mixin? We could rule that out for now if problematic

src/mo_frontend/typing.ml
```diff
@@ -3427,6 +3469,16 @@ and infer_block_exps env decs : T.typ =
 and infer_dec env dec : T.typ =
   let t =
   match dec.it with
+  | IncludeD (i, arg, n) ->
+    if not env.pre then begin
+      use_identifier env i.it;
+      if not env.in_actor then
```
crusso, line 3477: Note to self: This is the place to reject includeD with mixins, if we want - maybe by just changing in_actor to in_obj_sort... (or adding in_mixin)

src/mo_frontend/typing.ml
```diff
@@ -3460,7 +3512,7 @@ and infer_dec env dec : T.typ =
       let ve0 = check_class_shared_pat env shared_pat obj_sort in
       let cs, tbs, te, ce = check_typ_binds env typ_binds in
       let env' = adjoin_typs env te ce in
-      let in_actor = obj_sort.it = T.Actor in
+      let in_actor = obj_sort.it = T.Actor || obj_sort.it == T.Mixin in
```
crusso, line 3517: So I guess this means we can include in another mixin

src/mo_frontend/typing.ml
```diff
@@ -3500,9 +3552,18 @@ and infer_dec env dec : T.typ =
...
     end;
     T.normalize t
+  | MixinD (args, dec_fields) ->
+    if not env.in_prog then
+      error env dec.at "M0253" "mixins may only be declared at the top-level";
+    let t_pat, ve = infer_pat_exhaustive error env args in
+    let env' = adjoin_vals env ve in
+    let obj_sort : obj_sort = { it = T.Mixin ; at = no_region; note = { it = true; at = no_region; note = () } }  in
+    let t' = infer_obj { env' with check_unused = false } obj_sort None dec_fields dec.at in
+    (* Printf.printf "MIXIN TYPE: %s\n" (T.string_of_typ t'); *)
```
```suggestion
```

src/mo_frontend/typing.ml
```diff
@@ -3588,17 +3647,27 @@ and gather_dec env scope dec : Scope.t =
...
       obj_env = obj_env;
+      mixin_env = scope.mixin_env;
       fld_src_env = scope.fld_src_env;
     }
-  | LetD (pat, _, _) -> gather_pat env scope pat
+  | LetD (pat, exp, _) -> (match is_mixin_import env exp.it with
+    | None -> gather_pat env scope pat
+    | Some (imports, args, t, decs) ->
+      match pat.it with
+      | VarP id -> Scope.adjoin scope (Scope.mixin id.it (imports, args, t, decs))
```
crusso, line 3669: So I guess this is a bit dubious because we are confusing the name space of values with the name space of mixins....but I can live with that.

src/mo_frontend/typing.ml
```diff
@@ -3809,6 +3912,7 @@ and infer_dec_valdecs env dec : Scope.t =
       typ_env = T.Env.singleton id.it c;
       con_env = T.ConSet.singleton c;
     }
+  | MixinD (_, _) -> Scope.empty
```
crusso, line 3919: This seems slightly dubious, but I guess a top-level mixin is still checked somewhere or nothing would work.

Should this be an assert if !env.in_prog?

src/mo_frontend/typing.ml
```diff
@@ -2994,6 +3030,8 @@ and object_of_scope env sort dec_fields scope at =
         | Some src ->
           Field_sources.add_src env.srcs src.id_region;
           T.{lab = id; typ = T.Typ c; src = {depr = src.depr; track_region = src.id_region; region = src.field_region}}::tfs
+        | _ when sort = T.Mixin ->
```
crusso, line 3033: Ok, I guess this is just adding all types, not just public ones, in the case of a mixin?

src/mo_frontend/typing.ml
```diff
@@ -3004,6 +3042,8 @@ and object_of_scope env sort dec_fields scope at =
         | Some src ->
           Field_sources.add_src env.srcs src.id_region;
           T.{lab = id; typ = t; src = {depr = src.depr; track_region = src.id_region; region = src.field_region}}::tfs
+        | _ when sort = T.Mixin ->
```
crusso, line 3045: Same here for values?

src/mo_frontend/typing.ml
```diff
@@ -3068,6 +3108,7 @@ and infer_obj env obj_sort exp_opt dec_fields at : T.typ =
   leave_scope env (private_identifiers scope.Scope.val_env) initial_usage;
   let (_, tfs) = T.as_obj t in
   if not env.pre then begin
+    (* TODO: Mixins need a similar, but different check here *)
```
crusso: Is this still to be implemented?

I think we may not have to worry about system methods as long as they enter the mixin environment - then duplicate declarations between the mix and actor will get caught.

crusso: I can't remember, did we check the system inspect method has the correct domain (including all the mixins?) I think we did...

src/mo_frontend/typing.ml
crusso, line 3114: So by my reading, pub_fields will include any `Include`d fields from mixins, so this should check everything...
for the final actor (if not a mixin itself).

crusso, line 3127: I guess this check doesn't include Included fields

crusso, line 3132: Ditto here.

src/mo_frontend/typing.ml
```diff
@@ -3042,7 +3082,7 @@ and is_typ_dec dec : bool = match dec.it with
 
 and infer_obj env obj_sort exp_opt dec_fields at : T.typ =
   let s = obj_sort.it in
```
crusso, line 3084: Did you consider may exanding out the dec_field to include all Included dec_fields before doing this check?

src/mo_interpreter/interpret.ml
```diff
@@ -1000,7 +1000,10 @@ and interpret_block env decs ro (k : V.value V.cont) =
 and declare_dec dec : val_env =
   match dec.it with
   | ExpD _
-  | TypD _ -> V.Env.empty
+  | TypD _
+  | MixinD (_) -> V.Env.empty
+  | IncludeD (_, _, _) -> (* TODO *)
```
crusso: Might be better to just assert/fail rather than continue

src/lowering/desugar.ml
```diff
@@ -925,27 +939,39 @@ and block force_unit ds =
...
     end
-  | S.VarD (i, e) -> I.VarD (i.it, e.note.S.note_typ, exp e)
-  | S.TypD _ -> assert false
+  | S.VarD (i, e) -> [I.VarD (i.it, e.note.S.note_typ, exp e)]
+  | S.TypD _ -> []
+  | S.MixinD _ -> []
+  | S.IncludeD(_, args, note) ->
+    let { imports = is; pat = p; decs } = Option.get !note in
+    let ir_imports = List.concat_map transform_import is in
+    let renamed_imports, rho = Rename.decs Rename.Renaming.empty ir_imports in
```
crusso, line 968: I suspect this might have failed big time with full fat stable functions

crusso: (the renamings of args captured by persistent functions)

src/lowering/desugar.ml
```diff
@@ -925,27 +939,39 @@ and block force_unit ds =
...
+    let renamed_pat, rho = Rename.pat rho (pat p) in
+
+    (* TODO: Fix the positions on the generated let here *)
```
crusso, line 971: What's to fix?

src/mo_frontend/parser.mly
christoph-dfinity: Yeah I didn't want to go with `mixin` and `persistent mixin`. I think if we're going to make a big release with implicits and mixins that might be a reasonable time to pull the trigger on persistent by default as well.

src/mo_frontend/static.ml
christoph-dfinity: Decided I'd just make this report the same error we get for `VarD`s in module, even if we weren't supposed to hit this case the behaviour is consistent.

src/lowering/desugar.ml
christoph-dfinity: Yeah that's what I was worried about, I didn't know exactly when the names were taken. During type checking we still had all information, but starting from IR down it was going to be very messy.

src/pipeline/pipeline.ml
christoph-dfinity: As you wrote a little test and it worked, so I'm inclined to just allow it for now, and see if someone manages to break it?🔥

src/mo_frontend/typing.ml
christoph-dfinity: Yes, the type of the MixinD ends up recording both public and private fields.

src/mo_frontend/typing.ml
christoph-dfinity: Yes, the type of the MixinD ends up recording both public and private fields.

src/mo_frontend/typing.ml
christoph-dfinity: Yeah, let's allow it for now.

src/mo_frontend/typing.ml
christoph-dfinity: Yeah, let's allow it for now.

timohanke:
@christoph-dfinity I don’t have a strong enough use case to justify taking on what would be a substantial amount of work.

The idea came from my `stream` package. Users can define a stream between two canisters and the message type is a type parameter. The receiver needs a public `receive` function which could be in a mixin instead of the user having to define it manually in its actor. The `receive` function would have a parametric type. But as mentioned before, having to define a single public function manually is not enough overhead to justify this feature.

Two other questions I came across:
1. Can a mixin have an implicit argument? Apparently not. I could not find a way to define the default value.
2. Can mixins be nested? Can one mixin include another one? Apparently not. The public functions defined in the inner mixin do not seem to make it to the actor top-level.

PR #5460, opened 2025-09-09
chore: update `flake.nix` to unpin `matchers`

ggreif:
`core` is now included, so run latest `motoko-matchers` with a `core` package.

While `matchers` internally uses `core`, it has been [retrofitted](https://github.com/dfinity/motoko-base/pull/736) to work with `base` too. So we can now test both `core` and `base` with `matchers`.


flake.nix
```diff
@@ -42,15 +42,15 @@
       flake = false;
     };
     motoko-base-src = {
-      url = "github:dfinity/motoko-base/next-moc";
+      url = "github:dfinity/motoko-base/pull/736/head";
```
ggreif: revert

flake.nix
```diff
@@ -42,15 +42,15 @@
...
       flake = false;
     };
     motoko-core-src = {
       url = "github:dfinity/motoko-core";
       flake = false;
     };
     motoko-matchers-src = {
-      url = "github:kritzcreek/motoko-matchers/5ba5f52bd9a5649dedf5e2a1ccd55d98ed7ff982";
+      url = "github:kritzcreek/motoko-matchers/pull/19/head";
```
ggreif: unpin after merge

nix/base-tests.nix
```diff
@@ -1,3 +1,4 @@
+
```
ggreif:
```suggestion
```

flake.nix
ggreif:
```suggestion
      url = "github:dfinity/motoko-base/next-moc";
```

PR #5464, opened 2025-09-10
feat: Invariance limitations improvements

Kamirus:
1. When choosing between type `T` and `Any`, pick `T` when it has no proper supertypes (except `Any`).
Previously we needed the `T` to be *isolated* (no proper subtypes nor supertypes).
This change allows `Int` and `?T` as well! (Assuming `T` has no supertypes).
See the tests for more examples.

Likewise, when choosing between `Non` and `T`, pick `T` when it has no proper subtypes (except `Non`).

2. Error message improvements:
- We should start to suggest writing an explicit _type instantiation_. See the tests for examples.
- Start suggesting a concrete type instantiation so that it's easier for developers and AI -- they would just need to copy and paste it, perhaps modifying it slightly. This is better than making the compiler make the arbitrary choice. It would just suggest a solution. I think this might be a middle ground between developer experience and forcing the compiler to make an arbitrary choice (when there is no principal solution)
- For example:
```
type error [M0098], cannot apply ...
...
Hint: Add explicit type instantiation, e.g. <Nat, Text>
```
- Outdated:
~~I think the error message `because implicit instantiation of type parameter R is under-constrained with` is a bit cryptic, maybe we can find a better alternative that explains that there is no best solution and the compiler won't make an arbitrary choice...~~

<details>
<summary>The description below is no longer up to date, the error messages have been simplified. See the Changelog entry for an accurate description.</summary>

## For example (error message improvement)
```motoko
let _ = VarArray.map(va, func x = x); // Nat
```
### Before
```
cannot implicitly instantiate function of type
  <T, U>([var T], T -> U) -> [var U]
to argument of type
  ([var Nat], Nat -> U)
because implicit instantiation of type parameter U is under-constrained with
  Nat  <:  U  <:  Any
where
  Nat  =/=  Any
so that explicit type instantiation is required
```

### After
```
... cannot solve invariant type parameter U, no principal solution with
  Nat  <:  U  <:  Any
where
  Nat  =/=  Any
Add a return type annotation or an explicit type instantiation
Suggested return type annotation : 
  [var Nat]
```

In the future we could integrate it into the vscode extension to provide a 'quick fix' that would insert this type annotation.
**Limitations:** The return type might be too long, we would not use type aliases, maybe we should avoid the suggestion when the type is longer than X characters.
</details>

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -35,6 +35,9 @@
 
     Note that the error message suggests only one possible solution, but there may be alternatives. In the example above, `<Nat, Int>` would also be a valid instantiation.
 
+  * Fixes type inference of deferred funcs that use `return` in their body (#5615).
+    Avoids confusing errors like `Bool does not have expected type T` on `return` expressions. Should type check successfully now.
+
   * Add warning `M0239` that warns when binding a unit `()` value by `let` or `var` (#5599).
 
   * Use `self` parameter, not `Self` type, to enable contextual dot notation (#5574).
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -19,7 +19,6 @@ module S = Set.Make(String)
 type avl = Available | Unavailable
 
 type lab_env = T.typ T.Env.t
-type ret_env = T.typ option
 type val_info = T.typ * Source.region * Scope.val_kind * avl
 type val_env  = val_info T.Env.t
 
@@ -66,6 +65,10 @@ type env =
     errors_only : bool;
     srcs : Field_sources.t;
   }
+and ret_env =
+  | NoRet
+  | Ret of T.typ
+  | BimatchRet of (env -> exp -> unit)
 
 let env_of_scope ?(viper_mode=false) msgs scope =
   { vals = available scope.Scope.val_env;
@@ -75,7 +78,7 @@ let env_of_scope ?(viper_mode=false) msgs scope =
     cons = scope.Scope.con_env;
     objs = T.Env.empty;
     labs = T.Env.empty;
-    rets = None;
+    rets = NoRet;
     async = Async_cap.NullCap;
     in_actor = false;
     in_prog = true;
@@ -1908,7 +1911,7 @@ and infer_exp'' env exp : T.typ =
       let env'' =
         { env' with
           labs = T.Env.empty;
-          rets = Some codom;
+          rets = Ret codom;
           (* async = None; *) }
       in
       let initial_usage = enter_scope env'' in
@@ -1998,7 +2001,7 @@ and infer_exp'' env exp : T.typ =
       check_ErrorCap env "try" exp.at;
       if cases <> [] then
         coverage_cases "try handler" env cases T.catch exp.at;
-      Option.iter (check_exp_strong { env with async = C.NullCap; rets = None; labs = T.Env.empty } T.unit) exp2_opt
+      Option.iter (check_exp_strong { env with async = C.NullCap; rets = NoRet; labs = T.Env.empty } T.unit) exp2_opt
     end;
     T.lub ~src_fields:env.srcs t1 t2
   | WhileE (exp1, exp2) ->
@@ -2060,11 +2063,13 @@ and infer_exp'' env exp : T.typ =
   | RetE exp1 ->
     if not env.pre then begin
       match env.rets with
-      | Some T.Pre ->
+      | Ret T.Pre ->
         local_error env exp.at "M0084" "cannot infer return type"
-      | Some t ->
+      | Ret t ->
         check_exp_strong env t exp1
-      | None ->
+      | BimatchRet k ->
+        k env exp1
+      | NoRet ->
         local_error env exp.at "M0085" "misplaced return"
     end;
     T.Non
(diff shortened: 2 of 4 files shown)


src/mo_frontend/bi_match.ml
```diff
@@ -132,24 +136,27 @@ let fail_open_bound c bd =
...
   | Variance.Covariant -> lb
   | Variance.Contravariant -> ub
   | Variance.Bivariant -> lb
   | Variance.Invariant ->
     match normalize lb, normalize ub with
     (* Ignore [Any] when choosing a bound for the solution *)
-    (* Restrict to [isolated] types only, at least for now *)
-    | t, Any when isolated t ->
+    (* When the solution is between [t] and [Any], choose [t] when there are no other choices except [Any] *)
+    | t, Any when has_no_supertypes t ->
```
crusso, line 185: Can you do the symmetric thing for `Non, t when has_no_subtypes t`

src/mo_frontend/bi_match.ml
```diff
@@ -132,24 +136,27 @@ let fail_open_bound c bd =
...
       assert (t <> Non);
       lb
     | _ ->
-      raise (Bimatch (Format.asprintf
-        "implicit instantiation of type parameter %s is under-constrained with%a\nwhere%a\nso that explicit type instantiation is required"
-        (Cons.name c)
-        display_constraint (lb, c, ub)
-        display_rel (lb,"=/=",ub)))
+      if !error_msg = "" then
```
crusso: Why only report the first error? To avoid information overload?

src/mo_types/type.ml
```diff
@@ -1480,22 +1480,25 @@ and singleton_field co tf = singleton_typ co tf.typ
...
+  | NoSubtypes
+  | NoSupertypes
+
+let rec has_no_subtypes_or_supertypes m co = function
   | Con (c, ts) as t ->
     (match Cons.kind c with
     | Abs (_, bound) ->
-      (* Type parameters are isolated when unbounded. *)
+      (* Type parameters are isolated (no subtypes nor supertypes) when unbounded. *)
       bound = Any
```
crusso, line 1496:
```suggestion
      m = NoSubtypes || bound = Any
```

Otherwise T  :> Any has subtypes, right?

src/mo_frontend/bi_match.ml
```diff
@@ -45,6 +46,8 @@ type ctx = {
   bounds : typ ConEnv.t * typ ConEnv.t;
   (* Variances for type variables *)
   variances : Variance.t ConEnv.t;
+  (* Optional return type *)
+  typ_opt : typ option;
```
christoph-dfinity: Very minor nit, but I'd prefer something more descriptive like `ret_ty` here.

src/mo_frontend/bi_match.ml
Kamirus: We could, yes. I just did not see enough examples in practice that would benefit from that. So I wanted to limit this heuristic for now

src/mo_frontend/bi_match.ml
Kamirus: To retain the original behavior. We could gather all of them and report all. I just needed to finish the building of `env` so that I have the substitution.

src/mo_types/type.ml
Kamirus: I think I agree with the addition of `m = NoSubtypes || `, which means that a type variable has no proper subtypes.

My thinking was:
- A type variable `T` have a `bound` defined (default is `Any`)
- This means that `T <: bound`, T is a subtype of bound
- That is true that `T` has no proper subtypes (except `Non`), it has the `bound` as its supertype (which we ignore when its `Any`)

src/mo_frontend/bi_match.ml
Kamirus: that is a good suggestion! thanks

src/mo_frontend/bi_match.ml
Kamirus: But I guess it makes sense to make it symmetric. the `has_no_subtypes` is already ignoring the bottom type `Non` so it should be consistent.
I'll add some examples covering that. I guess something returning a function that takes a mutable data structure 😄

src/mo_frontend/bi_match.ml
Kamirus: OK, I've refactored and now we gather all 'under-constrained' and merge into a single error message, [see the test output here](https://github.com/dfinity/motoko/pull/5464/files#diff-f899dc59249826cb5cc0b79fd521ab48213a2b36cf3577e5f53a33cc53dd48b6R100-R112)

Kamirus:
@crusso @christoph-dfinity Hm, perhaps now with dot-notation we might actually prefer 'type instantiations' over 'return type annotations' because it is easier to write them with a dot-chain of functions like
```motoko
list
  .map<X,Y>(func x = ...)
  .filter(func y = ...)
  ...
```

Though we could probably easily allow partial type instantiations like
```motoko
list
  .map<_,Y>(func x = ...)
```

Because only the 2nd type parameter of map on a mutable list is essential to provide

PR #5466, opened 2025-09-11
feat: selectively suppress warning on dropped fields in migration function using named types

### LINKED ISSUE #5301: FR: add a way to suppress warning on migration dropping field
```
WARN: /home/runner/work/motoko-benchmark/motoko-benchmark/benchmarks/application/tmp/car-sharing-upgrade/main.mo:11.7-11.32: warning [M0207], migration expression consumes field `minCarCapacity` of type
  Nat
but does not produce it. The field is not declared in the actor.
This field will be removed from the actor, discarding its consumed value.
If this removal is unintended, declare the field in the actor and either remove the field from the parameter of the migration function or add it to the result of the migration function.
```

There is no way to suppress this warning, even if discard is intentional.

We could maybe look for an explicit comment on the field to suppress the warning.

/// @deleted
var x : Nat (edited) 

or use a specially named type  

var x : (deleted: Nat)

Suppressing all these warnings is probably a bad idea, since some data may be unintentionally lost.

crusso:
closes https://github.com/dfinity/motoko/issues/5301

NB: this is probably better than just disabling these warnings entirely since it forces users to consider each field on a case by case basis.

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -3141,7 +3141,7 @@ and check_migration env (stab_tfs : T.field list) exp_opt =
    (* This may indicate unintentional data loss. *)
    let is_named typ = T.(match typ with
         | Mut (Named _)
-        | Named (_, _) -> true
+        | Named _ -> true
         | _ -> false)
    in
    List.iter (fun T.{lab;typ;src} ->
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -3139,6 +3139,11 @@ and check_migration env (stab_tfs : T.field list) exp_opt =
      rng_tfs;
    (* Warn about any field in domain, not in range, and declared stable in actor *)
    (* This may indicate unintentional data loss. *)
+   let is_named typ = T.(match typ with
+        | Mut (Named _)
+        | Named (_, _) -> true
+        | _ -> false)
+   in
    List.iter (fun T.{lab;typ;src} ->
      match typ with
      | T.Typ c -> ()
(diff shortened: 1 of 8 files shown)


src/mo_frontend/typing.ml
```diff
@@ -3139,6 +3139,11 @@ and check_migration env (stab_tfs : T.field list) exp_opt =
      rng_tfs;
    (* Warn about any field in domain, not in range, and declared stable in actor *)
    (* This may indicate unintentional data loss. *)
+   let is_named typ = T.(match typ with
+        | Mut (Named _)
+        | Named (_, _) -> true
```
ggreif:
```suggestion
        | Named _ -> true
```

ggreif (approved):
LGTM, with a nit. A bit of a hack, but justified if we want caffeine to work smoother. It appears that a type-level sublanguage is emerging for warning suppression reasons.

src/mo_frontend/typing.ml
```diff
@@ -3139,6 +3139,11 @@ and check_migration env (stab_tfs : T.field list) exp_opt =
      rng_tfs;
    (* Warn about any field in domain, not in range, and declared stable in actor *)
    (* This may indicate unintentional data loss. *)
+   let is_named typ = T.(match typ with
+        | Mut (Named _)
+        | Named _ -> true
```
Kamirus, line 3144: If we do this I'd avoid warning only on a specific name, e.g.
"dropped" or "drop" or "nowarn" or sth like this. Just to avoid accidentally suppressing the warning

crusso: One could also use the warning number...

PR #5472, opened 2025-09-12
chore: use a larger worker for `linux` CI

alexandru-uta:
CI was stuck on linux x86, this PR adds a new worker type for the more resource intensive tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -138,7 +138,7 @@ jobs:
     needs: tests
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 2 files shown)


.github/actions/test-blueprint/action.yml
```diff
@@ -35,7 +35,7 @@ runs:
   using: "composite"
   steps:
     - name: Free Disk Space (Ubuntu)
-      if: startsWith(inputs.os, 'ubuntu-')
+      if: inputs.os == 'ubuntu-latest' || inputs.os == 'ubuntu-24.04-arm'
```
ggreif, line 38: this is crazy why doesn't this work?

PR #5477, opened 2025-09-12
feat: CI Optimization

alexandru-uta:
In this PR we:
* Add larger runners for all tests that were running on default ubuntu x86.
* Move `macos-latest` target to nightly because it's very expensive to run on all PR checks.
* Remove `macos-13` from nightly because github will remove them soon and will become unsupported.
* Various small job optimizations.


.github/workflows/nightly-macos-test.yml
```diff
@@ -6,45 +6,45 @@ on:
   
 jobs:
   common-tests:
-    runs-on: macos-13
+    runs-on: macos-latest
```
ggreif, line 9: Removing `macos-13` means we won't provide x64 macOS binaries with our releases any more. Will `dfx` cope?

ggreif: Or maybe we build them under Rosetta emulation.

ggreif: Oh, there may be another way: See https://github.com/dfinity/sdk/pull/4379/commits/41fe9b142a04629d43616fb0e4edcb09ed7317ee

alexandru-uta: There's 2 things:
1. this nighlty job just tests macos-13, doesn't have much to do with the release; we can still do that
2. macos-13 will be completely unsupported/deprecated by github, so we won't be able to offer it anyhow at some point in the near future

ggreif (approved):
LGTM!

PR #5480, opened 2025-09-12
experiment: batch upgrades

crusso:
to build the driver canister:
`moc driver.mo --package ic /path/to/mops-ic/src`

The latter part could be `npx ic-mops sources ic | grep -- '--package ic '`


ggreif:
This experiment can be pretty much closed, as a proper multi-migration architecture is in place. And evolving.

PR #5484, opened 2025-09-15
Fix `persistent func` typing to compile motoko-core

Kamirus:
1. Fixes typing of `persistent func` to get `motoko-core` (object-oriented branch) to compile ([see the successful test run here](https://github.com/dfinity/motoko-core/pull/393))
2. Makes the typing consistent between eop and non-eop. Removes implicit conversion from `persistent func` to `func` in non-eop setting. This is probably less confusing route. We could revert to the "implicit conversion" way but we would need to add more of these conversions in the compiler where they were missing...
3. Changes were pre-released ([here](https://github.com/dfinity/motoko/releases/tag/0.16.1-experimental-stable-functions-2)) and used to test the motoko-core object-oriented branch ([here](https://github.com/dfinity/motoko-core/pull/393))

### DIFF
--- a/test/run-drun/ok/persistent-actor-with-persistent-funcs-in-type-args-and-return.drun-run.ok
+++ b/test/run-drun/ok/persistent-actor-with-persistent-funcs-in-type-args-and-return.drun-run.ok
@@ -1,0 +1,2 @@
+ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
+ingress Completed: Reply: 0x4449444c0000


flake.nix
```diff
@@ -46,7 +46,7 @@
       flake = false;
     };
     motoko-core-src = {
-      url = "github:dfinity/motoko-core";
+      url = "github:dfinity/motoko-core/luc/oo-data-structures";
```
Kamirus: to be removed

src/mo_frontend/bi_match.ml
```diff
@@ -261,7 +261,7 @@ let bi_match_typs ctx =
     | Tup ts1, Tup ts2 ->
       bi_match_list bi_match_typ rel eq inst any ts1 ts2
     | Func (s1, c1, tbs1, t11, t12), Func (s2, c2, tbs2, t21, t22) ->
-      if s1 = s2 && c1 = c2 then
+      if rel_sort s1 s2 tbs1 && c1 = c2 then
```
crusso: Minor tweak - I think rel_sort should take either `rel` and `eq` and require equal sorts when `rel == eq` (or a boolean flag `is_invariant` set by `(rel == eq)`

Basically, when rel == eq you are comparing equality, otherwise subtyping.

test/fail/persistent-func-type-argument-and-return.mo
```diff
@@ -0,0 +1,8 @@
+module {
+  persistent func g<T>(x : T) : persistent() -> T {
+    persistent func h() : T { x };
+    h; // TODO: non-eop should still typecheck here
+  };
+  // don't allow `func() {}` to typecheck as `persistent() -> ()` for now
```
crusso, line 6: Yep, I think anonymous lambdas can never be persistent in Luc's design

Kamirus: They need a name, I have it in another test

src/mo_frontend/bi_match.ml
Kamirus:
```motoko
| Func (s1, c1, tbs1, t11, t12), Func (s2, c2, tbs2, t21, t22) ->
  (rel_sort s1 s2 tbs1 || incompatible_func_sorts d t1 t2) &&
  (c1 = c2 || incompatible_bounds d t1 t2) &&
  (match rel_binds d eq eq tbs1 tbs2 with
    | Some ts ->
      rel_list "function parameters" d rel_typ rel eq (List.map (open_ ts) t21) (List.map (open_ ts) t11) &&
      rel_list "return types" d rel_typ rel eq (List.map (open_ ts) t12) (List.map (open_ ts) t22)
    | None -> incompatible_funcs d t1 t2
  )
```
If so then we need to update the original use of rel_sort (mentioned above), yes?

crusso: Yeah, I think rel_sort was broken and only worked for subtyping, not equality. I guess you could also just require rel_sort in both direction when rel == lub, if you prefer.

test/fail/persistent-func-type-argument-and-return.mo
crusso: I think even just a name ain't enough. The name needs to be unique in the path (i.e. a declaration in a persistent context).

src/mo_frontend/bi_match.ml
Kamirus: Ok, fixed! Now both use the `rel == eq`

test/fail/persistent-func-type-argument-and-return.mo
Kamirus: But this test is passing https://github.com/dfinity/motoko/pull/5484/files#diff-fcef2042739c68f514af2918041769a33351fe675b46e578e11c72d23c210b06R7

christoph-dfinity: I think this is the same issue as the `B.mo` or the `C.mo` file in here: https://github.com/dfinity/motoko/pull/5354/files#diff-0e9e9083aa4cfa901cd7251a640e3a9b3f95eb13e0b10b34ed5bfeaa6b20ab1c

PR #5485, opened 2025-09-15
chore: delete `nix/emscripten-fix.patch`

ggreif:


ggreif:
> Is there anything I need to verify other than CI passing here?

I don't think so. We used to build RTS with a modified `emscripten`, but now `cargo` can build it.

christoph-dfinity (approved):

PR #5491, opened 2025-09-16
chore: Improve draft release workflow to enable manual prereleases

Kamirus:
Improves the draft release workflow to produce valid artifacts with correct tag.
Can be used to create pre-releases that work well with `mops toolchain moc`.

Includes a script that allows using the custom built moc with mops.
- builds local custom `moc`
- replaces it in mops (saving the original as backup)
- can restore the original
**Use case:** The script can be used to locally test the custom `moc` by running `mops test` in other projects, e.g. `motoko-core`. Without it we would need a `moc` [pre]release

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "candid-src": {
       "flake": false,
       "locked": {
-        "lastModified": 1756934685,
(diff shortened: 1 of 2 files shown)


christoph-dfinity (approved):
Nice, thank you!

PR #5468, opened 2025-09-11
feat: Redundant type instantiation warning

Kamirus:
Adds a warning for redundant type instantiations. Allowed (disabled) by default. Can be enabled via `-W M0223` moc flag.

### Perf tests
No measurable change comparing this feature branch to master using `time vessel verify`.
There are total of **3012** warnings now, **1338** redundant type instantiation warnings. Well... that is a lot of new warnings 

### Impact
Thanks to this check we can measure the impact of the 'improved type inference' (2-phase inference that defers func arguments + invariant improvements). Running `vessel verify` yields the following:
1. We used to have **907** redundant instantiations (before the 'improved type inference')
2. Which increased by **+431** (**+48%**) due to 2round inference and invariant improvements with isolated types (both contributing almost equally)

PR #5496, opened 2025-09-18
feat: Allow disabling/enabling selected warnings by message codes

Kamirus:
Notes:
- changelog entry includes the future warning code from the #5468 
- unfortunately we mix error and warning messages codes, some codes are used for both errors and warnings, etc. This complicates the definition of "warnings" when using error/message codes... For now I'm not addressing this as this would require a bigger refactor. As a result there is no validation whether the given message code via `-A`/`-W` is actually a warning. If that happens to be an error message code then it has no effect, as errors cannot be suppressed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
   common-tests:
     strategy:
       matrix:
-        os: [ ubuntu-24-large, ubuntu-24.04-arm ]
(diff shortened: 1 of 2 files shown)


crusso (commented):
Actually, we should probably document these in doc/md/15-compiler-ref.md (or wherever it lives).

Just add them to the table.

crusso (approved):

PR #5497, opened 2025-09-18
fix: Fix `test-runner` tests

alexandru-uta:
Run `test-runner` tests as part of building the package instead of separate tests. Simplifies nix code.


crusso (approved):
RS

PR #5506, opened 2025-09-22
feat: Extra moc flags in vscode extension

Kamirus:
Enables passing some of the `moc` flags in `moc.js` to be used in the vscode extension


src/js/common.ml
```diff
@@ -5,6 +5,11 @@ open Mo_config
 module Js = Js_of_ocaml.Js
 module Sys_js = Js_of_ocaml.Sys_js
 
+(** Extra flags from the [moc] CLI *)
+let moc_args = Mo_args.inclusion_args
+  @ Mo_args.warning_args
+  @ Mo_args.error_args
```
Kamirus, line 11: Any other flags we should include here?

src/mo_config/args.ml
```diff
@@ -1,6 +1,9 @@
 (* This module contains some argument parsing that is common between
 multiple executables *)
 
+(** suppress documentation *)
+let _UNDOCUMENTED_ doc = "" (* TODO: enable with developer env var? *)
```
rvanasa, line 5: Maybe some magic string such as `"<undocumented>"` for clarity?

src/js/common.ml
rvanasa: The current set of flags looks good for now.

rvanasa (approved):
LGTM with a few comments.

src/mo_config/args.ml
Kamirus: I wanted to preserve the original behavior. We can revisit that in another PR.
Changing it here would change the behavior of the `moc --help`. I guess there is some mechanism that skips options when undocumented so we'd have to change that too.

PR #5510, opened 2025-09-24
experiment: add dedup hook

crusso:


src/codegen/compile_enhanced.ml
```diff
@@ -13607,8 +13620,14 @@ and main_actor as_opt mod_env ds fs up =
 
     (* Compile the declarations *)
     let ae2, decls_codeW = compile_decs_public env ae1 ds v2en
-      Freevars.(captured_vars (system up))
+        Freevars.(captured_vars (system up))
+    in
+    let _ = match VarEnv.lookup_var ae2 "@add_cycles" (* CHANGE ME *) with
```
crusso, line 13625: You'll need to change the string to reference "@dedup"

PR #5517, opened 2025-09-26
feat: add support for implicit arguments

crusso:
* Add support for `implicit` argument declarations (#5517).

     Function and class parameters can be declared *implicit* using a specially named type of the form:

     * `implicit : <typ>` (implicit named or anonymous parameter).
     * `implicit : (<id> : <typ>)` (implicit parameter matching name `<id>` and `<typ>`).
     * `implicit : (_ : <typ>)` (implicit parameter matching `<typ>` under any name).

     This is just a special case of a named type in Motoko with special name "implicit".

     The arguments corresponding to implicit parameters can be omitted at call sites, provided the compiler can unambiguously resolve using a declaration of the given name and matching type, in either the current scope or the fields of a module in the current scope. By default, the name is determined by the parameter declared `implicit` but can be specified to be another name,
     `<id>`, using the second form of `implicit` type annotation above (see `Pair.toText` below for an example).
     An unnamed parameter or a named parameter declared with the third form of anonymous declaration matches any name in the context that has a matching type only.

     Any candidate from the current scope is preferred over candidates from nested scopes.

     In the rare case that a function has more than one implicit parameter, either all arguments must be specified or all implicit arguments omitted.

     Functions that require two implicits of the same (external) `name` can declare them with different argument names (to distinguish their implementations), and common (external) names, using a nested name.

     Example:
     ```motoko
     module Nat {
       public func toText(n : Nat) : Text { debug_show n };
     };
     module Array {
       public func toText<T>(as : [T], toText: (implicit : T -> Text)) : Text {
          var t = "";
          for (a in as.vals()) {
            t := t # (toText(a));
          };
          t
       }
     };
     module Pair {
       public func toText<T, U>(
         p : (T, U),
         toTextT : (implicit : (toText : T -> Text)),
         toTextU : (implicit : (toText : U -> Text)))
         : Text {
           "(" # toTextT(p.0) # "," # toTextU(p.1) # ")"
         };
     };
     Array.toText([1,2,3], Nat.toText) // explicit arguments
       |> debugPrint _;
     Array.toText([1,2,3]) // implicit arguments
       |> debugPrint _;
     Pair.toText((1,2))
       |> debugPrint _; // implicit arguments
     ```


crusso:
Recent commits lock good @christoph-dfinity!

src/mo_frontend/typing.ml
```diff
@@ -1281,6 +1282,136 @@ and combine_pat_srcs env t pat : unit =
...
+let disambiguate_resolutions (candidates : hole_candidate list) =
+  let add_candidate (frontiers : hole_candidate list) (c : hole_candidate) =
+    let rec go (fs : hole_candidate list) = match fs with
+      | [] -> [c]
+      | f::fs' ->
+         if T.sub c.typ f.typ then
+           if T.eq c.typ f.typ then
+             f :: go fs'
+           else
+             fs
```
crusso, line 1329:
```suggestion
             fs
```
Is this correct or a typo? My hunch is that it should maybe be `go fs'`? But I'm only going by smell.
I guess this branch is discarding `f` because `c` is more specific, but why not consider the remaining candidates in `fs'`

christoph-dfinity: It's actually the other way around. Here's how I explained it to myself:

All candidates are subtypes of the required type. The greatest (most super) of these types is the "closest" to the required type. If we can uniquely identify a single candidate that is the supertype of all other candidates we pick it.

```ocaml
let disambiguate_resolutions (candidates : hole_candidate list) =
  let add_candidate (frontiers : hole_candidate list) (c : hole_candidate) =
    let rec go (fs : hole_candidate list) = match fs with
      | [] -> [c]
      | f::fs' ->
         if T.sub c.typ f.typ then
           if T.eq c.typ f.typ then
             (* c = f, so we keep both *)
             f :: go fs'
           else
             (* c <: f, so f absorbs c *)
             fs
         else if T.sub f.typ c.typ then
           (* f <: c, so c absorbs f *)
           go fs'
         else
           (* no relation at all, so we keep both *)
           f :: go fs'
    in
    go frontiers
  in
  match List.fold_left add_candidate [] candidates with
  | [dom] -> Some dom
  | _ -> None
```

If you think that's too complicated I'm totally happy to go with "prefer candidates that are equal to the required type".

Changelog.md
```diff
@@ -1,5 +1,57 @@
 # Motoko compiler changelog
 
+ * motoko (`moc`)
+
+   * Add support for `implicit` argument declarations (#5517).
+
+     Function and class parameters can be declared *implicit* using a specially named type of the form
```
crusso, line 7:
```suggestion
     Function and class parameters can be declared *implicit* using a specially named type of the form:
```

test/fail/implicit-bug.mo
```diff
@@ -0,0 +1,29 @@
...
+
+func compare(n : Nat, m: Nat) : Int { 0 };
+
+let map : Map.Map<Nat,Text> = Map.Map();
+
+func test() {
+
+  let t : Nat = 0;
+
+//  let _ = map.get(t);
```
Kamirus: there is no bug anymore, right?

crusso: I think not. Let me rewrite the test.

Kamirus (approved):
LGTM!

PR #5521, opened 2025-09-27
experiment: add `implicit` keyword and dedicated grammar

crusso:
builds on #5517 

* add `implicit` keyword and grammer production (so a minor breaking change)

* expands `implicit t` to a named type. Since `implicit` is now  a keyword, users can't just use *identifier* `implicit` as name for a traditional named type (or anything else for that matter).

* extends pretty printer, parser and various other bits and bobs.

Let's you write and read:
```
func sort<T>( array : [T], compare : implicit (Nat, Nat) -> Order) {};
: <T>( array : [], compare : implicit (Nat, Nat) -> Order) -> ()
```
instead of raw syntax and types:
```
func sort<T>( array : [T], compare : (implicit : (Nat, Nat) -> Order)) {};
: <T>( array : [], compare : (implicit : (Nat, Nat) -> Order)) -> ()
```
(fewer `(..)` and `:` )

- [x] merge with master
- [x] fix merge (regen tests)
- [x] support old and new syntax (so we can (but don't need to) rewrite core or break old clients of core)


ggreif:
I'd suggest
``` Motoko
func sort<T>( array : [T], compare : implicit (Nat, Nat) -> Order) {};
: <T>( array : [], with compare : (Nat, Nat) -> Order) -> ()
```

- spares us from introducing a new keyword
- (to me at least) it reads naturally "with `compare : (Nat, Nat) -> Order` from lexical context"
- the keyword is right at the beginning
- most programmers will regard `implicit` as a very exotic word

test/fail/ok/implicit-import-hint.tc.ok
```diff
@@ -1,3 +1,3 @@
 implicit-import-hint.mo:11.8-11.18: type error [M0230], Cannot determine implicit argument 'compare' of type
   (Text, Text) -> Order
-Hint: Did you mean to import mo:core/Text?
+Hint: Did you mean to import core-stub/src/Text.mo.compare?
```
crusso: A regression?

test/fail/ok/implicit-import-hint-dot.tc.ok
```diff
@@ -11,8 +10,8 @@ and
   {var root : Node<Text, Text>; var size_ : Nat}  <:  (map : Map<K, V>)
 implicit-import-hint-dot.mo:17.20-17.25: type error [M0230], Cannot determine implicit argument 'compare' of type
   (Text, Text) -> Order
-Hint: Did you mean to import mo:core/Text?
+Hint: Did you mean to import core-stub/src/Text.mo.compare?
 implicit-import-hint-dot.mo:18.5-18.12: type error [M0072], field compare does not exist in type:
   {chars : () -> {next : () -> ?Char}; size : () -> Nat}
 The field compare is not available. Try something else?
-Hint: Did you mean to import mo:core/Text?
+Hint: Did you mean to import core-stub/src/Text.mo?
```
crusso: A regression?

Kamirus (approved):
LGTM

PR #5527, opened 2025-09-29
Components: Cleanups

Kamirus:


src/codegen/compile_classical.ml
```diff
@@ -14065,31 +13992,10 @@ let compile mode rts (prog : Ir.prog) : Wasm_exts.CustomModule.extended_module =
...
-        let wit_file_content = (imported_components_to_wit !imported_components) in
-        (* Generate the WAC file to compose the components*)
-        let wac_file_content = (imported_components_to_wac !imported_components) in
-        Some wit_file_content, Some wac_file_content
-      end
-    else
-      None, None
+  let wit_wac = 
+    if !Flags.wasm_components
+    then Some (Imported_components.generate_wit_wac (E.add_func_import env) prog)
```
przydatek: nit: maybe rename `generate_wit_wac` to something like `add_imports_and_generate_wit_wac`, as the "side-effect" of adding the imports is quite relevant here, and with the current name is "invisible".

src/wasm-exts/customModuleDecode.ml
```diff
@@ -1010,8 +1010,7 @@ let module_ s =
     candid;
     source_mapping_url = None;
     wasm_features = wasm_features;
-    wit_file_content = None;
-    wac_file_content = None;
+    wit_wac = None;
```
przydatek, line 1013: nit: I'd call the field  (and the corresponding variable elsewhere) `maybe_wit_wac`, to make it clear that it is an Option-type, but it seems that such a convention is not in use in the code base (cf.`source_mapping_url` above), so I guess it is ok to leave as-is ... 🤔

przydatek (approved):
A very nice refactoring, thanks!

test/wasm-components-test-project/ok/1_compile_motoko_with_components.stdout.ok
```diff
@@ -1,7 +1,11 @@
 --- Compiling Motoko that imports components but without -wasm-components flag...
-Wasm components support not enabled, use -wasm-components
+OOPS! You've triggered a compiler bug.
+Please report this at https://github.com/dfinity/motoko/issues/new with the following details:
+
+Motoko (source XXX)
```
przydatek: this line seems to cause problems with CI... 🤔

src/codegen/compile_classical.ml
```diff
@@ -12330,10 +12261,7 @@ and compile_prim_invocation (env : E.t) ae p es at =
 
   | ComponentPrim (_, component_name, function_name, arg_types, return_type), es ->
     if not !Flags.wasm_components then
-        begin
-          Wasm.Sexpr.print 80 (Wasm.Sexpr.Atom ("Wasm components support not enabled, use -wasm-components"));
-          exit 0
-        end;
+      failwith "Wasm components support not enabled, use -wasm-components";
```
przydatek, line 12264: nit: How about reverting this change (and undeleting the corresponding test for the flag), so that we're sure that the flag is working?  It seems to me that `failwith` is a bit misleading here (as the error message says that one has found a bug in the compiler, which is not accurate in this case), yet having some feedback about about the failure is useful.  (no strong opinion, if you prefer as-is, fine for me)

przydatek (approved):

PR #5531, opened 2025-09-30
test: mixins

crusso:
Adds some more tests to mixins PR #5459

* tests mixin capabilities in their body (a bit more restrictive than actor declarations)
* tests nested includes
* test system inspect message has an appropriate domain
* test this parameter of including actor affected.

All are static tests. I guess we could test compilation too.

Summary:
* looks like mixins can't declare system methods or use <system> capability, but could. Relax in future?
* We get warnings about unused privates from the mixin. I guess that makes sense, but maybe one could suppress them too.     (Probably enough to just mark them all used at the include?)
* nested include seems to work fine. Haven't tried cyclic ones, but I think import resolution will just catch that.

### DIFF
--- a/test/fail/mixins/Mixin1.mo
+++ b/test/fail/mixins/Mixin1.mo
@@ -1,0 +1,5 @@
+mixin() {
+  var f1 = 0;
+  public func m1() : async () { };
+  system func preugprade() {}; // reject system function in mixin
+};


test/fail/mixins/Mixin.mo
```diff
@@ -0,0 +1,5 @@
+mixin() {
+  var f1 = 0;
+  public func m1() : async () { };
+  system func preugprade() {};
```
crusso:
```suggestion
  system func preugprade() {}; // rejected (could be accepted?)
```

christoph-dfinity:
Thank you for writing these!

PR #5533, opened 2025-10-01
chore: refactor `Lib.List.safe_map` to use TMC

ggreif:
resolving a TODO

~~Sadly `coverage` builds/tests are breaking, as they use instrumentation that cannot deal (?) with this annotation.~~ — a9280eb12 disables instrumentation (on that particular `let`), so that fixes it.

Background: https://ocaml.org/manual/5.1/tail_mod_cons.html


src/lib/lib.ml
```diff
@@ -431,8 +431,11 @@ struct
        | hd' :: tl' -> equal hd hd' && is_prefix equal tl tl')
   
   (* tail-recursive map *)
-  (* TODO: can be replaced by [@tail_mod_cons] map (OCaml 0.4.14)/List.map once we upgrade OCaml *)
-  let safe_map f l = List.rev (List.rev_map f l)
+  let[@tail_mod_cons] rec safe_map f l = match l with
+    | [] -> []
+    | x :: xs ->
+      f x :: (safe_map[@tailcall]) f xs
+    [@@coverage off]
```
Kamirus, line 438: Why is this annotation here? Do we use some code coverage tools?

Kamirus: Ah yeah, without it the `make DUNE_OPTS="--instrument-with bisect_ppx" -C src moc` fails.
It probably tries to rewrite `safe_map` adding instrumentation into non-tail-call version and that produces warnings

ggreif: Yeah, maybe a future `bisect_ppx` will be able to cope with it. For now, it is not a big loss. I doubt anybody has looked at the code coverage metrics that we produce in the last 5 years...

ggreif: @Kamirus This is how the report [looks like now](https://dfinity.github.io/motoko/coverage/lib/lib.ml.html#L433).

Issue #5537, opened 2025-10-02
Improve error message for incorrect number of arguments (ctx-dot or implicits only)


Kamirus:
```motoko
let map = Map.empty<Text, Nat>();
map.add() // should be 2 or 3 arguments, because of implicits
```

Missing arguments cause the following error:

```
cannot implicitly instantiate function of type
  <K, V>(map : Map<K, V>, compare : (implicit : (K, K) -> Order), key : K,
         value : V) ->
    ()
to argument of type
  ()
because no instantiation of K, V makes
  ()  <:  (compare : (implicit : (K, K) -> Order), key : K, value : V)
and
  {var root : Node<Text, Nat>; var size_ : Nat}  <:  (map : Map<K, V>)
```

While it should just say that the number of arguments is incorrect.
Same when there are too many arguments.
Instantiation should be done after

PR #5539, opened 2025-10-02
feat: Improve import suggestions

Kamirus:
- fixes the `--ai-errors` flag that would break when there are any `--package` flags specified.
- improves the import suggestions to:
`Hint: Did you mean to import mo:core/Array?`
Instead of
`Hint: Did you mean to import .mops/_github/core#implicits-2@2d4c9f788c1e60554175b82b7cf240c141ae1ee9/src/Array.mo?`


crusso:
Thanks for unpicking this. 
I think there's probably similar package suggestions to fix for dot suggestions (search for eligible funcs)

src/pipeline/resolve_import.ml
```diff
@@ -283,8 +283,13 @@ let list_files : string -> string list =
     List.filter (fun f -> Filename.extension f = ".mo") all_files
 
 let package_imports base packages =
+  let abs p = Lib.FilePath.normalise (
+    if Filename.is_relative p then Filename.concat (Sys.getcwd ()) p else p)
```
crusso: Are you sure this should be relative to cwd? I really don't know, but if you've convinced yourself that's right, I'll go with that...

src/mo_frontend/typing.ml
```diff
@@ -1430,9 +1456,9 @@ let resolve_hole env at hole_sort typ =
              match t with
              | T.Obj (T.Module, fs) -> Some (n, fs)
              | _ -> None) |>
-         Seq.filter_map find_candidate |>
+         Seq.filter_map (fun (module_path, fs) -> find_candidate (module_path, fs)) |>
          List.of_seq in
-     Error (List.map (fun candidate -> candidate.desc) lib_candidates)
```
crusso: There is code similar to this for dot resolutions that may need changing (I basically copied and adapted Christoph's code).

src/pipeline/resolve_import.ml
Kamirus: Not convinced yet... for now this works locally when testing on motoko-core. I'll explore other options too

src/mo_frontend/typing.ml
Kamirus: the next change below does that, both use the `module_name_as_url`

src/pipeline/resolve_import.ml
```diff
@@ -283,15 +283,17 @@ let list_files : string -> string list =
...
 
 let package_imports base packages =
+  let base_norm = Lib.FilePath.normalise base in
   let imports = M.fold (fun pname url acc ->
-    if base = url then
+    let url_norm = Lib.FilePath.normalise url in
+    if base_norm = url_norm || Lib.FilePath.relative_to url_norm base_norm <> None then
       acc
     else
       let files = list_files url in
```
Kamirus, line 293: this is adding files found recursively, so the check above avoid adding relative sub-paths

test/fail/ok/implicit-import-hint-dot.tc.ok
```diff
@@ -0,0 +1,18 @@
...
+  <K, V>(map : Map<K, V>, compare : (implicit : (K, K) -> Order), key : K) ->
+    ?V
+to argument of type
+  (compare : (implicit : (K, K) -> Order), key : K)
+to produce result of type
+  ()
+because no instantiation of K, V makes
+  ?V  <:  ()
+and
+  {var root : Node<Text, Text>; var size_ : Nat}  <:  (map : Map<K, V>)
```
Kamirus, line 11: I want to address this confusing error in https://github.com/dfinity/motoko/issues/5537

src/pipeline/resolve_import.ml
crusso: Add a comment?

src/mo_frontend/typing.ml
```diff
@@ -1310,8 +1310,34 @@ type hole_candidate =
     desc: string;
     typ : T.typ;
     id : T.lab;
+    module_name_opt: string option;
   }
 
+(** Convert a filesystem path to a mo:<package>/<module> URL if it lies under any configured package path. *)
+let mo_url_of_path (path : string) : string option =
```
crusso: I wonder if there is a better home for this?
Maybe mo_typing/suggest.ml?

crusso (approved):

PR #5540, opened 2025-10-02
feat: improved error message for implicit arguments

crusso:
Improves error message for unresolved and ambiguous implicits with more detail and by mentioning available explicit candidates.

Modifies the search to find all candidates that match by type and then partitioning into those candidates
that match by name or not - the second is the set of explicit candidates that could be supplied instead.

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -1396,14 +1396,7 @@ let resolve_hole env at hole_sort typ =
     else None
   in
   let (eligible_ids, explicit_ids) =
-    let vals =
-      match hole_sort with
-      | Named id ->
-         env.vals
-      | Anon _ ->
-         env.vals
-    in
-    T.Env.to_seq vals |>
+    T.Env.to_seq env.vals |>
       Seq.filter_map find_candidate_id |>
       List.of_seq |>
       List.partition (fun (desc : hole_candidate) -> is_matching_lab desc.id)


src/mo_frontend/typing.ml
```diff
@@ -1362,85 +1363,96 @@ let resolve_hole env at hole_sort typ =
...
       | Named id ->
-        (* narrow env to search *)
-        (match T.Env.find_opt id env.vals with
-         | Some info -> T.Env.singleton id info
-         | None -> T.Env.empty)
+         env.vals
       | Anon _ ->
-         env.vals (* search entire env *)
+         env.vals
     in
```
Kamirus: It should be just `let vals = env.vals` now
I cannot make a GH suggestion unfortunately

src/mo_frontend/typing.ml
```diff
@@ -1362,85 +1363,96 @@ let resolve_hole env at hole_sort typ =
...
-         Seq.filter_map find_candidate |>
-         List.of_seq in
-     Error (List.map (fun candidate -> candidate.desc) lib_candidates)
-  | ocs -> begin
-     match disambiguate_resolutions ocs with
-     | Some oc -> Ok oc
+         Seq.map find_candidate_fields |>
+         List.of_seq |>
+         List.flatten |>
+         List.partition (fun (desc : hole_candidate) -> is_matching_lab desc.id)
```
```suggestion
         List.partition (fun (candidate : hole_candidate) -> is_matching_lab candidate.id)
```

Kamirus (approved):
Looks good! I've added two small suggestions

PR #5542, opened 2025-10-03
Claudio/implicit renamings

crusso:


src/mo_frontend/typing.ml
```diff
@@ -1417,6 +1420,21 @@ let resolve_hole env at hole_sort typ =
...
         explicit_ids @ explicit_fields)
   in
+  let renaming_hints env =
+    List.iter (fun candidate ->
+        match hole_sort with
+        | Anon _ -> ()
+        | Named id ->
+        if (candidate.region.left.file = at.left.file) then
+          info env candidate.region
+            "Consider renaming `%s` to `%s` so it can serve as an implicit argument `%s` in this call\n%s%s"
```
Kamirus: What if there are multiple compare functions already in scope, or what if this warning is shown in two places and renaming both would create a conflict.
Maybe we should suggest putting these in modules to avoid duplicate names.

crusso: Does caffeine generate modules?
After all, this is only  suggestion, but perhaps it will just confuse AI?

crusso: Defining a local module of the same name as the type will wind up hiding all bindings from the original module, I fear.
eg  compareNat to module Nat { public func compare .. } will hide all imported Nat bindings.

crusso: Ok, done it, trying to avoid a bad suggestion that would hide bindings.

src/mo_frontend/typing.ml
```diff
@@ -1417,6 +1420,42 @@ let resolve_hole env at hole_sort typ =
...
+               (match Source.read_region at with Some s -> ": " ^ s | None -> ".")
+        | DotE({ it = VarE {it = mid;_ }; _ }, _, _) ->
+               info env candidate.region
+               "Consider renaming `%s` to `%s` in `%s`. Then it can serve as an implicit argument `%s` in this call\n%s%s"
+               candidate.id
+               id
+               mid
+               id
+               (Source.string_of_region at)
+               (match Source.read_region at with Some s -> ": " ^ s | None -> ".")
```
Kamirus: I'd extract the common part to another function. It should reduce the complexity in this function

test/fail/ok/implicit.tc.ok
```diff
@@ -1,17 +1,27 @@
-implicit.mo:58.3-58.15: type error [M0230], Cannot determine implicit argument 'c' of type
+implicit.mo:58.1-58.15: type error [M0230], Cannot determine implicit argument 'c' of type
   (Bool, Bool) -> Order
-Hint: If you're trying to use an implicit argument you need to have a matching declaration named c in scope.
-implicit.mo:60.3-60.9: type error [M0230], Cannot determine implicit argument 'd' of type
+Hint: If you're trying to omit an implicit argument named `c` you need to have a matching declaration named `c` in scope.
+implicit.mo:60.1-60.9: type error [M0230], Cannot determine implicit argument 'd' of type
```
Kamirus: I'd change backticks to `'` for more consistency

crusso: Actually we use a random mixture of both, both mostly backtick. I'll make this consistent here and then we can fix all of them some other day

crusso:
@Kamirus (PTAL)

PR #5548, opened 2025-10-06
feat: Validate number of arguments for ctx-dot and implicits

### LINKED ISSUE #5544: Problem with detection of dot notation (bug)
Starting from this working code:
```
module M {
  public type Self = Nat;
  public func f(s : Self) {};
};
let s : M.Self = 1;
s.f()
```
If we change the type in this way then it stops working:
```
module M {
  public type Self = (Nat, Nat);
  public func f(s : Self) {};
};
let s : M.Self = (0, 0);
s.f()
```
If I then add a second argument then it starts working again:
```
module M {
  public type Self = (Nat, Nat);
  public func f(s : Self, _ : Nat) {};
};
let s : M.Self = (0, 0);
s.f(0)
```

Kamirus:
Fixes #5544
Closes #5550
Closes #5553

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -145,7 +145,7 @@ let display_many display p xs =
 
 let display_expected_arg_types fmt types =
   if types = [] then
-    Format.fprintf fmt "Expected no arguments\n"
+    Format.fprintf fmt "Expected no arguments"
   else
     Format.fprintf fmt "Expected arguments of type:%a" (display_many display_typ_expand) types
 
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -2617,6 +2617,7 @@ and infer_call env exp1 inst (parenthesized, ref_exp2) at t_expect_opt =
       ts, t_arg', t_ret'
     | _::_, None -> (* implicit, infer *)
       (* Before bimatching, validate the number of arguments *)
+      (* TODO: try to move it to the decompose so that it is closer to the bimatching *)
       validate_call_arg env ctx_dot exp2 t_args;
       infer_call_instantiation env t1 tbs t_arg t_ret exp2 at t_expect_opt extra_subtype_problems
   in
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -2533,6 +2533,26 @@ and infer_callee env exp =
      infer_exp_promote env exp, None
 
 and insert_holes at ts es =
+  let as_implicit = function
+    | T.Named ("implicit", T.Named (arg_name, t)) ->
+      Some arg_name
+    | T.Named ("implicit", t) ->
+      Some "_"
+    | T.Named (_inf_arg_name, (T.Named ("implicit", T.Named (arg_name, t)))) ->
+      (* override inferred arg_name *)
+      Some arg_name
+    | T.Named (inf_arg_name, (T.Named ("implicit", t))) ->
+      (* non-overriden, use inferred arg_name *)
+      Some inf_arg_name
+    | _ -> None
+  in
+
+  let implicits_n = List.to_seq ts |> Seq.filter_map as_implicit |> Seq.length in
+  if List.length ts <> List.length es + implicits_n then
+    let () = Printf.printf "+arity mismatch: %d <> %d + %d\n" (List.length ts) (List.length es) implicits_n in
+    es
+  else
+
   let mk_hole pos hole_id =
     let hole_sort = if hole_id = "" then Anon pos else Named hole_id in
     {it = HoleE (hole_sort, ref {it = PrimE "hole"; at; note=empty_typ_note });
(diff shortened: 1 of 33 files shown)


src/mo_frontend/typing.ml
```diff
@@ -2533,37 +2533,74 @@ and infer_callee env exp =
...
+
+    With ctx_dot: (exp1 = o.f)
+    1. o.f()  ~>  f(o)
+    2. o.f(e)  ~>  f(o, e)
+    3. o.f(e1 .. en)  ~>  f(o, e1 .. en)
+
+    With implicits: (exp1 = f)
+    1. f()  ~>  f(i1 .. ik)
+    2. f(e)  ~>  f(.. i .. e .. i ..)  (implicits inserted where implicit parameters)
+    3. f(e1 .. en)  ~>  f(.. i .. e1 .. i .. .. i .. en .. i ..)  (implicits inserted where implicit parameters)
```
crusso: maybe we should spell out the combination too.

test/fail/ok/implicit-import-hint-dot.tc.ok
```diff
@@ -1,14 +1,4 @@
...
-to produce result of type
-  ()
-because no instantiation of K, V makes
-  ?V  <:  ()
-and
-  {var root : Node<Text, Text>; var size_ : Nat}  <:  (map : Map<K, V>)
+implicit-import-hint-dot.mo:13.10-13.12: type error [M0233], wrong number of arguments: expected 2 but got 1
 implicit-import-hint-dot.mo:17.12-17.26: type error [M0230], Cannot determine implicit argument `compare` of type
   (Text, Text) -> Order
 Hint: Did you mean to import `core-stub/src/Text.mo.compare`?
```
crusso: Is that hint not supposed to be `mo:core/Text`

crusso:
Looks promising, thanks for refactoring and fixing my crap.

test/fail/ok/implicit-import-hint-dot.tc.ok
Kamirus: Ooops, I see https://github.com/dfinity/motoko/pull/5542 broke it when merging master

Kamirus: fix here https://github.com/dfinity/motoko/pull/5554

src/mo_frontend/typing.ml
```diff
@@ -2574,28 +2610,48 @@ and infer_call env exp1 inst (parenthesized, ref_exp2) at t_expect_opt =
...
-      match T.normalize t_arg with
-      | T.Tup([t'; t2]) -> t2, [(t, t')]
-      | T.Tup(t'::ts) -> T.Tup(ts), [(t, t')]
-      | t' -> T.unit, [(t, t')]
+      match t_args with
+      | [t'; t2] -> [t2], [(t, t')]
+      | t'::ts -> ts, [(t, t')]
+      | [] ->
+        (* Should be unreachable *)
+        [T.unit], [(t, T.unit)]
```
christoph-dfinity:
```suggestion
      match t_args with
      | t'::ts -> ts, [(t, t')]
      | [] -> assert false
```

The first case is already covered by the second one, and I agree that the last case is unreachable. Contextual dot search can't succeed if the function accepts no arguments.

src/mo_frontend/typing.ml
```diff
@@ -427,6 +442,17 @@ let check_closed env id k at =
...
+let can_args_match args t_args =
+  match args with
+  | [_] ->
+    (* A single argument could match any tuple *)
+    true
+  | _ ->
+    (* Case 1: No arguments, must match unit (empty tuple) *)
+    (* Case 2: Multiple arguments, must match n-tuple where n is the number of arguments *)
+    T.can_ntuple_match (List.length args) (T.seq t_args)
+
```
christoph-dfinity: I don't understand this function. Looks like we're condensing `args` into its length either via pattern match on shape or by calling List.length, so we could just accept the length up-front?

Also don't get the `A single argument could match any tuple`. How would `LitE 10` match any tuple? What cases do we miss out on if we remove this?

src/mo_frontend/typing.ml
```diff
@@ -2574,28 +2610,48 @@ and infer_call env exp1 inst (parenthesized, ref_exp2) at t_expect_opt =
...
+    match ctx_holes with
+    | None -> exp2, syntax_args
+    | Some args ->
+      let e' = match args with
+        | [e] -> e.it
+        | es -> TupE es
+      in
+      let exp2 = { exp2 with it = e'} in
+      exp2, args
+  in
```
christoph-dfinity: I think this code could be clearer if we moved the re-wrapping into `TupE` into `insert_holes`

src/mo_frontend/typing.ml
```diff
@@ -2574,28 +2610,48 @@ and infer_call env exp1 inst (parenthesized, ref_exp2) at t_expect_opt =
...
+
+  (* Match the arguments against the parameter types *)
+  let t_arg =
+    if (ctx_dot = None && ctx_holes = None && can_args_match args t_args) || List.length t_args = List.length args
```
christoph-dfinity: I think in general `Option.is_none` is to be preferred over polymorphic equality (code size and perf wise)

src/mo_types/type.ml
```diff
@@ -684,8 +684,8 @@ let as_pair_sub t = match promote t with
   | _ -> invalid "as_pair_sub"
 let as_func_sub default_s default_arity t = match promote t with
   | Func (s, c, tbs, ts1, ts2) ->
-    s, tbs, seq ts1, codom c (fun () -> Var((List.hd tbs).var, 0)) ts2
-  | Non -> default_s, Lib.List.make default_arity {var = "X"; sort = Type; bound = Any}, Any, Non
+    s, tbs, ts1, codom c (fun () -> Var((List.hd tbs).var, 0)) ts2
+  | Non -> default_s, Lib.List.make default_arity {var = "X"; sort = Type; bound = Any}, [Any], Non
```
christoph-dfinity, line 688: Well spotted 👍 This simplifies a lot of the downstream code!

src/mo_types/type.ml
```diff
@@ -1528,6 +1528,13 @@ let rec isolated_typ co = function
 
 and isolated t = isolated_typ (ref S.empty) t
 
+let can_ntuple_match n typ =
+  match promote typ with
+  | Tup ts -> List.length ts = n
+  | Any
+  | Var _ -> true
+  | _ -> false
```
christoph-dfinity: This feels pretty ad-hoc. If this is only used for error reporting I'd maybe move it closer to the reporting.

src/mo_frontend/typing.ml
Kamirus: Yeah this is tricky. I'd appreciate a 2nd pair of eyes on that part and we can definitely separate it out into another PR if we are not sure about it.
So this function is used to enable the arguments check for functions that dont use dot notation nor implicits.
This functions determines when it is safe to raise an error when `List.length args <> List.length t_args`.
The idea is to cheaply look at the number of arguments and the type that they must match.
A single argument could be used as a tuple, this check is not sufficient, it does not determine that `args` have type `t_args`. It is just a heuristic that detects obvious contradictions, like `args = [a1, a2]` cannot match `t_args = [t1, t2, t3]`, because of tuple arity mismatch.
when there are no args, the type must be unit, but it could also be a variable that could bimatch with unit etc.

Kamirus: Names for both `can_args_match` and `can_ntuple_match` could use some improvements.
`can_args_match args t_args` **does not imply** `args : T.seq t_args`.

Instead
`not (can_args_match args t_args)` implies `not (args : T.seq t_args)`
So if this check fails, we know that the later `check_exp` (or infer_exp + bimatch sub) would fail anyway, so it is safe to report an early argument mismatch error

christoph-dfinity: I see now. I'd prefer doing this in a follow-up PR. I'm sure there are some good heuristics we can come up with here.

src/mo_frontend/typing.ml
Kamirus: Really? How do you know ocamlc is not optimizing both to the same? `X = None` is more concise. I'd imagine `Option.is_none` to be worse if not inlined, but I guess they inline it

src/mo_frontend/typing.ml
Kamirus: but i'm using the `args` later, no I don't want to wrap them in the `insert_holes` to unwrap them again here 😄

Kamirus: yeah but maybe it is better to have just one of them, not both `exp2` and `args` ...

src/mo_frontend/typing.ml
Kamirus: Ok! I've reverted these changes

PR #5551, opened 2025-10-06
experiment: add a new method to core using stable functions

crusso:
This is a simple example demonstrating that just adding method to core (Map.copy) is a breaking change
and requires not just a migration, but a complex migration that holds on to both old and new core library.

### DIFF
--- a/test/run-drun/upgrade-core/core1/Map.mo
+++ b/test/run-drun/upgrade-core/core1/Map.mo
@@ -1,0 +1,176 @@
+import ImperativeMap "imperative/Map";
+import ImperativeIter "imperative/Iter";
+import Iter "Iter";
+import PureMap "pure/Map";
+import Order "Order";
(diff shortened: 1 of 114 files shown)


test/run-drun/upgrade-core/core1/Map.mo
```diff
@@ -0,0 +1,176 @@
...
+
+  public persistent class Map<K, V>(comparison : PersistentCompare<K>) {
+    let inner = ImperativeMap.empty<K, V>();
+
+    public func toPure() : PureMap.Map<K, V> {
+      PureMap.fromIter(ImperativeMap.entries(inner), comparison : TransientCompare<K>)
+    };
+
+    // some new method
+    public func copy() : Map<K,V> {
```
crusso, line 20: This is the only difference between core1 and core0 (a new `copy` method)

PR #5560, opened 2025-10-08
experiment: use ctx dot instantiation

Kamirus:


src/mo_frontend/typing.ml
```diff
@@ -2664,11 +2667,18 @@ and infer_call env exp1 inst (parenthesized, ref_exp2) at t_expect_opt =
       if not env.pre then check_exp_strong env t_arg' exp2
       else if typs <> [] && Flags.is_warning_enabled "M0223" &&
         is_redundant_instantiation ts env (fun env' ->
-          infer_call_instantiation env' t1 ctx_dot tbs t_arg t_ret exp2 at t_expect_opt extra_subtype_problems) then
+          infer_call_instantiation env' t1 tbs t_arg t_ret exp2 at t_expect_opt extra_subtype_problems) then
             warn env inst.at "M0223" "redundant type instantiation";
       ts, t_arg', t_ret'
     | _::_, None -> (* implicit, infer *)
-      infer_call_instantiation env t1 ctx_dot tbs t_arg t_ret exp2 at t_expect_opt extra_subtype_problems
+      match dot_full_inst with
```
crusso: Is this actually complete? The full instantiation might be too specific to match the other args?
I.e. if you had inferred with the arguments in hand, you might have chosen a different, weaker, instantiation to accommodate them as well as the receiver.

christoph-dfinity: I think it's "safe" in that it'll never infer anything non-principal, right? It might reject more programs than we accepted before. Maybe we could weaken this to a "second pass" in the error case in the future?

Kamirus: Yes, fixing parts of the instantiation early based on the receiver can be more restrictive.
If we want the same inference results for both dot and without dot syntax then we should not use the partial instantiation.
I'd like to explore an alternative of reporting better bimatch errors

test/run/ok/contextual-dot.tc.ok
```diff
@@ -0,0 +1,8 @@
+contextual-dot.mo:50.33-50.40: type error [M0050], literal of type
+  Text
+does not have expected type
+  B
```
Kamirus: and here is a bug, the `B` type parameter of a map function should not be fixed

PR #5561, opened 2025-10-09
experiment: implicit import all libs for dot&implicit resolution

Kamirus:
Adds undocumented implicit libs flag that allows resolution for ctx dot and implicits even with missing imports.
This flag adds all possible env.libs similarly to the --ai-errors flag.


src/mo_frontend/typing.ml
```diff
@@ -28,6 +28,16 @@ type visibility_env = visibility_src T.Env.t * visibility_src T.Env.t
 
 let available env = T.Env.map (fun (ty, at, kind) -> (ty, at, kind, Available)) env
 
+(* When AI errors are enabled, expose modules from libs as value bindings so they
+   can participate in contextual resolution and environment display. *)
+let lib_modules_as_vals (libs : Scope.lib_env) : val_env =
+  T.Env.fold (fun id ty acc ->
+    match T.normalize ty with
+    | T.Obj (T.Module, _) ->
+      T.Env.add ("file$" ^ id) (ty, Source.no_region, Scope.Declaration, Available) acc
```
Kamirus, line 38: they need to be Available, otherwise it is an error

src/mo_frontend/typing.ml
```diff
@@ -28,6 +28,16 @@ type visibility_env = visibility_src T.Env.t * visibility_src T.Env.t
 
 let available env = T.Env.map (fun (ty, at, kind) -> (ty, at, kind, Available)) env
 
+(* When AI errors are enabled, expose modules from libs as value bindings so they
+   can participate in contextual resolution and environment display. *)
+let lib_modules_as_vals (libs : Scope.lib_env) : val_env =
```
Kamirus, line 34: I found that we can either add them in the typing.ml or in the pipeline.ml...

test/run/core-stub/src/Text.mo
```diff
@@ -0,0 +1,14 @@
+/// Stub for Text.
```
Kamirus: Can we share these files between the `test/fail` and `test/run`?
Putting them in `test` does not work as nix is not including all the files in test...

test/run/implicit-import-dot.mo
```diff
@@ -0,0 +1,18 @@
...
+
+let map = Map.empty<Text, Nat>();
+map.add("abc", 3)
+
+// let ar = [3, 1, 2];
+// let ar2 = ar.sort()
+
+//SKIP run-low
+//SKIP run
+//SKIP run-ir
```
Kamirus: I could not support the interpret mode

src/pipeline/pipeline.ml
```diff
@@ -450,6 +450,19 @@ let chase_imports_cached parsefn senv0 imports scopes_map
         let* sscope = check_lib !senv cur_pkg_opt lib in
         libs := lib :: !libs; (* NB: Conceptually an append *)
         senv := Scope.adjoin !senv sscope;
+        (* Inject implicit alias as a value binding in the scope for AI diagnostics. *)
+        (* if !Flags.ai_errors then (
+          (* Inject alias values for all lib entries now in scope *)
+          Type.Env.iter (fun full_path typ ->
```
Kamirus: we could also inject these `vals` here. probably better in typing...

src/mo_frontend/typing.ml
Kamirus, line 38: "file$" just makes sure it matches the name added in the desugar.ml
if we go with this hack, then we would need to extract the `id_of_full_path` into a common place and call here

crusso: The Available thing is hack to prevent you from writing recursive actor classes, whose constructors are not availabe in the body of the class (but available to importers). IIRC

Fine to mark them as Available.

crusso: You might have to extend the typing of VarE (?) to look into env.libs when checking the "witnesses" for implicits and dots, or maybe that's taken care of by extending the inital scope already?

crusso: Another way to tackle this might be to use ImportE (of a lib as the witness module (not VarE) and extend desugar to allow ImportE  in other places beyond top-level (which I believe it once did anyway).

A minor danger, not to worry about now, is that we might be able to introduce cyclic dependencies between libraries this way, because imports are no longer explicit so we won't consider the implicit ones in the dependency analysis. Might not matter at all, but just occurred to me.

src/mo_frontend/typing.ml
```diff
@@ -1425,19 +1447,30 @@ let resolve_hole env at hole_sort typ =
...
       Seq.filter_map is_module |>
       Seq.map find_candidate_fields |>
       List.of_seq |>
       List.flatten |>
       List.partition (fun (desc : hole_candidate) -> is_matching_lab desc.id)
   in
+  let lib_fields () =
+    T.Env.to_seq env.libs |>
+      Seq.filter_map (fun (n, t) ->
+        match t with
```
```suggestion
        match T.normalize t with
```

src/mo_frontend/typing.ml
```diff
@@ -487,7 +507,7 @@ and check_obj_path' env path : T.typ =
   match path.it with
   | IdH id ->
     use_identifier env id.it;
-    (match T.Env.find_opt id.it env.vals with
+    (match find_id env id with
```
crusso: Hmm, I'm not sure I'd allow this here  and just look at env.vals as before

src/mo_frontend/typing.ml
```diff
@@ -1281,7 +1301,7 @@ let rec combine_pat_fields_srcs env t tfs (pfs : pat_field list) : unit =
       combine_pat_fields_srcs env t tfs' pfs';
 
 and combine_id_srcs env t id : unit =
-  match T.Env.find_opt id.it env.vals with
+  match find_id env id with
```
crusso: Same here (revert)

src/mo_frontend/typing.ml
```diff
@@ -1469,20 +1502,14 @@ let resolve_hole env at hole_sort typ =
...
-         List.flatten |>
-         List.partition (fun (candidate : hole_candidate) -> is_matching_lab candidate.id)
-     in
-     Error (List.map suggestion_of_candidate lib_terms,
-            List.map (fun candidate -> candidate.desc) explicit_terms,
-            renaming_hints)
+    (match disambiguate_resolutions (fst (eligible_fields env.implicit_lib_vals)) with
+    | Some term -> Ok term
+    | None ->
+      let (lib_terms, _) = lib_fields () in
```
crusso, line 1508: This is essentially recomputing the same thing as eligible_fields env.implicit_lib_vals, right?
Maybe just bind  eligible_fields env.implicit_lib_vals to `terms1` and match on terms1?

src/mo_frontend/typing.ml
```diff
@@ -1539,29 +1563,33 @@ let contextual_dot env name receiver_ty =
...
+    | _ ->
+      let lib_candidates =
+        T.Env.to_seq env.libs |>
+          Seq.filter_map (fun (n, t) ->
+              match t with
+              | T.Obj (T.Module, fs) -> Some (n, (t, fs))
+              | _ -> None) |>
+          Seq.filter has_matching_self_type |>
+          Seq.filter_map find_candidate |>
+          List.of_seq in
```
crusso, line 1586: Again, I think lib_candidates is just recomputing eligible_funcs env.implicit_lib_val
so bind the former and just re-use?

test/core-stub/src/Nat.mo
```diff
@@ -0,0 +1,11 @@
...
+
+import Types "Types";
+
+module {
+  public type Self = Nat;
+
+  public func compare(x : Nat, y : Nat) : Types.Order {
+    if (x < y) { #less } else if (x == y) { #equal } else { #greater }
+  };
+}
```
crusso:
```suggestion
}

```

test/core-stub/src/Int.mo
```diff
@@ -0,0 +1,11 @@
...
+
+import Types "Types";
+
+module {
+  public type Self = Int;
+
+  public func compare(x : Int, y : Int) : Types.Order {
+    if (x < y) { #less } else if (x == y) { #equal } else { #greater }
+  };
+}
```
crusso:
```suggestion
}

```

test/run/implicit-import-dot.mo
```diff
@@ -0,0 +1,49 @@
...
+mainText();
+mainNat();
+mainInt();
+mainCompareText();
+// mainCompareNat();
+mainCompareInt();
+
+//SKIP run-low
+//SKIP run
+//SKIP run-ir
```
crusso:
```suggestion
//SKIP run-ir

```

src/mo_frontend/typing.ml
Kamirus: they differ by the name. the implicit_lib_vals have the special "file$" name

PR #5569, opened 2025-10-13
fix: resolve issue #5507

### LINKED ISSUE #5507: bug: blob imports aren't working correctly
test.md
```
import pub = "blob:file:./test.md";
```

with a file next to the .mo file named test.md throws the following error.
```
❯ dfx deploy
Deploying all canisters.
All canisters have already been created.
Error: Failed while trying to deploy canisters.
Caused by: Failed to build all canisters.
Caused by: Failed while trying to build all canisters.
Caused by: Failed while trying to build all canisters in the canister pool.
Caused by: Failed to build dependencies graph for canister pool.
Caused by: Failed to get dependencies for canister 'motoko-canister-creation-backend'.
Caused by: Failed to find imports for canister at '/Users/moritz/projects/motoko-canister-creation/src/motoko-canister-creation-backend/main.mo'.
Caused by: Failed recursive dependency detection at /Users/moritz/projects/motoko-canister-creation/src/motoko-canister-creation-backend/main.mo.
Caused by: Failed to create MotokoImport.
Caused by: The dependency analyzer failed: Unknown import blob:file:./test.md
```

Also: 
 
> "blob:file:<filepath>" then <pat> is bound to a blob containing the contents of the file <filepath>. <filepath> is interpreted relative to the absolute location of the enclosing file. For example, import image "blob:file:/assets/image.jpg" defines image as the blob with bytes from local file ./assets/image.jpg. Unlike library imports, <filepath> should include the file’s extension, if any.

is incorrect - it actually seems to depend on the directory  moc is executed from, not the location relative to the imported file.

Also, VScode does not seem to resolve the file correctly:

Image

ggreif:
Fixes #5507.

The VSCode part will need another round, though. (I suspect)

### DIFF
--- a/src/lang_utils/error_codes.ml
+++ b/src/lang_utils/error_codes.ml
@@ -205,6 +205,7 @@ let error_codes : (string * string option) list =
     "M0231", None; (* Ambiguous implicit argument *)
     "M0232", None; (* Cannot infer type of implicit argument *)
     "M0233", None; (* Wrong number of arguments *)
+    "M0234", None; (* Field exists but is not a function *)
   ]
 
 (** Message codes that can be both used as warnings and errors *)
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -171,11 +171,21 @@ let display_given_arg_types fmt types =
   else
     Format.fprintf fmt "But got %d argument%s of type:%a" (List.length types) (plural_typs types) (display_many display_typ_expand) types
 
-let display_obj fmt (s, fs) =
-  if !Flags.ai_errors || (List.length fs) < 16 then
-    Format.fprintf fmt "type:%a" display_typ (T.Obj(s, fs))
-  else
-    Format.fprintf fmt "%s." (String.trim(T.string_of_obj_sort s))
+let display_obj fmt typ =
+  match T.normalize typ with
+  | T.Obj(s, fs) ->
+     if !Flags.ai_errors || (List.length fs) < 16 then
+       Format.fprintf fmt "type:%a" display_typ_expand typ
+     else
+       Format.fprintf fmt "%s."
+         T.(match s with
+          | Actor -> "actor"
+          | Module -> "module"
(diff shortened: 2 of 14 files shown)


test/run/issue-5507/inner.mo
```diff
@@ -0,0 +1,3 @@
+import _ = "blob:file:issue-5507/inner.mo";
```
ggreif:
```suggestion
import _ = "blob:file:./inner.mo";
```
should work this way

src/pipeline/resolve_import.ml
```diff
@@ -80,7 +80,7 @@ let err_file_does_not_exist' at full_path =
     at
     "M0009"
     "import"
-    (Printf.sprintf "file \"%s\" does not exist" full_path)
+    (Printf.sprintf "file \"%s\" does not exist (cwd: %s)" full_path (Sys.getcwd()))
```
ggreif: undo this

src/pipeline/resolve_import.ml
```diff
@@ -159,12 +159,13 @@ let add_value_import msgs imported ri_ref at path =
   let add_no_extension _file_exists f = f in
   match resolve_lib_import at path add_no_extension with
   | Ok full_path -> begin
+      (*Printf.eprintf "path: %s full_path: %s, cwd: %s\n" path full_path (Sys.getcwd());*)
```
ggreif:
```suggestion
```

src/pipeline/resolve_import.ml
```diff
@@ -205,7 +205,7 @@ let resolve_import_string msgs base actor_idl_path aliases packages imported (f,
     | None -> err_alias_not_defined msgs at alias
     end
   | Ok (Url.FileValue path) ->
-    add_value_import msgs imported ri_ref at path
+    add_value_import msgs imported ri_ref at (in_base base path)
```
ggreif, line 208: This is the fix. We do the same for `mo:` imports

PR #5573, opened 2025-10-14
[VSC-31] Remove intermediate conversion to S-exps

alexd1971:
Problem: The current process for creating a JS representation of the AST requires an intermediate conversion to S-expression. This hurts performance and makes it impossible to annotate all nodes with types.

Solution: Implement a library for converting the original AST directly to JS, eliminating the intermediate step.


rvanasa:
LGTM for now, postponing the `typeRep` cache for a future PR.

PR #5574, opened 2025-10-14
feat(breaking): removes the requirement for module Self type

christoph-dfinity:
Changes it so we require the first argument to a function be named `self` in order to be used with contextual-dot, while removing the `Self` type requirement.

Pros:
- Allows selectively opting into dot-notation for functions in a module
- Doesn't require adding an often duplicated type `Self` to modules
- Makes it a little easier to lint/enforce contextual-dot style
- Syntactically closer to Python/Rust, which might help with LLMs

Cons:
- Will cause a noisier change to `core`
- Somehow I feel less good about special value names than special type names

### DIFF
--- a/test/fail/contextual-dot-unused-self.mo
+++ b/test/fail/contextual-dot-unused-self.mo
@@ -1,0 +1,19 @@
+module F {
+ public func f(_ : (self : Nat)) {}; // no warning
+ public func g(self: Nat) {}; // warn unusued self
+ public func h(_self: Nat) {}; // suppress unused self, but also not dot enabled
+};
+
+func _testf() {
+  (1).f();
+};
+
+func _testg() {
+  (1).g();
+};
+
+
+func _testh() {
+  F.h(1); // works
+  // (1).h(); causes an error
+};
--- a/test/fail/ok/contextual-dot-unused-self.tc.ok
+++ b/test/fail/ok/contextual-dot-unused-self.tc.ok
@@ -1,0 +1,1 @@
+contextual-dot-unused-self.mo:3.16-3.20: warning [M0194], unused identifier self (delete or rename to wildcard `_` or `_self`)


src/mo_frontend/typing.ml
```diff
@@ -1558,15 +1558,8 @@ let contextual_dot env name receiver_ty =
...
     with _ ->
       None in
-  let has_matching_self tf = match tf with
-    | T.{ lab = "Self"; typ = T.Typ con; _ } ->
-       (match Cons.kind con with
-       | T.Def(tbs, t') -> permissive_sub receiver_ty (tbs, t') <> None
-       | _ -> false)
-    | _ -> false in
-  let has_matching_self_type (_, (_, fs)) = List.exists has_matching_self fs in
   let is_matching_func = function (* TODO: normalize first *)
```
```suggestion
  let is_matching_func typ = match T.normalize typ with
```
(Is what I meant here)

src/mo_frontend/typing.ml
```diff
@@ -1577,7 +1570,6 @@ let contextual_dot env name receiver_ty =
   let eligible_funcs vals =
     T.Env.to_seq vals |>
       Seq.filter_map is_module |>
```
crusso: So I guess this search is a little more expensive now since we have to test every value and can't just stop if looking the the module has no `Self` self-type. But then, we added it to almost every module...

But we can also place extensions on unrelated types in the same module...

christoph-dfinity: O-notation wise it's the same, as it's still a linear traversal over the module members (but yeah right now it would be more expensive, as the sorting on module members puts the types first). When/If we get around to making modules maps rather than lists the lookup would be `O(1)` in both versions.

src/mo_frontend/typing.ml
crusso: Actually, the fields are sorted and we know the label we want so we could do a faster search for the label first (especially if we used maps or arrays for the fields, not ordered lists as we do now.

christoph-dfinity: Cant really binary search a linked list, right?

crusso: (so room for future optimization, not now)

crusso: No, I know, hence future.

crusso:
| Allows selectively opting into dot-notation for functions in a module

A IRL example of this might be preferring Nat.compare(n1, n2) over n1.compare(n2). I like it.

crusso:
Another reason to like this is that it lets us put Iter.toSet into the Set module, as a method on self : Iter<T> instead of Set.fromIter.

PR #5575, opened 2025-10-14
feat: Partial explicit type instantiation

Kamirus:
Add partial explicit type instantiation using wildcard for types that must be inferred. 
Example:
```motoko
VarArray.map<_, Nat>(testArray, func x = x * 2)
```
1. Specifies the 2nd type argument in order to avoid the invariance limitation (Otherwise could be ambiguous: `Nat` vs `Int`)
2. Omit the 1st type argument, let the compiler infer it

Partial type instantiations are an improvement over explicit return type annotations:
- Works well with the dot-notation (return type annotations needed extra parenthesis or intermediate let-binding)
- Shorter, more concise syntax


test_partial_instantiation.mo
```diff
@@ -0,0 +1,36 @@
...
+let result1 = VarArray.map<Nat, Nat>(testArray, func x = x * 2);
+
+// Test 2: Partial instantiation with wildcard for first type argument
+let result2 = VarArray.map<_, Nat>(testArray, func x = x * 2);
+
+// Test 3: Partial instantiation with wildcard for second type argument  
+let result3 = VarArray.map<Nat, _>(testArray, func x = x * 2);
+
+// Test 4: Both wildcards (should infer all)
+let result4 = VarArray.map<_, _>(testArray, func x = x * 2);
```
christoph-dfinity: Should we warn about these? All wildcards is the same as not providing an instantiation, right?

Kamirus:
No time for pushing this now, we can revisit once this becomes a priority

PR #5576, opened 2025-10-14
feat: disambiguates overlapping context-dot resolutions similar to implicits

christoph-dfinity:


test/fail/M0224.mo
```diff
@@ -5,11 +5,18 @@ module A {
...
+
+module C {
+  public type Self = { x : Int };
+  public func getX1(self : Self) : Int { self.x + 1 };
 };
 
 let a : A.Self = { x = 10 };
 let b : B.Self = { x = +10 };
 
+ignore a.getX(); // Fine (with disambiguation)
```
crusso, line 20: So this resolves to A.getX(a)? right

christoph-dfinity: uhhhhh, my head hurts, but I think contextual-dot needs to disambiguate the "other direction" from implicits, because it's disambiguating functions and the type it's using to compare them is in contravariant position.

`Nat <: Int`

If you say `(10).toText()` you want `Nat.toText()`, while `(-10).toText()` can only pick `Int.toText()`

Given `zero : <T>(zero : implicit T) -> T`

`zero<Int>()` should pick `Int.zero`, while `zero<Nat>()` can only pick `Nat.zero`

src/mo_frontend/typing.ml
```diff
@@ -1352,19 +1352,21 @@ let suggestion_of_candidate candidate =
    these types is the "closest" to the required type. If we can
    uniquely identify a single candidate that is the supertype of all
    other candidates we pick it. *)
-let disambiguate_resolutions (candidates : hole_candidate list) =
-  let add_candidate (frontiers : hole_candidate list) (c : hole_candidate) =
-    let rec go (fs : hole_candidate list) = match fs with
+let disambiguate_resolutions (candidates : 'candidate list) (get_ty : 'candidate -> T.typ) =
```
crusso: I would just pass in `rel` instead of sub and pass T.sub for implicits, `T.super` or `inv T.sub` for dots.

test/fail/M0224.mo
Kamirus: Let's test what is the expected type from this 

```suggestion
let _ : Nat = a.getX(); // Fine (with disambiguation)
```

PR #5579, opened 2025-10-14
experiment: improved type argument inference errors

crusso:
builds on #5464 with terser messages and report of first failing subtype problem (typically an argument or checked return type).


Changelog.md
```diff
@@ -61,6 +61,7 @@
     Deprecates any public types and values with special doc comment
     `/// @deprecated M0235`.
     (allowed by default, warn with `-W 0235`).
+>>>>>>> master
```
```suggestion
```

test/fail/ok/invariant-inference.tc.ok
```diff
@@ -2,7 +2,9 @@ invariant-inference.mo:5.11-5.32: type error [M0098], cannot apply function of t
   (len : Nat, x : T) -> [var T]
 to argument of type
   (Nat, Nat)
+because type parameter `T` has no "best" solution.
 Hint: Add explicit type instantiation, e.g. <Nat>
+because Hint: Add explicit type instantiation, e.g. <Nat>
```
Kamirus: Duplicate here

test/fail/ok/inference.tc.ok
```diff
@@ -12,90 +12,116 @@ to argument of type
...
+because type parameter `T` has a bound 
+  U
+involving another type parameter. Please provide an explicit instantiation.
 inference.mo:111.8-111.17: type error [M0098], cannot apply function of type
   (x : T) -> T
 to argument of type
   Bool
 to produce result of expected type
   Any
+because type parameter `T` has no solution. Maybe try an explicit instantiation.
```
alexandru-uta: I'm not sure that adding this technical explanation `(because type parameter `T` has no solution)` helps a dev or an AI.

test/fail/ok/inference.tc.ok
```diff
@@ -12,90 +12,116 @@ to argument of type
...
+because there is no way to satisfy 
+  <V>(x : V) -> V  <:  <U>T -> U  (for argument `f`) 
 inference.mo:118.1-118.31: type error [M0098], cannot apply function of type
   (f : <U>U -> T) -> ()
 to argument of type
   <V>(x : V) -> V
 to produce result of expected type
   ()
+because there is no way to satisfy 
+  <V>(x : V) -> V  <:  <U>U -> T  (for argument `f`)
```
alexandru-uta: likewise, do people understand what this relation means?

test/fail/ok/lambdas-invariant.tc.ok
```diff
@@ -2,59 +2,83 @@ lambdas-invariant.mo:34.11-34.48: type error [M0098], cannot apply function of t
   ([var Nat], Nat -> C<U>) -> [var U]
 to argument of type
   ([var Nat], Nat -> C<U>)
+because type parameter `U` has no "best" solution.
 Hint: Add explicit type instantiation, e.g. <Int>
+because Hint: Add explicit type instantiation, e.g. <Int>
```
alexandru-uta: I'd remove "because" before "Hint"

test/fail/ok/wrong-call-args.tc.ok
```diff
@@ -58,33 +67,45 @@ to argument of type
...
   ()
+because there is no way to satisfy 
+  Nat  <:  (A, B)
 wrong-call-args.mo:112.5-112.17: type error [M0098], cannot apply function of type
   ((A, B)) -> ()
 to argument of type
   (Nat, Nat, Nat)
 to produce result of expected type
   ()
+because there is no way to satisfy
```
alexandru-uta: same as above

test/fail/ok/lambdas-invariant.tc.ok
crusso: Good catch

test/fail/ok/inference.tc.ok
crusso: I think this error is case where you need to actually see the bound on T to see why it has no solution. Otherwise the message is just wrong.

crusso: The error only makes sense when you see the bound. Compare with

https://github.com/dfinity/motoko/pull/5581/files#diff-4a534eb4b454798ebfec7657f42ee4a84fbfc80b86e1a8fec92cfd471f7f7814R44
```
inference.mo:111.8-111.17: type error [M0098], cannot apply function of type
  <T <: Int>(x : T) -> T
to argument of type
  Bool
  Any
because type parameter `T` has no solution. Maybe try an explicit instantiation.
```

test/fail/ok/inference.tc.ok
crusso: Well, to use Motoko, you do need to understand subtyping.

crusso: But we could say:
because ... is not a subtype of ...

alexandru-uta: Yes, I think that's much clearer.

test/fail/ok/inference.tc.ok
alexandru-uta: I agree, but this part `because type parameter `T` has no solution` I'm not so sure that it brings much info.

test/fail/ok/inf-error-small.tc.ok
```diff
@@ -4,18 +4,24 @@ to argument of type
...
+because there is no way to satisfy subtyping
+  ?V  <:  Text  (for the return type) 
 inf-error-small.mo:46.13-46.45: type error [M0098], cannot apply function `.get` of type
   (compare : (implicit : (Nat, Nat) -> Order), n : Nat) -> ?Text
 to argument of type
   ({map : [(Nat, [var Text])]}, Text)
 to produce result of expected type
   Text
+because there is no way to satisfy subtyping
+  ?V  <:  Text  (for the return type)
```
Kamirus, line 24: Hmm, now the message does not mention the type parameter `V` anymore.
Could we simplify the message to sth like in https://github.com/dfinity/motoko/pull/5630 ?
Something like `function 'peopleMap.get' returns '?Text' but 'Text' was expected`.
Maybe that is enough info?

test/fail/ok/inference.tc.ok
crusso: I've changed the working to there's not consistent choice for type parameter `T`.
Bool <: T <: Int is the actually inconsistency, which is only there because of the bound.
Again, this will be a rage message coz noone uses bounds (our test cover corner cases too).

test/fail/ok/inf-error-small.tc.ok
crusso: In #5581, I restore displaying of bounds, so there error does mention V:
```
inf-error-small.mo:48.13-48.45: type error [M0098], cannot apply function `.get` of type
  <K, V>(compare : (implicit : (K, K) -> Order), n : K) -> ?V
to argument of type
  ({map : [(Nat, [var Text])]}, Text)
  Text
  ?V  <:  Text  (for the return type) 
  ```

Comment (crusso) [reply test/fail/ok/inf-error-small.tc.ok:24]: Maybe we could just suppress:
```
  Text
```

File: src/mo_frontend/bi_match.ml [hunk 8]
```diff
@@ -192,10 +223,22 @@ let choose_under_constrained ctx er lb c ub =
...
-  
+
+  let rec bi_match_list_result p rel eq inst any xs1 xs2 =
+    match (xs1, xs2) with
+    | x1::xs1', x2::xs2' ->
+      (match p rel eq inst any x1 x2 with
+      | Some inst -> bi_match_list_result p rel eq inst any xs1' xs2'
+      | None -> Result.Error (inst, ([x1], [x2])))
+    | [], [] -> Ok inst
+    | _, _ -> Result.Error (inst, (xs1, xs2))
```
Kamirus: `List.combine` will trap when only one of these lists is nonempty.
If we just stop of first error then we don't need lists here, just a pair

crusso: Are these lists always of the same length?

Kamirus: actually yes, they should always be equal length. because they come from list of pairs

crusso: Anyway, I fixed it

test/fail/ok/inference.tc.ok
crusso: That looked ugly in the end so I added 
...
because there is no way to satisfy **subtyping**
  <V>(x : V) -> V  <:  <U>U -> T  (for argument `f`) 
  ...

PR #5580, opened 2025-10-15
design: adds an initial design for impl blocks

christoph-dfinity:


design/ImplBlock.md
```diff
@@ -0,0 +1,55 @@
...
+}
+```
+
+Allowing multiple `impl` blocks gives more flexibility then the OO version, as functions such as `empty`, `toArray`, ... that don't require the `compare` implicit can be defined in a separate block without the `using` clause, or as standalone definitions as before.
+Functions like `map` can be defined just fine and only need to specify the _additional_ type parameters they require.
+
+
+## Variations/Future extensions
+
+- Could move the `self` name into the impl block header `impl (self : T) {}`, to get even closer to the OO notation. Might give the wrong impression that closures are involved.
```
crusso, line 52: Probably not any more confusing than the `using` parameters though

christoph-dfinity: Yeah, I thought about that as I was writing the paragraph. But then I thought I don't have a good argument for why we can't also allow empty, or multi argument lists. But that feels more and more like code obfuscation...

```motoko
impl () ...
impl (self : Nat, other : Int)
```

PR #5581, opened 2025-10-15
experiment: type argument inference errors, displaying type parameters and bounds

crusso:
incorporates #5579, but adds display of bounds so suggested instantiations and reported inconsistencies always make sense.
Also removes display of expected return type since usually also reported as failed subtyping constraint (when it matters).

Might not look like much, but a huge simplification over what was in master before the whole caffeine effort kicked in.

### DIFF
--- a/src/mo_frontend/bi_match.ml
+++ b/src/mo_frontend/bi_match.ml
@@ -196,8 +196,6 @@ let choose_under_constrained ctx er lb c ub =
     | t, _ ->
       ErrorUnderconstrained.add er lb c ub;
       if t = Non then ub else lb
-    (* | _ ->
-     fail_under_constrained lb c ub *)
 
 let bi_match_typs ctx =
   let flexible c = ConSet.mem c ctx.var_set in


src/mo_frontend/bi_match.ml
```diff
@@ -153,28 +159,23 @@ end = struct
...
-      "implicit instantiation of type parameter `%s` is over-constrained with%a\nwhere%a\nso that no valid instantiation exists"
-      (Cons.name c)
-      display_constraint (lb, c, ub)
-      display_rel (lb, "</:", ub))
-(*  else
-    raise (Bimatch (Format.asprintf
-      "type parameter `%s` has no solution. Maybe try an explicit instantiation."
-      (Cons.name c))) *)
+  error (Format.asprintf
+           "there is no consistent choice for type parameter `%s`."
```
Kamirus, line 177: Is this message better then the one before?
This error is basically raised in situations like `expected 'Text' but got 'Nat'`. I'd leave the message in either form and improve in another PR
My idea was to start with this https://github.com/dfinity/motoko/pull/5640/files#diff-6d89983d022b561d8bcfe3f0add7c51eaca6df4d62dfd760cbb77468fc9ae187R7
However without incorporating the current solution-instantiation up until the error point the message looks wrong. That is why I'd like to solve the instantiation during an error and substitute. I continue working here https://github.com/dfinity/motoko/pull/5636

PR #5585, opened 2025-10-17
feat: suggest use of contextual dot notation

crusso:
* -W M0236 enables suggestions to use the contextual dot notation
* excludes suggestions for binary `M.compare(XXX)` and `M.equal`. 
* only suggest single rewrites, making multiple suggestions for nested occurrences.

This is only intended as a caffeine training data linter. YMMV

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,7 +7,7 @@
   * Add (caffeine) warning `M0236` (#5584).
     Warns if contextual dot notation could have been used,
     e.g. `Map.filter(map, ...)` vs `map.filter(...)`.
-    Does not warn for binary `M.equals(e1, e2)` or `M.compare(e1, e2)`.
+    Does not warn for binary `M.equals(e1, e2)` or `M.compareXXX(e1, e2)`.
     (allowed by default, warn with `-W 0236`).
 
   * Add (caffeine) deprecation code `M0235` (#5583).


src/mo_frontend/typing.ml
```diff
@@ -1530,21 +1530,41 @@ type ctx_dot_candidate =
...
+let has_matching_self receiver_ty tf = match tf with
+  | T.{ lab = "Self"; typ = T.Typ con; _ } ->
+     (match Cons.kind con with
+      | T.Def(tbs, t') -> permissive_sub receiver_ty (tbs, t') <> None
+      | _ -> false)
+  | _ -> false
+
+let check_can_dot env (exp : Syntax.exp) tys es at =
+  if not env.pre then
+  if Flags.get_warning_level "M0256" <> Flags.Allow then
```
Kamirus, line 1552: M0256? That is a bug, yeah?

Kamirus: this condition is always true now

Kamirus: https://github.com/dfinity/motoko/pull/5587
Fix here

crusso: Yikes, Thanks for fixing!

crusso: Thankfully this was harmless except for perf.

PR #5589, opened 2025-10-21
feat: Parser recovery and Improved record when block expected error

Kamirus:
1. Enables parser recovery like for `moc.js` (but without allowing errors in the checking phase to limit number of errors).
2. Provides a custom syntax error message when a record is provided where block was expected.

### DIFF
--- a/src/mo_frontend/parsing.ml
+++ b/src/mo_frontend/parsing.ml
@@ -12,8 +12,6 @@ module E =
     (Parser.MenhirInterpreter)
     (Error_reporting)
 
-module S = Source.Region_set
-
 (* Define a printer for explanations. We treat an explanation as if it
    were just an item: that is, we ignore the position information that
    is provided in the explanation. Indeed, this information is hard to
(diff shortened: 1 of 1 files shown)


src/mo_frontend/parsing.ml
```diff
@@ -159,23 +240,18 @@ let parse ?(recovery = false) mode error_detail start lexer lexbuf =
   Diag.with_message_store ~allow_errors:recovery (fun m ->
     Parser_lib.msg_store := Some m;
     Parser_lib.mode := Some mode;
+    (* Avoid repeated custom errors for the same position *)
+    let reported = ref S.empty in
```
christoph-dfinity: This is a bit weird... Do you know what's causing this, or is this a common occurrence and documented for Menhir somewhere?

test/fail/ok/parse-block-or-record.tc.ok
```diff
@@ -0,0 +1,10 @@
+parse-block-or-record.mo:14.12-14.13: syntax error [M0001], expected block but got record; wrap the record in braces: { { … } }
```
christoph-dfinity: Wrapping in braces is one option, one could also parenthesize... I don't mind being opinionated here, but just wanted to call it out in case you had any thoughts.

test/fail/parse-block-or-record.mo
christoph-dfinity, line 1: Maybe also add some examples for other language constructs (anything that references `exp_nest` in the grammar is a candidate).

test/fail/ok/parse-block-or-record.tc.ok
Kamirus: I think double braces would be better, so that the block is always there and a record inside it.
I suspect that this will be the style we want to enforce, that the case always starts with a block 🤷‍♂️

test/fail/parse-block-or-record.mo
Kamirus: ah yes, good suggestion!

src/mo_frontend/parsing.ml
Kamirus: Every encountered "unexpected token" is reported, so as the parser keeps recovering and discarding tokens it would e.g. report a syntax error for every `=` token in a record-block case.
I gather all of these syntax errors and discard them, emitting a single custom error at the first `{`.
And about Menhir: the main parser loop is not from menhir (it has no error recovery) but from a custom library that we use currently for `moc.js`

Kamirus: I've expanded the comment, maybe now it is more clear

src/mo_frontend/parser.mly
```diff
@@ -823,9 +832,17 @@ exp [@recover.expr mk_stub_expr loc] (B) :
     { e }
 
 block :
-  | LCURLY ds=seplist(dec, semicolon) RCURLY
+  | LCURLY ds=seplist_er(dec, block_dec_error, semicolon) RCURLY
     { BlockE(ds) @? at $sloc }
 
+// When block is actually a record, emit custom errors
```
crusso, line 838: I didn't know you could use line comments

Kamirus: Not everywhere, inside blocks you need `(* *)`, but outside the line comments work

PR #5597, opened 2025-10-22
feat: Warn typed anonymous function

Kamirus:
Add (caffeine) warning `M0238`
Warns if type annotations are used on anonymous functions.
Used to avoid unnecessary type annotations and rely on explicit type instantiations to propagate the type information instead.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -179,59 +179,31 @@ jobs:
           gh pr close ${{ github.event.pull_request.number }} --delete-branch --comment "CI looks good, this dependency bump would not cause problems, hence closing this trial PR."
           gh pr edit ${{ github.event.pull_request.number }} --remove-label "autoclose"
 
-# Handles open PRs that address `master` and only have one changed file. The body checks whether that is the `Changelog.md` (since the `if:` condition can't). If so, the PR gets approved with a comment.
-  approvals1:
+  approvals:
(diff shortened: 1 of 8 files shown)


test/run/warn-typed-anonymous-func.mo
```diff
@@ -0,0 +1,12 @@
+//MOC-FLAG -W M0238
+let n = 1;
+
+func foo<T>(x : T, f : (T, T) -> T) : T = f(x, x);
+
+ignore foo(n, func (x, y) = x + y); // no warning
+
+// With Warnings
+ignore foo(n, func (x : Nat, y : Nat) : Nat = x + y);
```
christoph-dfinity, line 9: What's special about anonymous functions here? Shouldn't this warn just the same?

```
ignore foo(n, func bar(x : Nat, y : Nat) : Nat = x + y);
```

Kamirus: Unfortunately it is difficult to distinguish top-level (func declarations) from lambda expressions (local?).
This is the easiest and close enough solution. Otherwise we would need to somehow pass that info, it is not even easily accessible in the parser ;c

christoph-dfinity: I don't think top-level vs local is really a thing. We could try to distinguish between funcs in a block scope and those without.

(The parser desugars one into the other at the moment 😢)

Wouldn't this be easier to do with a lintoko rule?

Kamirus: Is it possible with a lintoko rule?

Kamirus: aha, for anonymous functions it would be possible

PR #5599, opened 2025-10-23
feat: Warn unit binding

Kamirus:
Warn when binding a unit `()` value to let/var. For example:
```motoko
let va2 = VarArray.sortInPlace(va); // warning
```
This is only covering let/var declarations. One could imagine extending it to all contexts, e.g. `[VarArray.sortInPlace(va)]` or `f(VarArray.sortInPlace(va))` etc...

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,7 +4,13 @@
 
 * motoko (`moc`)
 
-  * Add warning `M0239` that warns when binding a unit `()` value to `let` or `var` (#5599).
+  * Add warning `M0239` that warns when binding a unit `()` value by `let` or `var` (#5599).
+
+  * Use `self` parameter, not `Self` type, to enable contextual dot notation (#5574).
+
+  * Enable parser recovery to gather more syntax errors at once (previously only enabled for `moc.js`) (#5589).
+
+  * Custom syntax error message when a record is provided where a block is expected (#5589).
 
   * Add (caffeine) warning `M0237` (#5588).
     Warns if explicit argument could have been inferred and omitted,
@@ -27,6 +33,8 @@
   * Experimental support for Mixins (#5459).
 
   * bugfix: importing of `blob:file:` URLs in subdirectories should work now (#5507, #5569).
+
+  * bugfix: escape `composite_query` fields on the Candid side, as it is a keyword (#5617).
 
 ## 0.16.3 (2025-09-29)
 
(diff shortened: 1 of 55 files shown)


src/mo_frontend/typing.ml
```diff
@@ -4024,28 +4032,31 @@ and infer_dec env dec : T.typ =
...
+        | VarP id -> use_identifier env id.it
+        | _ -> ())
+    | Some fail ->
+      if not env.pre then
+        check_exp env T.Non fail
+    );
     let t = infer_exp env exp in
     if !Flags.typechecker_combine_srcs then
       combine_pat_srcs env t pat;
+    if not env.pre && T.is_unit (T.normalize t) then
```
ggreif, line 4040: This would also trigger with `let () = ...`. Would we want that?

Kamirus: Hm, actually I don't think there is a value in `let () = e;` over just `e;`.
It's not like in ocaml that `let () = e in` can be actually useful.
I don't see value in allowing `let ()`.

I've thought about allowing `let <id> = ()`, but for now I've chosen the simplest solution

Changelog.md
```diff
@@ -4,6 +4,8 @@
 
 * motoko (`moc`)
 
+  * Add warning `M0239` that warns when binding a unit `()` value to `let` or `var` (#5599).
```
```suggestion
  * Add warning `M0239` that warns when binding a unit `()` value by `let` or `var` (#5599).
```

src/mo_frontend/typing.ml
```diff
@@ -4024,28 +4032,31 @@ and infer_dec env dec : T.typ =
...
-  | LetD (pat, exp, None) ->
-    (* For developer convenience, ignore top-level actor and module identifiers in unused detection. *)
-    (if env.in_prog && CompUnit.(is_actor_def exp || is_module_def exp) then
-      match pat.it with
-      | VarP id -> use_identifier env id.it
-      | _ -> ());
-    let t = infer_exp env exp in
-    if !Flags.typechecker_combine_srcs then
-      combine_pat_srcs env t pat;
-    t
```
ggreif, line 4026: This part is gone, intended?

Kamirus: It is still there, below.
I've refactored both LetD into a single one so I don't have to duplicate the code

src/mo_frontend/typing.ml
ggreif: Gotcha!

ggreif (approved):
Looks good!

PR #5604, opened 2025-10-23
experiment: use extra productions to detect records like blocks

crusso:


doc/md/examples/grammar.txt
```diff
@@ -345,6 +345,10 @@
     <exp_nondec>
     'let' <pat> '=' <exp> 'else' <exp_nest>
 
+<block_dec> ::= 
+    <dec>
+    <id> '=' <exp>
```
crusso, line 350: We should remove this production using the sed script

crusso:
subsumed by #5589

PR #5605, opened 2025-10-24
fix: Redundant type instantiation to check all places and avoid false positives

Kamirus:
Fix the `M0223` warning check for redundant type instantiations to avoid false positives.
This check is now correctly performed during `not env.pre` phase like all other warnings.
**Also:** Correctly handles nested expressions, further avoiding false positives, e.g. can report 2 warnings, where the 1st or the 2nd instantiation is actually necessary for type-checking.

Important notes:
- Redundant type instantiation needs backtracking, it first tries without the explicit instantiation, if it fails then continues with the instantiation
- It must be performed during `not env.pre` phase. Otherwise the program is not thoroughly checked (`env.pre` skips parts of the program)
- If the instantiation **IS** redundant (and the warning is reported) it **MUST NOT** check again with the instantiation as it can lead to **false positives**. So it means that `infer_exp` must be used instead of `check_exp` when redundant

Implementation:
- When backtracking from the redundancy check it reverts all committed state changes
- These state changes are tracked as `() -> ()` functions and added by 2 functions `commit` and `commit_note`


src/mo_frontend/typing.ml
```diff
@@ -2951,18 +2970,17 @@ and infer_call_instantiation env t1 ctx_dot tbs t_arg t_ret exp2 at t_expect_opt
...
         (* [codom] comes from [ts2] which might contain unsolved type variables. *)
-        let closed = Bi_match.is_closed remaining codom in
-        if not env.pre && (closed || body_typ <> codom) then begin
-          (* Closed [codom] implies closed [body_typ].
-            * [body_typ] is closed when it comes from [typ_opt] (which is when it is different from [codom])
-            *)
+        let closed_codom = Bi_match.is_closed remaining codom in
+        (* Closed [codom] implies closed [body_typ]. [body_typ] is closed when it comes from [typ_opt] *)
+        let closed_body_typ = closed_codom || Option.is_some typ_opt in
+        let env' = if closed_body_typ then env' else { env' with rets = None } in
```
crusso: Hmm, if `env'.rets = None`, that means the body can't `return` at all which is weird because functions can always `return`. There's only a few places where we disallow returns (e.g. object/class constructors) so maybe we should discuss.
Are you ever checking the body of a function with `env'.rets = None`? If not, probably ok.

crusso (commented):
Silly thought, and I haven't thought it through, but I wonder if one could get away with only updating the notes once, when they contain T.Pre (or a PreLit), and just not apply any further update.

src/mo_frontend/typing.ml
crusso: Oh, I think you are at line 2989

Kamirus: That is a good point. It's better to distinguish between None and Some here for `rets`
It should be more like `type rets = NotAllowed | RetBimatch | Ret typ`.
And a custom error when trying to check when `RetBimatch` saying that cannot infer type of return etc.

src/mo_frontend/typing.ml
```diff
@@ -2060,11 +2078,13 @@ and infer_exp'' env exp : T.typ =
   | RetE exp1 ->
     if not env.pre then begin
       match env.rets with
-      | Some T.Pre ->
+      | Ret T.Pre ->
         local_error env exp.at "M0084" "cannot infer return type"
```
crusso, line 2113: (Interesting - it seems we do allow Ret T.Pre, maybe that would actually be enough rather than the more complicated solution.
I wonder if this code was ever exercised by a test. I'm not sure what would cause it.)

src/mo_frontend/typing.ml
```diff
@@ -2060,11 +2078,13 @@ and infer_exp'' env exp : T.typ =
...
-      | Some t ->
+      | Ret t ->
         check_exp_strong env t exp1
-      | None ->
+      | BimatchRet k ->
+        k env exp1
```
crusso: This will actually wind up doing infer_exp rather than check_exp_strong (as on 2084). God knows if that would make a difference.

Kamirus: Doing infer is expected, but true I can pass the `weak = false` here
Continuation in a separate PR here https://github.com/dfinity/motoko/pull/5615

src/mo_frontend/typing.ml
Kamirus: Yeah we can provide a better error, but the `BimatchRet` is actually improving the type inference and avoids the error

Kamirus:
Backtracking trail in `pre` phase is good enough approximation for now, with limited harm.

PR #5607, opened 2025-10-24
feat(moc.js): add `blob:` import placeholder setting

rvanasa:
Possible (hacky) solution for https://github.com/dfinity/vscode-motoko/issues/394. Tested in VS Code and works as expected.

Adds a `setBlobImportPlaceholders(boolean)` function to `moc.js` which ignores unresolved `blob:` imports (trapping at runtime) when enabled. This is a workaround for the virtual file system currently only including Motoko and Candid files from the VS Code workspace.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,8 @@
 
 * motoko (`moc`)
 
+  * Use `self` parameter, not `Self` type, to enable contextual dot notation (#5574).
+
   * Enable parser recovery to gather more syntax errors at once (previously only enabled for `moc.js`) (#5589).
 
   * Custom syntax error message when a record is provided where a block is expected (#5589).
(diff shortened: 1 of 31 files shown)


ggreif (commented):
One could also use https://ocsigen.org/js_of_ocaml/latest/api/js_of_ocaml/Js_of_ocaml/Sys_js/index.html#val-create_file
to just male the file and then let the rest of the mechanism import it.

And there is the `mount` function too, which would actually allow importing the exact content.

_Edit_: Oh, I see, having that module conditionally included may be tricky.

src/lowering/desugar.ml
```diff
@@ -1215,9 +1215,13 @@ and transform_import (i : S.import) : Ir.dec list =
     | S.IDLPath (fp, canister_id) ->
       primE (I.ActorOfIdBlob t) [blobE canister_id]
     | S.ImportedValuePath path ->
-       let contents = Lib.FilePath.contents path in
-       assert T.(t = Prim Blob);
-       blobE contents
+       if !Mo_config.Flags.blob_import_placeholders then
+         raise (Invalid_argument ("blob import placeholder"))
```
ggreif, line 1219: Alright, block the desugaring!

PR #5609, opened 2025-10-25
experiment: implement auto-import for unbound variables that could be libs

crusso:
Flag `--implicit-package <package-name>` auto-imports
* libraries from `package-name` for implicit arguments
* libraries from `package-name` for contextual dots

but not libraries for unbound identifiers, e.g. lib `Nat` for `Nat` or `Nat.toText`.

This PR auto-imports for unbound identifiers too, make the approach more consistent and eliminating many 
M0057 errors seen with caffeine inference.

TODO:
- [x] testing
- [x] don't auto-import ambiguous List, pure/List etc.
- [x] issue a warning? Do we, for the other two cases? (No, postpone)
- [x] handle a few more cases (type paths?) (No, postpone)

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -2194,7 +2194,6 @@ and infer_bin_exp env exp1 exp2 =
    `Error` with the type of the receiver as well as the error message
    to report. This is used to delay the reporting for contextual dot resulution *)
 and try_infer_dot_exp env at exp id (desc, pred) =
-  (* CRUSSO *)
   let t0, t1 = infer_exp_and_promote env exp in
   let fields =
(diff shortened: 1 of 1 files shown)


christoph-dfinity:
With the combination of implicit search, contextual dot search, and now this, wouldn't it be easier to just have the `--implicit-package` add imports for every library in the package to the type checkers env?

src/mo_frontend/typing.ml
```diff
@@ -2173,6 +2194,7 @@ and infer_bin_exp env exp1 exp2 =
    `Error` with the type of the receiver as well as the error message
    to report. This is used to delay the reporting for contextual dot resulution *)
 and try_infer_dot_exp env at exp id (desc, pred) =
+  (* CRUSSO *)
```
```suggestion
```

PR #5615, opened 2025-10-28
fix: Type inference of `return` when bimatching

Kamirus:
Fixes type inference of deferred funcs that use return in their body.
Before: confusing error e.g. Bool does not have expected type T
Now: it should type check successfully

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -4007,6 +4007,14 @@ and infer_block_exps env decs : T.typ =
     if not env.pre then recover (check_dec env T.unit) dec;
     infer_block_exps env decs'
 
+and warn_unit_binding binder env (dec : dec) (exp : exp) =
+  let binder = match binder with
+    | `Let -> "let"
+    | `Var -> "var"
+  in
+  let at = Source.{dec.at with right = exp.at.left} in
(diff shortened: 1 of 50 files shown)


src/mo_frontend/typing.ml
```diff
@@ -2060,11 +2063,13 @@ and infer_exp'' env exp : T.typ =
...
       match env.rets with
-      | Some T.Pre ->
+      | Ret T.Pre ->
         local_error env exp.at "M0084" "cannot infer return type"
-      | Some t ->
+      | Ret t ->
         check_exp_strong env t exp1
-      | None ->
+      | BimatchRet k ->
+        k env exp1
```
Kamirus, line 2071: I'm not sure making `weak = false` is correct here.
I think it is better to leave it as it, so that the `return e` is inferred just like the rest of the body of the func.

For example assume this is a body of a deferred func:
```motoko
if (...) return e1;
e2
```
IMO both `e1` and `e2` (the rest of the body) should be inferred the same way.
If we infer `e2` then we should also infer `e1`.
When we have `closed_body_typ` then both `e1` and `e2` are checked
(with a small difference that `e2` would be `check_exp_strong` instead of `check_exp`)

When `weak=true` we might generate the `"M0155" "operator may trap for inferred type%a"` warning.
This is the only current use of `weak`

PR #5619, opened 2025-10-29
fix: catch up with Candid spec changes (`reserved` ↛ `null`)

### LINKED ISSUE #5543: candid: spec compliance
Test against Martin's recent changes regarding (not) decoding `reserved` to `null` type. 

https://github.com/dfinity/candid/pull/680

ggreif:
Reproduce issue #5543. Fixes #5543.

The spec says that every Candid value is a subtype of `opt T`, and gives `opt t` when `t <: T`. But the `null` type has no subtypes. I.e. we only get out a `null : Null` when the binary Candid has a `0x7F` there. In all other cases the coercion must fail, either trapping or indicating a coercion failure to the caller (when recoverable). However `null : ?T` can always be materialised when an argument is missing, similarly `null : Null`. So we have to treat missing arguments and coercion failures differently when the requested type is `Null` (or `null` respectively, in Candid).

This PR corrects the reading of (e.g. `\x70`, `<reserved>`) in the non-recoverable case and makes it a coercion failure.

Also adds improvements to the `candid-tests` runner to
- not ignore the case when a field type can be inferred, but not in the record type (eliminating `M0164`, ignored test)
- adds `null` values to top-level argument tuples where the type requests them
- adds `null` fields to record values (in argument sequences) where the type requests them

### DIFF
--- a/src/mo_idl/idl_to_mo_value.ml
+++ b/src/mo_idl/idl_to_mo_value.ml
@@ -90,17 +90,17 @@ let rec value v t =
     (Diag.error_message v.at "M0165" "import" "odd expected type"))
 
 let rec args vs = function
-  | ts when List.(compare_lengths vs.it ts < 0 && for_all null (Lib.List.drop (length vs.it) ts)) ->
+  | ts when List.(compare_lengths vs.it ts < 0 && for_all is_null (Lib.List.drop (length vs.it) ts)) ->
     let vs' = vs.it @ Lib.List.replicate { vs with it = NullV } List.(length ts - length vs.it) in
     args {vs with it = vs'} ts
   | ts when List.(exists (fun (t, v) -> apart t v.it) (combine ts vs.it)) ->
     args {vs with it = List.map2 enrich ts vs.it} ts
   | ts -> parens_comma (List.map2 value vs.it ts)
-and null t = t = T.(Prim Null)
+and is_null t = t = T.(Prim Null)
 and apart t v = match t, v with
   | T.(Obj (Object, tfs)), RecordV vfs ->
        let defaultable = diff tfs (List.map (fun {it; _} -> Idl_to_mo.check_label (fst it)) vfs) in
-       defaultable <> [] && List.for_all (fun {T.typ; _} -> null typ) defaultable
+       defaultable <> [] && List.for_all (fun {T.typ; _} -> is_null typ) defaultable
   | _ -> false
 and enrich t v = match t, v.it with
   | T.(Obj (Object, tfs)), RecordV vfs ->


src/codegen/compile_enhanced.ml
```diff
@@ -8540,6 +8545,7 @@ module Serialization = struct
         ) ^^
         get_x ^^
         Tagged.allocation_barrier env)
+      (*| Prim Null -> failwith "Null"*)
```
```suggestion
```

src/codegen/compile_enhanced.ml
```diff
@@ -8041,6 +8041,11 @@ module Serialization = struct
         compile_eq_const (Wasm.I64_convert.extend_i32_s (Int32.neg (Option.get (to_idl_prim Candid t))))
       in
 
+      let _check_reserved () =
+        get_idltyp ^^
+        compile_eq_const (Wasm.I64_convert.extend_i32_s (-16l))
+      in
+
```
```suggestion
```

src/mo_idl/idl_to_mo_value.ml
```diff
@@ -32,11 +32,21 @@ let is_tuple_rec fs =
...
-      (Diag.error_message f.at "M0164" "import" "unknown record or variant label in textual representation"))
+    match infer (snd (f.it)).it with
+    | Some t -> t
+    | _ ->
+      raise (UnsupportedCandidFeature
+               (Diag.error_message f.at "M0164" "import" "unknown record or variant label in textual representation"))
+
+let infer_typ = function
+  | NullV -> Some (T.(Prim Null))
+  (*| NumV n when n < 0 -> Some (T.(Prim Int))*)
```
ggreif: TODO!

src/mo_idl/idl_to_mo_value.ml
```diff
@@ -79,4 +89,23 @@ let rec value v t =
   | _ -> raise (UnsupportedCandidFeature
     (Diag.error_message v.at "M0165" "import" "odd expected type"))
 
-let args vs ts = parens_comma (List.map2 value vs.it ts)
+let rec args vs = function
+  | ts when List.(compare_lengths vs.it ts < 0 && for_all null (Lib.List.drop (length vs.it) ts)) ->
```
ggreif: Extends values to be checked when shorter than types and those are all `Null`.

src/mo_idl/idl_to_mo_value.ml
```diff
@@ -79,4 +89,23 @@ let rec value v t =
...
+    let vs' = vs.it @ Lib.List.replicate { vs with it = NullV } List.(length ts - length vs.it) in
+    args {vs with it = vs'} ts
+  | ts when List.(exists (fun (t, v) -> apart t v.it) (combine ts vs.it)) ->
```
ggreif, line 96: Adds `null`-ed fields when the object type calls for them.

src/idllib/idlHash.ml
```diff
@@ -1,8 +1,7 @@
-(* IDL field hashes *)
+(* IDL field hashes, called by unescape_hash *)
 let idl_hash : string -> Lib.Uint32.t = fun s ->
   let open Lib.Uint32 in
   List.fold_left
     (fun s c -> add (mul s (of_int 223)) (of_int (Char.code c)))
     zero
-    (* TODO: also unescape the string, once #465 is approved *)
```
ggreif, line 7: This was actually resolved back then.

src/codegen/compile_classical.ml
```diff
@@ -8322,21 +8322,25 @@ module MakeSerialization (Strm : Stream) = struct
...
 
         G.concat_map (fun t ->
-          let can_recover, default_or_trap = Type.(
+          let can_recover, argument_default_or_trap, coercion_default_or_trap = Type.(
+            let null_result _ = Opt.null_lit env in
             match normalize t with
-            | Prim Null | Opt _ | Any ->
-              (Bool.lit true, fun msg -> Opt.null_lit env)
+            | Prim Null ->
+              (get_can_recover, null_result, fun _ -> compile_unboxed_const (coercion_error_value env))
```
crusso, line 8329: I don't quite grok why this case is different. Can you explain?

ggreif: Yes, the spec says that every Candid value is a subtype of `opt T`, and gives `opt t` when `t <: T`. But the `null` type has no subtypes. I.e. we only get out a `null : Null` when the binary Candid has a `0x7F` there. In all other cases the coercion must fail, either trapping or indicating a coercion failure to the caller (when recoverable). However `null : ?T` can always be materialised when an argument is missing, similarly `null : Null`. So we have to treat missing arguments and coercion failures differently when the requested type is `Null` (or `null` respectively, in Candid).

These are coercion failures:
```
assert blob "DIDL\00\01\70"            !: (null) "parsing reserved as null";
assert blob "DIDL\00\02\70\70"         !: (null, null) "parsing (reserved, reserved) as (null, null)";
```

test/run-drun/ok/candid_opts.drun-run.ok
```diff
@@ -18,3 +18,15 @@ debug.print: {o1 = {x = null}; o2 = {x = ?null}; o3 = {x = ?(?null)}}
 debug.print: {o1 = {x = null}; o2 = {x = null}; o3 = {x = null}}
 debug.print: {o1 = {x = null}; o2 = {x = ?null}; o3 = {x = ?(?null)}}
 ingress Completed: Reply: 0x4449444c0000
+ingress Err: IC0503: Error from Canister rwlgt-iiaaa-aaaaa-aaaaa-cai: Canister called `ic0.trap` with message: 'IDL error: unexpected IDL type when parsing Null'.
```
ggreif, line 21: These are totally expected. A `<reserved>` marker (i.e. `\x70` = -16) is not digestible by `null : Null` pattern, similarly a `5 : nat` (here `x7D\x05`).

src/exes/candid_tests.ml
```diff
@@ -252,7 +252,7 @@ let () =
             (* Printf.printf "\n%s" src *)
             Unix.putenv "MOC_UNLOCK_PRIM" "yesplease";
             write_file "tmp.mo" src;
-            match run_cmd "moc -Werror -wasi-system-api tmp.mo -o tmp.wasm" with
+            match run_cmd "moc -A M0215 -Werror -wasi-system-api tmp.mo -o tmp.wasm" with
```
ggreif, line 255: The Motoko code generated by these
```
assert blob "DIDL\01\6c\01\00\7d\01\00\05" == "(record { 0 = 5 })" : (record { 1 : null }) "parsing into record with expected field that is greater than extra field on the wire";
assert blob "DIDL\01\6c\01\01\7d\01\00\05" == "(record { 1 = 5 })" : (record { 0 : null }) "parsing into record with expected field that is less than extra field on the wire";
```
would contain dead fields, which cause these warnings. I couldn't come up with a massaging so that the warning goes away, that doesn't change the meaning of the test. So I have just suppressed the warning so that `-Werror` won't kill compilation.

src/codegen/compile_classical.ml
ggreif: Though, I must admit, I cannot really wrap my head around this one. @mraszyk can you give a hint? (The test always passed, so just quenching my curiosity.)
```
// parsing reserved as null
assert blob "DIDL\00\01\70" == "(null)" : (reserved) "reserved";
```

ggreif: These test the `null`-materialisation for missing top-level args, i.e. `argument_default_or_trap`:
```
assert blob "DIDL\00\00" == "()" : (null) "parsing an empty top-level tuple into a longer top-level tuple";
assert blob "DIDL\00\02\7d\7d\05\06" == "(5, 6)" : (nat, nat, null) "parsing a top-level tuple into a longer top-level tuple";
```

src/mo_idl/idl_to_mo_value.ml
ggreif: In
```
assert blob "DIDL\01\6c\01\01\7d\01\00\05" == "(record { 1 = 5 })" : (record { 0 : null }) "parsing into record with expected field that is less than extra field on the wire";
```
The `record { 1 = 5 }` fragment need an augmenting to `record { 1 = 5; 0 = null }` so that the Motoko generation creates type-correct code.

src/mo_idl/idl_to_mo_value.ml
```diff
@@ -79,4 +89,23 @@ let rec value v t =
...
+    args {vs with it = List.map2 enrich ts vs.it} ts
+  | ts -> parens_comma (List.map2 value vs.it ts)
```
ggreif, line 98: This is the vanilla case, and we should exit here after all massaging from above is done.

src/mo_idl/idl_to_mo_value.ml
```diff
@@ -79,4 +89,23 @@ let rec value v t =
...
+and null t = t = T.(Prim Null)
```
ggreif: rename to `is_null`

src/mo_idl/idl_to_mo_value.ml
```diff
@@ -53,9 +63,9 @@ let rec value v t =
   | BlobV b, T.Array _ ->
     brackets_comma (List.of_seq (Seq.map (fun c -> Printf.sprintf "%d" (Char.code c)) (String.to_seq b)))
   | TextV s, _ -> text_lit s
-  | RecordV fs, T.Obj (T.Object, tfs) ->
+  | RecordV fs, T.(Obj (Object, tfs)) ->
     "{" ^ String.concat "; " (List.map (fun f ->
-      Idl_to_mo.check_label (fst f.it) ^ " = " ^ value (snd f.it) (find_typ tfs f)
+      Idl_to_mo.check_label (fst f.it) ^ " = " ^ value (snd f.it) (find_typ ~infer:infer_typ tfs f)
```
ggreif, line 68: This deals with previously errored (and ignored) test cases
```
construct:233 parsing into record with expected field that is greater than extra field on the wire ... ignored ((string):1.11-1.16: import error [M0164], unknown record or variant label in textual representation
)
construct:234 parsing into record with expected field that is less than extra field on the wire ... ignored ((string):1.11-1.16: import error [M0164], unknown record or variant label in textual representation
```
arising from
```
assert blob "DIDL\01\6c\01\00\7d\01\00\05" == "(record { 0 = 5 })" : (record { 1 : null }) "parsing into record with expected field that is greater than extra field on the wire";
assert blob "DIDL\01\6c\01\01\7d\01\00\05" == "(record { 1 = 5 })" : (record { 0 : null }) "parsing into record with expected field that is less than extra field on the wire";
```

src/codegen/compile_classical.ml
crusso:
```
// parsing reserved as null
assert blob "DIDL\00\01\70" == "(null)" : (reserved) "reserved";
```
Just to clarify, it think the comment should reading `// parsing reserved as null value of type reserved`, if that helps any.

ggreif: `\70` is a value of type `reserved`, so I don't see how a `null` appears suddenly in the game...

src/mo_idl/idl_to_mo_value.ml
```diff
@@ -79,4 +89,23 @@ let rec value v t =
...
 
-let args vs ts = parens_comma (List.map2 value vs.it ts)
+let rec args vs = function
+  | ts when List.(compare_lengths vs.it ts < 0 && for_all is_null (Lib.List.drop (length vs.it) ts)) ->
+    let vs' = vs.it @ Lib.List.replicate { vs with it = NullV } List.(length ts - length vs.it) in
+    args {vs with it = vs'} ts
+  | ts when List.(exists (fun (t, v) -> apart t v.it) (combine ts vs.it)) ->
+    args {vs with it = List.map2 enrich ts vs.it} ts
+  | ts -> parens_comma (List.map2 value vs.it ts)
+and is_null t = t = T.(Prim Null)
```
crusso, line 99:
```suggestion
and is_null_opt_reserved t = 
```
Should this maybe allow `Null, Opt t an Any` (after normalizing)?

src/mo_idl/idl_to_mo_value.ml
```diff
@@ -79,4 +89,23 @@ let rec value v t =
   | _ -> raise (UnsupportedCandidFeature
     (Diag.error_message v.at "M0165" "import" "odd expected type"))
 
-let args vs ts = parens_comma (List.map2 value vs.it ts)
```
crusso, line 82: I wonder if this code might be simpler if you just traversed the typs in ts and only allow trailing defaultable types?

src/mo_idl/idl_to_mo_value.ml
```diff
@@ -79,4 +89,23 @@ let rec value v t =
...
+let rec args vs = function
+  | ts when List.(compare_lengths vs.it ts < 0 && for_all is_null (Lib.List.drop (length vs.it) ts)) ->
+    let vs' = vs.it @ Lib.List.replicate { vs with it = NullV } List.(length ts - length vs.it) in
+    args {vs with it = vs'} ts
+  | ts when List.(exists (fun (t, v) -> apart t v.it) (combine ts vs.it)) ->
+    args {vs with it = List.map2 enrich ts vs.it} ts
+  | ts -> parens_comma (List.map2 value vs.it ts)
+and is_null t = t = T.(Prim Null)
+and apart t v = match t, v with
```
crusso, line 100: Ditto, this code might be simpler if you sorted the value and type fields by hash and then traverse the (type) fields in order, inserting preserving fields and inserting null fields when the type is defaultable.

Not sure.

src/mo_idl/idl_to_mo_value.ml
ggreif: Trying to decode your comment. Line 93 applies in this case:
https://github.com/dfinity/candid/blob/master/test/construct.test.did#L232
i.e.
```
assert blob "DIDL\00\02\7d\7d\05\06" == "(5, 6)" : (nat, nat, null) "parsing a top-level tuple into a longer top-level tuple";
```
by wanting to generate a Motoko assert
``` Motoko
assert (prim "deserialize" : Blob -> (Nat, Nat, Null)) "\44\49\44\4C\00\02\7D\7D\05\06" == ((5, 6) : (Nat, Nat, Null))
```
But `ts` has length 3 and `vs` has length 2, so that would be a type error. Besides `parens_comma` in line 98 uses `List.map2`, which crashes, as the lengths of the input lists don't line up. The solution is to append `null`s to `vs` to get the current
``` Motoko
assert (prim "deserialize" : Blob -> (Nat, Nat, Null)) "\44\49\44\4C\00\02\7D\7D\05\06" == ((5, 6, null) : (Nat, Nat, Null))
```
Am I missing something? There is nothing to be allowed here. We just tweak the expectation to make it make sense.

src/mo_idl/idl_to_mo_value.ml
ggreif: Well, I guess I can drop the `apart` guard and `enrich` unconditionally. Simpler.

ggreif: See #5638.

PR #5625, opened 2025-11-03
fix: record update should type-check like record creation

### LINKED ISSUE #5628: Improve type inference upon record update
(As already discussed with Kamil)
Type inference on record update is not as strong as on record creation:
```
  public shared ({ caller }) func addWorkout(id : Text, exercises : [Exercise], duration : Nat) : async () {
    let workout : Workout = {
      exercises = Map.fromIter(exercises.enumerate()); //here NO explicit type annotation (<Nat,Exercise>) necessary
      duration;
      timestamp = Time.now();
    };
    workouts.add(id, workout);
  };
```
vs. 
```
  public shared ({ caller }) func updateWorkout(id : Text, exercises : [Exercise], duration : Nat) : async () {
    let existingWorkout = getWorkoutInternal(id);
    let updatedWorkout : Workout = {
      existingWorkout with
      exercises = Map.fromIter<Nat,Exercise>(exercises.enumerate()); //here explicit type annotation (<Nat,Exercise>) necessary
      duration;
    };
    workouts.add(id, updatedWorkout);
  };
```

Would be great to also make this implicit typing consistent for both cases. 
Thank you!

Kamirus:
Improve type inference for record update syntax. Example
```motoko
let w : Workout = {
  workout with
  exercises = Map.empty(); // should typecheck without type instantiation
  duration;
};
```
fixes #5628

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -4021,7 +4021,7 @@ and warn_unit_binding binder env (dec : dec) (exp : exp) =
     | `Var -> "var"
   in
   let at = Source.{dec.at with right = exp.at.left} in
-  warn env at "M0239" "Avoid binding a unit '()' result; remove '%s' and keep the expression" binder
+  warn env at "M0239" "Avoid binding a unit `()` result; remove `%s` and keep the expression" binder
 
 and infer_dec env dec : T.typ =
   let t =
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,8 @@
 
   * Improved type inference of the record update syntax (#5625).
 
+  * New flag `--error-recovery` to enable reporting of multiple syntax errors (#5632).
+
   * Improved solving and error messages for invariant type parameters (#5464).
     Error messages now include suggested type instantiations when there is no principal solution.
 
--- a/src/codegen/compile_enhanced.ml
+++ b/src/codegen/compile_enhanced.ml
@@ -496,7 +496,7 @@ module E = struct
 
     (* Counter for deriving a unique id per constant function. *)
     constant_functions : int32 ref;
-    dedup2 : (unit -> int32) option ref;
+    dedup : (unit -> int32) option ref;
   }
 
 
@@ -531,7 +531,7 @@ module E = struct
     requires_stable_memory = ref false;
     global_type_descriptor = ref None;
     constant_functions = ref 0l;
-    dedup2 = ref None;
+    dedup = ref None;
   }
 
   (* This wraps Mo_types.Hash.hash to also record which labels we have seen,
@@ -802,13 +802,13 @@ module E = struct
       [ nr {mtype = MemoryType ({min = Int64.zero; max = None}, I64IndexType)} ]
     | _ -> []
 
-  let get_dedup (env : t) =
-    match !(env.dedup2) with
+  let get_dedup (env : t) : int32 =
+    match !(env.dedup) with
     | Some mk_fi -> mk_fi()
-    | _ -> assert false
+    | None -> assert false
 
-  let set_dedup (env : t) mk_fi =
-    env.dedup2 := Some mk_fi
+  let set_dedup (env : t) (mk_fi : unit -> int32)=
+    env.dedup := Some mk_fi
 
 end
 
@@ -6919,7 +6919,15 @@ module Internals = struct
   let add_cycles env ae = call_prelude_function env ae "@add_cycles"
   let reset_cycles env ae = call_prelude_function env ae "@reset_cycles"
   let reset_refund env ae = call_prelude_function env ae "@reset_refund"
-  let dedup2 env = G.i (Call (nr (E.get_dedup env))) 
+
+  let register_dedup env ae =
+    match VarEnv.lookup_var ae "@dedup" with
+    | Some (VarEnv.Const Const.Fun (_, mk_fi, _)) ->
+      E.set_dedup env mk_fi
+    | _ -> assert false
+
+  let dedup env = G.i (Call (nr (E.get_dedup env)))
+
 end
 
 module Serialization = struct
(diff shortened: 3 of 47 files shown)


src/mo_frontend/typing.ml
```diff
@@ -2237,7 +2237,7 @@ and infer_check_bases_fields env (check_fields : T.field list) exp_at exp_bases
 
   let infer_or_check rf =
     let { mut; id; exp } = rf.it in
-    match List.find_opt (fun ft -> ft.T.lab = id.it) check_fields with
+    match List.find_opt T.(fun ft -> ft.lab = id.it && not (is_typ ft.typ)) check_fields with
```
Kamirus: The `rf` are always value-level, yes? So we need to skip the `type <id> = <typ>` fields.
`normalize` not necessary, right?

src/mo_frontend/typing.ml
```diff
@@ -729,6 +729,10 @@ let infer_class_cap env obj_sort (tbs : T.bind list) cs =
   | _ ->
     C.NullCap, tbs, cs
 
+let find_field (ef : exp_field) (fts : T.field list) = 
+  let id = ef.it.id.it in
+  List.find_opt T.(fun ft -> ft.lab = id && not (is_typ ft.typ)) fts
```
Kamirus: `exp_field` is always value-level, yes? So we need to skip the `type <id> = <typ>` T.fields.
`normalize` not necessary, right?

src/mo_frontend/typing.ml
```diff
@@ -2411,22 +2414,24 @@ and check_exp' env0 t exp : T.typ =
...
+      |> List.map (fun ft -> Printf.sprintf "'%s'" ft.T.lab)
+    in
+    begin match missing_field_labs with
+    | [] -> check_inferred env0 env t t' exp
+    | fts ->
+      (* Future work: Replace this error with a general subtyping error once better explanations are available. *)
+      let s = if List.length fts = 1 then "" else "s" in
+      local_error env exp.at "M0151" "missing field%s %s from expected type%a" s (String.concat ", " fts) display_typ_expand t;
+      t'
+    end
```
Kamirus: All of this should be redundant (replaced with just `check_inferred env0 env t t' exp`) once we have time to complete https://github.com/dfinity/motoko/pull/5409

src/mo_frontend/typing.ml
crusso: Normalize should not be necessary to distinguish a type field from a value field, no. Ditto for T.mut

crusso: There should be a Type.find_val_field_opt or similar that does this for I think.

src/mo_frontend/typing.ml
```diff
@@ -2411,22 +2414,24 @@ and check_exp' env0 t exp : T.typ =
...
-    then detect_lost_fields env t e;
-    t
+  | ObjE (exp_bases, exp_fields), T.Obj(T.Object, fts) ->
+    let t' = infer_check_bases_fields env fts exp.at exp_bases exp_fields in
+    let fts' = match T.promote t' with
+      | T.Obj(T.Object, fts') -> fts'
+      | _ -> []
+    in
+    let missing_field_labs =
+      List.filter (fun ft -> not (List.exists (fun ft' -> ft.T.lab = ft'.T.lab) fts')) fts
```
crusso: I wonder if this should be distinguishing type fields from value fields. They can have the same name. Maybe this was a lurking bug?

src/mo_frontend/typing.ml
```diff
@@ -2411,22 +2414,24 @@ and check_exp' env0 t exp : T.typ =
   | TupE exps, T.Tup ts when List.length exps = List.length ts ->
     List.iter2 (check_exp env) ts exps;
     t
-  | ObjE ([], exp_fields) as e, T.Obj(T.Object, fts) -> (* TODO: infer bases? Default does a decent job. *)
-    check_ids env "object" "field"
```
crusso, line 2415: Has this check (check_ids) been removed? I think it checks the field names are distinct and needs to be done somewhere.

Kamirus: It is still there, in the `infer_check_bases_fields`

src/mo_frontend/typing.ml
Kamirus: Yes, another limitation of this extra check is that it is not checking whether the type matches.
Anyway, if this check does not catch the error, the `check_inferred` will. This is just an extra check.
But yeah I'll change to only val fields

src/mo_frontend/typing.ml
Kamirus: Thanks! `lookup_val_field_opt` is there, but it returns the type instead of the full field, but I've moved the `find_val_field_opt` next to that

src/mo_frontend/typing.ml
ggreif: So basically the whole infrastructure was there already, just had to be glued in?

src/mo_frontend/typing.ml
```diff
@@ -2411,22 +2411,24 @@ and check_exp' env0 t exp : T.typ =
   | TupE exps, T.Tup ts when List.length exps = List.length ts ->
     List.iter2 (check_exp env) ts exps;
     t
-  | ObjE ([], exp_fields) as e, T.Obj(T.Object, fts) -> (* TODO: infer bases? Default does a decent job. *)
```
ggreif, line 2414: Yeah, I remember punting on this back then. Good that Caffeine made an opening!

test/run/record-update-inference.mo
```diff
@@ -0,0 +1,62 @@
+// Issue #418: type inference on record update should be as strong as on record creation
```
ggreif: Did he migrate the issue? Please correct the number!

ggreif (approved):
Looks solid to me. I think this can go in as soon as @crusso's comments are addressed!

Thanks for filling in the gaping hole I left behind!

Issue #5591, opened 2025-10-22
dedup() does not exist when compiling module


alexandru-uta:
The program 

```
import Random "mo:core/Random";

module {
};
```

compiled with:
`moc test.mo  --package core /path/to/core/src/`

produces:

`Fatal error: exception File "codegen/compile_enhanced.ml", line 808, characters 11-17: Assertion failed
Raised at Codegen__Compile_enhanced.E.get_dedup in file "codegen/compile_enhanced.ml", line 808, characters 11-23
Called from Codegen__Compile_enhanced.Internals.dedup2 in file "codegen/compile_enhanced.ml", line 6922, characters 33-50
Called from Codegen__Compile_enhanced.Serialization.deserialize_go.(fun) in file "codegen/compile_enhanced.ml", line 8405, characters 10-30
Called from Codegen__Compile_enhanced.Func.of_body in file "codegen/compile_enhanced.ml", line 1024, characters 6-18
Called from CamlinternalLazy.force_lazy_block in file "camlinternalLazy.ml", line 31, characters 17-27
Re-raised at CamlinternalLazy.force_lazy_block in file "camlinternalLazy.ml", line 36, characters 4-11
Called from Lib.AllocOnUse.use in file "lib/lib.ml", line 633, characters 13-28
Called from Codegen__Compile_enhanced.E.built_in in file "codegen/compile_enhanced.ml" (inlined), line 630, characters 4-49
Called from Codegen__Compile_enhanced.Func.share_code in file "codegen/compile_enhanced.ml", line 1053, characters 20-41
Called from Codegen__Compile_enhanced.Serialization.deserialize_from_blob.(fun) in file "codegen/compile_enhanced.ml", line 8849, characters 16-36
Called from Codegen__Compile_enhanced.Stack.with_frame in file "codegen/compile_enhanced.ml", line 1593, characters 4-8
Called from Codegen__Compile_enhanced.Serialization.deserialize_from_blob.(fun) in file "codegen/compile_enhanced.ml", line 8839, characters 14-514
Called from Stdlib__List.map in file "list.ml", line 92, characters 20-23
Called from Codegen__InstrList.concat_map in file "codegen/instrList.ml" (inlined), line 187, characters 43-58
Called from Codegen__Compile_enhanced.Serialization.deserialize_from_blob.(fun) in file "codegen/compile_enhanced.ml", line 8821, characters 8-1023
[...]

PR #5629, opened 2025-11-04
fix: fix issue #5591 (missing `dedup` logic for actor-less code)

crusso:
Code that requires deserialization of blobs but is not part of a main actor would fail to register the necessary dedup function.

Builds on #5627

* avoids risk of not deduping depending on code gen order
* simplifies logic
- [x] TODO: rewrite test so it doesn't depend on core and runs in CI.


test/run-drun/issue-5591.mo
```diff
@@ -0,0 +1,8 @@
+//ENHANCED-ORTHOGONAL-PERSISTENCE-ONLY
+//MOC-FLAG --package core $MOTOKO_CORE
```
crusso: This probably won't work in CI...

PR #5630, opened 2025-11-05
feat: Extra error before inference when incompatible return type

Kamirus:
Instead of
```
cannot apply function of type
  (self : M, T) -> Bool
to argument of type
  (M, Nat)
  {a : Float; var b : Text}
```
Raise the following error that is the same when inference is not necessary
```
expression of type
  Bool
cannot produce expected type
  {a : Float; var b : Text}
```

### DIFF
--- a/src/mo_types/type.ml
+++ b/src/mo_types/type.ml
@@ -725,10 +725,14 @@ let is_immutable_obj obj_type =
   let _, fields = as_obj_sub [] obj_type in
   List.for_all (fun f -> not (is_mut f.typ)) fields
 
-
-let lookup_val_field_opt l tfs =
+let find_val_field_opt l tfs =
   let is_lab = function {typ = Typ _; _} -> false | {lab; _} -> lab = l in
   match List.find_opt is_lab tfs with
+  | Some tf -> Some tf
+  | None -> None
+
+let lookup_val_field_opt l tfs =
+  match find_val_field_opt l tfs with
   | Some tf -> Some tf.typ
   | None -> None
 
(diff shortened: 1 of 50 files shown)


src/mo_types/type.ml
```diff
@@ -572,6 +572,26 @@ let open_binds tbs =
   List.iter2 set_kind cs ks;
   ts
 
+let rec is_closed = function
```
crusso, line 575: This definition of is_closed isn't correct (the function case should be checking the bounds and allowing the bound vars in the bounds, dom and range and I think the async case is wrong too.

Might be better to add an index to check the debruijn  depth of Vars (perhaps we already have a predicate for that).

src/mo_types/type.ml
```diff
@@ -572,6 +572,26 @@ let open_binds tbs =
...
+  | Non
+  | Pre
+  | Typ _
+  | Prim _ -> true
+  | Array t
+  | Opt t
+  | Mut t
+  | Named (_, t)
+  | Weak t -> is_closed t
+  | Con (_, ts)
```
crusso, line 587: It would be safe to check the type function in a Con case, in case we every allow those to be open (we don't at the moment).

src/mo_types/type.ml
Kamirus: Oh yes, I forgot to make it right, this was just enough for experimenting. 
Do we want this PR anyway? Now that we have specific errors for return and arguments in bimatch it might not be that useful

crusso: Up to you. Maybe merge with master and then we can compare?

crusso: It might well still be an improvement...but a special case...

PR #5632, opened 2025-11-05
feat: hide syntax error recovery behind flag (default false)

crusso:
Multiple syntax errors confuses AI. Let's hide that option behind a flag:

c.f. https://github.com/dfinity/motoko/pull/5589


Kamirus (approved):
Lgtm

PR #5634, opened 2025-11-07
feat: Improve context dot instantiation used for error reporting, solve only receiver type variables

Kamirus:
- Fixes the documentation in `bi_match_subs` of the `must_solve` parameter.
- Solves only the receiver type variables, remaining stay unsolved `T := T`, not solved to `Any` or `Non`
- Fixes the type instantiation hint to have the correct arity

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,8 @@
 
   * Fixed the type instantiation hint to have the correct arity (#5634).
 
+  * Fix for #5618 (compiling dotted `await`s) (#5622).
+
   * Improved type inference of the record update syntax (#5625).
 
   * New flag `--error-recovery` to enable reporting of multiple syntax errors (#5632).
--- a/src/mo_def/syntax.ml
+++ b/src/mo_def/syntax.ml
@@ -391,11 +391,16 @@ let contextual_dot_args e1 e2 dot_note =
   let arity = match dot_note.note.note_typ with
     | T.Func(_, _, _, args, _) -> List.length args
     | _ -> raise (Invalid_argument "non-function type in contextual dot note") in
+  let effect eff =
+    match (e1.note.note_eff, eff) with
+    | T.Triv, T.Triv -> T.Triv
+    | _, _ -> T.Await
+  in
   let args = match e2 with
     | { it = TupE []; at; note = { note_eff;_ } } ->
-       { it = e1.it; at; note = { note_eff; note_typ = e1.note.note_typ } }
+       { it = e1.it; at; note = { note_eff = effect note_eff; note_typ = e1.note.note_typ } }
     | { it = TupE exps; at; note = { note_eff; note_typ = T.Tup ts } } when arity <> 2 ->
-       { it = TupE (e1::exps); at; note = { note_eff; note_typ = T.Tup (e1.note.note_typ::ts) } }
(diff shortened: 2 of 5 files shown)


src/mo_frontend/bi_match.mli
```diff
@@ -65,7 +65,9 @@ type ctx
   - [tbs] is the list of type parameters to instantiate
   - [ret_opt] is the optional return type mentioning tbs determining polarities
   - [subs] is the list of sub-type problems mentioning tbs either on left or right, but never both sides
-  - [deferred] is the list of types deferred until the next round; types containing tbs that are part of the subtype problems in the 2nd round
+  - [must_solve] is the list of types that must be solved in the 1st round.
+    If empty, all type variables in [tbs] will be solved in the 1st round.
+    Otherwise, only the type variables in [must_solve] will be solved in the 1st round.
```
crusso: must_solve is a list of typs, not type parameters? Is the interpretation that all variables in tbs that appear free in `must_solve` must be solved?

test/fail/ok/inf-error-ctx-dot-partial-instantiation.tc.ok
```diff
@@ -0,0 +1,13 @@
...
+  Text
+inf-error-ctx-dot-partial-instantiation.mo:15.23-15.47: type error [M0098], cannot apply function `.map` of type
+  (project : (Nat, Text) -> V2) -> Map<Nat, V2>
+to argument of type
+  {map : [(Nat, [var Text])]}
+inf-error-ctx-dot-partial-instantiation.mo:16.23-16.56: type error [M0098], cannot apply function `.map` of type
+  (project : (Nat, Text) -> V2) -> Map<Nat, V2>
+to argument of type
+  (project : (Nat, Text) -> V2)
+Hint: Add explicit type instantiation, e.g. <Nat>
```
crusso:
```suggestion
Hint: Add explicit type instantiation, e.g. <Nat,Text, V2>
```
map takes three type parameters, so this Hint is wrong I think.

crusso (commented):
I think the instantiation itself looks better but the hint is wrong (and was wrong before)

src/mo_frontend/bi_match.ml
```diff
@@ -384,12 +387,13 @@ let is_closed ctx t = if is_ctx_empty ctx then true else
...
 (** Raises when [er] is non-empty, optionally with a suggested type instantiation. *)
-let maybe_raise_underconstrained ctx env er =
+let maybe_raise_underconstrained ctx current_env unsolved er =
   let error_msg = ErrorUnderconstrained.to_string er in
   if error_msg = "" then () else
   let error_msg, hint =
-    let ts = List.map (fun c -> ConEnv.find c env) ctx.var_list in
-    if List.for_all (is_closed ctx) ts then
+    if ConSet.is_empty unsolved then
+      (* Future work: fill the unsolved and solved from previous rounds with holes, e.g. <_, Nat, _> *)
```
Kamirus, line 395: https://github.com/dfinity/motoko/pull/5575 might be useful for 
1. Skipping solved tvars in the previous round (not necessary to provide)
2. Skipping unsolved tvars, so that we can always provide the hint, but then such a hint might not provide enough type info, we could continue solving...

src/mo_frontend/bi_match.mli
Kamirus: Yes, I've updated the doc comment, maybe now it is clearer

PR #5635, opened 2025-11-07
experiment: multi-migration expressions

alexandru-uta:
Resolves LANG-1126

Adds a new type of migration expression `(with initializer = (m1, m2, m3, ..., mN))`.

**High-level Design Decisions**
* actors don't allow initializers on variables, e.g., declarations like `let x : Nat = 5` are forbidden in actor bodies.
* all actors have a migration function (now called initializer), even those deployed from scratch. All variables get initialized only in initializer functions. Newly deployed actors will have to initialize their stable variables via an initializer function.
* initializers can be a single function (which is migration from v(i) to v(i+1)), or a tuple of functions (which is migration from v(i), v(i+1), …, v(k)); all these functions should compose correctly, otherwise the compiler will reject the migration
* all initializers/migration functions will be identified by the file in which they are defined — a single initializer per file
* the compiler will not allow 2 identical migrations (i.e., same file) in the same chain — data loss possibility if we allow it
* the RTS will not allow 2 identical migrations to be run, the RTS will error at runtime if we try to apply the same migration twice — data loss possibility if we allow it

The **restrictions** for the multi-migration are the following:
1. Functions `mN(mN-1(...(m2(m1))..)` should be able to compose; The program fails to compile if not.
2. Input of m1 should be compatible with the initial program.
3. Output of mN should be compatible with the target program.
4. All stable fields need to be named explicitly in the input/output of migration functions. If a migration function input does not include all the output fields of the previous one, the migration will be rejected. Conversely, if a function does not specify a field in its output, but that field existed in the input, this means that the field will be explicitly dropped.
5. he multi-migration SHOULD NOT be combined with the regular migration!

**Usage**
The migrations are intended to be applied 1-by-1 in normal operation, when iterating across versions of the program, such as: `(with initializer = (m1))`. Then, at the end of an iteration chain, one would be able to simply fast-forward to the last program version by skipping all previous migrations and adding `(with initializer = (m1, m2, m3, m4))`.

**Example**

```
// Actor version v0, needs to initialize variables.
(with initializer = Init.new)
actor {
  let x : Nat;
  let y : Map.Map<X, Y>;
  var z : [var U];
}

// Actor version v1, adds migration function from previous version
(with initializer = Init.migration1)
actor {
  let x : Nat;
  let y : Map.Map<X, Y>;
  var z : [var U];
  // New field.
  let t : List.List<T>;
}

// Multi-migration
(with initializer = (Init.new, Init.migration1))
actor {
  let x : Nat;
  let y : Map.Map<X, Y>;
  var z : [var U];
  let t : List.List<T>;
}

// ------------------------- 

module {
  // The function outputs the initialized actor fields.
  public func new(_ : {}) : ActorType {
    {
      x = 0;
      y = Map.empty();
      var z = [var 1, 2, 3];
    };
  };

  public func migration1(old : ActorType) : ActorType2 {
    {
      old with
      // check this exists!
      y = List.empty();
    };
  };
};
```

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,9 @@
 * motoko (`moc`)
 
   * Allow `break` and `continue` in loops without labels (#5702).
+  * Report a better error for labeled `continue` targeting a non-loop (#5800).
+  * Deprecate older garbage collectors: generational, copying and compating GCs (#5806).
+  * Fix contextual dot type note, this should fix the hover hint in the vscode extension, showing the correct function type instead of `()` (#5809).
 
 ## 1.1.0 (2026-01-16)
 
--- a/doc/docusaurus/package-lock.json
+++ b/doc/docusaurus/package-lock.json
@@ -10857,9 +10857,9 @@
       }
     },
     "node_modules/lodash": {
-      "version": "4.17.21",
-      "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.21.tgz",
-      "integrity": "sha512-v2kDEe57lecTulaDIuNTPy3Ry4gLGJ6Z1O3vE1krgXZNrsQ+LFTGHVxVjcXPs17LhbZVGedAJv8XZ1tvj5FvSg=="
+      "version": "4.17.23",
+      "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.23.tgz",
+      "integrity": "sha512-LgVTMpQtIopCi79SJeDiP0TfWi5CNEc/L/aRdTh3yIvmZXTnheWpKjSZhnvMl8iXbC1tFg9gdHHDMLoV7CnG+w=="
     },
     "node_modules/lodash.debounce": {
       "version": "4.0.8",
@@ -25511,9 +25511,9 @@
       }
     },
     "lodash": {
-      "version": "4.17.21",
-      "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.21.tgz",
-      "integrity": "sha512-v2kDEe57lecTulaDIuNTPy3Ry4gLGJ6Z1O3vE1krgXZNrsQ+LFTGHVxVjcXPs17LhbZVGedAJv8XZ1tvj5FvSg=="
+      "version": "4.17.23",
+      "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.23.tgz",
+      "integrity": "sha512-LgVTMpQtIopCi79SJeDiP0TfWi5CNEc/L/aRdTh3yIvmZXTnheWpKjSZhnvMl8iXbC1tFg9gdHHDMLoV7CnG+w=="
     },
     "lodash.debounce": {
       "version": "4.0.8",
--- a/flake.lock
+++ b/flake.lock
@@ -71,11 +71,11 @@
     "ic-wasm-src": {
       "flake": false,
       "locked": {
-        "lastModified": 1768251536,
-        "narHash": "sha256-oilXW9k2IFDNXtxoIt9ZYvX5i75Ba55Uq9jLwCcDqW8=",
+        "lastModified": 1768961826,
+        "narHash": "sha256-wiKiVb5ujZbFDVe0mpiAI/kDI6EO269lWXTQExhM3v4=",
         "owner": "dfinity",
         "repo": "ic-wasm",
-        "rev": "a45d53291df5a05fc0cc509abfe22a2cbb5bba4d",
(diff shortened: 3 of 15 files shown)


src/lowering/desugar.ml
```diff
@@ -632,10 +632,58 @@ and build_actor at ts (exp_opt : Ir.exp option) self_id es obj_typ =
...
+              let result = callE e_migration_i [] arg_expr in
+              build_chain (i + 1) result
+          in
+          let body = blockE [letD tuple_var e_tuple] (build_chain 0 (varE input_var)) in
+          let e_composed = funcE "migration_chain" T.Local T.Returns [] [input_arg] [last_rng] body in
+          
+          T.multi_migration_lab, composed_typ, e_composed
+           | _ -> assert false)  (* typing.ml validates this is Func or Tup *)
+        | Some _, Some _ -> assert false  (* impossible: typing.ml rejects both *)
+        | None, None -> assert false      (* impossible: typing.ml requires one *)
```
crusso: Does it? You can't just have a` (with other = ...) actor ...` etc?

src/lang_utils/error_codes.ml
```diff
@@ -202,6 +202,9 @@ let error_codes : (string * string option) list =
     "M0228", None; (* Only top-level mixins *)
     "M0229", None; (* Non-var pattern for mixin import *)
     "M0230", None; (* Cannot determine implicit argument *)
+    "M0240", None; (* Migration chain broken - type mismatch *)
```
crusso: move down

src/lowering/desugar.ml
alexandru-uta: bad assumption I guess. Will fix

crusso: I'm actually not sure what we enforce - it was all implemented in a hurry for caffeine

alexandru-uta:
> I guess we could also avoid introduce another field name by just allowing the old `migration` field to contain a function or tuple of functions. Indeed, now I'm wondering why we can't just use a functional composition of migration functions.
> 
> ```
> (with migrate = (Migration1.run).then(Migration2.run2).then(Migration3.run))
> ```
> 
> where
> 
> `func then<A,B,C>(f : A-> B, g : B->C) : A->C = func (a:A) { g(f(a))};`
> 
> (I guess this wouldn't enforce that you are not losing data, buy maybe one could ensure that with a subtype bound or special bound that enforces something even stricter: `func then<A <: {},B <: A, C<: B>(f : A-> B, g : B->C) : A->C = func (a:A) { g(f(a))};`

I wanted to keep them separate so that we don't have people/AIs mix the two;

Re: the `.then()` -- does that work fine with hundreds of funcs?

rts/motoko-rts/src/persistence.rs
```diff
@@ -63,6 +63,10 @@ struct PersistentMetadata {
     /// We keep a pointer to this here so that we can keep if alive across upgrades.
     /// To keep the dedup table live, we need to add it to roots as well.
     dedup_table: Value,
+    /// Migration function hashes array. This stores an array of the the hashes (strings) of the migration functions.
```
crusso, line 66:
```suggestion
    /// Migration function hashes array. This stores an array of the hashes (strings) of the migration functions.
```

rts/motoko-rts/src/persistence.rs
```diff
@@ -63,6 +63,10 @@ struct PersistentMetadata {
...
+    /// This is for the purpose of multi-migration tracking such that a single migration function
+    /// cannot be executed multiple times, removing the risk of data loss.
```
crusso, line 68:
```suggestion
    /// cannot be executed multiple times, reducing the risk of data loss.
```

src/lowering/desugar.ml
```diff
@@ -637,10 +637,143 @@ and build_actor at ts (exp_opt : Ir.exp option) self_id es obj_typ =
...
+              let check_and_trap = ifE was_performed 
+                (primE (I.OtherPrim "trap") [error_msg])
+                (unitE())
+              in
+              
+              (* Extract migration i from tuple *)
+              let e_migration_i = projE (varE tuple_var) i in
+              
+              (* Execute the migration *)
+              let result = callE e_migration_i [] arg_expr in
```
crusso: Note to self: Since the record of stable variables can have different per-field mutability, we should probably at this point not pass arg_expr directly, but a new record constructed from the field values of arg_expr, but possibly differing in the mutability of fields.  This is because you can go from `var fld : T` to `fld : T` and vice versa on upgrades, eventhough this is verboten with vanilla subtyping.

PR #5636, opened 2025-11-07
feat: Further improve bimatch errors: solve bounds and make over-constrained errors redundant

Kamirus:
- On bimatch error: solve the bounds to provide a more concrete error
  This helps especially with context dot which solves the receiver first.
- This error substitution ignores variables solved to Any/Non as they are either unsolved yet or most likely a sign of error
- Reorders bimatch constraints to : receiver, expected return type and arguments starting from the first one
- Catches over-constrained errors early to pin point them to the correct argument or return type

Implementation details:
- the `env` in `bimatch.ml` does not contain unsolved type variables now.
  So that we can ConEnv.disjoint_union the previous solution with the new one.

### DIFF
--- a/src/mo_frontend/bi_match.ml
+++ b/src/mo_frontend/bi_match.ml
@@ -182,18 +182,12 @@ end = struct
 end
 
 
-(* TODO: is this dead code now? considering what the `check` is doing? *)
-let over_constrained_exn lb c ub =
-  if debug then
-    bimatch (Format.asprintf
-      "implicit instantiation of type parameter `%s` is over-constrained with%a\nwhere%a\nso that no valid instantiation exists"
-      (Cons.name c)
-      display_constraint (lb, c, ub)
-      display_rel (lb, "</:", ub))
-  else
-    bimatch (Format.asprintf
-      "there is no consistent choice for type parameter `%s`."
-      (Cons.name c))
+let impossible_over_constrained lb c ub =
+  error (Format.asprintf
+    "bug: impossible over-constrained type parameter `%s` with%a\nwhere%a\nPlease report this bug and supply an explicit instantiation instead."
+    (Cons.name c)
+    display_constraint (lb, c, ub)
+    display_rel (lb, "</:", ub))
 
 let choose_under_constrained ctx er lb c ub =
   match ConEnv.find c ctx.variances with
@@ -459,9 +453,8 @@ let solve_bounds on_error ctx to_defer l u =
         None
       end else
         Some (choose_under_constrained ctx er lb c ub)
-    else (
-      on_error (over_constrained_exn lb c ub);
-      None)
+    else
+      impossible_over_constrained lb c ub
   ) |> ConEnv.filter_map (fun c o -> o) in
   (* Join the previous solution with the new one *)
   let env = ConEnv.disjoint_union ctx.current_env env in
@@ -531,7 +524,6 @@ let solve ctx (ts1, ts2) must_solve =
       1. Unconstrained variables are solved to Any/Non, don't substitute them, the error was not there.
       2. Matching unrelated types, e.g. Nat and Text, is permitted (resulting in Any/Non bound),
         However, it usually indicates an error. Don't solve these variables.
-        These potential extra errors are attached when 
     *)
     let env = env |> ConEnv.filter (fun c t -> not (eq Any t || eq Non t)) in
     let pretty_sub (t1,t2) =


test/fail/ok/pretty-inference.tc.ok
```diff
@@ -4,36 +4,43 @@ to argument of type
   Nat
 to produce result of expected type
   ()
-because there is no consistent choice for type parameter `T`.
+because there is no way to satisfy subtyping
```
crusso, line 5: Note these errors are all longer that just saying T has no solution.

crusso: We could also just say `there is no valid choice for `T`` if solution is too fancy.

Kamirus: yes, these errors in this file are worse

Kamirus: These cases, where the error is in the bound of a variable, e.g. here `T <: ()` are tricky.
`T` cannot be bi-matched with `Nat` because there `()` and `Nat` are unrelated (over-constrained error).
The bounds are `Non <: T <: ()` and `Nat <: T` cannot be applied.
Solving `T` would yield `T := Non` because `T` is currently bivariant (maybe for error reporting we would need full variance info from the whole function type not just the return type...)
These cases are rare, I'd ignore them for now. There is enough info to figure out why there is an error

test/fail/ok/inference.tc.ok
```diff
@@ -37,12 +38,14 @@ inference.mo:111.8-111.17: type error [M0098], cannot apply function of type
...
-because there is no consistent choice for type parameter `T`.
+because there is no way to satisfy subtyping
+  Bool  <:  T  (for argument `x`) 
 inference.mo:112.1-112.10: type error [M0098], cannot apply function of type
   <T <: Int>(x : T) -> T
 to argument of type
   Bool
-because there is no consistent choice for type parameter `T`.
+because there is no way to satisfy subtyping
+  Bool  <:  None  (for argument `x`)
```
Kamirus: Why `None`?
- Expected return type is `()`
- Initial `bounds: None <: T <: Int`
- Initial `subs: T <: (@ret : ()), Bool <: (x : T)`
- So first the upper bound `()` is joined with `Int` resulting in `None`
- Current `bounds: None <: T <: None`, which is **not an error** 🤡

Kamirus: Fixed it by skipping all variables solved to Non/Any from the error substitution

src/mo_frontend/bi_match.ml
```diff
@@ -197,6 +215,23 @@ let choose_under_constrained ctx er lb c ub =
...
+    None
+  else
+    Some (l, u)
+
+let update binop c t ce =
+  let current = ConEnv.find c ce in
+  let updated = binop ?src_fields:None t current in
+  if debug then
+    Debug.print_update_bound c current t updated;
+  (* Future work: consider an error when joining two unrelated types, e.g. [lub Nat Text = Any], would be a breaking change *)
```
Kamirus, line 226: I've tried it and it adds extra details, but it requires adding an "error context" to the bi_match and then filtering out some of them to avoid duplicates and it only helps in rare cases.
Future work

src/mo_frontend/typing.ml
```diff
@@ -2935,10 +2935,13 @@ and infer_call_instantiation env t1 ctx_dot tbs t_arg t_ret exp2 at t_expect_opt
 
   (* Incorporate the return type into the subtyping constraints *)
   let ret_typ_opt, subs =
+    let subs = List.rev subs in
     match t_expect_opt with
     | None -> Some t_ret, subs
     | Some expected_ret -> None, (t_ret, Bi_match.name_ret_typ expected_ret) :: subs
   in
+  (* Make sure the order of constraints is: receiver, expected return type and arguments starting from the first one *)
+  let subs = extra_subtype_problems @ subs in
```
Kamirus, line 2944: TODO:
add a test that relies on that order, or check maybe it is already there

src/mo_frontend/bi_match.ml
```diff
@@ -172,10 +182,18 @@ end = struct
...
-           "there is no consistent choice for type parameter `%s`."
-           (Cons.name c))
+(* TODO: is this dead code now? considering what the `check` is doing? *)
+let over_constrained_exn lb c ub =
+  if debug then
+    bimatch (Format.asprintf
+      "implicit instantiation of type parameter `%s` is over-constrained with%a\nwhere%a\nso that no valid instantiation exists"
+      (Cons.name c)
+      display_constraint (lb, c, ub)
+      display_rel (lb, "</:", ub))
```
crusso: Do we still want this for debugging (the `then` branch)?

src/mo_frontend/bi_match.ml
```diff
@@ -398,20 +429,45 @@ let is_closed ctx t = if is_ctx_empty ctx then true else
...
   in
-  error error_msg ~hint
+  Some (bimatch error_msg ~hint)
+
+let solve_bounds on_error ctx to_defer l u =
+  if debug then Debug.print_solved_bounds l u;
+  let unsolved = ref ConSet.empty in
+  let er = ErrorUnderconstrained.empty () in
+  let env = l |> ConEnv.mapi (fun c lb ->
+    let ub = ConEnv.find c u in
```
crusso, line 446: Is this `find c u` guaranteed to succeed? Do we always declare both a lower and upper bound? Probably yes, but not I don't recall.

src/mo_frontend/bi_match.ml
```diff
@@ -489,8 +525,18 @@ let solve ctx (ts1, ts2) must_solve =
...
     end
-  | Error (inst, (t1, t2)) ->
-    let pretty_sub (t1, t2) =
+  | Error ((l, u), (t1, t2)) ->
+    let env, _unsolved = solve_bounds ignore ctx ConSet.empty l u in
+    (* Preprocess the substitution for better error messages. Drop variables solved to Any/Non.
+      1. Unconstrained variables are solved to Any/Non, don't substitute them, the error was not there.
+      2. Matching unrelated types, e.g. Nat and Text, is permitted (resulting in Any/Non bound),
+        However, it usually indicates an error. Don't solve these variables.
+        These potential extra errors are attached when
```
crusso: "These potential extra errors are attached when ???"
(broken comment)

Kamirus: TODO: remove this line

src/mo_frontend/bi_match.ml
Kamirus: yes

src/mo_frontend/bi_match.ml
Kamirus: It's dead code, I've replaced it with a bug-exception just like the bad instantiation.

PR #5637, opened 2025-11-07
[VSC-31] Remove intermediate conversion to S-exps

crusso:
Problem: The current process for creating a JSON representation of the AST requires an intermediate conversion to S-expression. This hurts performance and makes it impossible to annotate all nodes with types.

Solution: Implement a library for converting the original AST directly to JSON, eliminating the intermediate step.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -71,11 +71,11 @@
     "ic-src": {
       "flake": false,
       "locked": {
-        "lastModified": 1755350517,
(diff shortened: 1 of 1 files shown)


src/pipeline/test_field_srcs.ml
```diff
@@ -149,10 +149,6 @@ let%expect_test "" =
   [%expect {|
     Ok:
     Collected sources:
```
crusso, line 151: I don't really know why this has changed

PR #5638, opened 2025-11-07
chore: (review comment) simplify `mo_idl/idl_to_mo_value.ml`

ggreif:
This is an amendment to #5619, simplifying. It took me some time to realise that we can enrich record values with defaulted fields unconditionally.


src/mo_idl/idl_to_mo_value.ml
crusso, line 66: Why is there no insertion of nulls here too?

crusso, line 70: And possible in the Tup case too?

PR #5643, opened 2025-11-12
feat: provide an explanation when subtyping fails

crusso:
* adapts the machinery used to explain stable subtyping failure to ordinary subtyping, producing better error messages when subtyping fails.

Builds on #5409 

fixes LANG-37

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -121,11 +121,11 @@
     "motoko-core-src": {
       "flake": false,
       "locked": {
-        "lastModified": 1765968049,
-        "narHash": "sha256-lO8kPkUSQYqEd2pGPATlBIfQDsWZ65nif1hOYN3WGdo=",
+        "lastModified": 1767368472,
+        "narHash": "sha256-JP8kVRWEuM4cKAUpXnwWcQe2C0vT/54YYZwQs569q3E=",
         "owner": "caffeinelabs",
         "repo": "motoko-core",
-        "rev": "997746c93ae569fbf04e7b437ae1ae4fb13a59df",
+        "rev": "c87170edd5766dc9fa243d95a5ef197849f7956f",
         "type": "github"
       },
       "original": {


test/fail/ok/typ-field-lub-glb-fail.tc.ok
```diff
@@ -2,10 +2,14 @@ typ-field-lub-glb-fail.mo:14.31-14.35: type error [M0096], expression of type
   Cx = {x : Nat}
 cannot produce expected type
   {type T = Int; x : Nat}
+because: Unsupported additional field `T` in type 
+  {type T = Int; x : Nat}
```
Kamirus: @crusso do you think this message is helpful? Since encountering [LANG-45](https://linear.app/caffeinelabs/issue/LANG-45/unsupported-additional-field-when-missing-in-both-oldactor-and) I started to doubt whether `Unsupported additional field` is the right thing to say.

Given this message when coding I feel like I should remove this 'unsupported additional field' to fix the error, but that is not the case here.
But this is something to improve in another PR, just asking what do you think

src/mo_types/type.ml
```diff
@@ -2299,55 +2318,104 @@ let string_of_typ_expand typ : string =
...
+        |  _ -> true)
+    | _ -> true
+  in
+  let context = List.filter is_non_trivial_item context in
+  if context = [] then "" else
+  let emit_item item nested =
+    match item with
+    | (Field label) -> Printf.sprintf "`%s`" label, "in"
+    | (ConsType c) ->  Printf.sprintf "`%s`" (remove_hash_suffix (Cons.name c)), "in"
+    | (NamedType name)-> Printf.sprintf "`(%s : _)`" name, "in"
```
Kamirus:
```suggestion
    | Field label -> Printf.sprintf "`%s`" label, "in"
    | ConsType c ->  Printf.sprintf "`%s`" (remove_hash_suffix (Cons.name c)), "in"
    | NamedType name-> Printf.sprintf "`(%s : _)`" name, "in"
```

test/fail/ok/typ-field-lub-glb-fail.tc.ok
crusso: Yeah, I don't like it much either. That was just the original phrasing.
Maybe
Missing field `T` from type ...

Kamirus: maybe best to leave it for now as it, the model seems to understand what to do with it most of the time

crusso:
@Kamirus I took another pass, removed the weird error messages around unsupported fields and labels and adjusted the wording a bit.

test/fail/ok/typ-field-lub-glb-fail.tc.ok
crusso: Actually, I took the opportunity to fix it now. See what you thing.

src/mo_types/type.ml
```diff
@@ -2359,28 +2351,23 @@ let rec string_of_explanation explanation =
...
-  | UnexpectedTag (context, lab, t) ->
-    Format.asprintf "Unsupported additional tag `#%s` in type %a%s" lab display_typ t (string_of_context "of" context)
+    Format.asprintf "case `#%s` is missing from type %a%s" lab display_typ t (string_of_context "of" context)
   | MissingField (context, lab, t) ->
     let sort = if is_typ t then "type" else "field" in
-    Format.asprintf "Missing %s `%s` in type %a%s" sort lab display_typ t (string_of_context "of" context)
-  | UnexpectedField (context, lab, t) ->
-    let sort = if is_typ t then "type" else "field" in
-    Format.asprintf "Unsupported additional %s `%s` in type %a%s" sort lab display_typ t (string_of_context "of" context)
+    Format.asprintf "%s `%s` is missing from type %a%s" sort lab display_typ t (string_of_context "of" context)
```
Kamirus: This might be worse because now both messages are merged together.
Before at least we could know whether the field is missing from the actual or the expected.
Now you need to parse the whole type (which might be truncated not even showing the field).
As I said before, I'd leave it as it was because then at least we can control the RAG

test/cmp/ok/loss0-1.cmp.ok
```diff
@@ -2,16 +2,17 @@
  The previous type
   var {field : Nat}
  is not a stable subtype of
-  var {} because: Missing field `field` in type 
+  var {} because field `field` is missing from expected type
```
crusso: Add new line?

src/mo_types/type.ml
crusso: Ok, I think I've addressed this now with another bit on the error payload.

Kamirus (approved):
LGTM!

PR #5650, opened 2025-11-17
feat: Short bimatch errors arg/ret wrong where possible

Kamirus:
Prefer short errors like
```
inf-error.mo:46.45-46.51: type error [M0096], expression of type
  Text
cannot produce expected type
  Nat
```

Over long errors like
```
inf-error.mo:46.13-46.52: type error [M0098], cannot apply function of type
  <K, V>(self : Map<K, V>, compare : (implicit : (K, K) -> Order), n : K) ->
    ?V
to argument of type
  ({map : [(Nat, [var Text])]}, (n : Nat, m : Nat) -> Order, Text)
  Text  <:  Nat  (for argument `n`) 
```

Notes:
- Points the error to the wrong argument or the whole call when the return type is incompatible with the expected
- Simple errors only mention closed types, long error is displayed when there are unsolved type variables mentioned

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,10 @@
 
 * motoko (`moc`)
 
+  * Shorter, simpler error messages for generic functions (#5650).
+    The compiler now tries to point to the first problematic expression in the function call, rather than the entire function call with type inference details.
+    Simple errors only mention closed types; verbose errors with unsolved type variables are only shown when necessary.
+
   * Improved error messages for context dot: only the receiver type variables are solved, remaining type variables stay unsolved, not solved to `Any` or `Non` (#5634).
 
   * Fixed the type instantiation hint to have the correct arity (#5634).


test/fail/ok/inf-error-small.tc.ok
```diff
@@ -1,81 +1,62 @@
...
+inf-error-small.mo:48.13-48.45: type error [M0096], expression of type
+  ?Text
+cannot produce expected type
+  Text
+inf-error-small.mo:50.13-50.29: type error [M0096], expression of type
+  ?Text
+cannot produce expected type
+  Bool
+inf-error-small.mo:53.25-53.46: type error [M0230], Cannot determine implicit argument `compare` of type
+  (Any, Any) -> Order
```
Kamirus, line 29: bimatch succeeds with `K=Any`, which is dumb but valid ofc.
then the check_exp on the hole expression fails because there is no such implicit.

Kamirus: this is same on master
```
inf-error-small.mo:53.25-53.46: type error [M0230], Cannot determine implicit argument `compare` of type
  (Any, Any) -> Order
Hint: If you're trying to omit an implicit argument named `compare` you need to have a matching declaration named `compare` in scope.
```

test/fail/ok/pretty-inference.tc.ok
```diff
@@ -90,42 +90,45 @@ because there is no way to satisfy subtyping
    ((((Nat, Bool), (Nat, Bool)), ((Nat, Bool), (Nat, Bool))),
     (((Nat, Bool), (Nat, Bool)), ((Nat, Bool), (Nat, Bool)))))  <: 
     T  (for argument `x`) 
-pretty-inference.mo:42.1-42.7: type error [M0098], cannot apply function of type
-  <T, U>(x : T, u : U) -> Nat
+pretty-inference.mo:42.1-42.11: type error [M0098], cannot apply function of type
+  <T, U>(x : T, u : U, U -> ()) -> ()
```
Kamirus: I just added sth that would preserve the long error, because apparently this file tests pretty printing

src/mo_frontend/bi_match.ml
```diff
@@ -518,27 +522,32 @@ let solve ctx (ts1, ts2) must_solve =
...
-      | t2 ->
-        Format.asprintf "%a" display_rel (t1, "<:", t2)
+    let t1 = subst env t1 in
+    let t2 = subst env t2 in
+    let reason = match t2 with
+    | Named ("@ret", t2) ->
+      if is_closed ctx t1 then Some { actual = t1; expected = t2; at } else None
+    | Named (_, t2) ->
+      if is_closed ctx t2 then Some { actual = t1; expected = t2; at } else None
+    | t2 -> None
```
christoph-dfinity: Couldn't we treat this the same as the case above?

test/fail/inf-error-small.mo
```diff
@@ -50,6 +50,7 @@ persistent actor {
      ignore peopleMap.get(1) : Bool; // bad
   };
   func test2() {
+    func c0() { let _ = peopleMap.get("text") }; // wrong ret should be reported before wrong arg
```
Kamirus:
```suggestion
    func c0() { let _ = peopleMap.get("text") };
```

kritzcreek (approved):
I like it. When it kicks in the errors are more straight forward to fix.

src/mo_frontend/bi_match.ml
Kamirus: We could but we would need to have both `is_closed ctx t1 && is_closed ctx t2`.
That's a bit wasteful but safer so probably worth it :D I'll apply it tomorrow

christoph-dfinity: LGTM

PR #5652, opened 2025-11-17
LSP -- Track More Regions - Serokell Grant 2 Milestone 2

heitor-lassarote:
This PR changes the type checker code so that more regions are tracked. In particular, we missed `check_exp_field`. This code is necessary to make more test cases pass in the extension.

Related PRs:
* https://github.com/dfinity/node-motoko/pull/150
* https://github.com/dfinity/vscode-motoko/pull/408


crusso (approved):
Looks ok to me!

PR #5655, opened 2025-11-18
experiment: use prefix implicit keyword in patterns and types

crusso:
Builds on #5521, but allows prefix use of `implicit` in patterns and named type items (use in arguments and tuples)

- [x] merge parent with master
- [x] merge with parent
- [ ] adjust type pretty printer


crusso:
Closing in favour of #5659

PR #5659, opened 2025-11-19
feat: add `implicit` keyword and dedicated grammar

crusso:
Alternative to #5521 using `implicit` keyword to prefix argument declarations (rather than appearing in the type). 
Suggested by @christoph-dfinity.

Enforce restriction that `implicit` can only be used in argument positions, while preserving enough of the old syntax to have a non-breaking core.

* declare `implicit` as a keyword
* using dedicated argument patterns, with prefixed `implicit` keywords.
  
``` 
<implicit> ::= 
    'implicit'

<id_wild> ::=  
    '_'

<typ_item> ::= 
    <implicit> ':' <typ>
    <id> ':' <typ>
    <id_wild> ':' <typ>
    <implicit> <id> ':' <typ>
    <implicit> <id_wild> ':' <typ>

<pat_arg> ::= 
    <implicit> <id> ':' <typ>
    <implicit> <id> ':' <typ> '=' <pat>
    <implicit> <id_wild> ':' <typ> '=' <pat>
    ...

<pat_args> ::= 
    ...
    '(' <list(<pat_arg>, ',')> ')'

<func_pat> ::= 
    <id>? <typ_params_opt> <pat_args>
 ```
   
* allow `(implicit : typ)` named types, as before, to avoid breaking old clients of new core.

* rule out misplaced implicits in the type checker.  This turned out to be too difficult to do in the parser, though ugly to do in the checker.

Matches doc in #5674.

Example:

```motoko
module Array {
  public func toText<T>(as : [T], implicit toText : T -> Text) : Text {
     var t = "";
     for (a in as.vals()) {
       t := t # (toText(a));
     };
     t
  }
};
module Pair {
  public func toText<T,U>(
    implicit toText  :  T -> Text = toTextT,
    implicit toTextU : U -> Text = toTextU,
    p : (T, U))
    : Text {
      "(" # toTextT(p.0) # "," # toTextU(p.1) # ")"
    };
};

// types 
type F = <T, U>(implicit toText : T -> Text, implicit toText : U -> Text, p : (T, U)) -> Text;
```
[x] remove interim syntax
[x] rewrite tests
[x] add `implicit id : typ = pat` and `implicit  _ : typ = pat` (similar to field patterns) to support more syntaxtic manual override and wildcard pattern (without any new keyword).
[x] mo-doc works on core


test/fail/implicit-example.mo
```diff
@@ -0,0 +1,52 @@
+// test legacy syntax (using for back compat in core) works too
```
crusso:
```suggestion
// test legacy syntax (used for back compat of core) works too
```

christoph-dfinity (approved):
Needs a changelog entry I guess? Unless we just make one big one for all the features combined.

PR #5664, opened 2025-11-20
feat: allow specifying an extra 'implicit name' for value definitions

christoph-dfinity:
WIP


crusso:
Hmm, I think the change would be much less invasive if we just added a second class type constructor, ImplT of name * typ. and just extending the matching predicate appropriately. It would not be that different from MutT on fields (and presumable we wouldn't want both implicit and mutable anyway. Maybe one could even re-use the current ImplicitT, but not sure that's a good idea.

PR #5668, opened 2025-11-21
chore: fix `candid` flake `narHash`

ggreif:
After updating to `nix profile add nixpkgs/release-25.05#nix && nix profile upgrade nix`
```
$ nix --version
nix (Nix) 2.31.2
```

Looks like `nix (Nix) 2.28.5` and `nix-2.33.0pre20251121` both don't need this change, so this might be a bug in `2.31.2`.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -4,7 +4,7 @@
       "flake": false,
       "locked": {
         "lastModified": 1762175579,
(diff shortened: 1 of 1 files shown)


ggreif:
This was some fluke.
```
$ /nix/store/i73by4j33zs534f8f7p9gk9p9brnnv11-nix-2.31.2/bin/nix flake update candid-src
$ git diff
<nothing>
```
All is good.

PR #5669, opened 2025-11-22
chore: bump `pocket-ic`

ggreif:
Needed a bump to `rustc` stable. Now at 1.88.0.

Also removed an obsolete vendoring.

### DIFF
--- a/nix/pocket-ic.nix
+++ b/nix/pocket-ic.nix
@@ -42,29 +42,6 @@ pkgs: let
       [target.aarch64-apple-darwin]
       rustflags = [ "-C", "linker=c++" ]
       EOF
-
-      # Patch for pocket-ic-server to disable canister backtrace.
-      echo "Applying patch..."
-      patch -p1 << EOF
-      diff --git a/rs/pocket_ic_server/src/pocket_ic.rs b/rs/pocket_ic_server/src/pocket_ic.rs
-      --- a/rs/pocket_ic_server/src/pocket_ic.rs
-      +++ b/rs/pocket_ic_server/src/pocket_ic.rs
-      @@ -619,6 +619,14 @@ impl PocketIcSubnets {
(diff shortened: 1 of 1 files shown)


nix/pocket-ic.nix
```diff
@@ -42,29 +43,28 @@ pkgs: let
...
-      echo '+            .canister_backtrace = FlagStatus::Disabled;' >> pocket_ic_server.patch
-      echo '+        hypervisor_config' >> pocket_ic_server.patch
-      echo '+            .embedders_config' >> pocket_ic_server.patch
-      echo '+            .feature_flags' >> pocket_ic_server.patch
-      echo '+            .environment_variables = FlagStatus::Enabled;' >> pocket_ic_server.patch
-      echo '         let state_machine_config = StateMachineConfig::new(subnet_config, hypervisor_config);' >> pocket_ic_server.patch
-      echo '         StateMachineBuilder::new()' >> pocket_ic_server.patch
-      echo '             .with_runtime(runtime)' >> pocket_ic_server.patch
-
+      # Patch for pocket-ic-server to disable canister backtrace.
```
ggreif: Same patch as before, but without the intermediate file.

ggreif:
Works with Mac too: https://github.com/dfinity/motoko/actions/runs/19542656582

nix/pocket-ic.nix
alexandru-uta: I think we can safely drop this now. Could you try? 

I think I pushed the fix [here](https://github.com/dfinity/motoko/blob/master/test-runner/src/main.rs#L653) a while ago, but I forgot to take out the patch. Maybe a good time to clean up.

ggreif: Interesting. The patch didn't get rejected, though, so I never got suspicious.

PR #5678, opened 2025-11-24
experiment: track lowest upper bound and greatest lower bound potential errors

Kamirus:
Joining two unrelated types (using `glb` or `lub`) is not an immediate error, just results in `Non` or `Any`.
This can lead to loosing type information in errors, example: 
`Non <: T <: Any` starting point
`Nat <: T <: Any` `glb` with `Nat`
`Any <: T <: Any` `glb` with `Text`, we lost info about `Nat` and `Text`
Now, `lub` with `Float` fails with an over-constrained error, but at this point we only know that `T := Any` which is not helpful.

This PR adds tracking of 'failed' glb/lub by checking whether their result the the appropriate extreme and adds this extra error info to the error message.
Here it would say `'T' inferred to 'Nat' but incompatible with 'Text'`

However, this PR is not that helpful in common cases where our current error handling is enough.
So I'm not sure if the extra complexity is worth it.


test/fail/ok/inference.tc.ok
```diff
@@ -46,6 +46,7 @@ to argument of type
   Bool
 because there is no way to satisfy subtyping
   Bool  <:  T  (for argument `x`) 
+  `T` inferred to `Int` but incompatible with `()`
```
crusso, line 49: Where is the () coming from, the return type?

test/fail/ok/inference.tc.ok
```diff
@@ -76,6 +77,7 @@ to argument of type
   [var Nat]
 because there is no way to satisfy subtyping
   [var Nat]  <:  [var T]  (for argument `x`) 
+  `T` inferred to `()` but incompatible with `Nat`
```
crusso, line 80: Ditto

crusso:
Maybe not worth the complexity?

test/fail/ok/inference.tc.ok
Kamirus: the return type yes

Kamirus:

PR #5679, opened 2025-11-24
experiment: Show namespaces for Types

Kamirus:
Experiment to avoid error messages like
```
expression of type
  {_ : Map<Nat, Text>}
cannot produce expected type
  {_ : Map<Nat, Text>}
```
where different type aliases have the same name.

Adding a namespace from the module/import names gives:
```
expression of type
  {_ : M2.Map<Nat, Text>}
cannot produce expected type
  {_ : M1.Map<Nat, Text>}
```

where the source file is
```motoko
import M2 "mo:core/Map";

module M1 {
  public type Map<K, V> = { map : [(K, [var V])] };
};
```

### DIFF
--- a/src/mo_types/type.ml
+++ b/src/mo_types/type.ml
@@ -2240,9 +2240,9 @@ and pp_stab_sig ppf sig_ =
   let ds =
     let cs' = ConSet.filter (fun c ->
       match Cons.kind c with
-      | Def ([], Prim p) when Cons.name c = string_of_prim p -> false
-      | Def ([], Any) when Cons.name c = "Any" -> false
-      | Def ([], Non) when Cons.name c = "None" -> false
+      | Def ([], Prim p) when string_of_con c = string_of_prim p -> false
+      | Def ([], Any) when string_of_con c = "Any" -> false
+      | Def ([], Non) when string_of_con c = "None" -> false
       | Def _ -> true
       | Abs _ -> false) cs in
     ConSet.elements cs' in
--- a/test/run-drun/ok/composite-query.tc.ok
+++ b/test/run-drun/ok/composite-query.tc.ok
@@ -1,8 +1,8 @@
 composite-query.mo:31.11-31.13: warning [M0145], this pattern of type
-  ?Class
+  ?Class.Class
 does not cover value
   null
 composite-query.mo:36.11-36.13: warning [M0145], this pattern of type
-  ?Class
+  ?Class.Class
 does not cover value
   null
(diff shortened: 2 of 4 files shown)


test/fail/ok/type-alias-path.tc.ok
```diff
@@ -0,0 +1,20 @@
...
+  {m2 : M1.Map<Nat, Text>}
+type-alias-path.mo:21.5-21.53: type error [M0098], cannot apply function of type
+  <T>(t : T, T -> ()) -> ()
+to argument of type
+  (M1.Map<Nat, Text>, (m : MyMap.Map<Nat, Text>) -> ())
+because there is no way to satisfy subtyping
+  (m : MyMap.Map<Nat, Text>) -> ()  <:  M1.Map<Nat, Text> -> ()
+type-alias-path.mo:24.24-24.33: type error [M0096], expression of type
+  <K, V>(self : MyMap.Map<K, V>,
+         compare : (implicit : (K, K) -> Types.Order), key : K) ->
```
Kamirus, line 13: Note that `Types` is not imported in the source file.
So this can still be confusing because the name `Types` is gathered transitively from the imports.

test/fail/ok/type-alias-path.tc.ok
```diff
@@ -0,0 +1,20 @@
...
+    ?V
+cannot produce expected type
+  () -> ()
+type-alias-path.mo:29.10-29.12: type error [M0096], expression of type
+  {x : M.MyType}
+cannot produce expected type
+  {x : M.MyType}
```
Kamirus, line 20: It is still possible to create such errors, we cannot avoid that since we can rename modules

src/mo_def/syntax.ml
```diff
@@ -10,23 +10,23 @@ type typ_note = {note_typ : Type.typ; note_eff : Type.eff}
...
+
+type id = string Source.phrase
+
+(* type id_ref, see below *)
+type typ_id = (string, Type.con option) Source.annotated_phrase
+
 (* Resolved imports (filled in separately after parsing) *)
 
-type lib_path = {package : string option; path : string}
+type lib_path = {package : string option; path : string; id : id option}
```
Kamirus: Tracking of identifiers used in imports.
In future we could redesign imports in Motoko and use the path to construct the namespace.

crusso:
Hmm. Not too sure about this solution - it feels quite invasive and the types are always printer with their full paths, even when shorter ones are available, right? I'd like to try something else but let's keep this open in case that doesn't pan out.

PR #5684, opened 2025-11-26
chore: bump `rustc`-nightly again

ggreif:
We'll have `rustc-1.88.0-nightly-2025-04-19`.

### DIFF
--- a/doc/md/fundamentals/2-actors/3-data-persistence.md
+++ b/doc/md/fundamentals/2-actors/3-data-persistence.md
@@ -120,7 +120,7 @@ When upgrading a canister, it is important to verify that the upgrade can procee
 -   Introducing an incompatible change in stable declarations.
 -   Breaking clients due to a Candid interface change.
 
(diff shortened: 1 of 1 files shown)


test/bench/ok/alloc.drun-run.ok
```diff
@@ -1,9 +1,9 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
-debug.print: (+671_088_640, 5_506_115_609)
+debug.print: (+671_088_640, 5_522_892_825)
```
ggreif, line 3: This is a quite substantial pessimisation. Ca. 0,304%.

alexandru-uta: I wouldn't necessarily call 0.3% substantial :). I think it's fine

ggreif: Well, for nothing in exchange it looks expensive in face of a mere compiler version bump. But it seems the size went down, so maybe it is more clever and the resulting instruction stream is just more expensive. In that case it will probably stay, and there is little we can do anyway...

PR #5690, opened 2025-11-27
bug: Revert 'record when block expected error'

Kamirus:
Reverting because of the following misleading error:

```motoko
actor {
  public func main() : async () {
    let x = 1;
    y = 3; // syntax error [M0001], expected block but got record; wrap the record in braces: { { … } }
    var z = 2;
    ignore (x, y, z)
  }
}
```
Missing a `let` or `var` (which can easily happen for the AI) is misinterpreted as a record syntax.
Because of ambiguity it is hard to decide whether the user wanted to write a record or a block.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -52,10 +52,6 @@
 
   * Use `self` parameter, not `Self` type, to enable contextual dot notation (#5574).
 
-  * Enable parser recovery to gather more syntax errors at once (previously only enabled for `moc.js`) (#5589).
-
-  * Custom syntax error message when a record is provided where a block is expected (#5589).
-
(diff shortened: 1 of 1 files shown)


test/fail/parse-block-or-record.mo
```diff
@@ -69,4 +69,21 @@ func _conds() {
...
+  func _varFirst() {
+    var x = 1;
+    x = 2;
+    ignore x;
+  };
+  func _base() {
+    x = 2; // Missing let/var? Or assignment := operator? Or meant a record?
+    ignore x; // Actually a block
+  }
+}
```
crusso:
```suggestion
}

```

crusso (approved):

PR #5691, opened 2025-11-27
chore: fix `reports` creation

ggreif:
This was only triggered on `master`, so it regressed:
https://github.com/dfinity/motoko/actions/runs/19743469155/job/56572868022

This PR attempts to fix the slight fallout from #5032.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -94,7 +94,6 @@ jobs:
       - uses: cachix/cachix-action@v16
         with:
           name: ic-hs-test
-          authToken: ${{ inputs.cachix-auth-token }}
       - uses: nixbuild/nixbuild-action@v24
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -90,6 +90,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
+      - uses: cachix/cachix-action@v16
+        with:
+          name: ic-hs-test
+          authToken: ${{ inputs.cachix-auth-token }}
```
ggreif:
```suggestion
```
Do we need write back to `cachix`?

basvandijk: Probably indeed not needed

ggreif: We'll see!

PR #5693, opened 2025-12-01
Gabor/typtbl

ggreif:


rts/motoko-rts/src/buf.rs
```diff
@@ -35,7 +46,7 @@ pub(crate) unsafe fn read_word(buf: *mut Buf) -> u32 {
     // IDL buffer is still 32-bit-based.
     const WORD_SIZE: usize = core::mem::size_of::<u32>();
 
-    if (*buf).ptr.add(WORD_SIZE - 1) >= (*buf).end {
+    if (*buf).ptr.add(WORD_SIZE) > (*buf).end {
```
ggreif, line 49: this is equivalent, but can CSE with below

rts/motoko-rts/src/idl.rs
```diff
@@ -1258,7 +1265,7 @@ pub(crate) unsafe fn sub(
 
 #[no_mangle]
 unsafe extern "C" fn idl_sub_buf_words(typtbl_size1: usize, typtbl_size2: usize) -> usize {
-    return BitRel::words(typtbl_size1, typtbl_size2);
+    BitRel::words(typtbl_size1, typtbl_size2)
```
ggreif, line 1268: equivalent

src/codegen/compile_classical.ml
```diff
@@ -6776,6 +6776,8 @@ module MakeSerialization (Strm : Stream) = struct
     let idl_value_numerator = 1l
     let idl_value_denominator = 1l
     let idl_value_bias = 1024l
+    let idl_type_scaler = 3l
```
ggreif: `s/idl_type_/idl_typetbl_/g`

src/codegen/compile_classical.ml
```diff
@@ -6787,7 +6789,9 @@ module MakeSerialization (Strm : Stream) = struct
       E.add_global32 env "@@value_denominator" Mutable idl_value_denominator;
       E.add_global32 env "@@value_numerator" Mutable idl_value_numerator;
       E.add_global32 env "@@value_bias" Mutable idl_value_bias;
-      E.add_global64 env "@@value_quota" Mutable 0L
+      E.add_global64 env "@@value_quota" Mutable 0L;
+      E.add_global32 env "@@type_scaler" Mutable idl_type_scaler;
```
ggreif: similarly

src/codegen/compile_classical.ml
ggreif: Maybe we should have `{scaler = 1; bias = 512}`, I am not sure about the current defaults.

src/prelude/prim.mo
```diff
@@ -824,6 +824,24 @@ func getCandidLimits<system>() : {
   };
 };
 
+func setCandidTypeLimits<system>({
```
alexandru-uta, line 827: exposing this method to the user seems like a bit too much power/opportunity to do wrong things; no strong opinions but I wonder if this shouldn't perhaps be more like a compile-time setting, perhaps?

src/codegen/compile_classical.ml
alexandru-uta: probably fine to have a larger default;

src/prelude/prim.mo
ggreif: Yeah, it is dangerous if a library changes it. But so is `setCandidValueLimits`, and this is the reason for demanding `<system>` privileges.

test/bench/typtbl.mo
```diff
@@ -0,0 +1,36 @@
...
+  type H1<T> = [[[[[[[[[[[[[[[[T]]]]]]]]]]]]]]]];
+  type H2<T> = H1<H1<H1<H1<H1<H1<H1<H1<H1<H1<H1<H1<H1<H1<H1<H1<T>>>>>>>>>>>>>>>>;
+
+  type Nested256 = H2<Nat8>;
+  type Nested512 = H2<H2<Nat8>>;
+  type Nested768 = H2<H2<H2<Nat8>>>;
+  type Nested1024 = H2<H2<H2<H2<Nat8>>>>;
+
+  debugPrint(debug_show getCandidTypeLimits<system>());
+
```
ggreif, line 20: We could set a minimal limit here that lets us pass below decoding.

crusso:
@ggreif why does the PR think you are an external contributor? That's probably why CI hasn't run either.

src/prelude/prim.mo
crusso: I'm fine with the current approach, which is consistent with how we control the other candid decoding limits and avoids more mysterious compiler flags

src/codegen/compile_classical.ml
ggreif: @crusso suggested
```suggestion
    let idl_type_scaler = 16l
```

src/codegen/compile_enhanced.ml
```diff
@@ -7097,6 +7097,8 @@ module Serialization = struct
     let idl_value_numerator = 1L
     let idl_value_denominator = 1L
     let idl_value_bias = 1024L
+    let idl_typetbl_scaler = 3L
```
ggreif: @crusso suggested
```suggestion
    let idl_typetbl_scaler = 16L
```

src/codegen/compile_classical.ml
```diff
@@ -6776,6 +6776,8 @@ module MakeSerialization (Strm : Stream) = struct
     let idl_value_numerator = 1l
     let idl_value_denominator = 1l
     let idl_value_bias = 1024l
+    let idl_typetbl_scaler = 3l
```
ggreif: suggested by @crusso
```suggestion
    let idl_typetbl_scaler = 16l
```

PR #5698, opened 2025-12-03
feat: pretty print type constructors using local environment relative type paths

crusso:
Less invasive alternative to #5679.

Before printing diagnostics, registers a map from type constructor to path, constructed from the current env.

Displays type using paths relevant to the errror location.

- [x] represent paths as more accurate (id * lab list), not string list (which could be empty and is less descriptive)
- [x] possible enhancement: when constructing con_map, don't just update to current path but select shortest path

### DIFF
--- a/test/run-drun/ok/scope-example-implicit.tc.ok
+++ b/test/run-drun/ok/scope-example-implicit.tc.ok
@@ -3,6 +3,6 @@ scope-example-implicit.mo:10.12-10.19: type error [M0096], expression of type
 cannot produce expected type
   async<$f> Int
 because of incompatible async scopes: 
-  $@anon-async-8.12
+  $
  does not match 
   $f
--- a/test/run-drun/ok/scope-example.tc.ok
+++ b/test/run-drun/ok/scope-example.tc.ok
@@ -3,6 +3,6 @@ scope-example.mo:12.13-12.27: type error [M0096], expression of type
 cannot produce expected type
   async<$@anon-async-6.60> Int
 because of incompatible async scopes: 
-  $@anon-async-10.12
+  $
  does not match 
   $@anon-async-6.60


src/mo_frontend/typing.ml
```diff
@@ -128,6 +128,14 @@ let kind_of_field_pattern pf = match pf.it with
   | ValPF(id, { it = VarP pat_id; _ }) when id = pat_id -> Scope.FieldReference
   | _ -> Scope.Declaration
 
+let con_map env =
```
Kamirus, line 129: This could be costly to compute on every message right?
But we could cache it in the `env` and update it when the `env` changes

src/mo_types/type.ml
```diff
@@ -859,6 +860,40 @@ let concrete t =
...
+let paths p t =
+  let cm = ref ConEnv.empty in
+  let seen = ref S.empty in
+  let rec go p t =
+    if  S.mem t !seen then () else
+    begin
+      seen := S.add t !seen;
+      match t with
+      | Typ c ->
+        cm := ConEnv.add c (List.rev p) (!cm)
```
Kamirus: Aha so this is looking for type aliases definitions in modules

src/mo_frontend/typing.ml
crusso: Yes we could, but I don't even know if its worth it. Error reporting is on the slow path so we just make that a bit slower. Maybe a problem in the IDE...

crusso:
> Nice idea! I like it better than mine experiment. Yeah it is less invasive, but it does not cover not imported modules, e.g. still prints `Order` instead of `Types.Order` but that is fine. Maybe that could be included by iterating over `env.libs` too.
Yeah, good point. For types that are unreachable by a path, we just use the tycon name. Iterating over env.libs might help, but some type constructors will remain unreachable (consider a module with a private type definition that occurs in a public member).
 
> I'm worried about type-checking performance, computing `con_map` on every message might be a pain.
Yes, we could maintain it in env instead, was just going for minimal change just now (experiment).

src/mo_frontend/typing.ml
Kamirus: Yes, that is the problem in the IDE. Errors are almost always present when coding. But yeah this is just a hunch, we would need to measure how slow it is

test/fail/type-alias-path/importM1.mo
```diff
@@ -0,0 +1,5 @@
+import M "m1";
+
+module {
+  public func m() : M.MyType = M.m();
+}
```
crusso:
```suggestion
}

```

test/fail/type-alias-path/importM2.mo
```diff
@@ -0,0 +1,5 @@
+import M "m2";
+
+module {
+  public func m() : M.MyType = M.m();
+}
```
crusso:
```suggestion
}

```

src/mo_frontend/typing.ml
Kamirus: Ok, I've compared the compile times of the Map.test.mo (has 36 warnings currently) in the motoko-core and it is just slightly slower, 1.5s vs 1.6s

crusso: Thanks, that's not too bad I think.

PR #5702, opened 2025-12-04
feat: Support break/continue without labels

Kamirus:
Support `break` and `continue` without labels to skip/exist the inner most loop.

### DIFF
--- a/src/lowering/desugar.ml
+++ b/src/lowering/desugar.ml
@@ -322,11 +322,11 @@ and parenthetical send = function
   | Some par ->
     (* fishing for relevant attributes in the parenthetical based on its static type *)
     let cycles, clean_cycles =
-      if T.(sub par.note.note_typ (Obj (Object, [T.cycles_fld])))
+      if T.(sub par.note.note_typ (Obj (Object, [T.cycles_fld], [])))
       then [fun parV -> dotE parV T.cycles_lab T.nat |> assignVarE "@cycles" |> expD], []
       else [], [] in
     let timeout, clean_timeout =
-      if T.(sub par.note.note_typ (Obj (Object, [T.timeout_fld])))
+      if T.(sub par.note.note_typ (Obj (Object, [T.timeout_fld], [])))
       then [fun parV -> dotE parV T.timeout_lab T.nat32 |> optE |> assignVarE "@timeout" |> expD], []
       else [], [nullE () |> assignE (var "@timeout" T.(Mut (Opt nat32))) |> expD] in
     (* present attributes need to set variables, absent ones just clear out *)
@@ -426,9 +426,9 @@ and build_field {T.lab; T.typ;_} =
 
 and build_fields obj_typ =
     match obj_typ with
-    | T.Obj (_, fields) ->
-      (* TBR: do we need to sort val_fields?*)
-      List.map build_field (T.val_fields fields)
+    | T.Obj (_, fields, _) ->
+      (* TBR: do we need to sort fields? *)
+      List.map build_field fields
     | _ -> assert false
 
 and with_self i typ decs =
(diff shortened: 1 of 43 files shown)


src/mo_frontend/typing.ml
```diff
@@ -2047,24 +2049,28 @@ and infer_exp'' env exp : T.typ =
...
-  | LoopE (exp1, None) ->
+  | LoopE (exp1, None, loop_ref) ->
     if not env.pre then begin
-      check_exp_strong env T.unit exp1
-    end;
-    T.Non
-  | LoopE (exp1, Some exp2) ->
+      let env' = { env with closest_loop = Some loop_ref } in
+      check_exp_strong env' T.unit exp1;
+      if loop_ref.has_break then T.unit else T.Non
```
Kamirus: My idea to allow `break` in the infinite `loop`. I'm not too sure about this because during `env.pre` the type can be different than during `not env.pre`.
Alternatively we can also not allow `break` without a label here, for this infinite loop construct

crusso:
I wonder if there might be an unsoundness with break from a `loop` (investigating)

src/lowering/desugar.ml
```diff
@@ -261,16 +273,29 @@ and exp' at note = function
...
        [ letD th thunk ]
        { e1 with it = I.TryE (exp e1, cases cs, Some (id_of_var th, typ_of_var th)); note }).it
-  | S.WhileE (e1, e2) -> (whileE (exp e1) (exp e2)).it
-  | S.LoopE (e1, None) -> I.LoopE (exp e1)
-  | S.LoopE (e1, Some e2) -> (loopWhileE (exp e1) (exp e2)).it
-  | S.ForE (p, {it=S.CallE (None, {it=S.DotE (arr, proj, _); _}, _, (_, e1)); _}, e2)
-      when T.is_array arr.note.S.note_typ && (proj.it = "vals" || proj.it = "values" || proj.it = "keys")
-    -> (transform_for_to_while p arr proj (!e1) e2).it
-  | S.ForE (p, e1, e2) -> (forE (pat p) (exp e1) (exp e2)).it
+  | S.WhileE (e1, e2, flags) ->
```
crusso, line 276: Yikes, this is quite scary.

Kamirus: You mean too many changes? I could refactor it but it probably won't be easier to read, the base cases are the same as before when the flags are all false.

Kamirus: Basically when `has_continue` the body a loop changes, it gets a label.
when `has_break` the whole loop expression gets a label on top (this is the recursive step).
This is the same desugaring as we do in the parser for the labeled break/continue

ggreif (commented):
Before I dive deeper in the code changes there are a few fundamental comments
- I consider a labelless `break` and `continue` an antipattern with C heritage (adding a e.g. `for` loop around them changes the semantics)
- it sacrifices consistency with the more general mechanism of aborting the current expression's evaluation by supplying a premature result
- the abbreviated form won't allow supplying a surrogate value
- relying on the desugarer to supply semantics (in certain corner cases) makes the _reference interpreter_ moot
- we have to be careful not to sacrifice certain core principles of the language on the altar of the AI weaknesses (they might easily master the explicitly labelled `break` and `continue` the next month)

src/mo_frontend/typing.ml
crusso: I think that might be sensible. You can still break with an explicit label and value if desired, right, in which case the type is determined by the exlicit label.

christoph-dfinity (commented):
Will sit down and experiment with moving some code around in the evening, but my first high level thoughts:

- It's unfortunate that label-less break from a `loop` can't return a value (because that would syntactically overlap with `break label`.
- I'm wondering if we could make the desugaring pass simpler by having the parser insert synthetic labels for every loop (how expensive is an unused label per loop in the generated code?)

In general I understand that label-less break/continue is a bit of a refactoring hazard, but at the same it's just so convenient...

Kamirus:
> * by having the parser insert synthetic labels for every loop

You can do that, here is the old version https://github.com/caffeinelabs/motoko/pull/5702/changes/BASE..655dfc141b7f762452359cb947708c3381462053
You can also see the updated benchmark values
And you can also see that by doing so the simple `loop` construct cannot be supported, because as you auto-label it, you need to insert a type there and without typechecking you need to just pick `T.unit`.
I don't think desugaring should be done in parser in general, so I would not add to that.

Kamirus:
> I'm not saying we add labels to `continue` and `break` in the parser, just to `loop`, `for`, and `while`, so that we'd always have a target in the type checker (`closest_loop : label`) and could have `ref`s in the continue/break expressions that we modify locally, rather than storing references in the env. That would also make the desugaring more straightforward, right? (At the cost of a few unused labels)

Have a look at the old version ^ that's exactly what is being done there. The labels are added on top of loops, these are the break-labels. As you add them you need to pick a type for them. Without typing info you have to choose `T.unit` (or T.Non for simple `loop`, but then you cannot `break` out of this loop anymore, unless you make the type of a label mutable 😄)

PR #5704, opened 2025-12-05
docs/0.16.3 -> prepare split of marketing and dev-docs page

marc0olo:
I think it would be good to merge this rather sooner than later.

I made docs changes which are relevant for the split of the marketing and dev docs page and the related PR is here:
- https://github.com/dfinity/portal/pull/6163

IMO these changes here should not have a real impact on the current portal page. it's rather an improvement to avoid linking to specific domains wherever possible and use the repository internal references instead.

I am wondering about docs in this folder, does anybody know if they are somehow used somwhere?
- https://github.com/dfinity/motoko/tree/master/doc/attic

ideally we can keep this branch even after moving until a new Motoko release is published.


marc0olo:
ok I just pushed a copy of this branch to https://github.com/dfinity/motoko/tree/docs/0.16.3-keep so that I can merge the master into this one.

alexandru-uta (approved):
looks okay to me, thanks for the effort!

PR #5706, opened 2025-12-05
perf: spot neutral elements in numeric expressions

### LINKED ISSUE #5555: `moc` not optimising `i+0`
I noticed during profiling that the compiler does not seem to optimise the expression `i + 0` to simply `i`. Is that expected?

I think it does optimise `i / 2` to bit shifts, also for higher 2-powers such as `i / 64`. Not 100% sure but I got this impression during profiling. Hence I was wondering why not `i + 0`?

ggreif:
Fixes #5555.

TODO
- [x] `or` (`|`) — DONE: a26f4b31556330a59151bd23c207f43c5b7f0496
- [ ] `and` could be supported too, is it worth it?
- [ ] other obvious identities as listed by @christoph-dfinity, below

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # Motoko compiler changelog
 
+* motoko (`moc`)
+
+  * perf: Optimise a few arithmetic/logic operations involving neutral elements (#5706).
+
 ## 1.2.0 (2026-02-12)
 
 * motoko (`moc`)


christoph-dfinity (commented):
Why do this during desugaring as opposed to the peephole optimizations we do in `instrList.ml`? That way we'd probably also catch things like `x += 0`, ....

What about other identities/constant exprs? Is there a line we want to draw somewhere?
```
x << 0 = x
x >> 0 = x
x | 0 = x
x | x = x
x & 0 = 0
x & x = x
x ^ x = 0
...
```

test/run.sh
```diff
@@ -114,6 +114,8 @@ function run () {
   shift
 
   if grep -q "^//SKIP $ext$" $(basename $file); then return 1; fi
+  local FILTER_LINE=$(grep -E -w "^//FILTER $ext (tail|head) " $(basename $file) | cut -d' ' -f3-)
```
ggreif: Exclude shell meta characters! Otherwise we'll have a possible unwanted command execution security hole.

ggreif: DONE: e6c140e36

src/lowering/desugar.ml
```diff
@@ -314,6 +318,36 @@ and parenthetical send = function
     (* if all attributes are absent, we still have to evaluate the parenthetical for side-effects *)
     else [expD (exp par)], absent
 
+and neutral (op : (binop, binop) Either.t) : exp -> bool =
+  let add_like = function Either.(Left (AddOp | OrOp) | Right (AddOp | OrOp | SubOp)) -> true | _ -> false in
+  let mul_like = function Either.(Left MulOp | Right (MulOp | DivOp)) -> true | _ -> false in
+  let rec strip e = match e.it with
```
```suggestion
  let rec examine e = match e.it with
```

test/run.sh
```diff
@@ -114,6 +114,8 @@ function run () {
   shift
 
   if grep -q "^//SKIP $ext$" $(basename $file); then return 1; fi
+  local FILTER_LINE=$(grep -E "^//FILTER $ext [A-Za-z0-9 -]*$" $(basename $file) | cut -d' ' -f3-)
```
crusso, line 117: So what are these changes doing, and do we need to keep them?

src/lowering/desugar.ml
```diff
@@ -67,6 +67,10 @@ and exp' at note = function
   | S.LitE l -> I.LitE (lit !l)
   | S.UnE (ot, o, e) ->
     I.PrimE (I.UnPrim (!ot, o), [exp e])
+  | S.BinE (_ot, e1, op, e2) when neutral (Either.Left op) e1 ->
```
crusso, line 82: I seems a bit strange to do this in desugaring rather than a separe IR pass, but I guess one could easily move this logic to another pass in future.

test/run/issue-5555.mo
```diff
@@ -0,0 +1,35 @@
...
+
+ignore 25 : Nat8 * 1;
+ignore 25 * 1 : Nat16;
+ignore 25 * 1 : Nat32;
+ignore 25 * 1 : Nat64;
+
+//SKIP run
+//SKIP run-ir
+//SKIP run-low
+//FILTER comp tail -n 27
```
crusso, line 34: Ok, I guess its this.

ggreif:
> Maybe we should remove the run.sh hacks. It's complicated enough as is.

The problem is that only the tail of the IR dump is relevant for testing, and the front part is huge and if we always compare that too the brittleness of changes will surface. That's why I added this shell filtering. I considered `FileCheck` testing, but that is quite some harder (and those expressions may vanish in the optimiser).

src/lowering/desugar.ml
ggreif: Yeah, but out also an IR pass will burden everything with complete traversal over the full tree, while this way is very selective. But yes, we can move this out if we want some more comprehensive solution.

src/lowering/desugar.ml
```diff
@@ -339,6 +343,36 @@ and parenthetical send = function
...
+  let mul_like = function Either.(Left MulOp | Right (MulOp | DivOp)) -> true | _ -> false in
+  let rec examine e = match e.it with
+    | S.AnnotE (e, _) -> examine e
+    | S.LitE {contents} ->
+      let open Numerics in
+      (match contents with
+       | NatLit n | IntLit n when add_like op && Int.(eq n zero) -> true
+       | NatLit n | IntLit n when mul_like op && Int.(of_int 1 |> eq n) -> true
+       | Nat8Lit n when add_like op && Nat8.(eq n zero) -> true
+       | Nat8Lit n when mul_like op && Nat8.(of_int 1 |> eq n) -> true
```
crusso, line 357: I think there might be a `Nat8.one` etc.

src/lowering/desugar.ml
```diff
@@ -79,6 +79,10 @@ and exp' at note = function
   | S.LitE l -> I.LitE (lit !l)
   | S.UnE (ot, o, e) ->
     I.PrimE (I.UnPrim (!ot, o), [exp e])
+  | S.BinE (_ot, e1, op, e2) when neutral (Either.Left op) e1 ->
```
crusso, line 82: NIT: It might be a little faster to move the three binop cases into a single case with a nested if: I think pattern matching is slowed down by guards. Might be wrong though.

PR #5711, opened 2025-12-08
nix flake update

basvandijk:
Continuation of https://github.com/dfinity/motoko/pull/5692 to upgrade the dfinity/ic repo to upgrade the pocket-ic-server.


alexandru-uta:
I'd say let's just drop this, seems too much effort for the return -- we should probably just install pocket-ic as a binary from the ic releases; I can look into that

PR #5722, opened 2025-12-10
feat: Null coalesce operator

### LINKED ISSUE #5144: Syntax sugar for `Option`: Null coalescing `??`
## Null coalescing

Idea: `option1 ?? option2 ?? option3`

Instead of:
```motoko
func nullCoalescingWithoutDefault(option1 : ?Nat, option2 : ?Nat, option3 : ?Nat) : ?Nat {
  switch (option1) {
    case (?value) { ?value };
    case null {
      switch (option2) {
        case (?value) { ?value };
        case null { option3 }
      }
    }
  }
};
```

It should work with both option and non-option types for the last argument:
1. `option1 ?? option2 ?? option3 : ?T`
2. `option1 ?? option2 ?? default : T`

Alternative: Just use "mo:base"
```motoko
import Option "mo:base/Option";
func nullCoalescingWithDefault2(option1 : ?Nat, option2 : ?Nat, default : Nat) : Nat =
  Option.get(option1, Option.get(option2, default));
```
but this is not ideal:
- **not lazy**: it evaluates the whole chain of options, `??` should stop on the first non-null
- still a bit verbose
- needs additional `(?T, ?T) -> ?T` function as an alternative to `Option.get`, maybe named `Option.alt`?

### Potential problems with `??`
1. Ambiguous type: harder to typecheck
2. Operator might be difficult to parse

Kamirus:
Closes #5144 

Reduces verbosity of common expressions like
```motoko
switch(opt) {
  case (null) { defaultOrTrap(...) };
  case (?v) { v };
};
```
To just
```motoko
opt ?? defaultOrTrap(...)
```

The operator `??` is commonly used in other mainstream languages like TS/JS and C# so it is familiar for both humans and AI. 

Follow up: @rvanasa @christoph-dfinity we would need to update the grammar for tools like the formatter, lintoko, vscode etc, yes?


src/mo_frontend/parser.mly
```diff
@@ -453,6 +455,8 @@ typ_un :
     { t }
   | QUEST t=typ_un
     { OptT(t) @! at $sloc }
+  | NULLCOALESCE t=typ_un
```
Kamirus, line 464: Having both QUEST and NULLCOALESCE tokens causes `??` to be lexed as NULLCOALESCE, not double QUEST.
Just adding missing parsing where double `??` is allowed.
This slightly changes the region, but that is a small issue

crusso: I think you could actually fix-up the inner sloc

src/mo_frontend/parser.mly
```diff
@@ -733,6 +739,8 @@ exp_un(B) :
     { AndE(e1, e2) @? at $sloc }
   | e1=exp_bin(B) OR e2=exp_bin(ob)
     { OrE(e1, e2) @? at $sloc }
+  | e1=exp_bin(B) NULLCOALESCE e2=exp_bin(ob)
```
crusso: Since this is a bit like a conditional, we could also consider allowing a block on the rhs (like an else branch).
```
| e1=exp_bin(B) NULLCOALESCE e2=exp_nest
```
Maybe a bad idea

test/run-drun/migration-paths/new-installer.mo
```diff
@@ -6,33 +6,21 @@ actor {
...
 
     public func upgrade() : async () {
-        switch testCanister {
-            case null Prim.trap("null canister");
-            case (?canister) {
-                ignore await (system TestCanister.TestCanister)(#upgrade canister)();
-                Prim.debugPrint("Upgraded (default)");
-            };
-        };
+        let canister = testCanister ?? Prim.trap("null canister");
```
crusso, line 9: FTR this and the examples below could all be written already using let-else

```suggestion
        let ?canister = testCanister else Prim.trap("null canister");
```

crusso:
Looks nice enough to me - lets see what the others think. 

Are there any other (commonly used) symbols for `??` that avoid the parser hacks? `
Perhaps ` e !/ e` (since we already !  in do ? { ... e! })

alexandru-uta:
> Are there any other (commonly used) symbols for `??` that avoid the parser hacks? `Perhaps` e !/ e`(since we already use`!` in`do ? { ... e ! ... }`)`.

Since we try to help an AI write more concise code we should stick to `??` because then we can take advantage of the base model knowing JS/TS.

src/mo_frontend/parser.mly
Kamirus: yes we could set the correct region for the inner one, I'll change that if we decide on going forward with the `??` operator

src/mo_frontend/parser.mly
Kamirus: Oh! Maybe allowing a block on the left? so that
```motoko
do ? {
} ?? default
```
Would parse properly without a parenthesis? 
Not allowing a block on the RHS would parse records better yes? Without extra brackets

Kamirus:
TODO:
Should we allow blocks on the LHS or RHS? 

Examples:
RHS without block:
```motoko
dailyLogs.get(caller) ?? do {
  dailyLogs.add(caller, Map.empty<Nat, DailyRecord>());
  getDailyLogs(caller);
};
```

RHS without block vs with block
```motoko
foo.bar(qux) ?? { x = 0; z = null };
```
vs
```motoko
foo.bar(qux) ?? { { x = 0; z = null } };
```

LHS with block:
```motoko
do ? {
  ...
} ?? default
```
Does it work?

Kamirus:
> > I was not aware of the >> disambiguation
> 
> Right sorry, I should've linked it.

Thanks! No worries I would have found it 😄 I meant that I was not aware of that mechanism when I first added the `??` token 😅

alexandru-uta (approved):

PR #5725, opened 2025-12-10
feat: just make implicit a (contextual) keyword 

crusso:
If everyone agrees, I'd like to include this in Motoko 1.0.

This PR just makes `implicit` a keyword but lets you parse (implicit : T) as before.

If we do this now, with 1.0, then adding syntax that needs it will (likely) not be a breaking change in future (since we already have the keyword).

This is essentially a cut-down version of #5659 (already approved by Kamil), without the extra syntax and wildcard implicits removed.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "candid-src": {
       "flake": false,
       "locked": {
-        "lastModified": 1762175579,
-        "narHash": "sha256-ChlkmnamNG6LfpxEW6GwES6xigYwFgdBgzoNLrq0zJ4=",
+        "lastModified": 1765407289,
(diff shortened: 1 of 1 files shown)


doc/md/16-language-manual.md
```diff
@@ -2334,9 +2335,11 @@ the expanded function call expression `<parenthetical>? <exp1> <T0,…​,Tn>? <
     ```
     insert_holes(n ; <empty> ; <exps>) =
       <exps>
-    insert_holes(n ; ( (<id0> : (<implicit> : (<id> : U))), Us) ; <exps>) =
+    insert_holes(n ; ( ((implicit : (<id> : U))), Us) ; <exps>) =
+      hole(n, <id>, U), insert_holes(n + 1 ; Us ; exps)
```
crusso:
```suggestion
```
No longer necessary.

PR #5731, opened 2025-12-11
feat: allows resolving local definitions for context-dot

christoph-dfinity:
Resolves [LANG-12](https://linear.app/caffeinelabs/issue/LANG-12/allow-using-local-definitions-with-contextual-dot)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,4 @@
-* @dfinity/languages
+* @caffeinelabs/team-languages
 Changelog.md
 doc/docusaurus/package-lock.json
 flake.lock
--- a/.github/workflows/release-pr.yml
+++ b/.github/workflows/release-pr.yml
@@ -55,8 +55,8 @@ jobs:
       uses: actions/create-github-app-token@v2
       id: app-token
       with:
-        app-id: ${{ vars.PR_AUTOMATION_BOT_PUBLIC_APP_ID }}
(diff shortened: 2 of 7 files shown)


src/mo_frontend/typing.ml
```diff
@@ -1525,7 +1525,7 @@ let resolve_hole env at hole_sort typ =
      end
 
 type ctx_dot_candidate =
-  { module_name : T.lab;
+  { module_name : T.lab option;
```
crusso, line 1528: Minor nit: T.lab is used to describe field labels, but the module is really bound in the context to an *identifier*, not a label.  Not sure we have a T.id though...

crusso: Of course it doesn't really matter now since they are all strings but if we ever get our act together....

crusso (commented):
LGTM too.

BTW, even with this, I don't think we can quite emulate the special members using prelude define methods. 

The reason is the the current array.size is:
a) actually a function value.
b) cannot be redefined (since we prefer fields when present).
To fix a), I guess we could consider compiling `foo.method` (sans args) to a first class function fun args -> c(foo, args). where c is the candidate for foo.method.

crusso (approved):

christoph-dfinity:
> a) actually a function value.

I would be willing to break this, with the option of bringing the feature back more generally as you described. I'm a bit wary of introducing allocations for what looks like a projection.

PR #5732, opened 2025-12-11
feat: optimize pocket-ic usage

alexandru-uta:
Download pocket-ic-server instead of building it from source.

Uses pocket-ic from release of October 10th: https://github.com/dfinity/ic/releases/tag/release-2025-10-02_03-13-base

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -94,7 +94,7 @@ jobs:
         run: nix build --max-jobs 1 '.#"release-files-${{ matrix.os }}"'
 
       - name: Upload Release Artifacts
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: release-files-${{ matrix.os }}
           path: result/*
@@ -112,7 +112,7 @@ jobs:
     needs: [prepare, changelog, build]
     steps:
       - name: Download all release artifacts
-        uses: actions/download-artifact@v6
+        uses: actions/download-artifact@v7
         with:
           path: release-assets
           pattern: release-files-*
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -126,7 +126,7 @@ jobs:
       # https://github.com/actions/upload-artifact/issues/92
       - run: echo "UPLOAD_PATH=$(readlink -f result)" >> $GITHUB_ENV
       - name: upload artifacts
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: moc-${{ matrix.os }}
           path: ${{ env.UPLOAD_PATH }}


nix/pocket-ic.nix
```diff
@@ -1,73 +1,55 @@
...
-      [target.aarch64-apple-darwin]
-      rustflags = [ "-C", "linker=c++" ]
-      EOF
-    '';
-    nativeBuildInputs = with pkgs; [
-      pkg-config
-      cmake
+  server = pkgs.stdenv.mkDerivation rec {
+    pname = "pocket-ic-server";
+    version = "10.0.0"; # This is the pocket-ic production version.
```
christoph-dfinity: Which version controls what binary we get? This one, or the `releaseTag` above?

nix/pocket-ic.nix
```diff
@@ -1,73 +1,55 @@
...
+    
+    src = pkgs.fetchurl {
+      url = "${baseUrl}/${binaryName}.gz";
+      sha256 = sha256Map.${binaryName};
+      name = "pocket-ic-server.gz"; 
+    };
+    
+    dontUnpack = true;
```
christoph-dfinity, line 35: Mostly curious why you're setting this, only to run `gunzip` in the `installPhase`. Wouldn't it be easier to just let nix to do the unpacking here?

nix/pocket-ic.nix
alexandru-uta: It's the release tag above, this is just to signal what version of the pocket-ic we're at. Is it confusing?

christoph-dfinity: I could see myself debugging why changing this version doesn't pull in a new version of pocket-ic, yeah.

If we want to maintain this information let's put it in a comment above the `releaseTag` attribute. Otherwise these will just go out of sync.

nix/pocket-ic.nix
alexandru-uta: based on the documentation [here](https://nixos.org/manual/nixpkgs/stable/#ssec-unpack-phase), I didn't find a proper way to handle `.gz` besides manually doing it. Any ideas?

nix/pocket-ic.nix
alexandru-uta: okay, fixing it now.

nix/pocket-ic.nix
christoph-dfinity: Oh is that because this is not a gzipped archive, but just a single gzipped file? That makes sense then.

alexandru-uta: yes, unfortunately :(

nix/pocket-ic.nix
christoph-dfinity: You added the comment to the `shaMap`, I thought the `releaseTag` determined the version? How would I go about updating the `shaMap` when updating the pocketIc version. Would I need to download all the artifacts?

alexandru-uta: you gotta look here: https://github.com/dfinity/ic/releases/tag/release-2025-12-04_03-28-base; both the tag and the hashes are needed.

ggreif (approved):
Appears totally reasonable to download pinned binaries. When a measure is missing, we'll have to bump.

nix/pocket-ic.nix
alexandru-uta: I should probably add a few lines about the process and why it's done this way. I think this should be a more or less manual process because due to the newer pocket-ic features which could break our CI we need to consciously chose a version which works.

do you think this should be automated?

christoph-dfinity: Don't think automation is needed, but maybe some instructions for the next person that bumps these hashes/versions.

alexandru-uta: yup, done.

PR #5742, opened 2025-12-15
fix: Redundant type instantiation to check all places and avoid false positives (Alternative)

Kamirus:
An alternative to #5605 (see the description there)
this variant avoids mutating the expression state when performing the redundant instantiation check.

Risks: `coverage'` must be disabled during that check, because `PreLit` is not allowed there. Are there any other places where PreLit would cause `assert false` exception?


Kamirus:
Backtracking trail in `pre` phase is good enough approximation for now, with limited harm.

PR #5743, opened 2025-12-15
chore: bump `rustc-nightly`

ggreif:
Bumps:
- Rust nightly (huge: 2025-06-19 -> 2026-04-08)
- `nixpkgs` (small)
- minor syntactic changes (to remove warnings)

[macOS side](https://github.com/caffeinelabs/motoko/actions/workflows/nightly-macos-test.yml) looks good (but is plagued by build-machine related problems)

PR #5744, opened 2025-12-15
chore: bump `nixpkgs` to 25.11

ggreif:
this is the new stable branch, 25.05 is being deprecated

These will be done separately, (**fixed  — see #5743**)
```
evaluation warning: 'system' has been renamed to/replaced by 'stdenv.hostPlatform.system'
evaluation warning: 'runCommandNoCC' has been renamed to/replaced by 'runCommand'
```
to keep the diff small.

### DIFF
--- a/nix/mo-packages.nix
+++ b/nix/mo-packages.nix
@@ -16,7 +16,7 @@ let
         then "release-static"
         else "release";
       is_dyn_static =
-        is_static && pkgs.stdenv.hostPlatform.system == "aarch64-linux";
+        is_static && pkgs.system == "aarch64-linux";
     in
     staticpkgs.stdenv.mkDerivation {
       inherit name;
--- a/nix/pocket-ic.nix
+++ b/nix/pocket-ic.nix
@@ -6,7 +6,7 @@ let
     "aarch64-linux" = "pocket-ic-arm64-linux";
     "x86_64-darwin" = "pocket-ic-x86_64-darwin";
     "aarch64-darwin" = "pocket-ic-arm64-darwin";
-  }.${pkgs.stdenv.hostPlatform.system} or (throw "Unsupported system: ${pkgs.stdenv.hostPlatform.system}");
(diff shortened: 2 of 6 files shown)


src/codegen/instrList.ml
```diff
@@ -8,6 +8,13 @@ features are
 *)
 
 open Wasm_exts.Ast
+
+let cr Source.{left; right} =
```
ggreif, line 12: moved here to avoid record field ambiguities

nix/mo-packages.nix
```diff
@@ -16,7 +16,7 @@ let
         then "release-static"
         else "release";
       is_dyn_static =
-        is_static && pkgs.system == "aarch64-linux";
+        is_static && pkgs.stdenv.hostPlatform.system == "aarch64-linux";
```
ggreif:
```suggestion
        is_static && pkgs.system == "aarch64-linux";
```
Try reverting. @alexandru-uta didn't need this in his 25.11 migration...

src/codegen/instrList.ml
ggreif: Would be worthwhile to use the `Wasm.Source` one in `Wasm_exts.Ast` to not do this copying dance. I'll try.

src/exes/dune
```diff
@@ -31,6 +31,6 @@
 (executable
  (name candid_tests)
  (modules candid_tests)
- (libraries lib idllib mo_idl mo_values lang_utils)
+ (libraries lib idllib mo_idl mo_values lang_utils str)
```
ggreif, line 34: Needs a double-check

ggreif: Really needed:
```
File "_none_", line 1:                   
Error: No implementations provided for the following modules:
         Str referenced from exes/.candid_tests.eobjs/native/dune__exe__Candid_tests.cmx
```

src/codegen/instrList.ml
ggreif: I tried, not so easy. Will pursue this in a bit...

PR #5745, opened 2025-12-15
experiment: Try pretty-printing a diff when subtyping fails

Kamirus:
Idea / Specification 

```
type
  [Data = {... b : Text ...}]
 is not compatible with
  [Data = { ... }]
```

Print both types focusing on incompatible path found during subtype check.
It should print type aliases plus their normalized form.
It should skip most of the compatible fields to focus on the difference, but it should include some of the correct fields to keeps the surrounding (best to keep the 2 surrounding fields from the source, but what if they are sorted?)


Kamirus:
Experiment done

PR #5746, opened 2025-12-16
chore: drop `macos-15`

Kamirus:
Drop building MacOS Intel binaries


ggreif:
Here are a few lines that can be deleted too:
```
  CXX_x86_64-apple-darwin = CXX_aarch64-apple-darwin;
  CXXFLAGS_x86_64-apple-darwin = CXXFLAGS_aarch64-apple-darwin;
```

ggreif:
I'd suggest to keep this PR open a bit more. With `nixpkgs-26.11` support will be cut, and we have to drop it anyway. I'll remove it for good.

PR #5751, opened 2025-12-17
chore: remove viper

alexandru-uta:
Resolves LANG-54

### DIFF
--- a/.github/workflows/nightly-macos-test.yml
+++ b/.github/workflows/nightly-macos-test.yml
@@ -1,7 +1,7 @@
 name: "nightly-macos-test"
 on:
   schedule:
-    - cron: '0 0 * * *'  # Run at midnight UTC every day
+    - cron: "0 0 * * *"  # Run at midnight UTC every day
   workflow_dispatch:  # Allow manual triggering
   
 jobs:


christoph-dfinity (approved):
LGTM (but maybe have Claudio also take a look as he might know what other places to check)

We should also add a changelog entry

src/mo_def/syntax.ml
```diff
@@ -221,7 +219,7 @@ and exp' =
 and arg_exp = (bool * (exp ref))
 
 and assert_kind =
-  | Runtime | Static | Invariant | Precondition | Postcondition | Concurrency of string | Loop_entry | Loop_continue | Loop_exit | Loop_invariant
+  | Runtime
```
christoph-dfinity, line 222: Shouldn't we just delete `assert_kind`? It doesn't carry any information anymore.

alexandru-uta: Claudio said he'd keep it in case we want to add debug asserts later

christoph-dfinity: Thanks, that makes sense

.github/workflows/update-flake-lock.yml
```diff
@@ -13,10 +13,10 @@ on:
   schedule:
     # * is a special character in YAML so you have to quote this string
     # Run every day at midnight
-    - cron: '0 0 * * *'
+    - cron: "0 0 * * *"
 jobs:
   flake-updater:
-    name: 'Update flake.lock'
+    name: "Update flake.lock"
```
ggreif, line 19: Let me send these two separately to `master`, as they are independent changes.

ggreif: See #5755.

alexandru-uta:
> > We should also add a changelog entry
> 
> Looks like all is removed. We need to bump to 2.0.0 as functionality is gone.

Later, in a larger release.

PR #5752, opened 2025-12-17
chore: create token for `approvals`

ggreif:


.github/workflows/test.yml
```diff
@@ -179,7 +179,7 @@ jobs:
           ref: 'master'
       - name: Automatically closing successful trials
         env:
-          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          GH_TOKEN: ${{ github.token }}
```
ggreif, line 182: This is to undo the damage from #5749.

alexandru-uta (approved):
stamp

PR #5753, opened 2025-12-17
refactor: introduces helper functions for working with object types

christoph-dfinity:
Abstracting a few common patterns on object types in the hopes of making a move to separating value and type fields easier.

This way even if we don't find the big change worth it we might still have a few smaller sensible refactorings.

- refactor: introduces a helper function to enumerate object val fields
- refactor: helper function to align object fields
- refactor: uses object construction helper in a few more places

Resolves LANG-55

### DIFF
--- a/src/mo_types/type.ml
+++ b/src/mo_types/type.ml
@@ -1237,31 +1237,30 @@ let rec rel_typ d rel eq t1 t2 =
 
 and rel_fields t2 d rel eq tfs1 tfs2 =
   (* Assume that tfs1 and tfs2 are sorted. *)
-  let rel_fields = ref [] in
   let res = align_fields tfs1 tfs2 |>
     Seq.for_all (function
     | Lib.Both(tf1, tf2) ->
-      rel_fields := (tf1, tf2) :: !rel_fields;
+      add_src_field_update rel eq tf1 tf2;
       let d' = RelArg.push (Field tf2.lab) d in
       rel_typ d' rel eq tf1.typ tf2.typ
     | Lib.This tf1 ->
       if rel != eq && not (RelArg.is_stable_sub d) then true
       else missing_field d tf1.lab t2
-    | Lib.That tf2 -> unexpected_field d tf2.lab t2) in
-  if res then List.iter (fun (tf1, tf2) -> add_src_field_update rel eq tf1 tf2) !rel_fields;
+    | Lib.That tf2 ->
+      unexpected_field d tf2.lab t2) in
   res
 
 and rel_tags t2 d rel eq tfs1 tfs2 =
   (* Assume that tfs1 and tfs2 are sorted. *)
-  let rel_fields = ref [] in
   let res = align_fields tfs1 tfs2 |>
     Seq.for_all (function
       | Lib.Both(tf1, tf2) ->
-        rel_fields := (tf1, tf2) :: !rel_fields;
+        add_src_field_update rel eq tf1 tf2;
         rel_typ d rel eq tf1.typ tf2.typ
-      | Lib.This tf1 -> missing_tag d tf1.lab t2
-      | Lib.That tf2 -> if rel != eq then true else unexpected_tag d tf2.lab t2) in
-  if res then List.iter (fun (tf1, tf2) -> add_src_field_update rel eq tf1 tf2) !rel_fields;
+      | Lib.This tf1 ->
+        missing_tag d tf1.lab t2
+      | Lib.That tf2 ->
+        if rel != eq then true else unexpected_tag d tf2.lab t2) in
   res
 
 and rel_binds d rel eq tbs1 tbs2 =


src/lowering/desugar.ml
```diff
@@ -858,23 +857,21 @@ and obj obj_typ efs bases =
...
+        if T.is_mut typ then
+          refD id { it = I.DotLE(varE base_var, lab); note = typ; at = no_region }
+        else
+          letD id (dotE (varE base_var) lab typ) in
+      let f = { it = I.{ name = lab; var = id_of_var id }; at = no_region; note = typ } in
+      [d, f] in
 
   let dss, fs = map (exp_field obj_typ) efs |> split in
-  let ds', fs' = concat_map gap (T.as_obj obj_typ |> snd) |> split in
+  let ds', fs' = concat_map gap (T.as_obj obj_typ |> snd |> T.val_fields) |> split in
```
crusso, line 871: Was there a lurking bug here?

christoph-dfinity: I don't think so, I just removed the special handling for type fields from `gap` and filtered them out here instead.

src/mo_types/type.ml
```diff
@@ -1411,22 +1398,17 @@ let rec compatible_typ co t1 t2 =
 
 and compatible_fields co tfs1 tfs2 =
   (* Assume that tfs1 and tfs2 are sorted. *)
-  match tfs1, tfs2 with
-  | [], [] -> true
-  | tf1::tfs1', tf2::tfs2' ->
-    tf1.lab = tf2.lab && compatible_typ co tf1.typ tf2.typ &&
```
christoph-dfinity, line 1415: This was buggy before (would compare typ fields to value fields)

crusso: Good catch!

src/lib/lib.ml
```diff
@@ -439,13 +444,25 @@ struct
...
-  
+
   (* tail-recursive map *)
   let[@tail_mod_cons] rec safe_map f l = match l with
     | [] -> []
     | x :: xs ->
       f x :: (safe_map[@tailcall]) f xs
     [@@coverage off]
+
+  let align cmp xs ys =
```
crusso, line 455: I guess my only worry here is that there's quite a lot of allocation going on on each iteration. Maybe it doesn't matter at all, but we could try it on the cycleops benchmark (if I still have it around)

src/mo_types/type.ml
```diff
@@ -1235,62 +1238,32 @@ let rec rel_typ d rel eq t1 t2 =
...
   (* Assume that tfs1 and tfs2 are sorted. *)
-  match tfs1, tfs2 with
-  | [], [] ->
-    true
-  | tf1::_, [] when rel != eq && not (RelArg.is_stable_sub d) ->
-    true
-  | tf1::tfs1', tf2::tfs2' ->
-    (match compare_field tf1 tf2 with
-    | 0 ->
+  let rel_fields = ref [] in
```
crusso: This seems a lot nicer to read!

src/lib/lib.ml
christoph-dfinity: It would be great if you could run that for me. I'd hope that these are very cheap, since they're short-lived...

src/mo_types/type.ml
```diff
@@ -1653,35 +1621,30 @@ let rec combine rel lubs glbs t1 t2 =
...
+      | Lib.Both(f1, f2) ->
+        (match combine rel lubs glbs f1.typ f2.typ with
+        | typ ->
+          add_src_field rel lubs glbs f1 f2;
+          let src = {empty_src with track_region = f1.src.track_region} in
+          {lab = f1.lab; typ; src} :: acc
+        | exception Mismatch when rel == lubs -> acc)
+      | Lib.This f1 -> cons_if (rel == glbs) f1 acc
+      | Lib.That f2 -> cons_if (rel == glbs) f2 acc) []
+  |> List.rev
```
crusso, line 1633: Shame you need this extra step.

src/mo_types/type.ml
```diff
@@ -1653,35 +1621,30 @@ let rec combine rel lubs glbs t1 t2 =
...
+  |> Seq.fold_left (fun acc x -> match x with
+    | Lib.Both(f1, f2) ->
       let typ = combine rel lubs glbs f1.typ f2.typ in
       add_src_field rel lubs glbs f1 f2;
       let src = {empty_src with track_region = f1.src.track_region} in
-      {lab = f1.lab; typ; src} :: combine_tags rel lubs glbs fs1' fs2'
+      {lab = f1.lab; typ; src} :: acc
+    | Lib.This(f1) -> cons_if (rel == lubs) f1 acc
+    | Lib.That(f2) -> cons_if (rel == lubs) f2 acc) []
+  |> List.rev
```
crusso, line 1645: Ditto

src/lib/lib.ml
```diff
@@ -439,13 +444,25 @@ struct
...
+    let next (xs, ys) = match xs, ys with
+      | [], [] -> None
+      | x::xs', [] -> Some (This x, (xs', []))
+      | [], y::ys' -> Some (That y, ([], ys'))
+      | x::xs', y::ys' -> (match cmp x y with
+        | -1 -> Some (This x, (xs', ys))
+        | +1 -> Some (That y, (xs, ys'))
+        | _ -> Some (Both(x, y), (xs', ys')))
```
crusso, line 463:
```suggestion
    let next (xs, ys) = match xs, ys with
      | [], [] -> Done, [], []
      | x::xs', [] -> This x, xs', []
      | [], y::ys' -> That y, [], ys'
      | x::xs', y::ys' -> (match cmp x y with
        | -1 -> This x, xs', ys
        | +1 -> That y, xs, ys'
        | _ -> Both(x, y), xs', ys'
```
And then use that with explicit recursion.

crusso: Then I think you wouldn't need to accumulate stuff in refs nor reverse.

crusso: https://github.com/caffeinelabs/motoko/pull/5763 sketches this out (might be buggy)

src/mo_types/type.ml
christoph-dfinity: Yeah, this is necessary to re-establish the sorted order. It would go away if this folded into a map rather than a list.

src/mo_frontend/typing.ml
```diff
@@ -3510,36 +3495,35 @@ and check_pats_typ_dec env ts pats te at : Scope.typ_env =
...
-           let te1 = check_pat_typ_dec env typ p in
-           disjoint_union env at "M0017" "duplicate binding for %s in pattern" te te1
-        | _ -> assert false
+and check_pat_fields_typ_dec env t tfs pfs te at : Scope.typ_env =
+  let cmp (tf : T.field) pf = match tf, pf with
+    | T.{ lab; typ = T.Typ _; _}, { it = TypPF(id); _ } -> String.compare lab id.it
+    | _, { it = TypPF(id); _ } -> 1
+    | T.{ lab; _ }, { it = ValPF(id, _); _ } -> String.compare lab id.it in
+  let last_field = ref "" in
+  Lib.List.align cmp tfs pfs |>
```
crusso, line 3504: Do you actually know that pfs are sorted?

src/mo_types/type.ml
```diff
@@ -1235,62 +1237,32 @@ let rec rel_typ d rel eq t1 t2 =
...
-    )
-  | [], tf2::_ ->
-    unexpected_tag d tf2.lab t2
-  | tf1::_, [] ->
-    missing_tag d tf1.lab t2
+  let rel_fields = ref [] in
+  let res = align_fields tfs1 tfs2 |>
+    Seq.for_all (function
+      | Lib.Both(tf1, tf2) ->
+        rel_fields := (tf1, tf2) :: !rel_fields;
```
crusso: I think you could probably add the update here and avoid the ref. IIRC the updates are delayed and only applied if the root subtyp test succeeds (but good to double check).

src/mo_types/type.ml
```diff
@@ -1235,62 +1237,32 @@ let rec rel_typ d rel eq t1 t2 =
...
-  | [], tf2::_ ->
-    unexpected_field d tf2.lab t2
-  | tf1::_, [] ->
-    missing_field d tf1.lab t2
+      rel_typ d' rel eq tf1.typ tf2.typ
+    | Lib.This tf1 ->
+      if rel != eq && not (RelArg.is_stable_sub d) then true
+      else missing_field d tf1.lab t2
+    | Lib.That tf2 -> unexpected_field d tf2.lab t2) in
+  if res then List.iter (fun (tf1, tf2) -> add_src_field_update rel eq tf1 tf2) !rel_fields;
```
crusso: Same, here, I think you can get rid of the ref and record the update in the `Both` case. See comment below.

src/mo_frontend/typing.ml
```diff
@@ -3510,36 +3495,35 @@ and check_pats_typ_dec env ts pats te at : Scope.typ_env =
...
-      id.it
-      display_typ_expand t
-  | [], {it = ValPF(_); _}::pfs' ->
-     check_pat_fields_typ_dec env t tfs pfs' te at
-  | tf::tfs', pf::pfs' ->
-     match compare_pat_typ_field tf pf with
-     | -1 -> check_pat_fields_typ_dec env t tfs' pfs te at
-     | 1 ->
-        (match pf.it with
-        | TypPF(_) -> check_pat_fields_typ_dec env t [] pfs te at
```
crusso, line 3527: No sure I understand what this case was doing. Why tfs = [] in the recursive call?

christoph-dfinity: This case exists in both the `check_pat_fields` functions. It shortcuts to the error case for a missing field on the next round.

Rather ugly, really we should be reporting all field mismatches at once, instead of exiting at the first one.

src/mo_frontend/typing.ml
christoph-dfinity: The existing/replaced code also relied on it, so I didn't check.

src/mo_types/type.ml
christoph-dfinity: Yeah, it does look that way. Weird. I just tried to faithfully reproduce the previous behaviour here, but even better if that's not required.

src/mo_frontend/typing.ml
crusso: Oh, I now recall seeing Andreas' code use that trick elsewhere.

PR #5757, opened 2025-12-18
tooling: nix target picker

christoph-dfinity:
For when you can't remember all of the different build targets. Now you just run `nixb`!

first

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "candid-src": {
       "flake": false,
       "locked": {
-        "lastModified": 1765407289,
-        "narHash": "sha256-KBwXZWWgSIAbKpWNARNhbeAfypm9MLuj8ZlDVCGt5tg=",
+        "lastModified": 1766069143,
+        "narHash": "sha256-hllvzZ/phhWbxuHhA1tME1Nc0ZAT9fMAZgwWsa9BLNs=",
         "owner": "dfinity",
         "repo": "candid",
-        "rev": "fe9b0d86f40f9fd0c331ce8fabaa8b01b86381f3",
+        "rev": "2e4a2cf6a66593e2cb3289535899eb232635bef2",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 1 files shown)


basvandijk:
You could consider adding the following to the `flake.nix` (somewhere under the `formatter` attribute):
```nix
      # nix eval .#targets.$(nix-instantiate -E 'builtins.currentSystem' --eval --raw) --json | jq -r '.[]'
      targets =
        let
          # TODO: remove once motoko's nixpkgs has: https://github.com/NixOS/nixpkgs/commit/e5ac84e29c4b04ea5934d166f270dd6516ad76c7
          inherit (builtins) isAttrs concatMap attrNames;
          mapAttrsToListRecursiveCond =
            pred: f: set:
            let
              mapRecursive =
                path: value: if isAttrs value && pred path value then recurse path value else [ (f path value) ];
              recurse = path: set: concatMap (name: mapRecursive (path ++ [ name ]) set.${name}) (attrNames set);
            in
            recurse [ ] set;
        in
        mapAttrsToListRecursiveCond
          (path: as: !(pkgs.lib.isDerivation as))
          (path: _value: pkgs.lib.concatStringsSep "." path)
          self.packages.${system};
```
That evaluates pretty quickly for me 1-3s.

flake.nix
```diff
@@ -295,5 +295,25 @@
...
+            let
+              mapRecursive =
+                path: value: if isAttrs value && pred path value then recurse path value else [ (f path value) ];
+              recurse = path: set: concatMap (name: mapRecursive (path ++ [ name ]) set.${name}) (attrNames set);
+            in
+            recurse [ ] set;
+        in
+        mapAttrsToListRecursiveCond
+          (path: as: !(pkgs.lib.isDerivation as))
+          (path: _value: pkgs.lib.concatStringsSep "." path)
```
basvandijk: Nitpicking: since we know the value is a derivation at this point we might as well make that clear:
```suggestion
          (path: _drv: pkgs.lib.concatStringsSep "." path)
```

bin/nixb
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+nix eval --json .#targets.$system | jq -r ".[]" | fzf --bind 'enter:become(echo "Running nix build .#{}" && nix build .#{})'
```
basvandijk: Super nitpicking and purely a matter of taste so ignore if you wish: I like if the `--json` is close to the `jq` parsing like:
```suggestion
nix eval ".#targets.$system" --json | jq -r '.[]' | fzf --bind 'enter:become(echo "Running nix build .#{}" && nix build .#{})'
```
I also quoted the target to eval such that syntax highlighting doesn't get screwed up by the comment `#`.

basvandijk (approved):

PR #5763, opened 2025-12-22
experiment: alternative to align_fields

crusso:
Alternative to #5753 

Seems (suspicioulsy) faster than master:

real	0m25.783s
user	0m25.619s
sys	0m0.163s

real	0m26.191s
user	0m26.041s
sys	0m0.150s

real	0m26.949s
user	0m26.782s
sys	0m0.166s

- [ ] Use better constructor names...


crusso:
Relenting...

PR #5764, opened 2025-12-23
refactor: separate type and value fields in object type representation

christoph-dfinity:
Splits type and value fields in our object type representation and gets rid of the second class `Typ` constructor.

This lets us clean up a few `assert false` and enables changing to a Map rather than sorted list to speed up lookups in the future.

## Review notes

I tried to keep churn to a minimum, but unfortunately it's still a rather large PR. The commit history is messy so best to review the full change at once rather than trying to go commit by commit.

I'm also open to deciding it's not worth the risk of this causing regressions and scrapping the PR.

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -455,6 +455,10 @@ let coverage_cases category env cases t at =
 let coverage_pat warnOrError env pat t =
   coverage' warnOrError "pattern" env Coverage.check_pat pat t pat.at
 
+let coverage_pat_is_exhaustive pat t =
+  let uncovered, _ = Coverage.check_pat pat t in
+  uncovered = []
+
 (* Types *)
 
 let check_ids env kind member ids = Lib.List.iter_pairs
@@ -1949,6 +1953,8 @@ and infer_exp'' env exp : T.typ =
         error_in [Flags.ICMode; Flags.RefMode] env exp1.at "M0077"
           "a shared function is only allowed as a public field of an actor";
     end;
+    if not env.pre && T.is_shared_sort shared_pat.it && Option.is_none typ_opt then
+      warn env exp1.at "M0242" "this declares an implicit oneway function.\n  if this is intentional, annotate explicitly with `: ()`, otherwise annotate with `: async ()`";
     let typ = match typ_opt with
       | Some typ -> typ
       | None -> {it = TupT []; at = no_region; note = T.Pre}
@@ -4506,7 +4512,11 @@ and infer_dec_valdecs env dec : Scope.t =
      let t = infer_exp {env with pre = true; check_unused = false} exp in
      let ve' = match fail with
        | None -> check_pat_exhaustive (if is_import dec then local_error else warn) env t pat
-       | Some _ -> check_pat env t pat
+       | Some _ ->
+          let ve = check_pat env t pat in
+          if not env.pre && coverage_pat_is_exhaustive pat t then
+            warn env pat.at "M0243" "this pattern will always match, so the else clause is useless. Consider removing the else clause";
+          ve
      in
      Scope.{empty with val_env = ve'}
   | VarD (id, exp) ->
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,6 +16,11 @@ on:
         default: test-draft-release
         type: string
 
+permissions:
+  contents: write
+  issues: write
+  pull-requests: write
+
 jobs:
   prepare:
     runs-on: ubuntu-latest
@@ -154,10 +159,17 @@ jobs:
           echo "Contents of release-assets directory:"
           find release-assets -type f -ls || true
 
+      - name: Create GitHub App Token
+        uses: actions/create-github-app-token@v2
+        id: app-token
+        with:
+          app-id: ${{ vars.GENERIC_CI_RW_APP_ID }}
+          private-key: ${{ secrets.GENERIC_CI_RW_APP_PRIVATE_KEY }}
+
       - name: Upload Release Assets
         uses: svenstaro/upload-release-action@v2
         with:
-          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          repo_token: ${{ steps.app-token.outputs.token }}
           tag: ${{ github.event_name == 'workflow_dispatch' && steps.rewrite_draft_assets.outputs.tag || github.ref_name }}
           file: release-assets/*
           file_glob: true
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,11 +1,22 @@
 # Motoko compiler changelog
 
+## 1.1.0 (2026-01-16)
 
 * motoko (`moc`)
 
+  * Warn on unreachable let-else (#5789).
+
+  * bugfix: The source region for `do { ... }` blocks now includes the `do` keyword too (#5785).
+
   * Omit `blob:*` imports from `moc --print-deps` (#5781).
 
+  * Split unused identifier warnings into separate warnings for shared and non-shared contexts: `M0194` for general declarations, `M0240` for identifiers in shared pattern contexts (e.g. `c` in `shared({caller = c})`), `M0198` for unused fields in object patterns, and `M0241` for unused fields in shared patterns (e.g. `caller` in `shared({caller})`) (#5779). 
+
   * Print type constructors using available type paths (#5698).
+
+  * Warn on implicit oneway declarations (#5787).
+
+  * Make the type checker more lenient and continue accumulating typing errors, and try to produce the typed AST even with errors. Enabled only with a type recovery flag for the IDE (Serokell Grant 2 Milestone 3) (#5776).
 
   * Explain subtype failures (#5643).
 
@@ -23,10 +34,6 @@
   ```
 
   * Add privileged primitive for setting Candid type table cutoff (#5642).
-
-  * Split unused identifier warnings into separate warnings for shared and non-shared contexts: `M0194` for general declarations, `M0240` for identifiers in shared pattern contexts (e.g. `c` in `shared({caller = c})`), `M0198` for unused fields in object patterns, and `M0241` for unused fields in shared patterns (e.g. `caller` in `shared({caller})`) (#5779).
-
-  * bugfix: The source region for `do { ... }` blocks now includes the `do` keyword too (#5785).
 
 ## 1.0.0 (2025-12-11)
 
--- a/doc/md/base/Debug.md
+++ b/doc/md/base/Debug.md
@@ -14,7 +14,7 @@ func print(text : Text)
 Prints `text` to output stream.
 
 :::note
-When running on ICP, all output is written to the [canister log](https://internetcomputer.org/docs/building-apps/canister-management/logs) with the exclusion of any output
+When running on ICP, all output is written to the [canister log](https://internetcomputer.org/docs/current/developer-docs/smart-contracts/maintain/logs) with the exclusion of any output
 produced during the execution of non-replicated queries and composite queries.
 In other environments, like the interpreter and stand-alone Wasm engines, the output is written to standard out.
 :::
--- a/doc/md/base/ExperimentalInternetComputer.md
+++ b/doc/md/base/ExperimentalInternetComputer.md
@@ -31,7 +31,7 @@ let rawReply = await IC.call(ledger, method, to_candid(input)); // serialized Ca
 let output : ?OutputType = from_candid(rawReply); // { decimals = 8 }
 ```
 
-[Learn more about Candid serialization](../16-language-manual.md#candid-serialization)
+[Learn more about Candid serialization](https://internetcomputer.org/docs/current/motoko/main/reference/language-manual#candid-serialization)
 
 ## Function `isReplicated`
 ``` motoko no-repl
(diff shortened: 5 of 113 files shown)


src/mo_frontend/typing.ml
```diff
@@ -562,21 +559,22 @@ and check_obj_path' env path : T.typ =
...
+        display_obj (T.Obj(s, fs, tfs))
         (Suggest.suggest_id "field" id.it
-          (List.map (fun f -> f.T.lab) (T.val_fields fs)))
+          (List.map (fun f -> f.T.lab) fs))
 
 let rec check_typ_path env path : T.con =
   let c = check_typ_path' env path in
-  path.note <- T.Typ c;
+  (* TODO? *)
+  (* path.note <- T.Typ c; *)
```
christoph-dfinity: What should I do here? Can't annotate with a `con`

src/mo_types/type.ml
```diff
@@ -1677,8 +1696,19 @@ let rec combine rel lubs glbs t1 t2 =
...
 
+and combine_typ_fields rel lubs glbs tfs1 tfs2 =
+  align_fields tfs1 tfs2
+  |> Seq.fold_left (fun acc x -> match x with
+      | Lib.Both(f1, f2) ->
+        if eq_con f1.typ f2.typ then f1 :: acc else
+        if rel == glbs then raise Mismatch else acc
+      | Lib.This f1 -> cons_if_typ (rel == glbs) f1 acc
+      | Lib.That f2 -> cons_if_typ (rel == glbs) f2 acc) []
+  |> List.rev
```
Kamirus, line 1710: note to self: check later

src/mo_types/type.ml
```diff
@@ -2033,25 +2063,32 @@ and can_omit n t =
...
-  | (s, fs) ->
+      (pp_print_list ~pp_sep:semi (pp_print_either ~left:(pp_field vs) ~right:(pp_typ_field vs))) fs
+  | (s, fs, tfs) ->
+    let fields = List.map (fun f -> Either.Left f) fs in
+    let typ_fields = List.map (fun f -> Either.Right f) tfs in
+    let fs = typ_fields @ fields in
     fprintf ppf "@[<hv 2>%s{@;<0 0>%a@;<0 -2>}@]"
       (string_of_obj_sort s)
-      (pp_print_list ~pp_sep:semi (pp_field vs)) fs
+      (pp_print_list ~pp_sep:semi (pp_print_either ~left:(pp_field vs) ~right:(pp_typ_field vs))) fs
```
Kamirus, line 2091: not possible to `pp_print_list` one list and then the other? :D need to go through either?

christoph-dfinity: felt like more of a headache to figure out the optional/trailing separators when either of the lists are empty...

src/mo_types/type.ml
```diff
@@ -2257,6 +2296,7 @@ and pp_kind ppf k =
   let vs = vs_of_cs cs in
   pp_kind' vs ppf k
 
+(* TODO: I don't know what this means *)
```
Kamirus: remove?

src/mo_types/typ_hash.ml
```diff
@@ -108,8 +108,8 @@ let rec go = function
      this is useful for stable serialization *)
   | Mut t -> ((Unary, "M"), [t])
 
-  | Obj (s, fs) ->
-    ( ( Labeled (List.map (fun f -> f.lab ^ if is_mut f.typ then "!" else "") fs),
+  | Obj (s, fs, tfs) ->
+    ( ( Labeled (List.map (fun f -> f.lab ^ if is_mut f.typ then "!" else "") fs @ List.map (fun f -> f.lab) tfs),
```
Kamirus: Does the order matter here? Types should be added last?

src/mo_idl/idl_to_mo_value.ml
```diff
@@ -63,7 +63,7 @@ let rec value v t =
   | BlobV b, T.Array _ ->
     brackets_comma (List.of_seq (Seq.map (fun c -> Printf.sprintf "%d" (Char.code c)) (String.to_seq b)))
   | TextV s, _ -> text_lit s
-  | RecordV fs, T.(Obj (Object, tfs)) ->
+  | RecordV fs, T.(Obj (Object, tfs, [])) ->
```
Kamirus: so the type fields are not allowed?

src/lowering/desugar.ml
```diff
@@ -401,9 +401,9 @@ and build_field {T.lab; T.typ;_} =
 
 and build_fields obj_typ =
     match obj_typ with
-    | T.Obj (_, fields) ->
+    | T.Obj (_, fields, _) ->
       (* TBR: do we need to sort val_fields?*)
-      List.map build_field (T.val_fields fields)
+      List.map build_field fields
```
ggreif, line 406: the comment above now has a dangling reference

src/lowering/desugar.ml
```diff
@@ -507,7 +506,7 @@ and export_footprint self_id expr =
   let scope_con2 = Cons.fresh "T2" (Abs ([], Any)) in
   let bind1  = typ_arg scope_con1 Scope scope_bound in
   let bind2 = typ_arg scope_con2 Scope scope_bound in
-  let ret_typ = T.Obj(Object,[{lab = "size"; typ = T.nat64; src = empty_src}]) in
+  let ret_typ = T.obj T.Object [("size", T.nat64)] in
```
```suggestion
  let ret_typ = T.(obj Object [("size", nat64)]) in
```

src/mo_frontend/bi_match.ml
```diff
@@ -382,6 +381,18 @@ let bi_match_typs ctx =
...
+    (* Assume that tfs1 and tfs2 are sorted. *)
+    align_fields tfs1 tfs2 |>
+    Seq.fold_left (fun inst fs -> match inst with
+      | None -> None
+      | Some inst -> match fs with
+        | Lib.Both(tf1, tf2) ->
+          (* NB: we assume c1, c2 closed *)
+          if eq_con tf1.typ tf2.typ then Some inst else None
+        | Lib.This(_) -> if rel != eq then Some inst else None
+        | Lib.That(_) -> None) (Some inst)
```
ggreif:
```suggestion
        | Lib.This _ -> if rel != eq then Some inst else None
        | Lib.That _ -> None) (Some inst)
```

src/mo_types/typ_hash.ml
ggreif: If this is a hashing function then we want non-commutativity, so probably the list should be merged to produce one sorted list.

ggreif (commented):
Looks good, but I am ambivalent about what this buy us. The hashing function definitely needs another look (as @Kamirus spotted)

src/mo_frontend/typing.ml
```diff
@@ -901,7 +899,7 @@ and check_typ_def env at (id, typ_binds, typ) : T.kind =
   check_closed env id k at;
   k
 
-and check_typ_field env s typ_field : T.field = match typ_field.it with
+and check_typ_field env s typ_field : (T.field, T.typ_field) Either.t = match typ_field.it with
```
Kamirus, line 905: Syntax.typ_field could also be refactored (split into value and type fields)? Ofc not in this PR

src/mo_frontend/typing.ml
```diff
@@ -2315,73 +2293,51 @@ and infer_check_bases_fields env (check_fields : T.field list) exp_at exp_bases
...
-  let ambiguous_fields ft1 ft2 =
-    homonymous_fields ft1 ft2 &&
-    (* allow equivalent type fields *)
-    match ft1.T.typ, ft2.T.typ with
-    (* homonymous type fields are ambiguous when unequal *)
-    | T.Typ c1, T.Typ c2 ->  not (eq env exp_at ft1.T.typ ft2.T.typ)
-    (* homonymous value fields are always ambiguous *)
-    | _ -> true
-  in
+  let field_map = Hashtbl.create 0 in
```
Kamirus, line 2317: Some brand new code below: But looks good, cleaner. Worth taking another look from someone else though

src/mo_frontend/typing.ml
```diff
@@ -3375,7 +3331,8 @@ and check_pat_aux' env t pat val_kind : Scope.val_env =
     in check_pats env ts pats T.Env.empty pat.at
   | ObjP pfs ->
     let pfs' = List.stable_sort compare_pat_field pfs in
-    let s, tfs =
+    let s, fs =
+      (* TODO: Why not get the type fields out as well? *)
```
Kamirus: TODO here. I mean currently we return Scope.val_env here, so there is no space for types. Not really necessary to change that

src/mo_frontend/typing.ml
```diff
@@ -3556,35 +3510,37 @@ and check_pats_typ_dec env ts pats te at : Scope.typ_env =
   in
   go ts pats te
 
-and check_pat_fields_typ_dec env t tfs pfs te at : Scope.typ_env =
-  let cmp (tf : T.field) pf = match tf, pf with
-    | T.{ lab; typ = T.Typ _; _}, { it = TypPF(id); _ } -> String.compare lab id.it
-    | _, { it = TypPF(id); _ } -> 1
-    | T.{ lab; _ }, { it = ValPF(id, _); _ } -> String.compare lab id.it in
+and check_pat_fields_typ_dec env t fs tfs pfs te at : Scope.typ_env =
```
Kamirus, line 3517: New code also here: Another pair of eyes is welcome :D
IMO looks good

src/mo_frontend/typing.ml
```diff
@@ -4737,7 +4688,8 @@ let check_stab_sig scope sig_ : T.stab_sig  Diag.result =
                    match field.it with TypF (id, _, _) -> Some id | _ -> None)
                  sfs);
             let _ = List.map (check_typ_field {env1 with pre = true} T.Object) sfs in
-            let fs = List.map (check_typ_field {env1 with pre = false} T.Object) sfs in
+            (* TODO: Is it correct to drop type fields here? *)
```
Kamirus: TODO here, good question to Claudio probably. I guess that's ok since we only migrate value fields in migration functions.
Does it mean the `check_ids env "object type" "type field"` is dead code above?

src/mo_types/type.ml
Kamirus: Ok that's correct. You've inlined the `combine` from below in `combine_fields`.
And indeed the exception is not raised when `rel == lubs` which makes sense

src/mo_frontend/typing.ml
Kamirus: You can update the path's note type to `Type.con option` (option so that None is used as Type.Pre)
```ocaml
and path = (path', Type.typ) Source.annotated_phrase
```
Change to
```ocaml
and path = (path', Type.con option) Source.annotated_phrase
```
Nothing broke in our CI pipeline but maybe this is used by the vscode extension? Dunno

src/mo_types/type.ml
```diff
@@ -2033,25 +2063,32 @@ and can_omit n t =
...
     | Named (n, t) -> go i t
   in go n t
 
 let rec pp_typ_obj vs ppf o =
   match o with
-  | (Object, fs) ->
+  | (Object, fs, tfs) ->
+    let fields = List.map (fun f -> Either.Left f) fs in
+    let typ_fields = List.map (fun f -> Either.Right f) tfs in
+    let fs = typ_fields @ fields in
```
crusso, line 2082: I was wondering why the output hadn't changed...

src/mo_frontend/typing.ml
crusso: I think the way to fix this to alter Syntax.ml to have (recursive) path and  (non-recursive) typ_path, with different notes.

crusso: Or use an either type for the annotation (Type or Con) (maybe simplest).

crusso: Simplest might be to change the note to 'unit' and see if anything was using it

crusso: Simplest might be to change the note to 'unit' and see if anything was using it

crusso: Simplest might be to change the note to 'unit' and see if anything was using it

src/ir_def/check_ir.ml
```diff
@@ -299,10 +296,6 @@ and check_con env c =
 and check_typ_field env s tf : unit =
   match tf.T.typ, s with
   | T.Mut t, Some (T.Object | T.Memory) -> check_typ env t
-  | T.Typ c, Some _ ->
```
crusso, line 302: I think you've lost this check. Maybe move to line line 260 (T.Obj case)

src/mo_types/typ_hash.ml
christoph-dfinity: Yes, this is not great for a hashing function `{ type t = ... }` hashes to the same hash as `{ t : ... }`, but I'm just trying to preserve the behaviour. 

I did swap the order around accidentally, because previously type fields were always "smaller" than value fields, so they appeared first. I swapped it back, thanks for the catch!

src/mo_frontend/typing.ml
christoph-dfinity: 2aed09bb529d490e47592a4cb77cb7330d17a723 <- went for the first suggestion

christoph-dfinity: 2aed09bb529d490e47592a4cb77cb7330d17a723

src/mo_idl/idl_to_mo_value.ml
christoph-dfinity: I stared at it for a while, and I think this change was wrong. The only thing we do is look up value fields, so we don't care if there are type fields or not. I changed the code accordingly

src/mo_frontend/typing.ml
christoph-dfinity: Yes, I checked and it actually is dead code.

PR #5770, opened 2025-12-29
bugfix:  issue #5766

### LINKED ISSUE #5766: Bug: var function of a type fails to be called.
The code:

```motoko
module {
  public func f() {
    type Callbacks = {
      var call : Nat -> ();
    };
    let callbacks : Callbacks = {
      var call = func(_) {};
    };
    callbacks.call(1);
  };
};
```

receives an error:

```
 FAIL src/main.mo:9:15: type error [M0234], field call does exist in type:
  {var call : Nat -> ()}
but is not a function.

Did you mean field call?
   ...
   8	|     };
-> 9	|     callbacks.call(1);
   10	|   };
   ...
```

mops.toml:

```
[toolchain]
moc = "1.0.0"
```

However this worked in previous versions.

crusso:
Fixes #5766 by allowing also allowing mutable fields on lhs of applications.

### DIFF
--- a/src/js/common.ml
+++ b/src/js/common.ml
@@ -247,13 +247,14 @@ let js_parse_motoko_typed_with_scope_cache_impl enable_recovery paths scope_cach
            all. Hence, the use of [Obj.magic] is legitimate here. *)
         String_map_conversion.from_js scope_cache Js.to_string Obj.magic)
   in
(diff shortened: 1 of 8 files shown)


christoph-dfinity (approved):
LGTM

PR #5775, opened 2026-01-02
bugfix: remove SKIP_RUNNING=yes (a temporary hack?)

crusso:
make in test/run-drun was only executing the first test on drun, and skipping most other runs.

I think this is the culprit line, that appeared between 0.16.1 and 0.16.2, but I'm not sure why it was introduced. 

https://github.com/caffeinelabs/motoko/compare/0.16.1...0.16.2

Slack thread:

https://caffeine-ai.slack.com/archives/C09RAD20U83/p1766507061355159

The line appears to have been added here (@alexandru-uta): https://github.com/caffeinelabs/motoko/commit/6443756e5e692dc0951a593e2b1ba8e089d55dff#diff-608b6425c35171211ed108a6c30a5121be49f5f8f7bfdab4df44414377f26dd0R308

(as part of migration to pocket-ic)

I hate bash.

(let's hope this isn't the reason our CI is so much faster)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,7 +1,10 @@
 # Motoko compiler changelog
 
+
 * motoko (`moc`)
 
+  * Explain subtype failures (#5643).
+  
   * Removes the Viper support (#5751).
 
   * Allows resolving local definitions for context-dot (#5731).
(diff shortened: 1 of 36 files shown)


alexandru-uta (approved):
Thanks for finding this!

(I have probably added that for quick debugging mode but then forgot about it and checked it in..)

alexandru-uta:
> (let's hope this isn't the reason our CI is so much faster)

it's not; running make manually is different to what the CI does.

PR #5779, opened 2026-01-07
feat: Separate unused identifier warning for shared pattern

Kamirus:
Split unused identifier warnings into separate warnings for shared and non-shared contexts: `M0194` for general declarations, `M0240` for identifiers in shared pattern contexts (e.g. `c` in `shared({caller = c})`), `M0198` for unused fields in object patterns, and `M0241` for unused fields in shared patterns (e.g. `caller` in `shared({caller})`)

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -336,16 +336,11 @@ let _warn_in modes env at code fmt =
 (* Unused identifier detection *)
 
 let emit_unused_warnings env =
-  let is_in_shared_pat region =
-    let is_subregion sub super =
-      Source.Pos_ord.compare sub.left super.left >= 0 &&
-      Source.Pos_ord.compare sub.right super.right <= 0 &&
-      sub.left.file = super.left.file
-    in
-    List.exists (is_subregion region) !(env.shared_pat_regions)
+  let is_in_shared_pat pos = !(env.shared_pat_regions) |> List.exists Source.Pos_ord.(fun region ->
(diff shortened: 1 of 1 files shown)


src/mo_frontend/typing.ml
```diff
@@ -3283,34 +3289,34 @@ and check_class_shared_pat env shared_pat obj_sort : Scope.val_env =
...
     if mode = T.Query then
       error env shared_pat.at "M0109" "class cannot be a query";
-    check_pat_exhaustive local_error env T.ctxt pat
+    check_pat_exhaustive local_error env T.ctxt pat true
   | _, T.Memory -> assert false
 
 
-and check_pat_exhaustive warnOrError env t pat : Scope.val_env =
-  let ve = check_pat env t pat in
+and check_pat_exhaustive warnOrError env t pat shared: Scope.val_env =
```
christoph-dfinity: Let's maybe move the bool into `env` here. We've already got `in_actor` and `in_prog`, so `in_shared_pat` could fit there.

src/mo_frontend/scope.mli
```diff
@@ -1,7 +1,7 @@
 open Mo_types.Type
 
 type val_kind = Declaration | FieldReference
-type val_env = (typ * Source.region * val_kind) Env.t
+type val_env = (typ * Source.region * val_kind * bool) Env.t  (* bool = shared *)
```
christoph-dfinity: The value isn't `shared` it's "bound in a `shared_pat`".

PR #5785, opened 2026-01-13
fix: `do { ... }` block region to include the `do` keyword

Kamirus:
Fix the region for `do { ... }` blocks to include the `do` keyword too.
Using the `moc.js` I've found out that the `BlockE` expression has incorrect region, this is a fix for it.


Changelog.md
```diff
@@ -3,6 +3,8 @@
 
 * motoko (`moc`)
 
+  * Fix the region for `do { ... }` blocks to include the `do` keyword too (#5785).
```
ggreif: this should be at the bottom with "bugfix:" (like elsewhere)

ggreif (approved):
LGTM!

PR #5789, opened 2026-01-15
fix: warn on unreachable else in let-else

christoph-dfinity:


test/fail/ok/let-else-definedness.tc.ok
```diff
@@ -1 +1,2 @@
+let-else-definedness.mo:1.5-1.6: warning [M0243], this pattern will always match, so the else clause is useless. Consider removing the else clause
```
Kamirus, line 1: Can you fix the region to cover the whole `else { ... }` clause with the expression?
Right now it covers the variable

christoph-dfinity: I'm intentionally pointing at the pattern, because that's the first thing the error references. I know you want the else range for auto-fixing (FWIW I don't have the range for the `else` token in the compiler, because it's not part of the AST)

Kamirus: Maybe that is fine to point to the pattern, we already do that for `warning [M0146], this pattern is never matched`

PR #5790, opened 2026-01-15
fix: report multiple type errors for compound types at once

christoph-dfinity:
- fix: report multiple type errors per object type at once
- fix: report multiple type errors per tuple type
- fix: report multiple type errors for and/or types at once
- fix: report multiple type errors for function types at once

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,10 @@
 
 * motoko (`moc`)
 
+  * Report multiple type errors for compound types at once (#5790).
+
+    This means code like ``` type T = (Na, In)``` will fail with errors for both the misspelled `Na` and `In` types at once, so they can be fixed in one go, rather than having to re-run the compiler after fixing the first one.
+
   * Allow `break` and `continue` in loops without labels (#5702).
   * Report a better error for labeled `continue` targeting a non-loop (#5800).
   * Deprecate older garbage collectors: generational, copying and compating GCs (#5806).


src/mo_frontend/typing.ml
```diff
@@ -806,14 +806,18 @@ and check_typ' env typ : T.typ =
     let t = check_typ env typ in
     T.Array (infer_mut mut t)
   | TupT typ_items ->
-    T.Tup (List.map (check_typ_item env) typ_items)
+     let ts = List.filter_map (fun i -> try Some(check_typ_item env i) with Recover -> None) typ_items in
+     if List.length ts <> List.length typ_items then raise Recover;
+     T.Tup ts
```
Kamirus: I like this simple approach.
I'd just hide all these recovery details like:

```suggestion
    T.Tup (check_all (check_typ_item env) typ_items)
```
and I'd put this `check_all` function next to the recover functions.
I'm not sure about the name, maybe require_all/recover_all/traverse_all?

src/mo_frontend/typing.ml
```diff
@@ -860,11 +864,15 @@ and check_typ' env typ : T.typ =
...
+    if List.length fs <> List.length fields then raise Recover;
     T.Obj (sort.it, List.sort T.compare_field fs)
   | AndT (typ1, typ2) ->
-    let t1 = check_typ env typ1 in
-    let t2 = check_typ env typ2 in
+    let t1 = try Some(check_typ env typ1) with Recover -> None in
+    let t2 = try Some(check_typ env typ2) with Recover -> None in
+    let t1, t2 = match t1, t2 with
+      | Some t1, Some t2 -> t1, t2
+      | _ -> raise Recover in
```
Kamirus: Similar here: I'd use sth like
`let t1, t2 = check_both (check_typ env) typ1 typ2 in`

PR #5796, opened 2026-01-21
feat: generic data inspection via auto-generated, per stable variable view queries

crusso:
# Data inspection support

TODO:

- [x] Generate trivial queries for non-viewable data (don't just omit).
- [x] Or, cast to most precise shared supertype (this is constant time).

Adds `moc` flag `--generate-view-queries`.

With the flag enabled, `moc` will auto-generate one `__var` query method per stable variable `var` of type `<typ>`, provided:

* `__var` is not already declared; and
* `__var` does not have reserved prefix `__motoko_`; and
*  Either:
   1.   `var.view()` resolves to a function with shared argument `<typ1,...,typN>` and result `<ret-typ>`,
        generating:
         ```motoko
         public shared {caller} query func __var (arg1:<typ1>, ..., argN: <typN>) -> async <ret-typ> { 
           <access-control>;
           var.view()(arg1,...,argn)
         }
         ```
        Note that `var.view` can take implicit arguments (such as `compare`) provided they resolve the call.
    3. Or, if the type of var is `shared` (it may just be `stable` but not `shared`), generates the simple query
        ```motoko
          public shared query func __var() : async <typ> { 
            <access-control>;
            var
          };
       ```

The `.view()` method patterns accommodate both implicit local and additional external arguments and are intended to support the implementation of logical views of concrete, non-shared data structures (such as core `Map`s) using shared values with user-defined controls for, e.g.,  pagination.
For example, instead of reporting the raw data structure of a b-tree based map, we would display the logical content of the map, keys and values, in a paginated form.

# User defined view methods

As an example, here we provide a `.view` method for core maps with keys and values.

``` motoko
module MapView {
   public func view<K,V>(self : Map.Map<K, V>, compare : (implicit : (K,K) -> Order.Order)) :
      (ko : ?K, count : ?Nat) -> [(K, V)] =
      func (ko, count) {
        let entries = switch ko {
      	  case null {
            self.entries()
          };
          case (?k) {
          self.entriesFrom(k)
          };
        };
        switch count {
          case null { entries.toArray() };
          case (?c) { entries.take(c).toArray() };
        };
     }
  };
```

For example, for stable variable:

``` motoko
let customers : Map.Map<Text, Customer>;
```
The view above would generate a query of Motoko type:

``` motoko 
__customers: query (ko: ?Text, count: ?Nat) ->
	  async [(Text, Customer)]
     };
```
and Candid type:

``` candid
  __customers: (ko: opt text, count: opt nat) ->
   (vec
     record {
       text;
       Customer
     }) query;
```
# Candid

View queries are excluded from the custom section describing the (public) interface of a canister. This means that a new version of a canister will not have to provide view methods compatible with the previous version. Doing so would leak stable variable implementation details into the public interface of a canister.

Although a canister's public interface does not contain the generated queries, the `moc --idl` generated local interface *does* include them. This means that the local .did file can used in generic front-end display code, that parses any candid representation to present a type driven user interface. See the demo for an example, generic front-end UI.

# Access control

The injected `<access-control>` code limits access to any `caller` that is
1. a _controller_; or
2. _self_; 

## Possible extensions (implemented but removed for now)

3. a principal that satisfies the user-defined function `isAdmin(caller):Bool` (provided local function `isAdmin` is available in the actor or mixin and of the appropriate type `Principal -> Bool`).~
The optional `isAdmin` function lets users integrate with app specific notions of access control, if desired.

# Demo

App: https://fviwb-zqaaa-aaaaj-a5d5q-cai.icp0.io/
Code: https://github.com/crusso/motoko-stable-viewer (requires some setup)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -24,6 +24,10 @@
 
   * perf: type-based optimization of option creation and consumption, reducing cycle cost (#5947).
 
+* documentation (`mo-doc`)
+
+  * feat: doc comments on individual record fields and variant tags inside a `type` declaration are now extracted and rendered (#5983).
+
 ## 1.4.1 (2026-03-30)
 
 * motoko (`moc`)
--- a/doc/docusaurus/.npmrc
+++ b/doc/docusaurus/.npmrc
@@ -1,0 +1,1 @@
+legacy-peer-deps=true
--- a/doc/docusaurus/docusaurus.config.js
+++ b/doc/docusaurus/docusaurus.config.js
@@ -1,8 +1,9 @@
 // @ts-check
 // Note: type annotations allow type checking and IDEs autocompletion
 
-const lightCodeTheme = require('prism-react-renderer/themes/github');
-const darkCodeTheme = require('prism-react-renderer/themes/dracula');
+const {themes} = require('prism-react-renderer');
+const lightCodeTheme = themes.github;
+const darkCodeTheme = themes.dracula;
 
 /** @type {import('@docusaurus/types').Config} */
 const config = {
@@ -10,7 +11,7 @@ const config = {
   tagline: '(a work in progress)',
   url: 'https://your-docusaurus-test-site.com',
   baseUrl: '/',
-  onBrokenLinks: 'throw',
+  onBrokenLinks: 'warn',
   onBrokenMarkdownLinks: 'warn',
   favicon: 'img/motoko.png',
 
@@ -22,6 +23,10 @@ const config = {
   plugins: [
     'docusaurus-plugin-sass'
   ],
+
+  markdown: {
+    format: 'detect',
+  },
 
   // Even if you don't use internalization, you can use this field to set useful
   // metadata like html lang. For example, if your site is Chinese, you may want
--- a/doc/docusaurus/package-lock.json
+++ b/doc/docusaurus/package-lock.json
@@ -12,19 +12,20 @@
         "@docusaurus/core": "3.8.1",
         "@docusaurus/preset-classic": "3.8.1",
         "@docusaurus/theme-live-codeblock": "^3.8.1",
-        "@mdx-js/react": "^1.6.22",
-        "clsx": "^1.2.0",
+        "@mdx-js/react": "^3.0.0",
+        "clsx": "^2.0.0",
         "docusaurus-plugin-sass": "^0.2.2",
         "highlight.js": "^11.5.1",
-        "prism-react-renderer": "^1.3.5",
-        "react": "^16.14.0",
+        "prism-react-renderer": "^2.1.0",
+        "react": "^18.0.0",
         "react-chartjs-2": "^4.1.0",
         "react-codejar": "^1.1.2",
-        "react-dom": "^16.14.0",
-        "remark-code-import": "^0.4.0"
+        "react-dom": "^18.0.0",
+        "remark-code-import": "^0.4.0",
+        "sass": "^1.99.0"
       },
       "devDependencies": {
-        "@docusaurus/module-type-aliases": "2.0.0-beta.22"
+        "@docusaurus/module-type-aliases": "3.8.1"
       },
       "engines": {
         "node": ">=16.14"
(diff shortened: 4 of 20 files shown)


test/test-moc.js
```diff
@@ -192,6 +192,7 @@ type T = nat;
 service : {
   /// Function comment
   main: () -> (T) query;
+  x: () -> (T) query;
```
crusso:
```suggestion
```

src/mo_frontend/typing.ml
```diff
@@ -4144,6 +4154,54 @@ and check_stab env sort scope dec_fields =
...
+    assert (!viewer = None);
+    let lab = "__" ^ id.it in
+    if T.Env.mem lab scope.Scope.val_env || String.starts_with ~prefix:"__motoko" lab
+    then () (* avoid any clash with local or reserved `__motoko_XXX` members by omitting viewer *)
+    else
+      let viewer_field args ret =
+        T.{ lab; typ = Func (Shared Query, Promises, [scope_bind], args, ret); src = empty_src } in
+      let infer_dot_view =
+        Diag.with_message_store (recover_opt (fun msgs ->
+          let env = {env with msgs; used_identifiers = ref T.Env.empty } in (* don't record errors in outer env *)
```
christoph-dfinity: Could this cause unused warnings for `view` functions that are defined inline in the actor (not in a mixin/library)? They'd be resolved by the context-dot, but not marked as used afterwards.

src/lowering/desugar.ml
```diff
@@ -839,7 +894,7 @@ and stabilize stab_opt d =
   match s, d.it with
   | (S.Flexible, _) ->
     ([], fun _ -> d)
-  | (S.Stable, I.VarD(i, t, e)) ->
+  | (S.Stable viewer, I.VarD(i, t, e)) ->
```
christoph-dfinity:
```suggestion
  | (S.Stable _, I.VarD(i, t, e)) ->
```

(and do the same in the other branches that don't use the viewer)

src/mo_frontend/typing.ml
```diff
@@ -4144,6 +4154,54 @@ and check_stab env sort scope dec_fields =
...
+          match T.normalize viewer_typ with
+           | T.Func(T.Local, T.Returns, [], ts1, ts2)
+             when List.for_all T.shared ts1 && List.for_all T.shared ts2 ->
+              { viewer_body = DotViewV exp;
+                viewer_field = viewer_field ts1 ts2 }
+           | _ -> raise Recover))
+      in
+      match infer_dot_view with
+      | Ok (exp_typ, _) ->
+         (* info env at "viewer found for %s" id.it; *)
```
christoph-dfinity: remove?

src/mo_frontend/typing.ml
crusso: Good point. Do we care though?

crusso: Maybe the way to do this is rollback if resolution fails, and commit if it succeeds.
I was mostly worried about treating things as used that weren't used afterall. 
I'll try to fix.

src/lowering/desugar.ml
crusso: Done

src/mo_frontend/typing.ml
crusso: Done PTAL. For consistency, I consider any vars mentioned in generated queries used, although that might be weird from a users perspective.

christoph-dfinity: Thanks, that looks good!

src/mo_frontend/typing.ml
crusso: Done

doc/md/icp-features/7-view-queries.md
```diff
@@ -0,0 +1,173 @@
...
+
+```
+moc --generate-view-queries ...
+```
+
+The flag is **off by default**. When enabled, the compiler examines every stable variable in the actor and, where possible, emits a corresponding query function.
+
+## How it works
+
+For each stable variable `var` of type `<typ>`, the compiler attempts to generate a public query method named `__var`. Generation is skipped if:
```
christoph-dfinity: Use a different name that doesn't overlap with the keyword. `let`s get view methods as well.

doc/md/icp-features/7-view-queries.md
```diff
@@ -0,0 +1,173 @@
...
+```motoko
+public shared ({ caller }) query func __var() : async <typ> {
+  <access-control>;
+  var
+};
+```
+
+### 3. No generation
+
+If neither condition is met (no `.view()` and the type is not shared), no query is generated for that variable.
```
christoph-dfinity: Soon to be wrong?

doc/md/icp-features/7-view-queries.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Access control
+
+Every generated query enforces that the `caller` must be either:
+
+1. The canister **itself** (self-calls), or
```
christoph-dfinity, line 56: Why do we care about allowing `self` calls here? Is that for testing?

doc/md/icp-features/7-view-queries.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Signature pattern
+
+```motoko
+module MyView {
+  public func view<...>(self : <DataType>, ...) : (arg1 : T1, ..., argN : TN) -> R = ...
+}
+```
+
+Implicit arguments (like `compare` for ordered collections) are supported and resolved automatically by the compiler.
```
christoph-dfinity, line 78: Stop waffling

doc/md/icp-features/7-view-queries.md
crusso: yeah, I'm just waiting for you guys to decide on your preference.

doc/md/icp-features/7-view-queries.md
crusso: Largely yes. And its consistent with how we protect other things.

doc/md/icp-features/7-view-queries.md
crusso: Why is that waffling?

crusso: Why is that waffling?

doc/md/icp-features/7-view-queries.md
crusso: Changed to `id`

doc/md/icp-features/7-view-queries.md
christoph-dfinity: I think the "How it works" section describes the same thing with different words.

But it's fine, I'll approve

doc/md/icp-features/7-view-queries.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+The actor above exposes the following generated queries:
+
+| Stable variable        | Strategy        | Generated query                                          |
+|-------------------------|-----------------|----------------------------------------------------------|
+| `array`                 | `.view()`       | `__array(start : Nat, count : Nat) : async [(Nat, Text)]` |
+| `some_variant`          | shared fallback | `__some_variant() : async Tree`                           |
+| `some_record`           | shared fallback | `__some_record() : async {a : Nat; b : Text; c : Bool}`  |
+| `some_mutable_record`   | *(skipped)*     | *(none — type is not shared and no `.view()` available)*  |
```
crusso:
```suggestion
| `some_mutable_record`   | approximated  |  `__some_mutable_ record() : async Any`  |
```

doc/md/icp-features/7-view-queries.md
crusso: Going for just `Any`

doc/md/icp-features/7-view-queries.md
```diff
@@ -0,0 +1,180 @@
...
+```motoko
+public shared ({ caller }) query func __id() : async <typ> {
+  <access-control>;
+  id
+};
+```
+
+### 3. No generation
+
+If neither condition is met (no `.view()` and the type is not shared), no query is generated for that variable.
```
crusso:
```suggestion
### 3. Approximation to `Any`

If neither condition is met (no `.view()` and the type is not shared), the query just returns a value of the non-informative type `Any`.
```

doc/md/icp-features/7-view-queries.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+The actor above exposes the following generated queries:
+
+| Stable variable        | Strategy        | Generated query                                          |
+|-------------------------|-----------------|----------------------------------------------------------|
+| `array`                 | `.view()`       | `__array(start : Nat, count : Nat) : async [(Nat, Text)]` |
+| `some_variant`          | shared fallback | `__some_variant() : async Tree`                           |
+| `some_record`           | shared fallback | `__some_record() : async {a : Nat; b : Text; c : Bool}`  |
+| `some_mutable_record`   | approximated  |  `__some_mutable_ record() : async Any`  |
```
crusso:
```suggestion
| `some_mutable_record`   | approximated  |  `__some_mutable_record() : async Any`  |
```

PR #5797, opened 2026-01-21
feat: Contextual dot resolution in moc.js

Kamirus:
`moc.js` new exported functions:
- `contextualDotSuggestions` returns available context dot functions given a receiver type + scope
- `contextualDotModule` returns the string module reference for a given context-dot `DotE` expression

The moc.js AST nodes store now references to actual raw `exp`. These JS properties are not enumerable to avoid accidentally accessing them e.g. via `JSON.stringify`.
Also, the root node stores the scope with all transitive imports processed.
The idea behind it is to reuse the compiler functionality, e.g. context dot resolution, and not reimplement that in the vscode extension.

**Limitations**: For the context dot resolution only the imports are considered. These methods above don't have the access to the local scope of each expression, yet!

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "candid-src": {
       "flake": false,
       "locked": {
-        "lastModified": 1770135221,
-        "narHash": "sha256-7/5G25Ms2JrFGpzubtRwPJBAHAExCibw0PZ+ict4rHU=",
+        "lastModified": 1770754485,
+        "narHash": "sha256-wDzcJlOV+h2g7hGAEv5WdDEj8I31dDkwVptWlK+AgvQ=",
         "owner": "dfinity",
         "repo": "candid",
-        "rev": "022ec7e7f9b8061ed9b8342113638ad9b39e8c01",
+        "rev": "cb5e687e9b93ca1f78494311be1111e79c56cd81",
         "type": "github"
       },
       "original": {
--- a/src/lang_utils/diag.ml
+++ b/src/lang_utils/diag.ml
@@ -118,7 +118,7 @@ let print_message msg =
   if msg.sev <> Error && not !Flags.print_warnings
   then ()
   else match !Flags.error_format with
-  | Flags.Classic -> Printf.eprintf "%s%!" (string_of_message msg)
+  | Flags.Plain -> Printf.eprintf "%s%!" (string_of_message msg)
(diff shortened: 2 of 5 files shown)


src/js/common.ml
```diff
@@ -291,11 +293,12 @@ let js_parse_motoko_typed_with_scope_cache enable_recovery paths scope_cache =
   in
   js_result result (fun (progs, scope_cache) ->
     let progs =
-      progs |> Array.map (fun (ast, immediate_imports) ->
+      progs |> Array.map (fun (ast, immediate_imports, senv) ->
         object%js
           val ast = ast
           (* val typ = typ *)
           val immediateImports = immediate_imports
+          val scope = Js.Unsafe.inject senv
```
Kamirus, line 312: Notice that `senv` is the same for each program. We could just pass the `senv` as yet another result of `parseMotokoTyped` and `parseMotokoTypedWithScopeCache` but it would require breaking changes so it is easier to just add another field to each program root node

src/mo_frontend/typing.ml
```diff
@@ -1379,13 +1379,13 @@ type hole_candidate =
   { path: exp;
     desc: string;
     typ : T.typ;
-    module_name_opt: string option;
+    module_ref_opt: T.lab option; (* module name (from `vals`) or path (from `libs`) *)
```
Kamirus: I found out that `module_name` is actually misleading because this string is sometimes an actual module name e.g. `Map` and sometimes a full file path.
So I renamed all to `module_ref` and left comments.

src/js/astjs.ml
```diff
@@ -295,7 +304,7 @@ module Make (Cfg : Config) = struct
   let rec exp_js e =
     let open Syntax in
     let open Source in
-    exp'_js e |> add_type_annotation e.note.note_typ |> add_source e.at
+    exp'_js e |> add_raw_exp e |> add_type_annotation e.note.note_typ |> add_source e.at
```
crusso, line 307: Since `e` already carries e.note.note_type, perhaps one could remove the add_type_annotation in future.
If we added e.note.note_env, then we could even recover the local type environment for every expression, but might lead to space leaks (not sure).

Kamirus: Yes these are correct next steps, same for `add_source` probably

PR #5800, opened 2026-01-22
fix: report a "continue outside of loop" for labeled `continue`

christoph-dfinity:
The new error code introduced for unlabeled break/continue is a better fit for the case of a labeled continue targeting a non-loop label.

Fixes LANG-1128


Kamirus (approved):
Looks good!

PR #5807, opened 2026-01-30
experiment: extract common logic for classical and enhanced backends

alexandru-uta:


src/codegen/compile_common.ml
```diff
@@ -177,3 +177,34 @@ module Imports = struct
     func_imports, ni', remapping
 end
 
+type backend_arch = Wasm32 | Wasm64
```
alexandru-uta: not really needed, will remove

src/codegen/compile_classical.ml
```diff
@@ -156,13 +156,6 @@ module TaggingScheme = struct
       | _ -> assert false)
 end
 
-(*
```
crusso, line 159: To avoid changing all the references, you could also just do
``
open Pointers.W32_Pointers;
``
near the beginning.
ditto in compile_enhanced.

alexandru-uta: yeah, I see, will do

crusso: Or
module Pointers = WXX_Pointers (and always reference Pointers.<foo> in both files).

alexandru-uta: done :)

PR #5811, opened 2026-02-02
bug: Avoid moc.js crashing on exit

Kamirus:


rvanasa:
A safe/result-style API would be preferable but might be difficult to avoid breaking compatibility with previous `moc.js` versions. Finding a way to do this would make other error messages a lot more readable as well. Otherwise, should be fine to just throw an exception.

Kamirus:
@crusso I went with this idea to shadow the `Stdlib.exit` with my `Exit.exit` minimizing churn and covering more cases.
But I also added `fail` there so that if there is a message it will be added.
@rvanasa for now I'll stick to throwing an exception for simplicity and no breaking changes

PR #5814, opened 2026-02-03
feat: makes `system` capability available in mixin top-level

### LINKED ISSUE #5805: Expose `system` capability in mixin bodies
At the moment we don't allow code like:

```
mixin() {
   func f<system>(){};
   f<system>();
};
```

See `test/fail/mixin-capabilities.mo`.

In order to preserve the security guarantees we provide with the `system` capability, we should extend mixins to accept a `<system>` type parameter and require passing it on the `include` like so:

```
// MyMixin.mo
mixin<system>() {
   func f<system>(){};
   f<system>();
};

// Actor.mo
import MyMixin "MyMixin";

actor {
  include MyMixin<system>();
}
```

(edited by @christoph-dfinity)

christoph-dfinity:
Fixes #5805 

Could consider making this more strict at some point by passing the capability as a type parameter on the include.

## Implementation notes

I conjured the `C.top_cap` from "ether" like we do for actor classes. If you'd rather have me find a spot to insert it in the parser like we do for actors, I can give that a try as well.

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -4189,7 +4189,11 @@ and infer_dec env dec : T.typ =
     let t_pat, ve = infer_pat_exhaustive error env args in
     let env' = adjoin_vals env ve in
     let obj_sort : obj_sort = { it = T.Mixin ; at = no_region; note = { it = true; at = no_region; note = () } }  in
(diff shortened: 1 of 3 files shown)


crusso (approved):
Did you still want a review on this or going to add support for a single system type parameter?

PR #5816, opened 2026-02-03
feat: pretty errors with code snippets

christoph-dfinity:
image

Turned off by default for now, can be turned on by setting the repurposed `--print-source-on-error` flag.

## Implementation

Uses the [grace](https://github.com/johnyob/grace) library to pretty print snippets with markers and annotations. 

Extends our diagnostics type with spans and notes taking inspiration from https://rustc-dev-guide.rust-lang.org/diagnostics.html

I'd like to get this infrastructure in place first, and then incrementally make use of the spans to improve our existing errors. Once we're happy we can then flip the default for our human users.

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -11,6 +11,7 @@ module A = Effect
 module C = Async_cap
 
 module S = Set.Make(String)
+type usage = { assigned : bool }
 
 (* Contexts *)
 
@@ -34,7 +35,10 @@ let initial_scope =
     Scope.con_env = T.ConSet.singleton C.top_cap;
   }
 
-type unused_warnings = (string * Source.region * Scope.val_kind) List.t
+module UWSet = Set.Make(struct
+  type t = string * Source.region * Scope.val_kind
+  let compare (_, r1, _) (_, r2, _) = Region_ord.compare r1 r2
+end)
 
 type env =
   { vals : val_env;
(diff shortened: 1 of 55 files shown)


src/mo_frontend/typing.ml
```diff
@@ -250,11 +250,11 @@ let display_typs fmt typs =
   else
     Format.fprintf fmt ""
 
-let type_error at code text : Diag.message =
-  Diag.error_message at code "type" text
+let type_error at code text notes spans : Diag.message =
```
Kamirus, line 252: Wouldn't it be better if notes and spans were optional? Omitted when not needed?

src/mo_frontend/typing.ml
```diff
@@ -466,7 +474,7 @@ let coverage' warnOrError category env f x t at =
   let uncovered, unreached = f x t in
   List.iter (fun at -> warn env at "M0146" "this pattern is never matched") unreached;
   if uncovered <> [] then
-    warnOrError env at "M0145"
+    warnOrError ?notes:None ?spans:None env at "M0145"
```
```suggestion
    warnOrError env at "M0145"
```
That does not work?

src/lang_utils/diag.ml
```diff
@@ -1,19 +1,28 @@
 open Mo_config
+module G = Grace
 
 type error_code = string
 type severity = Warning | Error | Info
+type priority = Primary | Secondary
+type span = {
+  prio : priority;
+  at_span : Source.region;
+  text : string;
```
Kamirus: rename to `label`

src/mo_frontend/typing.ml
christoph-dfinity: As discussed I'll leave this non-optional for now, to encourage using the "top-level" functions instead, and to make it harder to miss providing these arguments for the more custom places.

PR #5819, opened 2026-02-04
feat: LetD/VarD Type Recovery during pre phase

Kamirus:
Problem: `let _ = myMap.ad` after checking this program `myMap` has `???` type so no method completions can be found for it. Programs like `myMap.ad` work just fine -- `myMap` has the correct type attached to the `exp`.
Goal: Preserve as much type info in expressions to improve vscode extension capabilities.

Note that this type recovery strategy is behind the existing flag.


kritzcreek (approved):
LGTM

PR #5824, opened 2026-02-06
feat: allow destructuring import from `actor` when only types are mentioned

ggreif:
Only error `M0114` when an object pattern tries to import a value from an `actor` (e.g. a `.did` file).

TODO:
- I will try to lift this restriction in a subsequent PR
- Maybe pinpoint the value in the object pattern that we cannot import (yet).

### DIFF
--- a/test/run-drun/issue-5824.mo
+++ b/test/run-drun/issue-5824.mo
@@ -1,5 +1,4 @@
 //MOC-FLAG --actor-idl issue-5824
-
 import { type T } = "ic:aaaaa-aa";
 
 actor {


test/fail/actor-match.mo
```diff
@@ -1,10 +1,13 @@
...
 };
 
 func foo() = switch a {
   case {f; g} { () }
 };
 
-assert ((switch (foo()) { case () 0 }) == 0)
+assert ((switch (foo()) { case () 0 }) == 0);
+
+let { type A } = a; // okay to import types from actor
```
alexandru-uta, line 13: could we add an example of how this is used later?

ggreif: This is just an _object pattern_ like one would use in a destructuring `import`.

test/fail/actor-match.mo
```diff
@@ -1,10 +1,13 @@
...
+  public func g() : () {};
+  public type A = Int
 };
 
 func foo() = switch a {
   case {f; g} { () }
 };
 
-assert ((switch (foo()) { case () 0 }) == 0)
+assert ((switch (foo()) { case () 0 }) == 0);
```
ggreif:
```suggestion
assert (switch (foo()) { case () 0 }) == 0;
```

christoph-dfinity (approved):
`import ... from "ic:..."` 

Okay I see. That actually boils down to importing from `did` files, I think? LGTM

test/run-drun/issue-5824.mo
```diff
@@ -0,0 +1,13 @@
+//MOC-FLAG --actor-idl issue-5824
+
```
ggreif:
```suggestion
```

test/fail/actor-match.mo
ggreif: See `test/run-drun/issue-5824.mo`

PR #5826, opened 2026-02-06
fix: don't miss `type` fields in pattern match

### LINKED ISSUE #5825: Sometimes I cannot import a `type` field
While debugging #5824, I found this obviously strange behaviour

```Motoko
> let { type C } : module { type A = Int; type B = Int; type C = Int } =
  module { public type A = Int; public type B = Int; public type C = Int; public type D = Int };
```

I get this error with v1.1.0:
```
stdin:1.7-1.13: type error [M0119], object type field C is not contained in expected type
  module {type A = Int; type B = Int; type C = Int}
```
The error message even says that there is `type C = Int` in the actor type!

I have a fix.

ggreif:
find `type` fields in imports, even when not alphabetically aligned

The problem was that `Lib.List.align` only works for both input lists sorted and when passed the (in spirit) same comparison as for sorting. This was not the case here, the comparison was flipped.

Fixes #5825.

Candidate for bugfix release.

### DIFF
--- a/test-runner/Cargo.lock
+++ b/test-runner/Cargo.lock
@@ -202,6 +202,28 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
+name = "console"
+version = "0.16.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "03e45a4a8926227e4197636ba97a9fc9b00477e9f4bd711395687c5f0734bec4"
+dependencies = [
+ "encode_unicode",
+ "libc",
+ "once_cell",
+ "unicode-width 0.2.2",
+ "windows-sys 0.61.0",
+]
+
+[[package]]
+name = "convert_case"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


src/mo_frontend/typing.ml
```diff
@@ -3548,8 +3544,8 @@ and check_pat_fields_typ_dec env t fs tfs pfs te at : Scope.typ_env =
   (* Assumes fs, tfs, and pfs are sorted *)
   let typ_pfs, val_pfs = List.partition_map (fun pf -> match pf.it with
     | ValPF(id, p) -> Either.Right(id, p)
-    | TypPF(id) -> Either.Left(id, pf.at)) pfs in
-  let cmp = fun tf (id, _) -> String.compare id.it tf.T.lab in
+    | TypPF id -> Either.Left(id, pf.at)) pfs in
+  let cmp tf (id, _) = String.compare tf.T.lab id.it in
```
ggreif, line 3560: This is the fix. The `compare` arguments were flipped. See other places where a `cmp` is passed to `Lib.List.align`.

test/run/import-type-field.mo
```diff
@@ -1 +1,5 @@
 import { type List; cons } = "lib/ListM";
+
+
```
ggreif:
```suggestion

```

christoph-dfinity (approved):
Whoops... Good catch. Thank you for fixing.

PR #5830, opened 2026-02-10
feat: [test-runner] Match output file content

alexandru-uta:
Adds clap for better input parsing and implements filtering tests by searching in the test `.ok` output. Also adds `--just-tc` option, which just type checks tests instead of running the whole suite, speeding it up for changes when functionality is not altered.

Screenshot 2026-02-10 at 11 32 37
Screenshot 2026-02-10 at 11 33 38

Closes LANG-1153

### DIFF
--- a/test-runner/src/main.rs
+++ b/test-runner/src/main.rs
@@ -288,7 +288,7 @@ fn main() {
         if !path.ends_with("motoko") {
             println!("Current path: {:?}", path.display());
             println!(
-                "test-runner --interactive should be run in the top-level motoko/ main repo directory only."
+                "test-runner should be run in the top-level motoko/ main repo directory only in interactive mode."
             );
             return;
         }


test-runner/src/main.rs
```diff
@@ -13,23 +14,43 @@ use std::sync::Arc;
...
+        long,
+        help = "Allows user to filter via pattern matching in the contents of the test output file.",
+        default_value = "false"
+    )]
+    pub in_file: bool,
+    #[arg(
+        long,
+        conflicts_with = "run",
+        help = "Just run type checking on tests.",
+        default_value = "false"
```
christoph-dfinity:
```suggestion
```

flags are false by default, no need to set this here.

test-runner/src/main.rs
```diff
@@ -60,15 +81,37 @@ fn run_interactive_mode(pre_filled_input: &str) {
...
+                    }
+                    final_output
+                } else {
+                    "".to_string()
+                };
+                TestFile {
+                    path,
+                    content: ok_file_content,
+                }
+            })
```
christoph-dfinity: I don't know about eagerly reading all these output files into memory. Shouldn't we do that conditionally on whether `search_in_file` is set?

Could just use an empty String in that case, as those don't allocate.

christoph-dfinity (changes requested):
Two small suggestions.

I like the idea of being able to run all tests that correspond to a given error code.

test-runner/src/main.rs
alexandru-uta: sure, doing that now.

PR #5831, opened 2026-02-10
feat: Emit `suggested_replacement` with the json error format

Kamirus:
Emit machine-applicable code fixes in `--error-format json` diagnostics for warnings M0223 (redundant type instantiation), M0236 (use dot notation), and M0237 (omit explicit implicit argument). The JSON span format now includes `is_primary`, `label`, `suggested_replacement`, and `suggestion_applicability` fields, enabling IDEs and tooling to offer automatic fixes

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -3446,17 +3446,18 @@ and check_pat_aux' env t pat val_kind : Scope.val_env =
     in check_pats env ts pats T.Env.empty pat.at
   | ObjP pfs ->
     let pfs' = List.stable_sort compare_pat_field pfs in
+    let vpfs = List.filter_map (fun pf ->
+      match pf.it with
+      | TypPF _ -> None
+      | ValPF(id, _) -> Some(id.it)) pfs' in
     let s, fs =
-      try T.as_obj_sub (List.filter_map (fun pf ->
-        match pf.it with
-        | TypPF _ -> None
-        | ValPF(id, _) -> Some(id.it)) pfs') t
+      try T.as_obj_sub vpfs t
       with Invalid_argument _ ->
         error env pat.at "M0113" "object pattern cannot consume expected type%a"
           display_typ_expand t
     in
-    if not env.pre && s = T.Actor then
-      local_error env pat.at "M0114" "object pattern cannot consume actor type%a"
+    if not env.pre && s = T.Actor && vpfs <> [] then
+      local_error env pat.at "M0114" "object pattern cannot consume values from actor type%a"
         display_typ_expand t;
     check_pat_fields env t fs pfs' T.Env.empty pat.at
   | OptP pat1 ->
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,7 @@
 
 * motoko (`moc`)
 
+  * Allow destructuring `type` imports from `actor`-valued URIs (#5824).
   * perf: Optimise a few arithmetic/logic operations involving neutral elements (#5706).
   * Adds `--error-format human` option to print pretty errors with code snippets and labels (#5816).
   * Emit machine-applicable code fixes in `--error-format json` diagnostics for warnings M0223 (redundant type instantiation), M0236 (use dot notation), and M0237 (omit explicit implicit argument). The JSON span format now includes `is_primary`, `label`, `suggested_replacement`, and `suggestion_applicability` fields, enabling IDEs and tooling to offer automatic fixes (#5831).
--- a/test/fail/actor-match.mo
+++ b/test/fail/actor-match.mo
@@ -1,10 +1,13 @@
-let a : actor {f : () -> (); g : () -> ()} = actor {
+let a : actor {f : () -> (); g : () -> (); type A = Int} = actor {
   public func f() : () {};
-  public func g() : () {}
+  public func g() : () {};
+  public type A = Int
 };
 
 func foo() = switch a {
   case {f; g} { () }
 };
 
-assert ((switch (foo()) { case () 0 }) == 0)
+assert (switch (foo()) { case () 0 }) == 0;
+
+let { type A } = a; // okay to import types from actor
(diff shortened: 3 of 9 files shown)


src/lang_utils/diag.ml
```diff
@@ -9,6 +9,10 @@ type span = {
   at_span : Source.region;
   label : string;
 }
+type edit = {
+  at_edit : Source.region;
+  suggested_replacement : string;
```
Kamirus, line 14: It is cleaner to have `edits` separate from `spans` because there is no priority nor label here.
It's easier to remember when serializing into diagnostics format to handle edits differently

design/JSON-Diagnostics.md
```diff
@@ -52,9 +50,33 @@ Each line on stdout is a JSON object with the following structure:
...
+            "label": null,
+            /* An optional string of a suggested replacement for this span.
+               When present, tools may replace the contents of the span with this text.
+            */
+            "suggested_replacement": null,
+            /* An optional indicator of the confidence of the suggested replacement.
+               Present only when suggested_replacement is non-null.
+
+               Possible values may be:
+               - "MachineApplicable": The suggestion is definitely what the user intended.
```
Kamirus, line 69: For now all emitted suggestions are machine applicable, but this allows us to extend the suggestions with other kinds in the future. Clients should always check the applicability before auto-applying.

design/JSON-Diagnostics.md
```diff
@@ -52,9 +50,33 @@ Each line on stdout is a JSON object with the following structure:
...
+                 This suggestion should be automatically applied together with
+                 all other machine-applicable suggestions in this diagnostic.
```
Kamirus, line 71: A single suggestion is not valid on its own (see the `test_edits.ml`). My idea was to emit as fine-grain suggestions as possible so that:
1. Format can be preserved
2. As many suggestions can be applied in one go, for now all suggestions (across multiple spans) are not overlapping so all can be applied in one go (note that this does NOT ensure fixpoint because the 'redundant type instantiation' is tricky and a bit broken, see https://github.com/caffeinelabs/motoko/pull/5605 or https://github.com/caffeinelabs/motoko/pull/5742)

src/mo_frontend/typing.ml
```diff
@@ -298,6 +298,9 @@ let primary env at fmt =
 let secondary env at fmt =
   Format.kasprintf env (fun s -> Diag.{ prio = Secondary; at_span = at; label = s }) fmt
 
+let edit replacement at : Diag.edit =
+  Diag.{ at_edit = at; suggested_replacement = replacement }
```
christoph-dfinity: Nit for consistency: our other constructors accept the `at` before accepting the "what"

test/fail/ok/contextual-dot-suggest.tc-human.ok
```diff
@@ -2,27 +2,41 @@ warning[M0236]: You can use the dot notation `(#odd).equal(...)` here
     ┌─ contextual-dot-suggest.mo:82:10
  82 │    ignore Odd.equal(#odd); // warn non-binary
     │           ^^^^^^^^^^^^^^^ 
+    = help: replace `Odd` with `(#odd)`
```
Kamirus: Actually, maybe we don't want suggestions that lead to 
`(#odd).equal()`
We might want to skip edits when the receiver needs to be wrapped in parenthesis 🤷‍♂️

src/lang_utils/diag.ml
christoph-dfinity: I could see edits being attached to messages/suggestions... But I see you've split the edits up into smaller chunks, in which case there's not a 1-1 mapping between edits and suggestions, so I think the current factoring makes sense.

test/fail/ok/contextual-dot-suggest.tc-human.ok
```diff
@@ -2,27 +2,41 @@ warning[M0236]: You can use the dot notation `(#odd).equal(...)` here
...
  83 │    ignore Odd.compare(#odd, #odd, #odd); // warn non-binary
     │           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
+    = help: replace `Odd` with `(#odd)`
+    = help: remove `#odd, `
 warning[M0236]: You can use the dot notation `peopleMap.get(...)` here
     ┌─ contextual-dot-suggest.mo:86:10
  86 │    ignore Map.get(peopleMap, Nat.compare, 1); // warn
     │           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
+    = help: replace `Map` with `peopleMap`
+    = help: remove `peopleMap, `
```
christoph-dfinity: I love this!

These would actually read even nicer if they we're a single diff. Could we maybe attach multiple edits to a single span? In the JSON format we could still stick to whatever rustc ends up doing.

src/mo_frontend/test_edits.ml
christoph-dfinity, line 1: The more I look at this test, the less I like it... It's doing basically the things we're doing with `run.sh`, like setting warning flags adding pkgs, etc... but then it's running in a different test-suite with different ergonomics.
Doesn't the printing of `help:` for the edits you're suggesting cover the same thing these tests do? Since we do snapshot the `--error-format=human` output now

test/fail/ok/contextual-dot-suggest.tc-human.ok
christoph-dfinity: That seems sensible to me. If the receiver can't just be moved "verbatim" we could suppress the suggestion.

src/mo_frontend/test_edits.ml
Kamirus: > Doesn't the printing of help: for the edits you're suggesting cover the same thing these tests do? Since we do snapshot the --error-format=human output now

You are right! I did not have that output before, let me refactor that to avoid these custom dune tests

test/fail/ok/contextual-dot-suggest.tc-human.ok
Kamirus: I'd do a single diff later, I don't want to delay this PR longer
It'd need some formatter and what not, for now this is simple enough and still useful

PR #5833, opened 2026-02-11
feat: reports a warning for `var` bindings that are never reassigned

christoph-dfinity:
Fixes [LANG-1140](https://linear.app/caffeinelabs/issue/LANG-1140/warn-on-var-declaration-that-is-never-assigned)

Also refactors the unused warnings code a bit to use a Set to check for duplicates instead of linear search.

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -111,10 +111,7 @@ let use_identifier env id =
   ) !(env.used_identifiers)
 
 let assign_identifier env id =
-  env.used_identifiers := T.Env.update id (function
-    | Some _ -> Some { assigned = true }
-    | None -> Some { assigned = true }
-  ) !(env.used_identifiers)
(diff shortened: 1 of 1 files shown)


src/mo_frontend/typing.ml
```diff
@@ -4372,6 +4372,8 @@ and gather_dec env scope dec : Scope.t =
       ) scope.typ_env tfs in
       let val_env = List.fold_left (fun ve T.{ lab; typ; _ } ->
         if T.Env.mem lab ve then error_duplicate env "" { it = lab; at = i.at; note = () };
+        (* We don't want to produce unused warnings for declarations introduced through mixins *)
```
alexandru-uta: why not? just so we don't get warnings that got introduced by "external"/"library" code (which might have been written by others)?

christoph-dfinity: Yes. It's kind of the same thing as never warning about unused declarations that are marked `public`, as we can't know all the call sites.

PR #5839, opened 2026-02-12
experiment: newtype

Kamirus:


design/WhitePaper.md
```diff
@@ -159,6 +159,60 @@ Finally, it would sometimes be desirable to have extended syntax for multi-line
...
+- **Constructor.** The newtype name is bound as both a type and a value (the constructor function). For `newtype T<A> = U<A>`, the constructor has type `<A>(U<A>) -> T<A>`.
+- **Unwrap.** Values of newtype are given a virtual `.unwrap` field (analogous to how arrays have `.size`). For a value `v : T<A>`, `v.unwrap : U<A>`.
+
+##### Candid Serialization
+
+Newtypes are a Motoko-only concept. When serializing to Candid, newtypes are transparent: `Time` serializes as `Int`, `Map<K, V>` serializes as its underlying record type.
+
+##### Stable Storage and Migrations
+
+For `.most` files (stable type signatures), newtypes are treated like type aliases. This allows users to freely add, rename, or remove newtypes across upgrades without breaking migration compatibility.
```
crusso, line 198: Hmm, I guess that might fix the problem I noticed with signatures. Feels a bit odd though.

crusso: At least it would mean moving from unbranded map to branded map would not be a breaking change, but I'm not sure that's a good enough motivation.

Kamirus:
This should be reopened once we have more time to pursue `newtype`s

PR #5840, opened 2026-02-12
feat: enhanced migration

alexandru-uta:
Resolves LANG-1126

Adds a new type of migration expression, where users specify a `migrations` folder via the compiler, e.g., `moc --enhanced-migration migrations_folder --enhanced-orthogonal-persistence.`

NB: the whole feature is gated behind the `--enhanced-migration` flag, so all the older functionality is retained.

**High-level Design Decisions**
* actors don't allow initializers on variables, e.g., declarations like `let x : Nat = 5` are forbidden in actor bodies.
* all actors have migration functions, even those deployed from scratch. All variables get initialized only in migration functions. Newly deployed actors will have to initialize their stable variables via an initializer function.
* migrations can be a single function (which is migration from v(i) to v(i+1)), or a sequence of functions (which is migration from v(i), v(i+1), …, v(k)); all these functions should compose correctly, otherwise the compiler will reject the migration
* all migration functions will be identified by the file in which they are defined — a single migration per file; in production, these should be timestamps (datetime).
* always, on every new deployment, _ALL_ the migrations from the migrations folder will then be applied in lexicographic order.
* the RTS will not allow to run the same migration twice as it will keep track of which migrations were already applied.

The **restrictions** for the multi-migration are the following:
* Each migration file should be a module with a function called `migration()` -- this will implement the migration logic.
* The output of each migration must be compatible (subtype) with the input of the next one. The program fails to compile if not.
* Each migration only needs to declare the fields it reads (input) and produces (output). Fields not mentioned by a migration are automatically carried through unchanged.
* The output of the migration chain must be compatible with the target actor's stable fields.
* Fields are only removed from the state when the final actor no longer declares them.
* The enhanced migration (--enhanced-migration) cannot be combined with the regular inline migration (with migration = ...).

```
migrations/                                    actor {
├── 20240101_init.mo                             var b : Text;
│   migration : {} → {a; b}                      var c : Bool;
│                                                var d : Nat;
├── 20240315_add_field.mo
│   migration : {a; b} → {a; b; c}              public func check() : async () {
│                                                  ...
├── 20240601_rename_a.mo                         };
│   migration : {a; c} → {c; d}                }
│
```

# Extended signatures

Adds notion of multi-migration signature recording sequence of named migration function signatures followed by actor signature.

Eg.
``` motoko
// Version: 4.0.0
{
  "1" : {} -> {a : Nat};
  "2" : {} -> {b : Int};
  "3" : (old : {b : Int}) -> {b : Bool};
  "4" : (_old : {a : Nat}) -> {};
  "5" : (_old : {}) -> {a : Text}
}
actor  {
  stable a : Text;
  stable var b : Bool
} ;
```

* migration signatures ordered by lexicographic ordering of (text literal) labels "1", ... "n" (typically filenames)
* implement multi sig parsing, checking and printing.
* extend stable-compatibility relation to check post signature of old .most is compatible with appropriate pre signature of new .most.
* When comparing two signatures, we use the (last) migration function name of the old signature (if any) to determine the requirements of the next migration function in the new signature.
* reflexive on multi-migrations (mirroring idempotence)

- [x] For now statically and dynamically prevent non-multimgration after multi-migration (or require a --version arg to push on  migration list.
- [x] Restrict actor decs to be static module system calls.
- [ ] ~~Did we ever review whether graph stabilization still works?~~ Stabilization doesn't work out of the box and we will tackle it in a later PR.

### DIFF
--- a/test/fail/ok/enhanced-migration-no-dir.tc-human.ok
+++ b/test/fail/ok/enhanced-migration-no-dir.tc-human.ok
@@ -1,1 +1,1 @@
-: No such file or directory
+(unknown location): import error [M0256], enhanced migration path 'enhanced-migration/bogus' is not a directory
--- a/test/fail/ok/implicit-stable-pattern.tc-human.ok
+++ b/test/fail/ok/implicit-stable-pattern.tc-human.ok
@@ -1,1 +1,4 @@
-(unknown location): type error [M0133], misplaced stability modifier: allowed on var or simple let declarations only
+error[M0133]: misplaced implicit stability modifier: allowed on var or simple let declarations only
+    ┌─ implicit-stable-pattern.mo:2:4
+  2 │     let (x,y) = (1,2) // illegal implicit stable pattern
+    │     ^^^^^^^^^^^^^^^^^ 
--- a/test/fail/ok/implicit-stable-pattern.tc.ok
+++ b/test/fail/ok/implicit-stable-pattern.tc.ok
@@ -1,1 +1,1 @@
-(unknown location): type error [M0133], misplaced stability modifier: allowed on var or simple let declarations only
+implicit-stable-pattern.mo:2.4-2.21: type error [M0133], misplaced implicit stability modifier: allowed on var or simple let declarations only


src/lowering/desugar.ml
```diff
@@ -767,128 +767,121 @@ and build_actor at ts (exp_opt : Ir.exp option) self_id es obj_typ =
...
+          let run_expr = dotE mod_expr "run" run_typ_k in
+          let extract_dom =
+            objectE T.Object
+              (List.map (fun T.{lab=i;typ=t;_} ->
+                let opt_dom_ty = T.Opt (T.as_immut t) in
+                let state_field_ty = T.lookup_val_field i type_prev_fields in
+                let vi = fresh_var ("v_"^i) (T.as_immut t) in
+                (i,
+                 switch_optE
+                   (primE (I.CastPrim (state_field_ty, opt_dom_ty))
```
crusso: Why is the cast necessary?

src/lowering/desugar.ml
```diff
@@ -767,128 +767,121 @@ and build_actor at ts (exp_opt : Ir.exp option) self_id es obj_typ =
...
+          let merge_result =
+            objectE T.Memory
+              (List.map (fun T.{lab=i;typ=t;_} ->
+                i,
+                match T.lookup_val_field_opt i rng_fields_k with
+                | Some rt ->
+                  optE (dotE (varE v_rng) i (T.as_immut rt))
+                | None ->
+                  let prev_ty = T.lookup_val_field i type_prev_fields in
+                  primE (I.CastPrim (prev_ty, t))
```
crusso: Ditto

src/lowering/desugar.ml
```diff
@@ -767,128 +767,121 @@ and build_actor at ts (exp_opt : Ir.exp option) self_id es obj_typ =
         ) chain
```
crusso: These still look wrong to me since they will include fields that were dropped and never re-introduced. Maybe it works out, but it's dodgy.

src/lowering/desugar.ml
```diff
@@ -767,128 +767,121 @@ and build_actor at ts (exp_opt : Ir.exp option) self_id es obj_typ =
...
           (List.map (fun T.{lab=i;typ=t;_} ->
-            i, primE (I.CastPrim (T.Opt T.Any, t))
-                 [dotE (varE state_var) i (T.Opt T.Any)])
+            i,
+            match T.lookup_val_field_opt i type_n_fields with
+            | Some tn ->
+              primE (I.CastPrim (tn, t))
+                [dotE (varE final_state) i tn]
+            | None ->
+              primE (I.CastPrim (T.Prim T.Null, t)) [nullE ()])
```
crusso: more weird casts

src/lowering/desugar.ml
alexandru-uta: fixed

src/lowering/desugar.ml
crusso: Actually, I would tell the AI to calculate type_{n-1} in a reverse loop from type type_n, where the base case type_n  is the content off all stable vars in the actor (optioned). The calculation should be similar to the pre_sig calculation elsewhere in the code.

src/lowering/desugar.ml
```diff
@@ -744,30 +744,36 @@ and build_actor at ts (exp_opt : Ir.exp option) self_id es obj_typ =
...
-            List.map (fun tf -> {tf with T.typ = T.Opt (T.as_immut tf.T.typ)}) accum_fields
-          in
-          T.Obj (T.Memory, accum_mem_fields, [])
-        ) chain
+        let undo_migration next_fields (_file, _mod_typ, run_typ) =
+          let (_, _, dom_fields_k, rng_fields_k) = decompose_run run_typ in
+          let prev_tbl = Hashtbl.create 32 in
+          List.iter (fun tf ->
+            if not (List.exists (fun rf -> rf.T.lab = tf.T.lab) rng_fields_k) then
+              Hashtbl.replace prev_tbl tf.T.lab tf
```
crusso: This looks good but I wonder if we still have a problem with allowing fields to omit initializers.

Consider:
```
{ a : #A} -> {}
actor { 
   a : #B [= #B]
}
```
In the existing semantics, this would (Ibelieve) remove `a : #A` and redefine `a` (at unrelated type `#B`) by running the initializer. But now, since we don't have an initializer, we can't redefine, but need to assume a : #B in the old actor, but this is inconsistent with the  { a : #A }, required by migration.

crusso: Maybe we just have rule that out statically (in typing.ml) for multi-migration: a stable field declared in the actor cannot occur in the domain of the migration function, unless it occurs in the range at a compatible type.

alexandru-uta: Already fixed in typing.ml

doc/enhanced-multi-migration.md
```diff
@@ -0,0 +1,189 @@
...
+    register_migration(m_{n-1})
+    merge(output_{n-1}, state_{n-1}) -> state_n
+
+ICStableStore(mem_ty)                             // update stored metadata
+result = project state_n to actor's mem_ty        // drop fields actor no longer declares
+```
+
+Each `merge(output_k, state_{k-1})` builds `type_k` by:
+- Taking range fields from `output_k` (wrapped in `?T`)
+- Carrying non-range fields from `state_{k-1}` unchanged
```
crusso:
```suggestion
- Carrying non-range and non-domain fields from `state_{k-1}` unchanged
```
The migration function *consumes* the domain fields, right? I think merge needs to depend on the dom type too to make this precise.
Perhaps the code gets this right already...

doc/enhanced-multi-migration.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Phase 1: Accumulated composition
+
+Given migration chain `[m_0, m_1, ..., m_n]`:
+
+For **every** k from 1 to n, the compiler checks
+`accumulated(m_0 .. m_{k-1}) <: input(m_k)`.
+Each migration has access to the full accumulated state (via merge semantics),
+not just the previous migration's output. A migration can reference a field
+produced by any earlier migration, not just the immediately preceding one.
```
crusso: I don't think this is correct, right? The output can only depend on the immediate input for each migration function

doc/enhanced-multi-migration.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Key Properties
+
+- **Idempotency.** Already-applied migrations are skipped; redeploying is a no-op.
+- **Fast-forward.** v1 → v100 directly, or v1 → v50 → v100, or step-by-step.
+- **Type changes.** Each level is strongly typed at `type_k`; type changes are
+  handled naturally by the migration's `run` function.
+- **Field drops.** Two ways: (1) consume the field without producing it
+  (`{a : T} -> {}`), which drops it from the accumulated state and prevents
+  later migrations from referencing it; (2) simply remove the field from the
```
crusso: I don't think our static checks allowed (2). To easy to lose data.

rts/motoko-rts/src/memory.rs
```diff
@@ -136,3 +136,126 @@ pub unsafe fn set_dedup_table<M: Memory>(mem: &mut M, dedup_table: Value) {
     }
     set_dedup_table_ptr(mem, dedup_table);
 }
+
+#[enhanced_orthogonal_persistence]
```
crusso: I wonder if this couldn't just be a single migration file name - the last migration applied.
When we look for the next migration function to apply, from highest to lowest, we compare against the last stored migration function name until we find it and then update with the (new) last migration function name when done.

That would be linear, not quadratic and just use 0(1) space.

But maybe I'm missing something.

src/lowering/desugar.ml
```diff
@@ -685,7 +693,200 @@ and build_actor at ts (exp_opt : Ir.exp option) self_id es obj_typ =
...
+      let chain = !T.migration_chain in
+
+      let decompose_run run_typ =
+        let dom_i, rng_i = T.as_mono_func_sub run_typ in
+        let (_ds, dom_fields_i) = T.as_obj (T.normalize dom_i) in
+        let (_rs, rng_fields_i) = T.as_obj (T.promote rng_i) in
+        (dom_i, rng_i, dom_fields_i, rng_fields_i)
+      in
+      let migration_hash file =
+        Digest.to_hex (Digest.string (Filename.basename file))
```
crusso: Since we seem to be hashing just the filename, why not  just use the filename?

src/mo_frontend/typing.ml
```diff
@@ -2284,7 +2284,14 @@ and infer_exp'' env exp : T.typ =
     T.unit
   | AnnotE (exp1, typ) ->
     let t = check_typ env typ in
-    if not env.pre then check_exp_strong env t exp1;
+    if not env.pre then
+      (* If exp1 is TupE [] (placeholder for no initializer) and --enhanced-migration is enabled, *)
+      (* allow it without type checking (it's used for stable variables without initializers) *)
+      (match exp1.it, !Flags.enhanced_migration with
```
crusso: This is kinda gross. Maybe it would be better to have a special prim that has any type (or `None`) and traps when executed.

src/mo_frontend/parser.mly
```diff
@@ -956,11 +956,19 @@ func_pat :
...
+    (* Type checker will verify this is only allowed for stable variables with --enhanced-migration *)
+    { let unit_exp = TupE([]) @? at $sloc in
+      VarD(x, annot_exp unit_exp t) @? at $sloc }
 
 dec_nonvar :
   | LET p=pat EQ e=exp(ob)
     { let p', e' = normalize_let p e in
       LetD (p', e', None) @? at $sloc }
+  | LET p=pat
+    { let p', e' = normalize_let p (TupE([]) @? at $sloc) in
```
crusso: Here it would be better use a dummy prim that has any type and traps when executed, I think.

src/mo_frontend/typing.ml
```diff
@@ -3909,6 +3929,121 @@ and infer_migration env obj_sort exp_opt =
...
+(* Validate the enhanced migration chain from --enhanced-migration directory.
+
+   Each incremental step v_i -> m_{i+1} -> v_{i+1} has the same semantics as
+   the old (with migration = fn) syntax: a migration only needs to mention fields
+   it transforms; everything else is retained from the accumulated state.
+
+   Checks:
+   (1) each run type is a valid local function with stable object input/output,
+   (2) accumulated(m_0..m_i) <: input(m_{i+1}) — each migration can reference
+       any field from the accumulated state, not just the previous output,
```
crusso: That ain't true is it?

src/mo_frontend/typing.ml
```diff
@@ -3909,6 +3929,121 @@ and infer_migration env obj_sort exp_opt =
...
+    (* Fold: for each migration i, check accumulated(m_{0}, m_{1}, ..., m_{i-1}). <: input(m_{i}) (skip first),
+       then update accumulated state. Dom-only fields (consumed but not
+       produced) are dropped — consuming a field = dropping it, matching
+       the old (with migration = fn) semantics. The result is the final
+       accumulated state after all migrations, used to check actor fields. *)
+    let accumulated = List.fold_left (fun acc (file, dom_tfs, rng_tfs, input, _) ->
+      if acc <> [] then begin
+        let acc_fields = List.sort T.compare_field (List.map snd acc) in
+        let acc_typ = T.Obj(T.Object, acc_fields, []) in
+        if not (T.sub acc_typ (T.normalize input)) then
```
crusso: This check is likely incorrect. We should allow change in mutability but disallow loss of data (using `stable_sub`, not `sub`).

crusso: c.f. `field_compat` below.

src/mo_frontend/typing.ml
```diff
@@ -4037,7 +4172,21 @@ and check_migration env (stab_tfs : T.field list) exp_opt =
...
-  if sort.it <> T.Actor then () else
+  if sort.it <> T.Actor then () else begin
+  (* With --enhanced-migration, stable variables must not have initializers *)
+  if Option.is_some !Flags.enhanced_migration then begin
+    List.iter (fun dec_field ->
+      match dec_field.it.stab, dec_field.it.dec.it with
+      | Some {it = Stable; _}, LetD (_, exp, _) 
+      | Some {it = Stable; _}, VarD (_, exp) ->
+        (match exp.it with
+         | AnnotE ({it = TupE []; _}, _) -> () (* placeholder for no initializer -- OK *)
```
crusso: This would be the place to check for the special prim instead of TupE[]

src/mo_frontend/typing.ml
```diff
@@ -4192,6 +4342,15 @@ and infer_dec env dec : T.typ =
         check_exp env T.Non fail
     );
     let t = infer_exp env exp in
+    (* Check if this is a placeholder for no initializer *)
+    if not env.pre then begin
+      match exp.it with
+      | AnnotE ({it = TupE []; _}, _) when Option.is_some !Flags.enhanced_migration ->
```
crusso: (look for special prim instead)

src/mo_frontend/typing.ml
```diff
@@ -4200,6 +4359,13 @@ and infer_dec env dec : T.typ =
   | VarD (id, exp) ->
     if not env.pre then begin
       let t = infer_exp env exp in
+      (* Check if this is a placeholder for no initializer *)
+      (match exp.it with
+       | AnnotE ({it = TupE []; _}, _) when Option.is_some !Flags.enhanced_migration ->
```
crusso: (same again)

src/mo_frontend/parser.mly
```diff
@@ -956,11 +956,19 @@ func_pat :
 dec_var :
   | VAR x=id t=annot_opt EQ e=exp(ob)
     { VarD(x, annot_exp e t) @? at $sloc }
+  | VAR x=id t=annot_opt
+    (* No initializer - use unit expression () as placeholder *)
+    (* Type checker will verify this is only allowed for stable variables with --enhanced-migration *)
+    { let unit_exp = TupE([]) @? at $sloc in
```
crusso: Here it would be better use a dummy prim that has any type and traps when executed, I think. (ditto below)

src/lowering/desugar.ml
```diff
@@ -685,7 +693,200 @@ and build_actor at ts (exp_opt : Ir.exp option) self_id es obj_typ =
...
+
+      let n = List.length chain in
+
+      (* type_0: base case for ICStableRead when no migrations were performed.
+         With migrations, use Init's domain — the pre-migration actor's expected
+         shape. On fresh install, ICStableRead returns defaults (all ?T fields
+         are None). On pre-migration upgrade, the RTS checks compatibility.
+         Without migrations, use enhanced_mem_ty (= actor's mem_ty). *)
+      let type_0 =
+        if n = 0 then enhanced_mem_ty
```
crusso: Isn't n = 0 impossible ? We only enter this code when !migration_chain <> []

src/lowering/desugar.ml
```diff
@@ -685,7 +693,200 @@ and build_actor at ts (exp_opt : Ir.exp option) self_id es obj_typ =
...
+            i,
+            match T.lookup_val_field_opt i type_n_fields with
+            | Some _tn ->
+              dotE (varE final_state) i t
+            | None ->
+              nullE ())
+          mem_fields)
+        mem_fields
+      in
+      T.Single stab_fields,
```
crusso: Ok, so this means we are only recording the post migration signature as the .most file. That's probably not what we want. The question is, what do we want? 

Maybe a new sort of signature that records all possible upgrade targets:
`
T.Multi( [(filename0:type_0... filenameN:typeN], stab_fields)
`
Then we could statically check upgrades against previous (live) versions, as well as the last (draft) one in caffeine.
```
T.Multi( [(filename0:type_0... filenameN:typeN], old_stab_fields) 
<:  
T.Multi( [(filename0:type_0..., filenameN : typeN, filenameN+1 : typeN+1, ... filename_{N+M}:type_{N+M}], stab_fields)
iff
old_stab_fields <: type_{N+1} 
```
But these files could get very large....

src/codegen/compile_enhanced.ml
```diff
@@ -1290,6 +1290,8 @@ module RTS = struct
     E.add_func_import env "rts" "weak_ref_is_live" [I64Type] [I32Type];
```
crusso, line 1294: Unfortunately, we'll also have to adapt the graph_stabilization upgrade path to multimigrations, or least argue why we don't need to.

src/lowering/desugar.ml
```diff
@@ -685,7 +693,200 @@ and build_actor at ts (exp_opt : Ir.exp option) self_id es obj_typ =
...
+      I.{pre = enhanced_mem_ty; post = mem_ty},
```
crusso: Maybe this can just be:
```suggestion
      I.{pre = type_at 0; post = mem_ty},
```
I think we need to figure out if we need enhanced_mem_ty or can just use type0 here.
I think this ay only be used for graph_stabilization now but we would need to check.
If it's used for graph_stabilization, then we may be able to use the old type descriptor stored in the the old actor instead (but needs careful thought).

doc/md/fundamentals/2-actors/8-enhanced-multi-migration.md
```diff
@@ -0,0 +1,380 @@
...
+
+- **Input only** — the migration consumes and removes this field. The field is dropped from the state. Later migrations can no longer reference it.
+
+- **Neither input nor output** — the field is untouched by this migration and carried through to the next migration (or the final actor) as-is.
+
+For example, given the state `{a : Nat; b : Text; c : Bool}` and a migration:
+
+```motoko no-repl
+module {
+  public func run({ a : Nat; b : Text } : { a : Nat; b : Text }) : { a : Int; d : Float } {
```
crusso:
```suggestion
  public func run({ a : Nat; b : Text }) : { a : Int; d : Float }
```

doc/md/fundamentals/2-actors/8-enhanced-multi-migration.md
```diff
@@ -0,0 +1,380 @@
...
+```
+
+### Adding a field
+
+To add a new field, write a migration with an empty (or minimal) input that produces the new field:
+
+```motoko no-repl
+// migrations/20250201_000000_AddEmail.mo
+module {
+  public func run(_ : {}) : { email : Text } {
```
crusso:
```suggestion
  public func run({}) : { email : Text } {
```

doc/md/fundamentals/2-actors/8-enhanced-multi-migration.md
```diff
@@ -0,0 +1,380 @@
...
+All existing fields are carried through automatically.
+
+### Changing a field's type
+
+To change the type of a field, read it at its current type and produce it at the new type:
+
+```motoko no-repl
+// migrations/20250301_000000_CountToInt.mo
+module {
+  public func run({ count : Nat } : { count : Nat }) : { count : Int } {
```
crusso:
```suggestion
  public func run({ count : Nat }) : { count : Int } {
```

doc/md/fundamentals/2-actors/8-enhanced-multi-migration.md
```diff
@@ -0,0 +1,380 @@
...
+Here `count` changes from `Nat` to `Int`. The compiler accepts this because `Nat` is a subtype of `Int`.
+
+### Renaming a field
+
+To rename a field, consume the old name and produce the new name:
+
+```motoko no-repl
+// migrations/20250401_000000_RenameLabel.mo
+module {
+  public func run({ label : Text } : { label : Text }) : { title : Text } {
```
crusso:
```suggestion
  public func run({ label : Text }) : { title : Text } {
```

doc/md/fundamentals/2-actors/8-enhanced-multi-migration.md
```diff
@@ -0,0 +1,380 @@
...
+The old field `label` is removed from the state and `title` takes its place.
+
+### Removing a field
+
+To drop a field entirely, consume it in the input without producing it in the output:
+
+```motoko no-repl
+// migrations/20250501_000000_DropEmail.mo
+module {
+  public func run({ email : Text } : { email : Text }) : {} {
```
crusso:
```suggestion
  public func run({ email : Text }) : {} 
```
FUTURE: maybe we should allow users to just use Any for deleted fields, to avoiding spelling out the old type.

rts/motoko-rts/src/persistence.rs
```diff
@@ -63,6 +63,10 @@ struct PersistentMetadata {
     /// We keep a pointer to this here so that we can keep if alive across upgrades.
     /// To keep the dedup table live, we need to add it to roots as well.
     dedup_table: Value,
+    /// Migration function hashes array. This stores an array of the the hashes (strings) of the migration functions.
```
alexandru-uta:
```suggestion
    /// Migration function list. This is an RTS root that holds the head of a linked list of the the names (strings) of the migration functions.
```

rts/motoko-rts/src/persistence.rs
```diff
@@ -113,6 +117,8 @@ impl PersistentMetadata {
         (*self).weak_ref_registry = NULL_POINTER;
         // Initialize the dedup table as the null pointer.
         (*self).dedup_table = NULL_POINTER;
+        // Initialize the migration functions array as the null pointer.
```
alexandru-uta:
```suggestion
        // Initialize the migration functions list as the null pointer.
```

rts/motoko-rts/src/persistence.rs
```diff
@@ -362,3 +374,23 @@ pub(crate) unsafe fn set_dedup_table_ptr<M: Memory>(mem: &mut M, dedup_table: Va
     // Use barrier in case the dedup table is set during a GC phase.
     write_with_barrier(mem, &mut (*metadata).dedup_table, dedup_table);
 }
+
+/// Accessor method for the migration functions array.
```
alexandru-uta:
```suggestion
/// Accessor method for the migration functions list.
```

rts/motoko-rts/src/persistence.rs
```diff
@@ -362,3 +374,23 @@ pub(crate) unsafe fn set_dedup_table_ptr<M: Memory>(mem: &mut M, dedup_table: Va
...
+pub(crate) unsafe fn get_migration_functions_ptr() -> &'static mut Value {
+    let metadata = PersistentMetadata::get();
+    &mut (*metadata).migration_functions
+}
+
+/// Setter method for the migration functions array.
```
alexandru-uta:
```suggestion
/// Setter method for the migration functions list.
```

rts/motoko-rts/src/persistence.rs
```diff
@@ -362,3 +374,23 @@ pub(crate) unsafe fn set_dedup_table_ptr<M: Memory>(mem: &mut M, dedup_table: Va
...
+pub(crate) unsafe fn set_migration_functions_ptr<M: Memory>(
+    mem: &mut M,
+    migration_functions: Value,
+) {
+    let metadata = PersistentMetadata::get();
+    // Use barrier in case the migration functions array is set during a GC phase.
```
alexandru-uta:
```suggestion
    // Use barrier in case the migration functions list is set during a GC phase.
```

doc/md/fundamentals/2-actors/8-enhanced-multi-migration.md
```diff
@@ -0,0 +1,380 @@
...
+```
+
+State: `{a : Nat; b : Int}`
+
+**Step 3 — Change `b` from `Int` to `Bool`:**
+
+```motoko no-repl
+// migrations/20250301_000000_ChangeBType.mo
+module {
+  public func run({ b : Int } : { b : Int }) : { b : Bool } {
```
alexandru-uta:
```suggestion
  public func run({ b : Int }) : { b : Bool } {
```

doc/md/examples/grammar.txt
```diff
@@ -373,6 +375,13 @@
...
     'stable'
     'in'
+
+<mig_lab> ::= 
+    <text>
+
+<mig_field> ::= 
+    <mig_lab> ':' <typ>
  ')'
+    <list(<typ_dec>, ';')> '{' <list(<mig_field>, ';')> '}' 'actor' '{' <list(<stab_field>, ';')> '}'
```
Kamirus: Looks like the `/^<parse_stab_sig> ::=/,+5d` rule in the `grammar.sed` needs updating.

Here is AI context:
> This sed rule was designed to delete the parse_stab_sig production from the generated grammar (since it's an internal format for .most stable signature files, not user-facing Motoko syntax). It deletes the line matching <parse_stab_sig> ::= plus the next 5 lines.

> The fix would be to either:
> 1. Increase the +5 to a larger number to cover all three alternatives (fragile — breaks again if another alternative is added), or
> 2. Add explicit deletion rules for the helper productions too (/^<mig_lab> ::=/,+2d, /^<mig_field> ::=/,+2d, /^<pre_stab_field> ::=/,+2d, /^<req> ::=/,+3d), or
> 3. Use a range-based deletion that's more robust, like /^<parse_stab_sig> ::=/,/^$/d (delete until the next blank line), which would be resilient to future changes

src/pipeline/resolve_import.ml
```diff
@@ -234,6 +234,19 @@ let resolve_package_url (msgs:Diag.msg_store) (pname:string) (f:url) : filepath
   then f
   else (err_package_file_does_not_exist msgs f pname;"")
 
+(* Get the migration files, filter by .mo extension and sort lexicographically. *)
+let get_migration_files dir : string list =
+   if not (Sys.is_directory dir) then
+    begin
+      Format.eprintf "The enhanced migration directory is not a directory.\n"; exit 1
```
Kamirus: We need to avoid `exit` otherwise moc.js and then vscode extension just dies.
Use `Exit.exit` or do `open Exit` on top of this file

src/pipeline/pipeline.ml
```diff
@@ -320,6 +324,11 @@ let validate_stab_sig s : unit Diag.result =
     | PrePost (pre1, post1), PrePost (pre2, post2) ->
       (* check we can at least self-upgrade,
          with a possibly different or no migration function *)
+       Stability.match_stab_sig (Single post1) (Single post2)
+    | Multi {chain=c1; post=post1}, Multi {chain=c2;post=post2} ->
+      (* check we can at least self-upgrade,
+         with a possibly different or no migration function *)
+      (* TODO: do more checks *)
```
Kamirus: future work?

src/pipeline/resolve_import.ml
Kamirus: No, actually we should return `Diag.result` like other functions in this module do, to properly handle errors not just abort

doc/md/examples/grammar.txt
crusso: Good catch

src/pipeline/resolve_import.ml
crusso: Ok - I'll see what I can do. Or maybe move this check to the command-line arg code in moc.ml

src/pipeline/pipeline.ml
crusso: I'll think of something or just remove the comment. It's just a quick check the emitted signature can be parsed. Better might be to parse and check it against the type representation that produced it. Or just drop the check

src/pipeline/resolve_import.ml
Kamirus: I think refactoring it to a normal `Diag` error would be the best as it would also work in the vscode extension and it would not abort the whole process 
It would look like this
```ocaml
let get_migration_files dir : string list Diag.result =
  if not (Sys.file_exists dir && Sys.is_directory dir) then
    Diag.error Source.no_region code "import" (* pick appropriate code *)
      (Printf.sprintf "enhanced migration path '%s' is not a directory" dir)
  else
    Sys.readdir dir
    |> Array.to_list
    |> List.sort String.compare
    |> List.filter (fun fname -> Filename.check_suffix fname ".mo")
    |> List.map (Filename.concat dir)
    |> Diag.return
```
then used like

```ocaml
let enhanced_migration_imports flags : resolved_import list Diag.result =
  match flags.enhanced_migration_path with
  | None -> Diag.return []
  | Some dir ->
    let open Diag.Syntax in
    let* migration_files = get_migration_files dir in
    Diag.return (List.map (fun path -> LibPath {package = None; path = path}) migration_files)
```

```ocaml
let* migration_imports = enhanced_migration_imports flags in
imported := (List.fold_right (fun ri rim -> RIM.add ri Source.no_region rim)
               migration_imports !imported);
```

src/pipeline/pipeline.ml
```diff
@@ -141,18 +141,21 @@ let parse_verification_file = parse_file' Lexer.mode_verification
...
 type resolve_result = (Syntax.prog * ResolveImport.resolved_imports) Diag.result
 
-let resolve_flags pkg_opt =
+let resolve_flags ?(enhanced_migration_path=None) pkg_opt  =
   ResolveImport.{
     package_urls = !Flags.package_urls;
     actor_aliases = !Flags.actor_aliases;
     actor_idl_path = !Flags.actor_idl_path;
     include_all_libs = pkg_opt = None && Flags.(!all_libs || !ai_errors || Option.is_some !implicit_package);
+    enhanced_migration_path = enhanced_migration_path
```
Kamirus: Why not just this single line change? Then the `~enhanced_migration_path` parameter is not needed right? because this is not used in other place right?
```suggestion
    enhanced_migration_path = !Flags.enhanced_migration
```

src/mo_values/prim.ml
```diff
@@ -341,6 +341,8 @@ let prim trap =
     | _ -> assert false
     )
 
+  | "get_migrations" ->
+    fun _ v k -> as_unit v; k Null
```
Kamirus, line 345: `set_migrations` not needed?

doc/md/examples/grammar.txt
crusso: Fixed

src/pipeline/pipeline.ml
crusso: I think this because I want the flag to only pertain to the main actor, not any imported actor classes. 
(Personally, I would have preferred the migrations to be in a folder of the same name main/*, next to the actor main.mo, and similarly for any imported classes.

I wonder it their is another calls to resolve_flags (without the default override), that is unchange so not shown in the diff.

src/pipeline/pipeline.ml
crusso: Right, I'll mark it FUTURE, not TODO

src/pipeline/resolve_import.ml
crusso: Thanks for that! Done!

src/pipeline/pipeline.ml
crusso: Ok, I made the arg mandatory (and removed _path) to make it more obvious where it's passed as None (for imports)

src/mo_values/prim.ml
crusso: None of this stuff will work properly in the interpreter without a lot of work (upgrades never did) but I'll implement it as doing nothing

src/pipeline/pipeline.ml
crusso: Actually, I just deleted the comment. Checking the signature already validates the chain. Not sure what else to check.

src/mo_frontend/parser.mly
```diff
@@ -956,11 +958,19 @@ func_pat :
 dec_var :
   | VAR x=id t=annot_opt EQ e=exp(ob)
     { VarD(x, annot_exp e t) @? at $sloc }
+  | VAR x=id t=annot_opt
```
Kamirus: So we allow parsing when the `annot` is missing? Do we catch these cases later in `typing.ml`?

src/mo_frontend/parser.mly
```diff
@@ -956,11 +958,19 @@ func_pat :
...
+    (* Type checker will verify this is only allowed for stable variables with --enhanced-migration *)
+    { let init_exp = PrimE "_" @? at $sloc in
+      VarD(x, annot_exp init_exp t) @? at $sloc }
 
 dec_nonvar :
   | LET p=pat EQ e=exp(ob)
     { let p', e' = normalize_let p e in
       LetD (p', e', None) @? at $sloc }
+  | LET p=pat
+    { let p', e' = normalize_let p (PrimE "_" @? at $sloc) in
```
Kamirus, line 973: Hmmm this `p=pat` is an arbitrary pattern so we allow `PrimE "_"` to appear in a bigger context than I've thought.
I thought we only allow `let/var field : Type;` but it can be an arbitrary pattern for `let` right? Do we handle all cases properly in typing.ml?

src/mo_frontend/typing.ml
```diff
@@ -1773,7 +1775,8 @@ and infer_exp'' env exp : T.typ =
   match exp.it with
   | HoleE (_, e) ->
     (* TODO: this should probably be an assert, not an error *)
-    error env exp.at "M0232" "cannot infer type of implicit argument"
+      error env exp.at "M0232" "cannot infer type of implicit argument"
+  | PrimE "_" -> T.Non
```
Kamirus, line 1738: `PrimE "_"` gets inferred to `T.Non` without any checks, this feels a bit fragile because we should only allow this prim to typecheck properly for a var/let dec in an actor/mixin.
Right now we always infer the prim and then have fragile checks for "is a var/let dec in actor/mixin".
If we ever change the AST and forget to update the check then we would have the code typechecking.
But maybe there are enough checks to detect this?

src/mo_frontend/typing.ml
```diff
@@ -2274,7 +2277,14 @@ and infer_exp'' env exp : T.typ =
...
     let t = check_typ env typ in
-    if not env.pre then check_exp_strong env t exp1;
+    if not env.pre then
+      (* If exp1 is PrimE "_" (placeholder for no initializer) and --enhanced-migration is enabled, *)
+      (* allow it without type checking (it's used for stable variables without initializers) *)
+      (match exp1.it, !Flags.enhanced_migration with
+       | PrimE "_", Some _ ->
+         exp1.note <- {exp1.note with note_typ = t}
+       | _ ->
+         check_exp_strong env t exp1);
```
Kamirus: this check only works when `PrimE "_"` is wrapped with `AnnotE`.
Can this prim appear without annot? or appear in other context?

src/mo_frontend/typing.ml
```diff
@@ -4186,6 +4309,7 @@ and infer_dec env dec : T.typ =
         check_exp env T.Non fail
     );
     let t = infer_exp env exp in
+    if not env.pre then check_init env exp;
```
Kamirus: Are these two cases the only places where `PrimE "_"` is allowed?
Maybe this should be the only place where we typecheck this prim?
Also `check_init` only works for `annot` on this `prim`, what if there is no `annot`?

src/mo_frontend/parser.mly
Kamirus: I like when parser is more forgiving, because parsing errors are a nightmare - but we should make sure we handle all cases in typing.ml later (only allow what we support)

crusso: Oh, good catch. That wasn't my understanding and I didn't notice. I assumed the type annotation was required.

I do (try) to restrict the patterns to only appear on stable variables (in typing.ml). There's a few tests, but not enough.

src/mo_frontend/parser.mly
crusso: Actually, for stable variables we already have a restriction that the let pattern must be a variable, so we are probably covered against more general patterns.

src/mo_frontend/typing.ml
crusso: The parser introduces the PrimE "_" only in unitialized `let` and `var`. I'll try to restrict them to always be typed.

src/mo_frontend/typing.ml
crusso: Don't blame me - blame the AI

Kamirus: No blame just questions 😁 just sharing my doubts. I'll think how this prim typing can be refactored. The current state is not bad, I think it covers most cases maybe all actually

crusso: I've started on restricting the grammar and will add some more tests for uninitialized decs in non-actor positions. Instead of Prim  "_" (which was an improvement on the ill-typed unit value), the obvious option is to put the rhs of a `let` under an option but that would affect even more code I expect.

Tbh, I'm not super crazy about this design and, if we have the time to it again, would prefer something like

upgrade old = "mo:upgrade/previous" in

actor new {

    var f = Old.f.map…

}

With no awkward migration function, type of old inferred from previous.most and no weird restrictions on side-effects.

The imported actors full initializer would run (transitively if nested) to produce the value of old (its stable field).

I think we could even make this work with both actor classes and imported classes by requiring that the new actor class argument is a subtype of the old and passing it along to the previous intializer.

The stable signature would just be a record of stable variable record (not functions) labelled by version.

No weird migration functions that require explicit argument and return types and no weird restrictions on side effects, and no weird absent initializers.

Basically, the new actor receives the old values in a let bound 'Old'.

To auther an upgrade, we just add a new main new.mo and make that the root of compilation, keeping old.mo right next to it.

src/mo_frontend/parser.mly
crusso: We do now.

src/mo_frontend/parser.mly
crusso: I think we do now

src/mo_frontend/typing.ml
crusso: I think it's ok. The prim is only introduced by the parser in restricted positions. Of course, we shouldn't make it available via Prim.mo.

Another option would be to add a new AbsurdE.

src/mo_frontend/typing.ml
crusso: Fixed.

src/mo_frontend/typing.ml
crusso: This hack is no longer needed - the ai inserted it when it use the placeholder `()` but since Prim "_" : None <: T for any T we don't need the hack anymore (I hope)

src/mo_frontend/parser.mly
crusso: Fixed

alexandru-uta:
> They are implicit side effects in the initialiser that we don't prevent

In the ethos of allowing \<system\> calls in the body of the actor, I think we can do the same for pre/post-upgrade, they are "system" concepts anyhow. Also, I would call them quite explicit, it's not implied functionality.

src/mo_frontend/typing.ml
Kamirus, line 4003: This is unrelated change, I'd revert that, this PR is already huge. IMO this extra info whether it is implicit/explicit is not more helpful than the original message

src/mo_frontend/typing.ml
```diff
@@ -4205,8 +4205,13 @@ and check_stab env sort scope dec_fields =
       List.map (fun id -> {it = id; at = pat.at; note = ()}) ids;
     | (T.Actor | T.Mixin), Some {it = Flexible; _} , (VarD _ | LetD _) -> []
     | (T.Actor | T.Mixin), Some stab, _ ->
-      local_error env stab.at "M0133"
-        "misplaced stability modifier: allowed on var or simple let declarations only";
+      let at, desc =
```
crusso, line 4156: @Kamirus Did you want me to revert this (the `desc`)?

Kamirus: I meant reverting this whole block 4156-4162. This is not needed for this PR right? It is just an improvement unrelated to migrations, yes?

src/mo_frontend/typing.ml
```diff
@@ -4113,6 +4226,25 @@ and warn_unit_binding binder env (dec : dec) (exp : exp) =
...
+    if Option.is_none env.enhanced_migration || not env.in_actor then
+      local_error env at "M0257"
+        "this declaration has no initializer. Omitting an initializer is only allowed in an actor and with flag --enhanced-migration.";
+    (* check pattern is simple identifier pattern *)
+    (match pat_opt with
+     | None
+     | Some { it = VarP _; _} -> ()
+     | Some pat ->
+       local_error env at "M0258"
+         "this uninitialized `let` can only a simple identifier pattern `let <id> : <typ>`")
```
```suggestion
         "this uninitialized `let` can only use a simple identifier pattern `let <id> : <typ>`")
```

rts/motoko-rts/src/memory.rs
```diff
@@ -136,3 +136,21 @@ pub unsafe fn set_dedup_table<M: Memory>(mem: &mut M, dedup_table: Value) {
...
+    *get_migration_functions_ptr()
+}
+
+/// Set the migrations list.
+#[enhanced_orthogonal_persistence]
+#[ic_mem_fn]
+#[cfg(feature = "ic")]
+pub unsafe fn set_migrations<M: Memory>(mem: &mut M, dedup_table: Value) {
+    use crate::persistence::set_migration_functions_ptr;
+    set_migration_functions_ptr(mem, dedup_table);
```
Kamirus:
```suggestion
pub unsafe fn set_migrations<M: Memory>(mem: &mut M, migrations: Value) {
    use crate::persistence::set_migration_functions_ptr;
    set_migration_functions_ptr(mem, migrations);
```

src/mo_frontend/stability.ml
```diff
@@ -18,89 +18,106 @@ let display_typ = Lib.Format.display Pretty.pp_typ
...
         display_typ_expand tf2.typ
         (Pretty.string_of_explanation explanation)
         migration_link))
 
-let error_required s tf =
+let error_required s at mig_lab_opt tf =
   Diag.add_msg s
-    (Diag.error_message Source.no_region "M0169" cat
-      (Format.asprintf "the previous program version does not contain the stable variable %s. The migration function cannot require this variable as input, please see %s."
+    (Diag.error_message at "M0169" cat (*TODO: this code is duplicated of above? Renumber? *)
```
Kamirus: todo comment here, needs addressing?

src/mo_frontend/stability.ml
```diff
@@ -18,89 +18,106 @@ let display_typ = Lib.Format.display Pretty.pp_typ
 
 let display_typ_expand = Lib.Format.display Pretty.pp_typ_expand
 
-let error_discard s tf =
+let desc mig_lab_opt =
+  match mig_lab_opt with
+  | None -> "the previous version"
+  | Some mig_lab -> "version `" ^ mig_lab ^ "`"
+
+(*TODO: could we use tf.src.region to locate the errors better *)
```
Kamirus: future work?

src/mo_frontend/stability.ml
alexandru-uta: @crusso I think the todo is wrong; we can probably keep the code, looks like we had this same error even in master. Or, the code is wrong in master as well?

crusso: This was an existing bug.  I just added a  TODO for future work.

crusso: I just noticed that we were using the same code for different errors, that's all

src/mo_frontend/stability.ml
crusso:
```suggestion
(* FUTURE: we could perhaps use tf.src.region to better locate the errors below *)
```

src/mo_frontend/stability.ml
```diff
@@ -18,89 +18,106 @@ let display_typ = Lib.Format.display Pretty.pp_typ
...
         display_typ_expand tf2.typ
         (Pretty.string_of_explanation explanation)
         migration_link))
 
-let error_required s tf =
+let error_required s at mig_lab_opt tf =
   Diag.add_msg s
-    (Diag.error_message Source.no_region "M0169" cat
-      (Format.asprintf "the previous program version does not contain the stable variable %s. The migration function cannot require this variable as input, please see %s."
+    (Diag.error_message at "M0169" cat (*TODO: this code is duplicated of above? Renumber? *)
```
crusso:
```suggestion
    (Diag.error_message at "M0169" cat
```

src/mo_frontend/typing.ml
crusso: I've adjust the code to reduce the diff in test output, while still reporting a better location for default stable modifier that don't have a region

PR #5841, opened 2026-02-12
experiment: Structural comparison

Kamirus:


src/mo_types/type.ml
```diff
@@ -1006,6 +1006,43 @@ let is_local_async_func typ =
 let shared t = serializable false t
 let stable t = serializable true t
 
+(* Does this primitive type support ordering? *)
+let orderable_prim = function
+  | Nat | Nat8 | Nat16 | Nat32 | Nat64
+  | Int | Int8 | Int16 | Int32 | Int64
+  | Float | Char | Text | Blob | Principal -> true
+  | Null | Bool | Error | Region -> false
```
crusso: Why not Null and Bool? Regions also have an id so could in principle be compared by id, but maybe we want to exclude them because mutable.

Kamirus:
closed in favor of https://github.com/caffeinelabs/motoko/pull/5966

PR #5855, opened 2026-02-18
test: snapshot the output for the new pretty errors

christoph-dfinity:
Adds an extra `—check` pass with `—error-format=human` when `-t` is set for the run.sh script.

The output does not include the ANSI escapes for colors, as the `grace` library detects a non-tty output and disables colors. I think this is a good thing as it makes the output/diffs more readable.

---
This is **part 1 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #5858 
- <kbd>&nbsp;1&nbsp;</kbd> #5855 👈


Kamirus (approved):
LGTM!

PR #5856, opened 2026-02-18
experiment: multi-migration-v2

crusso:


design/multi-migration.md
```diff
@@ -0,0 +1,143 @@
+# Enhanced Multi-Migration (`--enhanced-migration <dir>`)
+
+## Overview
+
+The enhanced multi-migration feature allows canister developers to define a chain of
+incremental migration steps as separate Motoko modules. The compiler reads all `.mo`
+migration modules from the specified directory, sorts them lexicographically (by
+convention, files are timestamp-prefixed to ensure deterministic ordering), and
```
crusso, line 8: Not sure I understand the timestamp bit.

design/multi-migration.md
```diff
@@ -0,0 +1,143 @@
...
+compiles them into the actor as a migration chain.
+
+Each migration module exposes a `public func run(old : {...}) : {...}` that consumes
+a subset of the actor's stable fields and produces a (possibly different) subset.
+
+## Compile-Time Validation (`typing.ml`)
+
+- Each migration's `run` function must have stable object input/output types.
+- The chain composes: each migration's output is a subtype of the next migration's
+  input (`output[i] <: input[i+1]`).
```
crusso, line 18: This should probably be more than just subtyping, to accommodate change in mutability.

design/multi-migration.md
```diff
@@ -0,0 +1,143 @@
...
+## IR Generation (`desugar.ml`)
+
+The desugaring phase transforms the validated migration chain into IR code that
+executes the migrations at upgrade time. The pipeline has five stages:
+
+### 1. Read the Old State
+
+The compiler computes `enhanced_mem_ty` — the union of every field mentioned across
+all migrations and the actor, with last-writer-wins semantics for types — and passes
+it to `ICStableRead` to load the old actor's persistent heap object.
```
crusso, line 32: This seems wrong to me. If the field evolved from `Text` to `[Text]`, then ICStableRead would try to read a `[Text]` value from the old state. What if the migration function now tries to read the first char from the `[Text]`, thinking it was `Text`. It might work, since with eop, ICStableRead doesn't use the types, but the typing seems wrong.

crusso: Is there any check that the old actor is compatible with the input of the first migration function?

design/multi-migration.md
```diff
@@ -0,0 +1,143 @@
...
+  changes a field's type (e.g., `x : ?Text` → `x : ?[Text]`), the new type will
+  differ from the stored old type.
+- The RTS would trap with "Memory-incompatible program upgrade" before the migration
+  code gets a chance to run.
+- The compile-time chain validation already ensures the migration chain is
+  well-formed.
+- The IC provides atomic rollback: if the upgrade traps for any reason, the canister
+  state reverts entirely.
+
+Therefore, the enhanced migration path uses `assign_stable_type` (no compatibility
```
crusso, line 93: There's my answer - there is no check. Would it not make sense to check the input to the first migration is compatible with the old actor type?

crusso: Ah, maybe you don't need this because you assume you have the whole history of migrations starting from initial input `{}`. But what if we need to split the migration into two steps, because the entire chain requires too many cycles?

design/multi-migration.md
```diff
@@ -0,0 +1,143 @@
...
+## Migration Directory Convention
+
+```
+migrations/
+  20250101_000000_Init.mo          # {} -> { initial fields }
+  20250115_120000_AddProfile.mo    # { ... } -> { ... + profile fields }
+  20250201_090000_ChangeTypes.mo   # { x : Text, ... } -> { x : [Text], ... }
+```
+
+Files are sorted lexicographically. Timestamp prefixes ensure correct ordering and
```
crusso, line 123: Why not just number them?

src/lowering/desugar.ml
```diff
@@ -685,83 +693,226 @@ and build_actor at ts (exp_opt : Ir.exp option) self_id es obj_typ =
...
+           5. Final projection: CastPrim(?Any, ?T_actor) for each actor field *)
+      let chain = !T.migration_chain in
+
+      (* Step 1: compute concrete enhanced mem type (for ICStableRead) *)
+      let all_fields_tbl = Hashtbl.create 32 in
+      List.iter (fun (_file, _mod_typ, run_typ) ->
+        let dom_i, rng_i = T.as_mono_func_sub run_typ in
+        let (_ds, dom_fields_i) = T.as_obj (T.normalize dom_i) in
+        let (_rs, rng_fields_i) = T.as_obj (T.promote rng_i) in
+        List.iter (fun tf -> Hashtbl.replace all_fields_tbl tf.T.lab tf) dom_fields_i;
```
crusso, line 798: What if the same label is in both dom and rng?

PR #5861, opened 2026-02-19
feat: `include_all_libs` flag for better diagnostics

Kamirus:
Add `--all-libs` flag to load all library files from all packages, enabling better diagnostics, e.g. hinting at non-imported items (increases compilation time) 

To be used in the vscode extension to provide completions about not imported modules (otherwise the user needs to remember to import by hand).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,8 +1,9 @@
 name: "build"
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request: {}
+  merge_group: {}
 
 permissions:
   contents: write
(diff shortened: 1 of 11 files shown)


Changelog.md
```diff
@@ -4,6 +4,7 @@
 
   * perf: Optimise a few arithmetic/logic operations involving neutral elements (#5706).
   * Adds `--error-format human` option to print pretty errors with code snippets and labels (#5816).
+  * Preload all library files from packages by default, enabling hints about non-imported items in diagnostics. Use `--no-packages-preload` to disable (#5861).
```
christoph-dfinity: Need to mark this as breaking, and mention the flag as a remedy. Could also include a provision to make the flag a temporary band-aid that'll go away eventually? (Just to keep the flag soup under control)

Kamirus: Good! I've added a section about a potential breaking change, when it might happen and how to resolve that. Mentions  in both --help and changelog that "This is a temporary workaround and may be removed in a future major release"

christoph-dfinity: Nice, thanks

PR #5864, opened 2026-02-19
feat: prettier unused errors

christoph-dfinity:
Fixes LANG-1167

If you'd prefer a separate PR for the changes to test/bench, I'm happy to move the changes there.

### DIFF
--- a/test/bench/nat16.mo
+++ b/test/bench/nat16.mo
@@ -1,6 +1,5 @@
 // test small scalar ops
 import {
-  Array_tabulate;
   performanceCounter;
   debugPrint;
   rts_heap_size;
--- a/test/bench/ok/nat16.tc.ok
+++ b/test/bench/ok/nat16.tc.ok
@@ -1,1 +1,0 @@
-nat16.mo:3.3-3.17: warning [M0198], unused field Array_tabulate in object pattern (delete or rewrite as `Array_tabulate = _`)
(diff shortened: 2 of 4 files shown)


src/mo_frontend/typing.ml
```diff
@@ -359,28 +365,21 @@ let emit_unused_warnings env =
...
-      | Scope.MutableNotAssigned ->
-         warn env region "M0244" "variable %s is never reassigned, consider using `let`" id
-      | Scope.MixinIncluded -> ()
+    match kind with
+    | Scope.Declaration ->
+      (* NOTE: Emitting special error codes for bindings in shared patterns for easier detection for Caffeine *)
+      let err_code = if is_in_shared_pat region.left then "M0240" else "M0194" in
+      warn env region err_code
+        ~spans: [primary env region  "help: if this is intentional, prefix it with an underscore: `_%s`" id]
+        "unused identifier: `%s`" id
```
Kamirus, line 374: After the refactor the message has changed `... in shared pattern` is not longer there.
Accidental or deliberate change?

christoph-dfinity: Deliberate. What do you think? I think caffeine only really cares about the code, right? I think singling out shared patterns doesn't really matter to the human user.

Kamirus: Sure let's have it

PR #5865, opened 2026-02-19
prettier duplicate name errors

christoph-dfinity:
When specifying duplicated names use spans to point at the duplicated
location as well as the initial location.

Fixes LANG-1168

### DIFF
--- a/test/fail/duplicate-field.mo
+++ b/test/fail/duplicate-field.mo
@@ -9,6 +9,20 @@ type T = {foo : Int; foo: Bool}
 };
 
 do {
+type T = {
+  foo : Int;
+  bar1 : Text;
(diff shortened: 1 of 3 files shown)


test/fail/ok/bad-type-comp.tc-human.ok
```diff
@@ -22,7 +22,11 @@ because the type
 error[M0018]: duplicate type field name T in object type
     ┌─ bad-type-comp.mo:7:73
   7 │  ignore ((module { public type T = Null}) : module { type T = Null; type T = Null });
```
Kamirus, line 24: What happens when these two locations are on two separate lines far apart?

PR #5874, opened 2026-02-23
fix: only add synthetic primary span if there are no others

christoph-dfinity:
Purely affects the human error format. Only add the “synthetic” span covering the original specified range when there are no other primary spans specified.

### DIFF
--- a/src/lang_utils/diag.ml
+++ b/src/lang_utils/diag.ml
@@ -121,7 +121,13 @@ let fancy_of_message msg =
     if msg.spans = [] then
       [G.Diagnostic.Label.primaryf ~range:(range msg.at) ""]
     else
-      List.map mk_span msg.spans in
+      let spans = List.map mk_span msg.spans in
+      let primary_spans = List.filter (fun span -> span.prio = Primary) msg.spans in
+      if primary_spans = [] then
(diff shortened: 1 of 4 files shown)


src/lang_utils/diag.ml
```diff
@@ -121,7 +121,13 @@ let fancy_of_message msg =
     if msg.spans = [] then
       [G.Diagnostic.Label.primaryf ~range:(range msg.at) ""]
     else
-      List.map mk_span msg.spans in
+      let spans = List.map mk_span msg.spans in
+      let primary_spans = List.filter (fun span -> span.prio = Primary) msg.spans in
+      if primary_spans = [] then
+        G.Diagnostic.Label.primaryf ~range:(range msg.at) "" :: spans
+      else
+        spans
```
Kamirus, line 129: yes, this part overlaps with https://github.com/caffeinelabs/motoko/pull/5831/changes#diff-a091910b7eb4fddceb45441e1ec1021035372250c367f9e29a9bc23fbe7421fbR114-R116
I just extracted it to a `nonempty_spans` function to reuse with json output.
But I see that you add the Primary when there are no Primary spans. This is reasonable I think.
I'll adapt my function to this behavior and rename to `ensure_primary`

PR #5875, opened 2026-02-23
feat: prettier unknown identifier suggestions

christoph-dfinity:
Use the span to point at the identifier for suggestions

Fixes LANG-1172

### DIFF
--- a/src/mo_frontend/suggest.ml
+++ b/src/mo_frontend/suggest.ml
@@ -16,11 +16,7 @@ let suggest_id desc id ids =
       ids
   in
   if !Flags.ai_errors then
-    Some(
-      Printf.sprintf
-        "\nThe %s %s is not available. Try something else?"
-        desc
-        id)
+    Some(Printf.sprintf "The %s %s is not available. Try something else?" desc id)
   else
   let suggestions =
     let limit = Lib.Int.log2 (String.length id) in
--- a/test/fail/ok/suggest-label-ai.tc-human.ok
+++ b/test/fail/ok/suggest-label-ai.tc-human.ok
@@ -2,5 +2,4 @@ error[M0083]: unbound label fo in label environment:
   {foo : ()}
     ┌─ suggest-label-ai.mo:3:9
   3 │    break fo;
-    │          ^^ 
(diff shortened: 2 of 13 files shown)


test/fail/ok/bad-type-comp.tc.ok
```diff
@@ -15,11 +15,12 @@ because the type
  in `T`
 bad-type-comp.mo:7.73-7.74: type error [M0018], duplicate type field name T in object type
 bad-type-comp.mo:11.15-11.16: type error [M0029], unbound type T
-Did you mean type Text?
+help: did you mean type Text?
 bad-type-comp.mo:17.15-17.16: type error [M0029], unbound type U
+
```
Kamirus: this is adding empty lines, need to filter them out?

christoph-dfinity: I think it's the `Format.display_typ` adding forced line-breaks. I've been thinking about making the error messages hold a `format` instead of a `string`, so I can decide the line-breakiness when actually printing the error.

test/fail/ok/migration.tc.ok
```diff
@@ -10,13 +10,12 @@ migration.mo:3.7-13.7: type error [M0204], migration expression produces field `
   var [var (Nat, Text)]
 migration.mo:3.7-13.7: type error [M0205], migration expression produces unexpected field `unstable2` of type
   () -> ()
-
 The actor should declare a corresponding `stable` field.
+
```
Kamirus: Empty line because there is no field to suggest?
e.g. it would be `help: did you mean field version?` if there was a 'version' field to suggest

test/fail/ok/bad-type-comp.tc.ok
christoph-dfinity: Sorry, that was nonsense. I need to actually check if any suggestions are returned.

test/fail/ok/migration.tc.ok
christoph-dfinity: Okay, I think I fixed all of the spurious newlines. PTAL

Kamirus (approved):
LGTM!

PR #5881, opened 2026-02-26
prettier pattern type mismatch errors

christoph-dfinity:
Uses the new spans to move the mismatch closer to the offending pattern and show the full type of the matched expression as secondary context (very much inspired by the equivalent rustc error)

Fixes LANG-1178

### DIFF
--- a/test/fail/ok/M0113.tc-human.ok
+++ b/test/fail/ok/M0113.tc-human.ok
@@ -3,4 +3,4 @@ error[M0113]: object pattern cannot consume expected type
   1 │  let {foo} = true
     │      ^^^^^   ---- this expression has type `Bool`
     │      │
-    │      expected object type, got `Bool`
+    │      expected `Bool`, got object type
--- a/test/fail/ok/pat-inconsistent.tc-human.ok
+++ b/test/fail/ok/pat-inconsistent.tc-human.ok
@@ -4,7 +4,7 @@ error[M0113]: object pattern cannot consume expected type
     │          --------- this expression has type `{#A}`
  41 │    case (#A) {};
  42 │    case ({B = _}) {};
-    │          ^^^^^^^ expected object type, got `{#A}`
+    │          ^^^^^^^ expected `{#A}`, got object type
 error[M0050]: literal of type
   Nat
 does not have expected type
@@ -17,7 +17,7 @@ error[M0116]: variant pattern cannot consume expected type
  53 │  switch (true : Bool) {
     │          ----------- this expression has type `Bool`
  54 │    case (#geese) {};
-    │          ^^^^^^ expected `{#geese : _}`, got `Bool`
+    │          ^^^^^^ expected `Bool`, got `{#geese : _}`
 warning[M0146]: this pattern is never matched
     ┌─ pat-inconsistent.mo:59:9
  59 │    case (#geese) {};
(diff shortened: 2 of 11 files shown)


test/fail/ok/M0113.tc-human.ok
```diff
@@ -1,5 +1,6 @@
 error[M0113]: object pattern cannot consume expected type
-  Bool
     ┌─ M0113.mo:1:5
   1 │  let {foo} = true
-    │      ^^^^^ 
+    │      ^^^^^   ---- this expression has type `Bool`
```
Kamirus, line 4: This is reverted, yes? 😄

test/fail/ok/pat-inconsistent.tc-human.ok
```diff
@@ -11,10 +13,11 @@ does not have expected type
  49 │    case 1 {};
     │         ^ 
 error[M0116]: variant pattern cannot consume expected type
-  Bool
     ┌─ pat-inconsistent.mo:54:9
+ 53 │  switch (true : Bool) {
+    │          ----------- this expression has type `Bool`
  54 │    case (#geese) {};
-    │          ^^^^^^ 
+    │          ^^^^^^ expected `{#geese : _}`, got `Bool`
```
Kamirus: flipped here too

test/fail/ok/pat-inconsistent.tc-human.ok
```diff
@@ -34,3 +37,15 @@ warning[M0146]: this pattern is never matched
     ┌─ pat-inconsistent.mo:64:10
  64 │      case (#geese) {};
     │           ^^^^^^^^ 
+error[M0112]: tuple pattern cannot consume expected type
+    ┌─ pat-inconsistent.mo:68:7
+ 68 │    let (_, _) = 10;
+    │        ^^^^^^   -- this expression has type `Nat`
```
Kamirus, line 43: and here

test/fail/ok/M0113.tc-human.ok
christoph-dfinity: No, the `----` marker marks the `true` expression. I added two test cases to show what this looks like for nested patterns (that's where it gets more interesting to see the full type of the expression being matched on).

But the `expected ...` points at the pattern. The pattern "expects" a value of a certain type, but gets a different one.

Kamirus: ah, right!

test/fail/ok/pat-inconsistent.tc-human.ok
Kamirus: I'd rework this `expected ... got ...` message. It shows the inferred type from a pattern and says that it was expected. This is confusing to me. Maybe sth like:

```suggestion
    │          ^^^^^^ this pattern has type `{#geese : _}`, but the matched expression has type `Bool`
```

Or 


```suggestion
    │          ^^^^^^ this pattern has type `{#geese : _}`, expected `Bool`
```

because from the perspective of the pattern the *expected* type  is the type of the scrutinee, right?

christoph-dfinity: I thought about it for a bit and I think I agree that we should flip these... The type of the scrutinee is expected

PR #5893, opened 2026-03-05
chore: bump `pocket-ic`

ggreif:


christoph-dfinity:
Probably also need to update https://github.com/caffeinelabs/motoko/blob/3d6c4b1ee50b327f3a6991d7bb967b05be0345bd/flake.nix#L150-L156

PR #5899, opened 2026-03-10
refactor: check imports separately

### LINKED ISSUE #5897: Bug importing nested modules
Consider these two files:
```
import { X } "module";

persistent actor Main {
  type T = X.Y;
};
```
where the file `module.mo` is:
```
module {
  public let X = module {
    public type Y = Nat;
  };
}
```
It fails with 
```
cannot infer type of forward variable reference X
```
at the `X` of `X.Y`.

A couple of things make this error go away:

1. Removing the actor name `Main`. Writing `persistent actor {` in that line works.
2. Not using the pattern matching in the import. Importing as `import M "module"` and using `M.X.Y` works.

christoph-dfinity:
Imports are syntactically guaranteed to preceed any other declarations in the unit body. They are guaranteed to not be mutually recursive, so we can check them separately.

Fixes #5897


src/mo_frontend/typing.ml
```diff
@@ -4688,6 +4664,40 @@ and infer_dec_valdecs env dec : Scope.t =
...
+    assert (t' <> T.Pre);
+    exp.note <- {note_typ = t'; note_eff = T.Triv};
+    dec.note <- {empty_typ_note with note_typ = t; note_eff = T.Triv};
+    scope
+  | _ -> assert false
+
+let infer_imports env ds =
+  List.fold_left (fun acc d -> Scope.adjoin acc (infer_import env d)) Scope.empty ds
+
+let infer_prog' env at check_unused imports decls =
```
```suggestion
let infer_split_prog env at check_unused imports decls =
```

src/mo_frontend/typing.ml
```diff
@@ -4704,7 +4714,8 @@ let infer_prog ?(enable_type_recovery=false) scope pkg_opt async_cap prog
           let env = {
              env0 with async = async_cap; type_recovery = enable_type_recovery;
           } in
-          let t, sscope = infer_block env prog.it prog.at true in
+          let imports, decls = split_imports prog.it in
+          let t, sscope = infer_prog' env prog.at true imports decls in
```
```suggestion
          let t, sscope = infer_split_prog env prog.at true imports decls in
```

PR #5900, opened 2026-03-10
chore: simplify RTS for enhanced multi-migration

alexandru-uta:


src/prelude/internals.mo
```diff
@@ -860,3 +860,50 @@ func @dedup(b : Blob) : Blob {
 
   result;
 };
+
+/// Function that returns the ID (e.g., file name) of the last applied migration.
+func @checkLastMigration(mId : Text) : Bool {
+  /// For future use, instead of keeping around just the last migration,
+  /// we keep a list of all the migrations applied, not locking us
+  /// out of showing the history of applied migrations or other functionality.
+  type MigrationsList = {
```
crusso:
```suggestion
  type MigrationsList = ?(Text,MigrationsList)
```
Might be a bit simpler and more idiomatic. But doesn't really matter much I guess.

src/prelude/internals.mo
```diff
@@ -860,3 +860,50 @@ func @dedup(b : Blob) : Blob {
...
+    migrationId : Text;
+    next : ?MigrationsList;
+  };
+
+  func getMigrationsObj() : ?MigrationsList {
+    (prim "get_migrations" : () -> ?MigrationsList)();
+  };
```
crusso:
```suggestion
```
You could just inline the prim below

src/prelude/internals.mo
```diff
@@ -860,3 +860,50 @@ func @dedup(b : Blob) : Blob {
...
+
+  // Get the migrations list from the RTS.
+  let ?list = getMigrationsObj() else {
```
crusso:
```suggestion
  let ?list = (prim "get_migrations" : () -> ?MigrationsList)() else {
```

src/prelude/internals.mo
```diff
@@ -860,3 +860,50 @@ func @dedup(b : Blob) : Blob {
...
+    migrationId : Text;
+    next : ?MigrationsList;
+  };
+
+  func getMigrationsObj() : ?MigrationsList {
+    (prim "get_migrations" : () -> ?MigrationsList)();
+  };
+  func setMigrationsObj(obj : MigrationsList) {
+    (prim "set_migrations" : (MigrationsList) -> ())(obj);
+  };
```
crusso:
```suggestion
```
delete and inline below

src/codegen/compile_enhanced.ml
```diff
@@ -12230,16 +12230,14 @@ and compile_prim_invocation (env : E.t) ae p es at =
     compile_exp_vanilla env ae dedup_table ^^
     E.call_import env "rts" "set_dedup_table"
 
-  | OtherPrim "register_migration", [hash] ->
-    SR.unit,
-    compile_exp_vanilla env ae hash ^^
-    E.call_import env "rts" "register_migration"
-
-  | OtherPrim "was_migration_performed", [hash] ->
+  | OtherPrim "get_migrations", [] ->
```
crusso, line 12215: It's possible we'll need dummy implementations for these in compile_classical too, unless the helpers that use them get dead-coded.

crusso: They could probably just trap.

alexandru-uta: we guard the --enhanced-migration flag with the --eop flag AFAIK, isn't that enough? we shouldn't be needing this I assume?

crusso: internals.mo contains (new) functions that call the new prims. If they are dead-coded by the time we generate code, that's not a problem. But it they aren't, then the classical backend might need to generate code for the prims, even if they are unreachable at runtime.

src/prelude/internals.mo
alexandru-uta: changed.

src/prelude/internals.mo
```diff
@@ -860,3 +860,30 @@ func @dedup(b : Blob) : Blob {
...
+/// Function that returns the ID (e.g., file name) of the last applied migration.
+func @setLastMigration(mId : Text) {
+  /// For future use, instead of keeping around just the last migration,
+  /// we keep a list of all the migrations applied, not locking us
+  /// out of showing the history of applied migrations or other functionality.
+  type MigrationsList = ?(Text, MigrationsList);
+
+  // Add the new migration ID at the beginning of the list.
+  let newList = (mId, (prim "get_migrations" : () -> MigrationsList)());
+  (prim "set_migrations" : (MigrationsList) -> ())(?newList);
```
crusso:
```suggestion
  let newList = ?(mId, (prim "get_migrations" : () -> MigrationsList)());
  (prim "set_migrations" : (MigrationsList) -> ())(newList);
```

src/prelude/internals.mo
```diff
@@ -860,3 +860,30 @@ func @dedup(b : Blob) : Blob {
 
   result;
 };
+
+/// Function that returns the ID (e.g., file name) of the last applied migration.
+func @checkLastMigration(mId : Text) : Bool {
```
crusso: Actually, we probably don't even need this and can move it all into the backend. I can do it after we merge.

crusso: Done in https://github.com/caffeinelabs/motoko/pull/5900/changes/08b451c69fac22dc3c46233c63f740c3e2696cc7

PR #5901, opened 2026-03-10
refactor: multimigration v2 static checks and typing.ml

crusso:
Builds on #5894 (which refactors desugar.ml)

Refactors typing.ml (mostly) to re-use more code and do chain checking in reverse order (as elsewhere).

Unlike original code, only warns if initial pre is non-empty so the chain can only be used for an upgrade of non-empty actor, not a fresh install. This is useful for dropping canister history by chopping a prefix of the migration chain.

 (affects test drun/enhanced-migration-bad-chain.drun  (version2 half) 

- [x] Fix global Type.migration_chain that breaks imported actor classes 
- [x] Remove code duplication in moc.ml and maybe pipeline.ml

### DIFF
--- a/test/cmp/enhanced-migration-incremental-v1-v1.cmp
+++ b/test/cmp/enhanced-migration-incremental-v1-v1.cmp
@@ -1,0 +1,1 @@
+files/enhanced-migration-incremental/version1.drun.most files/enhanced-migration-incremental/version1.drun.most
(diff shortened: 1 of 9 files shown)


src/mo_frontend/stability.ml
```diff
@@ -59,42 +69,46 @@ let error_required s tf =
    - Lossy promotion to any or dropping record fields is rejected (stricter than subtyping to prevent data loss).
  *)
 
+let match_stab_fields s at mig_lab_opt tfs1 tfs2 =
```
christoph-dfinity, line 72: Can this be written in terms of `align_fields`?

crusso: Yeah, probably. But it wasn't before ;->

crusso: Done it now. Indeed nicer.

PR #5903, opened 2026-03-11
feat: implicit argument derivation and structural record synthesis

Kamirus:
## Summary

Extends the Motoko compiler's implicit argument resolution with two new mechanisms:

### 1. Implicit derivation from functions with inner implicits

The compiler can now resolve implicit arguments by composing them from functions (possibly polymorphic) that themselves have implicit parameters. For example, an implicit `compare` for `[Nat]` is automatically derived from `Array.compare<Nat>` + `Nat.compare`, eliminating boilerplate wrapper modules. Works transitively (e.g., `[[Nat]]`). Depth-limited with `--implicit-derivation-depth` (default: 5).

**Before:**
```motoko
module MyArray {
  public func compare(a : [Nat], b : [Nat]) : Order { Array.compare(a, b) };
};
let m = Map.empty<[Nat], Text>();
m.add([1, 2, 3], "abc"); // uses MyArray.compare
```

**After:**
```motoko
let m = Map.empty<[Nat], Text>();
m.add([1, 2, 3], "abc"); // just works — derived from Array.compare<Nat> + Nat.compare
```

### 2. Structural derivation for records and tuples

A function whose sole explicit parameter is named `__record` (typed `[(Text, T)] -> R`) or `__tuple` (typed `[T] -> R`) acts as a *structural combiner*. When the compiler needs an implicit for a record or tuple type, it automatically decomposes the type, resolves a per-field/per-element implicit (using the same search label), and synthesizes a wrapper. Both unary (`X -> R`) and binary (`(X, X) -> R`) hole types are supported.

**Example — generic JSON serialization for any record:**
```motoko
import Json "mo:json/Json";
import IntJson "mo:json/IntJson";
import TextJson "mo:json/TextJson";

type Person = { name : Text; age : Int };

let p : Person = { name = "Alice"; age = 30 };
let json = p.toJson();
// #obj([("name", #text "Alice"), ("age", #number 30)])
```

The `Json` package defines a single `_toJson(__record : [(Text, Json)]) : Json` combiner. That's all — the compiler handles every record type automatically, as long as each field type has a `_toJson` instance.

## Resolution order

Direct matches are always preferred over derived ones. Within each tier, the most specific candidate (by subtyping) wins:

1. Direct local values
2. Direct module fields
3. Direct library fields (requires `--implicit-package`)
4. Derived from local values
5. Derived from module fields
6. Derived from library fields (requires `--implicit-package`)
7. Structural from local values (`__record` / `__tuple` combiner)
8. Structural from module fields
9. Structural from library fields (requires `--implicit-package`)

### Run tests
- `test/run/implicit-derivation.mo` — basic derivation, monomorphic, polymorphic, multiple implicits, priority ordering, subtyping
- `test/run/implicit-derivation-transitive.mo` — `[[Nat]]` through two derivation levels
- `test/run/implicit-derivation-recursive.mo` — single recursion, mutual recursion, recursive trees
- `test/run/implicit-derivation-record-variant.mo` — real-world record/variant comparison with `compareBy`
- `test/run/implicit-derivation-core.mo` — integration with `mo:core` library (`Array`, `Nat`, `Int`, `Text`)
- `test/run/implicit-derivation-implicit-package.mo` — derivation from unimported library modules
- `test/run/implicit-derivation-json.mo` — structural derivation for JSON serialization (records, tuples, maps, lists)
- `test/run/implicit-derivation-structural.mo` — basic structural record derivation (unary)
- `test/run/implicit-derivation-structural-compare.mo` — binary structural derivation for record/tuple comparison
- `test/run/implicit-derivation-structural-dispatch.mo` — `__record` vs `__tuple` dispatch and binary/unary disambiguation

### Fail tests (14 files)
- `test/fail/implicit-derivation.mo` — missing inner implicit, basic derivation errors
- `test/fail/implicit-derivation-ambiguous.mo` — ambiguous derived candidates at the head level
- `test/fail/implicit-derivation-ambiguous-deep.mo` — ambiguous inner implicit during derivation
- `test/fail/implicit-derivation-bimatch.mo` — bi-matching limits during type instantiation
- `test/fail/implicit-derivation-deep.mo` — derivation chain errors at multiple depths
- `test/fail/implicit-derivation-depth.mo` — `--implicit-derivation-depth` limit hit
- `test/fail/implicit-derivation-no-backtracking.mo` — no backtracking across derivation candidates
- `test/fail/implicit-derivation-json.mo` — real-world JSON errors: missing leaves, expanded Map types, nested records
- `test/fail/implicit-derivation-structural-missing.mo` — record with a field type that has no instance
- `test/fail/implicit-derivation-structural-ambiguous.mo` — two structural combiners in scope
- `test/fail/implicit-derivation-structural-hint.mo` — structural combiner in an unimported library (import hint)
- `test/fail/implicit-derivation-structural-depth.mo` — structural derivation depth limit
- `test/fail/implicit-derivation-structural-record2-missing.mo` — binary record derivation with missing field instance
- `test/fail/implicit-derivation-structural-tuple-missing.mo` — tuple derivation with missing element instance

### Test package
- `test/json-stub/` — minimal JSON package used by the structural derivation tests; demonstrates the `__record` convention and companion per-type modules (`IntJson`, `TextJson`, `ArrayJson`, `ListJson`, `MapJson`, `Tuple2Json`, `Tuple3Json`)


doc/md/fundamentals/11-implicit-parameters.md
```diff
@@ -167,13 +167,119 @@ The compiler infers an implicit argument by:
...
+
+The resolution depth is bounded to guarantee termination. If you encounter a depth limit, you can increase it with `--implicit-derivation-depth` or provide the argument explicitly.
+
+When derivation is attempted but fails (for example, because an inner implicit can't be resolved), the compiler includes this context in the error message, telling you which candidate was tried and which inner implicit was missing.
+
+### Structural derivation
+
+When an implicit is needed for a **record type**, the compiler can synthesise it automatically using a *structural combiner* — a function that converts a list of (field-name, converted-value) pairs into the target type.
+
+A structural combiner is any function whose sole explicit parameter is named `__record` and has type `[(Text, T)] -> R` for some element type `T` and result type `R`. The `__` prefix signals to the compiler that this function acts as a record-level builder.
```
crusso: Cool that you've got something generic working.

Is `T` ever different from `R`?

For an efficient derived `compare`, we may want to be less eager in computing the results from the fields, only compare later fields if needed. I  wonder if there is a nice way to do this.

Kamirus: `T` and `R` could be different but are usually equal I guess
e.g. `impl` hole on a record `{ a : A }` could have a combiner like `[(Text, T)] -> [T]`
`impl` on `A` would produce `T` but
`impl` on the whole record would produce `[T]`.

I would not worry about getting efficient compare of a record directly from this generic algorithm.
A fancy efficient package could provide some "lazy comparison results" that would make the combiner effectively launch the comparisons for each element one by one if needed, stopping on the first -- it would create some thunks which might have more overhead but that is a separate issue 😄

doc/md/fundamentals/11-implicit-parameters.md
```diff
@@ -167,13 +167,212 @@ The compiler infers an implicit argument by:
...
+
+When an implicit is needed for a **record or tuple type**, the compiler can synthesize it automatically using a *structural combiner* — a function whose single parameter name begins with `__` and encodes the structural decomposition kind. Structural combiners must not have implicit parameters.
+
+Two structural kinds are supported, distinguished by the combiner's parameter name:
+
+| Parameter name | Combiner type         | Hole type matched                        | Description                                    |
+|----------------|-----------------------|------------------------------------------|------------------------------------------------|
+| `__record`     | `[(Text, E)] -> R`    | `Rec -> R` or `(Rec, Rec) -> R`          | Record: one or two records, arity from hole    |
+| `__tuple`      | `[E] -> R`            | `(A, B, ...) -> R` or `((A,B,...), (A,B,...)) -> R` (≥ 2 elements) | Tuple: one implicit per element |
+| `__variant`    | —                     | —                                        | Reserved for future extension                  |
```
crusso:
```
| `__variant`    | —  (Nat, Text, () -> R)   | — `Var -> R`  or `(Var, Var) -> R`                                    | Reserved for future extension                  |
```
Might work for variants.... (Nat is the ordinal of the tag, which, for records, can be inferred from the array index)

Maybe arrays should just be delayed (or use an iterator if ok with forcing the ordering)
         `[(Text, () -> E)] -> R`

Kamirus: That's a really good idea to be go with lazy values `() -> R`, it'd give more flexibility to package authors (and they are the ones who write these `__record/tuple/variant` functions).
Better than `Iter` because they can choose which one to evaluate and when, yet still simple enough!

Just for the record: I'd implement variants in another PR. But I'm curious - Why would anyone need the `Nat` ordinal of the tag? Aren't we sorting tags in a variant anyway so that would make it tricky to preserve, right? Or am I misunderstanding or missing sth?

crusso: For efficient compare I would compare the ordinals before comparing the contents when ordinals match. Hash would work too but maybe more arbitrary and implementation specific than the ordinal of the label in the variants (which should be know statically, right?).

src/mo_frontend/typing.ml
```diff
@@ -1469,80 +1472,387 @@ let is_val_module (n, ((t, _, _, _) : val_info)) =
...
+
+  let as_structural_combiner_typ candidate_typ =
+    let with_thunk_elem kind thunk_typ ret_typ =
+      match T.normalize thunk_typ with
+      | T.Func (T.Local, T.Returns, [], [], [elem_typ]) ->
+        Some (kind, elem_typ, ret_typ)
+      | _ -> None in
+    match T.promote candidate_typ with
+    | T.Func (T.Local, T.Returns, [], [T.Named ("__record", inner_typ)], [ret_typ]) ->
+      (match T.promote inner_typ with
```
crusso: dubious promote on negative occurrence?

src/mo_frontend/typing.ml
```diff
@@ -1469,80 +1472,387 @@ let is_val_module (n, ((t, _, _, _) : val_info)) =
...
+      | T.Array (T.Tup [txt; thunk_typ]) when T.normalize txt = T.Prim T.Text ->
+        with_thunk_elem `Record thunk_typ ret_typ
+      | _ -> None)
+    | T.Func (T.Local, T.Returns, [], [T.Named ("__tuple", inner_typ)], [ret_typ]) ->
+      (match T.promote inner_typ with
```
crusso: Ditto: dubious promote on negative occurrence?

crusso:
Is there any way to present this PR as a diff to part 1 #5966? I tried to open a PR with this branch and the branch of #5966 as a base but it didn't reduce the diff for me.

Kamirus:
> Is there any way to present this PR as a diff to part 1 #5966? I tried to open a PR with this branch and the branch of #5966 as a base but it didn't reduce the diff for me.

Not easily no, after we merge the part1 then I'll update this PR
Because after reviews and such the part 1 got updates but they were not reapplied on this branch

nix/tests.nix
```diff
@@ -43,7 +43,8 @@ let
         type != "directory"
         || hasPrefix "${dir}/" "${relPath}/"
         || hasPrefix "core-stub/" "${relPath}/"
-        || hasPrefix "base-stub/" "${relPath}/";
+        || hasPrefix "base-stub/" "${relPath}/"
+        || hasPrefix "json-stub/" "${relPath}/";
```
christoph-dfinity, line 47: I don't think we should be adding to these. The top-level directories under `test/` would ideally just be test suites.

I'd rather get rid of the stubs at some point and just replace them with our `$MOTOKO_CORE` that's around in the tests.

test/json-stub/src/MapJson.mo
```diff
@@ -0,0 +1,14 @@
...
+module {
+  public func _toJson<K, V>(
+    self : Map.Map<K, V>,
+    toText : (implicit : K -> Text),
+    _toJson : (implicit : V -> Json.Json),
+  ) : Json.Json {
+    #obj(
+      self.entries().map(func(k, v) { (toText(k), _toJson(v)) }).toArray()
+    );
+  };
```
christoph-dfinity: mostly a nit as this is for tests, but this a poor encoding of `Map`, as you can't guarantee an injective `toText` for the key type.

If we're worried about AI's learning from this as an example it should take `toJson` for the key as well and encode to a shape of `[(Json, Json)]`

christoph-dfinity (approved):
I'm still a bit worried about code size/perf degradation if this feature is applied liberally in a large-ish codebase. If that becomes a problem I'll rely on you to write the skills to guide the AI.

Some smaller change requests, but looks good otherwise.

test/run/implicit-derivation-json.mo
```diff
@@ -0,0 +1,105 @@
...
+import Int "mo:core/Int";
+import Text "mo:core/Text";
+import Json "mo:json/Json";
+import RecordJson "mo:json/RecordJson";
+import IntJson "mo:json/IntJson";
+import TextJson "mo:json/TextJson";
+import ArrayJson "mo:json/ArrayJson";
+import ListJson "mo:json/ListJson";
+import MapJson "mo:json/MapJson";
+import TupleJson "mo:json/TupleJson";
```
Kamirus, line 14: Currently this multiple-imports and flat-modules hierarchy is the recommended way because of how our import hints work.
Our hints look like this currently `Did you mean to import mo:json/MapJson?`
Nested modules won't show up in these hints so we either accept this style or redesign.

PR #5906, opened 2026-03-12
feat: add `Float32` primitive type with arithmetics and conversions to `Float`

ggreif:
Implement some `Float32` functionality (basic arithmetics). This is experimental for now and somewhat starved in terms of supporting functions in the prelude (though basic arithmetic is there, everything else must pass via `Float` if missing, e.g. primitives like `sin`). But it should be efficient memory consumption-wise.

Can receive and send `float32` by Candid messages. This was the main motivator for the feature. Stable variables of `Float32` type work too.

TODO:
- [x] `safeFloatToFloat32 : Float -> Epsilon -> ?Float32`
- [ ] verify that runtime tagging works too — see 0abf97e26312f28cf3a36e363a9334546f4afcb9 + follow-ups (OBSOLETE, supplanted by f32 scalars on 64-bit backend)
- [x] `.most` process working?
- [x] (re)move plans (**opportunities inside**, please consult the last commit(s) about the unboxed `Float32` idea, and the unboxed `Float` – analogously on 64-bit backend!)


src/codegen/compile_enhanced.ml
```diff
@@ -11774,6 +11796,18 @@ and compile_prim_invocation (env : E.t) ae p es at =
       SR.UnboxedFloat64,
       compile_exp_as env ae (SR.UnboxedWord64 Int64) e ^^
       G.i (Convert (Wasm_exts.Values.F64 F64Op.ConvertSI64))
+
+    | Float, Float32 ->
+      SR.UnboxedFloat64,
+      compile_exp_as env ae SR.UnboxedFloat64 e ^^
+      G.i (Convert (Wasm_exts.Values.F32 F32Op.DemoteF64)) ^^
```
ggreif: To remove excess precision.

test/run-drun/float32-stable/main.mo
```diff
@@ -0,0 +1,21 @@
+import { debugPrint; floatToFloat32; float32ToFloat } = "mo:⛔";
+
+actor {
+  // scalar: exactly representable in f32
+  stable var x : Float32 = 1.5;
```
ggreif, line 5: Avoiding pilot errors. I checked the section contents:
```
$ wasm-objdump test/run-drun/_out/float32-stable/main.drun.wasm -j "icp:private motoko:stable-types" -s

main.drun.wasm:	file format wasm 0x1

Contents of section Custom:
0034054: 1f69 6370 3a70 7269 7661 7465 206d 6f74  .icp:private mot
0034064: 6f6b 6f3a 7374 6162 6c65 2d74 7970 6573  oko:stable-types
0034074: 2f2f 2056 6572 7369 6f6e 3a20 312e 302e  // Version: 1.0.
0034084: 300a 6163 746f 7220 7b0a 2020 7374 6162  0.actor {.  stab
0034094: 6c65 2076 6172 2061 7272 203a 205b 7661  le var arr : [va
00340a4: 7220 466c 6f61 7433 325d 3b0a 2020 7374  r Float32];.  st
00340b4: 6162 6c65 2076 6172 2078 203a 2046 6c6f  able var x : Flo
00340c4: 6174 3332 0a7d 3b0a                      at32.};.
```

src/prelude/prim.mo
```diff
@@ -531,6 +531,12 @@ func btstInt64(w : Int64, amount : Int64) : Bool = (prim "btstInt64" : (Int64, I
 func floatToFloat32(f : Float) : Float32 = (prim "num_conv_Float_Float32" : Float -> Float32) f;
 func float32ToFloat(f : Float32) : Float = (prim "num_conv_Float32_Float" : Float32 -> Float) f;
 
+// Checked Float -> Float32 conversion: returns ?Float32 if round-trip stays within epsilon, null otherwise
+func safeFloatToFloat32(f : Float, epsilon : Float) : ?Float32 {
+  let f32 = floatToFloat32 f;
```
ggreif, line 536: Nicely inlined:
``` Wasm
(func $safeFloatToFloat32 … (param $f i32) (param $epsilon i32) …
  local.get $f
  f64.load offset=5          ;; unbox f
  f32.demote_f64             ;; floatToFloat32 — inline
  f64.promote_f32            ;; float32ToFloat — inline (round-trip)
  local.tee $f32
  local.get $f
  f64.load offset=5
  f64.sub                    ;; round_trip − f — inline
  f64.abs                    ;; floatAbs — inline
  local.get $epsilon
  f64.load offset=5          ;; unbox epsilon
  f64.le                     ;; comparison — inline
  if … box ?f32 … else … null …
```

.claude/plans/float32-classical.md
```diff
@@ -0,0 +1,346 @@
...
+
+`const_lit_of_lit (Float32Lit f)` is the natural split point.  Check bit 0 of the F32 bit
+pattern at OCaml compile time:
+
+```ocaml
+| Float32Lit f ->
+  let bits = Int32.bits_of_float (Numerics.Float32.to_float f) in
+  if Int32.logand bits 1l = 0l
+  then Const.Vanilla bits   (* scalar: materialises as a Wasm i32.const / f32.const *)
+  else Const.Float32 f      (* heap:   shared_static_obj Bits32 F on first use *)
```
crusso: This smells funny to me. What's going on here and is this actually in the code?

.claude/plans/float32-classical.md
```diff
@@ -0,0 +1,346 @@
...
+IEEE 754 single-precision (F32):
+```
+ 31  30    23  22                    0
+  ┌───┬────────┬──────────────────────┐
+  │ S │ exp(8) │    mantissa (23)     │
+  └───┴────────┴──────────────────────┘
+                                     ╰── bit 0 is the LSBit of the mantissa
+```
+
+#### The scheme
```
crusso: This seems like overkill. Why don't we just box *all* float32 on 32-bit.
Also, I don't think this work with rtti on in 32-bit, since now you can't unboxed distinguish floats from other values since you've used 31 bits for the value.

.claude/plans/float32-classical.md
```diff
@@ -0,0 +1,346 @@
...
+| | Current (`Bits32 F` always) | Scalar-when-possible |
+|---|---|---|
+| Box cost | always alloc | branch + conditional alloc |
+| Unbox cost | always load | branch + conditional load |
+| `is_always_scalar` | false | still false (≈50 % box) |
+| `Opt.inject` | always needed | still needed |
+| GC pressure | 3 words per value | ≈1.5 words avg |
+
+This optimisation is deferred; the present plan establishes the `SR.UnboxedFloat32`
+foundation and always-heap `Bits32 F` boxing on which it can be layered.
```
crusso: Oh great, so the some-times unboxed rep for float32 is not implemented just a sketched optimization. That's good. Let's not go there.

.claude/plans/float32-scalar.md
```diff
@@ -0,0 +1,266 @@
...
+  get_addr ^^ get_f ^^
+  G.i (Store {ty = F32Type; align = 0; offset = 4L; sz = None}) ^^  (* f32 at upper half *)
+  get_addr ^^
+  ( if TaggingScheme.tagging_enabled ()
+    then compile_unboxed_const (TaggingScheme.tag_of_typ Type.Float32)  (* tag as i32 *)
+    else compile_unboxed_zero ) ^^                                        (* 0 when tagging off *)
+  G.i (Store {ty = I32Type; align = 0; offset = 0L; sz = None})          (* tag at lower half *)
+
+(* Load: plain f32.load at offset 4 — no shift, no reinterpret *)
+let load_field_float32 env get_addr =
```
crusso: Interesting load optimization. I wonder if we also do this for int8/nat8,...int32/nat32. 
Leave for future work though,

.claude/plans/float32-scalar.md
```diff
@@ -0,0 +1,266 @@
...
+
+### 8. Float32 literals
+
+Line 10847: `| Float32Lit f -> Const.Float64 f`
+
+Since `Const` has no `Float32` variant, emit as `Const.Float64 f` (keeping the full
+f64 bit pattern) and add a convert in the constant-to-SR path: when the expected SR is
+`UnboxedFloat32`, apply `f64.demote_f32` after loading the constant.
+
+Alternatively, add `Const.Float32 of F32.t` to `const.ml` — but that's a larger change;
```
crusso: Probably adding Const.Float32 is  a better change though...

ggreif: The plan is a-priori, use as orientation only. The real thing is implemented!

.claude/plans/float32-scalar.md
ggreif: There is no u8 or u16 data type in Wasm, so probably also no loads of that size.

.claude/plans/float32-classical.md
ggreif: Yeah, just throwing around ideas. Should they become prevalent, it might become interesting!

.claude/plans/float32-backend.md
```diff
@@ -0,0 +1,133 @@
+# Float32 Backend Plan
+
```
crusso: Should I read this or it is out of date?

.claude/plans/float32-classical.md
ggreif: The RTTI exemption is stated in the plan, though.

.claude/plans/float32-backend.md
ggreif: No need, you can return if something is not clear from the code...

Also this is quite outdated. This was the initial one for the enhanced backend.

src/codegen/compile_classical.ml
```diff
@@ -9214,6 +9305,12 @@ module StackRep = struct
     | UnboxedFloat64, Vanilla -> Float.box env
     | Vanilla, UnboxedFloat64 -> Float.unbox env
 
+    | UnboxedFloat32, Vanilla -> Float32.box env
+    | Vanilla, UnboxedFloat32 -> Float32.unbox env
+
+    | UnboxedFloat64, UnboxedFloat32 -> G.i (Convert (Wasm.Values.F32 F32Op.DemoteF64))
+    | UnboxedFloat32, UnboxedFloat64 -> G.i (Convert (Wasm.Values.F64 F64Op.PromoteF32))
```
crusso, line 9313: Curious, I wonder if we actually need this - I don't think we have similar widening/narrowing for other stackreps. Perhaps it depends on how the prims are implemented

ggreif: This is needed to make other prims accessible. I don't want to add `fsinf` etc. to the runtime system at the moment.

> In Rust, you can use the `fsinf` function from the `libm` crate to compute the sine of a floating-point number, which is part of the transcendental functions available in Rust's math libraries.

src/codegen/compile_classical.ml
```diff
@@ -11187,7 +11307,9 @@ and compile_prim_invocation (env : E.t) ae p es at =
 
   | OptPrim, [e] ->
     SR.Vanilla,
-    Opt.inject env (compile_exp_vanilla env ae e)
+    if StackRep.is_always_scalar e.note.Note.typ
```
crusso: StackRep.is_always_scalar isn't really a property of StackReps. Maybe there's a more suitable module to put it in? Maybe BitTagged?

src/codegen/compile_classical.ml
```diff
@@ -11187,7 +11307,9 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
+    then compile_exp_vanilla env ae e
```
crusso: Nice optimization!

src/codegen/compile_classical.ml
ggreif: Yeah, good point. It used to check for stackreps, but I found a counterexample. Now it only checks the type.

src/codegen/compile_classical.ml
ggreif: These are the backend for `prim "num_conv_Float_Float32"` and `"num_conv_Float32_Float"`. They are needed.

src/codegen/compile_enhanced.ml
crusso: Can you add a case to the comment here for the new tag, and maybe try to check the commented code compiles without a redundant match failure. Warning: I think the 64-bit version caused Ocaml do die.

src/codegen/compile_enhanced.ml
```diff
@@ -6492,17 +6502,30 @@ module StackRep = struct
...
     | Func (Shared _, _, _, _, _) -> Vanilla
     | p -> todo "StackRep.of_type" (Arrange_ir.typ p) Vanilla
 
+  (* True for types whose Vanilla encoding is always a bit-tagged scalar (bit 0 = 0),
+     so Opt.inject is a no-op and can be omitted at compile time.
+     Nat64/Int64 are excluded: values outside the 60-bit compact range are heap-boxed
+     as Bits64 (bit 0 = 1), so the scalar property does not hold for all values.
+     Opt.inject is still a no-op for Nat64/Int64 (branch_default returns Bits64 as-is),
+     but it cannot be eliminated statically. *)
+  let is_always_scalar t =
```
crusso: Nice

src/codegen/compile_enhanced.ml
```diff
@@ -6601,6 +6625,24 @@ module StackRep = struct
     | UnboxedFloat64, Vanilla -> Float.box env
     | Vanilla, UnboxedFloat64 -> Float.unbox env
 
+    (* Float32: inline tagged I64 like Nat32 — f32 bits in upper 32, tag in lower 32 *)
```
crusso, line 6647: Got here

Changelog.md
```diff
@@ -3,6 +3,7 @@
 * motoko (`moc`)
 
   * feat: provide a polymorphic `actorOfPrincipal` primitive (#5882).
+  * feat: add `Float32` primitive type with conversions to/from `Float`, Candid serialisation, and literal ascription (e.g. `(3.14 : Float32)`). This is an experimental feature and subject to change (#5906).
```
```suggestion
  * feat: add `Float32` primitive type with conversions to/from `Float`, Candid serialisation, and literal ascription (e.g. `(3.14 : Float32)`). This is experimental and subject to change (#5906).
```

src/codegen/compile_classical.ml
ggreif: See 5da35e0e13ad67ff2a8901fea837bfd061983736.

src/codegen/compile_enhanced.ml
ggreif: See: a691342d4

I still get
> Fatal error: exception Stack overflow

src/codegen/compile_classical.ml
```diff
@@ -11418,6 +11540,17 @@ and compile_prim_invocation (env : E.t) ae p es at =
...
+
+    | Float, Float32 ->
+      SR.UnboxedFloat32,
+      compile_exp_as env ae SR.UnboxedFloat64 e ^^
+      G.i (Convert (Wasm.Values.F32 F32Op.DemoteF64))
+
+    | Float32, Float ->
+      SR.UnboxedFloat64,
+      compile_exp_as env ae SR.UnboxedFloat32 e ^^
+      G.i (Convert (Wasm.Values.F64 F64Op.PromoteF32))
```
crusso, line 11555: Aren't these the conversions?

src/codegen/compile_classical.ml
crusso: That's weird, because those are also explicitly implemented below. See comment

src/codegen/compile_classical.ml
crusso: (in which case you don't need the other cases above, I think).

src/codegen/compile_classical.ml
```diff
@@ -12888,7 +13029,8 @@ and compile_lit_pat env l =
   | BlobLit t ->
     compile_lit_as env SR.Vanilla l ^^
     Blob.compare env (Some Operator.EqOp)
-  | FloatLit _ ->
+  | FloatLit _
+  | Float32Lit _ ->
```
crusso, line 13036: Do we disallow floats in pats? I hope so, just not sure.

crusso: Oh wow, we have a bug in the existing compiler:
```
crusso@crusso-Virtual-Machine:~/motoko/test/run$ more float-pat.mo
import Prim = "mo:prim";
let (0.1,0.0) = if true  (0.1,0.0) else (0.0, 1.0);
Prim.debugPrint ("ok")

crusso@crusso-Virtual-Machine:~/motoko/test/run$ ../run.sh float-pat.mo 
float-pat: [tc] [run] [run-ir] [run-low] [comp] [valid] [wasm-run]
--- float-pat.tc (expected)
+++ float-pat.tc (actual)
@@ -0,0 +1,5 @@
+float-pat.mo:2.5-2.14: warning [M0145], this pattern of type
+  (Float, Float)
+does not cover value
+  (0.100_000_000_000_000_01, _) or
+  (_, _)
--- float-pat.run (expected)
+++ float-pat.run (actual)
@@ -0,0 +1 @@
+ok
--- float-pat.run-ir (expected)
+++ float-pat.run-ir (actual)
@@ -0,0 +1 @@
+ok
--- float-pat.run-low (expected)
+++ float-pat.run-low (actual)
@@ -0,0 +1 @@
+ok
--- float-pat.comp (expected)
+++ float-pat.comp (actual)
@@ -0,0 +1,2 @@
+compile_lit_pat: (FloatLit 0)
+compile_lit_pat: (FloatLit 0.100_000_000_000_000_01)
--- float-pat.wasm-run.ret (expected)
+++ float-pat.wasm-run.ret (actual)
@@ -0,0 +1 @@
+Return code 134
--- float-pat.wasm-run (expected)
+++ float-pat.wasm-run (actual)
@@ -0,0 +1,9 @@
+TODO: compile_lit_pat
+Error: failed to run main module `_out/float-pat.wasm`
+
+Caused by:
+    1: error while executing at wasm backtrace:
+         0: init
+         1: _start
+    2: wasm trap: unreachable
Some tests failed:
float-pat.mo
```

crusso: Filed https://github.com/caffeinelabs/motoko/issues/5917.
No need to fix in this PR.

src/codegen/compile_enhanced.ml
crusso: Oh well, thanks for trying. 
The code is meant to check that all the tags are distinct.

src/codegen/compile_enhanced.ml
```diff
@@ -6601,6 +6626,24 @@ module StackRep = struct
...
+      BitTagged.sanity_check_tag __LINE__ env Type.Float32 ^^
+      compile_shrU_const 32L ^^
+      G.i (Convert (Wasm_exts.Values.I32 I32Op.WrapI64)) ^^
+      G.i (Convert (Wasm_exts.Values.F32 F32Op.ReinterpretInt))
+
+    (* Cross-width float conversions on the Wasm stack *)
+    | UnboxedFloat32, UnboxedFloat64 ->
+      G.i (Convert (Wasm_exts.Values.F64 F64Op.PromoteF32))
+    | UnboxedFloat64, UnboxedFloat32 ->
+      G.i (Convert (Wasm_exts.Values.F32 F32Op.DemoteF64))
```
crusso: Again, not sure these are actually needed.

src/codegen/compile_enhanced.ml
```diff
@@ -6609,6 +6652,12 @@ module StackRep = struct
     | Const Const.Lit (Const.Word64 (ty1, n)), UnboxedWord64 ty2 when ty1 = ty2 ->
         compile_unboxed_const n
     | Const Const.Lit (Const.Float64 f), UnboxedFloat64 -> Float.compile_unboxed_const f
+    | Const Const.Lit (Const.Float64 f), UnboxedFloat32 ->
```
crusso: Maybe just add a Const.Float32 as you did in compile_classical?

src/codegen/compile_enhanced.ml
```diff
@@ -6609,6 +6652,12 @@ module StackRep = struct
...
+      Float.compile_unboxed_const f ^^
+      G.i (Convert (Wasm_exts.Values.F32 F32Op.DemoteF64))
+    | Const Const.Lit (Const.Vanilla n), UnboxedFloat32 ->
```
crusso: I wonder how this case arises? Any idea?

src/codegen/compile_enhanced.ml
```diff
@@ -10823,6 +10884,12 @@ let const_lit_of_lit : Ir.lit -> Const.lit = function
   | TextLit t     -> Const.Text t
   | BlobLit t     -> Const.Blob t
   | FloatLit f    -> Const.Float64 f
+  | Float32Lit f  ->
+    (* EOP: Float32 lives as a bit-tagged I64 scalar in Vanilla (f32_bits<<32 | tag). *)
+    let f32_bits = Int32.bits_of_float (Numerics.Float32.to_float f) in
+    Const.Vanilla Int64.(logor
+      (shift_left (logand (of_int32 f32_bits) 0xFFFFFFFFL) 32)
```
crusso: Isn't
      (logand (of_int32 f32_bits) 0xFFFFFFFFL)
just the same as
     of_int32 f32_bits?

src/codegen/compile_enhanced.ml
```diff
@@ -11312,6 +11386,14 @@ let compile_binop env t op : SR.t * SR.t * G.t =
       (powInt64_shortcut (Word64.compile_unsigned_pow env))
   | Type.(Prim Nat),                          PowOp -> BigNum.compile_unsigned_pow env
   | Type.(Prim Float),                        PowOp -> E.call_import env "rts" "pow"
+  | Type.(Prim Float32),                      PowOp ->
+    (* stack: [e1:f32, e2:f32]; promote both to f64, call pow, demote back *)
```
crusso, line 11391: again, maybe worth having a dedicated float32 pow implementation (not sure if you get the same results).

src/codegen/compile_enhanced.ml
```diff
@@ -11388,6 +11473,18 @@ let compile_relop env t op =
   | Type.(Prim Float), GeOp -> compile_comparison_f64 F64Op.Ge
   | Type.(Prim Float), LeOp -> compile_comparison_f64 F64Op.Le
   | Type.(Prim Float), LtOp -> compile_comparison_f64 F64Op.Lt
+  | Type.(Prim Float32), GtOp ->
```
crusso: maybe worth defining `compile_comparison_f32` and following the pattern above, for similarity and concision.

src/codegen/compile_enhanced.ml
```diff
@@ -11555,7 +11652,12 @@ and compile_prim_invocation (env : E.t) ae p es at =
 
   | OptPrim, [e] ->
     SR.Vanilla,
-    Opt.inject env (compile_exp_vanilla env ae e)
+    (* Never-heap-boxed types (UnboxedWord64 / UnboxedFloat32) have bit 0 = 0
+       in their Vanilla encoding, so Opt.inject is a no-op — emit directly.
+       TODO: extend to compile_classical.ml once it gains SR.UnboxedFloat32 support *)
```
crusso:
```suggestion
    (* Types with exclusively scalar representation have a no-op injection *)
```
This comment seems weird though the code looks ok.

src/codegen/compile_enhanced.ml
crusso: (might also apply to compile_classical - take a quick look)

src/mo_frontend/typing.ml
```diff
@@ -1280,29 +1283,31 @@ let infer_lit env lit at : T.prim =
 
 let check_lit env t lit at suggest =
   match t, !lit with
```
crusso, line 1283: Nit: Maybe you could just do the entire match under T.(...) and avoid all the inner ones.

(no need to fix)

src/mo_types/typ_hash.ml
```diff
@@ -86,6 +87,7 @@ let prim = function
   | Int32 -> "i32"
   | Int64 -> "i64"
   | Float -> "f"
+  | Float32 -> "h"
```
crusso, line 90: How about F (for 64) and f for 32? I don't think we use `f` and changing from f to F is benign I think.

src/mo_values/numerics.ml
```diff
@@ -200,6 +200,11 @@ struct
 end
 
 module Float = MakeFloat(Wasm.F64)
+module Float32 = struct
```
crusso:
```suggestion
module Float32 = MakeFloat(Wasm.F32) 
```
doesn't work?

crusso: (asking, I don't know)

src/mo_values/numerics.mli
```diff
@@ -69,6 +69,10 @@ sig
 end
 
 module Float : FloatType with type bits = int64 and type t = Wasm.F64.t
+module Float32 : sig
+  include FloatType with type bits = int64 and type t = Wasm.F64.t
+  val demote : t -> t
+end
```
crusso, line 74:
```suggestion
module Float32 : sig
  include FloatType with type bits = int32 and type t = Wasm.F32.t
end
```
Doesn't work?

src/prelude/prelude.mo
```diff
@@ -18,6 +18,7 @@ type Int16 = prim "Int16";
 type Int32 = prim "Int32";
 type Int64 = prim "Int64";
 type Float = prim "Float";
+type Float32 = prim "Float32";
```
crusso, line 21: I guess other names might be "Single" or "float", though "float" breaks with our uppercase convention.

test/bench/ok/palindrome.drun-run.ok
```diff
@@ -1,10 +1,10 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
-debug.print: (true, +2_928, 25_412)
-debug.print: (false, +2_928, 24_081)
-debug.print: (false, +2_928, 25_386)
-debug.print: (true, +2_184, 23_432)
-debug.print: (false, +2_160, 21_507)
-debug.print: (false, +2_160, 23_232)
+debug.print: (true, +2_928, 25_181)
```
crusso, line 3: Is that no-op injection on scalars at work?

src/prelude/prim.mo
crusso: Could be even better - there's two unboxings of $f. Can we avoid that by let-binding f on entry? 

Not that it matters...

crusso: I don't actually know if let bound vars are stored in stack rep or not. Probably not.

crusso (commented):
Looks very nice. See comments. Probably needs another wee round of review but great work!

crusso (commented):
I don't see any changes to stable format in the rts. Maybe it's not needed because the values are always unboxed and fit in 64 bit but seems fishy and worth double-checking.

Maybe add a graph stabilization tests using some float32s in stable vars and check nothing goes wrong (or just extend an existing test with some float32s).

Might be worth bikeshedding the name Float32 on the channel...

UPDATE: 
- I took a quick look at the motoko/rts/motoko-rts/src/stabilization and it might be that no extra treatment for Float32 is needed. Luc seems to just preserver the scalar tagging, using the same tagging scheme as compile_enhanced.ml
- still worth writing a test though ;->

.claude/plans/float32-scalar.md
crusso: Ok, but I think compile_enhanced doesn't gave Const.Float32 (though _classical does). Or I'm misremembering...

src/codegen/compile_classical.ml
crusso: Oh, are just saying all the trigonemtric 64 bit prims just work if you add this? Ok, but what happens if the result is out of bounds. Do you get a trap or loss of precision converting from F64 to F32?

crusso: trap would be worse I guess. Not sure.

src/codegen/compile_classical.ml
crusso: Unless you used them to re-use F64 prims on F32. Ok

src/codegen/compile_classical.ml
ggreif: Another comment from @crusso (in Slack)
> Just a thought. Maybe  move the Opt.inject optimization in the function so it's always applied, eg also in candid deserialization if missing

src/codegen/compile_classical.ml
ggreif: The idea is to `Float32` -> `Float` --sin-> `Float` -> `safeFloatToFloat32` and handle the problem. Later we can reference the f32 library functions.

src/codegen/compile_classical.ml
ggreif: Sure!

src/codegen/compile_classical.ml
```diff
@@ -2975,6 +3006,47 @@ module Float = struct
...
+module Float32 = struct
+  (* We store Float32 values in immutable boxed 32-bit heap objects.
+
+     The heap layout of a Float32 is:
+
+       ┌────────────┬─────┐
+       │ obj header │ f32 │
+       └────────────┴─────┘
+
+     Tag = Bits32 F (45).  Incr GC: 3 words; no GC: 2 words.
```
crusso:
```suggestion
     Tag = Bits32 F (45).  Incr GC: 3 words; other GC: 2 words.
```

src/codegen/compile_classical.ml
```diff
@@ -10915,6 +11024,14 @@ let compile_binop env t op : SR.t * SR.t * G.t =
       (powInt64_shortcut (Word64.compile_unsigned_pow env))
   | Type.(Prim Nat),                          PowOp -> BigNum.compile_unsigned_pow env
   | Type.(Prim Float),                        PowOp -> E.call_import env "rts" "pow"
```
crusso, line 11025: Does this produce the same answers as a dedicated pow? Maybe its worth extending the rts for this now.

src/codegen/compile_enhanced.ml
```diff
@@ -82,6 +82,7 @@ module TaggingScheme = struct
     | ((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (_,_,_,_,I,I,O,O)) -> TInt64
```
crusso: Hmm, I wonder if Ocaml copes better if we replace each `(_,_,_,_,_,_,_,_)`  by ` _`.
I'll try that myself on master...

src/codegen/compile_enhanced.ml
```diff
@@ -2214,7 +2230,9 @@ module Opt = struct
   let alloc_some env get_payload =
     Tagged.obj env Tagged.Some [ get_payload ]
 
-  let inject env e =
+  let inject env t e =
+    if BitTagged.is_always_scalar t then e
```
crusso, line 2246: Hmm, we can probably optimize this even more. Even if t is not always scalar, as long as t is not null, an option type or generic, then injection is always a noop (I think). Best as separate PR, not here.

crusso: Opened tracking issue https://github.com/caffeinelabs/motoko/issues/5938

src/codegen/compile_enhanced.ml
```diff
@@ -6601,6 +6628,24 @@ module StackRep = struct
...
 
+    (* Float32: inline tagged I64 like Nat32 — f32 bits in upper 32, tag in lower 32 *)
+    | UnboxedFloat32, Vanilla ->
+      G.i (Convert (Wasm_exts.Values.I32 I32Op.ReinterpretFloat)) ^^
+      G.i (Convert (Wasm_exts.Values.I64 I64Op.ExtendUI32)) ^^
+      compile_shl_const 32L ^^
+      compile_bitor_const (TaggingScheme.tag_of_typ Type.Float32)
+    | Vanilla, UnboxedFloat32 ->
+      BitTagged.sanity_check_tag __LINE__ env Type.Float32 ^^
+      compile_shrU_const 32L ^^
```
crusso:
```suggestion
      BitTagged.un_tag env __LINE__ Type.Float32 ^^
```
might work, provided you teach BitTagged.tag about the new Float32.

src/codegen/compile_enhanced.ml
crusso: Is the TODO: done?

crusso (commented):
Took another pass, but you got some unresolved comments.

I think its very close to going in. Maybe teach BitTagged about the float tag in compile_enhanced.

src/codegen/compile_enhanced.ml
ggreif: Done in commit 08d355b65. The big-endian layout caused OCaml's exhaustiveness checker to stack-overflow because it works left-to-right through patterns — with MSB-first ordering the discriminating tag bits ended up at the *rightmost* position, hitting an exponential blowup. Switching to little-endian (LSB first within each byte, low bytes first) puts the tag bits on the left, making the check instant. See https://github.com/ocaml/ocaml/issues/14664#issuecomment-4135559828

The `TFloat32` case is added and the function is now uncommented (live), so the injectivity check is enforced at compile time.

Line-by-line correctness report (LE layout: `b0=(bit0..7)` … `b7=(bit56..63)`, bits LSB-first within each byte):

| Tag | Key bit(s) | Pattern byte | Correct? |
|-----|-----------|--------------|---------|
| TBool false `0` | all zero | all `(O…O)` | ✓ |
| TBool true `1` | bit0 → b0[0] | b0=`(I,O,…)` | ✓ |
| TRef `0b11` | bits 1:0=11 → b0[1:0] | b0=`(I,I,_…)` | ✓ |
| TNum `0b10` | bits 1:0=10 → b0[1:0] | b0=`(O,I,_…)` | ✓ |
| TNat64 `0b0100` | bit2 → b0[2] | b0=`(O,O,I,O,_…)` | ✓ |
| TInt64 `0b1100` | bits 3:2 → b0[3:2] | b0=`(O,O,I,I,_…)` | ✓ |
| TNat32 `0x40000000` | bit30 → b3[6] | b3=`(O,O,O,O,O,O,I,O)`, b4-b7=`_` | ✓ |
| TInt32 `0xC0000000` | bits 31:30 → b3[7:6] | b3=`(O,O,O,O,O,O,I,I)` | ✓ |
| TFloat32 `0x20000000` | bit29 → b3[5] | b3=`(O,O,O,O,O,I,O,O)` | ✓ |
| TChar `0b010_0…0` | bit41 → b5[1], bit40=bit42=0 | b5=`(O,I,O,_…)`, b6,b7=`_` | ✓ |
| TNat16 `0x400000000000` | bit46 → b5[6] | b5=`(O,O,O,O,O,O,I,O)` | ✓ |
| TInt16 `0xC00000000000` | bits 47:46 → b5[7:6] | b5=`(O,O,O,O,O,O,I,I)` | ✓ |
| TNat8 `0x40000000000000` | bit54 → b6[6] | b6=`(O,O,O,O,O,O,I,O)`, b7=`_` | ✓ |
| TInt8 `0xC0000000000000` | bits 55:54 → b6[7:6] | b6=`(O,O,O,O,O,O,I,I)` | ✓ |
| TUnit `0x4000000000000000` | bit62 → b7[6] | b7=`(O,O,O,O,O,O,I,O)` | ✓ |

src/mo_values/numerics.mli
ggreif: I'll try. I can revert if something goes awry.

src/codegen/compile_enhanced.ml
ggreif: Yes, done — `compile_classical.ml` gained `SR.UnboxedFloat32` in this PR (line 297), so `Opt.inject` already handles Float32 as a no-op in both backends. TODO removed in a05c3bba1.

src/codegen/compile_enhanced.ml
ggreif: Done in 9cd8d3cae — extended `BitTagged.untag` to handle `Float32` (unsigned logical shift, like `Nat32`), then replaced the open-coded `shl`/`bitor` and `sanity_check_tag`/`shrU` sequences with `BitTagged.tag` and `BitTagged.untag`.

src/codegen/compile_classical.ml
ggreif: I would prefer not to add `powf` for now. I guess (!) doing a series of single-precision multiplies will deliver poorer results than when doing it in double precision. Think underflow... Feel free to file a reminder issue if you think this is a problem.

src/codegen/compile_enhanced.ml
ggreif: You're right — removed in 8986f6bb4. Float↔Float32 conversion is always handled explicitly in `NumConvTrapPrim` codegen, which emits `DemoteF64`/`PromoteF32` inline. Since Float and Float32 have no subtyping relationship, `StackRep.adjust` can never be called with these cross-width float pairs.

src/prelude/prelude.mo
ggreif: Candid uses `float32`: https://github.com/dfinity/sdk/pull/4494/changes#r2916194615. I prefer to not introduce new names.

test/bench/ok/palindrome.drun-run.ok
ggreif: Yep. Let's see how your ideas improve on this :-)

src/prelude/prim.mo
ggreif: I'll leave it to the cycle counters to bitch about this 😛

src/mo_types/typ_hash.ml
ggreif: No idea. Are these hashes stored somewhere? Or are they per `moc` invocation. In the latter case I am inclined to change to `F`/`f`.

src/codegen/compile_enhanced.ml
ggreif: Claude did: baa4c0642

src/codegen/compile_enhanced.ml
ggreif: See https://github.com/caffeinelabs/motoko/pull/5906#discussion_r2999934769

src/mo_types/typ_hash.ml
crusso: I'm pretty sure these are just used internally, eg, to generate distinct function names for serialization and equality comparison and  such like. Using f and F might make them more easy to decode when debugging.

src/codegen/compile_enhanced.ml
ggreif: Done in both backends in commit 10636a421. Also extracted compile_comparison_f64 in compile_classical.ml for symmetry.

ggreif: Done — added compile_comparison_f32 in compile_enhanced.ml next to compile_comparison_f64, and symmetrically in compile_classical.ml too. Committed in 10636a421.

src/mo_types/typ_hash.ml
ggreif: The hash is used only to generate unique Wasm function names at compile time (e.g. `@buffer_size<h>`, `@deserialize_go<h>`) and as an ephemeral type-sanity scratch during a single deserialization pass — it is **not** persisted in stable memory between upgrades. So changing it is always safe.

After checking: `"F"` is taken (by functions), so we keep `"h"` for `Float32` — it hints at "half-width" and is prefix-free with `"f"`.

ggreif: This is what Claude says ^^^. (Now it changed its mind after following the code paths – check the edit history :-)

crusso: I don't think that's correct. Ask it to show you where

crusso: Claude is incorrect, but the hashes are reported in error messages so just stick with f and h as you planned (F and s can't be used because of Functions and (shared) Functions)

src/codegen/compile_classical.ml
crusso: It's no so much about poorer results as what users might expect... Perhaps they expect it all to be done in f32 land.
But yeah we change this later I guess or document.

src/codegen/compile_enhanced.ml
```diff
@@ -63,33 +63,36 @@ module TaggingScheme = struct
...
-    | ((_,_,_,_,_,_,_,_), (_,_,_,_,_,_,_,_), (I,I,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O)) -> TInt16
-    | ((_,_,_,_,_,_,_,_), (O,I,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O)) -> TNat8
-    | ((_,_,_,_,_,_,_,_), (I,I,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O)) -> TInt8
-    | ((O,I,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O)) -> TUnit
-    | _                                                                                                                                                        -> TUnused
-  *)
+    | ((I,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O), (O,O,O,O,O,O,O,O)) -> TBool (* true *)
```
crusso, line 81: Nice!

src/codegen/compile_enhanced.ml
```diff
@@ -6553,6 +6592,10 @@ module StackRep = struct
   | Const.Lit (Const.BigInt number) -> BigNum.constant env number
   | Const.Lit (Const.Word64 (pty, number)) -> BoxedWord64.constant env pty number
   | Const.Lit (Const.Float64 number) -> Float.constant env number
+  | Const.Lit (Const.Float32 f) ->
+    E.Vanilla Int64.(logor
+      (shift_left (logand (of_int32 (Wasm.F32.to_bits f)) 0xFFFFFFFFL) 32)
```
crusso: Isn't the logand redundant?

src/codegen/compile_enhanced.ml
```diff
@@ -11555,7 +11646,9 @@ and compile_prim_invocation (env : E.t) ae p es at =
 
   | OptPrim, [e] ->
     SR.Vanilla,
-    Opt.inject env (compile_exp_vanilla env ae e)
+    (* Never-heap-boxed types (UnboxedWord64 / UnboxedFloat32) have bit 0 = 0
+       in their Vanilla encoding, so Opt.inject is a no-op — emit directly. *)
```
crusso: I guess you can delete the comment or move to Opt.inject

src/codegen/compile_enhanced.ml
ggreif: Absolutely! I even let Claude correct such an instance in the past :-)

ggreif: Indeed redundant — `shift_left _ 32` shifts the sign-extended upper 32 bits completely out. Removed.

src/codegen/compile_enhanced.ml
ggreif: Deleted.

src/prelude/prelude.mo
crusso: I can live with Float32

test/bench/ok/palindrome.drun-run.ok
crusso: A fair amount it seems. 

debug.print: (true, +2_928, 24_290)
debug.print: (false, +2_928, 22_959)
debug.print: (false, +2_928, 24_264)
debug.print: (true, +2_184, 22_310)
debug.print: (false, +2_160, 20_385)
debug.print: (false, +2_160, 22_110)

crusso: Maybe it's broken ;->

crusso (commented):
LGTM apart from three comments

crusso (approved):
Fire away (but maybe improve the PR description and add a Changelog entry)

PR #5920, opened 2026-03-19
removes rename hints for implicit errors

christoph-dfinity:
This also lets us avoid expensive subtype checks across the entire environment

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -1421,7 +1421,6 @@ type hole_candidate =
     typ : T.typ;
     module_ref_opt: T.lab option; (* module name (from `vals`) or path (from `libs`) *)
     id : T.lab;
-    region : Source.region;
   }
 
 let desc_of_candidate candidate = quote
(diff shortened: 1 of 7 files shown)


src/mo_frontend/typing.ml
```diff
@@ -1506,81 +1504,42 @@ let resolve_hole env at name typ =
...
-          in
-            info env candidate.region
-             "Consider renaming `%s` to `%s.%s` in %s module `%s`. Then it can serve as an implicit argument `%s` in this call:\n%s%s"
-             (desc_of_candidate candidate) mid name mod_desc mid name call_region call_src)
-      explicit_terms
+      Seq.filter_map (find_candidate_field in_libs) |>
+      List.of_seq
   in
+  let eligible_terms = match T.Env.find_opt name env.vals with
+    | Some (t, region, _, _) when is_matching_typ t ->
```
Kamirus, line 1521: I think this is wrong: when there are two ambiguous local candidates this now just picks the first one instead of reporting ambiguity

christoph-dfinity: Only one value of a given name can exist in the local scope, right?

Kamirus: yes my mistake! btw AI was brutal when asked about this:
> You're wrong here, and the behavior is preserved.

😆

christoph-dfinity: You must've spoiled it... As a punishment you should make it do math without allowing python scripts :P

PR #5923, opened 2026-03-19
feat(test-runner): add batch mode, accept flag, and shared test infrastructure

Kamirus:
## Summary

- Add `-b`/`--batch` flag that skips the interactive picker and runs all matched tests directly
- Add `-a`/`--accept` flag that passes `-a` to `run.sh` to update expected output (`ok/`) files
- Extract shared infrastructure: `discover_tests()`, `compile_filter()`, `run_tests()` — used by both interactive and batch modes
- Auto-detect the correct `run.sh` mode flags based on directory (`fail/` → `-t`, `run-drun/` → `-d`, `run/` → no flag)
- Refactor `run_single_test` and `run_interactive_mode` to take `&TestRunnerArgs` instead of individual parameters

## Usage

Batch mode is the interactive mode minus the picker. The same `-f` filter selects tests in both modes.

```bash
# Interactive: explore tests with the picker
test-runner -f lambdas

# Batch: run all matching tests directly
test-runner -b -f lambdas

# Batch + accept changed outputs
test-runner -b -f implicit -a

# Batch + type-check only
test-runner -b -f lambdas --just-tc

# Batch: run ALL tests (warns about count)
test-runner -b

# Regex filter
test-runner -b -f "lambda.*core"

# Run all fail tests
test-runner -b -f /fail

# Search in test output contents
test-runner -b -f M0223 --in-file
```

Flags `-f`, `-a`, `--just-tc`, `--in-file` work with both modes. `--review` and `--dir` are interactive-only (conflict with `-b`).

Invalid regex patterns in batch mode produce a clear error instead of silently falling back to substring matching.

## Test plan

- [x] `test-runner -b -f lambdas` runs 7 tests, all pass
- [x] `test-runner -b -f "lambda.*core"` matches 2 tests via regex
- [x] `test-runner -b -f lambdas-core -a` runs with accept flag
- [x] `test-runner -b -f lambdas --just-tc` runs type-check only
- [x] `test-runner -b -f implicit` runs 17 tests across fail/run/run-drun
- [x] `test-runner -b --review` rejected by clap (conflict)
- [x] `test-runner -b --run` rejected by clap (conflict)
- [x] `test-runner -b -f "foo["` shows clear regex error, exits 1
- [x] `test-runner -b -f zzz_nonexistent` shows "no matches", exits 1
- [x] `test-runner --help` shows all flags correctly
- [x] `test-runner` (no args) still launches interactive mode

### DIFF
--- a/.github/actions/test-blueprint/action.yml
+++ b/.github/actions/test-blueprint/action.yml
@@ -7,12 +7,15 @@
 #   - run: nix profile install .#nix-build-uncached
 #   - name: "Run common tests"
 #     run: |
-#       cachix watch-exec ic-hs-test -- \
+#       cachix watch-exec ic-hs-test -- \  # only when auth token is available
 #         nix-build-uncached --max-jobs auto \
 #           -E '(builtins.getFlake (toString ./.)).packages.${builtins.currentSystem}.common-tests'
 
 # The main workflow from test.yml will call this workflow with different
 # inputs, and this workflow will call the tests defined in the flake.nix file.
+#
+# Fork PRs run without Cachix push (no auth token available), but still
+# benefit from read-only cache access via the `cachix use` substituter.
 
 name: Test Blueprint
 description: Runs Motoko tests using nix and cachix
@@ -28,8 +31,9 @@ inputs:
     required: true
     description: "Name of the test to display in logs"
   cachix-auth-token:
-    required: true
-    description: "Cachix authentication token"
+    required: false
+    default: ''
+    description: "Cachix authentication token (optional; fork PRs run without push)"
 
 runs:
   using: "composite"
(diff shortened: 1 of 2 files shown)


test-runner/src/main.rs
```diff
@@ -328,12 +368,17 @@ fn main() {
...
-        run_interactive_mode(
-            args.filter.as_deref().unwrap_or(""),
-            args.in_file,
-            args.just_tc,
-            args.review,
-        );
+        if !args.paths.is_empty() {
+            run_batch_mode(args.paths, args.just_tc, args.accept, args.review);
+        } else {
+            run_interactive_mode(
```
alexandru-uta: maybe just pass args and have the function pick and chose the fields it needs? also above.
when we started this there were few args, but now it's getting just longer and longer

Kamirus: Done — all functions now take `&TestRunnerArgs` and pick the fields they need. Also removed a needless `.clone()` on `Output` in `run_single_test`.

_This comment was written by Claude._

alexandru-uta: Thanks, Claude!

test-runner/src/main.rs
```diff
@@ -69,50 +85,63 @@ fn run_legacy_mode(subnet_type: SubnetType) {
     test_runner::run_cmdline_test(buffer, subnet_type);
 }
 
-/// The program offers the user a list of tests to choose from.
-/// A summary of the results of the tests is then printed out.
-fn run_interactive_mode(input_str: &str, search_in_file: bool, just_tc: bool, do_review: bool) {
-    let test_dirs = ["test/run-drun", "test/run", "test/fail"];
+const TEST_DIRS: [&str; 3] = ["test/run-drun", "test/run", "test/fail"];
```
alexandru-uta: since we're making a larger change -- maybe also add `test/trap`, should work I think

PR #5927, opened 2026-03-21
experiment: implement masked `br_table` dispatch for variant switches (n ≥ 4 arms)

ggreif:
## Summary

- Replaces the O(n) linear comparison chain for variant `switch` with O(1) `br_table` dispatch when there are 4 or more arms
- At compile time, finds a bitmask M (and shift S = ctz(M)) such that `(hash_i & M) >> S` are all distinct, then emits `i32.and M; [i32.shr_u S]; br_table`
- Mask-finding uses [Gosper's hack](https://programmingforinsomniacs.blogspot.com/2018/03/gospers-hack-explained.html) to iterate candidates in order of popcount and value, ensuring compact (low-index) masks are tried first
- Threshold: `max(64, 4n)` table size; falls back to linear chain if exceeded
- Strict win for all n ≥ 4 cases (n = 1 and 2 already handled by `single_case`/`simplify_cases`)

## Benchmark (vs moc 1.3.0, `test/bench/variant-switch.mo`)

Baseline: `PATH=/nix/store/71w3w2df8xv4x56dkff6sl5yfwd01ccc-moc/bin:$PATH`

| Workload | moc 1.3 | this branch | speedup |
|---|---|---|---|
| 9-arm switch loop ×10 000 (`go`) | 137,590,321 | 95,690,321 | **1.44×** |
| AST `eval` fib(7) ×100 | 24,492,248 | 22,034,248 | **1.11×** |
| AST→FT `transform` ×100 | 1,189,148 | 1,057,948 | **1.12×** |
| FT eval fib(7) ×100 | 21,519,148 | 21,519,148 | 1.00× (noise floor — no variant dispatch) |

The FT row is unchanged by design: the finally-tagless form has no `Expr` variant dispatch in its hot loop, confirming the other speedups are real and attributable to the switch optimisation.

## Test plan

- [ ] `make -C test/run variant_switch.only` — passes (4-arm Color, 7-arm Weekday, 4-arm Shape with payloads)
- [ ] `make -C test/run variants.only` — existing variant tests pass (no regressions)
- [ ] Inspect WAT output: `br_table` with `i32.and` emitted; `i32.shr_u` emitted when S > 0 (confirmed for Weekday: mask `0x15000`, shift 12)

## Key files

- `src/codegen/compile_classical.ml` — helpers `bits_needed`, `iter_masks_with_popcount` (Gosper's hack), `is_injective`, `compact_table_size`, `find_variant_mask`; new `SwitchE` case
- `test/run/variant_switch.mo` — new test
- `.claude/plans/variant-switch-br-table.md` — design plan (includes future work: same-body arm merging, or-pattern handling)

## TODOs
- [ ] effective branches (provably same result)
- [ ] distill not from IR, but from the `(^^^)` EDSL (maybe change to finally-tagless?)


ggreif:
## Reference: weekday variant hashes

Recovered from the compiled Wasm (`i64.const` at each `#Day` construction site) and cross-checked against the br_table slots in `isWeekend`/`isWeekendOr`. Useful as a concrete worked example when developing alternative `Dispatch` strategies.

| tag | hash (dec) | hash (hex) | `& 0x15000` | br_table slot |
|---|---:|---:|---:|---:|
| `#Mon` | 3_853_996 | `0x3ACEAC` | `0x04000` | 4 |
| `#Tue` | 4_203_428 | `0x4023A4` | `0x00000` | 0 |
| `#Wed` | 4_349_046 | `0x425C76` | `0x05000` | 5 |
| `#Thu` | 4_200_545 | `0x401861` | `0x01000` | 1 |
| `#Fri` | 3_506_557 | `0x35817D` | `0x10000` | 16 |
| `#Sat` | 4_149_254 | `0x3F5006` | `0x15000` | 21 |
| `#Sun` | 4_153_708 | `0x3F616C` | `0x14000` | 20 |

Gosper's mask picks the three bits at positions 12, 14, 16 (`mask = 0x15000`, `shift = 12`). Seven weekday tags land on seven distinct slot values `{0, 1, 4, 5, 16, 20, 21}`; slot `17` is the one valid-but-unused slot visible in the br_table's default-label entries — any future weekday whose hash happened to set bits 0 and 4 (post-shift) would land there without growing the table.

Pattern reading: bit 16 of the hash == weekend-side (set for Fri/Sat/Sun, clear for Mon/Tue/Wed/Thu); bits 12/14 distinguish days within each half.

Strategy notes for the dispatch handler:
- `MaskShift` (current default, Gosper): 23-slot table, 2 arithmetic ops (AND + SHR_U) + `br_table`.
- `ModPrime` could try small primes `p ≥ 7`; e.g. does `hᵢ mod 11` give 7 distinct residues? (Quick check: `3853996 % 11 = 2`, `4203428 % 11 = 9`, `4349046 % 11 = 10`, `4200545 % 11 = 6`, `3506557 % 11 = 3`, `4149254 % 11 = 7`, `4153708 % 11 = 6` — collision Thu↔Sun on `mod 11`; try 13, 17, …). Cheaper table (size `p`) but `rem_u` is ~3× an `and`.
- `RotLow bits=3 rot=?` — would need table size 8; worth trying for comparison once the handler protocol supports it.

## `startLetter`: 7-way distinct dispatch (flat vs or-pattern)

A richer 7-way switch added to `test/bench/variant-switch.mo`. Two forms produce the same `mask=0x15000, shift=12` dispatcher (same 7 weekday hashes), but partition the arm blocks differently:

### `startLetter` (7 flat arms) — 8 arm blocks

| slot | case idx | tag | outcome |
|---:|---:|---|---:|
| 0 | 1 | `#Tue` | `'T'` |
| 1 | 3 | `#Thu` | `'T'` |
| 4 | 0 | `#Mon` | `'M'` |
| 5 | 2 | `#Wed` | `'W'` |
| 16 | 4 | `#Fri` | `'F'` |
| 20 | 6 | `#Sun` | `'S'` |
| 21 | 5 | `#Sat` | `'S'` |

The Tue and Thu arm blocks are *byte-identical* (both load `i64.const 741070837121024` = `'T'`); likewise Sat and Sun both load `732274744098816` = `'S'`. Duplicated code, but each block is entered via exactly one br_table slot, so each call costs the same regardless of duplication.

### `startLetterOr` (5 arms, 2 or-patterns) — 5 arm blocks

| slot | case idx | tag | outcome |
|---:|---:|---|---:|
| 0 | 1 | `#Tue` | `'T'` (case shared with `#Thu`) |
| 1 | 1 | `#Thu` | `'T'` (same case) |
| 4 | 0 | `#Mon` | `'M'` |
| 5 | 2 | `#Wed` | `'W'` |
| 16 | 3 | `#Fri` | `'F'` |
| 20 | 4 | `#Sun` | `'S'` (case shared with `#Sat`) |
| 21 | 4 | `#Sat` | `'S'` (same case) |

Same br_table, fewer arm blocks (two pairs collapsed).

### Bench numbers (10 000 × 7-day iterations)

| metric | instructions |
|---|---:|
| `instr_isWeekend` | 7_070_321 (incl. outer `if` branch cost) |
| `instr_isWeekendOr` | 7_070_321 |
| `instr_startLetter` | 6_270_321 (clean dispatch + Char sink) |
| `instr_startLetterOr` | 6_270_321 |

**Key observation for strategy design:** `startLetter` and `startLetterOr` cost *exactly the same* at runtime despite the or-pattern form having 2 fewer arm blocks. Same-body arm duplication is a pure code-size bloat — each slot's br_table entry jumps directly into its own block, which executes the same ≈3 Wasm instructions regardless of whether a byte-identical block lives next to it.

**Where same-body merging *would* matter**: upstream of the dispatcher. If the recognizer were to merge equivalence classes *before* handing the token set to the handler, N = 5 instead of N = 7. Then `ModPrime` could try `mod 5` instead of `mod 7` (half the table bytes), `MaskShift` would have fewer injectivity constraints (possibly a smaller mask), and perfect-hash search gets cheaper. That's the lever the plan now captures under *Future Optimisation: Same-body arm merging* — with the refinement that the equivalence criterion should be raw Wasm byte sequences (each arm is already a `Block` internally and `(^^^)` is difference-list concatenation of `G.t`), not IR structural equality.

For V2 as shipped here, or-pattern merging is the *only* channel; cross-case merging stays user-driven (write the or-pattern to get the size win).

PR #5929, opened 2026-03-22
feat: support fixed-width Nat types as array and blob indices

q-uint:
## Summary

- Allow Nat8, Nat16, Nat32, and Nat64 values as array indices (and Nat64 for blobs), bypassing bignum conversion for direct word-sized index operations
- All fixed-width Nat types compile to direct indexing, reducing IC instruction counts by 30-59% compared to Nat indices
- Changes span the type checker, both codegen backends, IR checker, both interpreters, and include run/trap/fail/bench tests

## IC instruction counts (1000 iters x 256 elements)

| Type           | Instructions | vs Nat   |
|----------------|-------------|----------|
| Nat (baseline) | 25,687,306  |          |
| Nat8           | 11,036,370  | -57%     |
| Nat16          | 11,079,329  | -57%     |
| Nat32          | 10,567,324  | -59%     |
| Nat64          | 17,991,325  | -30%     |
| Nat64+toNat()  | 20,807,325  | -19%     |


src/lang_utils/error_codes.ml
```diff
@@ -207,6 +207,7 @@ let error_codes : (string * string option) list =
     "M0233", None; (* Wrong number of arguments *)
     "M0234", None; (* Field exists but is not a function *)
     "M0238", None; (* Misplaced break or continue *)
+    "M0245", None; (* Invalid array index type (expected Nat or Nat64) *)
```
q-uint: Comment OOD

src/codegen/compile_enhanced.ml
```diff
@@ -11430,10 +11430,87 @@ let rec compile_lexp (env : E.t) ae lexp : G.t * SR.t * G.t =
...
-    Arr.idx_bigint env
+    (match unwrap_toNat ae e2 with
+     | Some (pty, inner) ->
+       (* Peephole: recognize arr[NatN.toNat(x)] and elide the
+          conversion, compiling x directly as its native type. *)
+       compile_exp_as env ae (SR.UnboxedWord64 pty) inner ^^
+       TaggedSmallWord.lsb_adjust pty ^^
+       Arr.idx env
+     | None ->
+       (* Speculative fast path for plain Nat (bigint) array indices.
```
crusso: Are theseBigInt optimization still worth it?

test/bench/nat64-index.mo
```diff
@@ -0,0 +1,127 @@
+// Benchmark: NatN.toNat() peephole vs plain Nat array indexing
```
crusso, line 1:
```
//MOC-FLAG --package core $MOTOKO_CORE  
```
would let you reference core libs to test n.toNat() and Nat64.toNat(n) etc

src/codegen/compile_classical.ml
```diff
@@ -11062,10 +11088,86 @@ let rec compile_lexp (env : E.t) ae lexp : G.t * SR.t * G.t =
...
+            (32 - ubits) then recovers the raw index, which we pass
+            directly to Arr.idx (the same function used by the
+            fixed-width Nat8/Nat16/Nat32 paths).
+
+          Slow path:
+            Any value that fails the check (heap bigint, or a
+            negative Int that somehow got here) falls through to
+            Arr.idx_bigint, which handles the full conversion.
+
+          Both array and index are saved to locals because wasm
```
crusso: I wonder if that's actually true or just a limitation of our E.if_ helper that doesn't let you supply a operand  stack type  for each branch.

Wasm 1.0 didn't allow the branches to assume the stack is non-empty. But Wasm 2.0 does.
Unfortunately, I'm not entirely sure we are allowed to use Wasm 2.0 features.

@ggreif are we good with Wasm 2.0 instructions?

If so, we don't need new locals and can avoid the stores and loads of get_va and get_vi.

src/codegen/compile_enhanced.ml
q-uint: No, it introduces a lot of complexity.

crusso: If the numbers are better might still be worth it. I can have a play

src/codegen/compile_enhanced.ml
```diff
@@ -11430,14 +11430,29 @@ let rec compile_lexp (env : E.t) ae lexp : G.t * SR.t * G.t =
...
-        (Ir.NumConvTrapPrim (Type.(Nat8|Nat16|Nat32|Nat64 as pty), Type.Nat))))) ->
-      Some (pty, inner)
+    | Some (VarEnv.Const c) ->
+      Option.map (fun pty -> (pty, inner)) (unwrap_toNat_const c)
+    | _ -> None
+    end
+  | Ir.PrimE (Ir.CallPrim _,
+      [{it = Ir.PrimE (Ir.DotPrim name, [{it = Ir.VarE (_, mod_var); _}]); _}; inner]) ->
+    begin match VarEnv.lookup_var ae mod_var with
+    | Some (VarEnv.Const (Const.Obj fs)) ->
```
crusso, line 11451: I guess one could even iterate this for deeper projection paths, but probably not worth it.

q-uint:
Seems like it will not get merged.

PR #5935, opened 2026-03-25
fix: only allow system calls in actor bodies

alexandru-uta:


src/mo_frontend/typing.ml
```diff
@@ -3858,6 +3858,7 @@ and infer_obj env obj_sort exp_opt dec_fields at : T.typ =
       ) dec_fields;
     end;
     if s = T.Module then Static.dec_fields env.msgs dec_fields;
+    if s = T.Actor && Option.is_some env.enhanced_migration then Static.actor_dec_fields env.msgs dec_fields;
```
crusso: Will this catch the non-static expressions in mixins too?

alexandru-uta: nope, oversight, I think it's fixed now.

src/mo_frontend/static.ml
```diff
@@ -11,56 +11,75 @@ allow re-exporting names in modules.
...
+     use-case: multi-migration actor bodies where we want to allow e.g., timers.
+  *)
+  | CallE (_, callee, inst, (_, ref_args))
+    -> (match (env.allow_sys_call, inst.it) with
+        | (true, Some(true, _)) ->
+          (exp env callee;
+           exp env !ref_args)
+        | _ -> err env e.at)
 
   (* Clearly non-static *)
```
crusso, line 84: Future: we could probably relax many of these to be considered static, for example, I think we could probably allow some pure arithmetic etc, just not updates and calls hiding updates.

src/mo_frontend/static.ml
```diff
@@ -11,56 +11,75 @@ allow re-exporting names in modules.
...
+  let open Diag in
+  add_msg env.msg_store
+   (error_message
+      at
+      "M0014"
+      "type"
+      (if not (env.allow_sys_call) then
+         "non-static expression in library, module or migration expression"
+       else
+         "non-static expression in actor body compiled with enhanced migration capabilities. Only <system> capabilities allowed for non-static expressions."))
```
crusso:
```suggestion
         "non-static expression in actor body compiled with enhanced migration capabilities."))
```
Maybe get rid of the second half, since it'll mention <system> capabilities even for things that fail to be static for other reasons (like mutable arrays)

PR #5940, opened 2026-03-27
feat: preserve named types for VarP bindings in check_pat_aux

Kamirus:
## Summary
- Pass the original (unnormalized) type as `t_orig` through `check_pat_aux'` so that `VarP` bindings retain the user-written type alias instead of the expanded structural form
- This improves error messages by showing named types where the user declared them

### Before
```
expression of type
  {var root : Types.Map.Node<Text, Text>; var size_ : Nat}
cannot produce expected type ...
```

### After
```
expression of type
  Map.Map<Text, Text>
cannot produce expected type ...
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -180,7 +180,7 @@ jobs:
 
       - name: Notify node-motoko of new release
         if: ${{ github.event_name == 'push' }}
-        uses: peter-evans/repository-dispatch@v3
+        uses: peter-evans/repository-dispatch@v4
         with:
           token: ${{ steps.app-token.outputs.token }}
           repository: caffeinelabs/node-motoko


test/fail/ok/inf-error-small.tc-human.ok
```diff
@@ -23,7 +23,7 @@ error[M0233]: wrong number of arguments: expected 1 but got 3
 Expected 1 argument of type:
   (n : Nat)
 But got 3 arguments of type:
-  {map : [(Nat, [var Text])]}
+  Map.Map<Nat, Text> = {map : [(Nat, [var Text])]}
```
christoph-dfinity, line 26: This looks a bit weird, what's going on here?

Kamirus: A type alias expanded by the pprinter

christoph-dfinity: Did we show these anywhere before these changes?

Kamirus: yes, I guess this prettyprinter just expands all type aliases recursively
this PR is just oneline change in the typing 😄

christoph-dfinity: Checked that we report these in other places, by changing the pretty printer and re-running our test suite.

Good to go from my side.

Issue #5938, opened 2026-03-26
optimization: optimize most option injections to no-ops 


crusso:
Opt.inject t env can be optimized further to compile more option injections as noops, avoiding alot of code.

Gabor's float32 PR #5906 introduce the niceoptimization that compiles `?e` : `?t` to a noop  on `e` when `t` has an always scalar representation.
But even if `t` is not always scalar, as long as `t` is not `null`, an option type or generic (assume the worst), then injection is a noop too, and we can use the type to optimize this statically.

PR #5947, opened 2026-03-27
perf: statically optimize no-op option injections and projections FTW

crusso:
(Cherrypicked reconstruction of #5939 after base was squash merged)

Builds on the Opt.inject optimization in https://github.com/caffeinelabs/motoko/pull/5906, but extends it to more payload types.

Implements #5938.

* less code
* more perf

same correctness?

Saves 10%(!) of instructions on bench/alloc (creating a large linked list, so not surprising)

Background on our option representation: https://www.joachim-breitner.de/blog/787-A_mostly_allocation-free_optional_type

- [x] Consider promoting, not just normalizing, to exploit bounds that are better than Any.
- [x] Tests... Many iterators should get faster too.
- [ ] Should we load_forwarding_pointer in fast path or not. If so, need to special case  (or slow-path) bools coz of true literal looks like a pointer but can't be forwarded. For some reason, the existing Opt.simple_inject (a no-op) always forwards (but is only every used on reference types).
- [x] Optimize Opt.project too.

### DIFF
--- a/src/codegen/compile_enhanced.ml
+++ b/src/codegen/compile_enhanced.ml
@@ -2243,10 +2243,7 @@ module Opt = struct
   *)
   let injection_is_free env t =
     Type.(match promote t with
-         | Prim Null
-         | Opt _
-         | Any
-         | Con(_, _) -> false
+         | Prim Null | Opt _ | Any | Con _ -> false
          | _ -> true)
 
   let inject env t e =


test/bench/ok/alloc.drun-run.ok
```diff
@@ -1,9 +1,9 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
-debug.print: (+671_088_640, 5_522_892_825)
+debug.print: (+671_088_640, 5_069_907_993)
```
ggreif, line 3: These are fantastic!

src/codegen/compile_enhanced.ml
```diff
@@ -2242,8 +2231,16 @@ module Opt = struct
   let alloc_some env get_payload =
     Tagged.obj env Tagged.Some [ get_payload ]
 
+  let injection_is_free env t =
+    Type.(match promote t with
```
crusso, line 2245: Double check promote normalizes too.

UPDATE: Yes it does.

test/bench/iter.mo
```diff
@@ -0,0 +1,33 @@
...
+import Nat "mo:core/Nat"
+
+actor _alloc {
+
+  func counters() : (Int, Nat64) = (rts_heap_size(), performanceCounter(0));
+
+  public func go() : async () {
+    let (m0, n0) = counters();
+    for (i in Nat.range(0, 1024)) {
+      assert (i < 1024)
```
ggreif:
```suggestion
      assert i < 1024
```

test/bench/ok/alloc.drun-run.ok
crusso: It gets even better if we optimize projections too 

https://github.com/caffeinelabs/motoko/pull/5947/commits/2160e92be3175ade89f6e6264e70e7383afd29c1

src/codegen/compile_enhanced.ml
```diff
@@ -2287,7 +2287,10 @@ module Opt = struct
     Tagged.load_forwarding_pointer env ^^
     Tagged.load_field env some_payload_field
 
-  let project env =
+  let project env typ =
```
ggreif, line 2289: This will boost equally!

src/codegen/compile_enhanced.ml
```diff
@@ -2281,7 +2287,10 @@ module Opt = struct
...
+    if injection_is_free env typ
+    then G.i Nop
```
```suggestion
    then G.nop
```

src/codegen/compile_classical.ml
```diff
@@ -2407,7 +2416,10 @@ module Opt = struct
     Tagged.load_forwarding_pointer env ^^
     Tagged.load_field env (some_payload_field env)
 
-  let project env =
+  let project env typ =
+    if injection_is_free env typ
+    then G.i Nop
```
```suggestion
    then G.nop
```

test/run-drun/ok/options.tc.ok
```diff
@@ -0,0 +1,4 @@
+options.mo:483.5-483.9: warning [M0239], Avoid binding a unit `()` result; remove `let` and keep the expression
+options.mo:764.16-764.21: warning [M0146], this pattern is never matched
+options.mo:776.16-776.22: warning [M0146], this pattern is never matched
+options.mo:789.16-789.23: warning [M0146], this pattern is never matched
```
christoph-dfinity: Please fix these warnings to avoid churn if we change their messages in the future.

crusso: Done, I think.

test/run-drun/options.mo
```diff
@@ -0,0 +1,798 @@
+import Prim "mo:prim";
+
+// Test (type optimized) option injection and projection works correctly
+
+// Basically the same test repeated at different,
+// representative type instantiations (staticially unboxed, statically mixed,
+// statically unknown)
+// Intended to test compile_enhanced/classical/Opt.injection_is_free type based
```
crusso:
```suggestion
// Intended to test compile_enhanced/classical/Opt.injection_is_free type-based
```

src/codegen/compile_classical.ml
```diff
@@ -2373,32 +2365,49 @@ module Opt = struct
...
+     So for any T other than Null, ?U, Any, or generic bound T,
+     injection is a no-op and just returns v at type ?T.
+     For type that may contain null or ?w, we need to do some work.
+  *)
+  let injection_is_free env t =
+    Type.(match promote t with
+    | Prim Null
+    | Opt _
+    | Any
+    | Con(_, _) -> false
```
ggreif: These could save some vertical space
```suggestion
    | Prim Null | Opt _ | Any | Con _ -> false
```

src/codegen/compile_enhanced.ml
```diff
@@ -2244,26 +2233,43 @@ module Opt = struct
...
+     So for any T other than Null, ?U, Any, or generic bound T,
+     injection is a no-op and just returns v at type ?T.
+     For a type T that may contain null or ?w, we need to do some work.
+  *)
+  let injection_is_free env t =
+    Type.(match promote t with
+         | Prim Null
+         | Opt _
+         | Any
+         | Con(_, _) -> false
```
```suggestion
         | Prim Null | Opt _ | Any | Con _ -> false
```

PR #5951, opened 2026-03-27
perf: eliminate adjacent `PromoteF32;DemoteF64` pairs via peephole optimiser

ggreif:
## Summary

- Add a peephole rule to `instrList.ml` that removes consecutive `f64.promote_f32; f32.demote_f64` instructions — a Float32→Float→Float32 round-trip that is provably identity (mirrors the existing `ExtendI32;WrapI64` rule).
- Add a `floatToFloat32 (float32ToFloat pi) == pi` round-trip test to `test/run/float32.mo` that exercises the peephole.
- Fix `test/run.sh`: `grep -F -q ^//CHECK` was silently broken (GNU grep does not anchor `^` in fixed-string mode), so `[FileCheck]` never fired for any test. Fixed to `grep -q '^//CHECK'`.

## Details

The optimisation fires when `floatToFloat32 (float32ToFloat x)` is compiled via direct nested application. The pipe operator `|>` (we don't use it any more here) desugars to intermediate let-bindings which insert `local.tee` between the two conversions, breaking adjacency — so the test uses nested application specifically to expose the pattern.

Verified with a `failwith` probe before committing the real `go l' r'` elimination.

## Test plan

- [x] `make -C test/run float32.only` passes (all of `[tc]`, `[run]`, `[run-ir]`, `[run-low]`, `[comp]`, `[FileCheck]`, `[wasm-run]`)
- [x] `[FileCheck]` appears in the output, verifying the `pi` assertion compiles to two bare `f32.const` + `f32.eq` with no conversions
- [x] Confirm `f64.promote_f32; f32.demote_f64` no longer appears adjacent in `_out/float32.wasm`

### DIFF
--- a/test/run/ok/float32.diff-ir.ok
+++ b/test/run/ok/float32.diff-ir.ok
@@ -1,6 +1,6 @@
 --- float32.run
 +++ float32.run-ir
-@@ -1 +1,76 @@
+@@ -1 +1,81 @@
 +        (LitE (TextLit Float32 precision tests passed: ))
 +  (LetD (VarP lit32) (LitE (Float32Lit 3.140_000_104_904_174_8)))
 +  (LetD (VarP a2) (LitE (Float32Lit 2)))
--- a/test/run/ok/float32.diff-low.ok
+++ b/test/run/ok/float32.diff-low.ok
@@ -1,6 +1,6 @@
 --- float32.run
 +++ float32.run-low
-@@ -1 +1,106 @@
+@@ -1 +1,113 @@
 +        (LitE (TextLit Float32 precision tests passed: ))
 +  (LetD (VarP lit32) (LitE (Float32Lit 3.140_000_104_904_174_8)))
 +  (LetD (VarP a2) (LitE (Float32Lit 2)))


test/run/ok/float32.comp.ok
```diff
@@ -28,6 +29,7 @@
     (PrimE (BinPrim Float32 PowOp) (VarE a2) (LitE (Float32Lit 0.5)))
       (LitE (Float32Lit 1.414_213_538_169_860_8))
         (LitE (Float32Lit 1.000_000_013_351_432e-10))
+  (LetD (VarP pi) (LitE (Float32Lit 3.141_592_741_012_573_2)))
```
alexandru-uta, line 32: I don't think PI is correct :D 
I vaguely remember `...9265` since I was a kid and google says this:
`pi = 3.141_592_653_589_793_238 `

test/run.sh
```diff
@@ -449,7 +449,7 @@ do
           # Check filecheck
           if [ "$SKIP_RUNNING" != yes ]
           then
-            if grep -F -q ^//CHECK $mangled
+            if grep -q '^//CHECK' $mangled
```
alexandru-uta, line 452: nice catch!

test/run/ok/float32.comp.ok
ggreif: Yeah, the code has the ClaudePi. Gives starry circles!

test/run/float32.mo
```diff
@@ -82,6 +82,16 @@ assert a2 <= a2;
 assert b2 >  a2;
 assert b2 >= b2;
 
+// PromoteF32;DemoteF64 round-trip is identity (peephole target)
+// Direct nested application avoids the pipe's intermediate let-binding (which would insert local.tee)
+let pi : Float32 = 3.14159265358979323846;
```
crusso, line 87: @ggreif why is the pi in src, different from the pi in the debug output? Is there a bug in the conversion from literal text to float representation?

crusso: 3.14159265358979323846 vs
3.141_592_741_012_573_2

ggreif: Because of precision. `Float32` has about 7-8 digits of precision everything behind that becomes 0 in the `Float`'s mantissa. Then when you display that you get something different. We don't have `format` and `toText` for native `Float32` yet. That's why stuff is experimental. There are a few rough edges still.

crusso: Ok, that makes sense I guess.

Just as a general feature, would it make sense to warn about float literals that don't roundtrip to text format? I don't know what other languages do.

ggreif: Rather, I'd avoid going to textual, as it is notoriously difficult to get right. Better convert to `F64` and `F32` in Wasm format, then promote the `F32` to `F64` and if that is different to the former we have loss of precision. That is a good reason to warn. But I wonder if that'd give lots of false positives. See #5973.

PR #5962, opened 2026-03-31
fix: run BimatchRet during pre-pass so return infers generic types

Kamirus:
## Problem

When an unannotated lambda with `return` statements is passed to a generic function (e.g. `"abc".chars().map(func(c) { if (c == ' ') return #space else return #char(c) })`), the type checker resolves the generic type parameter to `Non` (bottom type) instead of the actual return type. This produces an IR type error after desugaring:

```
Ill-typed intermediate code after Desugaring:
  {next : () -> ?{#char : Char; #space}}
  {next : () -> ?None}
```

### Root cause

The `RetE` handler in `typing.ml` is guarded by `if not env.pre`, so during the preliminary type-checking pass (Pass 4), `return` expressions skip the `BimatchRet` callback entirely. The callback is what feeds the return value's type into `infer_body` for bi-directional type matching. Without it, `return` contributes `Non` as the body type, and the generic parameter resolves to `Non`.

Note: running `BimatchRet` during the pre-pass does not guarantee we always visit every `return` — the pre-pass still skips most sub-expressions. But the issue was specifically that when a `return` _was_ reached, its type constraint was silently discarded.

## Fix

Add an `else` branch to the `RetE` handler that runs `BimatchRet` callbacks even when `env.pre = true`. Other `rets` cases (`Ret`, `NoRet`) are intentionally skipped during the pre-pass.

## Test plan

- [x] New regression test `test/run/return-in-deferred-lambda.mo` — exercises `return` in lambdas passed to `Iter.map` and `Iter.filterMap`
- [x] Existing test suite passes without regressions

### DIFF
--- a/doc/docusaurus/package-lock.json
+++ b/doc/docusaurus/package-lock.json
@@ -12531,9 +12531,9 @@
       "integrity": "sha512-UtJcAD4yEaGtjPezWuO9wC4nwUnVH/8/Im3yEHQP4b67cXlD/Qr9hdITCU1xDbSEXg2XKNaP8jsReV7vQd00/A=="
     },
     "node_modules/minimatch": {
-      "version": "3.1.2",
-      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-3.1.2.tgz",
-      "integrity": "sha512-J7p63hRiAjw1NDEww1W7i37+ByIrOWO5XQQAzZ3VOcL0PNybwpfmV/N05zFAzwQ9USyEcX6t3UO+K5aqBQOIHw==",
+      "version": "3.1.5",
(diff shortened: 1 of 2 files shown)


src/mo_frontend/typing.ml
```diff
@@ -2206,6 +2206,11 @@ and infer_exp'' env exp : T.typ =
         k env exp1
       | NoRet ->
         local_error env exp.at "M0085" "misplaced return"
+    end else begin
+      (* Pre-pass must still run BimatchRet to infer return types for generic lambdas *)
+      match env.rets with
+      | BimatchRet k -> k env exp1
```
crusso, line 2212: OMG, is our type-checker in cps? ;->

Kamirus: now that I think of this, I'd rather name this `infer` instead of `k` 😅

PR #5966, opened 2026-04-01
feat: implicit argument derivation

Kamirus:
## Summary

Part 1 of #5903 — implicit argument derivation only, without structural derivation (`__record`/`__tuple` combiners).

The compiler can now derive implicit arguments from functions that themselves have implicit parameters. For example, an implicit `compare : ([Nat], [Nat]) -> Order` is automatically derived from `Array.compare<Nat>` when `Nat.compare` is in scope.

- Derivation works transitively (e.g. `compare` for `[[Nat]]` chains through `Array.compare<[Nat]>` → `Array.compare<Nat>` → `Nat.compare`)
- Recursive derivation cycles are detected and handled via `rec_bindings`
- Depth is bounded by `--implicit-derivation-depth` (default 100)
- Refactored `resolve_hole` and `disambiguate_resolutions` to support the new derivation tiers while preserving existing direct-match behavior
- Improved error messages: derivation failures report which inner implicits are missing and suggest imports

Structural derivation (`__record`/`__tuple` combiners for records and tuples) will follow in a separate PR on top of this one.


src/mo_config/args.ml
```diff
@@ -73,7 +73,8 @@ let inclusion_args = [
 let ai_args = [
   "--ai-errors", Arg.Set Flags.ai_errors, " emit AI tailored errors";
   "--all-libs", Arg.Set Flags.all_libs, " load all library files from all packages, enabling better diagnostics, e.g. hinting at non-imported items (increases compilation time)";
-  "--implicit-package", Arg.String (fun s -> Flags.implicit_package := Some s), _UNDOCUMENTED_ " allow contextual dot and implicits resolution from all modules in the given package"
+  "--implicit-package", Arg.String (fun s -> Flags.implicit_package := Some s), _UNDOCUMENTED_ " allow contextual dot and implicits resolution from all modules in the given package";
+  "--implicit-derivation-depth", Arg.Set_int Flags.implicit_derivation_depth, _UNDOCUMENTED_ " maximum depth for recursive implicit derivation (default: 100)"
```
crusso, line 77: Do we want this to be _UNDOCUMENTED_ ?

If not we should also add a description to compiler-ref.md (the doc that describes all compiler options)

src/mo_config/flags.ml
```diff
@@ -87,6 +87,7 @@ let experimental_stable_memory_default = 0 (* _ < 0: error; _ = 0: warn, _ > 0:
 let experimental_stable_memory = ref experimental_stable_memory_default
 let typechecker_combine_srcs = ref false (* useful for the language server *)
 let blob_import_placeholders = ref false (* when enabled, blob:file imports resolve as empty blobs *)
+let implicit_derivation_depth = ref 100
```
crusso, line 90: I guess this will affect all imported code too. Maybe that's ok but seems fragile.

doc/md/fundamentals/11-implicit-parameters.md
```diff
@@ -3,25 +3,25 @@
 ## Overview
 
 Implicit parameters allow you to omit frequently-used function arguments at call sites when the compiler can infer them from context. This feature is particularly useful when working with ordered collections like `Map` and `Set` from the `core` library, which require comparison functions but where the comparison logic is usually obvious from the key type.
```
crusso, line 5: Ugh, I must have forgotten to use ispell - sorry!

src/mo_frontend/typing.ml
```diff
@@ -1470,79 +1473,386 @@ let is_val_module (n, ((t, _, _, _) : val_info)) =
...
+      | [arg] -> arg | args -> mk (TupE args) in
+    func_ ~name (List.rev param_names_rev) (call candidate_path call_arg_exp)
+
+end
+
+(** Checks [args -> rets <: req_args -> req_rets] via subtyping or
+    bidirectional matching when [tbs] are present. Returns [Some inst] or [None]. *)
+let sub_or_bimatch_func tbs args rets req_args req_rets =
+  if tbs = [] then
+    if List.for_all2 (fun a b -> T.sub a b) req_args args
```
crusso, line 1599:
```suggestion
    if List.for_all2 T.sub req_args args
```

src/mo_frontend/typing.ml
```diff
@@ -1470,79 +1473,386 @@ let is_val_module (n, ((t, _, _, _) : val_info)) =
...
+    && List.for_all2 (fun a b -> T.sub a b) rets req_rets
```
crusso, line 1600:
```suggestion
    && List.for_all2 T.sub rets req_rets
```

src/mo_frontend/typing.ml
```diff
@@ -1470,79 +1473,386 @@ let is_val_module (n, ((t, _, _, _) : val_info)) =
...
+    match deriv_err with
+    | DepthLimited ->
+      Printf.sprintf "depth limit reached (increase with `--implicit-derivation-depth`, current: %d)"
+        !(Flags.implicit_derivation_depth) :: acc
+    | InnerErrors inner_errors -> List.fold_left collect_error acc inner_errors
+  and collect_error acc (name, typ, err) = match err with
+    | HoleSuggestions (lib_terms, note_opt) ->
+      let imports = List.filter_map import_suggestion_of_candidate lib_terms in
+      if imports <> [] then
+        Format.asprintf env "`%s : %a` — not found, try importing %s"
```
crusso, line 1537:
```
     Format.asprintf env "`%s : %a` — not found, try importing %s"
```

I smell AI... ;->

src/mo_frontend/typing.ml
```diff
@@ -1470,79 +1473,386 @@ let is_val_module (n, ((t, _, _, _) : val_info)) =
...
+module SynthesizeWrapper = struct
+  (* Fresh AST nodes are required at each use site — type-checking annotates in
+     place, so sharing a node triggers an "already-annotated" assertion. *)
+  let mk e = e @? no_region
+  let var n = mk (VarE (n @~ no_region))
+  let inst () = Source.annotate [] None no_region
+  let var_pat n = VarP (n @@ no_region) @! no_region
+  let call path arg =
+    mk (CallE (None, path, inst (), (false, ref arg)))
+  let func_ ~name param_names body =
```
crusso, line 1563: FYI I'm pretty sure ~name is actually not used by the type-checker (e.g. for recursion) but just an annotation used by the interpreter to produce traces. In case you are relying on it...

src/mo_frontend/typing.ml
```diff
@@ -1470,79 +1473,386 @@ let is_val_module (n, ((t, _, _, _) : val_info)) =
...
+  | T.Named (_inf_arg_name, (T.Named ("implicit", T.Named (arg_name, t)))) ->
+    (* override inferred arg_name *)
+    Some arg_name
+  | T.Named (inf_arg_name, (T.Named ("implicit", t))) ->
+    (* non-overriden, use inferred arg_name *)
+    Some inf_arg_name
+  | _ -> None
+
+(* Like [as_implicit] but also returns the inner type that needs to be resolved *)
+let as_implicit_with_type = function
```
crusso, line 1499: Maybe rename this to as_implict and adjust the as_implicit call-sites (ie. refactor)

src/mo_frontend/typing.ml
```diff
@@ -1470,79 +1473,386 @@ let is_val_module (n, ((t, _, _, _) : val_info)) =
...
+    cand_args : T.typ list; (* Not substituted! Use only to determine the argument name *)
+    holes : hole list;
+    func_without_holes : T.typ;
+  }
+
+  (* Only Local/Returns functions are eligible for derivation:
+     Shared and Composite functions (actors, async) are excluded
+     since implicits are a local-scope, synchronous mechanism. *)
+  let is_matching_typ_with_holes hole candidate_typ =
+    match T.promote hole.hole_typ, T.promote candidate_typ with
```
crusso, line 1643:
```
    match T.promote hole.hole_typ, T.promote candidate_typ with
```
I'm not sure you should promote the type of the hole, only the candidate. If this was a variable, it would replace it by its bounds (in what I think is a negative position). Suppose we had a hole of type T <: U  and a candidate of type U. Then U <: T only if we substitute U for T. But we could actually substitute any W < U (strictly less than) for T too.

src/mo_frontend/typing.ml
```diff
@@ -1470,79 +1473,386 @@ let is_val_module (n, ((t, _, _, _) : val_info)) =
...
+      |> Option.map (fun holes -> holes, make_field_candidate module_ref field))
+
+  end
+
+  let make_val_candidate id t =
+    let path = VarE (id @~ no_region) @? no_region in
+    { path; typ = t; module_ref_opt = None; id }
+
+  let matching_val hole (vals : val_env) =
+    let* (t, _, _, _) = T.Env.find_opt hole.hole_name vals in
```
crusso, line 1714: We probably want to rule out mutable ids as candidates. The is matching_typ predicate might do it already, but perhaps it's worth adding an explicit check in matching_val and matching_val_with_holes

src/mo_frontend/typing.ml
```diff
@@ -1470,79 +1473,386 @@ let is_val_module (n, ((t, _, _, _) : val_info)) =
...
+    if not (is_matching_typ hole t) then None else
+    Some (make_val_candidate hole.hole_name t)
+
+  let matching_val_with_holes hole (vals : val_env) =
+    let* (t, _, _, _) = T.Env.find_opt hole.hole_name vals in
```
crusso, line 1719: We probably want to rule out mutable ids as candidates (see above).

src/mo_frontend/typing.ml
```diff
@@ -2599,6 +2886,41 @@ and check_exp' env0 t exp : T.typ =
...
+    exp_ref := path;
+    check_exp env hole_typ path
+  | Ok ({path; _}, bindings) ->
+    let env_rec = List.fold_left (fun env (name, typ, _) ->
+      { env with vals = T.Env.add name
+          (typ, Source.no_region, Scope.Declaration, Available) env.vals }
+    ) env bindings in
+    List.iter (fun (_, typ, func_exp) -> check_exp env_rec typ func_exp) bindings;
+    check_exp env_rec hole_typ path;
+    exp_ref := mk_recursive_block at bindings path hole_typ
```
crusso, line 2901: Hmm, so each witness is reconstructed at the hole site. I wonder how expensive this will get. For example, in some Map client code, each call into map will reconstruct the ordering function if complicated. I guess the internal calls of Map will at least re-use the passed in compare argument.

Maybe for a later PR we could introduce a little dsl for the witnesses so we can deduplicate them in an ir pass. Might be hard though, since they can close over unknown functions in general, so you can't just gather them all and declare up-front.

src/mo_frontend/typing.ml
```diff
@@ -1470,79 +1473,386 @@ let is_val_module (n, ((t, _, _, _) : val_info)) =
...
+    assert (remaining = []);
+    let call_arg_exp = match List.rev call_args_rev with
+      | [arg] -> arg | args -> mk (TupE args) in
+    func_ ~name (List.rev param_names_rev) (call candidate_path call_arg_exp)
+
+end
+
+(** Checks [args -> rets <: req_args -> req_rets] via subtyping or
+    bidirectional matching when [tbs] are present. Returns [Some inst] or [None]. *)
+let sub_or_bimatch_func tbs args rets req_args req_rets =
```
crusso, line 1597: maybe assert the args/req_args (rets/req_rets) lengths are equal (the call site guarantees this I think)

crusso (commented):
Some interim comments before GH loses them...

src/mo_config/flags.ml
Kamirus: yes, but we don't have flags per file/package and we need some default cap

src/mo_frontend/typing.ml
Kamirus: good catch!

src/mo_frontend/typing.ml
Kamirus: I want to add caching/hoisting of these derived candidates so that they are statically allocated at the enclosing `module` (My idea is to emit these wrappers as normal module functions)

src/mo_frontend/typing.ml
Kamirus: because of `?src_fields` that can't be done

src/mo_config/args.ml
Kamirus: good catch, I'll change that, this should be documented

src/mo_frontend/typing.ml
crusso: Ok

src/mo_frontend/typing.ml
crusso: I'll see if I can construct a counter-example...

... Hmm, not having much luck.

src/mo_frontend/typing.ml
```diff
@@ -1470,79 +1473,386 @@ let is_val_module (n, ((t, _, _, _) : val_info)) =
...
   else
     ImplicitLibE module_ref
 
 let dot_module_exp module_exp name =
-  DotE({
-    it = module_exp;
-    at = name.at;
-    note = empty_typ_note
-  }, name, ref None) @? name.at
+  DotE(module_exp @? name.at, name, ref None) @? name.at
```
crusso, line 1500:
```suggestion
  DotE(module_exp @? no_region, name, ref None) @? no_region
```
Not sure why we used name.at?

src/mo_frontend/typing.ml
```diff
@@ -1470,79 +1473,386 @@ let is_val_module (n, ((t, _, _, _) : val_info)) =
...
+    holes : hole list;
+    func_without_holes : T.typ;
+  }
+
+  (* Only Local/Returns functions are eligible for derivation:
+     Shared and Composite functions (actors, async) are excluded
+     since implicits are a local-scope, synchronous mechanism. *)
+  let is_matching_typ_with_holes hole candidate_typ =
+    match T.promote hole.hole_typ, T.promote candidate_typ with
+    | T.Func (T.Local, T.Returns, [], req_args, req_rets),
```
crusso, line 1657: Is there any reason the hole type can't be a generic function - this seems to require the hole to be non-generic and I guess the witness to be monomorphic? I guess we could extend later if need.

crusso: I guess it's just because we don't support inference for unannotated generic lambdas, right?

doc/md/fundamentals/11-implicit-parameters.md
```diff
@@ -167,13 +167,57 @@ The compiler infers an implicit argument by:
...
+   1. Local values in the current scope.
+   2. Module fields (e.g., `Array.compare<T>`).
+   3. Fields of unimported modules (requires `--implicit-package`).
+Within each tier, if multiple candidates match, the compiler picks the most specific one (by subtyping). If no unique best candidate exists, the call is rejected as ambiguous.
+
+This ordering guarantees that direct matches are always preferred over derived ones, and local definitions take precedence over imported or unimported module definitions.
+
+### Implicit derivation
+
+When no direct match exists, the compiler can **derive** an implicit argument from a function that itself has implicit parameters. This eliminates the need for boilerplate wrapper functions. The candidate function can be polymorphic (the compiler infers the type instantiation) or monomorphic.
```
crusso, line 195: I don't think anything here clearly states that the hole must have a (monomorphic) function type although perhaps its implied by saying that stripping the implicit arguments from the candidate produces a matching type.

I now realize that's probably why my example in https://github.com/caffeinelabs/motoko/pull/6054 fails.

src/mo_frontend/typing.ml
crusso: Indeed, this requires the hole to have a function type, though we don't require this elsewhere. That's what tripped me up in https://github.com/caffeinelabs/motoko/pull/6054.

I guess this approach of inserting  implicit arguments is needed to make the existing (flattened) `Array.compare`  work without changes to core.

Another option might be to look for curried applications that produce the desired hole type directly, and then extend, e.g., core Array with:

```
   // existing Array.compare
   public func compare<T>(a1 : [T], a2 : [T], compare : (T, T) -> Order) : Order  { ... };

   // derivation
   oublic func deriveCompare<A>(compareA : (implicit : (compare : T -> T -> Order))) :
     (compare : ([T], [T]) -> Order)) =  func( a1, a2) { compare( a1, a2, compareA) };
```

Then our witnesses are not lambdas that insert arguments but partial applications of functions that only take implicit arguments and produce a matching witness (without constraining the dictionary type to be a function),

Then the search for derivations only considers  functions with (exlusively) implicit arguments that produces
a matching type.

But maybe our bimatch is not up to that or this winds up requiring backtracking.

 It also means we would need to add some bindings to core to derive `compare`, `eq` and `toText` for generic containers, instead of it just working with our existing core.

doc/md/fundamentals/11-implicit-parameters.md
Kamirus: I've explored this extension in https://github.com/caffeinelabs/motoko/pull/5991 (see comments there)
but this approach of deriving values (records/modules with funcs) instead of funcs is weaker, as it does not extend to `__record` to allow generic folds over records/tuples/etc
We could have both but this complicates the PR even more
So I'd just refactor your example to have two separate implicits, instead of a single record that ships both
This approach is simpler and extends to `__record`. 
I'd just define `toCandid` and `fromCandid` separately instead of `rep`.

btw, if you want to allow context-dot syntax like `myObj.toCandid()` then you'd need an indirection like
```motoko
func toCandid<A>(self : A, toCandidImpl : (implicit : A -> Blob)) : Blob {
  toCandidImpl(self)
}
```

src/mo_frontend/typing.ml
Kamirus: > Is there any reason the hole type can't be a generic function

No, just simplicity, we can extend if we ever need that

Kamirus: > I guess this approach of inserting implicit arguments is needed to make the existing (flattened) Array.compare work without changes to core.

Not only that, see https://github.com/caffeinelabs/motoko/pull/5966#discussion_r3147562857

test/fail/ok/implicit-derivation-ambiguous-deep.tc-human.ok
```diff
@@ -0,0 +1,7 @@
+error[M0230]: Cannot determine implicit argument `compare` of type
+  ([Nat], [Nat]) -> Order
+    ┌─ implicit-derivation-ambiguous-deep.mo:31:8
+ 31 │  ignore needsCompare([1], [2]); // Array.compare is unique head, but inner Nat compare is ambiguous
+    │         ^^^^^^^^^^^^^^^^^^^^^^ 
+    = note: Implicit derivation failed:
+                 `compare : (Nat, Nat) -> Order` ambiguous: `MyNat.compare`, `Nat.compare`
```
crusso:
```suggestion
                 `compare : (Nat, Nat) -> Order` has ambiguous solutions: `MyNat.compare`, `Nat.compare`
```
Might be clearer.

src/mo_frontend/typing.ml
```diff
@@ -1470,79 +1473,382 @@ let is_val_module (n, ((t, _, _, _) : val_info)) =
...
+      if imports <> [] then
+        Format.asprintf env "`%s : %a` not found, try importing %s"
+          name display_typ_oneline typ (List.hd imports) :: acc
+      else (match note_opt with
+      | Some note -> collect_note note acc
+      | None ->
+        Format.asprintf env "`%s : %a` not found"
+          name display_typ_oneline typ :: acc)
+    | HoleAmbiguous ambiguous_candidates ->
+      Format.asprintf env "`%s : %a` ambiguous: %s"
```
```suggestion
      Format.asprintf env "`%s : %a` has ambiguous solutions: %s"
```

Kamirus: I'll apply this later once you finish the review

crusso (approved):
I should stop worrying and learn to love the bomb.

PR #5970, opened 2026-04-01
feat: adds prims for accessing caller info

christoph-dfinity:
Fixes LANG-1246


alexandru-uta (approved):
That looks good to me, thanks!

In case we ever want to change how this works, we can do so by introducing new prims and deprecating the old.

One thing for the future: probably need to do something to support multiple principles in that env var for trusted signers. But that perhaps can also go in the core library.

christoph-dfinity:
> One thing for the future: probably need to do something to support multiple principles in that env var for trusted signers. But that perhaps can also go in the core library.

Yes, that was my plan. Comma-separated principals are handled in: https://github.com/caffeinelabs/motoko-core/pull/491

PR #5987, opened 2026-04-08
experiment: add do #lab variant blocks and / # narrowing operator

crusso:
If you can't beat 'em.... experimenting with cursor.

Generalize do ? { ... } option blocks to do #lab { ... } for variants. Inside a do #lab block, ! on a variant extracts the #lab payload and short-circuits (propagates) all other tags. Also add e / #lab for narrowing a variant type by removing #lab (trapping if matched).

Made-with: Cursor

Both exciting and depressing at the same time. But mostly depressing.

[ ] Manually check test output
[ ] Consider add SlashVariant prim to IR.prim, instead of relying on safe use of unsafe cast prim
[ ] Consider removing  `e / # lab` syntax. It's useful to avoid reallocation, but also trapping. I suggested it to the AI as a means of implementing `!` efficiently, without necessarily adding it to the surface language (but it preferred casts anyway).

Related (old) PR adding option blocks (`do ? ...`) https://github.com/caffeinelabs/motoko/pull/2174


src/lowering/desugar.ml
```diff
@@ -79,6 +79,8 @@ let desugar_loop_flags at note body flags with_body =
   let () = flags.has_break <- false in
   `Rec (S.LabelE (S.auto_s @@ at, unit_typ at, { it = with_body body; at; note = { note_typ = note.Note.typ; note_eff = note.Note.eff } }))
 
+let current_variant_do : string option ref = ref None
```
crusso: Yuck, added a global

crusso (commented):
This seems incorrect as is.
You can't use the same ! at two unrelated "#ok" values.

Nope, it's fine

doc/md/examples/grammar.txt
```diff
@@ -219,6 +219,7 @@
 <exp_bin> ::= 
     <exp_un>
     <exp_bin> <binop> <exp_bin>
+    <exp_bin> DIV'#' <id>
```
crusso, line 222: Fix sed script

test/fail/ok/parse-block-or-record.tc-human.ok
```diff
@@ -10,6 +10,7 @@ error[M0001]: unexpected token '=', expected one of token or <phrase> sequence:
   <unassign> <exp(ob)>
   <relop> <exp_bin(ob)>
   . <id>
+  <binop> <id>
```
crusso, line 13: This is odd - I wonder why this is now an item? Can't (myself) tell how it follows from the grammar.

doc/md/fundamentals/3-types/10-options.md
```diff
@@ -211,6 +211,42 @@ To guard against division by 0 without trapping, the `eval` function returns an
...
+// result == #err "division by zero"
+```
+
+Inside `do #ok { ... }`, the `!` operator checks if a variant value carries the `#ok` tag. If it does, `!` extracts the payload and continues. If it carries any other tag (e.g. `#err`), `!` immediately exits the block with that variant value.
+
+The `!` operator's behavior depends on context: inside `do ? { ... }` it unwraps options; inside `do #lab { ... }` it unwraps variants. Variant and option blocks can be nested, with `!` always referring to the nearest enclosing block.
+
+### Variant narrowing (`/ #lab`)
+
+The expression `e / #lab` removes tag `#lab` from a variant type, trapping if the value actually carries that tag. This is useful for asserting that a variant is *not* a particular case:
```
Kamirus, line 242: Why would we want such an operator? It seems risky to have because it traps.
Also, why is this together with the 'variant blocks'? They seem like two independent features

test/fail/ok/parse-block-or-record.tc-human.ok
Kamirus: From `DIVHASH x=id` ?

doc/md/fundamentals/3-types/10-options.md
crusso: yeah, I suggested this operator to the AI as a way to implement `!` efficiently, without reconstructing the variant value to propagate (needed when the success payload type changes. I think it would actually be more appropriate at the IR, not source level, to avoid the unchecked cast in desugaring.

But it's not totally silly as a source level feature either, tbh.

test/fail/ok/parse-block-or-record.tc-human.ok
crusso: Oh, I see, its the remainder of the production on 221, I guess `<exp_bin> <bin_op> <exp_bin>` maybe.

doc/md/fundamentals/3-types/10-options.md
crusso: I mean, it traps, but it's not unsafe.

crusso: The idea was that 
```
e !  ~~>
  switch e {
     #success v -> v
     other -> break <lab> (other / #success)
 ```
(but the AI doesn't actually use that, it does:

```
e !  ~~>
  switch e {
     #success v -> v
     other -> break <lab> (Cast<typ, typ / {# success _}>  other)
```
(exploiting our unsafe cast as IR level)

PR #5991, opened 2026-04-09
feat: implicit derivation of non-function values (objects, typeclasses)

Kamirus:
## Summary

Extends implicit derivation to produce values directly, not just wrapper functions. This enables typeclass-style patterns where bundled operations are represented as object types and derived via classes/functions with implicit constructor parameters.

**Problem**: Implicit derivation only supported function-typed holes — the compiler could synthesize a wrapper function, but couldn't produce a non-function value like an `Ord<[Nat]>` object from a class `ArrayEx.Ord<T>(Ord : (implicit : Ord<T>))`.

**Solution**: A new "value derivation" path recognizes candidates whose *return type* (not full function type) matches the hole type. Instead of wrapping in a `FuncE`, the compiler calls the candidate directly to produce the value. This works for candidates with implicit parameters (resolved recursively) and for candidates with only unit-typed explicit parameters and type parameters inferred from context.

**Refactoring**: Unified the two code paths to reduce duplication:
- `derived_wrapper` and `derived_call` merged into `SynthesizeWrapper.synthesize ~is_value`
- `sub_or_bimatch_return` eliminated (was a special case of `sub_or_bimatch_func`)
- `is_matching_typ_with_holes` refactored from two branches into demand-extraction + shared matching path
- Removed `List.rev` on derived bindings to preserve dependency order (critical for eager value evaluation)

## Test plan

- [x] New `test/run/implicit-derivation-object.mo` — covers `Ord` (with implicit element Ord) and `Monoid` (zero-implicit) typeclass derivation, including transitive derivation and explicit argument passthrough
- [x] Adjusted `test/run/implicit-derivation.mo` — scoped `Const.compare` to avoid ambiguity with value derivation candidates
- [x] All existing `run/implicit-derivation-*` and `fail/implicit-derivation-*` tests pass


doc/md/fundamentals/11-implicit-parameters.md
```diff
@@ -214,6 +216,52 @@ With derivation, the compiler handles this automatically. It recognizes that `Ar
 
 This works transitively: a `compare` for `[[Nat]]` is derived via `Array.compare<[Nat]>`, which needs `[Nat]` compare, which is derived via `Array.compare<Nat>`, which needs `Nat.compare` — all resolved automatically.
 
+#### Value derivation
+
+Derivation also works for non-function values such as objects. This enables typeclass-style patterns where related operations are bundled in an object type:
+
+```motoko no-repl
+type Ord<T> = {
```
crusso, line 224: If we allow (and perhaps even restrict) this for modules
```
type Ord<T> = module {...
```
Then most of it not all of our primitive modules would satisfy Ord<T>.
And `module classes` would be a neat way of defining "derived instances".
Also restricting to module and module classes might restrict the search space a teeny bit (we only have to consider functions that produce modules.. before checking if the type matches.).

Kamirus: Nice insight, so it is better to have modules, nice.

Unfortunately getting these dictionaries to replace the func-derivation from the main PR is tricky... Because how would we support `__record` instance for `Ord<T>`? These `__record` combinators only really work with deriving `func` directly, deriving modules with some funcs inside gets really tricky...

So I'd stay with the original approach and abandon this PR so that we have a single way of doing 'implicit entailment/derivation'.

crusso: The problem with record instances is equally true for object vs module dictionaries, right?
I think Christoph's take is that we should just use Haskell like `deriving` for special operations on records/variants. Maybe he's got a point because otherwise we will get (incorrect) equality on Maps just because they are implemented as records.

crusso: Although if maps and sets were "named" types, then I guess we could provide bespoke compare etc  by keying on the name, not the implementations under the names.

Kamirus: Direct candidates are preferred over `__record` which have lowest priority.
Haskell's `deriving` does not really fit, yet, in Motoko because we don't have nominal types and it would not solve the problem of writing general folds over arbitrary records.

These combiners allow generic things like this data viewer, having a compare for records by just by adding an import, allow package authors to provide generic toJson and all that for people to use by just importing it.

So with the `__record` combiners we get both the `deriving` and the general folds over records to do stuff like data-viewer and toJson and what not.
And these combiners are only used if there are no direct candidates -- just like we prefer direct values over module fields now

crusso:
Curious. If you have generic Map (a record) with (generic, paramterized) Map.compare and \__record, will your algorithm prefer to use Map.compare over \__record or report an ambiguity. I guess the former but just asking

Kamirus:
> Curious. If you have generic Map (a record) with (generic, paramterized) Map.compare and __record, will your algorithm prefer to use Map.compare over __record or report an ambiguity. I guess the former but just asking

Yes, this is by design, here is the doc that describes that https://github.com/caffeinelabs/motoko/pull/5903/changes#diff-c89de36493c8bfd5e65417b19ebc295ead3dae55345a3656249bd6485e7a9299R177-R192
and here is the code https://github.com/caffeinelabs/motoko/pull/5903/changes#diff-3f553f492c5fe9acdbed07b9024025f63fffac45f2ce15ef36f2bd7e2f376dc2R2008-R2021

`__record` has the lowest priority, if there is a matching `Map.compare` (direct or derivable by resolving and plugging implicit arguments) it will be used without ambiguity

Kamirus:
Cancelled

PR #5993, opened 2026-04-09
feat: fix graph copy stab/destab for multi-migration and dedup table

alexandru-uta:
The graph copy stabilization/destabilization algorithm basically does a BFS walk to first serialize the heap to stable memory, then after an upgrade a deserialization bringing back up the heap from stable memory. This BFS was initially designed to just start the BFS walk out of the stable actor record.

However, later additions to the enhanced OP incremental GC added 2 GC-only roots: the `blob deduplication table` and the `migration functions list`. Without proper handling of the serialization/deserialization, these objects would be lost and the new actor version would work erroneously.

This PR modifies the BFS algorithm to add the two objects as roots in the serialization/deserialization graph copy algorithm such that they are first saved properly to stable memory and then restored correctly into the heap. Other future GC roots can be treated similary.

To verify the functionality works correctly, two tests were modified to trigger this behavior and check if the new upgraded actor works correctly.

**Attention point:**
This PR also fixes a corner-case. For enhanced multi-migration, which migration function (in the chain) is matched against the old actor type is only determined at runtime (see `desugar.ml`), depending on which migration function in the chain it's been brought at previously. The old code in `compile_enhanced.ml` was performing compatibility checks against the new actor `.pre`, which can be wrong for multi-migration in case there might be multiple steps missing. Originally, the code made a check against old actor and new actor `.pre` types early in the `start destabilization` code path. This cannot work with the new migration. 

The solution is to restore into `metadata` the old actor type from stable memory and let the checks happen at `ICStableRead` time (via `register_stable_type`, see `compile_enhanced.ml`). Previously, this metadata was wiped and `ICStableRead` was not performing any checks, just recovering the actor.

This way, the checks happen correctly at runtime for both enhanced multi-migration and regular migration (or upgrade).


src/codegen/compile_enhanced.ml
```diff
@@ -10391,12 +10391,12 @@ module IncrementalGraphStabilization = struct
   let partial_destabilization_on_upgrade env actor_type =
     (* TODO: Verify that the post_upgrade hook cannot be directly called by the IC *)
     (* Garbage collection is disabled in `start_graph_destabilization` until destabilization has completed. *)
-    GraphCopyStabilization.start_graph_destabilization env actor_type.Ir.pre ^^
+    GraphCopyStabilization.start_graph_destabilization env actor_type.Ir.post ^^
```
crusso: Why the change from pre to post? With single migration, destabilizing at pre would (I think) check the old actor type is compatible with the input to the migration function. This seems like it might be dangerous.

alexandru-uta: check once more; it's only for `--enhanced-migration`

test/run-drun/enhanced-migration-incremental.drun
```diff
@@ -11,5 +11,19 @@ ingress $ID check "DIDL\x00\x00"
 upgrade $ID enhanced-migration-incremental/version4.mo ""
 ingress $ID check "DIDL\x00\x00"
 
+# Add a call to stabilization to check if graph copy works properly
+# with the dedup table.
```
crusso:
```suggestion
# with the rts migration list.
```

crusso (changes requested):
After staring long and hard at this I don't think it's quite right. It's probably consistent internally, but I think it breaks backward compatibility.

Essentially, we are changing the graphical format  (from old single record to new record and two more value) without versioning it properly, which is asking for trouble.

For instance, if we upgrade from the graph format dumped by an older version, we will try to deserialize a de_dup table and migration list that just aren't there. 

I think we instead need to add a new (minor) version of the graph format:

VERSION_GRAPH_COPY_V1_NO_REGIONS, VERSION_GRAPH_COPY_V1_REGIONS
(and version_graph_copy_v1_no_regions etc in compile.ml)

and act accordingly.

Serialization would emit a triple (or record) of stable variable record, dedup_table and migration list and write the XXX_V1_XXX version to meta-data.

Deserialization would look at the (graphical) version and either parse just a record, and construct and  empty dedup table and migration list, or (for  XXX_V1_XXX), parse a full triple.

Most of the action could probably happen in compile_enhanced.ml, reverting some of the changes to the rts, if that's any easier (that's just a hunch though)

crusso:
Took a quick look at recent changes. Look good, but might take another closer look on Friday.

crusso (approved):
LGTM

Thanks for adding the test!

PR #5996, opened 2026-04-09
feat: opt-in non-Candid wire format via `encoder` / `decoder` parenthetical

ggreif:
## Summary

Per-public-method, opt-in escape hatch from Candid in both directions:

```motoko
persistent actor {
  (with encoder = <T -> Blob> ;
        decoder = <Blob -> A>)
  public func go(_ : A) : async T = …
}
```

- `encoder : T -> Blob` replaces Candid **serialization on reply**.
- `decoder : Blob -> A` replaces Candid **deserialization on ingress**.

Both fields are independently optional; either, both, or neither can appear, and unrecognised fields warn (M0212). The strategic driver is direct OpenAPI/Web2 interfacing — see `.claude/plans/non-candid.md` for the design memo, including the actor-level inheritance idea and the motivation. (The OpenAPI integration is its own follow-on.)

### Frontend (`src/mo_frontend/typing.ml`)
- `check_vis_parenthetical` does per-field bidirectional checking against a known-fields table `[("encoder", T -> Blob); ("decoder", Blob -> A)]`. Encoder type is driven by the method's return type, decoder by its ingress type. Direction (method → codec) is documented in a comment alongside the alternative inverse direction (codec types driving the method signature) for future exploration.
- M0214 type-mismatch (per-field), M0215 effect-free check named per field, M0211 redundant-empty parenthetical, M0212 unrecognised attribute.

### IR (`src/ir_def/ir.ml`)
- `FuncE`'s trailing `exp option (* encoder *)` replaced by a labeled record `codecs = { encoder : exp option; decoder : exp option }`. `no_codecs` helper for the common no-annotation case. Future codec-shaped fields (e.g. inbound-cycles caps, schema pins) land additively.
- `check_ir` independently type-checks both fields.

### IR passes
- `rename`, `subst_var`, `freevars`, `await`, `async`, `const`, `tailcall`, `erase_typ_field`, `eq`, `show` — pass-through, retyped from `exp option` to `codecs`.
- `arrange_ir.ml` prints both fields side by side.

### Desugaring (`src/lowering/desugar.ml`)
- `find_codec_in_par` factored out from the existing `find_encoder_in_par`.
- New `find_decoder_in_par` mirror.
- `build_codecs` (replacing `build_encoders`) returns `(enc_opt, dec_opt) list`.
- `build_actor` stamps both fields onto each public method's `FuncE.codecs`, preserving any pre-existing per-field override.

### Async lowering (`src/ir_passes/async.ml`)
- The CPS transform consults `codecs.encoder` when synthesising the reply continuation, lifting it into `ICReplyPrim(ts, Some enc)`. Decoder unaffected here — it lives only on `FuncE`.

### Codegen (`src/codegen/compile_classical.ml` & `compile_enhanced.ml`)
- **Encoder** (existing): `ICReplyPrim` branches on `enc_opt` — `Some` → call closure, `IC.reply_with_data`; `None` → `Serialization.serialize`.
- **Decoder** (new): `FuncDec.{lit, closed, compile_const_message}` gain a `?(decoder=None)` thunk parameter `(E.t -> VarEnv.t -> G.t) option`. The thunk is constructed at the FuncE call site (where `compile_exp_vanilla` is in scope) wrapping the decoder expression. Inside `compile_const_message`, branch at the argument-decoding step: `None` → `Serialization.deserialize`; `Some compile_dec` → `compile_dec env ae0 ; closure-call` on raw `IC.arg_data`.

### AST interpreter (`src/mo_interpreter/interpret.ml`)
- Codec parentheticals are intentionally not visited (the AST interpreter doesn't model Candid serialization). One-line comment surfaces this for reviewers.

## Tests

Run-drun:
- `parenthetical-public.mo` — encoder, returns `()`, all phases.
- `parenthetical-decoder.mo` — full pipeline. Method's ingress is `?Nat`; decoder is the flow `Blob -> ?Text -> ?Nat` composed via `decodeUtf8` and `Nat.fromText`. The `//CALL` payload is the raw three ASCII bytes `"123"` (`0x313233`) — *not* a Candid envelope. With the decoder active that blob deserialises as `?123`, and the reply is Candid `Nat 123` (`0x4449444c00017d7b`). Without the decoder Candid would reject `0x313233` as malformed input — so the green test is end-to-end proof that ingress Candid is bypassed.

Fail (matched pairs, encoder ↔ decoder):
- `parenthetical-{encoder,decoder}-effect.mo`   — M0215 effect-free check fires.
- `parenthetical-{encoder,decoder}-mismatch.mo` — M0095 (finer than field-level M0214) on a wrong codec signature.

## Test plan

- [ ] `make -C test/run-drun parenthetical-public.only` passes
- [ ] `make -C test/run-drun parenthetical-decoder.only` passes
- [ ] `make -C test/fail` includes the four new fail tests with stable output


ggreif:
## Grammar conflict on the encoder parenthetical — and a conflict-free fix (verified prototype)

Attaching the encoder *before* `public` — `(base with encoder = …) public func …` — collides with actor-level tuples: a field beginning with `(` is ambiguous between a parenthesized/tuple expression field and the leading encoder parenthetical, so it trips `menhir --strict`.

### Proposal: hang the (optional) encoder off `public`

`public (base with encoder = …) func …`. `public` is the natural anchor — only public (shared) members are marshalled, so the encoder is inherently a public-function concern.

### Why the naive version still conflicts

Even after `public`, `public (` is ambiguous: `(` can begin the encoder **or** a bare parenthesized-expression field, because a field's `dec` can start with `(` — via `exp_nondec`, and via `dec_nonvar`'s existing `(base with …) obj/class` extension. `menhir --strict` reports a shift/reduce in exactly this spot (`reduce vis → PUBLIC` vs `shift LPAR`). Parametrising alone doesn't remove it — the ambiguity lives *inside* the actor-context variant.

### The fix (conflict-free)

Make `vis` own its continuation and parametrise the field by context — **`dec_field(ac|lo)`** (two variants):

- **`lo`** — *local*: module / object / object-class. `vis` + general `dec` — unchanged.
- **`ac`** — *actor-context*: actor / actor-class / **mixin** (a mixin's public members become the actor's, so it shares this variant). `public = PUBLIC parenthetical? dec_pub`, where `parenthetical` is the *existing* `(base with … = …)` nonterminal and `dec_pub = { let, type, func }`. Non-public members (`private`/`system`/default) keep the general `dec`, so private funcs, state, etc. are unaffected.

Actor, actor-class, and mixin bodies have identical public-member rules (both accept `func` / `let = func` / `type`; both reject a public value with M0124) and share the same field grammar, so there is **no `ac` vs `mi` distinction** — one restricted variant, wired at three sites.

Because `dec_pub` never begins with `(`, `public (` is unambiguously the encoder — no reduce alternative. `menhir --strict`: **no shift/reduce conflict.**

Two dead-ends worth recording:
- `dec_func` (func-only) is conflict-free but **too narrow** — it rejects `public type` and `public let = func` (the latter being the desugaring of `public func`, so it must be admitted).
- `dec` minus `exp_nondec` is **too broad** — `dec_nonvar`'s `(base with …) obj/class` extension is also `LPAR`-leading and re-introduces the conflict.

`dec_pub = { let, type, func }` is the sweet spot: exactly the valid public members, none `LPAR`-leading.

### A nice side effect

This enforces **M0125** ("a public actor field needs to be a manifest function") at the *grammar* level: `public 5` / `public (1,2)` become parse errors rather than type errors — arguably a virtue (grammar-as-spec). `public func` / `public let = func` / `public type` all keep working.

### Verified prototype

Pushed to **`gabor/vis-experiment`** (off `master`), commit `987ff934c` — the `dec_pub` variant, `menhir --strict` clean. (`f83e32de3` is the earlier too-narrow func-only cut.) It wires the `mixin` body as the demo `ac` site; the actor block + actor class route to the same variant.

Remaining to productionise: route the actor block + actor class to the `ac` variant; thread the `parenthetical` value into the field AST for the marshalling machinery; add the type-checker guard that the encoder attaches only to a public *manifest function* (not `public type`).

PR #6001, opened 2026-04-10
chore: deduplicate `pos` and `region`

ggreif:
This is very much a huge patch, but potentially removes a big thorn from our side. The `Wasm.Source` vs. `Lang_Utils.Source` ambiguity was nasty as hell! Everything here is completely mechanic.

We now use `pos` and `region` from `Wasm.Source` throughout, no own generative type in `Source`. They are the same type now. I most cases `open`ing `Source` is enough.

Finished the `Source.phrase` -> `Ir.phrase` transition that was done half-way. (Modulo where `note` is unit, there we reuse `Source.phrase`.)

TODO:
- [x] remove thousands of `Source.` qualifications. Not needed any more as there are no more ambiguities.

### DIFF
--- a/src/codegen/instrList.ml
+++ b/src/codegen/instrList.ml
@@ -7,12 +7,6 @@ features are
  * Some simple peephole optimizations.
 *)
 
-(*let _cr Source.{left; right} =
-  let left, right =
-    (let Source.{ file; line; column } = left in Wasm.Source.{ file; line; column }),
-    (let Source.{ file; line; column } = right in Wasm.Source.{ file; line; column })
-  in Wasm.Source.{ left; right }*)
-
 open Wasm_exts.Ast
 open Wasm.Source
 open Wasm_exts.Values
--- a/src/codegen/instrList.ml
+++ b/src/codegen/instrList.ml
@@ -11,8 +11,6 @@ open Wasm_exts.Ast
 open Wasm.Source
 open Wasm_exts.Values
 
-let cr region = region
-
 let combine_shifts const op = function
   | I32 opl, ({it = I32 l'; _} as cl), I32 opr, I32 r' when opl = opr ->
     let l, r = Int32.(to_int l', to_int r') in
@@ -194,7 +192,7 @@ let table n f = List.fold_right (^^) (Lib.List.table n f) nop
 (* Region-managing combinator *)
 
 let with_region (pos : Source.region) (body : t) : t =
-  fun d _pos rest -> body d (cr pos) rest
+  fun d _pos rest -> body d pos rest
 
 (* Depths-managing combinators *)
 
--- a/src/codegen/instrList.ml
+++ b/src/codegen/instrList.ml
@@ -7,16 +7,17 @@ features are
  * Some simple peephole optimizations.
 *)
 
-open Wasm_exts.Ast
-
-let cr Source.{left; right} =
+(*let _cr Source.{left; right} =
   let left, right =
     (let Source.{ file; line; column } = left in Wasm.Source.{ file; line; column }),
     (let Source.{ file; line; column } = right in Wasm.Source.{ file; line; column })
-  in Wasm.Source.{ left; right }
+  in Wasm.Source.{ left; right }*)
 
+open Wasm_exts.Ast
 open Wasm.Source
 open Wasm_exts.Values
+
+let cr region = region
 
 let combine_shifts const op = function
   | I32 opl, ({it = I32 l'; _} as cl), I32 opr, I32 r' when opl = opr ->
@@ -331,7 +332,7 @@ let dw_tag die body =
 let dw_tag_no_children = dw_tag_open (* self-closing *)
 
 (* Marker for statement boundaries *)
-let dw_statement { Source.left; Source.right } =
+let dw_statement { left; right } =
   let open Wasm.Source in
-  let left = { file = left.Source.file; line = left.Source.line; column = left.Source.column } in
+  let left = { file = left.file; line = left.line; column = left.column } in
   i (Meta (StatementDelimiter left))
--- a/src/codegen/instrList.ml
+++ b/src/codegen/instrList.ml
@@ -324,7 +324,5 @@ let dw_tag die body =
 let dw_tag_no_children = dw_tag_open (* self-closing *)
 
 (* Marker for statement boundaries *)
-let dw_statement { left; right } =
-  let open Wasm.Source in
-  let left = { file = left.file; line = left.line; column = left.column } in
+let dw_statement { left; _ } =
   i (Meta (StatementDelimiter left))
--- a/src/idllib/lexer.mll
+++ b/src/idllib/lexer.mll
@@ -1,17 +1,18 @@
 {
+open Wasm.Source
 open Parser
 module Utf8 = Lib.Utf8
 
 let convert_pos pos =
-  { Source.file = pos.Lexing.pos_fname;
-    Source.line = pos.Lexing.pos_lnum;
-    Source.column = pos.Lexing.pos_cnum - pos.Lexing.pos_bol
+  { file = pos.Lexing.pos_fname;
+    line = pos.Lexing.pos_lnum;
+    column = pos.Lexing.pos_cnum - pos.Lexing.pos_bol
   }
 
 let region lexbuf =
   let left = convert_pos (Lexing.lexeme_start_p lexbuf) in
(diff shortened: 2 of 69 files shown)


src/mo_def/arrange.ml
```diff
@@ -33,11 +33,15 @@ module Make (Cfg : Config) = struct
...
 
   let pos p =
+    let open Wasm.Source in
     let file = match Cfg.main_file with
     | Some f when f <> p.file -> p.file
     | _ -> ""
     in "Pos" $$ [Atom file; Atom (string_of_int p.line); Atom (string_of_int p.column)]
-  let source at it = if Cfg.include_sources && at <> Source.no_region then "@" $$ [pos at.left; pos at.right; it] else it
+
+  let source at0 it =
```
ggreif:
```suggestion
  let source at' it =
```
The problem is shadowing below.

src/codegen/instrList.ml
```diff
@@ -7,17 +7,18 @@ features are
...
     (let Source.{ file; line; column } = left in Wasm.Source.{ file; line; column }),
     (let Source.{ file; line; column } = right in Wasm.Source.{ file; line; column })
-  in Wasm.Source.{ left; right }
+  in Wasm.Source.{ left; right }*)
 
+open Wasm_exts.Ast
 open Wasm.Source
 open Wasm_exts.Values
 
+let cr region = region
```
ggreif: Now we can drop `cr`! :-)

src/idllib/lexer.mll
```diff
@@ -1,17 +1,18 @@
 {
+open Wasm.Source
 open Parser
 module Utf8 = Lib.Utf8
 
 let convert_pos pos =
-  { Source.file = pos.Lexing.pos_fname;
-    Source.line = pos.Lexing.pos_lnum;
-    Source.column = pos.Lexing.pos_cnum - pos.Lexing.pos_bol
+  { file = pos.Lexing.pos_fname;
```
ggreif:
```suggestion
  Lexing.{ file = pos.pos_fname;
```

src/js/astjs.ml
```diff
@@ -307,8 +304,7 @@ module Make (Cfg : Config) = struct
 
   let rec exp_js e =
     let open Syntax in
-    let open Source in
```
ggreif, line 310: What about re-adding?

src/lang_utils/source.mli
```diff
@@ -1,5 +1,5 @@
-type pos = {file : string; line : int; column : int}
-type region = {left : pos; right : pos}
+type pos = Wasm.Source.pos
```
ggreif: I just found this: https://github.com/ocaml/ocaml/issues/7025#issue-421232737

ggreif: And it works as advertised!

src/codegen/instrList.ml
ggreif: Done!

ggreif:
## Pre-Review: `chore: deduplicate pos and region`

Overall approach is sound and well-motivated — making `Source.pos`/`Source.region` type-aliases for `Wasm.Source.pos`/`Wasm.Source.region` eliminates the previous need for the `cr` conversion function in `instrList.ml` and removes a whole class of boilerplate. The mechanical nature of most changes is a good sign.

---

### Critical: missing `wasm` dependency in `src/lang_utils/dune`

`src/lang_utils/source.ml` and `source.mli` now reference `Wasm.Source.pos` and `Wasm.Source.region` directly:

```ocaml
(* source.ml / source.mli *)
type pos = Wasm.Source.pos = { file : string; line : int; column : int }
type region = Wasm.Source.region = { left : pos; right : pos }
open Wasm.Source
```

But the `lang_utils/dune` file is unchanged:

```
(libraries lib mo_config yojson grace grace.ansi_renderer)
```

`wasm` is not listed. This will fail to compile. The `wasm` library needs to be added to `lang_utils/dune`'s `libraries` stanza.

> Strangely it _does_ compile. There might be some defaulting mechanism in `dune`. I added it and it still compiles just fine. Keeping it: see 585262f0f.

(Note: the PR correctly adds `lang_utils` to `wasm_exts/dune`, but the reverse direction — `wasm` into `lang_utils` — is missing.)

---

### Likely bug: incorrect record construction syntax in `src/idllib/lexer.mll`

The `convert_pos` function is changed to:

```ocaml
let convert_pos pos =
  Lexing.{ file = pos.pos_fname;
    line = pos.pos_lnum;
    column = pos.pos_cnum - pos.pos_bol
  }
```

The prefix `Lexing.{ ... }` in OCaml means "disambiguate field names using the `Lexing` module". But `Lexing` has no record type with fields `file`, `line`, `column` — those belong to `Source.pos`. This should be:

```ocaml
let convert_pos pos =
  { file = pos.Lexing.pos_fname;
    line = pos.Lexing.pos_lnum;
    column = pos.Lexing.pos_cnum - pos.Lexing.pos_bol
  }
```

Since `open Source` is now in effect at the top of the lexer, the field names `file`, `line`, `column` are unambiguous and the `Source.` prefix on the fields is no longer needed.

> I do `Lexing.{` to access `pos_fname` and co. Won't change.

---

### `open Wasm.Source` violation outside `codegen/`: `src/linking/linkModule.ml` (pre-existing, not addressed)

Per the stated policy, `open Wasm.Source` should be restricted to `codegen/`. The following file outside that directory still uses it:

- `src/linking/linkModule.ml:7` — `open Wasm.Source`

This is pre-existing and not introduced by this PR, but since the PR's stated goal is to eliminate the ambiguity, it should ideally be addressed here too. After the type unification, `linkModule.ml` can simply use `open Source` (from `lang_utils`) or rely on whatever `Source` is in scope.

Similarly, `src/wasm-exts/ast.ml:37` and `src/wasm-exts/operators.ml:18` still use `open Wasm.Source`, but these are in `wasm-exts/` which directly wraps `wasm`, so the argument for keeping `Wasm.Source` there is stronger.

> As the linker comes after codegen this is acceptable too.

---

### Missed simplification in `src/codegen/instrList.ml`, `dw_statement`

After the type unification, `Source.pos` and `Wasm.Source.pos` are the same type. The `dw_statement` function still has redundant field-by-field reconstruction:

```ocaml
let dw_statement { left; right } =
  let open Wasm.Source in
  let left = { file = left.file; line = left.line; column = left.column } in
  i (Meta (StatementDelimiter left))
```

Since `left` is already a `pos` (= `Wasm.Source.pos`), the `let open Wasm.Source in` and the record copy are unnecessary. This could simply be:

```ocaml
let dw_statement { left; _ } =
  i (Meta (StatementDelimiter left))
```

(The `right` binding is not used anyway.)

> Well spotted! Simplified!

---

### Minor: redundant `let no_region = Source.no_region` in `src/ir_def/ir.ml`

`ir.ml` now has `open Source` and also:

```ocaml
let no_region = Source.no_region
```

After `open Source`, `no_region` is already directly in scope. The `let` re-binding is redundant (though harmless — it may be intentional for explicit re-export into the `Ir` module's namespace).

> Removing `Source.`, this still exports.

---

### Minor: awkward `Source.(fun ...)` in `src/pipeline/test_field_srcs.ml`

The change uses a local-open-in-expression form:

```ocaml
Seq.iter
  Source.(fun (define, origins) ->
    Format.printf "%s:" (string_of_region define);
    ...
    (Region_set.to_seq origins);
```

This is valid OCaml but inconsistent with the rest of the PR's approach of adding `open Source` at the module top level. A simple `open Source` at the top of the file (which already has `open Source` added elsewhere in the diff for other files) would be cleaner.

> Done!

---

### Summary

The overall design is correct and the type-alias approach for `pos`/`region` is the right mechanism. The main blockers before merge are the missing `wasm` build dependency in `lang_utils/dune` and the malformed record construction in `idllib/lexer.mll`. The `open Wasm.Source` in `linking/linkModule.ml` is a policy gap worth closing in this PR.

src/idllib/lexer.mll
ggreif: Done.

src/js/astjs.ml
ggreif: No!

src/mo_def/arrange.ml
ggreif: Done!

Kamirus (approved):
skimmed thru, looks good but too many changes to be sure, stamping
Looks like a nice cleanup

PR #6003, opened 2026-04-10
experiment: incremental compilation

Kamirus:
## Experiment: Incremental Compilation

Experimental incremental compilation support for `moc`. Two caching strategies to reduce compile times when dependencies (packages like `mo:base`, `mo:core`) stay the same. Both use the `--moi-cache <dir>` flag.

---

### Experiment 1: Scope caching (`.moi`) — `--check` mode

**Goal:** Speed up repeated type-checking by caching each library's type-checked scope.

**Approach:** Serialize `Scope.t` per library to `.moi` files using a custom binary format, handling cyclic type constructors. Cache key is a Merkle hash of the library's source content and its transitive dependencies. On cache hit, type-checking for that library is skipped entirely.

**Results (Map.test.mo, motoko-core, `$(mops sources)`, 78 dependency files):**

| Scenario | Time |
|---|---|
| `--check` (no cache) | ~287ms |
| `--check` + `.moi` cache (cold) | ~287ms |
| `--check` + `.moi` cache (warm) | ~229ms |

The scope cache saves ~58ms in check mode (~20% of check work).

**Conclusion:** Modest speedup. Type-checking is already fast, so caching it saves little in absolute terms.

#### Profiling: where does compile time go?

This experiment also revealed the full compile-time breakdown, motivating experiment 2:

| Phase | Time | % |
|---|---|---|
| Parse + check (per-lib) | 167ms | 11% |
| Lowering | 61ms | 4% |
| 7 IR passes (whole-program) | 1,076ms | **73%** |
| Codegen | 164ms | 11% |
| **Total (`-c`)** | **~1,760ms** | |

The seven whole-program IR passes (`erase_typ_field`, `show`, `eq`, `await`, `async`, `tailcall`, `const`) dominate compile time at 73%.

---

### Experiment 2: IR caching (`.moic`) — `-c` compile mode

**Goal:** Speed up repeated compilation by caching post-IR-pass library declarations.

**Approach:** After lowering and running all seven IR passes on library code, serialize the resulting `Ir.dec list` to `.moic` files using `Marshal`. On warm cache, the compiler skips lowering and all IR passes for unchanged libraries — only the main program goes through the full pipeline, then gets linked with cached library IR before codegen.

**Results (Map.test.mo, motoko-core, `$(mops sources)`):**

| Build | CPU time | Speedup |
|-------|----------|---------|
| Baseline (`-c`) | 1.56s | — |
| Cold cache | 1.74s | 0.9x (write overhead) |
| **Warm cache** | **0.68s** | **2.3x** |

Cold and warm produce byte-identical Wasm. Cache invalidation works correctly.

**Where the remaining time goes (warm, 0.68s):**
- ~0.20s — parse + typecheck all files (re-done every time)
- ~0.48s — lower main program + codegen on the full combined program

**Codegen is now the bottleneck** — it still walks all declarations (cached + main). Eliminating this would require separate Wasm compilation + linking, which is a fundamentally larger change.

---

### Overall conclusion

~350 lines of new code for a 2.3x compile-mode speedup. The IR caching approach works but hits diminishing returns: the next meaningful improvement requires Wasm-level separate compilation, not more caching. This POC uses `Marshal` (tied to exact compiler binary) and has no cache eviction — not production-ready as-is.

### Test plan

- [x] `test/run/moi-cache.mo` — roundtrip: counter, types with generics, recursive tree types
- [x] `test/fail/moi-cache-error.mo` — type errors still reported correctly with cache
- [x] Compile-mode IR caching — byte-identical Wasm on Map.test.mo, cache invalidation verified

### DIFF
--- a/.cursor/plans/compile-mode_ir_caching_8a12d666.plan.md
+++ b/.cursor/plans/compile-mode_ir_caching_8a12d666.plan.md
@@ -1,0 +1,367 @@
+---
+name: Compile-mode IR caching
+overview: "Cache the post-IR-pass lowered IR for all dependency libraries to disk. On subsequent compiles where only the main program changes, load the cached IR and skip parsing, type-checking, lowering, and all 7 IR passes for dependencies. Expected savings: ~1,300ms out of 1,760ms (~74% faster) on Map.test.mo."
+todos:
+  - id: cli-update
+    content: Allow --moi-cache with -c mode in moc.ml (remove the --check restriction)
+    status: pending
+  - id: fresh-scoping
+    content: Add ir_scope to construct.ml fresh to namespace generated names by library context, and add bump_stamps to cons.ml for post-deserialization stamp safety
+    status: pending
+  - id: const-initial-env
+    content: Modify const.ml analyze to accept optional initial env (known_const binding names from cached lib decs)
+    status: pending
+  - id: ir-serialization
+    content: Implement IR cache serialization using Marshal with compiler version validation and Cons stamp bumping (in ir_cache.ml or extending moi_cache.ml)
+    status: pending
+  - id: pipeline-restructure
+    content: "Restructure compile_progs in pipeline.ml: split library and main IR processing, integrate IR cache load/save, handle cache-miss fallback by bypassing scope cache"
+    status: pending
+  - id: show-eq-dedup
+    content: Implement show/eq helper deduplication when linking cached lib decs with main decs (filter @show/@eq LetD duplicates)
+    status: pending
+  - id: check-ir-adjustment
+    content: Adjust Check_ir to run per-fragment only (before caching for libs, after passes for main), skip on combined IR
+    status: pending
+  - id: tests
+    content: "Add compile-mode cache tests: warm cache produces identical Wasm, cache invalidation on dep change, actor-class libs bypass cache"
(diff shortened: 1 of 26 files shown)


ggreif:
Do you use https://github.com/ocaml-ppx/ppx_deriving_protobuf#usage ?

Kamirus:
> Do you use https://github.com/ocaml-ppx/ppx_deriving_protobuf#usage ?

No, it was a quick POC measuring the speedup potential by caching the typing env and the IR.
So it was easier to do the de/serialization manually than to refactor our current codebase to fit in to the 'deriving' requirements

crusso:
Something is a bit fishy since I don't think we currently run IR-passes on library code at all - we just do them on the combined IR. 

Won't this produce more and redundant code, like multiple versions of the same `eq`  and `show` functions?

Kamirus:
> Something is a bit fishy since I don't think we currently run IR-passes on library code at all - we just do them on the combined IR.
> 
> Won't this produce more and redundant code, like multiple versions of the same `eq` and `show` functions?

Yes it would. I was just experimenting with this idea first trying to measure how faster compilation would get if all lib-IR was cached.

Kamirus:
Experiment done

PR #6009, opened 2026-04-12
chore: bump `clang` to v20

ggreif:
The linker `wasm-ld` has changed. It is necessary to check whether the now missing export is in any way a problem. So far no visible problem — Good!

Wait till we are on `nixpkgs-26.05`.

After this bump I don't expect problems with moving to `clang-21`. Then the `nix` closure will only contain a single `llvm`/`clang`!


rts/Makefile
```diff
@@ -286,6 +285,12 @@ EXPORTED_SYMBOLS=\
   fmodf \
   log
 
+EXPORTED_SYMBOLS_64=\
```
ggreif, line 288: This is dead code. None of our RTS libraries export it.

ggreif: Dropped in `6373bcf91` — confirmed `EXPORTED_SYMBOLS_64` was defined but referenced nowhere in `rts/Makefile`, and `LINKER_OPTIONS_64` already uses plain `EXPORTED_SYMBOLS`. Whole 4-line block gone.

crusso (approved):

ggreif:
On the back of the just-landed `gc-tests` matrix-passed `max-jobs` fix (commit `afd242d4a` plus the `951911982` matrix restructure): we could apply the exact same shape to the `common-tests` and `tests` matrices in `test.yml` to preempt future cold-cachix OOMs on `arm64-linux-16`.

```yaml
common-tests:
  strategy:
    matrix:
      include:
        - os: ubuntu-24-large
          max_jobs: auto
        - os: arm64-linux-16
          max_jobs: 1
    fail-fast: false
  ...
  with:
    ...
    max-jobs: ${{ matrix.max_jobs }}
```

(and the same for the `tests` matrix, with `build_type: [release, debug]` preserved.)

**Pros:**
- Consistent pattern across all three matrices — no surprise when the next matrix grows a third memory-pressure hotspot.
- Future-proofs `common-tests` / `tests` if their resource footprint grows on a cold cachix (e.g. via a stdlib bump, an additional RTS variant, or a larger test suite).
- One less "why does only `gc-tests` use the include form?" question for future contributors.

**Cons:**
- Wall-clock penalty on cold-cachix arm64 runs. `gc-tests` paid ~7 min for the serial run vs. the ubuntu sibling's ~40 s, so call it ~6 minutes per affected matrix. That's only paid when cachix is genuinely cold.
- Currently neither `common-tests` nor `tests` exhibits the OOM, so we'd be paying a hypothetical-prevention tax.

I'd lean **for** the change since the cost is bounded (cachix-cold only, single-digit minutes) and the consistency win is concrete. If you want, I can make this the next commit on this branch; or we can let it land as a follow-up after #6009 merges. Your call.

PR #6005, opened 2026-04-10
chore: add `build-and-test` Cursor skill for AI agents

Kamirus:
## Summary
- Adds a `.cursor/skills/build-and-test/SKILL.md` that teaches AI agents how to build `moc` and run tests
- Uses `test-runner` (parallel, auto-flags) instead of raw `make`/`run.sh`
- Worktree-safe: uses `git rev-parse --show-toplevel` instead of hardcoded paths
- Covers environment setup (direnv/nix), building (dune), running tests, test file conventions, and troubleshooting common failures

## Test plan
- [x] Verified `dune build --root src` works from repo root
- [x] Verified `test-runner -baf fail/` runs 359 tests successfully (~23s)

Made with [Cursor](https://cursor.com)

PR #6014, opened 2026-04-13
chore: apply review feedback to `build-and-test` skill

Kamirus:
## Summary

Follow-up to #6005 — applies @ggreif's review feedback and moves the skill to a shared location.

- **Move skill** from `.cursor/skills/` to `.agents/skills/` so both Cursor and Claude agents discover it
- **Mention `Makefile`** as a build alternative alongside raw `dune` commands
- **Fix time estimate** for the `run/` suite (~1–2 min, not "close to an hour")
- **Document `//FILTER`** test convention for piping output through a command before comparison
- **Document `//CALL` / `//OR-CALL`** conventions for `drun` tests, with a warning that at least one `//CALL` is required for `run-drun/` tests

## Test plan

- [ ] Verify Cursor agent can still discover and use the skill from `.agents/skills/`
- [ ] Verify Claude agent can discover and use the skill from `.agents/skills/`

### DIFF
--- a/.agents/skills/build-and-test/SKILL.md
+++ b/.agents/skills/build-and-test/SKILL.md
@@ -53,6 +53,13 @@ that `dune build exes/moc.exe` would miss.
 
 `src/moc` is a symlink to `_build/default/exes/moc.exe`.
 
+Alternatively, `src/Makefile` wraps dune with convenient targets:
+
+```bash
+make -C src          # build all native exes + grammar (equivalent to dune build)
+make -C src moc      # build just moc
+```
+
 The user typically runs `dune build -w` in a separate terminal for continuous rebuilds.
 If the user says they have a watch build running, skip the build step — `src/moc` is
 already up to date.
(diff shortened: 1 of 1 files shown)


.agents/skills/build-and-test/SKILL.md
```diff
@@ -87,9 +94,10 @@ test-runner -bf fail/              # run without accepting (check for diffs)
...
-**Avoid running full `run/` or `run-drun/` suites locally** — they take close to
-an hour. The full `fail/` suite is fine (~25 seconds). For `run/` and `run-drun/`,
-run only the specific tests relevant to your change:
+**Avoid running full `run/` or `run-drun/` suites locally** — `run/` takes ~1–2
+minutes, `run-drun/` significantly longer. The full `fail/` suite is fine (~25
+seconds). For `run/` and `run-drun/`, prefer running only the specific tests
+relevant to your change:
 
 ```bash
 test-runner -baf some-test         # targeted by name pattern
```
ggreif, line 103: `-baf` will update the `.ok` files. Maybe that is not what you want for just checking. That is not the regular mode.

ggreif (approved):
Left a comment. With relation to CI failures I normally just prompt a spawned agent to check the PR (`gh` tool) and run `nix build` on the failed flake. Then it corrects all kinds of stuff.

.agents/skills/build-and-test/SKILL.md
Kamirus: I actually want that, I always run with check so I see which files have changed. I'd rather look at the github diff locally than terminal output

PR #6020, opened 2026-04-15
moc.js: add `resetExtraFlags` API

Kamirus:
## Problem

`setExtraFlags(["--enhanced-migration=migrations"])` sets `Flags.enhanced_migration := Some "migrations"`. A subsequent `setExtraFlags(["--all-libs"])` does **not** clear `enhanced_migration` — it stays `Some "migrations"` forever. There is no way to "unset" a flag once set via `setExtraFlags`.

This breaks multi-canister projects (e.g. in the VS Code extension) where only some canisters use `--enhanced-migration`. Flags from one canister's type-check bleed into the next.

## Fix

Two new APIs on the moc.js `Motoko` object:

### `clearExtraFlag(name)`

Clears a single flag by its CLI name (the same format used in `mops.toml` and `setExtraFlags`). Throws on unknown flag names.

```js
Motoko.clearExtraFlag("--enhanced-migration");  // resets to None
Motoko.clearExtraFlag("--error-detail");        // resets to 4 (moc.js default)
Motoko.clearExtraFlag("--nonexistent");          // throws
```

Supported flags: `--hide-warnings`, `-Werror`, `-A`/`-W`/`-E`, `--error-detail`, `--error-recovery`, `--error-format`, `--ai-errors`, `--all-libs`, `--implicit-package`, `--enhanced-migration`, `--default-persistent-actors`/`--require-persistent-actors`/`--legacy-actors`.

### `resetExtraFlags()`

Resets **all** flags touchable by `setExtraFlags` back to their moc.js defaults. Implemented by iterating over all known flag names and calling the same clear logic. Convenience method for the common pattern: `resetExtraFlags()` then `setExtraFlags(perCanisterFlags)`.

Flags managed by dedicated APIs (`setProjectRoot`, `addPackage`, `setCandidPath`, etc.) are intentionally not touched by either method.

## Caveat — js_of_ocaml calling convention

`method m () = ...` in js_of_ocaml compiles to a 2-arity JS function (`this` + unit). Calling `Motoko.resetExtraFlags()` with 0 arguments triggers OCaml currying and returns a partial application. Callers must pass a dummy argument: `Motoko.resetExtraFlags(0)`. Same applies to the existing `clearPackage`. node-motoko's `invoke()` handles this transparently.

## Downstream changes needed

- **node-motoko**: expose `clearExtraFlag(name)` via `invoke('clearExtraFlag', false, [name])` and `resetExtraFlags()` via `invoke('resetExtraFlags', false, [])`
- **vscode-motoko**: call `resetExtraFlags()` then `setExtraFlags(perCanisterFlags)` before each type-check, or use surgical `clearExtraFlag("--enhanced-migration")` when switching canisters

## Test plan

- [x] `clearExtraFlag("--error-detail")` restores detail to 4 after `setExtraFlags(["--error-detail", "0"])`
- [x] `clearExtraFlag("--nonexistent")` throws
- [x] `resetExtraFlags` restores all flags
- [x] Flags don't bleed through `resetExtraFlags` into subsequent `setExtraFlags`
- [x] Verified locally with `dune build` + `node`
- [ ] CI: `nix build .#js.moc` runs `test/test-moc.js` end-to-end

### DIFF
--- a/src/js/common.ml
+++ b/src/js/common.ml
@@ -81,6 +81,19 @@ let js_set_extra_flags flags =
   let tokens = flags |> Js.to_array |> Array.map Js.to_string in
   parse_extra_flags tokens
 
+let js_reset_extra_flags () =
+  Flags.print_warnings := true;
+  Flags.warnings_are_errors := false;
+  Flags.warning_levels := Flags.default_warning_levels;
+  Flags.error_detail := 4;
+  Flags.error_recovery := false;
+  Flags.error_format := Flags.Plain;
+  Flags.ai_errors := false;
+  Flags.all_libs := false;
+  Flags.implicit_package := None;
+  Flags.actors := Flags.RequirePersistentActors;
+  Flags.enhanced_migration := None
+
 let js_set_run_step_limit limit =
   Mo_interpreter.Interpret.step_limit := limit
 
--- a/src/js/common.ml
+++ b/src/js/common.ml
@@ -81,18 +81,32 @@ let js_set_extra_flags flags =
   let tokens = flags |> Js.to_array |> Array.map Js.to_string in
   parse_extra_flags tokens
 
+let js_clear_extra_flag name =
+  match name with
+  | "--hide-warnings" -> Flags.print_warnings := true
+  | "-Werror" -> Flags.warnings_are_errors := false
+  | "-A" | "-W" | "-E" -> Flags.warning_levels := Flags.default_warning_levels
+  | "--error-detail" -> Flags.error_detail := 4
+  | "--error-recovery" -> Flags.error_recovery := false
+  | "--error-format" -> Flags.error_format := Flags.Plain
+  | "--ai-errors" -> Flags.ai_errors := false
+  | "--all-libs" -> Flags.all_libs := false
+  | "--implicit-package" -> Flags.implicit_package := None
+  | "--enhanced-migration" -> Flags.enhanced_migration := None
+  | "--default-persistent-actors"
+  | "--require-persistent-actors"
+  | "--legacy-actors" -> Flags.actors := Flags.RequirePersistentActors
+  | _ -> failwith (Printf.sprintf "clearExtraFlag: unknown flag %s" name)
+
+let all_extra_flag_names = [
+  "--hide-warnings"; "-Werror"; "-W";
+  "--error-detail"; "--error-recovery"; "--error-format";
+  "--ai-errors"; "--all-libs"; "--implicit-package";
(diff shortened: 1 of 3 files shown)


src/js/common.ml
```diff
@@ -81,6 +81,19 @@ let js_set_extra_flags flags =
   let tokens = flags |> Js.to_array |> Array.map Js.to_string in
   parse_extra_flags tokens
 
+let js_reset_extra_flags () =
+  Flags.print_warnings := true;
```
ggreif, line 85:
```suggestion
  let open Flags in
  print_warnings := true;
```
and the rest too.

src/js/moc_js.ml
```diff
@@ -17,6 +17,7 @@ let () =
       val version = js_version
       method setProjectRoot path = Flags.js_project_root := Some (Js.to_string path)
       method setExtraFlags argv = js_set_extra_flags argv
+      method resetExtraFlags () = js_reset_extra_flags ()
```
ggreif, line 20: I wanted to say "eta-reduce these", but then I saw `method` and thought WTF is this? New fangled OO?

test/test-moc.js
```diff
@@ -288,3 +288,35 @@ assert.throws(
   () => Motoko.setExtraFlags(["-W=MMM"]),
   /moc: invalid warning code: MMM/
 );
+
+// js_of_ocaml note: `method m () = ...` compiles to a 2-arity function
+// (this + unit). Calling from JS requires a dummy argument to avoid partial
+// application. This matches how node-motoko's invoke() calls these methods.
+const resetFlags = () => Motoko.resetExtraFlags(0);
```
ggreif, line 295: @Claudio fixed an environment-bleeding issue back in the day by runtime instantiating a dynamic module, getting all the default settings fresh. Maybe that trick could work here too and would get rid of all those fragile assignments...

ggreif (approved):
I think this is LGTM, but we should consult @crusso when he is back. @claude do you have a better idea?

test/test-moc.js
ggreif: I think it was an improvement for the motoko-playground, maybe two years ago, but probably for `moc.js`.

Kamirus: you mean that instead of resetting the flags, we could spin up another `moc.js`? what might break the cache... I'm not sure, but seems riskier

src/js/moc_js.ml
Kamirus: that;s `js` stuff

Kamirus:
Closing for now, trying to handle vscode without this

PR #6026, opened 2026-04-16
feat: expose `LibTomMath` modular arithmetic as primitives

ggreif:
## Summary

Expose LibTomMath modular arithmetic operations as Motoko primitives:

- `Prim.intAddMod(a, b, m)` — `(a + b) mod m` via `mp_addmod`
- `Prim.intSubMod(a, b, m)` — `(a - b) mod m` via `mp_submod`
- `Prim.intMulMod(a, b, m)` — `(a * b) mod m` via `mp_mulmod`
- `Prim.intPowMod(base, exp, m)` — `base^exp mod m` via `mp_exptmod`

All operate on `Int` (arbitrary precision). `intPowMod` is particularly useful for cryptographic applications (RSA, Diffie-Hellman, etc.) where modular exponentiation with large integers is needed.

### Changes

| Layer | File | What |
|-------|------|------|
| RTS/C | `rts/Makefile` | Add libtommath source files + bindgen allowlist |
| RTS/Rust | `rts/motoko-rts/src/bigint.rs` | Wrapper functions calling mp_* |
| Compiler | `compile_enhanced.ml`, `compile_classical.ml` | Wasm imports + OtherPrim codegen |
| Interpreter | `mo_values/prim.ml` | OCaml reference implementation |
| Prelude | `src/prelude/prim.mo` | Public Prim declarations |
| Test | `test/run/modular-arith.mo` | Happy-path tests incl. Fermat's little theorem |

## Test plan

- [x] `nix build .#tests.unit` passes
- [x] `nix build .#tests.release` passes
- [ ] CI green

## TODO
- [ ] Amend `Changelog.md`


ggreif:
cc @timohanke — this was your request 🎯

rts/Makefile
```diff
@@ -20,7 +20,13 @@ TOMMATHFILES = \
...
    s_mp_balance_mul s_mp_toom_mul s_mp_toom_sqr s_mp_karatsuba_sqr s_mp_sqr_fast s_mp_sqr s_mp_karatsuba_mul \
    s_mp_mul_digs_fast s_mp_mul_digs mp_init_multi mp_clear_multi mp_mul_2 mp_div_2 mp_div_3 mp_lshd mp_incr \
-   mp_decr mp_add_d mp_sub_d
+   mp_decr mp_add_d mp_sub_d \
+   mp_mod mp_addmod mp_submod mp_mulmod mp_exptmod mp_invmod \
+   mp_reduce mp_reduce_setup mp_montgomery_setup mp_montgomery_calc_normalization mp_montgomery_reduce \
+   mp_dr_is_modulus mp_dr_setup mp_dr_reduce \
+   mp_reduce_is_2k mp_reduce_2k_setup mp_reduce_2k \
+   mp_reduce_is_2k_l mp_reduce_2k_setup_l mp_reduce_2k_l \
+   mp_cnt_lsb s_mp_exptmod s_mp_exptmod_fast s_mp_montgomery_reduce_fast
```
ggreif:
```suggestion
   mp_cnt_lsb s_mp_exptmod \
   s_mp_exptmod_fast s_mp_montgomery_reduce_fast
```

ggreif:
@timohanke — your modular arithmetic primitives are ready! 🎉

`Prim.intAddMod`, `intSubMod`, `intMulMod`, `intPowMod` backed by LibTomMath's `mp_addmod`, `mp_submod`, `mp_mulmod`, `mp_exptmod`. Build artifacts attached to the CI run.

Getting here was a bit of a Grothendieck experience — rather than cracking the nut with a hammer, we let the rising sea of understanding slowly submerge the problem. From missing libtommath transitive dependencies, through WASI allocator incompatibilities, to discovering that Motoko's compact BigNum representation requires proper unboxing for ternary operations — each wrong theory peeled back a layer until the fix emerged naturally from the architecture.

Build artifacts (macOS, Ubuntu, ARM) are available from the CI run linked above.

E.g. macOS ARM64 is here: https://github.com/caffeinelabs/motoko/actions/runs/24552026361/artifacts/6490288624

timohanke:
Thanks @ggreif . This is awesome. Looking forward to try it out.

We should also consider exposing `mp_sqr`. And additionally, make the compiler detect `x ** 2` or `x * x` in the source code and call it. (Off-topic for the feature, which is about modular arithmetic and `mp_sqr` isn't modular, but on-topic application-wise because it is useful for the same cryptographic applications.) 

Further down the road we can also consider exposing
```
mp_montgomery_setup
mp_montgomery_calc_normalization
mp_montgomery_reduce
```
but let's see first how the ones already added perform in practice.

ggreif:
> We should also consider exposing `mp_sqr`. And additionally, make the compiler detect `x ** 2` or `x * x` in the source code and call it. (Off-topic for the feature, which is about modular arithmetic and `mp_sqr` isn't modular, but on-topic application-wise because it is useful for the same cryptographic applications.)
> 

Good point. We could do
We sure could:

- spot both args `VarE` and same -> call `mp_sqr`
- spot literal `**` exponent `2` and call `mp_sqr`
- popcnt exponent and if =1 do iterated `mp_sqr`
Both of the latter are possible, but does TomMath already do the second? Still it would be a win, because `** 2` is internally `0b100` and `ctz` gives 2 then iterate `mp_sqr` only once. For `0b1000` (a.k.a. `4`) iterate twice. Etc. Maybe worth it. When heap pointer `popcnt = 1` will never trigger.

timohanke:
Also `mp_invmod` should be exposed as well. That fits right into this PR.

timohanke:
After testing this in some benchmarks I would like to see this function exposed as well: `mp_invmod`.

The benchmarks showed an advantage of `intPowMod = mp_exptmod` over an implementation in user code. That's why I suspect that access to `mp_invmod` would be an advantage over implementing EEA in user code.

The benchmarks showed no advantage of `intAddMod, intSubMod, intMulMod` over using the operators `+, -, *, %` in user code.

ggreif:
> Also `mp_invmod` should be exposed as well. That fits right into this PR.

Go for it! Thanks for the feedback!

timohanke:
Great, thank you. `intInvMod` is indeed faster than user code EEA. By how much differs widely depending on input. Hence it is more meaningful to benchmark in actual applications. 

In raw EC arithmetic the results still differ, between 5-20% improvement. In the full final application of ECDSA-verify the improvement shrinks down to 1%.

I then tried using Montgomery form throughout ECDSA-verify and the result was a 20% improvement. That was by implementing everything in use code. So my next request would be to expose these four from libtommath:

- mp_montgomery_setup
- mp_montgomery_calc_normalization
- mp_montgomery_reduce
- mp_sqr

Then maybe we can speed up Montgomery form even further.

ggreif:
@timohanke Montgomery suite + `intSqr` landed in 40e804d0f, golden updates in 441e0f63f. Five new primitives, all `Int`-typed:

- `intSqr(a)` → `mp_sqr` (specialised squaring)
- `intMontgomerySetup(m)` → `mp_montgomery_setup` (the `rho` constant)
- `intMontgomeryCalcNormalization(m)` → `mp_montgomery_calc_normalization` (the `R` value)
- `intMontgomeryReduce(t, m, mp)` → `mp_montgomery_reduce` (`t · R⁻¹ mod m`; non-mutating wrapper)

Plumbed through both backends (classical + enhanced) plus an OCaml interpreter implementation that's round-trip-consistent (`reduce(a · R mod m, m, mp) ≡ a mod m` holds for any `digit_bit`, so the test passes in both interpreter and compiled even though wasm32 uses 28-bit digits and wasm64 uses 60-bit). Test coverage in [`test/run/modular-arith.mo`](https://github.com/dfinity/motoko/blob/gabor/modular-arithmetic/test/run/modular-arith.mo) — single-reduce round-trip, Montgomery multiplication (`reduce(aR · bR, m, mp) ≡ abR mod m`), and double-reduce. All seven phases green.

If you have a chance to re-run the ECDSA-verify benchmark on top of this branch, very curious to hear the numbers — the prior Motoko-side Montgomery you measured (20% improvement) should now be comparable against C-direct-loop reductions on the inner ops.

timohanke:
On it, trying it out. But I already observed:

> so the test passes in both interpreter and compiled

This is not the case for me in the interpreter. Tests only pass in wasi mode for me.

ggreif:
Interesting — locally `make -C test/run modular-arith.only` reports `[tc] [run] [run-ir] [run-low] [comp] [valid] [wasm-run]: All tests passed.` so the OCaml interpreter (`run`/`run-ir`/`run-low` phases) exercises [`prim.ml`](https://github.com/dfinity/motoko/blob/gabor/modular-arithmetic/src/mo_values/prim.ml#L246-L319)'s digit_bit=28 implementations cleanly on the test cases I added.

What's the specific failure on your side — failing assertion, the exact input, the produced-vs-expected values? Two early hypotheses:

1. **Stale `moc`**: did the new `prim.ml` cases get compiled into your local binary? `make -C src` after a fresh `git pull` on `gabor/modular-arithmetic` should pick them up.
2. **A test case my coverage doesn't reach** — I focused on round-trips because they cancel the digit_bit choice. If your benchmark uses `intMontgomerySetup` / `intMontgomeryCalcNormalization` and asserts on the raw `rho` / `R` values directly, the interpreter (digit_bit=28) and wasm32 compiled (digit_bit=28) agree but wasm64 compiled (digit_bit=60) doesn't — that'd reproduce as "interpreter ≠ compiled" symmetric to "interpreter passes but `comp` fails", though.

If you can paste the failing line + the value mismatch I can ground the next iteration on data rather than guesswork.

timohanke:
Seems to pass now in the interpreter as well. Not totally sure but it might have been related to the constants.

Benchmark for ecdsa_verify improved significantly against the Motoko-side implementation. Slightly more than 2x improvement. (That is on top of the 20% improvement of using Montgomery form at all.)

Same for GC pressure, slightly above 2x improvement.

timohanke:
`Prim.invMulMod` now gives a larger improvement percentage-wise after we switched to built-in Montgomery than it did before. Probably because the baseline has dropped in absolute terms the same difference in absolute terms now becomes more noticeable percentage-wise.

So I would keep these four functions in the compiler:

invMulMod
intMontgomerySetup
intMontgomeryCalcNormalization
intMontgomeryReduce

I didn't get anything out of intSqr (using it made everything slower).

PR #6028, opened 2026-04-17
chore: bump `OCaml` 4.14 → 5.3

ggreif:
## Summary

Upgrade the OCaml compiler from 4.14 to 5.3.

### Changes

- **`nix/pkgs.nix`**: `ocamlPackages_4_14` → `ocamlPackages_5_3`
- **`src/wasm-exts/customModuleEncode.ml`**: Inline the VLQ Base64 encoder (from `ocaml-vlq`, MIT license). The external `vlq` package depends on `Stream` which was removed in OCaml 5.x. Only the encode path is needed by `moc` (for source maps); decode is not used.
- **`src/wasm-exts/dune`**: Drop `vlq` library dependency
- **`flake.nix`**: Remove `vlq` from build inputs

### Status

- [x] `nix build .#moc` — passes
- [x] `nix build .#js.moc` — passes
- [x] Full CI test suite — passes

### Notes

The `ocaml-vlq` package is marked as broken for OCaml 5.x in nixpkgs due to its use of `Stream.t` (removed in OCaml 5.0). Since `moc` only uses `Vlq.Base64.encode` (4 call sites for source map generation), inlining the ~20 lines of encode logic is simpler than maintaining a compatibility shim.

The `ocaml-vlq-src` flake input has been dropped (separate PR #6027) as it was already unused — the package came from nixpkgs directly.

### DIFF
--- a/src/mo_def/syntax.ml
+++ b/src/mo_def/syntax.ml
@@ -427,18 +427,18 @@ let contextual_dot_args e1 e2 dot_note =
   let arity = match dot_note.note.note_typ with
     | T.Func(_, _, _, args, _) -> List.length args
     | _ -> raise (Invalid_argument "non-function type in contextual dot note") in
-  let effect eff =
+  let effec eff =
     match (e1.note.note_eff, eff) with
     | T.Triv, T.Triv -> T.Triv
     | _, _ -> T.Await
   in
   let args = match e2 with
     | { it = TupE []; at; note = { note_eff;_ } } ->
-       { it = e1.it; at; note = { note_eff = effect note_eff; note_typ = e1.note.note_typ } }
+       { it = e1.it; at; note = { note_eff = effec note_eff; note_typ = e1.note.note_typ } }
     | { it = TupE exps; at; note = { note_eff; note_typ = T.Tup ts } } when arity <> 2 ->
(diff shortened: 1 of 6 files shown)


src/mo_def/syntax.ml
```diff
@@ -427,18 +427,18 @@ let contextual_dot_args e1 e2 dot_note =
   let arity = match dot_note.note.note_typ with
     | T.Func(_, _, _, args, _) -> List.length args
     | _ -> raise (Invalid_argument "non-function type in contextual dot note") in
-  let effect eff =
+  let effec eff =
```
alexandru-uta, line 430: I'm probably missing something, but why the rename?

christoph-dfinity: OCaml 5 adds effect handlers, so `effect` became a keyword.

ggreif: That's exactly what you are not missing any more ;-)

christoph-dfinity:
Anything we can change from our custom `Lib.*` modules to use new functionality in the std lib? Or rather in a follow-up PR?

crusso (commented):
Nice that it was that easy to move from 4 to 5!

crusso (approved):

ggreif:
> Anything we can change from our custom `Lib.*` modules to use new functionality in the std lib? Or rather in a follow-up PR?

That will be a follow-up soon, `take` and `drop` are now in. See #6034.

PR #6034, opened 2026-04-20
chore: drop `Lib.List.take/drop` — now in OCaml 5.3 `stdlib`

ggreif:
## Summary

Follow-up to #6028 (OCaml 4.14 → 5.3 bump). Five isolated commits, each removing a `Lib.List*` function that is now redundant with the stdlib (or dead code). Easy to review one-by-one.

### 1. `drop Lib.List.take/drop` — now in OCaml 5.3 stdlib

OCaml 5.3 adds `List.take` / `List.drop` to the stdlib. Three call sites in the tree retargeted; `Lib.List.split_at` now delegates to the stdlib versions.

**Semantics note.** The custom versions raised `Failure` on short list; stdlib returns the short result. All three call sites guard `n ∈ [0, length xs]`, so the difference is not observable:

- `src/mo_idl/idl_to_mo_value.ml:93` — `drop (length vs.it) ts` guarded by `compare_lengths vs.it ts < 0`.
- `src/mo_types/type.ml:2018-2019` — `take (max/2)` / `drop (len − max/2)` guarded by `len >= max`.

### 2. `drop Lib.List.equal` — in stdlib since 4.12

No callers in tree. Stdlib's `List.equal eq l1 l2` returns `false` on length mismatch, matching the custom `try List.for_all2 … with _ -> false`.

### 3. `drop Lib.List.compare` — in stdlib since 4.12

Stdlib's `List.compare` has identical lexicographic semantics (shorter list < longer). One caller in `src/mo_values/value.ml` updated.

### 4. `drop dead Lib.List32.take / Lib.List32.drop`

No callers anywhere in the source tree. The `int32`-indexed variants are not provided by stdlib (so this isn't a stdlib-alignment change like its int-variant siblings), just dead-code cleanup while the surrounding module is being touched.

### 5. `drop Lib.List.safe_map` — stdlib `List.map` is TMC-tail-recursive since OCaml 5.1

Our `safe_map` was a hand-rolled `[@tail_mod_cons]` variant of `List.map` to avoid stack overflow on long lists. OCaml 5.1 reimplemented `Stdlib.List.map` using the same attribute, so the wrapper is now dead weight. All ~30 call sites (most of them in `src/linking/linkModule.ml`, a few in `src/wasm-exts/ast.ml`) replaced with `List.map`. Per @crusso's review.

## Test plan

- [x] `nix develop -c make -C src moc` — builds clean after each commit.
- [x] Broader test execution deferred to CI.

### DIFF
--- a/src/lib/lib.ml
+++ b/src/lib/lib.ml
@@ -443,13 +443,6 @@ struct
        | [] -> false
        | hd' :: tl' -> equal hd hd' && is_prefix equal tl tl')
 
-  (* tail-recursive map *)
-  let[@tail_mod_cons] rec safe_map f l = match l with
-    | [] -> []
-    | x :: xs ->
-      f x :: (safe_map[@tailcall]) f xs
-    [@@coverage off]
-
   let align cmp xs ys =
     let next (xs, ys) = match xs, ys with
       | [], [] -> None
--- a/src/lib/lib.ml
+++ b/src/lib/lib.ml
@@ -415,16 +415,6 @@ struct
 
   let index_of x = index_where ((=) x)
 
-  let rec compare f xs ys =
-    match xs, ys with
-    | [], [] -> 0
-    | [], _ -> -1
-    | _, [] -> +1
-    | x::xs', y::ys' ->
-      match f x y with
-      | 0 -> compare f xs' ys'
-      | n -> n
-
   let rec is_ordered f xs =
     match xs with
     | [] | [_] -> true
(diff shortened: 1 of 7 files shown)


crusso (approved):
I think you can replace safe_map by map now too, if you like

ggreif:
Thanks @crusso — good call. Pushed as `4b91461d8`. Dropped `Lib.List.safe_map` and replaced ~30 call sites (mostly in `linkModule.ml`, a few in `wasm-exts/ast.ml`) with `List.map`. The stdlib version has been TMC-tail-recursive since OCaml 5.1, so the wrapper was carrying its weight only under 4.14 — a debt that became payable on the 5.3 bump.

PR description updated with a new commit-5 entry.

PR #6044, opened 2026-04-22
fix: improve error message for install over non-Motoko canister

alexandru-uta:
This PR improves the error message for installing EOP canister on top of non-Motoko canister.

Before, in Caffeine, due to erroneous installs, we were getting messages like:
```
Deployment failed: Failed to install canister XYZ due to Failed to install canister XYZ with version 1.0: Failed to call management canister method 'install_code' for 'XYZ': The replica returned a rejection error: reject code CanisterError, reject message Error from Canister XYZ: Canister called `ic0.trap` with message: 'Detected implicit upgrade from classical orthogonal persistence to enhanced orthogonal persistence. Recompile with explicit flag --enhanced-orthogonal-persistence and redeploy to enable this irreversible migration.'.
```

This was misleading and did not help for identifying the root cause of the issue.

Closes LANG-1271.


crusso (approved):
LGTM

PR #6046, opened 2026-04-22
test-runner: auto-detect persistence mode per test

alexandru-uta:
## Summary
- `test-runner` now infers applicable persistence mode(s) from markers in each test file; no more manual `EXTRA_MOC_ARGS=--enhanced-orthogonal-persistence`.
- Unmarked tests run in both classical and EOP against the shared `ok/*.ok`. Marker/flag conventions match those already honored by `test/run.sh`.
- In multi-mode `-a` accept, only the first mode accepts; later modes diff against the fresh golden to catch cross-mode divergence.
- Legacy `EXTRA_MOC_ARGS` env path preserved for CI/Nix (inference skipped, single inherited invocation).
- Zero changes to `test/run.sh`.

## Test plan
- [x] `cargo test --bin test-runner mode::` (13 tests)
- [x] Smoke: `test-runner -bf bad-field-dec` inside nix, runs both modes, passes
- [x] Smoke: `test-runner -baf bad-field-dec` inside nix, accept-then-verify, passes

closes LANG-1273

### DIFF
--- a/.agents/skills/build-and-test/SKILL.md
+++ b/.agents/skills/build-and-test/SKILL.md
@@ -103,13 +103,24 @@ relevant to your change:
 test-runner -baf some-test         # targeted by name pattern
 ```
 
-**EOP (Enhanced Orthogonal Persistence) tests**: By default, `run-drun/` tests
-run in non-EOP mode. To run the EOP variant, prefix the command with
-`EXTRA_MOC_ARGS="--enhanced-orthogonal-persistence"`:
+**EOP (Enhanced Orthogonal Persistence) tests**: `test-runner` detects the
+persistence mode automatically from markers in the test file — you don't
+need to set any environment variable. The rules (mirrored from `test/run.sh`):
 
-```bash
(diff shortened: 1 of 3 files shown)


.agents/skills/build-and-test/SKILL.md
```diff
@@ -103,13 +103,24 @@ relevant to your change:
...
 ```
 
-**EOP (Enhanced Orthogonal Persistence) tests**: By default, `run-drun/` tests
-run in non-EOP mode. To run the EOP variant, prefix the command with
-`EXTRA_MOC_ARGS="--enhanced-orthogonal-persistence"`:
-
-```bash
-EXTRA_MOC_ARGS="--enhanced-orthogonal-persistence" test-runner -baf run-drun/some-test
-```
+**EOP (Enhanced Orthogonal Persistence) tests**: `test-runner` detects the
```
Kamirus: maybe this whole section could be removed now? agents running the test runner needs to know about this? it should be now transparent right?

alexandru-uta: sure, makes sense; updated.

Kamirus (approved):

PR #6049, opened 2026-04-23
feat: `and`-patterns (conjunctive patterns)

ggreif:
Adds `p1 and p2` — conjunctive pattern, dual to existing `p1 or p2`. Matches when both legs match; bindings from both are available. Legs must bind *disjoint* identifiers (or-patterns require the *same* set); `and` binds tighter than `or`.

```motoko
switch o {
  case (?x and s) (debug_show {s; x});  // payload + whole value
  case null "none";
};
```

## Implementation

End-to-end: parser → Syntax → typing/definedness/coverage → desugar → IR → all ir_passes (rename, subst_var, freevars, check_ir, await, async, const, tailcall, erase_typ_field) → both backends (classical + enhanced) → wasm. AST and IR interpreters both use `Option.(bind … map …)`.

## Tests

Positive (`test/run/`):

- `and-pattern-binds.mo` — passes `[tc] [run] [run-ir] [run-low] [comp] [valid] [wasm-run]`. Refutable-leg failures (both positions), func-body, TupP-nested, three-way with refutable middle.
- `and-pattern-coverage.mo` — exhaustiveness regression: `InAnd1` walks both legs cleanly.
- `and-pattern-infer.mo` — inference position with ≥1 explicit leg (typer's `true, _` / `_, true` arms).

Negative (`test/fail/`):

- `and-pattern-typemismatch.mo` — let-context, check-side type mismatch (M0117 from per-leg check).
- `and-pattern-infer-typemismatch.mo` — inference-position, both explicit annotations disagree (M0262).
- `and-pattern-infer.mo` — inference-position, neither leg explicit (M0261).
- `and-pattern-overlap.mo` — let-context duplicate binding (M0260, frontend).
- `and-pattern-overlap-case.mo` — case-context duplicate (M0260, from the check_pat path).
- `and-pattern-type-overlap.mo` — duplicate type-identifier in legs (M0260, typ-id arm).

## Docs

Changelog entry + language-manual section + fundamentals-table row.

See also the [independent FP-savvy pre-review pass](https://github.com/caffeinelabs/motoko/pull/6049#issuecomment-4303457170) for concrete correctness/coverage traces.


ggreif:
## Independent FP-savvy review (pre-human-review pass)

Ran an independent pass over the branch looking for correctness bugs, style issues, and coverage gaps. Summary below.

### Overall verdict

- ✅ **Frontend plumbing (syntax, parser, definedness, desugar, arrange)** — complete and parallel to AltP. Precedence correct: `%left OR` at `parser.mly:268` then `%left AND` at `parser.mly:269` — later-declared = tighter, so `and` binds tighter than `or`.
- ✅ **Typing** (`src/mo_frontend/typing.ml`) — `check_pat`, `gather_pat_aux`, `vis_pat`, `combine_pat_srcs`, `is_explicit_pat`, `check_pat_typ_dec` all have AndP arms. Overlap rejection via M0189 on both val and typ envs.
- ✅ **AST interpreter** (`src/mo_interpreter/interpret.ml:929`) — monadic form `bind ... (fun ve1 -> map (V.Env.adjoin ve1) ...)` is idiomatic.
- ✅ **IR interpreter** (`src/ir_interpreter/interpret_ir.ml:762`) — symmetric; `declare_pat` adjoins both legs.
- ✅ **IR check** (`src/ir_def/check_ir.ml:1093`) — `t <: pat1.note; t <: pat2.note` + `disjoint_union` is correct and enforces the invariant IR-side.
- ✅ **Freevars / rename / subst_var** — `+-` is left-biased union on maps; on disjoint keys it coincides with true union, so the result is correct. Rename threads `rho → rho' → rho''` via `id_bind` which allocates globally-fresh names, so no fresh-rename collision possible.
- ✅ **await.ml** — `disjoint_union` in `rename_pat`; `declare_pat pat1 (declare_pat pat2 exp)` and `define_pat` concat are correct. AltP's freevars-empty assertion correctly does not extend to AndP.
- ✅ **Codegen** — `(^^^)` at `compile_enhanced.ml:10635` threads the *same* fail continuation to both legs' `CanFail` arms, so if either leg fails the whole match aborts. Stack discipline is balanced (`set_i` consumes scrutinee; each leg starts with `get_i` and consumes it). Classical and enhanced are identical.
- ⚠️ **Coverage** (`src/mo_frontend/coverage.ml:272`) — explicit TODO stub delegating to `pat1` only. Concrete consequences: (a) `case (#a _ and #b _)` is treated as covering `#a _`, so an explicitly-unreachable `case` won't be flagged redundant and a following `case (#a _)` may be spuriously flagged as redundant; (b) exhaustiveness-wise it's *over*-approximating coverage of the first leg, so "missing cases" warnings may be missed. **Diagnostic-quality regression only, not soundness.** — **Fixed** with 7a757525d8e0dde8ee387a12e1c570adad1bfbe8!

### Correctness concerns

1. `gather_pat` in `check_ir.ml:1040` uses `go (go ve pat1) pat2` without any collision check, trusting the frontend. `check_pat` further down does enforce `disjoint_union`, so any bug that slipped past the frontend would surface there — **defensible but worth a one-line assertion or comment noting the redundancy**. — **Fixed**: the "(frontend enforces)" comments at `check_ir.ml:1041` and `check_ir.ml:1094` have been rewritten to name the actual enforcer (`check_pat` / `verify_pair`), so the invariant-holder is explicit for separate-compilation scenarios where IR is loaded from disk. `check_pat`'s AndP arm now uses an explicit `verify_pair` helper (cheap iter + predicate, names the first clashing key) instead of `disjoint_union`'s raise/catch control flow.
2. `gather_pat_aux` in `typing.ml:4622` also quietly merges without collision check; the real enforcement happens in `check_pat` at 3444. Same comment applies: gather runs before the overlap check. — **Fixed** in `753404cac`: `gather_pat_aux`'s AndP arm now gathers each leg independently against the outer scope and enforces disjoint leg-contributions with the AndP-specific M0260, matching `check_pat`'s diagnostic (previously `gather_id` fired the generic M0051 first in let-context).
3. `infer_pat` at 3271 outright errors with M0184. Even when one leg is fully annotated (`(a : Nat) and b`) we could propagate its type to the other leg; currently this forces users to annotate both. Parity with AltP's behaviour, but surprising in function-arg position: `func f ((a : Nat) and b) = …` works; `func f (a and b) = …` is a hard error. OK to leave; document it. — **Partially fixed**: now uses dedicated `M0261` (not M0184) in `1b332d74a`; parity with AltP's `M0184` preserved, so left undocumented.

### Missing tests (the biggest gap)

The current positive test (`and-pattern-binds.mo`) exercises *only* irrefutable patterns in `let`-position, so the interesting codegen path (`CanFail is1 ... k; is2 k` with actual failure) is untested. Suggested additions:

- **Refutable AndP in switch**: `case (?x and s) …; case null …` over a `?Nat` — exercises `(^^^)` on `CanFail` codes. — **Fixed** in `bfe6a06aa`.
- **Runtime-failing leg**: switch where the AndP fails on a value that would have matched the other case — verifies fail-through from leg 1 or leg 2 reaches the next `case`, not a trap. — **Fixed** in `bfe6a06aa` (`leftFails` / `rightFails` over `Status`).
- **AndP in function arg**: `func f ((x : Nat) and y) = x + y`. — **Fixed** in `1355414ea` (`test/fail/and-pattern-infer.mo` inference-position; positive func-arg covered in `and-pattern-binds.mo`).
- **AndP inside TupP**: `let ((a : Nat) and b, c) = (5, 6)`. — **Fixed** in `bfe6a06aa`.
- **Type-mismatched legs**: `let (x : Nat) and (y : Text) = 5` — expect subtype error, not a confused message. — **Fixed** in `3e730ff7c` (`test/fail/and-pattern-typemismatch.mo`): produces a clean M0117 "pattern of type Text cannot consume expected type Nat" pointing at the offending leg.
- **Three-way conjunction with one refutable leg** (e.g. `((x : Nat) and ?y and z)` over `?7`) — exercises the threaded `rho` in rename/subst. — **Fixed** in `bfe6a06aa`.

### Style nits

- Both interpreters' monadic form is fine.
- `subst_var`'s `pat (pat rho p1) p2` threading and `rename`'s `(p1', rho'); (p2', rho'')` are both idiomatic.

### Final verdict

**Ready for human review, not ready to merge.** Correctness of the core implementation is solid. Before merge I'd want:

1. At least one refutable-AndP switch test exercising `(^^^)` on `CanFail` codes. — **Fixed**
2. At least one AndP-in-func-arg and one AndP-in-TupP test. — **Fixed**
3. Either flesh out `coverage.ml`'s AndP arm or add a FileCheck/warn-expectation test confirming the current stub's behaviour is known. — **Fixed**
4. The typing changes include a new use of error code `M0184` for AndP; M0189 for overlap — both reuse existing codes, which is fine, but worth a quick grep that no diagnostics index needs updating. — **Fixed**

🤖 Review performed by an independent agent (not the author).

ggreif:
## Round 2 Independent FP-savvy re-review

**Referencing:** [round 1](https://github.com/caffeinelabs/motoko/pull/6049#issuecomment-4303457170). Scope: the two new commits `753404ca` (check_ir rigour + M0260 let-parity) and `3e730ff7` (type-mismatch test), plus anything round 1 missed.

### Overall verdict

- The two round-1 concerns that remained open are now cleanly resolved. `verify_pair` in `check_ir.ml:155` is a faithful stand-in for `disjoint_union`'s raise/catch — same domain, same diagnostic, no behavioural drift. The new `gather_pat_aux` AndP arm at `typing.ml:4648` genuinely restores M0260 parity between `let` and `case` context, and the regenerated `and-pattern-overlap.tc{,-human}.ok` proves it (region is now the whole AndP `1.5-1.28`, code flipped `M0051 → M0260`).
- `T.Env.adjoin` is right-biased (`lang_utils/env.ml:35`, `Some x2`). Since `verify_pair` proves disjointness *before* `adjoin` runs, the semantics are identical to the old `disjoint_union`; if a future regression let overlap through, `adjoin` would silently drop `ve1`'s binding, but the explicit `verify_pair` step prevents that. Fine.
- **Scope-component coverage in the new `gather_pat_aux` arm is complete.** `gather_pat_aux` + `gather_pat_field` + `gather_typ_id` only ever touch `val_env`, `typ_env`, and `con_env`. The first two are explicitly checked. `con_env` uses `Cons.fresh` per `TypPF`, so the constructors are always distinct identities; collisions are detected by-name through the `typ_env` iter before `con_env`'s `disjoint_add` is asked to swallow them. `lib_env`, `obj_env`, `mixin_env`, `fld_src_env` are unreachable from patterns — no gap. Verified against `scope.ml:43` and `typing.ml:4669–4691`.
- Nested AndP (`(a and b) and c`), AndP-in-TupP (`( x and y, z )`), AndP-in-AltP-leg, AndP-with-ObjP-leg-containing-TypPF, and duplicate-against-outer-scope (caught earlier by `gather_id`'s M0051) all work as intended — I traced each.
- Parser precedence `%left COLON` < `%left OR` < `%left AND` (`parser.mly:267/269/270`) is sound: `(a : Nat) and b` and `?x and s` and `#A x and #B y` all bind the way round 1 asserted. Nothing to add.

### Concerns

1. **Stale comment in `test/run/and-pattern-binds.mo:62–63`** — still says "infer_pat raises M0184, matching AltP's behaviour", but the round-1 fix in `1b332d74a` moved the code to the dedicated **M0261**. Trivial, one-line fix.
2. **`verify_pair` is slight over-abstraction.** Its only call site (`check_ir.ml:1108`) passes `fun k ve2 -> not (T.Env.mem k ve2)`. A narrower `check_disjoint env at fmt env1 env2` would be equally rigorous and shorter. Not a blocker; ignore if you plan to reuse the predicate form elsewhere.
3. **New M0117 diagnostic for `let (x : Nat) and (y : Text) = 5` does not mention `and`-pattern context.** The output reads "pattern of type Text cannot consume expected type Nat" and underlines `(y : Text)`. The underline is helpful, but a first-time AndP user sees a bare AnnotP error and has to reason about *why* that leg saw `Nat`. This is shared with AltP's equivalent, and not a new issue introduced by these commits — just worth noting that the test locks in a slightly user-unfriendly diagnostic. I would not block merge on it.

### Things round 1 skipped that I re-checked and cleared

- **Coverage residual over-approximation.** `coverage.ml:272, 345, 374` (`InAnd1`): traced slices for `((#a _) and (#b _))` — pat1 accepts `#a` slice, `succeed(InAnd1)` delegates to pat2 on `#a`-narrowed desc, pat2's TagP fails → `fail(InAnd1) → fail ctxt'` → fall-through. Case's `at` is not added to `reached_cases`, so a vacuous AndP case *is* correctly flagged as unreachable. No residual over-approximation.
- **Freevars** (`freevars.ml:88–89`): `+-` on disjoint keys = union; combined with `check_pat`/`verify_pair`'s disjointness guarantee, the asymmetric left-biased semantics of `+-` are irrelevant. Clean.
- **`await.ml`** `rename_pat`/`declare_pat`/`define_pat` all still symmetric between AndP legs. No AltP-style "bindings-empty" assumption leaks into AndP.
- **`arrange.ml:172` / `arrange_ir.ml:143`** — identical to AltP, fine.

### Final verdict

**Ready to merge** once the stale M0184 comment in `and-pattern-binds.mo:62–63` is updated. Everything else is either cosmetic (over-general helper) or shared with AltP's long-standing behaviour (AndP-flavour-less M0117). The rigour/parity story is now coherent end-to-end: frontend gathers strictly, check_pat enforces strictly, and `check_ir`'s `verify_pair` validates disk-loaded IR without trusting the frontend.

🤖 Round-2 review by an independent agent (not the author).

src/ir_def/rename.ml
```diff
@@ -144,6 +144,10 @@ and pat' rho = function
     end;
     let rho' = ids_bind rho is1 in
     (AltP (pat_subst rho' p1, pat_subst rho' p2), rho')
+  | AndP (p1, p2) ->
+    let (p1', rho')  = pat rho p1 in
+    let (p2', rho'') = pat rho' p2 in
```
ggreif:
```suggestion
    let p1', rho'  = pat rho p1 in
    let p2', rho'' = pat rho' p2 in
```

src/ir_passes/await.ml
```diff
@@ -597,6 +598,10 @@ and rename_pat' pat =
     assert(Freevars.(M.is_empty (pat pat1)));
     assert(Freevars.(M.is_empty (pat pat2)));
     (PatEnv.empty,pat.it)
+  | AndP (pat1, pat2) ->
+    let (patenv1, pat1') = rename_pat pat1 in
+    let (patenv2, pat2') = rename_pat pat2 in
```
ggreif:
```suggestion
    let patenv1, pat1' = rename_pat pat1 in
    let patenv2, pat2' = rename_pat pat2 in
```

src/mo_frontend/typing.ml
```diff
@@ -3879,6 +3933,18 @@ and check_pat_typ_dec env t pat : Scope.typ_env =
...
+  | AndP (pat1, pat2), _ ->
+    (* Type-level bindings: union, with error on overlap. *)
+    let te1 = check_pat_typ_dec env t pat1 in
+    let te2 = check_pat_typ_dec env t pat2 in
+    T.Env.iter (fun k _ ->
+      if T.Env.mem k te2 then
+        error env pat.at "M0260"
+          "type identifier `%s` bound in both branches of and-pattern" k
+    ) te1;
+    T.Env.merge (fun _ o1 o2 -> match o1, o2 with
```
ggreif: Tests needed for type-level in `and`-patterns.

ggreif: Tests added!

Changelog.md
```diff
@@ -4,6 +4,8 @@
 
   * feat: Implicit argument derivation — the compiler can derive implicit arguments from functions that themselves have implicit parameters (e.g., `compare` for `[Nat]` from `Array.compare<Nat>` + `Nat.compare`). Works transitively and is depth-limited via `--implicit-derivation-depth` (#5966).
 
+  * feat: and-patterns — `(p1 and p2)` matches when both legs match, binding from both (#6049).
```
```suggestion
  * feat: and-patterns — `p1 and p2` matches when both legs match, binding from both (#6049).
```

src/codegen/compile_enhanced.ml
```diff
@@ -13508,6 +13508,16 @@ and fill_pat env ae pat : patternCode =
...
+  | AndP (p1, p2) ->
+      (* Both legs must match the same scrutinee; bindings from both
+         are available. Stash the scrutinee once, feed it to each leg
+         in sequence. `(^^^)` threads failure through — either leg's
+         failure cancels the whole match. *)
+      let code1, code2 = fill_pat env ae p1, fill_pat env ae p2 in
+      let set_i, get_i = new_local env "and_scrut" in
+      CannotFail set_i ^^^
+      (CannotFail get_i ^^^ code1) ^^^
+      (CannotFail get_i ^^^ code2)
```
christoph-dfinity, line 13516: For code that is the same between `compile` and `compile_enhanced` let's also have the same comments or no comments.

src/codegen/compile_enhanced.ml
```diff
@@ -13783,6 +13793,10 @@ and destruct_const_pat ae pat const : VarEnv.t option = match pat.it with
     let l = destruct_const_pat ae p1 const in
     if l = None then destruct_const_pat ae p2 const
     else l
+  | AndP (p1, p2) ->
+    (* Both legs must accept; fold bindings left-to-right. *)
+    Option.bind (destruct_const_pat ae p1 const) (fun ae' ->
+      destruct_const_pat ae' p2 const)
```
christoph-dfinity, line 13794: Same here, either have the comment in both files, or none.

src/ir_def/check_ir.ml
```diff
@@ -1087,14 +1098,25 @@ and check_pat env pat : val_env =
         error env pat.at "set of bindings differ for alternative pattern";
     let common i1 i2 = { typ = lub env i1.typ i2.typ; loc_known = i1.loc_known && i2.loc_known; const = i1.const && i2.const } in
     T.Env.merge (fun _ -> Lib.Option.map2 common) ve1 ve2
+  | AndP (pat1, pat2) ->
+    (* both legs must accept the scrutinee; verify_pair enforces
+       disjoint bindings without trusting the frontend *)
```
christoph-dfinity: No need for this comment. The entire IR checking is independent of the frontend and runs on code in between rewrites/traversals, right?

src/mo_frontend/coverage.ml
```diff
@@ -268,6 +269,15 @@ let rec match_pat ctxt desc pat t sets =
...
     sets.alts <- AtSet.add pat1.at (AtSet.add pat2.at sets.alts);
     match_pat (InAlt1 (ctxt, pat1.at, pat2, t)) desc pat1 t sets
+  | AndP (pat1, pat2) ->
+    (* AndP accepts v iff both pat1 and pat2 accept v.
+       Match pat1 first under an InAnd1 context; on success, continue
+       with pat2 on the narrowed `desc` under the outer context; on
+       failure, the whole AndP fails for that desc-slice. Non-covering
+       content in pat2 (e.g. a variant tag incompatible with what pat1
+       already admitted) surfaces as uncovered via the normal ctxt
+       flow — no special casing needed. *)
```
christoph-dfinity: None of the other branches have explanations, so this is a bit jarring. It also just describes the code rather than providing any higher level insights.

src/mo_frontend/coverage.ml
```diff
@@ -332,6 +342,9 @@ and succeed ctxt desc sets : bool =
   | InAlt2 (ctxt', at2) ->
     sets.reached_alts <- AtSet.add at2 sets.reached_alts;
     succeed ctxt' desc sets
+  | InAnd1 (ctxt', pat2, t) ->
+    (* pat1 accepted `desc`; now pat2 must also accept it *)
```
christoph-dfinity: Same here

src/mo_frontend/typing.ml
```diff
@@ -1178,6 +1178,7 @@ let rec is_explicit_pat p =
   | TupP ps -> List.for_all is_explicit_pat ps
   | ObjP pfs -> List.for_all is_explicit_pat_field pfs
   | AltP (p1, p2) -> is_explicit_pat p1 && is_explicit_pat p2
+  | AndP (p1, p2) -> is_explicit_pat p1 && is_explicit_pat p2
```
christoph-dfinity: This seems to disagree with your code in 3609. There you declare it's enough for either of these patterns to be explicit.

src/mo_frontend/typing.ml
```diff
@@ -3602,6 +3606,41 @@ and infer_pat' name_types env pat : T.typ * Scope.val_env =
...
+    let t, ve1, ve2 =
+      match is_explicit_pat pat1, is_explicit_pat pat2 with
+      | true, true ->
+        let t1, ve1 = infer_pat false env pat1 in
+        let t2, ve2 = infer_pat false env pat2 in
+        if not (T.compatible t1 t2) then
+          error env pat.at "M0104"
+            "and-pattern legs have incompatible types,\nleft accepts%a\nright accepts%a"
+            display_typ_expand t1
+            display_typ_expand t2;
```
christoph-dfinity: Couldnt this just fall through into a `true, _` case, checking the type of the second pattern against the first inferred one?

Are we repurposing the `M0104` code here? We're somewhere in the 200s range by now, and and-patterns are new in this PR.

src/mo_frontend/typing.ml
```diff
@@ -4911,6 +4972,26 @@ and gather_pat_aux env val_kind scope pat : Scope.t =
...
+      && T.Env.mem k scope2.Scope.val_env then
+        error env pat.at "M0260"
+          "variable `%s` bound in both branches of and-pattern" k)
+      scope1.Scope.val_env;
+    T.Env.iter (fun k _ ->
+      if not (T.Env.mem k scope.Scope.typ_env)
+      && T.Env.mem k scope2.Scope.typ_env then
+        error env pat.at "M0260"
+          "type identifier `%s` bound in both branches of and-pattern" k)
+      scope1.Scope.typ_env;
```
christoph-dfinity: It might be nicer to use the combinators for collecting and reporting multiple errors at once here, rather than bailing at the first one.

christoph-dfinity (changes requested):
In general I think this would be a fine feature to have. Did you consider `as` as the keyword over `and` for familiarity?

The AI generated code is pretty wordy with its comments. Maybe take a pass over the code and try to be consistent with the surrounding code.

src/codegen/compile_enhanced.ml
ggreif: No comments.

src/mo_frontend/typing.ml
ggreif: I prefer to compute the most precise subtype which we get with `glb`.

The other comment is addressed with a fresh code.

src/mo_frontend/typing.ml
ggreif: Fixed. Thanks for spotting!

src/mo_frontend/typing.ml
ggreif: Using `local_error` now, as this mismatch is not a killer.

christoph-dfinity:
I don't think your change in 12d50d2fdefa387b4422e1d91e14199bea541a64 works as you expect it to. That's why I was talking about the combinators in my comment (see `try_both`/`try_all`)

test/fail/ok/and-pattern-overlap-multi.tc-human.ok
```diff
@@ -0,0 +1,12 @@
+error[M0260]: variable `a` bound in both branches of and-pattern
+    ┌─ and-pattern-overlap-multi.mo:8:5
+  8 │  let (a : Nat, b : Nat, c : Nat) and (a : Nat, b : Nat, c : Nat) = (1, 2, 3);
```
christoph-dfinity, line 3: Looks like you can pick more precise error ranges here, rather than targeting the full pattern.

ggreif: Now pointing at the name conflict.

test/fail/ok/and-pattern-type-overlap.tc-human.ok
```diff
@@ -0,0 +1,4 @@
+error[M0260]: type identifier `T` bound in both branches of and-pattern
+    ┌─ and-pattern-type-overlap.mo:1:5
+  1 │  let { type T } and { type T } : module { type T = Nat } =
+    │      ^^^^^^^^^^^^^^^^^^^^^^^^^
```
ggreif: Attempting here too...

ggreif: Worked.

christoph-dfinity (approved):
LGTM

PR #6051, opened 2026-04-23
testing: sunset `run.sh`

alexandru-uta:
This PR retires `test/run.sh` (Bash) in favor of `run-test`, a Rust drop-in built
from the `test-runner` crate and shipped from the same Nix derivation.

- `run-test` matches `run.sh` on flags, positional args, cwd semantics,
  stdout format, exit codes, and `_out/` / `ok/` layout.
- drun / pocket-ic tests run in-process; each persistence mode spawns a
  fresh `run-test` subprocess so `--all-modes` correctly captures
  `debug_print` output from `pocket-ic-server`.
- Output normalization is ported verbatim from `run.sh`'s `sed`/`grep`
  filters; persistence-mode inference moves out of
  `test-runner::run_single_test` and into `run-test`.
- Makefiles, `quick.mk`, and `test-feature.sh` are updated. Rule
  prerequisite `../run.sh` becomes `$(shell command -v run-test)` so
  rebuilds track the binary's Nix-store path.
- `profile-report.sh` absorbs the logic of `drun-wrapper.sh`; both
  `test/run.sh` and `test/drun-wrapper.sh` are deleted.

Closes LANG-1275.


ggreif (approved):
LGTM! I don't think there are in-flight changes that could break because of this. Low risk.

_EDIT_: where did all the `sed` normalisation rules go? I didn't see them, but they must be there!
_FOUND THEM_: https://github.com/caffeinelabs/motoko/pull/6051/changes#diff-1d066b7b9c428a9574ab5f02c7dbf3424bc5cee3622b04e73d7a03266cbb0478

PR #6054, opened 2026-04-24
experiment: implicit-derivation-rep example of type reps failing inference

crusso:
Part of reviewing #5966.

I thought this example (inferring a derived dictionary of functions( would work, but perhaps it's expected not to. 

I guess we only apply inference if the hole has a function type - it can't be a record of methods. But is that restriction necessary?

I updated the review with some comments/questions


Kamirus:
Answered here https://github.com/caffeinelabs/motoko/pull/5966#discussion_r3147562857

PR #6073, opened 2026-04-30
fix: strip motoko:stable-types section from wasm under --enhanced-migration

alexandru-uta:
Strips the `motoko:stable-types` custom section from the wasm whenever `--enhanced-migration` is active. The section grows linearly with the migration chain (~280 KB at length 200, ~1.4 MB at length 1000), duplicates the `.most` file content emitted under `--stable-types`, and under enhanced orthogonal persistence is redundant for upgrade safety because the runtime system already enforces stable-type compatibility at upgrade time. Long migration chains were pushing the wasm past the IC's ~1 MB metadata limits with this section alone.

## Behaviour preserved
- Compile-time `validate_stab_sig` still runs (reads from `stable_types_text`).
- `--stable-types` still writes `.most` next to the wasm (reads from `stable_types_text`).
- The JS API (`js_compile_wasm`) still returns the stable signature string.
- `customModuleDecode` mirrors the in-section text into `stable_types_text` on round-trip, so any consumer that loads a wasm that *does* embed the section still works unchanged.
- Classical persistence and non-`--enhanced-migration` builds keep embedding the section; behaviour there is unchanged.


Kamirus (approved):

Kamirus:
I've revived the tests to make sure we test that
1. We don't loose .most with this flag on
2. Stable section is not there for EM, but it is there for non-EM

PR #6074, opened 2026-04-30
feat: dedup structurally-equal types in stable signatures

Kamirus:
## Why

Long enhanced-migration chains redeclare the same type shapes under fresh nominal names at every step, exploding the emitted stable signature — the `.most` file used by `prev.most` baselines / `check-stable` / `--stable-baseline`, and the `motoko:stable-types` Wasm custom section where still emitted (see [#6073](https://github.com/caffeinelabs/motoko/pull/6073), which strips the section under `--enhanced-migration`). On the internal `enhanced-migrations/long-rich` stress chain (N=500) the signature is 1.14 MB; a real staging app (`roomreserve`) carries a 429 KB `.most` of which ~36% was duplicate declarations.

## What the printer now does

`pp_stab_sig` collapses structurally-equal zero-arity type aliases at render time. Grouping is purely structural — a `Typ_hash` bucket confirmed by `Type.eq`, so hash injectivity is not load-bearing — with three name-aware refinements:

- a name lost to a cross-name merge is preserved as a one-line residual alias;
- nullary bare aliases dedup in a second pass keyed by `(name, final target)`, so they can never collapse onto each other (`type X = X`) or hijack a prim's class;
- a re-export alias whose final target carries the same name (`type IntegrationDomain = Mod.IntegrationDomain`, core's `type Nat = ...`) is dropped entirely and its uses point at the target.

Before:

```motoko
type Event__1 = {#added : Nat; #removed : Nat};   // one copy per migration step
type Event__2 = {#added : Nat; #removed : Nat};
type OldActor__3 = {events : [Event__1]; total : Nat};
type NewActor__4 = {events : [Event__2]; total : Nat};
type Registry__5 = Map__9<Text, Event__1>;
type Registry__6 = Map__9<Text, Event__2>;
type Nat__7 = Nat;
stable v : {a : OldActor__3; b : NewActor__4; r : Registry__5; n : Nat__7};
```

After:

```motoko
type Event__1 = {#added : Nat; #removed : Nat};
type NewActor__4 = OldActor__3;
type OldActor__3 = {events : [Event__1]; total : Nat};
type Registry__5 = Map__9<Text, Event__1>;
stable v : {a : OldActor__3; b : OldActor__3; r : Registry__5; n : Nat};
```

`test/run-drun/stable-sig-dedup-showcase.mo` demonstrates every behavior in one example with a checked-in golden.

## Rendering never mutates the cons graph

Rewriting happens on a parallel graph: each surviving cons that needs rewriting gets a `Cons.clone`, kinds are rewritten in-clone via `Cons.unsafe_set_kind`, and the clones are discarded when the printer returns (same pattern as the `.did` dedup in [#5013](https://github.com/caffeinelabs/motoko/pull/5013)). Clones are minted in a dedicated `Cons.session` stamp scope: the per-name counters are restored on session exit, so a clone minted in the enclosing scope could reuse a checker-era stamp and `Cons.eq`-collide with a live cons, silently dropping its decl (regression-tested in `stable-sig-clone-stamp.mo`).

## Impact

| build | `long-rich` N=500 `.most` |
| --- | --- |
| 1.7.0 (no dedup) | 1,140,909 B |
| this PR | 360,752 B (−68%) |

On `roomreserve` (staging), the remaining duplication after the name-scoped draft was 32% cross-name `OldActor`/`NewActor` pairs plus repeated parameterized and nullary aliases — all covered now.

## Compatibility

Signature matching is structural, so old and new outputs are mutually compatible; verified in both directions between 1.7.0-, draft-, and this-build-produced signatures on `long-rich`, including 1.7.0 parsing the alias-bearing output. Same-shape types under *different* user names still collapse (the alias keeps the name in the signature); semantically-distinct same-shape names like `type Ms = Nat` vs `type Cents = Nat` remain separate declarations.

## What is unchanged

`Typ_hash` itself (its partiality — `Async`, type components — is tolerated at the call site, and dedup falls back to not collapsing), `validate_stab_sig`, `match_stab_sig`, and parameterized (arity > 0) type declarations, which are never grouped.


ggreif:
I did a similar deduplication with great success back then for `.did` output. Let me find the PR.

_EDIT_: here is is #5013!

PR #6082, opened 2026-05-05
fix: macOS nightly debug test time out

alexandru-uta:
For MacOS nightly test runs, the `trap-on-oneway-failure.mo` test was timing out after `600 seconds`.

Bumping the RTS optimization level to `opt-level = 1`, speeds it up and makes the CI [green](https://github.com/caffeinelabs/motoko/actions/runs/25365404435).

We also retain the debug assertions explicitly on the debug builds.


ggreif (approved):
Awesome!

PR #6084, opened 2026-05-05
feat!: searches nested modules for implicits

christoph-dfinity:
## Motivation

This PR allows defining implicits in nested modules to allow library authors to define their own "type classes" without requiring an explicit import for every single instance. As an example this would allow us to move the viewer functions into a module and add an import, rather than "abusing" mixins to bring them into scope.

Before:
```motoko
import Hash "mo:siphash/Hash";
import HashNat "mo:siphash/Nat";
import HashInt "mo:siphash/Int";
import HashText "mo:siphash/Text";

Hash.hash(1);
Hash.hash(-1);
Hash.hash("hello");
```

After:
```motoko
import Hash "mo:siphash/Hash";

Hash.hash(1);
Hash.hash(-1);
Hash.hash("hello");
```
 
It searches up to a depth of 8 (hardcoded for now) to avoid looping on recursive module type definitions.

## Design considerations
As is, this PR is a breaking change, as it could theoretically change what existing code using implicits resolves to.

I've been thinking about adding `implicit module` definitions, that explicitly mark which modules are allowed to be recursively searched. They'd be semantically equivalent to `NamedT("implicit", module { ...})`

Compile performance wise we will search more modules for implicits then before, but nested modules are used rarely in Motoko at the moment. If this is a concern it could also be alleviated with the aforementioned `implicit module`s, as those would constrain the search space further.


Changelog.md
```diff
@@ -4,6 +4,7 @@
 
 * motoko (`moc`)
 
+  * feat: Search nested modules for implicit candidates. When module `M` is in scope, implicit resolution will now consider candidates like `M.N.compare` (#6084).
```
Kamirus: damn AI reviewer again did not flag editing an already released version 😆 
I guess we should enforce that even more, maybe as a separate explicit step for the agent to check that the changelog edit is correct...

Kamirus:
`implicit module` would save us from a breaking change and it might be worth doing if this is the direction we want to take.
If not or if we want to merge this PR faster, I think it's ok to just make a major bump together with some other breaking changes that we can group together for 2.0.0

Kamirus (commented):
1. Should this nested resolution work with the implicit-package flag? Should we resolve nested candidates from implicit modules or maybe only search for them and only suggest imports (so that we rely less on the implicit flag)? Regardless what we choose we should have a test covering this - either as an import suggestion or a successful resolution
2. Could you add a test that suggests an import for a nested candidate? maybe a test with `--all-libs`?

PR #6085, opened 2026-05-05
fix: allow resolving context-dot for async functions

christoph-dfinity:
By hardcoding the scope to `None` before, we made it so async functions would never be resolved for context-dot.

Threading the scope through fixes that.

Reported by @timohanke 

@Kamirus I didn't see an obvious way to get a hold of a `scope` for the context dot suggestions in the LSP, so for now completions for async functions won't appear.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,10 @@
 # Motoko compiler changelog
 
+## Unreleased
+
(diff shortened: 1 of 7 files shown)


alexandru-uta (approved):
thanks, looks good

PR #6089, opened 2026-05-06
chore: serialise `tests` builds on `arm64-linux-16`

ggreif:
## Summary
Extends the same `max_jobs=1` workaround used by `gc-tests` in #6009 to the `tests` matrix on `arm64-linux-16`. `ubuntu-24-large` keeps `max-jobs=auto`.

## Why
`tests (arm64-linux-16, release)` and `tests (arm64-linux-16, debug)` have been OOM-killed by the runner whenever `nix-build-uncached --max-jobs auto` schedules three parallel ~3 GB derivation builds. Most recent observation: https://github.com/caffeinelabs/motoko/actions/runs/25409615273/job/74528169567 (initial CI on the `pocket-ic` bump PR #6088 — failure was unrelated to the bump itself).

## Why max-jobs=1 (not 2)
We probed a higher ceiling but it's not stable enough to ship:

- PR #6090 ran `gc-tests` and `tests` at `max-jobs=2` against a cache-busted `test-runner` and **appeared** to pass.
- PR #6088 with a *different* cache-buster (forcing a different content hash than #6090's, and therefore a real cold rebuild of the GC-variant derivations) hit OOM on `gc-tests (arm64-linux-16, 2)` after ~2.5 minutes:
  ```
  building 'test-drun-compacting-gc.drv'...
  building 'test-drun-generational-gc.drv'...
  Killed   nix-build-uncached -build-flags "--max-jobs 2"
  ```
  Run: https://github.com/caffeinelabs/motoko/actions/runs/25411132891/job/74532926605

The result depends on which derivations happen to be cachix-cached vs cold-rebuilt at any given moment. An asymmetric `gc-tests=1, tests=2` configuration would also be brittle: a future test addition could shift the memory profile of `release-systems-go` and silently reintroduce OOMs. Going uniform `max-jobs=1` is slower but stable.

## Test plan
- [x] CI green
- [x] No `tests (arm64-linux-16, *)` OOM on the resulting run

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6090, opened 2026-05-06
[experiment] arm64-linux-16 max-jobs=2

ggreif:
Probing whether two parallel ~3 GB derivation builds fit on the 16 GB runner. If green, we can relax PR #6089 from max-jobs=1 to max-jobs=2.

Stacked on top of #6089. Will close once we've learned.


ggreif:
Experiment served its purpose — `max-jobs=2` works on `arm64-linux-16` with cold rebuilds. Results folded back into PR #6089 (commit [`501449704`](https://github.com/caffeinelabs/motoko/commit/501449704)).

PR #6094, opened 2026-05-06
spec: align IDL-Motoko.md with Float32 support

Kamirus:
## Summary

- Adds `e(Float32) = float32` and `i(float32) = Float32` to the type mapping tables.
- Removes the note that said importing `float32` was unsupported/failing.

## Why

`Float32` support was shipped in #5906, which updated both `idl_to_mo.ml` and `mo_to_idl.ml` to handle the `float32 ↔ Float32` mapping. The spec doc (`design/IDL-Motoko.md`) was not updated at the time, leaving it contradicting the implementation.

This is a companion to a fix in [dfinity/candid](https://github.com/dfinity/candid/pull/new/fix/motoko-binding-float32) where `didc bind --target mo` panics on any `.did` file containing `float32`, because it (correctly) follows the now-outdated spec.

### DIFF
--- a/design/IDL-Motoko.md
+++ b/design/IDL-Motoko.md
@@ -85,6 +85,7 @@ e(Nat) = nat
 e(Int) = int
 e(Nat<n>) = nat<n> for n = 8, 16, 32, 64
 e(Int<n>) = int<n> for n = 8, 16, 32, 64
+e(Float32) = float32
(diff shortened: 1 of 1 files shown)


ggreif (approved):
Indeed, this was an omission! Thanks for catching it!

PR #6096, opened 2026-05-07
fix: M0236 dot-notation auto-fix preserves call parentheses for unparenthesized calls

Kamirus:
## Problem

`mops check --fix` on `Module.f arg` (single arg, no parens) silently rewrote the call into a function reference:

```motoko
ignore List.reverse b;       // before
ignore b.reverse ;           // after  ← no longer calls reverse
```

`b.reverse` type-checks as a function value, so the bug propagates without error. The parenthesized form was unaffected.

## Fix

The remove-argument edit deleted the lone arg's range; for `f(arg)` the surrounding `(…)` survived, but for `f arg` no parens existed to keep. Detect this case and replace the arg with `"()"` instead of `""`.

After the fix: `List.reverse b` → `b.reverse ()`. Type instantiation is preserved (`List.reverse<Nat> b` → `b.reverse<Nat> ()`).


christoph-dfinity (approved):
LGTM

PR #6086, opened 2026-05-05
chore: bump `clang` to v21

ggreif:
## Summary

Mechanical bump from clang/llvm 20 to 21 across the nix flake and the RTS Makefile. Same shape as PR #6009 (clang 19 → 20).

- `flake.nix` — `WASM_CLANG`, `WASM_CLANG_LIB`, `CLANG`
- `nix/rts.nix`, `nix/tests.nix`, `nix/ic-wasm.nix` — `llvmPackages_20` → `llvmPackages_21`
- `rts/Makefile`, `rts/README.md`, `test/ld/Makefile` — `clang-20`/`wasm-ld-20` → `clang-21`/`wasm-ld-21`
- `rustStdDepsHash` unchanged (rust-nightly didn't move).

## Test plan

- [x] CI green: `tests`, `gc-tests`, `common-tests` all pass with the v21 toolchain on linux and macOS runners.
- [x] `nix build .#rts` succeeds locally.
- [x] `nix build .#tests.ld` regenerates the same `representative.linked.wat.ok` (clang-21's `wasm-ld` may shift the golden — accept if so).

Draft until #6009 merges; converting + merge in a week.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6099, opened 2026-05-08
fix(nix): link `ic-wasm` against `llvmPackages_21.libcxx` on darwin

alexandru-uta:
The clang 21 bump (#6086) broke `ic-wasm`, and therefore `nix develop`, on apple-silicon.


ggreif (approved):
Rubber-stamping. But this is strange.

PR #6103, opened 2026-05-09
fix(codegen): see through `i32.wrap_i64` in LSB-and-1 → `ctz` peephole

ggreif:
## Summary

Enhanced (EOP) inserts an `i32.wrap_i64` before every `if` (since the i64 condition produced by `i64.and 1` is too wide for `if`'s i32-typed condition operand), interposing between `Binary And` and `If`. The existing i64 peephole at `instrList.ml:99-100` required adjacency, so it never fired in EOP and the discriminator emitted by `BitTagged.if_both_tagged_scalar` (driving Int `==`, comparisons, `pow`, …) stayed as `or; const 1; and; wrap; if(slow,fast)` instead of the desired `or; ctz; wrap; if(fast,slow)`.

The wrap is semantically irrelevant for the LSB test — `wrap(x & 1)` and `wrap(ctz x)` give the same boolean answer to `if` — so adding a sibling case that pattern-matches across the wrap is sound. Second commit drops the now-unreachable bare-i64 peephole (`[i64.and 1; if]` is ill-typed wasm — `if` requires an i32 condition, so this shape never reaches the optimizer).

Third commit extends the same rewrite to the `br_if` consumer (i32 + two i64 forms), no leg-swap since `br_if` has only one branch direction. No moc codegen currently emits `br_if`, but the upcoming variant-switch `BitTest` strategy for n=2 dispatch (planned on `gabor/variant-switch`) will — see [BitTest entry in the plan](https://github.com/caffeinelabs/motoko/blob/gabor/variant-switch/.claude/plans/variant-switch-br-table.md).

## Empirical (disc-actor, Nat add+mul)

| | ctz emitted | wasm size |
|---|---:|---:|
| before | 1 | 60 888 |
| after | 52 | 60 786 |

Classical (legacy persistence) was already emitting ~58 ctz per actor — this PR brings EOP up to parity.

## CI bench deltas

All wins or flat — no regressions:

| bench | Δ cycles | rel |
|---|---:|---:|
| alloc (×3) | −16.78 M each | −0.33% |
| region-mem | −25.17 M | −0.37% |
| region0-mem | −25.17 M | −0.38% |
| stable-mem | −25.17 M | −0.55% |
| heap-64 (×2) | −700 K each | −0.05% |
| heap-32 (×2) | −499 K, −600 K | −0.04% |
| variant-switch (×3) | −10 K to −16 K | −0.06% to −0.11% |
| iter (×3) | −2 K each | −1.03% |
| palindrome (6) | −1 to −33 | sub-noise to −0.17% |
| bignum / candid / nat16 | sub-noise | flat |

Bench `.ok` files updated by the CI cycle-counting bot in `78286dcf2`.

## Test plan

- [x] `test/run/lsbit-ctz-peephole.mo` — FileCheck regression test (EOP-only): asserts `i64.or; i64.ctz; i32.wrap_i64; if` adjacency inside `B_eq`. Verified to fail when the new peephole case is reverted.
- [x] CI bench numbers updated and committed.
- [ ] `br_if` peepholes have no moc-emitted consumers today (no `br_if` in current output); they're forward-looking infrastructure for `BitTest`. Will be exercised by the variant-switch follow-up PR.


ggreif:
**Wasm-coverage audit completed (LLM-assisted Wasm-spec review)** — coverage of i32-condition consumers in the local-peephole envelope is complete.

Full enumeration across the core spec (1.0/2.0) plus all live proposals (reference-types, GC, exception-handling, threads/atomics, tail-call, function-references, SIMD, relaxed-SIMD, memory64, multi-memory):

| consumer | status |
|---|---|
| `if cond` | ✅ covered (i32 + i64-with-wrap) |
| `br_if L cond` | ✅ covered (i32 + i64-eqz + i32-wrap-then-i32-eqz) |
| `select` / `select t` | ⚠️ deferred — needs reordering the two value operands pushed before the condition, outside the local zipper window |
| `br_table` | ❌ numeric index, not a boolean consumer |

Not i32-condition consumers (correctly out of scope): GC `br_on_null`/`br_on_non_null`/`br_on_cast`/`br_on_cast_fail` (reference-typed), `try_table` (tag-typed), `call_indirect` and the tail-call family (table-index dispatch), atomics/SIMD/Memory64 (no new condition shapes). `select_spectre_guard` is a Cranelift IR node, not a Wasm instruction.

**Additional note on \`select\`**: moc doesn't emit \`select\` at the moment (zero occurrences in the disc-actor and Result-switch samples) — same situation as \`br_if\` before the variant-switch BitTest work motivates it. So even the deferred case has no current consumer to feed it.

alexandru-uta (approved):
looks safe as far as I can tell; thanks!

PR #6111, opened 2026-05-12
perf: write-barrier improvements (multi-value `if` + cached running-GC flag)

ggreif:
## Summary

Two layered optimizations of the EOP backend's `Tagged.write_with_barrier`, the inline guard emitted at every `var := value` assignment.

### Commit 1 — multi-value `if (param i64 i64)`

The previous shape stashed both `[location, value]` operands into locals before the branch and reloaded them in each arm — 8 local-set/get instructions per call site plus two i64 locals declared in the host function.

This commit uses a **multi-value `if` block-type** (`(param i64 i64) (result)`, wasm 1.0) to thread the operands directly through the branch:

```text
before:                       after (commit 1):
  set_value                     <gc-running?>
  set_location                  if (param i64 i64)
  <gc-running?>                   call write_with_barrier
  if (empty)                    else
    get_location                  i64.store
    get_value                   end
    call write_with_barrier
  else
    get_location
    get_value
    i64.store
  end
```

| | before | after | Δ |
|---|---:|---:|---:|
| sample 2-method actor with one `var` field | 58,293 B | 56,065 B | **−2,228 B (~4%)** |

The win is both **code size and cycles**: every assignment to a pointer-typed mutable variable used to pay 4× `local.set` + 4× `local.get` to stash and reload the operands across the branch. The IC's instruction counter ticks wasm ops, so those 8 saved ops are a direct cycle reduction at every `var := value` site (proportional to assignment density, not just barrier-firing density). The transform also drops two declared i64 locals from the enclosing function, eliminating their slot allocation cost. (`<gc-running?>` is `call running_gc` at this commit; commit 2 abolishes the call and replaces it with `global.get __running_gc`.)

### Commit 2 — cache the running-GC flag in a backend global

`Tagged.write_with_barrier` previously checked GC state via an RTS-imported `running_gc()` call at every assignment. That import is now abolished in favor of an i32 global `__running_gc` maintained on the backend side:

- New `set_running_gc(state)` export, generated alongside `bigint_trap` in `RTS_Exports.system_exports`.
- RTS-side `Phase` state's `phase` field renamed to `phase_inner` and routed through `State::phase()` / `State::set_phase()`. `set_phase` detects every `Pause ↔ non-Pause` transition and pushes the flag through `set_running_gc`.
- Backend's `write_with_barrier` now reads the global directly via `GlobalGet "__running_gc"`, dropping the indirect-call round-trip.
- Classical+incremental backend mirrors the same change, gated on `Flags.gc_strategy = Incremental`.

The obsolete `pub unsafe extern "C" fn running_gc()` in `rts/motoko-rts/src/gc/incremental/barriers.rs` is removed.

```text
after (commit 2):
  global.get __running_gc
  if (param i64 i64)
    call write_with_barrier
  else
    i64.store
  end
```

A `call` + function entry/exit overhead is replaced by a single `global.get` per assignment site — another cycle win, this one proportional to barrier-firing density rather than assignment density.

### Commits 3–5

- **Constant-fold `ptr_unskew + MutBox.field * word_size`** in the two enhanced + four classical `Var.set_val` sites — one `compile_add_const` instead of two.
- **Drop redundant inner parens** in `Some ((Foo bar), typ)` pattern matches — 9 sites in enhanced, 11 in classical.
- **Rustfmt wrap** of `mark_completed` condition after the `.phase` → `.phase()` rename.

## Why now

`VarBlockType` in `wasm-exts/ast.ml` has handled param-carrying block types since the start (per wasm 1.0 spec), but no codegen site was using it. This is the first one. It doubles as a feasibility probe:

- Other barrier sites with single-i64 input (`Tagged.allocation_barrier`) could use `if (param i64) (result)` similarly.
- The `compile_eq` early-`i64.eq` shortcut (which currently uses `E.if1 I64Type` plus stashing) is a candidate.

The cached-flag approach also lays groundwork for further per-call-site barrier elision once we know statically (or via shape analysis) that the GC cannot be running at a given site.

## Test plan

- [x] Local `test/run` quick suite passes after RTS rebuild.
- [x] `gc-random-test-force-gc` passes — exercises the backend↔RTS handshake under forced GC cycles.
- [x] Classical+incremental `self-upgrade.drun` passes locally with `EXTRA_MOC_ARGS="--sanity-checks --incremental-gc"` — matches the CI `test-drun-incremental-gc` configuration that originally caught the missing classical-side mirror.
- [x] CI on this PR — `wasm-validate`, the IC runtime, and bench numbers (the latter should show measurable per-assignment cycle reductions).


src/codegen/compile_enhanced.ml
```diff
@@ -2134,19 +2141,16 @@ module Tagged = struct
...
-    ) (
-      get_location ^^ get_value ^^
+    (* Stack on entry: [location, value]. Read the backend-cached
+       running-GC flag (i32) and dispatch via a multi-value
+       `if (param i64 i64)` block-type so the operands flow through
+       without locals. The RTS pushes the flag via `set_running_gc`
+       on every Pause↔non-Pause transition. *)
+    G.i (GlobalGet (nr (E.get_global env "__running_gc"))) ^^
+    G.if_
+      (VarBlockType (nr (E.func_type env (FuncType ([I64Type; I64Type], [])))))
```
ggreif: We'll want a pretty helper: `if_param [x2 I64Type]` or such.

ggreif:
Bench bot numbers per-commit, after both commits landed on this branch:

**Commit 1** — multi-value `if (param i64 i64)` ([c687933f6](https://github.com/caffeinelabs/motoko/commit/c687933f63fdafd4387df9f68839f6b49d06a995)):

| bench | delta | % |
|---|---:|---:|
| **iter** (each of 3 prints) | **−4,100** | **−2.05%** |
| heap-32 (line 1) | −600,916 | −0.053% |
| heap-32 (line 2) | −600,288 | −0.049% |
| heap-64 (line 1) | −599,868 | −0.046% |
| heap-64 (line 2) | −599,232 | −0.047% |
| bignum (line 1) | −144 | −0.006% |
| bignum (line 2) | −4 | noise |

**Commit 2** — cache the running-GC flag in a backend global ([85e25513f](https://github.com/caffeinelabs/motoko/commit/85e25513fa3ac427aecf06cccdfff0e4d668862e)):

| bench | delta | % |
|---|---:|---:|
| **iter** (each of 3 prints) | **−11,275** | **−5.77%** |
| heap-32 (line 1) | −1,652,519 | −0.145% |
| heap-32 (line 2) | −1,650,792 | −0.134% |
| heap-64 (line 1) | −1,649,637 | −0.126% |
| heap-64 (line 2) | −1,647,888 | −0.129% |
| bignum (line 1) | −372 | −0.015% |
| bignum (line 2) | −11 | noise |

**Cumulative vs master baseline:**

| bench | total delta | % |
|---|---:|---:|
| **iter** | **−15,375** | **−7.69%** |
| heap-32 (line 1) | −2,253,435 | −0.198% |
| heap-64 (line 1) | −2,249,505 | −0.172% |

The pattern: cycle savings track barrier-firing density per loop iteration. `iter` is the canonical case (tight `var := value` inner loop). Heap-* benches see ~0.18% because heap-grow dominates; bignum is at noise floor because bigint ops dominate.

Commit 2 is the bigger of the two, as expected: replacing a `call running_gc` (RTS function call — call setup, branch indirection, RTS body, return) with a single `global.get __running_gc` saves ~10 wasm-op-equivalents per barrier vs commit 1's ~8 from local-stash elimination.

ggreif:
## Review: write-barrier improvements (#6111)

### Overview

This PR layers three concrete improvements onto the incremental-GC write barrier:

1. **Multi-value `if (param i64 i64)`** keeps `[location, value]` on the wasm stack across the GC-running branch instead of stashing/reloading via 4× `local.set` + 4× `local.get`.
2. **Cached `__running_gc` i32 global** replaces the RTS `running_gc()` extern call at every barrier site. The RTS pushes the flag via a new `set_running_gc(state)` export from inside `State::set_phase()` on every Pause↔non-Pause transition.
3. **Cleanups**: constant-fold `ptr_unskew + MutBox.field * word_size`, drop redundant inner parens in `Some ((Foo bar), typ)` matches, add `E.if'` / `E.i64s` helpers, classical+incremental mirror.

### GC invariants — `set_running_gc` and `phase` mutation paths

Audited every site that touches the phase field:

- **All runtime mutations go through `set_phase`**: 6 call sites — `start_marking` (→Mark), `start_evacuating` (→Evacuate), `start_updating` (→Update), `complete_run` (→Pause), `stop_gc_on_upgrade` (→Stop), `resume_gc_after_upgrade` (→Pause). No direct `state.phase_inner = …` writes outside `set_phase`.
- **Two struct-initializer sites** write `Phase::Pause` directly: the static `STATE` initializer (`incremental.rs:192`) and `initial_gc_state` (`incremental.rs:224`). Both are pre-runtime; the backend initializes the global to `0l` to match, so startup is consistent and no transition needs signaling.
- **Pause↔non-Pause is correctly the only transition that fires `set_running_gc`**: the `was_running != now_running` guard in `set_phase` skips Stop↔Pause (both not-running) and Mark↔Evacuate↔Update (all running), saving the extern call when it would be a no-op.
- **Upgrade flow**: pre_upgrade does `Mark/Evacuate/Update → Stop` (was=now=running, no call), state persists, new wasm instance loads with global=`0`, state loads with phase=`Stop` — transient mismatch but no user code runs between pre/post. Post_upgrade does `Stop → Pause` (was=running, now=not), `set_running_gc(0)` fires, global re-syncs with phase. Sound.
- **`#[cfg(feature = "ic")]` gating** is correct on both the extern declaration and the `set_phase` call site, so RTS unit tests don't need a stub.

Single residual question: if some future RTS code path writes `phase_inner` directly (bypassing `set_phase`), the global will desync silently. The field's privacy (no `pub`) limits the blast radius to this module, but a comment on the field block — "DO NOT assign directly; use `set_phase`" — would future-proof it.

### Code correctness

- **Multi-value `if` operand types**: stack on entry to `write_with_barrier` is `[location: i64, value: i64]`; after `global.get __running_gc` it's `[location, value, cond: i32]`. The `if (param i64 i64)` block-type consumes the i32 cond, both arms see `[location, value]` and consume both (call into RTS / `i64.store`). Types check.
- **`E.if'` deliberately omits `prepare_branch_condition`** (the `i32.wrap_i64` that `E.if_` applies for Motoko-Bool conds) — correctly suited to a global-i32 cond. The docstring calls this out explicitly. Other callers (Motoko Bool) keep using `E.if_`.
- **Classical+incremental mirror**: same shape, gated on `Flags.gc_strategy = Incremental`. The `set_running_gc` export is conditional on the same flag, so non-incremental classical builds don't generate it (which is fine since the RTS `set_running_gc` extern is itself behind `cfg(feature = "ic")` + incremental code paths).
- **Constant fold**: `ptr_unskew + MutBox.field * Heap.word_size` are both compile-time constants; folding is semantics-preserving and removes one wasm `i64.add` per `Var.set_val` lvalue site.

### Performance

Bench-bot results posted on the PR: **−7.69% cycles on `test/bench/iter`** (cumulative), proportional savings on heap-32/64 (~0.18%), at the noise floor on bignum. Numbers track barrier-firing density, which is the expected story.

### Tests

- `gc-random-test-force-gc.drun` covers the backend↔RTS handshake under forced GC cycles.
- Classical+incremental path verified via `EXTRA_MOC_ARGS="--sanity-checks --incremental-gc"` on `self-upgrade.drun` — caught a bug at PR time (missing classical-mirror) and is now passing.
- Full CI green: common-tests, gc-tests, tests × {debug, release} × {ubuntu, arm}.
- Possible additional coverage: a `drun` test that explicitly forces a Pause→Mark→Pause transition during a barrier-heavy workload and asserts the global flips — but `gc-random-test-force-gc` likely already exercises this implicitly.

### Suggestions (non-blocking)

- Consider adding a doc comment above `State.phase_inner` field warning future RTS contributors to use `set_phase()` (privacy enforces it within the module; a comment makes the rule explicit).
- The PR body's "Why now" mentions `Tagged.allocation_barrier` and `compile_eq`'s i64.eq shortcut as natural follow-up sites for the multi-value-if pattern — worth filing as a follow-up issue to keep the trail visible.

### Verdict

**LGTM** modulo the optional comment-suggestion. The GC invariant story is tight (all phase mutations centralized, init states match, upgrade flow safe). The codegen side composes cleanly with the existing infrastructure and the bench numbers validate the rationale.

ggreif:
Follow-up: bench-bot numbers ([a8494e592](https://github.com/caffeinelabs/motoko/commit/a8494e59208cf67b096812a423a0dfece651f3aa)) after [0a5fdb45f](https://github.com/caffeinelabs/motoko/commit/0a5fdb45f) (inline running-GC fast path in `Tagged.allocation_barrier`):

| bench | line | before | after | delta | % |
|---|---|---:|---:|---:|---:|
| **alloc** | each of 3 prints | ~5,069M | ~4,885M | **−184,549,376** | **−3.64%** |
| **palindrome** | line 1 | 21,903 | 21,144 | −759 | −3.46% |
| **palindrome** | line 2 | 20,902 | 20,143 | −759 | −3.63% |
| **palindrome** | line 3 | 21,877 | 21,118 | −759 | −3.47% |
| **palindrome** | line 4 | 19,923 | 19,329 | −594 | −2.98% |
| **palindrome** | line 5 | 18,328 | 17,745 | −583 | −3.18% |
| **palindrome** | line 6 | 19,723 | 19,140 | −583 | −2.96% |
| **variant-switch** | `instr_transform` | 1,189,148 | 1,140,748 | −48,400 | **−4.07%** |
| variant-switch | `instr_evalFT` | 21,536,248 | 21,150,148 | −386,100 | −1.79% |
| variant-switch | `instr_eval` | 24,509,348 | 24,123,248 | −386,100 | −1.58% |
| candid-subtype-cost | only line | 1,118,481 | 1,107,206 | −11,275 | −1.01% |
| iter | each of 3 | 184,215 | 184,182 | −33 | noise |
| heap-32 / heap-64 | various | ... | ... | −44 each | noise |
| bignum | line 2 | 54,911,654 | 54,911,599 | −55 | noise |

The winners are **allocation-heavy** benches — exactly where eliding the `call allocation_barrier` per-object overhead dominates. `iter` is flat because its hot loop is write-barrier-bound (already addressed by earlier commits); `heap-*` is bound by heap-grow, `bignum` by libtommath ops.

`palindrome` at **−3.5%** and `alloc` at **−3.64%** per print are the headline gains — bigger per-bench payoff than the write-barrier commits delivered, on workloads they didn't touch. Different barrier, different bench class.

Cumulative effect on motoko-typical allocation-heavy paths (alloc + palindrome + variant-switch transform/eval): solid 1-4% reduction across the board, with the dramatic single-bench hit on `instr_transform` (−4.07%) — a hot AST→FT translation loop that allocates per node.

src/codegen/compile_classical.ml
```diff
@@ -610,13 +610,20 @@ module E = struct
...
-    G.if_
-      (VarBlockType (nr (func_type env (FuncType (tys1, tys2)))))
-      thn els
-  *)
+  (* Multi-value-aware `if`: each of [param] and [return] is an
+     optional `(env, types)` bundle. With both omitted this is just
+     a nullary `if`. Mirror of `compile_enhanced.ml`'s `E.if'`. *)
+  let if' ?param ?return thn els =
+    let bt = match param, return with
+      | None,           None              -> ValBlockType None
```
```suggestion
      | None,           None
```

src/codegen/compile_enhanced.ml
```diff
@@ -678,6 +686,24 @@ module E = struct
...
+  (* [if' ?param ?return thn els] — block-type-aware multi-value `if`.
+     Each of [param] and [return] is an optional `(env, types)`
+     bundle. With both omitted the result is `G.if_ (ValBlockType None)
+     thn els` — a regular nullary-result `if`. Unlike `if_`, does
+     *not* `prepare_branch_condition` (no `i32.wrap_i64`): the caller
+     must put an i32 condition on top of the wasm stack themselves.
+     Use this for raw-wasm conditions like `global.get` of an i32 flag. *)
+  let if' ?param ?return thn els =
+    let bt = match param, return with
+      | None,           None              -> ValBlockType None
```
```suggestion
      | None,           None
```

rts/motoko-rts/src/gc/incremental.rs
```diff
@@ -151,19 +162,74 @@ enum Phase {
 /// Use a long-term representation by relying on C layout.
 #[repr(C)]
 pub struct State {
```
alexandru-uta, line 164: just a first brief look: are you sure this is not going break compatibility with existing programs? What happens to this data structure after an upgrade to the new version of moc? we basically break its layout.

I would try out some tests around this and some more scrutiny around modifying these data structures.

ggreif: Good question, and it made me build the test that was missing. Short answer: the layout is unchanged, but you were right that this deserved scrutiny — and the mid-cycle upgrade case turns out to be reachable and worth a regression test.

**The layout does not change.** `State` is `#[repr(C)]` ("Use a long-term representation by relying on C layout"), so its layout is fixed by field order and types, and this PR changes neither — `phase: Phase` merely becomes `phase_inner: Phase` in the same position. A Rust field rename cannot move a byte under `repr(C)`. Field-by-field against master:

```
phase{,_inner}: Phase, partitioned_heap: PartitionedHeap, allocation_count: usize,
mark_state: StableOption<MarkState>, iterator_state: StableOption<PartitionedHeapIterator>,
statistics: Statistics
```

identical in order and type. A genuine layout change would need `persistence::VERSION` (currently `1`) bumped, which `check_version` enforces with *"Incompatible persistent memory version"*.

Rather than leave that as prose, `75aec06f8` pins it where it cannot rot:

```rust
const _: () = assert!(core::mem::offset_of!(State, phase_inner) == 0);
const _: () = assert!(core::mem::size_of::<Phase>() == 4);
```

so a future reordering breaks the RTS build — with a comment pointing at the VERSION bump as the fix — instead of misreading persisted bytes in the field. Both hold for wasm32 and wasm64.

**What genuinely needed a test** is the other half of your question: what happens to a canister that upgrades *mid-GC-cycle*. Under an in-place EOP upgrade the persisted `State` keeps its phase, but the new module's `__running_gc` global starts at zero — so it must be re-seated, or every barrier takes the paused fast path while the GC is still marking. Note that `stop_gc_before_stabilization` / `start_gc_after_destabilization` only bracket the graph-copy path, so nothing sets the phase around an ordinary upgrade.

`bc002338c` adds `test/run-drun/gc-cache-upgrade.drun` for exactly that, and the sizing is the whole point: marking advances the increment clock once per element, so the heap is 192 × 64Ki ≈ 12.6M elements to exceed the 5,000,000-unit increment limit. My first attempt used 4.19M and **passed even with the re-seat deleted** — marking finished in one increment, the phase was `Pause` at upgrade time, and a stale cache agreed by accident. Above the limit it bites:

```
panic occurred in file 'src/gc/incremental.rs' at line 82: assertion `left == right` failed
  left: 0     <- cached __running_gc, stale after the upgrade
 right: 1     <- authoritative phase != Pause
```

That is with `resync_running_gc_cache()` commented out; restored, the test passes in both the plain and `--sanity-checks` EOP arms. So the upgrade path is now covered by a test that demonstrably fails without the fix, and the `--sanity-checks` oracle compares the cache against the authoritative phase at every scheduling point.

One caveat I would flag rather than paper over: this proves coherence for a *new* compiler reading its own persisted state. A canister built by an older `moc` and upgraded to this one has no `__running_gc` global in the old module at all, so there is nothing stale to inherit — the new module's `(start)` re-seats from the persisted phase before any barrier runs. If you want that cross-version path exercised too, it needs a two-compiler harness we do not currently have in-tree, and I would rather add that as a follow-up than fake it here.

ggreif:
Benchmark effect of the cached running-GC flag, from the `test/bench` expectations in this PR. The three biggest movers:

| benchmark | before | after | delta | % |
| --- | ---: | ---: | ---: | ---: |
| `iter` | 197,531 | 182,123 | −15,408 | **−7.80%** |
| `alloc` | 5,053,458,197 | 4,868,908,821 | −184,549,376 | **−3.65%** |
| `palindrome` | 21,894 | 21,135 | −759 | **−3.47%** (range −2.96% … −3.63% over its six measurements) |

The ordering follows the mechanism rather than being incidental. `iter` is `var`-heavy, so it is almost entirely write barriers — per-operation, replacing the `rts.running_gc` call with a `global.get` on the paused path is the largest relative saving available, and 7.8% is what a barrier-bound loop looks like. `alloc` is allocation-barrier dense at ~5G cycles, so its 3.65% is the biggest absolute win here (−184.5M cycles). `palindrome` sits in between, mixing mutable structure updates with real work.

For contrast at the other end: `bignum` moves −0.02% and −0.00%, which is the expected non-result for a benchmark that barely touches barriers — a useful sanity check that the win is coming from where we think it is.

Across every line the bench expectations move: **20,346,055,489 → 19,780,249,854 cycles, −565,805,635 (−2.78%)**, with no line regressing. `heap_bytes` is byte-identical on both sides throughout (`+671_088_640`, `+2_928`, …), so this is purely fewer instructions, not a change in allocation behaviour.

One footnote so the numbers reconcile with the other open PR: `candid-subtype-cost` shows −0.99% here, while #6311's Candid length bound moved that same benchmark +1.28%. Different PRs, opposite directions, roughly cancelling.

Kamirus (commented):
Re-checked after the master merge. Conflicts are gone and the bench expectations are current, so the rebase item is done. What remains:

- PR body calls the param-carrying `if` block type "wasm 1.0"; it is the multi-value proposal (wasm 2.0), and this is the first `VarBlockType` the classical backend ever emits. Worth a truthful sentence.
- The new comments are far too long and restate the diff, the discussion, and the code below them; the test comment also claims the cache is checked "on every barrier" when it is checked at GC schedule points. Fixed in #6366, which targets this branch.
- Optional: `#[cfg(debug_assertions)]` on a `debug_assert_eq!` is double gating; a plain `assert_eq!` under the cfg reads cleaner.

Otherwise LGTM, pending a GC-owner sign-off.

Kamirus (approved):

PR #6113, opened 2026-05-12
feat: re-enable `--experimental-multi-value` flag

ggreif:
## Summary

The `Flags.multi_value` ref and `FakeMultiVal` infrastructure for function-level multi-value codegen have been live in both backends since 2020, but the CLI flag was commented out in [src/exes/moc.ml](src/exes/moc.ml) back in [PR #1038 (2019, ggreif)](https://github.com/caffeinelabs/motoko/pull/1038) — "Temporarily disable -(no-)multi-value options" — pending two upstream blockers:

1. **OCaml `wasm` library supporting multi-value** — long done; `wasm-exts` already exercises `VarBlockType` for *block-level* multi-value, shipped via #3865 (2023) and #6111 (currently in review on `gabor/barrier`).
2. **IC replica accepting multi-result function signatures** — @nomeata's 2021 reproducer in the closed [PR #1459](https://github.com/caffeinelabs/motoko/pull/1459) showed `qr.mo` failing at install with `Failed to deserialize wasm module with Invalid or unknown value type 34`. PR #1459 was force-closed 2024-12-03 by @luc-blaeser as "inactive" with the note "Please re-open if it is still relevant".

## What I verified

Blocker (2) is gone. A throwaway probe compiled

```motoko
//MOC-FLAG --experimental-multi-value
persistent actor {
  func pair(n : Nat64) : (Nat64, Nat64) = (n, n + 1);
  public func go() : async () {
    let (a, b) = pair(42);
    Prim.debugPrint(debug_show a # " " # debug_show b);
  };
}
```

| backend | install on pocket-ic | runtime output |
|---|:-:|---|
| classical+legacy | ✓ | `debug.print: 42 43` |
| EOP | ✓ | `debug.print: 42 43` |

The classical wasm contains **30 `(result iN iN)` multi-result function signatures** when the flag is on; zero with the flag off (every multi-result helper falls back to `FakeMultiVal`'s `multi_val_*` global stash).

## What this PR does

- Re-enables `--experimental-multi-value` (and `--no-experimental-multi-value`) in [src/exes/moc.ml](src/exes/moc.ml), reverting the temporary disable from [PR #1038](https://github.com/caffeinelabs/motoko/pull/1038) and effectively reviving the broader effort tracked in [PR #1459](https://github.com/caffeinelabs/motoko/pull/1459) (force-closed as inactive in 2024). The flag stays **opt-in** — default behavior is unchanged. No codegen path touched; the infrastructure was already there.
- Keeps the original `-multi-value` / `-no-multi-value` single-dash spellings as deprecated aliases that print a stderr warning pointing at the new flag.
- Adds `test/run/multi-value-experiment.mo` with FileCheck assertions that multi-result function types appear and `multi_val_*` globals do not.
- Adds `test/run-drun/multi-value-experiment.drun` for an end-to-end install+ingress check against pocket-ic.

## Follow-ups (out of scope here)

- Flip the default to multi-value-on once we have wider canister coverage.
- Delete `FakeMultiVal` and its `multi_val_*` globals after the default flip.
- Wire `--experimental-multi-value` into the IC-replica feature-flag matrix the test harness uses, so the EOP and classical CI paths both exercise it.


ggreif:
**Correction**: my earlier bench numbers in this comment were contaminated. I captured them on a working tree that also had the [PR #6111](https://github.com/caffeinelabs/motoko/pull/6111) `__running_gc` global-cache RTS built — so the egcd "18×" headline was actually that PR's per-barrier elision, not multi-value. Clean `gabor/multi-value-probe` (master RTS, per-barrier `call running_gc`) numbers, captured with the same `test/bench/multi.mo`, EOP backend:

| bench | MV on | MV off | delta |
|---|---:|---:|---:|
| `fibPair(40)` × 1k | 7,036,422 | 7,364,430 | **−4.5%** |
| `egcd(46368, 75025)` × 10k | 916,800,538 | 920,900,546 | **−0.45%** |
| `leb128Length(0xDEADBEEFCAFEBABE)` × 100k | 178,100,313 | 186,100,321 | **−4.3%** |
| **total `rts_lifetime_instructions`** | 1,147,414,744 | 1,158,743,092 | **−0.98%** |

Heap allocations are essentially identical between modes — bignum tuples get boxed regardless of how the multi-value return is encoded, so `FakeMultiVal` global stash vs native wasm multi-result is just a register/stack-shape question, not an allocation-pressure one. The earlier "0 B vs 85 MB" claim was also wrong; both modes allocate ~85 MB for the egcd test.

So on EOP today, `--experimental-multi-value` is a small wasm-size and cycles win across these three benchmarks (~1% total cycles, low single-digit per-bench). The dramatic egcd savings I posted come from the unrelated barrier-cache change in #6111, which is its own PR.

test/bench/Makefile
```diff
@@ -1,5 +1,5 @@
 RUNFLAGS = -p
-export EXTRA_MOC_ARGS = --enhanced-orthogonal-persistence
+export EXTRA_MOC_ARGS = --enhanced-orthogonal-persistence --experimental-multi-value
```
ggreif, line 2: Maybe `test-runner` needs a parallel change?

flake.nix
```diff
@@ -238,7 +238,21 @@
...
+      packages =
+        let
+          # `-mv` helpers append `--experimental-multi-value` to every
+          # constituent test derivation's `EXTRA_MOC_ARGS`. Used by
+          # `nightly-macos-test.yml` to exercise the multi-value
+          # codegen path across the test matrix on macOS. Non-test
+          # derivations accept the override harmlessly — they don't
+          # read `EXTRA_MOC_ARGS`. Hidden in a `let` so they don't show
+          # up as flake outputs.
+          appendMultiValue = drv: drv.overrideAttrs (oa: {
```
ggreif: This is only for a moderately short transition period.
- we'll deprecate the `FakeMultiVal` contraption in 2 major releases distance and after 2 more we'll flip the default. After another two the old mechanism will be removed. The modern flags will exist for a while more but will be inactive and only warn.

test/bench/Makefile
ggreif: No — `test-runner` is flag-agnostic. The `//MOC-FLAG` extractor (`test-runner/src/run_test.rs:477`, `mode.rs:88`) treats any directive string as opaque and passes it to `moc`; the `EXTRA_MOC_ARGS` env path (used here by `appendMultiValue` in `flake.nix`, and already by `--enhanced-orthogonal-persistence` on `test/bench/Makefile:2`) is concatenated onto the `moc` command line unchanged. New flag = no harness work.

.github/workflows/nightly-macos-test.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         uses: ./.github/actions/test-blueprint
         with:
           os: macos-latest
-          test-target: common-tests
+          test-target: common-tests-mv
```
alexandru-uta: not clear to me; why did we need to rename the target name?

flake.nix
```diff
@@ -291,6 +305,36 @@
               ++ builtins.attrValues js;
         };
 
+        common-tests-mv = pkgs.releaseTools.aggregate {
```
alexandru-uta: are we effectively doubling the number of tests? 

I think we can run tests in both ways (with and without the flag) by playing with test-runner and run-test instead of playing with the nix files. sort of in the same way as we run tests with `eop` and without.

alexandru-uta: i.e., overriding `EXTRA_MOC_ARGS` and adding similar support to `eop_only`

ggreif: Well, we are doubling but half is running on Linux (without `-mv`) and the others (with `-mv`, nightly) on MacOS.

This way all is tested. The distinction will go away when the `moc` flag is removed.

Or do you have a better suggestion with overriding `EXTRA_MOC_ARGS`? C.f. **line 244**...

.github/workflows/nightly-macos-test.yml
ggreif: These are macOS-only targets with `--experimental-multi-value` flag. You can run them on Linux too, but that is not part of current CI (not needed, Wasm is platform agnostic).

flake.nix
alexandru-uta: look at `test-runner/src/mode.rs`, we run tests in both ways if need be, or choose a single way. I would actually do it this way, instead of the nix way.

alexandru-uta (approved):
thanks, looks good!

flake.nix
ggreif: Did that!

PR #6116, opened 2026-05-13
fix(test-runner): recognise `// CHECK` (space) as `FileCheck` directive

ggreif:
## Summary

Both FileCheck gates in `test-runner/src/run_test.rs` (`parse_directives:283` and `filecheck:620`) used `line.starts_with("//CHECK")` with **no space** — but 24 of the 27 FileCheck-annotated tests in `test/run/` use the conventional `// CHECK:` form. Result: FileCheck was silently skipped on the majority of would-be FileCheck tests. The `[FileCheck]` phase never appeared in their output, and `// CHECK` directives were unenforced.

Fix: trim leading `/` and whitespace before matching `CHECK`, so both `// CHECK` and `//CHECK` activate.

## Test plan

- [x] Manually invoked FileCheck on the 19 silenced classical-mode tests (the 5 EOP-only ones have no wasm under default classical mode) — all 19 pass; fix is risk-free.
- [x] `make -C test/run -j8 quick` still green.
- [x] After the fix, the phase line shows `[FileCheck]` for all `// CHECK` tests, e.g. `top-bit: [tc] [run] [run-ir] [run-low] [comp] [valid] [FileCheck] [wasm-run]`.

### DIFF
--- a/test-runner/src/run_test.rs
+++ b/test-runner/src/run_test.rs
@@ -280,7 +280,7 @@ fn parse_directives(src: &str, is_drun: bool) -> Directives {
             {
                 d.filter.insert(ext.to_string(), cmd.trim().to_string());
             }
-            d.check |= line.starts_with("//CHECK");
+            d.check |= re(r"^// *CHECK").is_match(line);
             d.no_force_gc |= tagged("MOC-NO-FORCE-GC");
             d.no_skip_gc_deprecation_warning |= tagged("NO-SKIP-GC-DEPRECATION-WARNING");
             d.generational_gc_only |= tagged("GENERATIONAL-GC-ONLY");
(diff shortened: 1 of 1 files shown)


test-runner/src/run_test.rs
```diff
@@ -280,7 +280,10 @@ fn parse_directives(src: &str, is_drun: bool) -> Directives {
             {
                 d.filter.insert(ext.to_string(), cmd.trim().to_string());
             }
-            d.check |= line.starts_with("//CHECK");
+            d.check |= line
```
ggreif: Maybe just use a _regex_? `// *CHECK`?

ggreif: Or if at most one space: `// ?CHECK`

Kamirus (approved):

PR #6117, opened 2026-05-13
perf(codegen): MSB/LSB → `clz`/`ctz` peephole family

ggreif:
## Summary

Extends the existing LSB-and-1 → `ctz` peephole family (PRs #3873 / #6103) with the symmetric MSB-extraction → `clz` direction, the LSB-via-`shl (N-1)` route, and a new `if [Br N] []` → `br_if (N-1)` rule that finally lets moc emit `BrIf` at all. Also unsilences the FileCheck gate (commit 5265cd41e) so the new `// CHECK` directives are actually validated end-to-end — that fix is also shipping standalone as #6116.

### MSB-extraction → `Clz` family (`cf729f046`)

- `i32.(shr_u|shr_s) 31; if t e` → `i32.clz; if e t` (or-pattern on the shift opcode)
- `i32.and 0x80000000; if t e` → `i32.clz; if e t`
- `i32.(shr_u|shr_s) 31; eqz; br_if l` → `i32.clz; br_if l`
- `i32.and 0x80000000; eqz; br_if l` → `i32.clz; br_if l`
- i64 mirrors with `wrap_i64` (EOP path) and the `i64.eqz`-route variants
- Drops the dead `i64.and 1; if` rule (#6103's commit message claimed removal; diff didn't)

### LSB-via-`shl (N-1)` family (`cf729f046`)

- `i32.shl 31; if t e` → `i32.ctz; if e t` — fires from `compile_classical.ml`'s `pow` body
- `i32.shl 31; eqz; br_if l` → `i32.ctz; br_if l`
- `i64.shl 63; i64.eqz; if t e` → `i64.ctz; wrap; if t e` — fires from `compile_enhanced.ml`'s `pow` body (Word64 + TaggedSmallWord variants)
- `i64.shl 63; i64.eqz; br_if l` → `i64.ctz; wrap; br_if l`

Originally landed with `failwith` probes; the probes fired immediately on `aardvark.mo` (i64) and `assign-ops.mo` (i32), pinning the source sites to each backend's `pow` square-and-multiply LSB-test. Comments now name those exact emit sites.

### `i64.load*; i32.wrap_i64` → `i32.load*` (`c60e99159`)

Wasm is little-endian, so `i64.load addr` reads bytes `[addr, addr+8)` with the low 32 bits at `[addr, addr+4)` — exactly what `i32.load addr` would load. Three cases:

- `i64.load` (`sz=None`) → `i32.load`, `align` clamped to `≤ 2`.
- `i64.load32_{u,s}` → `i32.load` (Pack32 + i32 is degenerate).
- `i64.load{8,16}_{u,s}` → `i32.load{8,16}_{u,s}` (`sz` preserved).

Packed variants are fully spec-preserving (same byte count read either way; `wrap` just discards the i64 extension). The full `i64.load → i32.load` narrows the read from 8 to 4 bytes — same OOM caveat as the existing `Load; Drop → drop` rule.

**Impact** — EOP corpus scan over `test/run/_out/` (268 wasm files):

| Stat | Before | After |
|---|---|---|
| Full `i64.load; i32.wrap_i64` instances | 201 (in 12 files) | **0** |
| Top offenders | `candid.wasm` 53, `idl-opt-tests.wasm` 31, `idl-decoding-bug.wasm` 23, `idl-extra-variant-bug.wasm` 22, `idl.wasm` / `idl-ops.wasm` / `issue-5606.wasm` 16 each, … | — |

The ~3 packed-load+wrap per file that remain are RTS-side (Rust toolchain output, linked verbatim) and outside the OCaml peephole's reach. CI's auto-bench captured a small cycle win on `typtbl.drun-run.ok` (−2 cycles uniformly across 5 measurements) attributable to this rule.

### `i64.and (-1<<n); i64.eqz` → `i64.shrU n; i64.eqz` (`30426ab0a`)

When the constant is a "high mask" (`-1 << n`, n ∈ [1, 63]), `eqz` tests "are bits `[n, 63]` all zero?". `i64.shrU n` brings those same bits down to positions `[0, 63−n]` and `eqz` asks the identical question. Same predicate, fewer bytes — sleb128 of small `n` is 1 byte; sleb128 of `-1 << n` is up to 9.

The detector `c ≠ 0 ∧ c ≠ -1 ∧ ctz(c) + clz(~c) = 64` (via `Wasm.I64.{ctz, clz}`) is the compact characterisation of a contiguous high mask. The `c ∈ {0, -1}` edge cases both pass the bare formula (Wasm spec: `ctz(0) = clz(0) = 64`) but yield meaningless `n`; they're ruled out explicitly.

**Placement** matters: must come *after* the existing MSB `and (1<<63); eqz; (if|br_if) → clz; wrap; (if|br_if)` rules. For `n = 63` in a branch context the MSB rule is strictly better (produces `clz` directly, no shift). After all the MSB rules fail to match, this rule catches the remaining high-mask-and-eqz shapes.

**Impact** (EOP build, selected files):

| File | hits |
|---|---|
| `candid.wasm` | 18 |
| `words-enhanced.wasm` | 7 |
| `issue-5606.wasm` | 5 |
| `assign-ops.wasm` | 4 |

Each hit saves ~7 bytes of code and one mask materialisation. Fires on every `enforce_unsigned_bits env n` in `compile_enhanced.ml`'s small-word arithmetic kernels (Nat8/Int8/Nat16/Int16/Nat32/Int32 wrapping arithmetic + overflow check).

### Tests (`cf729f046`, `e8d8b80e2`, `5265cd41e`)

- `test/run/top-bit.mo` — classical MSB i32 (Nat32/Int32) → `clz; if`
- `test/run/top-bit-i64.mo` — EOP MSB i64 (Nat64/Int64) → `clz; wrap; if`
- `test/run/shl-lsb.mo` — classical LSB-via-shl i32 in `wpow_nat<Nat32>` → `ctz; if`
- `test/run/shl-lsb-i64.mo` — EOP LSB-via-shl i64 in `wpow_nat<Nat64>` → `ctz; wrap; if`

The FileCheck-gate fix in `5265cd41e` (= #6116) is what makes those `// CHECK` directives actually enforce.

## Test plan

- [x] `make -C test/run -j8 quick` green.
- [x] `EXTRA_MOC_ARGS="--enhanced-orthogonal-persistence" make -C test/run top-bit-i64.only shl-lsb-i64.only` green.
- [x] Manually verified each `~~>` rewrite in disassembled wat (`wasm2wat`) before adding it.
- [x] All four new tests' `// CHECK` lines pass FileCheck after the gate fix lands.
- [ ] Run-eop / drun / mo-idl on CI.


ggreif:
## Asymmetric coverage = codegen lever, not just gap

While investigating whether moc handles `i64.and 1; i64.eqz; br_if` correctly, I noticed the i64-and-1 rule matrix here covers some shapes but not others. The gaps are worth recording as **design hints**, not just as TODOs.

### Matrix today

| Pattern (source order) | Direction | Covered |
|---|---|---|
| `i64.and 1; wrap_i64; if T E` | T = LSB **set** | ✅ line 150 → `ctz; wrap; if E T` (legs swap) |
| `i64.and 1; eqz; br_if L` | LSB **clear** | ✅ line 162 → `ctz; wrap; br_if` |
| `i64.and 1; wrap_i64; i32.eqz; br_if L` | LSB **clear** (wrap-route) | ✅ line 165 → `ctz; wrap; br_if` |
| `i64.and 1; wrap_i64; br_if L` | LSB **set** (direct br_if) | ❌ **GAP 1** |
| `i64.and 1; i64.eqz; if T E` | LSB **clear** (if-form) | ❌ **GAP 2** |
| `i64.and 1; wrap_i64; i32.eqz; if T E` | LSB **clear** (wrap+eqz if) | ❌ **GAP 3** |

### Why this matters beyond "implement them too"

For any 2-variant discriminator whose variants differ in LSB (`Result.#ok / .#err`, `Option.?v / null` if `null` is LSB-tagged, …) we have *freedom* in which side to test: testing for `#err` vs testing for `#ok` produce the same control flow, just with opposite branch directions.

So when moc has that choice, it should pick **the direction the peephole covers** — currently LSB-clear via eqz. Then the post-peephole shape is `i64.ctz; i32.wrap_i64; br_if`, which is also exactly what cranelift v45's #13343-style lowering recognises for `test x, 1; jz`. **One design choice buys both the moc-level rewrite AND the cranelift-level machine-code collapse.**

Conversely, if moc ends up emitting (say) `i64.and 1; wrap_i64; br_if L` for the LSB-set direction, it stays as 3 wasm instructions AND cranelift has nothing to recognise — falls back to `and 1; test; jne`. Same instruction count at the wasm level, worse machine code.

### Practical convention

The asymmetric matrix above suggests moc should consistently:

1. **Assign tag 0 to the cold / failure / less-common arm** (`#err = 0`, `null = 0`, …) so the natural "is it the cold arm?" test is **LSB-clear**.
2. **Emit LSB-clear-tests via `eqz; br_if`** (covered) rather than LSB-set-tests via `wrap; br_if` (not covered).
3. **Prefer `if T E` only when the legs-swap rule fires** (i.e., the `wrap; if` family at line 150) — which is the LSB-**set** direction; this is symmetric to the br_if conventions because `if T E` *can* swap.

### So: should we add the missing rules?

I'd argue **no, not yet** — at least not before adding `failwith` probes (the same technique that surfaced the original peephole sites in `compile_enhanced.ml`'s `pow` body) to confirm whether moc actually emits any of these gap shapes today. If GAP 1/2/3 never fire, they're defensive rules without tests; the missing-rule matrix is more useful as documentation than as code.

A FYI for reviewers and future me. Cross-ref my comment thread with MG on binaryen#8562 (the i32 vs runtime-bias debate) — the same principle applies here on the moc side: prefer the shape the downstream toolchain already handles cleanly.

MaxGraey:
Hey @ggreif could you teach your Claude's agent (via CLAUDE.md I guess) doesn't ping me, especially regarding repositories and PRs that I have nothing to do with? Thanks!

ggreif:
> Hey @ggreif could you teach your Claude's agent (via CLAUDE.md I guess) doesn't ping me, especially regarding repositories and PRs that I have nothing to do with? Thanks!

Sorry for that! Won't happen again!

PR #6119, opened 2026-05-14
experiment: `await*` on `public` self-actor methods (worker/wrapper split)

ggreif:
## Summary

Permits `await*` on a call to a `public` method of the enclosing
actor/actor-class that returns `async T` (not `async*`), with the goal of
eliding IC message dispatch on self-calls. Three syntactic call shapes
are recognised by the typechecker; lowering produces an
optimisation-appropriate IR for each.

| call shape | classification | lowering |
|---|---|---|
| `await* foo()` where `foo` is in `env.public_methods` | Obvious self (`VarE`) | direct `await* foo*(args)` |
| `await* self.foo()` where `self = env.self_id` | Obvious self (`DotE-self`) | direct `await* foo*(args)` |
| `await* e.foo()` where `e : ActorT`, `e <> self` | Maybe self | `if thisActor == principalOfActor e then await* foo*(args) else await e.foo(args)` |
| `await* abbrev()` (let-bound) or `await* m()` (computed callee) | reject | M0088 |
| `await* localHelper()` (private async) | reject | M0088 |

## Mechanism

**Typing (`mo_frontend/typing.ml`).** Extends `env` with
`self_id : string option` and `public_methods : S.t`, populated at
`infer_obj` entry for `Actor`/`Mixin`. The `AwaitE` rule classifies the
inner `CallE`'s callee syntactically; the test was settled on after
discovering that two instances of an `actor class` share method-type
identity, so any type-only physical-equality check would conflate
`self.foo` with `other.foo`.

**Lowering (`lowering/desugar.ml`).** For each actor body:
- `collect_obvious_self_demand` walks the dec_fields via
  `Mo_frontend.Traversals.over_dec_field`, collecting the set of public
  method names that are `await*`'d somewhere (Obvious or Maybe self).
- `split_demanded_method` synthesises, per demanded method, a private
  `foo*` worker (`Local`, `Returns`, body wrapped in `AsyncE Cmp`) and a
  modified wrapper whose body becomes `async { await* foo*(args) }`.
- The `S.AwaitE` handler consults `current_actor_context` and the
  callee's syntactic shape:
  - Obvious self -> emit `await* foo*(args)` directly, instantiating the
    worker's outer scope binder with the current `AsyncE`'s scope.
  - Maybe self -> let-bind the receiver and emit
    `if selfRefE Principal == principalOfActor recv then await* foo*(args)
    else await recv.foo(args)`. Slow path uses `ActorDotPrim` (not
    `DotPrim`) since the receiver has actor sort.
- AST interpreter is made polymorphic on the runtime value:
  `AwaitE (AwaitCmp, ...)` falls back from `V.as_comp` to the regular
  `await env ... false ...` when the value is an `Async`, so the
  interpreter runs the same source program without needing the IR-level
  worker split.

**Minor.** `Construct.primE`'s type table gains
`OtherPrim "principalOfActor" -> T.principal`. Both codegen backends
already implement the primitive.

## Tests

- `test/run-drun/await-self-worker.mo` — singleton actor exercising
  Obvious-self via unqualified `VarE`. Full pipeline (with `//OR-CALL`):
  `[tc][run][run-ir][run-low][comp][valid][drun-run]`.
- `test/run-drun/await-self-worker-class.mo` — `actor class A(prior :
  ?A) = self`; two instances created at top level; `go()` exercises all
  three classifier branches with a 3-way add (`42+42+21=105`) that
  encodes closure-capture of `prior` through the worker. `[comp]` is
  `SKIPped` because multi-instance creation at top level would need a
  separate driver-actor + helper-file split; orthogonal to the feature.
- `test/fail/await-self-worker-bad.mo` — three M0088 rejections:
  private/local async helper (sort = Local, no canonical match);
  let-bound async value (inner shape is `VarE`, not `CallE`); direct
  `await*` on the local helper.

Regression-checks `test/run-drun/do-async.mo` and
`test/fail/bad-async-sort.mo` stay green.

## Out of scope

- When the [Candid opt-out](https://github.com/caffeinelabs/motoko/pull/5996) lands, `Maybe_self` should tighten to reject
  (the slow path can't marshal). Filed as a follow-up.
- Multi-file `[comp]`-passing version of the class test (driver actor
  importing the actor class) — purely test-structural.

## Commits

Ten WIP commits, intended to be squashed or rebased into a single feature commit
before merge.

## Test plan
- [x] `nix develop --command make -C src`
- [x] `test/run-drun/await-self-worker.{only}` — full pipeline including `[drun-run]`
- [x] `test/run-drun/await-self-worker-class.only` — `[tc][run][run-ir][run-low]`
- [x] `test/fail/await-self-worker-bad.only` — three M0088 rejections
- [x] `test/run-drun/do-async.only` — full-pipeline regression
- [x] `test/fail/bad-async-sort.only` — pre-existing M0088 cases

### DIFF
--- a/src/ir_def/construct.ml
+++ b/src/ir_def/construct.ml
@@ -136,6 +136,7 @@ let primE prim es =
     | OtherPrim "env_var" -> T.text
     | OtherPrim "caller_info_signer" -> T.blob
     | OtherPrim "caller_info_data" -> T.blob
+    | OtherPrim "principalOfActor" -> T.principal
     | OtherPrim "set_migrations" -> T.unit
     | OtherPrim "get_migrations" -> T.text_list
     | _ -> assert false (* implement more as needed *)
--- a/src/lowering/desugar.ml
+++ b/src/lowering/desugar.ml
@@ -9,7 +9,45 @@ open Operator
 module S = Syntax
 module I = Ir
 module T = Type
+module Traversals = Mo_frontend.Traversals
 open Construct
+
+(* String set used for collecting the per-actor "worker demand": names of
+   public methods that are `await*`'d as Obvious-self call sites somewhere
+   in the actor body. Built by `collect_obvious_self_demand`; consumed by
+   the wrapper/worker split in `build_actor` / `build_obj`. *)
+module StringSet = Set.Make (String)
+
+(* Walk the surface dec_fields of an actor body and collect the names of
+   public methods that appear as the target of an Obvious-self `await*`
+   (matching the same syntactic classifier as `typing.ml`'s AwaitE rule).
+
+   Quadratic in the number of methods (each AwaitE call site checks
+   membership in `public_methods`), but N is small in practice. *)
+let collect_obvious_self_demand
+    (self_id_opt : string option)
+    (public_methods : StringSet.t)
+    (dec_fields : S.dec_field list)
+    : StringSet.t =
+  let demand = ref StringSet.empty in
+  let visit (e : S.exp) : S.exp =
+    (match e.it with
+     | S.AwaitE (T.AwaitCmp, e1) ->
+        (match e1.it with
+         | S.CallE (_, callee, _, _) ->
+            (match callee.it with
+             | S.VarE id when StringSet.mem id.it public_methods ->
+                demand := StringSet.add id.it !demand
+             | S.DotE ({ it = S.VarE recv; _ }, mname, _)
+                when self_id_opt = Some recv.it ->
+                demand := StringSet.add mname.it !demand
+             | _ -> ())
+         | _ -> ())
+     | _ -> ());
+    e
+  in
+  List.iter (fun df -> ignore (Traversals.over_dec_field visit df)) dec_fields;
+  !demand
 
 (*
 As a first scaffolding, we translate imported files into let-bound
@@ -765,6 +803,20 @@ and build_stabs (df : S.dec_field) : stab option list = match df.it.S.dec.it wit
 
 and build_actor at chain ts (exp_opt : Ir.exp option) self_id es obj_typ0 =
   let fs0 = build_fields obj_typ0 in
+  (* Worker/wrapper-split scaffolding: identify which public methods are
+     `await*`'d as Obvious-self call sites somewhere in this actor body.
+     Currently collected but not consumed — the actual IR transformation
+     (emit private `foo*`, replace wrapper body with `async { await* foo*(...) }`)
+     is a follow-up. The AST-interpreter polymorphic `await*` and the
+     desugar fallback (`AwaitCmp` on `async _ T` → `AwaitFut false`) keep
+     soundness while the optimisation is offline. *)
+  let _demand =
+    let self_id_opt = Option.map (fun (i : S.id) -> i.it) self_id in
+    let (_, tfs0, _) = T.as_obj' obj_typ0 in
+    let public_methods =
+      List.fold_left (fun s T.{lab; _} -> StringSet.add lab s) StringSet.empty tfs0 in
+    collect_obvious_self_demand self_id_opt public_methods es
+  in
   let stabs = List.concat_map build_stabs es in
   let ds = decs (List.map (fun ef -> ef.it.S.dec) es) in
   let pairs = List.map2 stabilize stabs ds in
--- a/src/lowering/desugar.ml
+++ b/src/lowering/desugar.ml
@@ -29,8 +29,14 @@ let current_actor_context : (StringSet.t * string option) option ref = ref None
 (* ----- helpers used by the self-actor worker split -------------------- *)
 
 (* Walk the surface dec_fields of an actor body and collect the names of
-   public methods that appear as the target of an Obvious-self `await*`
-   (matching the same syntactic classifier as `typing.ml`'s AwaitE rule).
+   public methods that need a worker emitted because they appear as the
+   target of an `await*` somewhere in the body. The classifier matches:
+
+   - Obvious self via unqualified `VarE id` with `id ∈ public_methods`;
+   - Obvious self via `DotE (VarE recv, _, _)` with `recv = self_id`;
+   - Maybe-self via `DotE (recv, _, _)` where `recv` has actor type and
+     the method name is in `public_methods` (recv may or may not be self
+     at runtime; phase-2 lowering emits a runtime-checked dual path).
 
    Quadratic in the number of methods (each AwaitE call site checks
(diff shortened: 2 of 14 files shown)


ggreif:
**Productisation roadmap (design notes from review)**

The PR ships a working PoC. Three refinements are needed before this stands up to production scrutiny; capturing them here so the design isn't lost.

### 1. Refine `Shared` with `by_whom` provenance — a flat lattice

The current Obvious-self gate compares the receiver's binding name to `env.self_id` as a string. That's brittle to shadowing — `let self = otherActor; await* self.foo()` would (incorrectly) be classified Obvious-self by the syntactic test.

Replace it with a type-level *by-whom* annotation on `Shared` method types, structured as a flat lattice:

```
                Dunno  (top)
                / | \
        Self(srcloc1, Singleton)
        Self(srcloc2, Class)
        ...
```

- **`Self(srcloc, Singleton)`** — value reachable only from a singleton `actor a { ... }` at that srcloc. There is exactly one runtime instance, so direct-worker lowering is sound *without* a runtime check.
- **`Self(srcloc, Class)`** — reachable from an `actor class A` at that srcloc. Multiple instances possible; lowering still needs the runtime `principalOfActor` check.
- **`Dunno`** — let-aliased without provenance, function-arg boundary, imported, etc. No optimisation; plain `await`.

Unification: `x | x = x`, `x | y = Dunno` for distinct lattice points, `Dunno` absorbing. Subtype-widening at function-arg signatures naturally drops to `Dunno`, which is the desired refinement-loss behaviour.

Mechanically, fits in the polymorphic `'a` payload of `func_sort = (shared_sort * by_whom) shared`, leaving the `shared_pat` instantiation untouched.

The Obvious-self test then becomes purely lattice-driven: `callee.note.typ` has `by_whom = Self(s, _)` and the enclosing scope is `Self(s, _)` for the same `s`. No string compares, no shadowing pitfall.

### 2. Algebraic effects for "what scope am I in" queries

The PoC threads `self_id` / `public_methods` through `typing.ml`'s `env`, and uses a module-level `current_actor_context : ... ref` in `desugar.ml` for the same purpose at lowering. Both are essentially queries for *context inherited from a surrounding scope* — exactly what OCaml 5.3 algebraic effects model without the threading:

```ocaml
type _ Effect.t +=
  | EnclosingActor : actor_provenance Effect.t
  | PublicMethods  : SS.t Effect.t
```

Handlers installed at actor-body entry; effects performed at the `AwaitE`/`CallE` rules deep inside. Nested actors compose for free (inner handler shadows outer). No `env` fields, no save/restore boilerplate, no module-level mutable ref. Consistent with the effects style already used in the variant-switch dispatch experiment.

### Interaction (1) + (2)

The two are complementary:

| concern | mechanism |
|---|---|
| *what is this value's provenance* | `by_whom` payload on `Shared` (type-level, flows with the value) |
| *what scope is currently being typed/lowered* | algebraic effect (control-level, queried at AwaitE / DotE / VarE) |

### 3. AST-level worker split, not IR-level

`split_demanded_method` currently lives in `desugar.ml` and rewrites IR after the dec_fields have been lowered. That was pragmatic — the `Construct` module's typed builders (`funcE`, `asyncE`, `awaitE`, `callE`) made note-construction painless, and I avoided hand-rolling typed surface-AST nodes with their `S.typ_note` fields.

For production this should move to a *pre-desugar* pass over the typed surface AST, conceptually:

```
let method = func (<type-part>, <body-part>)
becomes
let method' = func (<async*-type-part>, <body-part>)
let method  = func (<type-part>,        <forwarding-body-derived-from-type>)
```

Reasons:
- Plays nicely with the "pre-desugaring run" hook we already need for generation-ordering resolution.
- The forwarding body is just the natural typed AST one would write by hand once `Shared (_, by_whom)` exists — no special IR-builder dance.
- Lets the standard desugar handle the resulting `LetD`s with no IR-level surgery.
- Eliminates the `current_actor_context` ref entirely (its only consumer is the IR-level retargeting at AwaitE).

---

For now the PoC is sufficient and gives a good impression of what the feature feels like end-to-end. Once management buy-in lands, the production push is the three items above.

ggreif:
**Edit / addendum** — the lattice + dynamic-check angle has a nice symmetric story:

Shared method values cross actor boundaries through Candid (e.g. as a `shared func` argument to an `update`). The Candid wire encoding is just `(principal, method_name)`; any compile-time provenance (`by_whom` from §1) is necessarily stripped on the way in. So values received over Candid type as `Shared (_, Dunno)` on the receiving side.

`await*` on a `Dunno`-provenance shared-func value would still be *unlocked* — and the lowering becomes uniform with the Maybe-self case from §2 of the PR body, just generalised:

```
let r = received_func;
let p = principalOfActor r in            -- comes off the (principal, method_name) wire repr
let m = methodNameOf r in                -- ditto
if p == thisActor && m == "foo"   then await* foo*(args)
elif p == thisActor && m == "bar" then await* bar*(args)
...                                       -- one branch per public method
else await r(args)                       -- slow path: marshal & dispatch normally
```

i.e. a dynamic `(principal, method_name)` dispatch against the actor's static `public_methods` set. When the received value happens to be a method on `self`, the static workers handle it without a round-trip; otherwise the original message-send semantics are preserved.

This generalises naturally:
- §1's `Self(srcloc, Singleton)` ⇒ no runtime check at all (statically known unique instance).
- §1's `Self(srcloc, Class)`     ⇒ principal-only check (method name is statically known from the call shape).
- `Dunno`                        ⇒ `(principal, method_name)` check against the public-method table.

The check itself stays cheap (`Blob` equality on principal, `Text` equality on method name) and the dispatch table is bounded by the actor's interface. So the same conceptual framework — "fast-path when we recognise ourselves" — extends to opaque shared func values without losing the slow-path soundness guarantee.

**Static rejection falls out for free.** The lattice + dynamic-dispatch framing makes one more observation crisp: `await*` on a `Dunno`-provenance shared-func value can be *statically rejected* at typing when the enclosing actor exposes no method (or `Shared` lambda) whose type matches the callee's signature. The fast-path dispatch table is then guaranteed empty, so no worker can ever exist for that type — even runtime can't recover. The check is purely type-level: lookup the callee's `func_sort` shape (args ⊢ async T) in the enclosing actor's public-method table, reject if none matches.

This sharpens the contract: `await*` on `Dunno` is *well-typed iff* some method of ours could serve as a fast-path target. The runtime check just picks among them; if there's nothing to pick from, the call shouldn't have type-checked in the first place. Users wanting plain message dispatch on an arbitrary shared func keep using `await` — `await*` becomes a static declaration of intent: "I expect this could be me."

PR #6121, opened 2026-05-15
perf: smart `immut_local` constructor for switch scrutinees

ggreif:
## Summary

Implements the `immut_local` smart constructor proposed in commit `08ef20b77` (plan section of `gabor/variant-switch`'s `.claude/plans/variant-switch-br-table.md`) and threads it through the pattern-matching emission paths on **both** the enhanced and classical backends.

`immut_local env name init`: if `init` materialises to a single pure-read instruction (`LocalGet` / `GlobalGet` / `Const`), no local is allocated — `init` is re-emitted verbatim at each use. Otherwise falls back to the `new_local` + `set/get` shape. An algebraic effect (`InstrList.Inspecting`) routed through a new `G.effectful` marker on `remember_depth` / `branch_to_` guards the inspection against unfulfilled depth-promises (e.g. scrutinees containing `break` out of a `do ?` block).

Refactor surface:
- `fill_pat` now takes `init : G.t` instead of consuming the stack value; per-arm `*_scrut` bindings (TagP / OptP / TupP / ObjP / AltP / AndP) go through `immut_local`.
- `compile_pat_local` threads `init` through.
- `compile_unboxed_pat` passes `G.nop` as `init` to preserve the stack convention for let-patterns.
- SwitchE single_case + general use `immut_local "switch_in"` and pass the resulting (always-pure) getter as `init`.
- Side fix: `GlobalGet ^^ Drop` added to the existing `Const|LocalGet ^^ Drop` peephole for symmetry on the irrefutable-nonbinding fallback.

## Wasm before/after — 5-arm variant switch with binding patterns

```motoko
type E = { #r : Nat; #g : Nat; #b : Nat; #y : Nat; #p : Nat };
func payload(c : E) : Nat =
  switch c {
    case (#r x) x; case (#g x) x + 1; case (#b x) x + 2;
    case (#y x) x + 3; case (#p x) x + 4;
  };
```

| | locals declared | function body lines |
|---|---:|---:|
| **before** | `$x, $tag_scrut, $x.1, $tag_scrut.1, $x.2, $tag_scrut.2, $x.3, $tag_scrut.3, $x.4, $tag_scrut.4, $x.5` (**11**) | 143 |
| **after** | `$x, $x.1, $x.2, $x.3, $x.4, $x.5` (**6**) | 138 |

Per arm: `local.tee $tag_scrut.N` elided, `local.get $tag_scrut.N` → `local.get $c`. **6 locals + 7 instructions saved per such switch site.**

Single-arm `switch (#only x)` and switches over non-pure scrutinees are unchanged (TagP("",_) fast-path + materialise fallback respectively).

## Bench delta (`test/bench/variant-switch.mo`, full numbers in commit `9148c4c`)

| metric | Δ instr | Δ % |
|---|---:|---:|
| `isWeekendOr` (or-pattern, 5 cases) | −740,000 | **−6.70%** |
| AST `eval` loop ×10k | −4,070,000 | **−3.0%** |
| `fib7_eval` | −453,300 | −1.85% |
| `isWeekend` (7 arms) | −140,000 | −1.40% |

Follow-up tracked in #6124: narrowing the variant-hash compare from `i64.eq` to `i32.eq` on the enhanced backend (classical already does i32).

## Test plan

- [x] `test/run` quick — green on both classical (default) and `--enhanced-orthogonal-persistence`
- [x] Bench-updater commit `9148c4c` — wins above
- [x] Full CI


src/codegen/compile_enhanced.ml
```diff
@@ -13571,16 +13575,20 @@ and compile_unboxed_pat env ae how pat
...
     | VarP name ->
       let pre_code, sr, code = Var.set_val env ae1 name in
       pre_code, Some sr, code
     (* The general case: Create a single value, match that. *)
     | _ ->
+      let set_x, get_x = new_local env "scrut" in
       G.nop,
       Some SR.Vanilla,
-      orPatternFailure env (fill_pat env ae1 pat) in
+      set_x ^^ orPatternFailure env (fill_pat env ae1 get_x pat) in
```
ggreif: This looks like a pessimisation. Can we have the old version of `fill_pat` back as `fill_pat'` that doesn't need the `init`, but receives on the stack?

ggreif: Or maybe pass `G.nop` to `init` if that results in just popping the stack?

ggreif: I would be okay with an `~init:` argument that is optional.

src/codegen/compile_enhanced.ml
```diff
@@ -13571,16 +13575,20 @@ and compile_unboxed_pat env ae how pat
       (* We have to fill the pattern in reverse order, to take things off the
          stack. This is only ok as long as patterns have no side effects.
       *)
-      G.concat_mapi (fun i p -> orPatternFailure env (fill_pat env ae1 p)) (List.rev ps)
+      G.concat_mapi (fun _ p ->
+        let set_x, get_x = new_local env "tup_unboxed" in
```
ggreif: See below!

test/bench/ok/variant-switch.drun-run.ok
```diff
@@ -1,9 +1,9 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
-debug.print: {heap_diff = 0; instr_diff = 137_580_319; total = 1_090_000}
+debug.print: {heap_diff = 0; instr_diff = 133_510_319; total = 1_090_000}
 ingress Completed: Reply: 0x4449444c0000
-debug.print: {fib7_eval = 13; fib7_evalFT = 13; fib7_xform = 13; instr_eval = 24_493_446; instr_evalFT = 21_520_346; instr_transform = 1_187_846}
+debug.print: {fib7_eval = 13; fib7_evalFT = 13; fib7_xform = 13; instr_eval = 24_040_146; instr_evalFT = 21_384_946; instr_transform = 1_171_546}
 ingress Completed: Reply: 0x4449444c0000
-debug.print: {acc1 = 20_000; acc2 = 20_000; instr_isWeekend = 10_000_319; instr_isWeekendOr = 11_040_319}
+debug.print: {acc1 = 20_000; acc2 = 20_000; instr_isWeekend = 9_860_319; instr_isWeekendOr = 10_300_319}
```
ggreif: Bench wins from this update (relative to master):

| Bench | Δ instr | Δ % |
|---|---:|---:|
| `instr_isWeekendOr` (or-pattern, 5 cases) | −740,000 | **−6.70%** |
| AST `eval` loop ×10k (`instr_diff`) | −4,070,000 | **−3.0%** |
| `instr_eval` (fib7) | −453,300 | **−1.85%** |

The or-pattern wins biggest in relative terms because each arm allocates a `tag_scrut` under the old shape — eliding all of them per dispatch compounds. The AST eval loop has the biggest absolute saving (4M cycles).

src/codegen/compile_enhanced.ml
```diff
@@ -13433,87 +13460,87 @@ and compile_lit_pat env l =
...
+      orElse code1 code2
   | AndP (p1, p2) ->
-      let code1, code2 = fill_pat env ae p1, fill_pat env ae p2 in
-      let set_i, get_i = new_local env "and_scrut" in
-      CannotFail set_i ^^^
-      (CannotFail get_i ^^^ code1) ^^^
-      (CannotFail get_i ^^^ code2)
+      let scr_prelude, get_i = immut_local env "and_scrut" init in
+      let code1 = fill_pat env ae get_i p1 in
+      let code2 = fill_pat env ae get_i p2 in
```
ggreif: let code1, code2 = ...

ggreif: Above too! Line 13535.

src/codegen/compile_enhanced.ml
```diff
@@ -13433,87 +13460,87 @@ and compile_lit_pat env l =
...
+          ( with_fail fail_code
+              (fill_pat env ae (get_x ^^ Opt.project env p.note) p)
           )
           fail_code
       )
   | TagP ("", p) -> (* these only come from known_tag_pat *)
     if Ir_utils.is_irrefutable_nonbinding p
-    then CannotFail (G.i Drop)
-    else CannotFail (Variant.project env) ^^^ fill_pat env ae p
+    then CannotFail (init ^^ G.i Drop)
```
ggreif, line 13470: can you verify that peephole will clean this for the optimal case?

src/codegen/compile_enhanced.ml
```diff
@@ -13209,19 +13237,19 @@ and compile_exp_with_hint (env : E.t) ae sr_hint exp =
...
 
   | SwitchE (e, cs) ->
     let code1 = compile_exp_vanilla env ae e in
-    let (set_i, get_i) = new_local env "switch_in" in
+    let scr_prelude, get_i = immut_local env "switch_in" code1 in
 
     (* compile subexpressions and collect the provided stack reps *)
     let codes = List.map (fun {it={pat; exp=e}; _} ->
-      let (ae1, pat_code) = compile_pat_local env ae pat in
+      let (ae1, pat_code) = compile_pat_local env ae get_i pat in
```
ggreif: parens redundant

src/codegen/compile_enhanced.ml
```diff
@@ -932,6 +932,34 @@ let new_local32 env name =
...
+let immut_local env name (init : G.t) : G.t * G.t =
+  let open Wasm.Source in
+  let exception Bail in
+  let inspect () =
+    let open Effect.Deep in
+    try_with
+      (fun () -> init 0l no_region [])
+      ()
+      { effc = fun (type a) -> function
+          | (InstrList.Inspecting : a Effect.t) ->
```
ggreif: parens redundant?

ggreif: Required, sadly — OCaml's `function | pat -> …` parses `pat` as a regular pattern; a type annotation on a constructor pattern needs `(C : T)` parens or the colon is mis-parsed. Without them this is a syntax error.

The annotation is *load-bearing* though: it's the bridge between the locally-abstract type `a` introduced by `fun (type a)` and the GADT-like return type the `effc` field expects. Drop it and OCaml can't refine `Some (… : (a, _) continuation)` to a concrete continuation type.

ggreif: Thanks, Claude!

src/codegen/compile_classical.ml
```diff
@@ -12855,50 +12883,44 @@ and compile_exp_with_hint (env : E.t) ae sr_hint exp =
   | SwitchE (e, cs) when single_case e cs ->
     let code1 = compile_exp_vanilla env ae e in
     let [@warning "-8"] [{it={pat={it=TagP (_, pat');_} as pat; exp}; _}] = cs in
-    let ae1, pat_code = compile_pat_local env ae {pat with it = known_tag_pat pat'} in
+    let ae1, pat_code = compile_pat_local env ae code1 {pat with it = known_tag_pat pat'} in
     let sr, rhs_code = compile_exp_with_hint env ae1 sr_hint exp in
 
-    (* Use the expected stackrep, if given, else infer from the branches *)
```
ggreif, line 12861: Don't remove pre-existing comments.

src/codegen/compile_classical.ml
```diff
@@ -12855,50 +12883,44 @@ and compile_exp_with_hint (env : E.t) ae sr_hint exp =
...
     let final_sr = match sr_hint with
       | Some sr -> sr
       | None -> sr
     in
 
     final_sr,
-    (* Run rest in block to exit from *)
```
ggreif, line 12868: same

src/codegen/compile_classical.ml
```diff
@@ -12855,50 +12883,44 @@ and compile_exp_with_hint (env : E.t) ae sr_hint exp =
...
+       ) [sr, pat_code ^^^ CannotFail rhs_code]) ^^
+       G.i Unreachable
     )
 
   | SwitchE (e, cs) ->
     let code1 = compile_exp_vanilla env ae e in
-    let (set_i, get_i) = new_local env "switch_in" in
+    let scr_prelude, get_i = immut_local env "switch_in" code1 in
 
-    (* compile subexpressions and collect the provided stack reps *)
```
ggreif, line 12880: same

src/codegen/compile_classical.ml
```diff
@@ -12855,50 +12883,44 @@ and compile_exp_with_hint (env : E.t) ae sr_hint exp =
...
     let codes = List.map (fun {it={pat; exp=e}; _} ->
-      let (ae1, pat_code) = compile_pat_local env ae pat in
-      let (sr, rhs_code) = compile_exp_with_hint env ae1 sr_hint e in
-      (sr, CannotFail get_i ^^^ pat_code ^^^ CannotFail rhs_code)
+      let ae1, pat_code = compile_pat_local env ae get_i pat in
+      let sr, rhs_code = compile_exp_with_hint env ae1 sr_hint e in
+      sr, pat_code ^^^ CannotFail rhs_code
       ) (simplify_cases e cs) in
 
-    (* Use the expected stackrep, if given, else infer from the branches *)
```
ggreif, line 12887: and here

src/codegen/compile_classical.ml
```diff
@@ -12855,50 +12883,44 @@ and compile_exp_with_hint (env : E.t) ae sr_hint exp =
...
     final_sr,
-    (* Run scrut *)
-    code1 ^^ set_i ^^
-    (* Run rest in block to exit from *)
+    scr_prelude ^^
     FakeMultiVal.block_ env (StackRep.to_block_type env final_sr) (fun branch_code ->
        orsPatternFailure env (List.map (fun (sr, c) ->
           c ^^^ CannotFail (StackRep.adjust env sr final_sr ^^ branch_code)
        ) codes) ^^
-       G.i Unreachable (* We should always exit using the branch_code *)
```
ggreif, line 12901: keep comment

src/codegen/compile_classical.ml
```diff
@@ -13188,18 +13208,9 @@ and alloc_pat env ae how pat : VarEnv.t * G.t  =
     in (ae1, code0 ^^ code1)
   ) d (ae, G.nop)
 
-and compile_pat_local env ae pat : VarEnv.t * patternCode =
```
ggreif, line 13194: keep

src/codegen/compile_enhanced.ml
```diff
@@ -13530,18 +13555,9 @@ and alloc_pat env ae how pat : VarEnv.t * G.t  =
     in (ae1, code0 ^^ code1)
   ) d (ae, G.nop)
 
-and compile_pat_local env ae pat : VarEnv.t * patternCode =
```
ggreif, line 13536: keep

src/codegen/compile_classical.ml
```diff
@@ -12891,9 +12919,7 @@ and compile_exp_with_hint (env : E.t) ae sr_hint exp =
     in
 
     final_sr,
-    (* Run scrut *)
-    code1 ^^ set_i ^^
-    (* Run rest in block to exit from *)
+    scr_prelude ^^
```
ggreif, line 12904:
```suggestion
    (* Run scrut *)
    scr_prelude ^^
```

src/codegen/compile_enhanced.ml
```diff
@@ -13231,8 +13259,7 @@ and compile_exp_with_hint (env : E.t) ae sr_hint exp =
     in
 
     final_sr,
-    (* Run scrut *)
-    code1 ^^ set_i ^^
+    scr_prelude ^^
```
ggreif, line 13244:
```suggestion
    (* Run scrut *)
    scr_prelude ^^
```

PR #6126, opened 2026-05-16
experiment: GADTs in Motoko

ggreif:
Exploratory branch. Adds GADT-style refinement + existential clauses to
variant arms and (in M10) to top-level type definitions, plus an
information-hiding "black-hole" story at the Candid boundary.

Status, design rationale, milestones, surprises, and open pitfalls
all live in [`.claude/plans/GADTs.md`](.claude/plans/GADTs.md) on the
branch. That's the reading order; everything in code is in service
of that document.

Unlaid eggs — please don't pile in expectations yet.

Two very related issues:
- shared generics (reified types) — some planning document included in this PR
- strongly-typed `Principal`, by inspecting the principal's internal type bits and exposing it as a GADT


ggreif:
## Background — a sanity-check conversation

Before committing more time to this experiment I ran the GADT pitch
past Caffeine's assistant in the context of a real OAuth /
TweetMyMoment-style application. The reception was warmer than I
expected, and the back-and-forth surfaced specific patterns where
GADTs would pay rent. Summary below; full Q&A folded inside.

---

### High-value patterns identified

- **Type-safe state machines** — OAuth and similar flows where
  invalid transitions become compile-time errors.
- **Typed query/command** — `Query<T>` always returns `T`,
  mismatches caught before Candid sees them.
- **Validated domain types** — `Token<Validated>` can't be passed
  where a raw `Token<Raw>` is needed.
- **Typed interpreters / mini DSLs** — single correct
  implementation, no runtime type errors.

The OAuth example maps onto a phantom state machine directly:

```mermaid
stateDiagram-v2
  [*] --> Idle
  Idle --> Pending : initiate(verifier)
  Pending --> Authed : complete(token)
  Authed --> [*]
  state "State&lt;Idle&gt;"   as Idle
  state "State&lt;Pending&gt;" as Pending
  state "State&lt;Authed&gt;"  as Authed
```

…with `getVerifier` taking only `State<Pending>` so
`getVerifier(start())` is a type error rather than a runtime crash.

See [below](https://github.com/caffeinelabs/motoko/pull/6126#issuecomment-4466799010) for a more fleshed-out example.

<details><summary>Full Q&A — would backend code with GADTs be useful?</summary>

> GADTs (Generalized Algebraic Data Types) would be a meaningful
> addition to Motoko. From the perspective of generating backend code
> for apps like TweetMyMoment, here's where GADTs would be most
> useful:
>
> **High value areas:**
> - **Type-safe state machines** — OAuth flows (exactly what your
>   app uses) involve strict state transitions
>   (`unauthorized → code_pending → token_stored`). GADTs could
>   encode those transitions so invalid states are compile-time
>   errors, not runtime surprises.
> - **Typed query/command patterns** — expressing that a query of
>   type `Query<T>` always returns `T`, catching mismatches at
>   compile time rather than in Candid serialization.
> - **Validated domain types** — e.g. a `Token` that carries proof
>   it was validated, so you can't accidentally pass an unvalidated
>   token where a validated one is required.
>
> **Lower value areas:**
> - Simple CRUD apps — the added expressiveness wouldn't be exercised
>   much.
> - Frontend-heavy apps where the backend is thin glue — the
>   complexity cost may outweigh the benefit.
>
> **Honest caveat:** the generated code would only benefit if the code
> generator learns to use GADTs idiomatically. Naive use would produce
> noisier code with no safety gain.

</details>

<details><summary>Full Q&A — practical concerns for the implementer</summary>

> Most useful GADT patterns for Motoko backends:
> - **Indexed state machines** — the OAuth case is real. Right now,
>   generated code uses runtime checks to guard against calling
>   `exchangeToken` before `storeVerifier`. A GADT over a state index
>   would make that impossible to express incorrectly.
> - **Heterogeneous collections with type recovery** — canister
>   storage often needs to hold mixed types and recover the type on
>   retrieval. GADTs make this safe without downcasting hacks.
> - **Phantom-type validation witnesses** — marking values as
>   "validated" or "sanitized" at the type level. Useful for anything
>   touching external input (HTTP outcall responses, user-supplied
>   text before it goes into a tweet).
> - **Typed interpreter pattern** — if a canister implements a mini
>   DSL or rule engine, GADTs let the interpreter be written once,
>   correctly, without runtime type errors.
>
> **Practical concerns:**
> - **Candid serialization boundary** — GADTs are great inside the
>   canister, but Candid is not GADT-aware. The boundary needs a
>   clear story.
> - **Type inference interaction** — Motoko's inference is already
>   doing a lot. GADTs with equality constraints can require more
>   annotations. Decide early how much annotation burden is
>   acceptable.
> - **Upgrade safety** — if GADT-typed values end up in stable
>   storage, the upgrade path when the type index changes needs to
>   be well-defined.

</details>

---

### The Candid story (M12 in the plan)

Confirmed acceptable: monomorphise at the boundary, existentials are
rejected outright as "black-hole types", type-table mismatch traps.

```mermaid
flowchart LR
  subgraph "Internal (GADT-rich)"
    A[State&lt;Pending&gt;]
    B[Query&lt;T&gt;]
    C[Token&lt;Validated&gt;]
    D[Tree&lt;A&gt; with ∃B]
  end
  subgraph "Wire (Candid)"
    E[monomorphic Variant]
    F[record / opt]
  end
  A -->|monomorphise| E
  B -->|monomorphise| F
  C -->|monomorphise| F
  D -.->|black-hole<br/>not shareable| X[rejected at is_shared]
  style X fill:#fdd,stroke:#900
  style D fill:#ffd
```

<details><summary>Caffeine's response on the Candid boundary</summary>

> Monomorphisable-only at the Candid boundary is a clean constraint.
> It means GADT-typed values used in canister APIs must be
> instantiated to a concrete type before serialization — the GADT
> expressiveness lives entirely inside the canister, which is
> actually the right place for it.
>
> Existentials being out is the right call for now. They would
> require Candid decoders to handle open type variables, which is
> a much heavier lift and would break most existing tooling.
>
> Net effect: the constraint is workable. The main discipline it
> imposes is a clean separation between:
> - Internal GADT-typed logic (state machines, witnesses, typed
>   interpreters)
> - Candid-facing types that are always fully monomorphic

</details>

<details><summary>On trap-on-type-mismatch as the enforcement model</summary>

> If the type system is strong enough that only a compiler bug
> produces ill-typed wire bytes, then the trap is an internal
> invariant violation, not a user-facing error path. No need to
> engineer around it.
>
> It also means generated code doesn't need any defensive handling
> at that boundary — if it compiles, it's correct. That's a
> considerably cleaner guarantee than most Candid interactions today,
> where the type system and the wire format are only loosely coupled.

</details>

---

### Scope: OCaml-power GADTs without unification

Confirmed sensible as a starting point. The patterns that pay rent
(state machines, witnesses, typed interpreters) all work in this
subset; heterogeneous list/tree traversals that lean on unification
to propagate recovered types outward are out of scope for v1 and
require explicit annotations.

<details><summary>What's kept / what's harder without unification</summary>

> **What you keep without unification:**
> - Phantom types and validation witnesses — full value.
> - Indexed state machines where the index is always known at the
>   call site.
> - Typed interpreters where the return type is determined by the
>   constructor, not inferred across branches.
>
> **What becomes harder:**
> - Heterogeneous list/tree traversals where the element type is
>   recovered by matching — those lean on unification to propagate
>   the recovered type outward. Without it, the user needs explicit
>   annotations at each branch.
>
> For code generation this is fine. Generated code is not written by
> hand, so annotation burden is irrelevant.

</details>

---

## Suggested test corpus

Caffeine offered a complete test corpus in this branch's syntax.
About a third of these are in `test/run/` today; the rest fill in
alongside M11a/b.

```mermaid
flowchart LR
  Nil["List&lt;A, Zero&gt;<br/>(#nil)"]
  L1["List&lt;A, Succ&lt;Zero&gt;&gt;<br/>(#cons)"]
  L2["List&lt;A, Succ&lt;Succ&lt;Zero&gt;&gt;&gt;<br/>(#cons)"]
  Nil -->|prepend| L1
  L1 -->|prepend| L2
  L1 -.->|head &nbsp;✓| A1[A]
  L2 -.->|head &nbsp;✓| A1
  Nil -.->|head &nbsp;✗ ill-typed| X[compile error]
  style X fill:#fdd,stroke:#900
```

<details><summary>gadt-phantom-state.mo — OAuth state machine</summary>

```motoko
type State<S> = {
  #idle    : type S = Idle    in ();
  #pending : type S = Pending in Text;  // code_verifier
  #authed  : type S = Authed  in Text;  // access_token
};

type Idle    = ();
type Pending = ();
type Authed  = ();

func start()                  : State<Idle>    = #idle ();
func initiate(verifier : Text) : State<Pending> = #pending verifier;
func complete(token : Text)    : State<Authed>  = #authed token;

func getVerifier(s : State<Pending>) : Text =
  switch s { case (#pending v) v };

func getToken(s : State<Authed>) : Text =
  switch s { case (#authed t) t };
```

</details>

<details><summary>gadt-equality-witness.mo — Refl</summary>

```motoko
type Equal<A, B> = {
  #refl : type A = B in ();
};

func coerce<A, B>(eq : Equal<A, B>, x : A) : B =
  switch eq { case (#refl ()) x };
```

</details>

<details><summary>gadt-typed-list.mo — length-indexed Vec (committed as <code>test/run/gadt-typed-list.mo</code>)</summary>

```motoko
// Distinct singleton variants — NOT `type Zero = ()` (see fix note below).
type Zero    = { #zero };
type Succ<N> = { #succ : N };

type List<A, N> = {
  #nil  : type N = Zero      in ();
  #cons : type M, type N = Succ<M> in (A, List<A, M>);
};

func head<A, M>(l : List<A, Succ<M>>) : A =
  switch l { case (#cons (x, _)) x };

func tail<A, M>(l : List<A, Succ<M>>) : List<A, M> =
  switch l { case (#cons (_, xs)) xs };

// Reassembly axiom — destructure and re-cons preserves the index.
// The existential `M` in the `#cons` arm is what makes this non-trivial:
// destructuring binds a fresh M for the tail, and re-consing has to pack
// it back into the same `Succ<M>` index as the input.
func reassemble<A, M>(l : List<A, Succ<M>>) : List<A, Succ<M>> =
  switch l { case (#cons (x, xs)) #cons (x, xs) };
```

**Fix notes vs the original sketch:**

1. **`type Zero = ()` collapses the index.** Unit aliases compare structurally equal under `T.eq`, so `Zero` and `Succ<N>` become indistinguishable — refinement can't discriminate the arms. Use distinct singleton variants (`{ #zero }`, `{ #succ : N }`) — the same gotcha bit the OAuth phantom-state example.
2. **Constraint order matters.** The original `type N = Succ<M>, type M` forward-references `M`. Declare `M` first: `type M, type N = Succ<M>`.
3. **Parameterized refinement RHS (`Succ<M>`)** required a new `T.arm_compat` helper (treats arm existentials as wildcards during pruning) — committed alongside on the branch. Without it, `arm_compat (Succ<M_func>) (Succ<M_schema>)` would prune `#cons` away when the inner cons differ.

</details>

<details><summary>gadt-validation-witness.mo — sanitized-input</summary>

```motoko
type Raw       = ();
type Sanitized = ();

type Input<S> = {
  #raw       : type S = Raw       in Text;
  #sanitized : type S = Sanitized in Text;
};

func sanitize(i : Input<Raw>) : Input<Sanitized> =
  switch i { case (#raw t) #sanitized t };  // real impl: strip/escape

func publish(i : Input<Sanitized>) : Text =
  switch i { case (#sanitized t) t };
```

</details>

<details><summary>gadt-mutual-recursive.mo — Expr ↔ Stmt</summary>

```motoko
type Expr<A> = {
  #lit  : type A = Nat  in Nat;
  #flag : type A = Bool in Bool;
  #ifE  : type B        in (Stmt<Bool>, Expr<B>, Expr<B>);
};

type Stmt<A> = {
  #eval   : type A = () in Expr<Nat>;
  #assert : type A = Bool in Expr<Bool>;
  #seq    : type A = () in (Stmt<()>, Stmt<()>);
};
```

</details>

<details><summary>gadt-nested-existential-internal.mo — B hidden inside #node</summary>

```motoko
type Tree<A> = {
  #leaf : type A = Nat          in Nat;
  #node : type A = Bool, type B in (Tree<B>, (B -> Bool), Tree<B>);
};

func anyTrue<A>(t : Tree<A>) : Bool =
  switch t {
    case (#leaf n)         n > 0;
    case (#node (l, _, r)) anyTrue l or anyTrue r;
  };
```

</details>

<details><summary>gadt-stable-roundtrip.mo — internal GADT, Candid-safe projection</summary>

```motoko
type Expr<A> = {
  #int  : type A = Nat  in Nat;
  #bool : type A = Bool in Bool;
};

type FlatExpr = { #int : Nat; #bool : Bool };  // Candid-safe

func flatten<A>(e : Expr<A>) : FlatExpr =
  switch e {
    case (#int n)  #int n;
    case (#bool b) #bool b;
  };

func restoreNat(f : FlatExpr) : ?Expr<Nat> =
  switch f {
    case (#int n)  ?(#int n);
    case (#bool _) null;
  };
```

</details>

<details><summary>Negative tests (should be rejected)</summary>

```motoko
// gadt-neg-wrong-arm.mo — #int produces Expr<Nat>, not Expr<Bool>
let bad : Expr<Bool> = #int 7;
```

```motoko
// gadt-neg-existential-escape.mo — B from #node can't leak out
func badExtract<A>(t : Tree<A>) : Tree<B> =  // B unbound here
  switch t {
    case (#node (l, _)) l;
    case (#leaf n)      #leaf n;
  };
```

```motoko
// gadt-neg-index-mismatch.mo — wrong index across branches
func bad<T>(e : Expr<T>) : Expr<Nat> =
  switch e {
    case (#int n)  #int n;
    case (#bool b) #bool b;  // Expr<Bool> where Expr<Nat> expected
  };
```

</details>

---

## Takeaway

The Candid-boundary story is the load-bearing question, and the
answer (monomorphise / reject existentials / trap on type-table
mismatch) is reception-positive. The scope (OCaml-power without
unification) covers the patterns that actually pay rent for code
generation. The test corpus above is roughly the implementation
target.

ggreif:
## Addendum — OAuth state machine with the refresh-token dance

The two-state `Idle → Pending → Authed` flow in the Caffeine excerpt
is the textbook GADT demo but doesn't reflect what production OAuth
actually does. Real flow has a token-expiry loop and graceful
revocation:

```mermaid
stateDiagram-v2
  [*] --> Idle
  Idle    --> Pending  : initiate(verifier)
  Pending --> Authed   : complete(token, refresh, expiry)
  Authed  --> Expired  : expire()  /* 401 / clock-driven */
  Expired --> Authed   : refreshToken(new, expiry)
  Authed  --> Revoked  : revokeAuthed()
  Expired --> Revoked  : revokeExpired() /* refresh fail */
  Revoked --> Idle     : reset()
  Revoked --> [*]
  state "State&lt;Idle&gt;"    as Idle
  state "State&lt;Pending&gt;" as Pending
  state "State&lt;Authed&gt;"  as Authed
  state "State&lt;Expired&gt;" as Expired
  state "State&lt;Revoked&gt;" as Revoked
```

State payloads differ — `Pending` carries the PKCE verifier, `Authed`
the full `Session`, `Expired` just the refresh token, `Idle` /
`Revoked` nothing. Accessors are signatured at specific states, so
`getToken(s : State<Expired>)` is a compile-time error.

Lives at [`test/run/gadt-phantom-state.mo`](../tree/gabor/gadt/test/run/gadt-phantom-state.mo)
on this branch — passes `[tc] [run] [run-ir] [run-low] [comp]
[valid] [wasm-run]`.

<details><summary>The full implementation</summary>

```motoko
// Distinct singleton-variant tags so refinement can discriminate.
// `type Idle = ()` would alias all states to the unit type and the
// GADT would lose its discriminating power.
type IdleT    = { #idle };
type PendingT = { #pending };
type AuthedT  = { #authed };
type ExpiredT = { #expired };
type RevokedT = { #revoked };

type Session = { token : Text; refresh : Text; expiry : Int };
type Stale   = { refresh : Text };

type State<S> = {
  #idle    : type S = IdleT    in ();
  #pending : type S = PendingT in Text;       // PKCE verifier
  #authed  : type S = AuthedT  in Session;
  #expired : type S = ExpiredT in Stale;
  #revoked : type S = RevokedT in ();
};

func start () : State<IdleT> = #idle ();

func initiate (verifier : Text) : State<PendingT> = #pending verifier;

func complete (token : Text, refresh : Text, expiry : Int) : State<AuthedT> =
  #authed { token; refresh; expiry };

// 401 / clock-driven: keep the refresh token, drop the access token.
func expire (s : State<AuthedT>) : State<ExpiredT> =
  switch s {
    case (#authed sess) #expired { refresh = sess.refresh };
  };

// Refresh-token exchange: trade Stale for a fresh Session.
func refreshToken (s : State<ExpiredT>, newToken : Text, newExpiry : Int)
  : State<AuthedT> =
  switch s {
    case (#expired { refresh }) #authed { token = newToken; refresh; expiry = newExpiry };
  };

// Two revoke entry points — without unification, a single function
// can't accept both Authed and Expired in one signature. The types
// still enforce "you've reached a state where revocation is
// meaningful" (no revoking an Idle session).
func revokeAuthed  (s : State<AuthedT>)  : State<RevokedT> { ignore s; #revoked () };
func revokeExpired (s : State<ExpiredT>) : State<RevokedT> { ignore s; #revoked () };

func reset (s : State<RevokedT>) : State<IdleT> { ignore s; #idle () };

// Accessors only available in the right state:
func getToken (s : State<AuthedT>) : Text =
  switch s { case (#authed sess) sess.token };
func getRefreshFromExpired (s : State<ExpiredT>) : Text =
  switch s { case (#expired stale) stale.refresh };
```

</details>

<details><summary>The driving call sequence</summary>

```motoko
let s0 : State<IdleT>    = start ();
let s1 : State<PendingT> = initiate "verifier_abc";
let s2 : State<AuthedT>  = complete ("tok_v1", "rfr_v1", 1000);
assert getToken s2 == "tok_v1";

// Token expires → still have the refresh.
let s3 : State<ExpiredT> = expire s2;
assert getRefreshFromExpired s3 == "rfr_v1";

// Refresh succeeds → back to Authed with new access token,
// refresh token preserved across the call.
let s4 : State<AuthedT>  = refreshToken (s3, "tok_v2", 2000);
assert getToken s4 == "tok_v2";

let s5 : State<RevokedT> = revokeAuthed s4;
let s6 : State<IdleT>    = reset s5;

// Alternate path: refresh-token call fails → revoke from Expired.
let _alt : State<IdleT> = reset (revokeExpired (expire (complete ("t", "r", 0))));
```

</details>

### What this exercises that the simpler version doesn't

- **Distinct payload per state.** `Authed` carries a full record,
  `Expired` only the refresh token, `Pending` only a Text. The
  refinement plus destructure mints bindings at the right
  payload type without further annotation.
- **State-preserving field-level operations.** `expire` reads
  `sess.refresh` out of the `Authed` record and packs it into the
  `Stale` record for `Expired` — runtime data carried across the
  transition, types updated to match.
- **Branching from one state to multiple.** From `Authed` you can
  go to either `Expired` (clock) or `Revoked` (logout). The
  caller's choice; no ambient state required.
- **Honest limit without unification.** A single `revoke(s : State<???>)`
  taking both `Authed` and `Expired` isn't expressible without
  union refinement — surfaces the OCaml-power-without-unification
  trade-off in concrete code.

crusso:
Haven't looked at code but it's worth checking you respect scoping of quantifiers. In a case branch, the return type should not depend on locally quantified types (no escape) since the type of the entire case needs to be well formed  in the outer context.

ggreif:
> scoping of quantifiers

I guess you don't mean "scoping qualifiers" (the `async` scope marker). Just to be sure I don't misunderstand.

> Haven't looked at code but it's wortth checking you respect scoping of quantifiers. In a case branch, the return type should not depend on locally quantified types (no escape) since the type of the entire case needs to be well formed in the outer context.

We are very careful not to expose existential to the outer world via the type signature, similarly we check for skolem-typed values being part some (observable) expression and error on those. The `skolem -> Any` conversion is intercepted and an error (when dealing with declared existentials).

PR #6134, opened 2026-05-20
fix: split M0169 — distinct error code for missing migration input

Kamirus:
## Summary

Splits stable-signature compatibility error **M0169** into two codes so the two genuinely distinct failure modes can be told apart.

**Before** (both reported as `M0169`):

```
Compatibility error [M0169], the stable variable `x` of the previous version cannot be implicitly discarded. ...
Compatibility error [M0169], the previous version does not contain the stable variable `y`. The migration function cannot require this variable as input. ...
```

**After**:

```
Compatibility error [M0169], the stable variable `x` of the previous version cannot be implicitly discarded. ...
Compatibility error [M0263], the previous version does not contain the stable variable `y`. The migration function cannot require this variable as input. ...
```

## Why

The two cases describe **opposite** problems and require **different fixes**:

| | M0169 (kept) | M0263 (new) |
|---|---|---|
| Trigger | Field present on the previous side, missing on the new side | Field marked required on the new side, missing on the previous side |
| Root cause | New code dropped a stable field | Migration/actor declares a stable input that prior state never had |
| Severity | Potential **data loss** | **Unsatisfiable input** |
| User fix | Write a migration that consumes the old field | Make input optional / supply default / restructure migration chain |

The comment in `error_codes.ml` for M0169 even said *"Stable variable will be discarded. This may cause data loss."* — only fitting the discard case.

This also mirrors the existing split between `M0170` (must subtype) and `M0216` (must stable subtype), which were factored apart for the same reason.

## Migration notes

External tooling that filters on `M0169` will need to add `M0263` if it cares about the "missing required input" case. The error message texts are unchanged, so anything matching on message text rather than code is unaffected.

## Test plan

- [x] Targeted `enhanced-migration-bad-chain` + `enhanced-migration-bad-let` tests pass.
- [x] Full `test/cmp/` suite (`make -C test/cmp`) passes.
- [x] All 16 `enhanced-migration*` tests pass.
- [x] No new failures in `test/fail/` (one pre-existing `implicit-derivation-deep` failure exists on master too — unrelated).
- [x] CI green.


ggreif (approved):
LGTM! Feel free to cut `1.8.1`!

PR #6132, opened 2026-05-19
docs: prepare doc/md for transform-free sync with docs.internetcomputer.org

marc0olo:
## Summary

Transforms `doc/md/` into a Starlight-ready state so the consumer site (`docs.internetcomputer.org`) can sync Motoko language docs via a plain rsync; zero bash/Node.js post-processing pipeline.

### §1 — Frontmatter migration (Docusaurus → Starlight)
- `sidebar_position: N` → `sidebar: { order: N }`
- `title:` and `description:` added to all pages
- Docusaurus-specific keys removed (`hide_table_of_contents`, `custom_edit_url`, `keywords`)

### §2 — Admonition syntax
- `:::info` → `:::note`, `:::warn` / `:::warning` → `:::caution` throughout

### §3 — Code fence cleanup
- ` ``` motoko ` (with leading space) → ` ```motoko `
- Bare ` ```motoko ` (no annotation) → ` ```motoko no-repl ` where appropriate

### §4 — `_category_.yml` removal
- All 18 `_category_.yml` files deleted; ordering now lives in `sidebar.order` frontmatter

### §5 — Numeric prefix removal
- All `N-filename.md` → `filename.md` via `git mv` (history preserved)
- Slug collision resolved: `types/3-functions.md` → `types/function-types.md`

### §6 — Internal link updates
- All inter-page links updated to use new paths after §5 renames
- Broken `mops.one/core/docs/index` URL fixed to `https://mops.one/core`

### §7 — `mo:base` → `mo:core` migration
- All `import X "mo:base/..."` → `import X "mo:core/..."` in code examples

### §8 — Em-dash removal
- All `—` in prose replaced with colons, semicolons, or parentheses (ban from consumer site applies)

### §9 — Docusaurus infrastructure deleted
- `doc/docusaurus/` (31 files) — full Docusaurus site deleted
- `doc/md/old/` (65 files) — deprecated content deleted
- `doc/md/base/` (51 files) — auto-generated mo:base API docs deleted
- `doc/md/core/` (49+ files) — auto-generated mo:core API docs deleted (canonical home: mops.one/core)
- `doc/Makefile` — removed `base:`, `preview:`, `doc:` Docusaurus targets

### §10 — Changelog page
- `reference/3-changelog.mdx` → `reference/changelog.md` using `md file=<motokoRoot>/Changelog.md` fence
- The consumer site's `remark-include-file` plugin inlines `Changelog.md` at build time

### §11 — Index stubs
[...]

PR #6135, opened 2026-05-20
fix(ci): skip GitHub-App token + perf step on fork PRs

Kamirus:
## Why

Every PR from a fork currently fails CI before tests even start. The `tests` matrix job dies at the `Create GitHub App Token` step because GitHub never exposes `vars.*` / `secrets.*` to fork workflows — `client-id` is empty and the action errors out. Since `tests` is a required check via `verify-main-tests`, no external contributor can land a change, regardless of what they touched. Concrete example: #6132.

This isn't a contributor-permission problem — there's no GitHub setting or org allowlist that hands secrets to fork workflows. The workflow itself has to handle it, the way the `artifacts` / `autoclose` / `approvals` jobs in this same file already do.

## What changes

Fork PRs now run the full test suite to completion and get a green check. They don't get the perf comment or the `test/bench` auto-commit, because those need to push to / comment on the PR — which the action genuinely can't do from a fork (it's also why the perf action already has `is_fork` gating internally).

Same-repo PRs and `master` are unchanged.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -81,7 +81,10 @@ jobs:
           cachix-auth-token: ${{ secrets.CACHIX_AUTH_TOKEN }}
           max-jobs: ${{ matrix.max_jobs }}
 
+      # Secrets/vars are not exposed to fork PRs, so this step (and the perf
+      # step that consumes its output) is skipped on fork PRs.
       - name: Create GitHub App Token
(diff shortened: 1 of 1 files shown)


christoph-dfinity (approved):
Stamp

ggreif (commented):
Cool!

PR #6136, opened 2026-05-20
chore: drop dead docs gh-pages pipeline + stale Docusaurus refs

Kamirus:
Follow-up to #6132.

## Why

After #6132, the `report-site` GitHub Pages deployment at
<https://caffeinelabs.github.io/motoko/> stopped serving anything useful:

- `base-doc/` — generated `mo:base` API HTML. The standard library is now `core`, with canonical docs on mops.one. `mo:base` is legacy.
- `docs/md/` — a verbatim copy of `doc/md/`. The real rendered docs are at <https://docs.internetcomputer.org/languages/motoko/>; the raw markdown is unreadable in a browser.
- `coverage/` — useful in principle, but the deployed site currently shows "5 directories, 0 files" (the symlinks resolve to empty content), so even this is broken in production.

Net: every link on the live site is a 404. Killing the pipeline is the right call.

## What

- Delete `nix/base-doc.nix`, `nix/core-doc.nix` (orphan since #6132), `nix/report-site.nix`.
- Drop `base-doc` and `report-site` from `flake.nix` `common-constituents` (verified `nix eval` still resolves the remaining outputs cleanly).
- Remove the `reports:` job from `.github/workflows/test.yml`.

While here, also clean up stale references to the removed Docusaurus site:

- `.gitignore`: drop `.docusaurus`.
- `.github/CODEOWNERS`: `doc/docusaurus/package-lock.json` → `doc/site/package-lock.json` (matches `BOT_APPROVED_FILES`, which #6132 already updated).
- `README.md`: replace dead `[Base package documentation](doc/md/base/index.md)` link with `[Core package documentation](https://mops.one/core/docs)` — `core` is the active standard library; the legacy `base` repo is still linked in the README body.

## Out of scope

- The `gh-pages` branch and the Pages setting on the repo are left alone — those require repo-admin action and are unrelated to the build.
- Other follow-ups from #6132 (homepage use-case section, remaining `mo:base/HashMap` examples in `system-functions.md`, `doc/attic/`) tracked separately.


christoph-dfinity (approved):
LGTM (these kind of did serve double duty as a means of E2E testing mo-doc, but we can come up with proper testing for that at some point)

Kamirus:
Thanks @christoph-dfinity. Confirmed that `test/mo-doc/` still has golden-file coverage for the basic shape (2 fixtures), so we're not flying completely blind. Happy to revisit broader integration testing if a regression slips through, but won't add speculative coverage now.

EDIT: What my agent^ tried to say is: maybe we should just rely on normal mo-doc tests and improve them, dropping these base/core docs makes releases easier and faster

PR #6138, opened 2026-05-20
chore: drop obsolete `make -C doc base` from release flow

ggreif:
## Why

Step 0 of the release procedure in Building.md says to run:

\`\`\`bash
make -C rts
make -C src
make -C doc base
git diff
\`\`\`

The third line fails today: \`doc/Makefile\` only has \`preview\` /
\`build\` / \`serve\` / \`_site-install\` targets.  Anyone following
the documented release flow trips on
\`make: *** No rule to make target 'base'.  Stop.\` mid-release.

## What

Drop the stale step.  Keep the still-valid rts/src build sanity
check (catches local build breakage before a release).  Add a note
pointing at the existing \`niv-updater-action\` that handles
downstream \`motoko-base\` / \`motoko-core\` markdown sync, so
nobody wonders where the base-docs-sync went.

Caught while cutting 1.8.1 (#6137).

### DIFF
--- a/Building.md
+++ b/Building.md
@@ -113,17 +113,16 @@ git switch master
 git pull
 ```
 
-Make sure the markdown doc for base is up-to-date:
-For now, in a nix shell `$ nix develop` (or _re-enter_ if you already have one open):
+Make sure the compiler, runtime, and any generated docs still build
+cleanly and the working tree stays clean:
 
 ```bash
-  make -C rts
-  make -C src
-  make -C doc base
-  git diff
+nix develop -c bash -c "
+  make -C rts &&
+  make -C src &&
+  git -C doc diff
+"
 ```
-
-If not, create and merge a separate PR to update the doc (adding any new files) and goto step 0.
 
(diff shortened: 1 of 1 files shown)


Building.md
```diff
@@ -113,18 +113,18 @@ git switch master
...
-  git diff
+nix develop -c bash -c "
+  make -C rts &&
+  make -C src &&
+  make -C doc core &&
+  git -C doc diff
+"
 ```
 
-If not, create and merge a separate PR to update the doc (adding any new files) and goto step 0.
```
ggreif, line 126: I think it is streamlining the release workflow if smallish doc changes may enter in the release PR. WDYT?
For major changes (not on the horizon) of course one should drop a proper PR.

Kamirus: Recent Marco's PR dropped all base and core docs from this repo, we just link now to mops
So yeah we can safely drop this, I also have a followup cleanup PR but this PR looks good

Kamirus (approved):

PR #6144, opened 2026-05-21
fix: don't suggest M0236 dot notation when receiver isn't postfix

Kamirus:
## What changes for users

Compiling

```motoko
ignore Nat.toText((pos * 6364136223846793005 + 1442695040888963407) % 4294967296);
```

with `-W M0236` (e.g. via `mops check`) used to print

```
warning[M0236]: You can use the dot notation `((pos * 6364136223846793005 + 1442695040888963407) % 4294967296).toText(...)` here
```

with no machine-applicable edits attached, so `mops check --fix` could
not apply it. The suggestion itself was also debatable — the rewrite
adds synthetic outer parens (`(expr).toText()`) purely to satisfy the
parser, which is at best a wash against the original `Module.f(expr)`
form.

After this PR, M0236 only fires when the receiver is already a
postfix expression (variable, literal, call, indexing, `.field`,
`!`, array literal), i.e. exactly the cases where the rewrite is a
clear readability win **and** we can hand `mops check --fix` an edit
it can apply. Every emitted M0236 now carries machine-applicable
edits.

Likewise gone: `Odd.equal(#odd)` / `Odd.compare(#odd, #odd, #odd)`
(TagE receivers) used to warn `(#odd).equal(...)` with no fix; they
now warn-free.

Trivial-receiver cases (`Map.size(m)`, `Map.get(m, …)`,
`Map.clone(x).size(...)`, etc.) are unchanged.

## Why

Two complaints conflated into one fix:

1. The warning is meant to be machine-applicable (it's how `mops
   check --fix` enforces dot-notation style). Emitting it without an
   edit defeats the workflow.
2. For non-postfix receivers the rewrite arguably hurts readability
   — `Module.f(<complex>)` reads better than `(<complex>).f()`
   because the verb (`f`) shows up before you've finished scanning
   the long expression.

The cutoff is `Syntax.is_postfix_exp`, which is exactly "can appear
to the left of `.` without parenthesization" (matches the
`exp_post` parser rule). That's the right invariant for both
points.

## Test plan

- [x] `test-runner -baf contextual-dot` — passes; two snapshot lines removed.
- [x] `test-runner -baf edit-suggestions` — passes; added a `BinE` and `IfE` receiver case as negative regression tests.
- [x] `test-runner -bf fail/` — only pre-existing unrelated failure (`implicit-derivation-deep` whitespace, present on master too).

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -1997,12 +1997,9 @@ let check_can_dot env ctx_dot (exp : Syntax.exp) tys es at =
           DotE ({ it = VarE {it = mod_id1; note = (Const, _); _};_ } as old_receiver,
                 { it = id1; _},
                 _)  when mod_id0 = mod_id1 && id0 = id1 ->
-          (* Skip non-postfix receivers: `(complex).f()` is a debatable style change and we'd emit no autofix anyway. *)
-          if not (Syntax.is_postfix_exp e) then () else
-          (match
-             if e.at.left.line <> e.at.right.line then None
-             else read_region e.at
-           with
+          (* Skip non-postfix or multi-line receivers: `(complex).f()` is a debatable style change and we'd emit no autofix anyway. *)
+          if not (Syntax.is_postfix_exp e) || e.at.left.line <> e.at.right.line then () else
+          (match read_region e.at with
            | None -> ()
            | Some receiver_text ->
              let replace_receiver = edit old_receiver.at receiver_text in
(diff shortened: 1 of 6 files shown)


src/mo_frontend/typing.ml
```diff
@@ -1997,25 +1997,24 @@ let check_can_dot env ctx_dot (exp : Syntax.exp) tys es at =
...
-          in
-          warn env at "M0236" "You can use the dot notation `%s.%s(...)` here"
-            ~edits
-            receiver_text
-            id.it
+          (* Skip non-postfix receivers: `(complex).f()` is a debatable style change and we'd emit no autofix anyway. *)
+          if not (Syntax.is_postfix_exp e) then () else
+          (match
+             if e.at.left.line <> e.at.right.line then None
+             else read_region e.at
```
christoph-dfinity: Maybe it would be cleaner to also move the multi-line condition up into this if?

christoph-dfinity (approved):

PR #6153, opened 2026-05-26
feat(typing): allow free-text message after M0235 deprecation code

Kamirus:
## What changes for users

Two related deprecation-warning improvements that share a single output format.

### 1. M0235 (caffeine deprecation) can now carry a message

Module authors can attach a free-text message after the `M0235` code in a `/// @deprecated` doc comment. The text is surfaced as a `note:` sub-diagnostic at every use site, telling consumers *why* something is deprecated or what to use instead.

**Before:**

```motoko
module C {
  /// @deprecated M0235 use \`bar\` instead
  public let foo = 5;
};
ignore C.foo;
```

```
warning [M0235], field foo is deprecated for caffeine
```

The trailing \" use \`bar\` instead\" was silently dropped — `check_deprecation` only matched the exact string `\"M0235\"`, so any extra text fell through to the generic M0154 path with a nonsensical message (`\"M0235 use bar instead\"`).

**After (human format):**

```
warning[M0235]: field foo is deprecated for caffeine
    ┌─ test.mo:5:9
  5 │  ignore C.foo;
    │         ^^^^^
    = note: use \`bar\` instead
```

Bare `/// @deprecated M0235` keeps the previous one-line message.

### 2. M0154 (generic deprecation) message rendering aligned with the rest of the compiler

`/// @deprecated <free text>` previously rendered the author's text as a second line glued on with `:\n`:

```
warning [M0154], field foo is deprecated:
do not use foo anymore
```

It now uses the same structured-note machinery that M0230 / M0236 / etc. already use:

```
warning [M0154], field foo is deprecated
note: do not use foo anymore
```

In the human/fancy renderer this becomes `= note: do not use foo anymore` indented under the source span. The author-provided text is unchanged; only the framing differs.

## Why

1. M0154 already supported a free-text message; M0235 didn't. The result was that caffeine deprecations could tell consumers *that* something was deprecated, not *what to do about it*. This makes them as expressive as plain `@deprecated`, while still routing under the dedicated M0235 code (so projects opting into `-W M0235` keep one code to filter on).
2. M0154's `:\n`-glued-message format was a one-off. Every other diagnostic in this codebase that wants to attach explanatory text uses `~notes`, which the fancy renderer prints as a structured `= note:` line. Aligning M0154 means LSP hovers, JSON output (`json_string_of_message`), and the human renderer all treat the author's message as first-class metadata.

## Implementation

The whole feature is two lines in `typing.ml`:

```ocaml
| Some msg ->
  match Lib.String.chop_prefix \"M0235 \" msg with
  | Some m -> warn env at \"M0235\" ~notes:[m] \"%s %s is deprecated for caffeine\" desc id
  | None -> warn env at \"M0154\" ~notes:[msg] \"%s %s is deprecated\" desc id
```

The M0235-exact and M0199 branches are untouched. M0199 has a prescriptive hardcoded message that wouldn't compose well with user text, so it's intentionally left alone.

## Test plan

- [x] `test-runner -baf caffeine-deprecated` — passes; added a `withMsg` case in its own `module C` at the bottom of the fixtures so existing snapshot lines stay byte-identical and the new warning just appends.
- [x] `test-runner -baf deprecated|disable-enable-warnings` + `test/perf` make accept — all three pre-existing M0154 snapshots updated mechanically (5 files total, transformation is `is deprecated:\n<msg>` → `is deprecated\nnote: <msg>`).
- [x] `test-runner -bf fail/` — only pre-existing unrelated failure (`implicit-derivation-deep` whitespace, present on master too).
- [x] Changelog entries for both changes under `## Unreleased`.


christoph-dfinity (approved):
LGTM

PR #6165, opened 2026-06-02
perf: default to WasmMV (real multi-value)

ggreif:
## Summary

Flips `Flags.multi_value` default from `false` (FakeMV — Globals-stash emulation) to `true` (WasmMV — real Wasm multi-value returns).

- Builds on #6113 (merged), which introduced the `--experimental-multi-value` / `--no-experimental-multi-value` flag pair and the FakeMV/WasmMV test-runner modes.
- `--no-experimental-multi-value` still selects the FakeMV path; nothing is removed.
- The one-line change in `src/mo_config/flags.ml` compiles cleanly (`make -C src`).

## Test plan

- [x] CI green across `tests.run`, `tests.run-drun`, `tests.run-wasm-opt` etc. (covers both WasmMV and FakeMV test modes)
- [x] Bench suite shows no regression (no change, because it already had the same `--experimental-multi-value` setting before)


ggreif:
**`test/bench` vs `master` — no regressions.**

Checked the multi-value-default flip against `master`:
- **Committed `test/bench/ok/*.ok`: identical** on both — CI's auto-accept regenerated nothing, so the in-canister cycle counts are unchanged by the flip.
- **Perf bot** (external gas/size, base `e7c78d7c` → `6e0e167c`): **5 benchmarks improved, 0 worsened**; mean **−0.4% gas, −1.1% size**.

Every benchmark that moved moved the beneficial way (smaller wasm, fewer gas units); the rest were within noise. A per-test table isn't available from the committed `.ok` (unchanged) or the perf bot (aggregate only) without a local bench run.

Verdict: net improvement, nothing regressed.

Kamirus (approved):
stamp

PR #6166, opened 2026-06-02
fix: M0237 unsoundness — verify suggested call still typechecks

Kamirus:
M0237 ("argument can be inferred and omitted") could suggest removing an implicit argument that the compiler then rejects with M0098. When a type parameter is pinned only by the argument being suggested for removal — e.g. inside a polymorphic wrapper `id<A>(a : A) : A` — dropping it leaves the parameter underconstrained, so applying the suggested edit breaks compilation.

The warning now re-runs inference with the candidate arguments replaced by holes and fires only when the trial recovers the same instantiation.

**Before**

```motoko
ignore id(Map.fromIter(Iter.fromArray<(Nat, Text)>([]), Nat.compare));
// warning [M0237]: `Nat.compare` can be inferred and omitted here

// applying the suggested removal:
ignore id(Map.fromIter(Iter.fromArray<(Nat, Text)>([])));
// error [M0098]: no "best" choice for type parameter `K`
```

**After** — no warning for the wrapped form; legitimate suggestions (void context, `implicit-suggest.mo`) still fire. M0223 (redundant type instantiation) now shares the same trial helper.

## Limitation

The trial runs in `pre` mode, so it can't re-enter an already-inferred expression (that would trip `assert exp.note.note_typ = T.Pre`) — the same constraint blocking [#5742](https://github.com/caffeinelabs/motoko/issues/5742). A proper fix needs real backtracking in the non-pre phase (snapshot/restore notes) plus a flag to cap nested trials; both have stalled as too invasive. The pre-mode approximation fixes the reported case but is not guaranteed to match non-pre results in every case.

Running trials in `pre` also bounds the cost: because a trial never spawns nested trials, deeply nested redundant calls stay quadratic (O(n²) in call nesting), not exponential.

## Also in this PR

Unrelated: tightens the changelog rule in the PR-review prompt (`.github/prompts/pr-review-prompt.md`) to require the `(#NNNN)` PR-number reference.

Linear: [LANG-1323](https://linear.app/caffeinelabs/issue/LANG-1323/fix-m0237-unsoundness-trial-based-candidate-validation).


src/mo_frontend/typing.ml
```diff
@@ -3383,6 +3387,39 @@ and infer_call env exp1 inst (parenthesized, ref_exp2) at t_expect_opt =
...
+      let exp2_with_holes =
+        { exp2 with it = insert_holes at t_args syntax_args_no_implicits;
+                    note = empty_typ_note }
+      in
+      let env_pre = { env with pre = true } in
+      match
+        Diag.with_message_store (recover_opt (fun msgs ->
+          let env' = { env_pre with msgs } in
+          let ts', _, _ =
+            infer_call_instantiation env' t1 ctx_dot tbs (T.seq t_args) t_ret
```
Kamirus: Calling infer twice could result in exponential explosion, we might want to have a flag in the env to skip this check in recursive calls

christoph-dfinity (approved):
LGTM, as discussed

PR #6168, opened 2026-06-02
fix: diagnostic columns count codepoints, not UTF-8 bytes

Kamirus:
JSON diagnostic spans now carry `byte_start`/`byte_end` (0-based UTF-8 byte offsets from start of file, matching `rustc`); tools applying `suggested_replacement` spans should prefer these — bytes are encoding-independent and need no de/re-encoding. `column_start`/`column_end` stay 1-based codepoint columns for human display (the plain text `5.31` format moves to codepoints too, matching what editors show and what the Grace-rendered human format was already doing).

Before, `column_start`/`column_end` (and the plain text `line.col`) were raw lexer UTF-8 byte columns. Tools like `mops check --fix` that index by codepoint or UTF-16 code units over-deleted on lines with multi-byte characters — e.g. `Char.toNat32('京')` lost its closing `)`.


christoph-dfinity:
Tools should generally prefer byte offsets for applying diffs, as those are independent of encodings and can be applied without de- and re-encoding the source code.

For positions presented to humans I agree the codepoint (USV) offset should be preferred (but don't confuse that with utf16 codeunits, as the PR description does). If you want to adhere to `rustc`'s format, I'm fine adding `byte_start` and `byte_end` to the json errors (where those represent bytes from start of file).

We already have `uutf` in the dependency tree, no need to hand-role utf8 decoding in here.

Kamirus:
Thanks @christoph-dfinity — addressed all three points in d318c51:

- Added `byte_start`/`byte_end` (0-based, from start of file, matching `rustc`). Spec calls them out as the preferred field for diff application. Kept `column_start`/`column_end` as codepoints for human display.
- Swapped the hand-rolled UTF-8 stepping for `uutf`.
- Fixed the PR description — codepoints/USVs ≠ UTF-16 code units.

christoph-dfinity:
This still looks like a lot of code for counting codepoints and offsets. IIRC `uutf` does most of that work. Check the "Character position" chapter in https://ocaml.org/p/uutf/latest/doc/uutf/Uutf/index.html#decode

christoph-dfinity (approved):
LGTM, thanks for looking into this!

PR #6173, opened 2026-06-04
fix: don't suggest M0236 dot-rewrite for literal receivers

Kamirus:
M0236 suggested `lit.f()` rewrites that didn't compile. Three failure modes for literal receivers:

- **Signed**: `Float.isNaN(-1.1)` → `-1.1.isNaN()` parses as `-(1.1.isNaN())`.
- **Hex**: `Int.abs(0xff)` → `0xff.abs()` lexes as a single hex-float token.
- **Coercion-dependent**: `Blob.isEmpty("\00")` → `"\00".isEmpty()` fails because contextual-dot lookup is weaker than argument-position inference — it never re-tries `check_lit`'s Text→Blob / Nat→Nat8 coercion.

Skip `LitE` receivers; other receiver forms unchanged. The coercion case is pinned by `test/fail/M0236-coercion-autofix-broken.mo`.


Changelog.md
```diff
@@ -2,6 +2,8 @@
 
 * motoko (`moc`)
 
+  * bugfix: M0236 dot-notation suggestion no longer fires for literal receivers — the `lit.f()` rewrite could misparse (`-1.1.isNaN()` → `-(1.1.isNaN())`), mis-lex (`0xff.abs` as a hex float), or fail to type-check when it lost a literal coercion (`Blob.isEmpty("\00")` → `"\00".isEmpty()`).
```
Kamirus:
```suggestion
  * bugfix: M0236 dot-notation suggestion no longer fires for literal receivers — the `lit.f()` rewrite could misparse (`-1.1.isNaN()` → `-(1.1.isNaN())`), mis-lex (`0xff.abs` as a hex float), or fail to type-check when it lost a literal coercion (`Blob.isEmpty("\00")` → `"\00".isEmpty()`) (#6173).
```

test/fail/multibyte-columns.mo
```diff
@@ -6,10 +6,10 @@
...
-  public func go() {
-    ignore Char.toNat32('A');   // ASCII
-    ignore Char.toNat32('京');  // 3-byte UTF-8 inside span
-    ignore Char.toNat32('💩'); // 4-byte UTF-8 inside span (non-BMP)
-    ignore "京京"; ignore Char.toNat32('D');  // multibyte BEFORE the span
+  public func go(c : Char) {
+    ignore Char.toNat32(c);                       // ASCII baseline
+    ignore "京"; ignore Char.toNat32(c);          // 3-byte UTF-8 before span
+    ignore "💩"; ignore Char.toNat32(c);          // 4-byte UTF-8 (non-BMP) before span
+    ignore "京京"; ignore Char.toNat32(c);        // multibyte text before span
```
christoph-dfinity, line 13: How is this change related?

Kamirus: To still test the fix suggestions are counting the UTF-8 characters correctly.
Since this PR stops suggesting context-dot rewrites for literals we would loose the suggestions in the output.
But now that I think of it and look at the updated output it's hard to know if the `byte_start":486,"byte_end":501` values are correct or not 🤔

christoph-dfinity: I see. Let's assume they're correct and treat as pure regression tests.

christoph-dfinity (approved):

Issue #6163, opened 2026-06-01
Allow `label` with fallthrough default value


ggreif:
Currently, using a labeled early-exit with a fallthrough default requires wrapping in some clumsy boilerplate:

```motoko
label found : Bool do {
  for (candidate in values.vals()) { if (cmp(v, #eq, candidate)) break found true };
  false
}
```

The `false` is purely mechanical — it's the fallthrough value when no `break found` fires. The `do { ...; default }` wrapper exists only to express it.

**Proposal:** Allow the fallthrough default to be written directly on the `label`:

```motoko
label found = false
  for (candidate in values.vals()) { if (cmp(v, #eq, candidate)) break found true }
```

Type of `found` is inferred from the default expression (and must agree with any `break found <val>` sites). Desugaring is exactly:

```motoko
label found : T do { <body>; <default> }
```

**Use cases beyond `Bool`:** The pattern generalises naturally to any sentinel fallthrough:

```motoko
// Option: null when nothing found
label result = null
  for (x in xs.vals()) { if (pred x) break result (?x) }

// Result: error when nothing found
label result = #err "not found"
  for (x in xs.vals()) { if (pred x) break result (#ok x) }
```

In all cases the fallthrough default is the "nothing happened" sentinel and the `break` carries the success value — a common and idiomatic pattern that currently requires the mechanical `do { ...; default }` wrapper.

ggreif:
An extended form with an explicit type annotation should also be supported, for cases where the type cannot be inferred from the default alone (e.g. the default is `null` which has type `Null`, not `?T`):

```motoko
label found : ?Item = null
  for (x in xs.vals()) { if (pred x) break found (?x) }
```

So the full syntax would be:

```
label <id> [: <type>] [= <default>] <expr>
```

with the constraint that at least one of `: <type>` or `= <default>` must be present when a default is intended (the bare `label <id> <expr>` form stays unchanged).

PR #6176, opened 2026-06-04
experiment: infer label type from its fallthrough default (#6163)

ggreif:
**Experimental**, stacked on #6174 (the parser desugaring). #6174 makes the *annotated* `label x : T = <dflt> <body>` work; this completes the issue by letting the **compact** `label x = <dflt> <body>` take its type from the default — no `: T`, and it works in synthesis positions too:

```motoko
let v = label found = false                         // v : Bool, inferred from the default
  for (x in xs.vals()) { if (x == t) break found true };
```

### Mechanism

`LabelE` grows a `bool`, set by the parser action **only when a default is present and no annotation is written** (accepted grammar unchanged). For that flag the typechecker infers the block's final expression (the default — it can't mention the label) and uses it as the label type; `break`s check against it.

To infer that sub-expression without committing its type notes (so the whole block is then checked intact), `env` grows `commit_notes : bool` (default `true`, cleared for that one inference). It threads inward via the usual `{env with …}`; record-completeness means no note-setter can silently skip it. No global state.

### Semantics ("type wins")

| form | label type |
|---|---|
| `label x = false …; break x true` | `Bool` (from the default) |
| `label r = (null : ?Nat) …; break r (?x)` | `?Nat` — annotate the **default** to widen a sentinel |
| `label p : ?Int = null …; break p (?1)` | `?Int` — **annotation wins** (flag not consulted) |
| `label r = null …; break r (?x)` | error: `?Nat ⊄ Null` (`Null` isn't bottom) |
| `label s = (return …) …; break s` | bottom default is **forgotten** → unit fallback |

No lub over break sites; the default (optionally annotated) is authoritative — like a `var`/`let` initializer.

### Tests
`test/run/label-default.mo` (compact infer in synthesis + checking, annotation-wins, annotated-default sentinel, bottom-default fallback, laziness, bare label) — all backends; `test/fail/label-default-compact.mo` (bare narrow sentinel rejects a wider break).


src/mo_def/arrange.ml
```diff
@@ -126,7 +126,7 @@ module Make (Cfg : Config) = struct
     | LoopE (e1, None, _)    -> "LoopE"   $$ [exp e1]
     | LoopE (e1, Some e2, _) -> "LoopE"   $$ [exp e1; exp e2]
     | ForE (p, e1, e2, _)    -> "ForE"    $$ [pat p; exp e1; exp e2]
-    | LabelE (i, t, e)       -> "LabelE"  $$ [id i; typ t; exp e]
+    | LabelE (i, (t, b), e)  -> "LabelE"  $$ [id i; typ t; exp e; Atom (string_of_bool b)]
```
ggreif, line 129: the `, defaulted` should not be at the end

PR #6177, opened 2026-06-04
fix: don't suggest M0236 dot-rewrite when receiver cannot infer or changes resolution

Kamirus:
M0236's dot-rewrite suggestion was unsound when the call relied on coercion the receiver position can't reproduce. [#6173](https://github.com/caffeinelabs/motoko/pull/6173) fixed the literal flavour; container/branch receivers hit the same hazard. Reduced from `motoko-core/Base64.encode`:

```motoko
Array.toBlob([b[i], if (i == 0) b[i] else 61]) // [Nat8] only because `61 : Nat` coerces against the param
```

Rewriting to `[…, 61].toBlob()` reinfers the array as `[Any]`, so `toBlob` no longer resolves.

Fix: re-infer the receiver standalone via `with_backtracking` (no param type as context) and suggest the rewrite only when its type is **exactly equal** to the parameter type. Exact equality — not `<:` — also blocks resolution changes, where a merely-compatible subtype would dispatch to a different module function (`id(1).tricky()` → `Nat` vs `Int`). Receivers that can't be inferred at all are skipped too.

## Out of scope

- The `LitE` skip in `is_postfix_exp` stays — the trial covers its coercion rationale but not the lexical one (`-1.1.isNaN()` reparses, `0xff.abs()` lexes as a hex float).
- Pre-mode inference is approximate — same caveat [#6166](https://github.com/caffeinelabs/motoko/pull/6166) carries.


ggreif (approved):
Looks benign to me, the worst that can happen is to give a misleading edit hint. The test cases seem conclusive. LGTM.

PR #6182, opened 2026-06-10
ci: offload Linux builds to nixbuild.net (cost/perf)

marko-k0:
## Goal

Make CI more cost-effective, starting with the jobs that take the most time.

## Context / change of approach

This PR originally switched the heavy jobs to [Namespace](https://namespace.so) runners. Per review feedback from @ggreif and @basvandijk, it now instead aligns with the **nixbuild.net** approach from [#5032](https://github.com/caffeinelabs/motoko/pull/5032) (which was later reverted in [#5694](https://github.com/caffeinelabs/motoko/pull/5694)). nixbuild.net is a remote Nix builder *optimised for Nix*: each derivation gets its own machine (up to 16 CPUs) rather than sharing the 16 CPUs of one big GitHub runner — faster **and** cheaper. Namespace is great but not Nix-optimised; its main advantage here would be macOS runners, which nixbuild.net doesn't offer.

## Which jobs take the most time?

All in `test.yml`, running on every PR on the expensive `ubuntu-24-large` / `arm64-linux-16` runners:

| Job | ~Duration |
| --- | --- |
| `common-tests` | ~16–18 min |
| `tests (release/debug)` | ~6–16 min each |
| `gc-tests` | ~5–7 min each |

## Changes

- **Offload Linux nix builds to nixbuild.net** via `nixbuild/nixbuild-action` + remote store builds (helper scripts `nixbuild.sh` / `nixbuildcopy.sh`, mirroring #5032).
- Because the runner now does little work, move the heavy Linux jobs off `ubuntu-24-large` / `arm64-linux-16` onto the **free** `ubuntu-latest` / `ubuntu-24.04-arm` runners (`test.yml`, `build.yml`, `release.yml`, nightly `motoko-core-tests`).
- `test-blueprint`: uses nixbuild.net for Linux when `NIXBUILD_TOKEN` is set, and **falls back to a local build for macOS and for fork PRs** (forks can't read secrets) so fork CI keeps working.
- `performance` action: builds via `nixbuildcopy.sh`.
- **Build-abort concurrency** (per @ggreif): `test.yml` cancels a PR's in-progress run on a new push; `master` and merge-queue runs are never cancelled.
- macOS stays on GitHub-hosted runners (nixbuild.net has no macOS); noted that Namespace could cover macOS later if those minutes get expensive.
- `CI.md` documents the setup.

---

## Estimated benefits

> [!IMPORTANT]
> **These are estimates only.** Job durations are averaged from a handful of recent runs and the costs use list prices; the nixbuild.net figure in particular cannot be measured without a live run. **All numbers below should be verified once `NIXBUILD_TOKEN` is configured and the new pipeline has run for real** (check `nixbuild.net> usage` for actual CPU-hours).

### 1. Cost (headline)

This repo is **public**, which is the key lever: `ubuntu-24-large` / `arm64-linux-16` are *larger* GitHub runners and are **billed even in public repos**, while `ubuntu-latest` / `ubuntu-24.04-arm` are *standard* runners and are **free** in public repos. So the 8 heavy `test.yml` jobs move from billed → free, leaving only nixbuild.net compute.

Per-run cost on today's runners (avg of 3 substantive runs; GitHub bills per job rounded up to the minute; Jan-2026 rates: x64 16-core $0.042/min, ARM 16-core $0.026/min):

| Job | x64 (`ubuntu-24-large`) | ARM (`arm64-linux-16`) |
| --- | --- | --- |
| common-tests | ~17 min | ~18 min |
| tests (release) | ~11 min | ~14 min |
| tests (debug) | ~8 min | ~16 min |
| gc-tests | ~5 min | ~7 min |
| **billed subtotal** | **~41 min → ~$1.72** | **~58 min → ~$1.51** |

- **≈ $3.2 per substantive run** (code change, partial cache); **≈ $0.27** per trivial warm-cache run.
- **Volume:** ~342 `test.yml` runs in the last 30 days → with a conservative blend (~30–50% substantive), **≈ $400–600 / month** on these jobs alone *(estimate)*.

**After:**
- GitHub Actions cost for these 8 jobs → **$0** (free standard runners).
- Replaced by **nixbuild.net at €0.12/CPU-hour (~$0.13), with 25 free CPU-hours/month**, billed only for actual (uncached) build CPU-time.
- *Estimate, must be verified:* nixbuild only rebuilds changed derivations (same caching benefit as cachix today), so the marginal cost per PR should be small. A deliberately pessimistic ceiling — every derivation pegging 16 cores for the entire current wall time (never happens in practice) — is ~25 CPU-hours ≈ €3 per cold run; realistic cached runs should be a fraction of that and often within the free tier.

**Net (estimated):** the ~$400–600/mo of guaranteed larger-runner spend is eliminated; the new nixbuild cost is expected to be materially lower, but **this is the one number that needs a real run to confirm.**

### 2. Wall-clock / developer feedback

Mechanism is concrete; magnitude needs validation. Today all 8 jobs share 16-CPU runners and the **critical path is `common-tests` (~17–18 min)** (plus ARM `tests` ~14–16 min). With nixbuild.net each derivation gets its own up-to-16-CPU machine instead of contending, so the long-pole jobs should shrink (the explicit performance motivation of #5032). No speedup figure is claimed here — **to be measured.**

### 3. Build-abort concurrency

- **11 runs were cancelled in the last 30 days**, i.e. superseding already happens manually. Auto-cancel formalizes it: each obsoleted push avoids up to **~95 runner-minutes (~$3 at today's rates)** of wasted compute and frees queue capacity.

### 4. Non-cost benefits

- **Simpler config:** removes the ARM OOM workaround (`max_jobs: 1` serialization for the ~3 GB RTS builds).
- **Fork PRs preserved:** the local fallback keeps fork CI working (forks can't read `NIXBUILD_TOKEN`), which #5032 didn't handle.

### Caveats on the numbers
- Assumes `ubuntu-24-large` / `arm64-linux-16` are GitHub *larger* runners (consistent with #5032's "large expensive GitHub runners" description); if any are self-hosted, the GitHub-side figure shifts to VM hosting cost instead.
- Covers `test.yml` only (the PR cost drivers); `build.yml` / `release.yml` already run mostly on standard/macOS runners.
- The nixbuild.net cost is the only fully estimated figure — confirm from the first real run's `nixbuild.net> usage` output.

---

## Required before this can run

- Add a **`NIXBUILD_TOKEN`** repository secret (nixbuild.net auth). Without it, Linux jobs fall back to slow local builds on the small runners.

## Open questions for reviewers

- **Why was #5032 reverted?** The revert (#5694) was bundled with a "fix `reports` creation" revert and doesn't state a reason. Worth confirming there wasn't an underlying nixbuild.net issue before re-landing. (cc @ggreif / @basvandijk)
- This is a faithful re-application of #5032 adapted to the current workflows (plus the fork fallback and concurrency); I could **not** run CI here (no `NIXBUILD_TOKEN` / runners), so it needs a real CI run to validate — especially the cachix-push interaction with remote-store builds in `build.yml`.
- The label-gated `artifacts` job in `test.yml` was left on the local cachix path (rarely run); say the word if you'd like it on nixbuild.net too.

<div><a href="https://cursor.com/agents/bc-4493d637-1a6e-45cc-9802-1c69f1ce9f92"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-4493d637-1a6e-45cc-9802-1c69f1ce9f92"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


ggreif:
@marko-k0 Some months back @basvandijk had very concrete suggestions (with serious savings) in a similar but more sophisticated (caching) setup. Maybe worth synchronising?

We could possibly also activate build abort (concurrency setting) on rapid-shot commits.

basvandijk:
I think @ggreif was referring to https://github.com/caffeinelabs/motoko/pull/5032 which switched builds to [nixbuild.net](https://nixbuild.net/). We have great experience with them on [Farm](https://github.com/dfinity-lab/farm/blob/master/.github/workflows/actions.yml#L10). They're cheap, very fast and have good personal support. Namespace is great too but not optimized for Nix. However they do offer MacOS runners which nixbuild.net doesn't.

marko-k0:
@ggreif, @basvandijk, do changes right now make sense?

nixbuild.sh
```diff
@@ -0,0 +1,20 @@
...
+    NIX_ARGS+=(
+        --print-build-logs
+        --builders ""
+        --max-jobs 2
+        --eval-store auto
+        --store ssh-ng://eu.nixbuild.net
+    )
+fi
+
+nix build "${NIX_ARGS[@]}" "$@"
```
basvandijk: Maybe:
```suggestion
exec nix build "${NIX_ARGS[@]}" "$@"
```

alexandru-uta:
> @ggreif, @basvandijk, do changes right now make sense?

makes sense, and I even am surprised it worked so smoothly. 👍

alexandru-uta:
my only question is @ggreif @basvandijk and @marko-k0 : who pays for nixbuild.net? do we have a "corporate" account for it? last time when I was talking to Bas about this, Kepler suggested we keep the admin side of things simpler. But if the SRE people have a solution for it, let's go for it.

basvandijk (approved):
LGTM.

As a follow up you could look into using their [`nixbuild/nixbuild-action`](https://github.com/nixbuild/nixbuild-action) `.github/workflows/ci-workflow.yml` workflow like we [use in Farm](https://github.com/dfinity-lab/farm/blob/master/.github/workflows/actions.yml#L11). I expect you could significantly simplify things that way.

basvandijk:
> who pays for nixbuild.net? do we have a "corporate" account for it? 

DFINITY has an account. It would make sense for Caffeine Labs to [sign up](https://nixbuild.net/#register) as well and attach a credit card.

alexandru-uta:
> > who pays for nixbuild.net? do we have a "corporate" account for it?
> 
> DFINITY has an account. It would make sense for Caffeine Labs to [sign up](https://nixbuild.net/#register) as well and attach a credit card.

ok, but then I guess @marko-k0 will take care of this

ggreif (approved):
[Nightly MacOS](https://github.com/caffeinelabs/motoko/actions/runs/28030679578) still works, LGTM!

PR body still needs a touchup!

marko-k0:
> > > who pays for nixbuild.net? do we have a "corporate" account for it?
> > 
> > 
> > DFINITY has an account. It would make sense for Caffeine Labs to [sign up](https://nixbuild.net/#register) as well and attach a credit card.
> 
> ok, but then I guess @marko-k0 will take care of this

I'll pick this up in coming days.

PR #6183, opened 2026-06-10
fix: handles mixins for enhanced migrations

christoph-dfinity:
## Motivation

The `--enhanced-migrations` flag did not have an effect on mixins. It allowed initializers for stable variables in mixins, and didn't consider stable fields in mixins when checking the migration chain. This meant the code would compile fine, but (thankfully) fail on the dynamic checks when upgrading the canister.

This PR properly propagates the flag to mixins, and considers their stable fields when checking the migration chain.


alexandru-uta:
As per the offline discussion -- let's hold on to this for a bit until we relax the `<system>` constraint for transients under the static checks.

alexandru-uta (approved):
works, thank you!

PR #6184, opened 2026-06-11
feat(parser): `func`-field sugar — `func name(…)` as a record field

ggreif:
## Motivation

Currently many real-world records look like

Screenshot 2026-06-16 at 16 47 57

The idea is to refactor a bit and make them look like

Screenshot 2026-06-16 at 16 52 34

which reads awesome IMHO.

### Anecdotal

This mirrors a question I asked (around 2017) Andres Löh (Haskell guru) “why not have Haskell’s records `R { a = \b -> …, … }` to allow `R { a b = …, … }`” , and he replied, “because nobody filed a PR yet”! He was super supportive of the idea.

Some ask "why not use `object { ... }` instead?" Several reasons:
- it has much more ceremony: `public`/`private`, `let`, `object`/`class`
- doesn't support `{ <base> with ... }`
- records are the no-fuss version of full-fledged objects, but bringing the two closer _empowers_

Also the improved field syntax is inherited by _parentheticals_, so `(with func migration(...) ) actor ...` becomes immediately  available.

### Migration

The new syntax is _strictly conservative_, the **user opts-in**. caffeinelabs/tree-sitter-motoko#12 adds tooling preparation and caffeinelabs/lintoko#37 has examples for rewriting if the user chooses to do so.

## Summary

Phase 1 of the **func-fields** feature: allow

```motoko
let r = {
  func incr(x : Nat) : Nat = x + 1;   // instead of:  incr = func(x : Nat) : Nat = x + 1
  func double(n : Nat) : Nat { n + n };
};
```

i.e. `func name(params) : ret { body }` (and `= expr` body) as a **record-literal field**. This removes the verbose `name = func(…){…}` boilerplate that dominates arrow-typed fields.

## What this is (and isn't)

This is **pure parse-time sugar**. The new `exp_field` alternative builds the *same* `FuncE` node that `name = func(…){…}` already produces, so:

- **No AST change** (`syntax.ml` untouched)
- **No typechecker change** (`typing.ml` untouched)
- **No lowering change** (`desugar.ml` untouched)

It is **Phase 1 only** — there is deliberately **no sibling capture**. `{ a = 1; b = a }` still errors (M0057, and not intended to be changed!); record fields do not see each other. (Sibling capture for `func`-fields' bodies is a possible gated Phase 2, not in scope here.)

## Implementation

A single 5-line alternative on the `exp_field` non-terminal in `src/mo_frontend/parser.mly`, reusing the existing `desugar_func_body` / `func_exp` helpers from the `func` *declaration* form:

```
| FUNC x=id tp=typ_params_opt p=pat_plain t=annot_opt fb=func_body
  { let sp = Type.Local @@ no_region in
    let is_sugar, e = desugar_func_body sp x t fb in
    { mut = Const @@ no_region; id = x;
      exp = func_exp x.it sp tp p t is_sugar e @? at $sloc; } @@ at $sloc }
```

Requires a name (no anonymous form), no `var`/`shared`/`query` qualifiers. **Menhir reports zero new conflicts** (`FUNC` was a dead lookahead in `exp_field`). The `{ base with … }` extension form inherits it automatically (same non-terminal).

## Scope notes

- `with`-extension form works for free: `{ b with func f() : Nat = 7 }`.
- The parenthetical `(with …)` form (`encoder`/`decoder`/`migration`) also inherits the sugar by sharing `exp_field` — a genuine ergonomic win there, no special handling needed.

## Test

`test/run/func-field.mo` covers the plain-record, block-body, and `with` forms across **all** phases — `[tc] [run] [run-ir] [run-low] [comp]` — and is green on the full CI matrix (release + debug, both arches). No `ok/*` fixtures are needed: the test is assert-only, so every phase (codegen included) runs silently and returns 0.

## Status

**Draft / proof-of-concept.** Full test matrix is green. Opening for review of the *approach* before polishing (`CHANGELOG`, docs, and a decision on whether the gated Phase 2 sibling-capture is wanted). Design rationale is committed at `design/func-fields.md`.


design/func-fields.md
```diff
@@ -0,0 +1,197 @@
...
+  but carries a `var`-field aliasing pitfall. Whole-record recursive scoping is a
+  breaking change; opt-in via `func`-syntax is safe.
+
+**Cleanest design — two phases:**
+1. **Phase 1 (pure sugar):** `func f(args):ret{body}` in a record literal
+   desugars to `f = func(…){…}` at parse time. No typer change, no sibling
+   capture. Low risk. (Solves the user's stated dislike of the `= func` form.)
+2. **Phase 2 (sibling capture, gated):** extend `infer_check_bases_fields`
+   (`typing.ml:2654`) to gather `func`-declared field names and adjoin them for
+   `func`-field bodies only. Medium risk (hybrid scoping rule + `var` semantics).
```
ggreif: This would allow recursive functions. We should also think about capturing all fields – not only functions – in the augmented `env`, even fields of the bases `{ b1 and b2 with f1 = ...; func f2() = b1sFieldA }`, since `b1` could be a computed expression that is expensive.

Kamirus:
I'm skeptical towards this because:
- it strays from the standard record syntax, has no known precedent in other languages that I know of -- records are always just with fields, if a user wants more then they should use object
- it maybe needs an update to the vscode extension to support this properly dunno

christoph-dfinity:
I gave this a try, but I'm not sure it's a net win:

```motoko
{ indexedElement "ord " with
  otherField = 10;
  lookup = func(parent : Smurf) {
    ...
  };
}

{ indexedElement "ord " with
  otherField = 10;
  func lookup(parent : Smurf) {
    ...
  };
}
```

It only saves a single token, and I find it a bit harder to tell which fields end up in the record. The improvement in the screenshots really comes from `with` omitting fields, which is independent of this change.

So I'm leaning against merging, happy to be convinced otherwise if I'm missing something!

PR #6186, opened 2026-06-15
fix(typing): derive implicits for covariant-only result type variables

Kamirus:
Implicit derivation failed when a candidate's type variable occurred only in a covariant result position (e.g. decoders `Text -> ?T`): the variable was solved to `None` (bottom) instead of the required type, so the implicit had to be passed explicitly.

Solving `?A -> ?B  <:  ?Nat -> ?Int`, when instantiating candidates, should pick the maximal solution.
So `Nat <: A <: Any` and `None <: B <: Int` should solve as `A := Nat, B := Int` — i.e. lower bound for argument vars, upper bound for result vars.

Without a polarity witness `ret_opt`, a result-only var like `B` was treated as bivariant and collapsed to its lower bound `None`. Passing a flipped function type as the witness treats `args` as covariant and `rets` as contravariant, recovering the maximal solution.

### DIFF
--- a/src/mo_frontend/typing.ml
+++ b/src/mo_frontend/typing.ml
@@ -1649,7 +1649,7 @@ end
 
 (** Checks [args -> rets  <:  req_args -> req_rets] via subtyping or
     bidirectional matching when [tbs] are present. Returns [Some inst] or [None].
-    
+
     [inst] is the maximal solution, s.t [?A -> ?B  <:  ?Nat -> ?Int] solves [Nat <: A <: Any] and [Non <: B <: Int] as [A := Nat, B := Int]:
     To achieve this, we treat [args] as covariant and [rets] as contravariant.
     *)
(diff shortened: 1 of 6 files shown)


src/mo_frontend/typing.ml
```diff
@@ -1659,8 +1663,9 @@ let sub_or_bimatch_func tbs args rets req_args req_rets =
   else
     let arg_subs = List.map2 (fun ra ea -> (ra, ea, no_region)) req_args args in
     let ret_subs = List.map2 (fun cr rr -> (cr, rr, no_region)) rets req_rets in
+    let ret_opt = Some (T.Func (T.Local, T.Returns, [], rets, args)) in (* Note: Flipped to get the maximal solution! *)
```
christoph-dfinity, line 1666: Could you explain what's going on here? The function type we're constructing here doesn't make sense right? Is this just to put type variables in co/contravariant positions?

Kamirus: yes, this is just the "polarity witness ret_opt"

christoph-dfinity (approved):

PR #6191, opened 2026-06-17
feat: disables static check for transient fields under enhanced migration

christoph-dfinity:
This means `transient let x = Map.empty<Nat>();` is now allowed again under `--enhanced-migration`


crusso:
I think one potential danger is that you can now immediately  overwrite the stable vars to lose data. The purity restriction prevented this.

More seriously, the real issue is that if you allow  side effects in the actor initializer then fast forwarding is not the same as single stepping the migrations, as you won't execute the side effects of the intermediate initializer. All the side effects should be restricted to the migration functions, which do get run. 

The `<system>` escape  hatch was to let you run odd system side-effects like setting timers.

Buyer beware

christoph-dfinity:
> I think one potential danger is that you can now immediately overwrite the stable vars to lose data.

As in this, overwriting whatever value was set in the migrations?
```
actor {
  var x : Nat;
  x := 20;
}
```
 We weren't worried about that with our previous migration strategy though. What changed?

> ...as single stepping the migrations, as you won't execute the side effects of the intermediate initializer.
All the side effects should be restricted to the migration functions, which do get run.

There are no side-effects in migration functions, right? They're synchronous and don't have `<system>` capabilities... Is this referring to a previous draft in which we compiled multiple actor definitions into a single migration function?

alexandru-uta (approved):
thanks!

PR #6192, opened 2026-06-17
feat(typing): structural implicit derivation for variants via `__variant`

Kamirus:
Structural implicit derivation previously covered records (`__record`) and tuples (`__tuple`) but not variants. This fills that gap: a single combiner now lets the compiler derive an implicit for any variant type whose case payloads have instances, removing per-variant boilerplate (serialization, `toText`, etc.).

A combiner is a function whose parameter is named `__variant` with type `(Text, () -> E) -> R`. Because a variant value is exactly one case at runtime, the synthesized wrapper switches on the active case and applies the combiner once to its `(tag, payload thunk)`:

```
func ($v) {
  combiner(switch ($v) {
    case (#t1 x) ("t1", func() { inst1(x) });
    ...
    case (#tn x) ("tn", func() { instn(x) });
  })
}
```

Each per-case implicit (`Ti -> E`) is resolved by the same search label, recursively, so nested types work (e.g. a variant whose payload is a record). A no-payload case `#t` has payload `()`, so it needs a `() -> E` instance — the same "every component type needs an instance" rule that already applies to record fields and tuple elements.

### Before / After

```motoko
func show(__variant : (Text, () -> Text)) : Text {
  let (tag, payload) = __variant;
  "#" # tag # "(" # payload() # ")"
};
module TextShow { public func show(self : Text) : Text = self };
module NatShow  { public func show(self : Nat)  : Text = debug_show self };

func inspect<T>(x : T, show : (implicit : T -> Text)) : Text = show(x);

type Shape = { #circle : Nat; #named : Text };
```

- Before: `inspect<Shape>(#named "hi")` — no candidate; must pass `show` explicitly.
- After: `inspect<Shape>(#named "hi")` → `"#named(hi)"`, derived automatically.

## Scope

Unary (`Var -> R`) only. Binary `(Var, Var) -> R` (e.g. `compare`/`equal`) is intentionally not derived structurally: the two values may be in different cases, so there is no per-case pairing and the cross-case ordering would require a separate, more complex combiner convention. Write those explicitly.


alexandru-uta (approved):
nice, thanks!

PR #6198, opened 2026-06-18
feat: warn on excess precision in `Float32` literals (M0266)

### LINKED ISSUE #5973: Warn on loss of precision for literal `Float32`
Rather, I'd avoid going to textual, as it is notoriously difficult to get right. Better convert to `F64` and `F32` in Wasm format, then promote the `F32` to `F64` and if that is different to the former we have loss of precision. That is a good reason to warn. But I wonder if that'd give lots of false positives.

ggreif:
Closes #5973.

A `Float32` literal written with more significant digits than F32 can hold has those digits **silently discarded** by rounding. The type checker now warns and suggests the shortest equivalent:

```
0.123456789 : Float32   →  M0266, rounds to 0.12345679
0.100000001 : Float32   →  M0266, rounds to 0.1
0.1  /  3.14  /  1.0     →  (no warning)
```

## The concept — *excess digits*, not *rounding distance*

The issue floated "convert to F32 and back, warn if it differs from the F64 by more than F32’s epsilon." That can’t work: rounding to F32 changes a value by **at most ½ ulp by definition**, so that difference is bounded regardless of how the literal was written. Concretely `0.1` and `0.1000000000000000001` parse to the *same* F64 — a value-only test cannot tell a minimal literal from an excess one, because the excess lives in the **digit string**, which parsing discards.

So the criterion is: **does the literal carry more significant digits than the F32 value needs?** We compute the shortest decimal that round-trips to the F32 value and warn only when the written literal is longer.

- Minimal literals (`0.1`, `3.14`, `1.5`, `1.0`) *are* their own shortest form → quiet. No false-positive storm.
- This is behaviourally Rust’s `clippy::excessive_precision` (on by default, precisely because shortest-round-trip doesn’t cry wolf).

## Implementation

`check_float32_precision` at the `Prim Float32, PreLit (s, …)` site in `typing.ml`. "Shortest" is **synthesised from primitives moc already has** — a 1..9-digit `Printf "%g"` / `Float32.of_string` round-trip search — so **no shortest-float (Ryu/Grisu) dependency** is added. (`Wasm`’s `%.17g` printer is fixed-width, not shortest, so it couldn’t serve directly.) Significant digits are counted off the lexeme (underscores / exponent / point / leading+trailing zeros normalised away, so `1.0`/`1e3`/`0.10` count as 1). Hex floats are not analysed.

Scope: `Float32` only (generalising to `Float`/F64 is the same code with `N ∈ 1..17`). Test: `run/float32-precision.mo` — `excess` warns, `minimal`/`exact` stay clean.


src/mo_frontend/typing.ml
```diff
@@ -1353,7 +1404,9 @@ let check_lit env t lit at suggest =
   | Prim Float, PreLit (s, (Nat | Int | Float)) ->
     lit := FloatLit (check_float env at s)
   | Prim Float32, PreLit (s, (Nat | Int | Float)) ->
-    lit := Float32Lit (check_float32 env at s)
+    let v = check_float32 env at s in
+    check_float32_precision env at s v;
```
Kamirus, line 1413: Just a sanity check: did you verify that we don't need to run the same check in the `infer_exp` path? I guess there we just default to normal float, right?

ggreif: We should basically do a `Float64` excess precision check too, but I deferred that to later. Inference will priorise the default, which is `Float64`.

ggreif: There is a single producer for `Float32Lit`, and it is this code path. So all is correct.

src/lang_utils/error_codes.ml
```diff
@@ -269,6 +269,7 @@ let warning_codes = [
   "M0244", None, "Mutable variable is never reassigned";
   "M0254", None, "Initial actor requires field";
   "M0265", None, "system capability not required by this mixin";
```
ggreif, line 271:
```suggestion
  "M0265", None, "`system` capability not required by this mixin";
```

test/fail/ok/warn-help.run-ir.ok
```diff
@@ -0,0 +1,46 @@
...
+M0236 (A) Suggest contextual dot notation
+M0237 (A) Suggest redundant explicit arguments
+M0239 (W) Avoid binding a unit `()` result
+M0240 (W) Unused identifier in shared pattern warning
+M0241 (W) Unused field in shared pattern warning
+M0242 (W) Implicit oneway declaration
+M0243 (W) Unreachable else in let-else
+M0244 (W) Mutable variable is never reassigned
+M0254 (W) Initial actor requires field
+M0265 (W) system capability not required by this mixin
```
ggreif:
```suggestion
M0265 (W)  The `system` capability is not required by this mixin
```

src/lang_utils/error_codes.ml
```diff
@@ -268,7 +268,8 @@ let warning_codes = [
   "M0243", None, "Unreachable else in let-else";
   "M0244", None, "Mutable variable is never reassigned";
   "M0254", None, "Initial actor requires field";
-  "M0265", None, "system capability not required by this mixin";
+  "M0265", None, "`system` capability not required by this mixin";
```
ggreif:
```suggestion
  "M0265", None, "The `system` capability is not required by this mixin";
```

test/fail/ok/warn-help.comp.ok
```diff
@@ -0,0 +1,46 @@
...
+M0236 (A) Suggest contextual dot notation
+M0237 (A) Suggest redundant explicit arguments
+M0239 (W) Avoid binding a unit `()` result
+M0240 (W) Unused identifier in shared pattern warning
+M0241 (W) Unused field in shared pattern warning
+M0242 (W) Implicit oneway declaration
+M0243 (W) Unreachable else in let-else
+M0244 (W) Mutable variable is never reassigned
+M0254 (W) Initial actor requires field
+M0265 (W) system capability not required by this mixin
```
ggreif:
```suggestion
M0265 (W) The `system` capability is not required by this mixin
```

test/fail/ok/warn-help.run-low.ok
```diff
@@ -0,0 +1,46 @@
...
+M0236 (A) Suggest contextual dot notation
+M0237 (A) Suggest redundant explicit arguments
+M0239 (W) Avoid binding a unit `()` result
+M0240 (W) Unused identifier in shared pattern warning
+M0241 (W) Unused field in shared pattern warning
+M0242 (W) Implicit oneway declaration
+M0243 (W) Unreachable else in let-else
+M0244 (W) Mutable variable is never reassigned
+M0254 (W) Initial actor requires field
+M0265 (W) system capability not required by this mixin
```
ggreif:
```suggestion
M0265 (W) The `system` capability is not required by this mixin
```

test/fail/ok/warn-help.run.ok
```diff
@@ -0,0 +1,46 @@
...
+M0236 (A) Suggest contextual dot notation
+M0237 (A) Suggest redundant explicit arguments
+M0239 (W) Avoid binding a unit `()` result
+M0240 (W) Unused identifier in shared pattern warning
+M0241 (W) Unused field in shared pattern warning
+M0242 (W) Implicit oneway declaration
+M0243 (W) Unreachable else in let-else
+M0244 (W) Mutable variable is never reassigned
+M0254 (W) Initial actor requires field
+M0265 (W) system capability not required by this mixin
```
ggreif:
```suggestion
M0265 (W) The `system` capability is not required by this mixin
```

test/fail/ok/warn-help.run-low.ok
christoph-dfinity, line 1: Don't think these `run` outputs should be here. This test lives in `test/fail`

test/run/float32.mo
```diff
@@ -1,5 +1,9 @@
 import { floatToFloat32; float32ToFloat; debugPrint } = "mo:⛔";
 
+// This test deliberately writes over-precise Float32 literals (e.g. a 21-digit
+// pi) to exercise rounding/conversion, so silence the excess-precision warning.
+//MOC-FLAG -A M0266
+
```
christoph-dfinity, line 5: Usually we put these flag comments at the top of the file, let's do that here as well for consistency?

christoph-dfinity (changes requested):
Good warning to add. The implementation looks a bit complicated, but as this is just a warning it's low risk. (I compared it to the clippy version, which basically roundtrips through parser and formatter and makes sure it comes out the same)

Some wrong test files have been committed, but otherwise LGTM

test/fail/ok/warn-help.run-low.ok
ggreif: #6219 is it now.

test/fail/ok/warn-help.run-ir.ok
ggreif: Moved.

christoph-dfinity (approved):
LGTM

PR #6206, opened 2026-06-19
chore: deprecate experimental multi-value flags; retire the `test-runner` MV axis

ggreif:
Multi-value Wasm codegen has been the default since #6165.

- Drop the legacy single-dash `-multi-value`/`-no-multi-value` aliases; deprecate `--(no-)experimental-multi-value` (still functional, now warn).
- Retire the FakeMV/WasmMV test-mode product from the test-runner (2nd commit) — persistence is now the only mode axis. The same shape will return as `TcMode` for the upcoming `--experimental-tailcalls` bifurcation.
- Fold the redundant `multi-value-experiment` tests into `multi-value.mo`.

Note: FakeMultiVal codegen loses test coverage (intended — the emulation is being retired).


christoph-dfinity (commented):
This is a breaking change, isn't it?

christoph-dfinity (approved):
Actually nevermind, looks like the flags still work, they just warn.

PR #6211, opened 2026-06-23
feat: allows requiring `<system>` capability on mixins

### LINKED ISSUE #5805: Expose `system` capability in mixin bodies

christoph-dfinity:
Allows defining mixins that require the `system` capability, which require a `<system>` to be passed when included.

The implementation is somewhat crude, as it is limited to `<system>` and does not permit any other type bounds (which greatly simplifies it).

Fixes #5805


src/mo_def/arrange.ml
```diff
@@ -298,8 +298,8 @@ module Make (Cfg : Config) = struct
         obj_sort s;
         id i
       ] @ List.map dec_field dfs)
-    | MixinD (_, dfs) -> "MixinD" $$ List.map dec_field dfs
-    | IncludeD (i, e, _) -> "IncludeD" $$ [id i; exp e]))
+    | MixinD (sys, p, dfs) -> "MixinD" $$ (if sys then [Atom "system"] else []) @ [pat p] @ List.map dec_field dfs
```
ggreif, line 301: Not a killer, but a bit of a smell, this is boolean-blind. If we ever wanted to allow the parametric degrees of freedom, we could use the corresponding data type now. The parser then would just fill in the `system`-related part. Just an idea...

src/mo_frontend/typing.ml
```diff
@@ -5030,13 +5040,13 @@ and infer_dec env dec : T.typ =
...
     T.normalize t
-  | MixinD (args, dec_fields) ->
+  | MixinD (sys, args, dec_fields) ->
     if not env.in_prog then
       error env dec.at "M0228" "mixins may only be declared at the top-level";
     let t_pat, ve = infer_pat_exhaustive error env args in
     let env' = adjoin_vals env ve in
     let obj_sort : obj_sort = { it = T.Mixin ; at = no_region; note = { it = true; at = no_region; note = [] } }  in
-    let t' = infer_obj { env' with check_unused = false } obj_sort None dec_fields dec.at in
+    let t' = infer_obj { env' with check_unused = false; async = if sys then C.SystemCap C.top_cap else C.NullCap } obj_sort None dec_fields dec.at in
```
ggreif, line 5066:
```suggestion
    let t' = infer_obj { env' with check_unused = false; async = C.(if sys then SystemCap top_cap else NullCap) } obj_sort None dec_fields dec.at in
```

src/mo_def/arrange.ml
ggreif: Not sure how much of the downstream machinery that would make reusable, though. You don't have much logic over there...
So maybe not an issue.

ggreif (approved):
Simple and effective, LGTM!

src/mo_def/arrange.ml
christoph-dfinity: Yeah, I started out with `typ_bind`, but it's basically just noise downstream, because none of the scope machinery applies at the moment. I think the best time to think about this probably is when actually implementing type parameters (if we ever decide that's needed).

src/mo_frontend/typing.ml
christoph-dfinity: Sorry, I prefer not having extra parenthesis here.

PR #6214, opened 2026-06-24
feat: emit standardized `target_features` Wasm custom section

### LINKED ISSUE #6213: moc should emit a standardized `target_features` Wasm custom section
## Problem

`moc` produces Wasm that uses non-MVP features (multi-value by default; `memory64` + 64-bit tables under enhanced orthogonal persistence; bulk-memory), but it does **not** emit a standardized `target_features` custom section.

As a result, every Binaryen-based consumer must be told about these features via per-tool flags:

- `wasm-opt` needs `--enable-multivalue` (etc.), else it errors.
- `dfx` `optimize = "cycles"`, `icp-cli` recipes, and any raw `wasm-opt` invocation hit the same wall.
- `ic-wasm optimize` **doesn't even expose** such a flag, so today the failure is unrecoverable through `dfx`.

Under enhanced orthogonal persistence there's additionally a `Tables may not be 64-bit` error from Binaryen — same root cause (no declared `memory64` feature).

## Why this regressed (Binaryen behavior)

Binaryen does **not** infer features by scanning the binary. With no `target_features` section it falls back to a fixed default — and that default changed: in Binaryen **≥ v122** the absent-section default went from "all features" to **MVP** ([changelog](https://github.com/WebAssembly/binaryen/blob/version_122/CHANGELOG.md): *"Change default feature set in the absence of a target features section from all features to MVP"*). So older Binaryen-based tools accepted moc output (defaulted to all-features); newer ones default to MVP and error on the first non-MVP instruction. `--detect-features` only means "read the section" — with no section it equals `--mvp-features`, so it doesn't help.

When a section **is** present (and no CLI feature flags), Binaryen uses **exactly MVP + the listed features** for validation and passes — authoritative, not additive on top of a broad default.

## Proposal

Emit the standardized `target_features` custom section, exactly as LLVM/Rust and Emscripten do. Binaryen reads this section and auto-configures its feature set on read, so **it fixes every Binaryen-based consumer at once with no per-tool flags**.

## Findings from a codebase pass (notes for whoever picks this up)

This is **not** greenfield — there's existing plumbing, but it's the wrong section and incomplete:

- A Motoko-private section named `wasm_features` already exists (`src/wasm-exts/customModuleEncode.ml`), emitted as comma-separated text and used only to raise warning `M0191` (`src/pipeline/pipeline.ml`). **Binaryen does not read this** — only the standardized `target_features` section. Keep it (the warning is useful) but populate from one source of truth.
- Feature tracking exists via `E.add_feature` / `E.get_features` (`FeatureSet`), but `add_feature` is currently called in **exactly one place** (`bulk-memory`/`multi-memory` for WASI/Wasm stable memory). **`multivalue` is on by default (`Flags.multi_value`) but is never registered** — the single most important feature is untracked today.

Per-claim verification:

- **multivalue** — needed; must be registered wherever `Flags.multi_value` drives codegen.
- **bulk-memory** — needed; RTS built with `-C target-feature=+bulk-memory` (`rts/Makefile`), and moc emits `MemoryCopy`/`MemoryFill`.
- **memory64 (enhanced OP)** — needed; EOP uses a 64-bit main memory. In Binaryen, 64-bit tables are gated by the `memory64` feature (no separate flag), so listing `memory64` should clear the `Tables may not be 64-bit` error too.
- **sign-ext** — **probably not needed from codegen, but harmless to include.** All "extend" ops in `src/codegen` are MVP `i64.extend_i32_s/u`, not the sign-extension extension (`i32.extend8_s`, …). It may still come from the linked RTS. Since over-listing is harmless (see caveat 1), include it if in doubt.

## Risks / caveats

1. **List a superset of what's used — under-listing just moves the error.** Because a present section is authoritative (MVP + exactly the listed features), omitting a used feature doesn't miscompile — it leaves the consumer broken on *that* feature instead, so the fix looks done but isn't. The asymmetry: **over-listing an unused feature is harmless** (enabling an unused feature passes validation), **under-listing is a latent break**. So derive the list from what moc **and the linked RTS** actually emit, and err toward inclusion when unsure (e.g. `sign-ext`).
2. **Name mapping required.** Internal names are wasmtime-style (`multi-memory`, `bulk-memory`); `target_features` uses Binaryen/LLVM names (`multivalue` no-hyphen, `memory64`, `sign-ext`). Need an explicit internal→Binaryen mapping. Do **not** blanket-emit `multi-memory` for IC output (IC stable memory uses the system API, not a second Wasm memory).
3. **Per-mode accuracy.** Classical vs enhanced-OP and IC vs WASI differ; compute the section per compilation (the `env.features` approach already supports this).
4. **Encoding.** `target_features` is a `vec` of (`'+'` prefix byte, length-delimited name) — not the comma-separated text used by the existing `wasm_features` section.

## Suggested scope

1. Determine the used feature set per mode empirically (e.g. dump the linked Wasm; check what `wasm-opt` needs vs MVP) and emit a superset of it — especially confirm `memory64` alone clears the table error.
2. Add the missing `add_feature` calls (multivalue, memory64) at the right codegen points and an internal→Binaryen name mapping.
3. Add a `target_features` encoder next to `wasm_features_section`, plus decoder + round-trip.
4. Tests/fixtures + Changelog entry.

Effort: small-to-moderate, low-risk. The encoder is a few lines; the real work is auditing the complete feature set.

Kamirus:
Binaryen-based tools (`wasm-opt`, `ic-wasm optimize`, `dfx`'s `optimize`) default to the MVP feature set when a module carries no `target_features` section. Motoko output uses `multivalue`, `bulk-memory`, `sign-ext`, `nontrapping-fptoint` (and `memory64` under enhanced orthogonal persistence), so these tools rejected it outright — and `ic-wasm optimize` has no flag to override, making the optimization unrecoverable.

`moc` now emits the standardized [`target_features`](https://github.com/WebAssembly/tool-conventions/blob/main/Linking.md#target-features-section) custom section listing exactly the features the emitted code and linked RTS use. Binaryen reads it and self-configures, so optimizing Motoko output needs no per-tool feature flags.

### Before / After

```console
$ moc -o canister.wasm canister.mo
$ wasm-opt -O3 canister.wasm -o opt.wasm
# Before: [wasm-validator error] ... memory.copy requires bulk memory [--enable-bulk-memory-opt]
# After:  ok
```

The declared set is empirically minimal: stripping the section makes the same module fail MVP validation again. It adapts to the build mode — `memory64` only under enhanced orthogonal persistence (classical is 32-bit), `multivalue` gated on `--multi-value`, and `multimemory` only in WASI/Wasm modes that emulate stable memory with a second linear memory (kept in sync with the existing dynamic `wasm_features` set).

## Scope

- Fully unblocks classical (32-bit) Motoko output for current `ic-wasm` (0.9.11).
- For enhanced-persistence (64-bit) output, the section is a prerequisite but not sufficient for `ic-wasm` 0.9.11: it bundles Binaryen 116, and 64-bit tables need Binaryen ≥118. That's a downstream toolchain bump tracked in [dfinity/ic-wasm#102](https://github.com/dfinity/ic-wasm/pull/102), not fixable here.

Closes #6213.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,10 +1,12 @@
 # Motoko compiler changelog
 
-## 1.10.1 (2026-06-24)
-
 * motoko (`moc`)
 
   * feat: `moc` now emits the standardized `target_features` Wasm custom section, so Binaryen-based tools (`wasm-opt`, `ic-wasm optimize`, `dfx`'s `optimize`) accept and optimize Motoko output without per-tool feature flags. Previously these tools defaulted to MVP and rejected the `multivalue`/`bulk-memory`/`memory64` features moc relies on (#6213).
+
+## 1.10.1 (2026-06-24)
+
+* motoko (`moc`)
 
   * bugfix: M0223 ("redundant type instantiation") and M0237 ("implicit argument can be omitted") no longer emit suggestions that are individually valid but break compilation when applied together. In nested calls where an inner instantiation or implicit is only inferable thanks to an outer one (e.g. `List.fromArray(Array.tabulate(...))`), only a jointly-applicable subset is now suggested, so `mops check --fix` no longer rewrites the code into an M0098 type error. The check is conservative: a few genuinely-redundant cases may go unreported in exchange for soundness (#6209).
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,7 +2,7 @@
 
 * motoko (`moc`)
 
-  * feat: `moc` now emits the standardized `target_features` Wasm custom section, so Binaryen-based tools (`wasm-opt`, `ic-wasm optimize`, `dfx`'s `optimize`) accept and optimize Motoko output without per-tool feature flags. Previously these tools defaulted to MVP and rejected the `multivalue`/`bulk-memory`/`memory64` features moc relies on (#6213).
+  * feat: `moc` now emits the standardized `target_features` Wasm custom section, so Binaryen-based tools (`wasm-opt`, `ic-wasm optimize`, `dfx`'s `optimize`) accept and optimize Motoko output without per-tool feature flags. Previously these tools defaulted to MVP and rejected the `multivalue`/`bulk-memory`/`memory64` features moc relies on (#6214).
 
 ## 1.10.1 (2026-06-24)
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # Motoko compiler changelog
 
+* motoko (`moc`)
+
+  * feat: `moc` now emits the standardized `target_features` Wasm custom section, so Binaryen-based tools (`wasm-opt`, `ic-wasm optimize`, `dfx`'s `optimize`) accept and optimize Motoko output without per-tool feature flags. Previously these tools defaulted to MVP and rejected the `multivalue`/`bulk-memory`/`memory64` features moc relies on (#6213).
+
 ## 1.10.0 (2026-06-19)
 
 * motoko (`moc`)
--- a/src/codegen/compile_classical.ml
+++ b/src/codegen/compile_classical.ml
@@ -13732,6 +13732,12 @@ and conclude_module env set_serialization_globals start_fi_o =
       };
       source_mapping_url = None;
       wasm_features = E.get_features env;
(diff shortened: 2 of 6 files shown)


ggreif (approved):
Looks awesome to me!

PR #6216, opened 2026-06-24
test: feature-aware Wasm validation via wasm-opt in run-test

Kamirus:
moc emits a `target_features` Wasm custom section ([#6214](https://github.com/caffeinelabs/motoko/pull/6214)) so Binaryen-based tools can self-configure. Nothing in CI actually checked that section: `run-test` validates with wabt's `wasm-validate --enable-memory64 --enable-multi-memory`, which ignores `target_features` and is told the feature set by hand. So if codegen started using a Wasm feature without declaring it (or the encoder broke), every test would still pass — the exact regression the section exists to prevent would go unnoticed until a downstream `wasm-opt`/`ic-wasm`/`dfx` user hit it.

This adds a flagless `wasm-opt` phase right after `wasm-validate`. Binaryen reads the `target_features` section and validates against *exactly* that set, so a declared-vs-used mismatch fails the test at the source. It runs on every compiled test across all modes (classical, enhanced-persistence `memory64`, WASI multi-memory), giving broad coverage rather than a hand-picked sample.

On failure the phase fails with a short hint:

```
wasm-opt rejected the module (exit 1): a used Wasm feature is likely missing from moc's `target_features` section.
Reproduce: wasm-opt _out/<test>.wasm
```

## Cost

Negligible. The phase runs **no optimization passes** — input validation happens on read, so `-O3` is unnecessary (it would only exercise Binaryen's optimizer, not moc). Measured per ~240 KB module: `wasm-validate` ~12 ms, no-pass `wasm-opt` ~38 ms, `wasm-opt -O3` ~185 ms. At ~26 ms over the existing validate, 8-way parallel, the added wall-time is a few seconds across the suite — lost in the noise next to `drun`/`pocket-ic`. (CI timing comparison to follow on this PR.)

## Notes

- Adds `pkgs.binaryen` to the dev shell and the shared test derivations; we always run tests via nix, so the tool is guaranteed present (no runtime gating).
- `wasm-validate` stays — the two are complementary: wabt checks structural validity allowing all features; `wasm-opt` checks that moc declared the features it uses.
- Binaryen ≥122 treats the section as authoritative (absent ⇒ MVP), which is what makes this strict; nixpkgs ships 129.

### DIFF
--- a/test-runner/src/run_test.rs
+++ b/test-runner/src/run_test.rs
@@ -575,6 +575,37 @@ fn handle_mo(ctx: &mut Ctx, cli: &Cli, src: &str, d: &Directives) {
         let mut valid = Command::new("wasm-validate");
         valid.args(["--enable-memory64", "--enable-multi-memory"]).arg(&wasm_path);
         let _ = run_phase(ctx, "valid", d, Some("wasm"), |outp| spawn_to_file(outp, valid));
+
+        // Feature-aware validation. `wasm-validate` above is told which features
+        // to allow via flags; Binaryen instead reads the `target_features`
+        // section moc emits and validates against exactly that set, so this
+        // catches codegen that uses a Wasm feature without declaring it. No
+        // optimization passes are needed — input validation runs on read. The
+        // exit code drives the phase; wasm-opt's output is discarded (it dumps
+        // the whole module on error), with a short hint written on failure.
+        let wasm_arg = wasm_path.clone();
+        let _ = run_phase(ctx, "opt", d, Some("wasm"), move |outp| {
+            let code = match Command::new("wasm-opt")
+                .arg("-q").arg(&wasm_arg).args(["-o", "/dev/null"])
+                .stdin(Stdio::null()).stdout(Stdio::null()).stderr(Stdio::null())
+                .status()
+            {
+                Ok(s) => s.code().unwrap_or(-1),
+                Err(_) => 127,
+            };
+            let hint = if code == 0 {
+                String::new()
+            } else {
+                format!(
+                    "wasm-opt rejected the module (exit {code}): a used Wasm feature is \
+                     likely missing from moc's `target_features` section.\n\
+                     Reproduce: wasm-opt {}\n",
+                    wasm_arg.display()
+                )
+            };
+            let _ = fs::write(outp, hint);
+            code
+        });
     }
 
     if wasm_path.exists() && !skip_running && d.check {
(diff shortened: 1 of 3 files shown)


Kamirus:
## CI timing

Short version: **CI wall-clock can't isolate this phase**, and a controlled local A/B puts the real cost at **~90 ms per compiled module (serial) — a few seconds per suite under CI's parallelism**, with zero recurring cost on cached runs.

### Why wall-clock comparison is misleading

nix caches test-derivation *outputs*, so unchanged code re-runs in seconds (the suite never executes). This PR changes `test-runner` and adds a dependency, so it forced a one-time **cold** re-execution of every suite on nixbuild.net, while recent baselines were warm cache hits. Concretely, the `tests (ubuntu-latest, release)` job: this PR **executed** `drun-release` (494 tests, 211 s) and `drun-eop-release` (494 tests, 274 s), whereas the #6214 baseline job emitted **no** run lines at all — pure cache hit. Comparing the two measures caching, not the phase. That cold cost is one-time (this PR); unrelated PRs cache-hit again.

### Controlled A/B (same machine, same suite, only diff = the `opt` phase)

Full `run/` suite, run serially, 267 compiled modules, nix `binaryen-129`:

| | wall |
|---|---|
| without `opt` phase | 167.5 s |
| with `opt` phase | 191.3 s |
| **delta** | **+23.8 s (~+90 ms/module)** |

Under CI parallelism (8-way for `run/`, `-j4` for `run-drun`) that's a few seconds per suite; for `pocket-ic`-driven `run-drun` it's dwarfed by execution. No optimization passes are run — this is parse + feature validation only; `-O3` would be ~5× this and buys nothing for the contract.

All checks passed green with the phase active across classical, enhanced-persistence (`memory64`), and WASI (`multimemory`) modes.

test-runner/src/run_test.rs
```diff
@@ -575,6 +575,37 @@ fn handle_mo(ctx: &mut Ctx, cli: &Cli, src: &str, d: &Directives) {
...
+        // Feature-aware validation. `wasm-validate` above is told which features
+        // to allow via flags; Binaryen instead reads the `target_features`
+        // section moc emits and validates against exactly that set, so this
+        // catches codegen that uses a Wasm feature without declaring it. No
+        // optimization passes are needed — input validation runs on read. The
+        // exit code drives the phase; wasm-opt's output is discarded (it dumps
+        // the whole module on error), with a short hint written on failure.
+        let wasm_arg = wasm_path.clone();
+        let _ = run_phase(ctx, "opt", d, Some("wasm"), move |outp| {
+            let code = match Command::new("wasm-opt")
```
ggreif, line 588: Why not just link against the `binaryen` crate and call it? But this is also okay!

ggreif: I know. The maintainer has stopped maintaining.

PR #6223, opened 2026-06-30
fix: fix codegen for nested mixins

christoph-dfinity:
This is required so that the stabilities and declarations line up when desugaring actor bodies.

We didn't notice this earlier because we only had a test in `test/fail` which didn't actually involve the code generator.

## Note for reviewer

If you review the commits individually, you can see the compiler crash on the commit that just adds the test.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,7 @@
 
 * motoko (`moc`)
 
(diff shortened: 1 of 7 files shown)


alexandru-uta (approved):
Thank you!

PR #6224, opened 2026-06-30
chore: parallelize the `test-coverage` target

ggreif:
## TODOs

- [ ] move coverage tests into the nightly runs because they are never looked at anyway :-)
- [ ] the natural home of coverage testing should be `test-runner`
- [x] we will need a few more tests to light up the dark corners of coverage, especially in the type-checker, in the middle end, and for the code generators — see #6225 for a first batch

## What

The coverage report was built strictly serially: `make coverage` ran `make -C <dir>` for all 11 test directories chained with `&&`, each invoking the full-glob serial `all:` recipe.

This reshapes the target so each directory is a `cov-%` prerequisite invoked as a recursive `$(MAKE)`. Run under `make -jN`, GNU make's **jobserver** is inherited by every sub-make, so the *total* number of concurrent tests — across directories **and** across files within a directory — is bounded at N. No oversubscription, no thread-pool coordination.

## Coverage fidelity

The per-directory target is chosen so the test set stays **identical** to before:

| dir | target | why |
|-----|--------|-----|
| `run-drun` | `parallel` | its `quick` (from `quick.mk`) silently omits `*.drun` — would lose all actor/upgrade/migration coverage |
| `cmp` | `all` | no per-file rule for `*.cmp`; stays on the (tiny) serial recipe |
| all others | `quick` | per-file rule covers the whole glob |

`bisect_ppx` already appends a unique per-process suffix to `$BISECT_FILE` — the serial run spawned thousands of `moc` processes sharing one prefix and aggregated fine — so concurrent writes introduce **no** collision risk.

## nix

Driven with `make -j"$NIX_BUILD_CORES"` plus `enableParallelBuilding`. The daemon sets `NIX_BUILD_CORES` to the core count granted to the build, so it adapts to whatever the builder provides (nixbuild.net or a runner) without any workflow change.


ggreif:
## Coverage build time: serial → parallel

Measured as the `test-coverage` derivation's wall-clock on nixbuild.net, taken from the build-log timestamps and validated against nix's own `building '…-test-coverage.drv'` marker (first build output lands within +0.2s of it in every case, so the window is exact).

| arch | serial — 1.11.0 (#6220) | serial — nested-mixins (#6223) | **parallel (this PR)** | speedup |
|------|------|------|------|---------|
| **x86_64** (`ubuntu-latest`) | 15m47s | 10m36s | **6m50s** | **1.6–2.3×** |
| **ARM64** (`ubuntu-24.04-arm`) | 23m09s | 23m19s | **9m24s** | **~2.5×** |

**ARM64** is the cleanest result: both independent serial baselines agree at ~23m, and the parallel build lands at **9m24s — ~2.5×, roughly 14 minutes saved** per run. The bigger ARM gain is consistent with those builders having more cores for the jobserver to fill.

**x86_64** serial varied between the two baselines (10m36s vs 15m47s — nixbuild machine-to-machine variance), but the parallel build at 6m50s beats even the faster serial baseline, so it's a solid **1.6–2.3×** win.

Notes:
- Both serial baselines ran the full sweep (~2950 build-log lines each). The 1.11.0 figure is from its final, successful run — the three earlier ones were cancelled by force-pushes.
- The parallel run shows far fewer log lines (202) only because the per-file `quick`/`parallel` targets wrap `run-test` in `chronic`, which suppresses output on success — the build window itself is confirmed real by the drv-start marker.
- Single parallel sample so far; x86 baseline carries machine variance. Direction and magnitude are unambiguous, and the wins land where the constituents are biggest (`run`, `run-drun`), as the jobserver design intends.

ggreif:
## Why this PR adds `--all-modes`: enhanced backend was structurally uncovered

The coverage report showed `compile_enhanced.ml` at **68%** vs `compile_classical.ml` at **92%**. Root cause: `make coverage` invoked `run-test` *without* `--all-modes`, so every unmarked test compiled single-pass under the harness default `--legacy-persistence` (classical). The bisect hit-count on the backend dispatch (`pipeline.ml`) confirmed it — classical **798×**, enhanced **34×**. Enhanced-backend coverage in CI otherwise comes only from the separate `run-eop-*` derivations, which are **not** constituents of the `coverage` derivation.

Exporting `EXTRA_RUNFLAGS=--all-modes` for the coverage run (appended only to the `.mo`/`.drun` moc-codegen recipes) makes unmarked tests compile in **both** persistence modes. Measured locally:

| | before | after `--all-modes` |
|---|---|---|
| `Compile_enhanced.compile` invocations | 34 | **1025** |
| `Compile_classical.compile` invocations | 798 | 800 |
| **`compile_enhanced.ml`** | **68%** (8717/12650) | **92%** (11713/12650) |
| `compile_classical.ml` | 92% (11550/12487) | 92% (11550/12487) |

Both backends now sit at **92%**. The residual 8% on enhanced is the same *kind* of gap classical has — genuinely feature-gated regions (`compile_prim_invocation`, bignum fast/slow paths, `compile_binop`), not a structural blind spot.

### Build-time impact (parallel + `--all-modes`, nixbuild.net)

`--all-modes` ~doubles the comp work, which the parallelization absorbs:

| arch | serial single-mode (master) | parallel single-mode | parallel + `--all-modes` |
|------|------|------|------|
| x86_64 | 10m36s – 15m47s | 6m50s | 23m23s |
| ARM64 | 23m09s – 23m19s | 9m24s | 18m18s |

On ARM64, the parallel **dual-backend** run (18m18s) is still **faster than the original serial single-backend** run (~23m) — double the coverage in less wall-clock.

The companion commit also makes `SKIP_RUNNING` skip `.drun` execution (pocket-ic), which the coverage build was spawning pointlessly — drun execution exercises the RTS, not the instrumented compilers, so it adds no coverage. That trims the build further and keeps it compilation-focused.

PR #6225, opened 2026-06-30
test: cover dark corners in type-checker, enhanced codegen, and middle end

ggreif:
## What

Adds targeted tests that light up previously-dark regions of the **type-checker**, **enhanced code generator**, and **middle end**. Each was found from the bisect-ppx coverage report and each fill was confirmed with an instrumented `moc` (before/after `data-count`), not just by construction.

This is the test-content counterpart to the coverage-infrastructure work in #6224 (parallelization + `--all-modes` + `SKIP_RUNNING` for `.drun`); it stands alone and needs none of it to pass.

## Tests

**Type-checker — and-pattern edge branches** (`src/mo_frontend/typing.ml`, −70 uncovered spans)
- `test/run/and-pattern-infer-shapes.mo` — `AndP` in inference position drives `is_explicit_pat`/`is_explicit_pat_field` across LitP/OptP/TagP/TupP/ObjP/nested-AndP/ValPF (previously only `AnnotP` was hit).
- `test/run/and-pattern-vis-pat.mo` — `vis_pat`/`vis_pat_field` over every composite pattern shape.
- `test/fail/and-pattern-type-nested{,-valp,-tup,-annot}.mo` — M0260 "type bound in both branches" with the type id nested inside AndP / ValPF / TupP / AnnotP, exercising `find_typ_id_at`'s recursion arms.

**Enhanced codegen — IC cost-estimation primitives** (`src/codegen/compile_enhanced.ml`)
- `test/run-drun/cost-prims-enhanced.mo` — exercises `SystemCyclesBurnPrim` and `OtherPrim "cost{Call,CreateCanister,HttpRequest,SignWithEcdsa,SignWithSchnorr}"` in `compile_prim_invocation`, all of which were uncovered (the existing `cost.drun` is `CLASSICAL-PERSISTENCE-ONLY`, so it never touches the enhanced backend). EOP-only, compile-phase only.

**Middle end — typed-int neutral-element lowering** (`src/lowering/desugar.ml`)
- `test/run/desugar-neutral.mo` — the add-`0`/mul-`1` identity-elimination arms of `neutral` for `Nat8`/`Int8`/…/`Nat64`/`Int64` literals (16 arms, previously all dark), plus the `Left MulOp` path.

## Notes on what was deliberately *not* added

- `apply_sign`'s `Int8Lit`…`Int64Lit` arms and `check_ir.ml`'s `error …` branches are **structurally unreachable** from valid source (the parser folds `-128` into a `LitP` before `apply_sign`; `check_ir`'s error arms only fire on malformed IR the type-checker already rejects). They are left dark on purpose.
- The `-128:Int8` / `128:Int8` boundary asymmetry is already pinned by `test/run/neg-boundary.mo` + `test/fail/word-range.mo`, so no new boundary test was added.


test/run/ok/type-close.wasm-run.ok
```diff
@@ -0,0 +1,10 @@
+assertion failed at type-close.mo:43.1-43.62
+Error: failed to run main module `_out/type-close.wasm`
```
christoph-dfinity: Should these be trapping?

ggreif: No, they lipped in due to a wrong test! Fixed now...

ggreif:
@christoph-dfinity I have consolidated the test files where possible, but leaving thematic grouping intact. Much less noise.

christoph-dfinity (approved):
Thanks, this looks more reasonable!

PR #6228, opened 2026-07-01
feat: `actor { … }` <: `Principal`

### LINKED ISSUE #2264: actor t <: principal
Users currently have to write `Principal.fromActor(Foo)` to get the `principal` of an actor (including their own `self`). This is a bit silly, since an actor reference inherently is represented by a principal (in a way, it’s a principal with some phantom type information attached).

So we could add `actor t <: principal` to our subtyping rules.

Benefit: No need for `Principal.fromActor(Foo)`.

Implication: Representation has to be the same for `actor …` and `principal`, but that’s true now, and I don’t see us changing that soon.

Interaction with Candid: Darn, here is an interaction. As per https://github.com/dfinity/motoko/blob/master/design/IDL-Motoko.md we need

    ∀ t1 t2 : dom(e), t1 <: t2 ⟹ e(t1) <: e(t2)

i.e. Motoko subtyping must be contained in IDL subtyping. So we can only add this to Motoko’s subtyping if we also add it to Candid’s subtyping, which does not have `service <: principal …` yet.

ggreif:
Implements nomeata's idea from (Resolves #2264) — an actor reference upcasts to its canister-id principal, so you no longer need `Principal.fromActor(Foo)` (or `Prim.principalOfActor`).

**Status:** the type rule plus the matching runtime support (compiled backends + Candid) are complete and **CI-green**; interpreter parity and docs/changelog remain (see below).

## The type rule (`type.ml`)

- **`rel_typ`**: `Obj (Actor, _, _) <: Prim Principal` — subtyping-only and Actor-sort-only. The reverse (`Principal ⊀ actor`) stays rejected, so the downcast remains explicit and unsafe.
- **`combine`**: `lub(actor, Principal) = Principal`, `glb(actor, Principal) = actor`. The existing actor-vs-actor `lub` already yields a common actor interface (itself `<: Principal`) — left as-is, since it's *more* precise than widening to `Principal`.
- **Coercion-free (identity)**: as #2264 notes, an actor and a principal have the **same representation** (a `Vanilla` tagged blob differing only by heap tag), so the upcast inserts no coercion.

## Runtime realization — `service <: principal` at all three checkers

The rule bites at three distinct runtime sites; each gets a minimal arm mirroring the type rule — covariant/widening only, so the reverse (`principal` as `service`) stays rejected:

| Site | Where | Arm |
|---|---|---|
| `.did` wire, **top-level** `principal` | `compile_enhanced.ml` deser `Prim Principal` case | also accept an IDL `service` (composite `idl_service`), reading the wire-identical byte-tagged principal |
| `.did` wire, **nested** in a func signature | `rts/motoko-rts/src/idl.rs` `sub` (`idl_sub`) | `(IDL_CON_service, IDL_REF_principal) => true`, parallel to `nat <: int` |
| **EOP stable upgrade** | `rts/motoko-rts/src/idl.rs` `memory_compatible` | `(IDL_CON_service, IDL_REF_principal) => variance != Invariance` |

`flake.lock` bumps `candid-src` to the rev that carries the upstream `service <: principal` rule and its test vectors.

## Tests

- **Type system**: `test/run/actor-sub-principal.mo` — upcast in arg / return / first-class / `lub` / array positions, through all phases (`tc … wasm-run`); `test/fail/actor-sub-principal.mo` — downcast rejected (M0096), plain **object** and **module** are *not* `<: Principal`.
- **Candid conformance** (bumped suite): **467 tests, 458 ok, 0 failed** — including `reference:31/32/42` (top-level, incl. recursive service) and `subtypes:162/164/170` (`service<:principal` nested in a func result). Guardrails hold: `subtypes:166 principal </: service {}` rejected; `reference:46 principal: reject func wire type` traps — moc accepts **only** `service` at `principal`, matching the spec (`func` is *not* `<: principal`, so no func arm was added).
- **Both backends**: `test/run-drun/actor-sub-principal-stable.drun` — a stable var typed `actor {}` in v0 migrates into a `Principal`-typed var in v1 across an actor-class upgrade. Previously `# CLASSICAL-PERSISTENCE-ONLY`; the `memory_compatible` arm makes the EOP in-place reinterpretation succeed, so it now runs on **classical and EOP** (both print `aaaaa-aa`).
- **CI**: full `build` suite green on every leg — `common-tests`, `gc-tests`, `tests` (release/debug) × {ubuntu-latest, ubuntu-24.04-arm}.

## The Candid crux (resolved)

#2264 flagged that Motoko subtyping must be contained in IDL subtyping (`∀ t1 t2, t1 <: t2 ⟹ e(t1) <: e(t2)`), and Candid historically had **no `service <: principal`**. That precondition is now satisfied: upstream Candid carries the rule — a **conservative extension** (it only enlarges `<:`, changes no existing encoding, and is decode-sound because a `service` reference is already encoded as its principal bytes, so decoders merely become strictly more accepting). With the rule in Candid, moc's three arms above are justified for Candid interface subtyping — service imports and actor-class upgrade compatibility — not merely Motoko-internal.

## Remaining

- **Interpreter parity** — identity on interpreter values (the compiled backends + Candid paths are done and CI-verified; the interpreter value path is the last audit item).
- **Docs + changelog.**

(`lub`-widening golden reconciliation is covered by the green suite; the earlier "Candid soundness guard" item is moot now that the precondition holds.)


ggreif:
## Experience report: stable-var upgrade `actor {} → Principal` (run-drun, `e5e1acd8a`)

Added `test/run-drun/actor-sub-principal-stable` exercising an actor-class upgrade where a stable var typed `actor {}` in v0 migrates into a `Principal`-typed var in v1. The two backends diverge:

- **Classical: passes.** Classical re-serialises the stable record through Candid (stabilize/destabilize); the old `service`-encoded value reads back as a `principal`. The representation coincides (an actor reference *is* its canister-id principal — no `A`→`P` re-tag), so it round-trips. Value survives (`aaaaa-aa` before and after).
- **Enhanced (EOP): traps** — `IC0503: RTS error: Memory-incompatible program upgrade`.

### Why EOP rejects it — and the connection to the Candid gap

EOP doesn't re-serialise; it gates the upgrade with the RTS `memory_compatible` check (`rts/motoko-rts/src/idl.rs:673`, from `persistence.rs:249`). That check **is a Candid-subtyping checker over the serialized type table** (it walks `IDL_CON_opt/vec/func/record/service` and `IDL_PRIM_*` with variance). It has **no `(IDL_CON_service, IDL_PRIM_principal)` case**, so the upgrade falls through to "incompatible" and traps.

So the missing Candid rule `service <: principal` (#2264) bites in **two** places: the `.did` wire *and* the EOP upgrade gate. The type-system change in this PR is necessary but **not sufficient** for EOP stable upgrades.

The EOP fix is a one-case addition mirroring the type rule:
```rust
(IDL_CON_service, IDL_PRIM_principal) => variance != TypeVariance::Invariance,
```
…but it governs what in-place reinterpretations upgrades may perform, so it should land deliberately, in lockstep with the Candid spec rule.

### Plan

- Pursue `service <: principal` in the Candid definition. It's a **conservative extension**: it only enlarges `<:` (never removes a pair), changes no existing encoding, and is decode-sound because a `service` reference is already encoded as its principal bytes — the decoder merely becomes strictly more accepting. Nothing that typechecked/decoded before stops doing so.
- Land the matching `memory_compatible` case + flip the run-drun test to run on both backends, in lockstep.

The run-drun test is `# CLASSICAL-PERSISTENCE-ONLY` for now, documenting the working path and the EOP gap in its header.

ggreif:
## Verified EOP fix (local, not committed here)

Applied the one-case `memory_compatible` addition locally and rebuilt the RTS — the EOP `actor {} → Principal` stable-var upgrade then **passes**, matching the classical golden exactly (both print `aaaaa-aa`). So the type rule + this single RTS case fully enable EOP stable upgrades for the new subtyping.

```diff
--- a/rts/motoko-rts/src/idl.rs
+++ b/rts/motoko-rts/src/idl.rs
@@ -987,6 +987,9 @@ pub(crate) unsafe fn memory_compatible(
             }
             variance != TypeVariance::Invariance || n1 == 0
         }
+        // An actor reference upcasts to its canister-id principal (service <: principal),
+        // mirroring the Motoko type rule (type.ml rel_typ). Widening — sub direction only.
+        (IDL_CON_service, IDL_REF_principal) => variance != TypeVariance::Invariance,
         // default
         (_, _) => false,
     }
```

`IDL_REF_principal` (= -24) is the principal reference code; `IDL_CON_service` (= -23) is the service/actor constructor. The `variance != Invariance` guard matches the existing widening arms (e.g. `nat <: int`), so it fires only in the sub (covariant) direction — the reverse (`principal <: service`) stays rejected.

**Reverted locally** — this lands in lockstep with the Candid `service <: principal` rule, at which point the run-drun test drops its `# CLASSICAL-PERSISTENCE-ONLY` marker and runs on both backends.

ggreif:
## `service <: principal` realised at runtime — candid suite green (commit `91b7813cc`)

Now that upstream Candid carries the rule (bumped `candid-src` → `d24fb18`), I implemented the runtime counterpart of this PR's `actor {} <: Principal` type rule. **Key finding: the gap bites at *three* distinct runtime checkers**, not two — each needed its own arm:

| Site | Checker | Fix | Candid tests fixed |
|---|---|---|---|
| `.did` wire, **top-level** `principal` | deserializer `Prim Principal` case (`compile_enhanced.ml`) | accept an IDL `service` (composite `idl_service`), reading the wire-identical byte-tagged principal | `reference:31/32/42` |
| `.did` wire, **nested** in a func signature | RTS `sub` / `idl_sub` relation (`idl.rs`) | `(IDL_CON_service, IDL_REF_principal) => true`, a covariant widening arm parallel to `nat <: int` | `subtypes:162/164/170` |
| **EOP stable upgrade** | RTS `memory_compatible` gate (`idl.rs`) | `(IDL_CON_service, IDL_REF_principal) => variance != Invariance` (the arm verified earlier) | the `actor{}→Principal` upgrade |

The `memory_compatible` arm alone left `subtypes:*` failing — proving the deserializer routes func/service structural checks through `idl_sub`, **not** the upgrade gate. (RTS confirmed recompiled: the build rebuilt `moc-rts.drv`, so the arms were genuinely in the moc under test.)

### Verification
- **Candid suite ran green**: `467 tests: 458 ok, 9 ignored, 0 failed`.
- **Guardrails hold** (not just the happy path):
  - `subtypes:166 principal </: service {}` → **rejected** — the reverse stays unsound-free; the `sub` arm is covariant-only, contravariance handled by `sub`'s type-swap.
  - `reference:46 principal: reject func wire type` → **traps** — moc accepts **only** `service` at `principal`, matching the spec (`func` is *not* `<: principal`), so no func arm was added.
  - record/opt wire types at `principal` still rejected.
- **EOP `.drun` upgrade**: relied on the earlier local verification ("both print `aaaaa-aa`") + the arm confirmed compiled; flipped `actor-sub-principal-stable.drun` off `# CLASSICAL-PERSISTENCE-ONLY` so it now runs on both backends.

Watching CI on `91b7813cc`.

.claude/plans/actor-sub-principal.md
```diff
@@ -0,0 +1,149 @@
+# Plan: make `Principal` a supertype of every `actor { … }`
```
ggreif: Remove before merging!

src/mo_types/type.ml
```diff
@@ -1288,6 +1288,11 @@ let rec rel_typ d rel eq t1 t2 =
     (s1 = s2 || incompatible_obj_sorts d s1 s2) &&
     rel_fields t1 t2 d rel eq fs1 fs2 &&
     rel_typ_fields t1 t2 d rel eq tfs1 tfs2
+  | Obj (Actor, _, _), Prim Principal when rel != eq ->
+    (* An actor reference upcasts to its canister-id principal. Coercion-free:
+       both are the same Vanilla tagged blob, differing only by heap tag.
+       One-directional — Principal ⊀ actor keeps the downcast explicit. *)
+    true
```
ggreif, line 1292:
```suggestion
  | Obj (Actor, _, _), Prim Principal when rel != eq ->
    true
```

src/codegen/compile_enhanced.ml
```diff
@@ -8526,13 +8526,27 @@ module Serialization = struct
           Internals.dedup env          (* Call dedup *)
         )
       | Prim Principal ->
-        with_prim_typ t
-        begin
+        (* Candid `service <: principal`: accept either an IDL `principal`
+           (prim -24) or a `service` reference (composite `idl_service`). A
+           service value is wire-identical to a principal — a byte-tagged
+           principal blob (cf. `read_actor_data`), differing only in the heap
+           tag — so both decode here to a `Principal` (Tagged.P). *)
```
```suggestion
        (* Candid `service <: principal`: accept either an IDL `principal`
           or a `service` reference. They are wire-identical, differing only in
           the heap tag — so both decode here to a `Principal` (Tagged.P). *)
```

ggreif:
## On the inference concern — you're each partly right (verified on this branch's `moc`)

Adding `actor <: Principal` **is** conservative for the two things "purely additive" usually means:

- **Subtype *checking*** — monotone: you only add pairs to `<:`, so no check that passed can start failing.
- **Inference *satisfiability*** — also monotone: bi-matching never *loses* a solution, so nothing "fails to get inferred" that inferred before.

**But inference doesn't only check — it computes a `lub`/`glb` *witness*, and the witness moves:**

| | before | after |
|---|---|---|
| `lub(actor, Principal)` | `Any` | `Principal` |
| `glb(actor, Principal)` | `Non` | `actor` |

A moved witness is **safe in covariant and contravariant positions** (covariantly `Principal <: Any` fits anywhere `Any` did; contravariantly the requirement only *weakens*). It is **breaking only in an *invariant* position** — and the failure mode is *over-precise* inference (`Principal` instead of `Any`) tripping a later invariant use, **not** "instantiation can't be inferred."

### Minimal regressions (all compiled before the rule; all fail with it now)

```motoko
// 1. direct lub binding, invariant use — `var x` gets type `lub`, fixed eagerly
var x = if b anActor else aPrincipal;   // was : Any, now : Principal
x := ("hello" : Text);                  // M0096: Text cannot produce expected type Principal

// 2. mutable array literal (invariant `[var _]`)
let m = [var anActor, aPrincipal];      // was [var Any], now [var Principal]
m[0] := ("hello" : Text);               // M0096

// 3. bi-matching with the invariant occurrence in the RESULT (nothing pins T)
func mk<T>(x : T, y : T) : [var T] = [var x, y];
let arr = mk(anActor, aPrincipal);      // T = lub = Principal (was Any)
arr[0] := ("hello" : Text);             // M0096
```

### The corner that stays safe — bounds the blast radius

```motoko
// invariant occurrence in an ARGUMENT pins T; bi-matching solves jointly, so T = Any. Still compiles.
func g<T>(x : T, y : T, m : [var T]) {};
g(anActor, aPrincipal, (someVarAnyArray : [var Any]));   // ok
```

So it isn't "bi-matching" as such — joint solving with an invariant *argument* dominates the `lub`. The dangerous shape is an **unpinned invariant occurrence**: the result/output of a `lub`-instantiated call, or a direct `var`/array-literal binding over an actor and a principal.

### Two more notes

- **The `glb` dual** (which I haven't exercised yet, worth a test): a type var with two upper bounds `T <: Principal` and `T <: actor {}`. Its witness flips `Non → actor`; breaking if that now-`actor` witness flows *covariantly* into a non-actor context where `Non` (bottom) previously fit. Symmetric to #1–3, likely rarer.
- **No hidden "sealed" assumption in the inference code** — I scanned `type.ml`; there is no special-case treating `Principal` (or primitives) as having no proper subtypes. The entire effect flows through `combine` (`lub`/`glb`) + `rel_typ` — the functions already touched here — so the surface is fully accounted for; there's no separate matcher shortcut that silently assumed `Principal` sealed.

**Net:** conservative for *checking* and *satisfiability*, not for *inferred witnesses*; the only breakage is a `lub`/`glb` result landing in an **invariant** position, surfacing as over-precise inference failing a later invariant assignment. Repros #1–3 are the regression tests to add. If that (contrived) breakage is acceptable, it's a **bounded, documentable** one — a Changelog `breaking:` note rather than a silent surprise. Happy to add the tests + a changelog caveat, or to gate the rule if you'd rather not take even the contrived break.

.claude/plans/actor-sub-principal.md
ggreif: Done!

Kamirus:
The reason why this PR is/should be a breaking change is that allowing `actor { .. } <: Principal` implies we have to change `has_no_subtypes_or_supertypes` function and remove Principal from there, because it would be no longer true that Principal has no proper sub/super-types except `Non` and `Any`.
That change would mean that some type instantiations stop being redundant, e.g. creating a mutable container with Principals compiles now without type instantiation, but after the change it would show the "under-constrained type error" with `there is no "best" choice for type parameter`.

So the question is: is this PR worth it? Trading `fromActor` explicit calls for explicit instantiations. I'd say it's not worth it. We could choose to ignore updating the `has_no_subtypes_or_supertypes`, assume that `actor ...` is not a "proper"  subtype of Principal, and whenever the compiler has `Non <: T <: Principal` (and `T` is invariant) to solve it would arbitrarily pick Principal as the solution. I this that is the pragmatic solution if we want to keep this PR and be consistent with Candid rules

PR #6242, opened 2026-07-15
fix: restrict implicit lib resolution to --implicit-package

Kamirus:
With `--implicit-package core`, implicit arguments and contextual dot should resolve from core modules only. Transitive imports from another package (e.g. `import Extra "mo:base/Extra"`) still load that package's dependencies into `env.libs`, so `base/Text.compare` could compete with `core/Text.compare` and produce an ambiguous `M0230` even though the flag names a single package.

Import-time filtering in `resolve_import.ml` skips bulk-loading non-implicit packages, but cannot exclude modules pulled in transitively. Typing now filters lib candidates with `Suggest.is_implicit_lib`, which checks whether a loaded lib path lies under the `--implicit-package` root (same rule as `mo:pkg/...` URL resolution).

**Before**

```motoko
//MOC-FLAG --package core ../core-stub/src
//MOC-FLAG --package base ../base-stub/src
//MOC-FLAG --implicit-package core
import Map "mo:core/Map";
import Extra "mo:base/Extra";

map.add("abc", 3);
// M0230: Cannot determine implicit argument `compare`
// note: Did you mean to import mo:base/Text or mo:core/Text?
```

**After**

```motoko
// same setup — compiles; compare resolves from core/Text only
map.add("abc", 3);
```

## What is unchanged

Explicit imports remain in scope regardless of package. Error suggestions may still mention non-implicit-package libs as import hints; only resolution is restricted. Storing package metadata on `lib_env` entries (see pipeline TODO) is a possible follow-up but out of scope here.


src/mo_frontend/scope.ml
```diff
@@ -54,8 +55,8 @@ let adjoin scope1 scope2 =
 
 let adjoin_val_env scope ve = {scope with val_env = T.Env.adjoin scope.val_env ve}
 
-let lib f t =
-  { empty with lib_env = T.Env.singleton f t }
+let lib ?package path typ =
```
christoph-dfinity: I think I'd rather make this a required argument. Too easy to forget otherwise.

src/mo_frontend/typing.ml
```diff
@@ -2317,20 +2335,21 @@ and infer_exp'' env exp : T.typ =
       let candidate_libs =
         if Option.is_some(!Flags.implicit_package) then
           T.Env.to_seq env.libs |>
-            Seq.filter (fun (name, typ) ->
-              name <> "@prim" &&
+            Seq.filter (fun (name, info) ->
+              is_implicit_package info.lib_package &&
```
christoph-dfinity, line 2346: This is always false when `!Flags.implicit_package` is None, so the surrounding `if` check is redundant, right?

src/mo_frontend/typing.ml
```diff
@@ -113,6 +114,16 @@ let env_of_scope msgs scope =
     enclosing_removal = false;
   }
 
+let is_implicit_package pkg =
+  match !Flags.implicit_package, pkg with
+  | Some ip, Some p -> ip = p
+  | _ -> false
+
+let implicit_lib_p env path =
```
christoph-dfinity: Assming the `_p`  is for "predicate"? Is that a convention we use anywhere else? Feels a bit lisp-y

src/mo_frontend/typing.ml
```diff
@@ -2188,8 +2205,9 @@ let contextual_dot env name receiver_ty : (ctx_dot_candidate, 'a context_dot_err
       let modules = String.concat ", " (List.filter_map (fun c -> c.module_ref) cs) in
       error env name.at "M0224" "overlapping resolution for `%s` in scope from these modules: %s" name.it modules))
     | `Empty ->
+      (* Resolve only implicit-package libs; error suggestions may still list others. *)
       let lib_candidates = candidates true env.libs is_lib_module in
-      match if Option.is_some !Flags.implicit_package then disambiguate_candidates lib_candidates else `Empty with
+      match if Option.is_some !Flags.implicit_package then disambiguate_candidates (List.filter (fun c -> Option.fold ~none:false ~some:(implicit_lib_p env) c.module_ref) lib_candidates) else `Empty with
```
christoph-dfinity: Might be time to move this into a `let`. Large expressions as a match scrutinee are a bit weird to read.

christoph-dfinity (changes requested):
Mostly stylistic nits. Looks good in general!

src/mo_frontend/typing.ml
Kamirus: Just to short-circuit the computation, when the flag is off there is no need to compute.

src/mo_frontend/typing.ml
Kamirus: true, changed to `is_implicit_lib`

christoph-dfinity (approved):

PR #6249, opened 2026-07-21
feat: `--stable-baseline` turns forgotten EM fields into M0267

Kamirus:
Tracks [LANG-1348](https://linear.app/caffeinelabs/issue/LANG-1348/check-time-stable-baseline-for-em-m0267-single-invocation-stable).

Converted legacy→enhanced-migration projects can't promote M0254 to an error: the migration chain often still "requires" fields that already lived on the last deployed canister. Agents then add a new actor field without a migration, get only a soft warning, and fail at deploy time.

`--stable-baseline <last.most>` keeps the same `initial_required` set: fields explained by the baseline (stable subtype) stay **M0254**, and unexplained ones escalate to error **M0267**.

### Before / After

```bash
# converted project: field `x` already on last deploy, no Init for it
moc --check --enhanced-migration migrations/ actor.mo
# warning [M0254], initial actor requires field `x`

moc --check --enhanced-migration migrations/ \
  --stable-baseline last.most actor.mo
# warning [M0254] for `x` (carried by baseline)

# agent adds `email` with no migration
moc --check --enhanced-migration migrations/ \
  --stable-baseline last.most actor.mo
# type error [M0267] on the actor field `email` ...
```

### Diagnostic location (intentional)

**M0254** stays on the first migration file (same as today — “required at the start of the chain”).

**M0267** points at the **actor field** instead. The walked “first migration” locus would only guess `m1.mo:1:1`; for a forgotten new field we want the declaration the agent just added, not the oldest migration in the chain.

### What is unchanged

- No baseline → today's M0254 warning only
- `--stable-compatible` upgrade checking is unchanged
- Prototype: no mops/caffeine wiring yet; `moc.js` does not share the CLI flag guard

Follow-up (check-time stable-compat / single-invocation mops): [#6253](https://github.com/caffeinelabs/motoko/pull/6253).


alexandru-uta (approved):
useful feature, thanks!

PR #6251, opened 2026-07-23
chore(test): remove 17 orphan `.ok` files; resurrect loss0-1 stable-compat test

ggreif:
Standard gardening of `test/*/ok`. A repo-wide sweep of all 14 `test/*/ok` directories found exactly 18 expected-output files with **no backing source** (`.mo`/`.cmp`/`.drun`/…) — all in `run`/`fail`/`cmp`/`run-drun`; the other directories are clean.

The harness only reads `ok/NAME.*.ok` for an existing source `NAME.*`, so a `.ok` with no source never participates in any test.

**17** are genuinely dead and are removed:
- `test/run`: `pkg-import-relative-diffcase`
- `test/fail`: `actor-match`, `system`
- `test/cmp`: `bad`
- `test/run-drun`: `id-sub-ho`, `mod-rebind`, `self-describe`, `stable-mem-warning`, `stable-region-misc`, `tuple-message`, `weak-classical`

The **18th** (`test/cmp/ok/loss0-1.cmp.ok`) is **not** dead. It was the sole witness of error code `M0216`; its `.most` fixtures still exist, and only its one-line `.cmp` driver was never committed. So rather than delete it, this PR **restores the driver** (`test/cmp/loss0-1.cmp`), turning phantom coverage into a real, emitting test. See the comment below for detail.

Independent of #6225.


ggreif:
Note on why this PR restores one file instead of removing all 18 candidates.

While confirming each `.ok` was truly dead, `test/cmp/ok/loss0-1.cmp.ok` turned out to be the **sole witness of error code `M0216`** ("the new type of stable variable implicitly drops data", `mo_frontend/stability.ml`). Removing it made CI's `check-error-codes` fail with `untested error codes: {'M0216'}`.

It is not really an orphan: both fixtures (`test/cmp/files/loss0.most`, `loss1.most`) and the expected output (`loss0-1.cmp.ok`) are present — only the one-line `.cmp` driver (`files/loss0.most files/loss1.most`) was never committed, so the test has simply never run. `moc --stable-compatible loss0.most loss1.most` does emit M0216 (twice), matching the committed `.ok` byte-for-byte.

It stayed invisible because `check-error-codes.py` scans `.ok` file *contents* for codes rather than codes emitted by *executed* tests — the `M0216` string in the stray `.ok` counted as coverage, so nothing flagged the absent driver.

So this PR **restores the driver** (`test/cmp/loss0-1.cmp`) rather than deleting the `.ok`, turning phantom coverage into a real, emitting test. Net: **17** genuine orphans removed, `loss0-1` resurrected.

Possible follow-up (out of scope here): having `check-error-codes.py` key off executed-test output rather than `.ok` contents would close the gap that let a driver-less fixture look "covered".

christoph-dfinity (approved):

PR #6255, opened 2026-07-24
fix: refresh broken docs links in compatibility errors

Kamirus:
Compatibility and Region diagnostics pointed at old `internetcomputer.org/docs/...` URLs that now redirect to the marketing homepage, so the “please see …” links in upgrade errors were dead.

They now use the current docs:

- Explicit migration: [compatibility § migration function](https://docs.internetcomputer.org/languages/motoko/fundamentals/actors/compatibility/#explicit-migration-using-a-migration-function)
- Leaving enhanced multi-migration (M0255): [enhanced multi-migration](https://docs.internetcomputer.org/languages/motoko/fundamentals/actors/enhanced-multi-migration/)
- Deprecated `ExperimentalStableMemory` (M0199): [stable memory and regions](https://docs.internetcomputer.org/languages/motoko/icp-features/stable-memory/)

**Before (M0169):**
```
the stable variable `three` of the previous version cannot be implicitly discarded. The variable can only be dropped by an explicit migration function, please see https://internetcomputer.org/docs/motoko/fundamentals/actors/compatibility#explicit-migration-using-a-migration-function
```

**After:**
```
stable variable `three` of the previous version cannot be implicitly discarded — use an explicit migration function to drop it
See https://docs.internetcomputer.org/languages/motoko/fundamentals/actors/compatibility/#explicit-migration-using-a-migration-function
```

Same pass also shortens the related M0170 / M0216 / M0263 / M0255 messages along those lines.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,8 @@
     stable subtype of the required type keep warning M0254 (prototype for legacy→EM
     conversions) (#6249).
 
+  * fix: refresh the broken docs links in compatibility and stable-memory diagnostics (#6255).
+
 ## 1.11.2 (2026-07-22)
 
 * motoko (`moc`)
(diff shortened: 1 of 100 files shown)


ggreif (approved):
Verified the links, LGTM!

PR #6256, opened 2026-07-24
ci: align AI PR review to shared reusable workflow (comment-only + ponytail)

Kamirus:
motoko's AI PR review was a ~390-line self-contained copy of logic that now lives centrally in `caffeinelabs/devtools` (`reusable-ai-pr-review.yml`). Swapping the copy for a thin wrapper that calls the shared workflow — the setup `app` already uses — keeps Cursor install, diff materialization, verdict parsing, commenting, and the ponytail pass maintained in one place instead of drifting per repo.

Review behavior is unchanged in intent: advisory only. The AI posts a verdict comment but never submits a GitHub approving review (`enable_approval: false`), so a human is still required to approve and merge. The over-engineering "ponytail" pass is now enabled as a separate non-gating comment. motoko keeps its own compiler-specific prompt via `prompt_path`, and the model is pinned to `grok-4.5` to match `app`.

## What changes behaviorally

- Trigger moves from `pull_request` to `pull_request_target` so the review can read secrets; fork PRs are still gated out inside the reusable workflow.
- The check only goes red on `REVIEW_ERROR` / incomplete runs, no longer on `REQUEST_CHANGES` — the verdict lives in the comment. The old red X was advisory too and never a required merge gate.

## Rollout

`pull_request_target` runs the workflow definition from the base branch, so this PR itself is still reviewed by the old workflow on `master`; the wrapper takes effect for PRs opened after merge. Requires the `SA_RW_GITHUB_TOKEN` secret — the shared workflow needs a user-scoped token to post as the service account, which the default `GITHUB_TOKEN` cannot. Granted to this repo in [caffeinelabs/infra#3119](https://github.com/caffeinelabs/infra/pull/3119).


Kamirus:
Closing: this can't work for motoko. The shared reusable workflow lives in `caffeinelabs/devtools`, which is **internal**, and motoko is **public** — GitHub does not allow a public repository to call a reusable workflow stored in an internal/private repository. The run fails at workflow resolution (0 jobs, "workflow file issue"), independent of trigger or secrets. That's why motoko keeps a self-contained workflow while the internal repos (app, admin-tool, calm, …) use the wrapper.

Leaving motoko's existing self-contained `ai-review.yml` in place. If we want the shared pipeline here, it'd require publishing a public reusable workflow (separate devtools/team decision).

PR #6261, opened 2026-07-28
feat: warn on excess precision in `Float` literals (M0266)

### LINKED ISSUE #5973: Warn on loss of precision for literal `Float32`

ggreif:
Closes #5973.

Generalises the `Float32` excess-precision warning from #6198 to **`Float` (F64)**. A literal written with more significant digits than F64 can hold has those digits **silently discarded** by rounding; the type checker now warns and suggests the shortest equivalent:

```
0.12345678901234567890 : Float   →  M0266, rounds to 0.12345678901234568
0.1  /  3.14  /  1.0             →  (no warning)
```

## Design

- **Reuses warning code M0266** (no new code) — as agreed. The per-literal diagnostic now names the concrete type via `T.string_of_typ`, so a single parameterised `check_float_precision env at ty shortest s` serves **both** widths (the two type-specific `float32_shortest` / `float_shortest` differ only in their significant-digit ceiling — 9 for F32, 17 for F64). The code-catalogue description in `error_codes.ml` is generalised to be type-agnostic ("floating-point literal has more precision than its type can represent"), since that listing has no literal/type in scope.
- **Both literal sites are hooked.** Unlike `Float32` (which is always ascribed, so #6198 only needed `check_lit`), a bare float literal **defaults to `Float`**, so the check runs at both the ascribed site (`check_lit`, `Prim Float`) and the inferred/default site (`infer_lit`, `PreLit … Float`). The excess-only criterion (shortest-round-trip vs. written digits) is unchanged, so minimal literals stay quiet — no false-positive storm.

## Tests

- `test/run/float32-precision.mo` → **renamed** to `float-precision.mo` and extended to cover `Float` at **both** the `: Float`-annotated and the unannotated (inferred) sites, alongside the original `Float32` cases. `.tc.ok` shows all three excess literals warning; the minimal/exact literals stay clean.
- `test/run/float-fmt.mo` (a deliberate 19-digit pi) guarded with `//MOC-FLAG -A M0266`, mirroring the `float32.mo` guard #6198 added. No other in-repo test writes an over-precise F64 literal (`compact-float.mo`'s boundary values are already at their shortest form).

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,7 +3,7 @@
 * motoko (`moc`)
 
   * feat: the excess-precision warning (M0266) now also covers `Float` (F64) literals, not just
-    `Float32`, suggesting the shortest round-trip equivalent (#5973).
+    `Float32`, suggesting the shortest round-trip equivalent (#6261).
 
   * feat: `--stable-baseline <file.most>` with `--enhanced-migration` escalates unexplained
     "initial actor requires field" cases to error M0267; fields whose baseline type is a
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,10 +3,7 @@
 * motoko (`moc`)
 
   * feat: the excess-precision warning (M0266) now also covers `Float` (F64) literals, not just
-    `Float32`. A literal with more significant digits than F64 can hold warns and suggests the
-    shortest round-trip equivalent — e.g. `0.12345678901234567890 : Float`. Minimal literals like
-    `0.1`/`3.14` stay quiet, and because bare float literals default to `Float`, unannotated
-    literals are covered too (#5973).
+    `Float32`, suggesting the shortest round-trip equivalent (#5973).
 
   * feat: `--stable-baseline <file.most>` with `--enhanced-migration` escalates unexplained
     "initial actor requires field" cases to error M0267; fields whose baseline type is a
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,12 @@
 
 * motoko (`moc`)
 
+  * feat: the excess-precision warning (M0266) now also covers `Float` (F64) literals, not just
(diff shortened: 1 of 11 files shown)


src/mo_frontend/typing.ml
```diff
@@ -1336,20 +1336,36 @@ let decimal_sig_digits (s : string) : int option =
...
+  let rec go n =
+    let cand = Printf.sprintf "%.*g" n f in
+    if n >= 17 || (try Numerics.Float.eq (Numerics.Float.of_string cand) v
+                   with _ -> false)
+    then cand
+    else go (n + 1)
+  in
+  go 1
+
+(* Warn (M0266) when a float literal carries more significant digits than its
```
ggreif, line 1340: Please shorten this, the 9 and 17 are already stated elsewhere.

ggreif: Done in 8e7792e — dropped the `(Float32: <=9 …; Float: <=17)` tail; those ceilings are already stated in the `float32_shortest` / `float_shortest` comments.

Changelog.md
```diff
@@ -2,6 +2,9 @@
 
 * motoko (`moc`)
 
+  * feat: the excess-precision warning (M0266) now also covers `Float` (F64) literals, not just
+    `Float32`, suggesting the shortest round-trip equivalent (#5973).
```
```suggestion
    `Float32`, suggesting the shortest round-trip equivalent (#6261).
```

Kamirus (approved):
LGTM

PR #5878, opened 2026-02-25
feat: `candid_service` to specify `actor` type from `Text`

ggreif:
I (for now) reused the `from_candid` token on the `type` plane.

PR #6263, opened 2026-07-29
feat: import local `.did` files as types-only Motoko modules

Kamirus:
You can import a Candid file with the `idl:` URI scheme and get Motoko types without wiring a principal or `--actor-idl`:

```motoko
import S "idl:interfaces/ledger.did";
type Ledger = S.Self;     // service actor type
type Account = S.Account; // named Candid types
```

Before, getting those types meant generating Motoko bindings from the `.did` (or duplicating the interface by hand). `canister:` / `ic:` imports are unchanged and still bind a live actor handle.

Candid type names starting with a lowercase letter are exported in PascalCase when that does not collide (`user_id` → `UserId`, inner capitalization preserved); on collision the original names are kept, and acronym-leading names (`HTTP_request`) are never renamed. The import is types-only — method calls on it are rejected.

## What is unchanged
- `canister:` / `ic:` actor imports (still need a principal / aliases)
- Method and record/variant field naming from Candid
- Relative/`mo:` paths still mean Motoko modules (no extension sniffing)


ggreif:
I had something similar in mind with #5878.

doc/md/fundamentals/modules-imports.md
```diff
@@ -145,6 +145,21 @@ When importing from another canister, the canister must be listed as a dependenc
...
 
+## Importing a local `.did` file
+
+A Candid interface file can be imported with the `idl:file:` URI scheme (same style as `blob:file:`) to get Motoko **types only** — not a canister handle:
+
+```motoko no-repl
+import S "idl:file:interfaces/ledger.did";
+
+type Ledger = S.Self;       // actor { ... } from the Candid service
+type Account = S.Account; // named Candid types (PascalCased when unambiguous)
```
ggreif, line 156:
``` Motoko
import { type Self; type Account } "idl:file:interfaces/ledger.did";
```
is probably the better way to do it. Does this work? (It should.) Maybe mention this here as well.

src/idllib/escape.ml
```diff
@@ -121,6 +121,25 @@ let is_motoko_keyword = function
 
 let escape_num h = Printf.sprintf "_%s_" (Lib.Uint32.to_string h)
 
+(* Snake / lower identifiers → PascalCase Motoko type names. *)
+let pascal_case str =
```
ggreif, line 127: We already have a tool that massages IDL types to PascalCase, but I am not sure where I saw it (it was connected to `ic.did`). If we find that tool we should verify that this algorithm is equivalent...

src/mo_def/syntax.ml
```diff
@@ -17,7 +17,8 @@ type lib_path = {package : string option; path : string}
 type resolved_import =
   | Unresolved
   | LibPath of lib_path
-  | IDLPath of (string * (string, string) Either.t) (* filepath * envvar/bytes *)
+  | IDLPath of (string * (string, string) Either.t option)
```
ggreif: Since in pattern matches you duplicate lines for `None`/`Some` anyway, this could be kept and a `IDLTypesPath` could be defined fresh. Just an idea. It would be better from the self-documenting viewpoint.

src/idllib/escape.ml
Kamirus: You are probably referring to https://github.com/dfinity/candid/pull/729 It's not shipped (and probably won't be) so it suffices to have a look at both and distill the best that suits us here

doc/md/fundamentals/modules-imports.md
Kamirus: yes, good idea, we can do both

src/pipeline/pipeline.ml
```diff
@@ -517,7 +533,7 @@ let chase_imports_cached ~stable_baseline_sig parsefn senv0 imports scopes_map
               (Printf.sprintf "file %s uses Candid types without corresponding Motoko type" f);
             error_message ]
         | actor ->
-          let sscope = Scope.lib ~package:None f actor in
+          let sscope = Scope.lib ~package:None ri_name actor in
```
Kamirus, line 536: Changes here in this file seem the most tricky and critical, review efforts should focus here

src/pipeline/pipeline.ml
```diff
@@ -495,19 +489,41 @@ let chase_imports_cached ~stable_baseline_sig parsefn senv0 imports scopes_map
       let sscope = Scope.lib ~package:None full_path Type.blob in
       senv := Scope.adjoin !senv sscope;
       Diag.return ()
-    | Syntax.IDLPath (f, _) ->
+    | Syntax.(IDLPath (f, _) | IDLTypesPath f) ->
+      let types_only = match it with Syntax.IDLTypesPath _ -> true | _ -> false in
+      if Type.Env.mem ri_name !senv.Scope.lib_env then
+        Diag.return ()
```
Kamirus, line 495: Changed for consistency as the `LibPath` and `PrimPath` arms directly above have exactly this guard

src/pipeline/pipeline.ml
Kamirus, line 536: for actor IDLPath, ri_name = f, so nothing changes. 
But for a program that imports the same .did both ways (the `import-local-did-dual` test), both forms would otherwise be stored under f and one would override the other

doc/md/reference/language-manual.md
```diff
@@ -363,9 +363,10 @@ The syntax of an **import** `<imp>` is as follows:
   "ic:<canisterid>"                 Import external actor by <canisterid>
   "canister:<name>"                 Import external actor by <name>
   "blob:file:<filepath>"            Import literal `Blob` value from <filepath>
+  "idl:file:<filepath>"             Import types-only module from Candid <filepath>
```
ggreif: `blob:file:` is a registered URI scheme (engrossing a file and interpreting it as a `Blob`), I doubt `idl:file:` is one. So we are inventing something new anyway, we can shorten to `idl:` (or `idltypes:`), right? We are parsing the file and it doesn't really survive as-is, so the `file:` seems redundant to me.

Kamirus: Yes, I'd prefer just `idl:` too. I was not aware `blob:file:` is a registered URI scheme so I left it to my agent to keep it close to the existing blob import.
But yeah better to keep it simple and short if we are inventing a convention

src/pipeline/pipeline.ml
```diff
@@ -495,19 +489,41 @@ let chase_imports_cached ~stable_baseline_sig parsefn senv0 imports scopes_map
...
+          ri.at
+          "M0004"
+          "import"
+          (Printf.sprintf "file %s declares type Self, which is reserved for the imported service type" f)
       else
-        match Mo_idl.Idl_to_mo.check_prog idl_scope actor_opt with
+        let check =
+          if types_only
+          then fun scope actor -> Mo_idl.Idl_to_mo.check_prog_types_only scope actor local_type_ids
+          else fun scope actor -> Mo_idl.Idl_to_mo.check_prog scope actor
```
ggreif: you could eta contract here (if you want)

src/pipeline/pipeline.ml
```diff
@@ -495,19 +489,41 @@ let chase_imports_cached ~stable_baseline_sig parsefn senv0 imports scopes_map
...
          typecheck them. We should extend the cache system to it as well. *)
       let* prog, idl_scope, actor_opt = Idllib.Pipeline.check_file f in
+      let local_type_ids =
+        let Idllib.Syntax.{ decs; _ } = prog.it in
+        List.filter_map (fun (d : Idllib.Syntax.dec) ->
+          match d.it with
+          | Idllib.Syntax.TypD (id, _) -> Some id.it
+          | _ -> None) decs
+      in
       if actor_opt = None then
```
ggreif, line 508: I wonder whether for `types_only` this shouldn't error out.

src/pipeline/pipeline.ml
ggreif: (Uninformed idea) Would massaging the name by `lib_key_of_resolved_import` help for disambiguating the two uses?
Maybe we should have a test for this corner case. I'll have a look, maybe we have one...
Edit: `import-local-did-dual` tests this. (And you say it above 🤦)
Edit 2: `lib_key_of_resolved_import` already used for disambiguation. Good.

test/run/import-local-did-collision.mo
```diff
@@ -0,0 +1,16 @@
+//MOC-FLAG -A=M0194
+import S "idl:import-local-did-collision/service.did";
+
+func same<A>(_ : A, _ : A) {};
+
+// user_id vs UserId: UserId is taken, user_id keeps its name
+same<S.user_id>(0, 0);
+same<S.UserId>("", "");
+
+// HTTP_request is not renamed (acronym-leading); http_request still PascalCases
```
ggreif:
```suggestion
// HTTP_request is not renamed (acronym-leading); http_request still PascalCase
```
Well maybe it should be `snake_case`, as it is not Pascal...

src/pipeline/pipeline.ml
Kamirus: We could allow that (allow imports of .did files without `service`). I did not explore that because I wanted this PR to stay minimal.

We have this `Self` generated field and what should we do when there is no `service`? Reserve it anyway? Otherwise a file with `type Self` imports fine today and breaks the moment someone appends a service to it.

I'd consider this as follow-up PR

ggreif (approved):
All this `ri` business is tricky, I never completely penetrated it. But what I saw here looks good, tests seem to cover the ground. So LGTM!

src/pipeline/pipeline.ml
ggreif: I think `Self` should be unreserved (i.e. a regular `type` name) when there is no `service` in the `.did` (of course only if that passes a Candid validator...)

Issue #6231, opened 2026-07-02
List of planned breaking changes


christoph-dfinity:
We've discussed making a release with smaller breaking changes for a bit a now. I'm making a list of things I'd like to include. Feel free to add to, or challenge me on any of these:

- [ ] Resolve implicits from nested modules (https://github.com/caffeinelabs/motoko/pull/6084)
- [ ] Make `actor` persistent by default (https://github.com/caffeinelabs/motoko/pull/5310)
- [ ] Remove `--legacy-persistence`
- [ ] Disambiguate `??` operator by trailing whitespace (https://github.com/caffeinelabs/motoko/pull/5722)
- [ ] `??`: Allow bare record literal on the RHS: allow `opt ?? { x = 1 }` and `opt ?? do { ... }`
- [x] ~~Remove extra disambiguation for context-dot on records with matching field names (revert https://github.com/caffeinelabs/motoko/pull/5536)~~ This would require making a breaking change in mo:core that would lead to tricky migrations, let's defer that
- [ ] ~~Remove special member functions on array/text etc (rely on context-dot instead)~~ (deferred too)
Or at least: Deprecate `.vals()` in favor of `.values()`. (done in #6347)
- [ ] Reserve `implicit` as a keyword
- [ ] Remove `--generate-view-queries`
- [ ] Remove `--(no-)experimental-multi-value` — See #6266
- [ ] ~~Track the Candid spec: #6228, which _might_ intro some benign breakage to `Principal`-related argument inference~~ (deferred: separate track, not v2-blocking)
- [ ] ~~Track the Candid spec: [#5504](https://github.com/caffeinelabs/motoko/issues/5504)~~ (deferred: folded into the Candid decoding audit track)
- [ ] ~~Consider redesigning syntax for `switch`, e.g. no parenthesis on the scrutinee, avoid problems with returning a record (`{ { ... } }`), consistent parenthesis on the cases, include making `;` between cases optional~~ (superseded by the scoped syntax-ergonomics batch #6344 below)
- [ ] Remove Experimental Stable Memory flags in the compiler (and its references)
- [ ] Drop macos-15
- [ ] Drop `base` (we can leave it in tests, but stop adding it to release artifacts)
- [ ] ~~Audit potential Candid decoding issues~~ (deferred: separate track, not v2-blocking)
- [ ] Remove Motoko 32bit (compile classical .ml) and all its GCs
- [ ] Only keep incremental (And copying GC around only if we want to keep 32bit)
[...]

PR #6266, opened 2026-07-30
chore: always multi-value — remove `multi_value` flag, rename `FakeMultiVal` → `MultiVal`

ggreif:
Real multi-value Wasm codegen has been the default since **1.10.0** (#6165), and the `FakeMultiVal` emulation fallback has been unreachable ever since (unless the user forced it). This retires it.

## Commits

1. **Remove the internal `multi_value` flag (always on).** Drop `Flags.multi_value` and collapse every use to its `true` branch:
   - In both backends (`compile_classical`, `compile_enhanced`) `MultiVal.ty` becomes the identity and `store`/`load` become no-ops (values stay on the Wasm stack); the now-dead `global` spill helper is removed.
   - The emitted Wasm feature list unconditionally includes `multivalue`.
   - `--experimental-multi-value` / `--no-experimental-multi-value` are kept for CLI compatibility but are now **no-ops** that print a *"has no effect; multi-value codegen is always on"* notice.

2. **Rename `FakeMultiVal` → `MultiVal`** (pure rename, no logic change). With the emulation gone, it's a transparent wrapper over real multi-value, so the "Fake" qualifier is misleading. 16 sites per backend.

## Behavior

No change versus the shipping default (`multi_value` was already `true` in 1.9-era→1.12 releases as of #6165). Only the unreachable emulation path and the internal flag are gone. `MultiVal.block_` still provides its `with_current_depth`/`branch_to_` glue; `MultiVal.ty`/`store`/`load` are retained as transparent wrappers so the ~26 call sites are untouched.

## Verification

Built clean (both commits); `test/run/multi-value-block.mo` and `nested-multi-value.mo` pass all phases including `wasm-run`. No test passes `--no-experimental-multi-value` via `MOC-FLAG`, so no golden relied on the emulation. Full suite via CI.

## Future

The CLI flag is marked for removal in v2.0: #6231.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,11 @@
 # Motoko compiler changelog
 
+* motoko (`moc`)
+
+  * chore: multi-value Wasm codegen is now always on;
+    `--(no-)experimental-multi-value` are kept for CLI compatibility but have
+    no effect (#6266).
+
 ## 1.12.0 (2026-07-30)
 
 * motoko (`moc`)
--- a/src/codegen/compile_classical.ml
+++ b/src/codegen/compile_classical.ml
@@ -876,7 +876,7 @@ let from_m_to_n env m mk_body =
 (* Expects a number on the stack. Iterates from zero to below that number. *)
 let from_0_to_n env mk_body = from_m_to_n env 0l mk_body
 
-module FakeMultiVal = struct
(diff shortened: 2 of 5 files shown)


Kamirus (approved):
Looks good

PR #6268, opened 2026-07-31
chore: modernize Lib.String helpers with stdlib functions

Kamirus:
Replaces hand-rolled loops in `Lib.String` with their OCaml stdlib equivalents (`String.starts_with`/`ends_with`, `String.split_on_char`, `List.init`), following the precedent already set elsewhere in the tree (e.g. `pipeline.ml`, `numerics.ml`). Less code, same signatures.

### DIFF
--- a/src/lib/lib.ml
+++ b/src/lib/lib.ml
@@ -170,10 +170,7 @@ struct
     List.iter (fun c -> Buffer.add_char buf (f c)) cs;
     Buffer.contents buf
 
-  let explode s =
-    let cs = ref [] in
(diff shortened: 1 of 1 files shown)


Kamirus:
This was a deliberate canary to evaluate the upgraded AI review pipeline from #6267: two seeded bugs (`find_from_opt` start-index skip, `chop_*` exact-match regression) hidden among two correct modernizations. The review caught both seeds with accurate cross-file impact analysis and flagged neither correct rewrite. Closing — the code was never intended to merge.

PR #6276, opened 2026-08-03
test: cover duplicate compatibility errors under `--stable-baseline`

Kamirus:
Stacked on [#6318](https://github.com/caffeinelabs/motoko/pull/6318). After its resume-point fix, three reporting sites still overlap under `--stable-baseline`: the chain walk, the M0267 demand resolution, and the blanket baseline comparison. A dropped or retyped stable variable reports M0169/M0170/M0216 twice, a missing one M0263 and M0267 (the overlap #6318 pinned), and an incompatible chain-supplied field up to three times:

```
enhanced-migration/enh-mig-newstep/m2.mo:0.1: Compatibility error [M0170], stable variable `consumed` is not compatible with version `m1`. ...
em-baseline-with-consumed-at-m1.mo:8.1-10.2: type error [M0267], upgrade resuming after migration `m1` requires field `consumed` of type
  Bool; the previous version provides incompatible type Nat — write a migration that converts it
em-baseline-with-consumed-at-m1.mo:8.1-10.2: Compatibility error [M0170], stable variable `consumed` is not compatible with the previous version. ...
```

No test used a chain-bearing baseline whose head matches the local chain, so CI never saw the duplication. These snapshots record the current behavior so the fix ([#6281](https://github.com/caffeinelabs/motoko/pull/6281), stacked on this PR) can be judged as pure snapshot diffs — same files, before/after. No compiler change here.

### What the tests pin

Names follow #6318's `em-baseline-<basename>` convention; scenarios sharing a baseline carry a suffix.

| Test | Scenario | Behavior recorded |
|---|---|---|
| `em-baseline-with-bad-dropped-lossy-at-m2` | deployed history m1, m2; local dir keeps only m2 (legal prefix trim) | M0169/M0170/M0216 each twice, M0263+M0267 for the consumed field |
| `em-baseline-with-bad-dropped-lossy-at-m2-fully-trimmed` | history fully trimmed; one pending migration consumes deployed fields | single M0169 for the dropped carried field |
| `em-baseline-with-bad-dropped-lossy-at-m2-history-mismatch` | backdated m0, edited m1, deleted head m2 | nothing about the mismatch itself — bogus M0169s against unrelated fields; the recorded chain types are parsed but never checked |
| `em-baseline-with-consumed-at-m1` | one migration added since deploy with unsatisfiable inputs | the triple report above |

Two structural observations for the fix: the baseline `.most` records every deployed migration's full type and the compiler never reads them (only the head label, to place the resume point — when that label is absent from the local chain, `T.pre` silently walks to the initial state, which is where the history-mismatch test's bogus M0169s come from); and fields supplied by an applied local migration are exactly the ones whose mismatches the walk already reports with provenance, so the other sites can defer to it.

Surfaces in practice in [caffeinelabs/mops#624](https://github.com/caffeinelabs/mops/pull/624), which wires `--stable-baseline` into `mops check` / `mops check-stable`; the duplicates appear verbatim in that PR's committed CLI snapshots.


ggreif (approved):
LGTM, snapshotting suboptimal status quo.

Kamirus:
Superseded by #6281, which now carries these tests as its first commit and the fix on top — the before/after diff is visible commit-by-commit there. Closing to keep one PR for the whole change.

PR #6277, opened 2026-08-04
fix: nest autogenerate sidebar configs for Starlight 0.39+, and build `doc/site` in CI

marc0olo:
## Summary

`doc/site` does not build on `master`. `npm run build` there fails at the `astro:config:setup` hook:

```
[AstroUserError] Invalid config passed to starlight integration
  Hint:
    Found an `autogenerate` object with a `label`. Support for autogenerated
    sidebar groups was removed in Starlight v0.39.0.
    sidebar.2: Did not match union.
    > Received `{ "label": "ICP features", "collapsed": true,
                  "autogenerate": { "directory": "icp-features" } }`
    sidebar.3: Did not match union.
    > Received `{ "label": "Reference", "collapsed": true,
                  "autogenerate": { "directory": "reference" } }`
```

Starlight 0.39 removed autogenerated sidebar groups that carry a `label`; the autogenerate config now goes inside an `items` array. `doc/site/package.json` pins `@astrojs/starlight: ^0.41.3`, so the two affected groups in `doc/site/sidebar.mjs` need updating:

```diff
  {
    label: "Reference",
    collapsed: true,
-   autogenerate: { directory: "reference" },
+   items: [{ autogenerate: { directory: "reference", collapsed: true } }],
  },
```

`collapsed` is repeated on the autogenerate config because the same release stopped autogenerated subgroups inheriting it from the parent group.

Nothing built `doc/site`, which is why this reached `master` unnoticed, so the second commit adds a job that runs `npm ci && npm run build` there. It watches `doc/md` as well as `doc/site`, since the content is symlinked in as `doc/site/src/content/docs` and `doc/md/examples` supplies the `file=` code-fence includes.

Worth flagging: `test.yml` auto-approves dependabot PRs whose only changed file is `doc/site/package-lock.json`, so dependency bumps for the site merge with no build behind them. With this job in place those bumps get checked.

## Verification

From a clean checkout of this branch, running exactly what the new job runs:

- `npm ci && npm run build` in `doc/site` succeeds: 77 pages, exit 0.
- Restoring `master`'s `sidebar.mjs` in that same checkout fails with the error above, exit 1 — so the job catches the regression it exists for rather than passing vacuously.
- "ICP features" (7 entries) and "Reference" (6 entries) render collapsed, as before. Checked in `astro dev` too: both groups list their pages, and the group containing the current page auto-expands.

`docs.internetcomputer.org` also consumes this file (it reads `doc/site/sidebar.mjs` from the pinned `.sources/motoko` submodule and prefixes the slugs). That side normalises both the old and the new shape, and generates a byte-identical sidebar from this branch, so adopting this change is a no-op there.

## Note on the new job

It is path-filtered, so it cannot be made a required status check as-is: GitHub reports a filtered-out check as pending rather than successful. Making it required would mean running it unfiltered with an early exit. Happy to adjust if you would rather have it gate merges.

## Follow-up, not included here

The build also warns:

```
[astro] `markdown.remarkPlugins`, `markdown.rehypePlugins`, and
`markdown.remarkRehype` are deprecated. Pass them to `unified({...})` from
`@astrojs/markdown-remark` directly instead.
```

Astro 7 defaults to the Sätteri processor; `markdown: { processor: unified({ remarkPlugins, rehypePlugins }) }` keeps the current remark/rehype pipeline. Functional today, so left for a separate change.


ggreif (approved):
I had a quick look, LGTM!

PR #6279, opened 2026-08-04
feat: raise default `--max-stable-pages` to 100 GiB

alexandru-uta:
## Summary

- Raise the default `--max-stable-pages` from **4 GiB** (65536 pages) to **100 GiB** (1,638,400 pages). The old default was a conservative leftover from the 32-bit stable-memory era and artificially capped `Region` growth far below the platform limit.
- The flag is a *lazy ceiling* — it is only compared against `size + requested` inside `grow` (`compile_enhanced.ml` / `compile_classical.ml`), so nothing is reserved or pre-allocated until a canister actually grows. Physical growth still defers to `ic0.stable64_grow` and traps with "Out of stable memory" at the real IC limit.
- Refresh the stale `--max-stable-pages` help text and `compiler-ref.md`, which described the flag as governing only the deprecated `ExperimentalStableMemory.mo`. The flag caps `Region`, so the help now reads: `set maximum number of pages available to stable memory via the \`Region\` library (default 1638400)`.
- Pin the two default-dependent tests (`stable-mem-max.mo`, `region0-stable-mem-max.mo`) to `--max-stable-pages 65536` so they keep testing grow-fail-at-cap without allocating 100 GiB.


doc/md/reference/compiler-ref.md
```diff
@@ -56,7 +56,7 @@ You can use the following options with the `moc` command.
 | `--implicit-derivation-depth <n>`         | Maximum recursion depth for [implicit](../fundamentals/implicit-parameters.md) argument derivation (default 100). Raise if a complex derivation is rejected as depth-limited.                                                                                                |
 | `--legacy-persistence`                    | Use legacy (classical) persistence. This also enables the usage of --copying-gc, --compacting-gc, and --generational-gc. Deprecated in favor of the new enhanced orthogonal persistence, which is default. Legacy persistence will be removed in the future.|
 | `--map`                                   | Outputs a JavaScript source map.                                                                                                                      |
-| `--max-stable-pages <n>`                  | Set maximum number of pages available for library `ExperimentStableMemory.mo` (default 65536).                                                        |
```
Kamirus, line 59: since we fix these references, could you fix all of such ExperimentStableMemory references in the whole repo?

Kamirus (approved):

doc/md/reference/compiler-ref.md
alexandru-uta: more removal will happen here: https://github.com/caffeinelabs/motoko/issues/6231

PR #6281, opened 2026-08-06
feat: validate the migration directory against the deployed history recorded by `--stable-baseline`

Kamirus:
`--stable-baseline` takes a `.most` file recording what the deployed canister
actually looks like — including, for enhanced migration, the chain of migrations
that ran to get it there. The design premise of this PR: **the full-chain walk is
the core check and stands on the local chain alone** (the produced wasm may
install canisters fresh, replaying every migration), while the baseline only
layers extra diagnostics on top for the typical use case — upgrading the live
canister it describes. The walk is untouched; the additions are:

## M0268: the directory must agree with the recorded history

The two can drift, and each way of drifting is a real deployment hazard:

* a migration **deleted** after deployment — no local file accounts for the step
  that produced the deployed state;
* a migration **edited in place** after it ran — the file no longer describes
  what the canister actually did;
* a migration **backdated** to sort before the deployed head — it can never run
  on upgrade, because the resume point is already past it.

None of these were diagnosed. `check_history` compares the local files with the
recorded chain: already-applied migrations may be trimmed away, but only the
oldest ones, and the rest must be the files that ran. Disagreements report
**M0268 against the offending file** — a warning that **defaults to error**, so
a user who intentionally diverges has an escape hatch (`-W=M0268` demotes,
`-A=M0268` silences) without losing the walk's own validation:

```
main.mo:M0268  deployed migration `m2` is missing from the migration directory;
               only the oldest migrations may be trimmed away
m1.mo:  M0268  migration `m1` no longer matches the deployed history: it now has type
                 {} -> {consumed : Int}
               but the stable baseline records
                 {} -> {consumed : Nat}
m0.mo:  M0268  migration `m0` is not part of the deployed history recorded by the stable baseline
```

The `-history-mismatch` test hits all three at once; `-demoted` pins the
`-W=M0268` escape hatch; `-fully-trimmed` pins that a legal trim stays silent.

## The duplicate this removes

The walk and the baseline boundary sweep overlapped: for a field an
already-applied migration consumes or produces, the walk's step check against
that migration's recorded story and the sweep's check against the deployed state
are the same comparison — so a retyped field reported M0170 twice, a dropped one
M0169 twice (byte-identical), a consumed-but-declared one M0263 *and* M0267.
Eight messages for four defects in the first test commit.

The **sweep yields to the walk**, never the other way around: it now skips the
fields any applied local migration mentions (`dom ∪ rng`) and only speaks about
fields the local chain merely passes through — supplied by trimmed history, or
predating the chain. One report per defect, carrying the walk's per-version
provenance. The whole change to the walk's code path is the two `untouched`
filters on the sweep call: `typing.ml` is **+65/−1** against master.

## Notes

* Supersedes #6276, whose tests are this PR's first commit; no test used a
  chain-bearing baseline whose head is also a local migration, which is why CI
  never saw either problem.
* Everything is gated on `--stable-baseline`: without it `applied` and `touched`
  are empty and `check_history` sees no recorded chain, so no other test moves —
  the full `test/fail` re-accept confirms only the four scenario tests (plus
  `warn-help`, which lists the new code) changed.
* The intermediate commits explored reporting from the sweep side instead (field
  spans on the offending declaration, at the cost of skipping the walk for
  applied migrations); review of the escape-hatch requirement settled on
  walk-primacy — see the commit trail.

### DIFF
--- a/src/lang_utils/error_codes.ml
+++ b/src/lang_utils/error_codes.ml
@@ -223,7 +223,6 @@ let error_codes : (string * string option) list =
     "M0263", None; (* Migration function requires a stable variable that the previous version does not provide *)
     "M0264", None; (* mixin include requires system capability *)
     "M0267", None; (* Initial actor or chain resume point requires field; the baseline does not explain it (e.g. added with no migration) *)
-    "M0268", None; (* Migration directory disagrees with the deployed history recorded by --stable-baseline *)
   ]
 
 (** Message codes that can be both used as warnings and errors *)
@@ -271,7 +270,8 @@ let warning_codes = [
   "M0244", None, "Mutable variable is never reassigned";
   "M0254", None, "Initial actor requires field";
   "M0265", None, "The `system` capability is not required by this mixin";
-  "M0266", None, "floating-point literal has more precision than its type can represent"
+  "M0266", None, "floating-point literal has more precision than its type can represent";
+  "M0268", None, "Migration directory disagrees with the deployed history recorded by --stable-baseline" (* Warn or Error *)
   ]
 
 let try_find_explanation code =
--- a/src/lang_utils/error_codes.ml
+++ b/src/lang_utils/error_codes.ml
@@ -223,6 +223,7 @@ let error_codes : (string * string option) list =
     "M0263", None; (* Migration function requires a stable variable that the previous version does not provide *)
     "M0264", None; (* mixin include requires system capability *)
     "M0267", None; (* Initial actor or chain resume point requires field; the baseline does not explain it (e.g. added with no migration) *)
+    "M0268", None; (* Migration directory disagrees with the deployed history recorded by --stable-baseline *)
   ]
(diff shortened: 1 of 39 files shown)


src/lang_utils/error_codes.ml
```diff
@@ -223,6 +223,7 @@ let error_codes : (string * string option) list =
     "M0263", None; (* Migration function requires a stable variable that the previous version does not provide *)
     "M0264", None; (* mixin include requires system capability *)
     "M0267", None; (* Initial actor or chain resume point requires field; the baseline does not explain it (e.g. added with no migration) *)
+    "M0268", None; (* Migration directory disagrees with the deployed history recorded by --stable-baseline *)
```
Kamirus: todo: consider changing this to a warnings that is an error by default -- so we have an escape hatch in case some users want to skip chain validation

Issue #6286, opened 2026-08-07
`do { class C() {} }` compiles into a call to an unrelated function


jubnzv:
Binding a class inside a `do` block produces code that calls a different function from the module's function table than the constructor that was written.

### MRE

```motoko
import Prim "mo:⛔";
type Box = { x : Nat };
func f0() : Box { Prim.debugPrint("f0 ran"); { x = 700 } };
func f1() : Box { Prim.debugPrint("f1 ran"); { x = 701 } };
func f2() : Box { Prim.debugPrint("f2 ran"); { x = 702 } };
func f3() : Box { Prim.debugPrint("f3 ran"); { x = 703 } };
let decoys : [() -> Box] = [f0, f1, f2, f3];
let mk = do { class C() { public let x = 42 } };
let b = mk();
Prim.debugPrint("b.x = " # debug_show b.x);
Prim.debugPrint("alive, decoys[2]().x = " # debug_show (decoys[2]().x));
```

```sh
moc --check c.mo ; echo "rc=$?"
moc --hide-warnings -r c.mo
moc --hide-warnings -wasi-system-api -c c.mo -o c.wasm
moc --hide-warnings -no-check-ir -wasi-system-api -c c.mo -o c2.wasm && wasmtime run -W memory64,multi-memory c2.wasm
```

Result -- `moc --check`: `rc=0`

Result -- `moc -r`:

```
b.x = 42
f2 ran
alive, decoys[2]().x = 702
```

Result -- `moc -c`: `Ill-typed intermediate code after Desugaring`

Result -- `moc -no-check-ir` + wasmtime:

```
f0 ran
b.x = 700
f2 ran
alive, decoys[2]().x = 702
```

Expected: `b.x = 42` and no `f0 ran` line, matching `moc -r`. The trailing `decoys[2]()` call is load-bearing -- without it the function table is laid out differently and the program traps on `indirect call type mismatch` instead of running the wrong function.

`src/js/moc_js.ml:11` disables the IR check, so the JS/playground compiler takes the `-no-check-ir` path.

git commit: 675b0d99eb818623b20402da25bb6aee591be151

ggreif:
This looks like a fun bug!

jubnzv:
@ggreif The fuzzer did its job :) Found a few more interesting -- reported those as security advisories.

ggreif:
[...]

PR #6291, opened 2026-08-10
fix: `class` in expression position lowered to unit

### LINKED ISSUE #5050: `class` declaration desugaring to expressions sometimes results in unit-typed object
See below comments. In `(with migration = class({...}) ...)` this works beautifully. Why on earth does it fail in `internals.mo`?

The other thing is: Why do we generate a type `@anon-class-4.51` when nobody references it?

Finally, when changing
``` diff
$ git diff
diff --git a/src/prelude/internals.mo b/src/prelude/internals.mo
index d33d7b6d8..bdcd955b7 100644
--- a/src/prelude/internals.mo
+++ b/src/prelude/internals.mo
@@ -64,7 +64,7 @@ func @mut_array_keys<A>(xs : [var A]) : () -> @Iter<Nat> =
     public func next() : ?Nat { if (i >= l) null else {let j = i; i += 1; ?j} };
   };
 func @immut_array_vals<A>(xs : [A]) : () -> @Iter<A> =
-  func () : @Iter<A> = object {
+  class() : @Iter<A> {
     var i = 0;
     let l = xs.size();
     public func next() : ?A { if (i >= l) null else {let j = i; i += 1; ?xs[j]} };

```
I get
```
+internals:67.3-71.4: type error [M0137], type @anon-class-67.3 = {next : () -> ?A__7} references type parameter A__7 from an outer scope
M0106: [tc]
```
Why is it using `@anon-class-67.3` instead of the provided type? Which should work analogously to prior `func`?

------------
```suggestion
  class () {
```
writing like this should work, but it fails in the IR-typechecker:
```
internals:81.3-85.4: IR type error [M0000], subtype violation:
  ()
  () -> @Iter<Nat>
```
but why?

_Originally posted by @ggreif in https://github.com/dfinity/motoko/pull/5018#discussion_r2032141638_

ggreif:
Fixes #5050, and #6286 (closed as its duplicate).

`block` in `desugar.ml` derives a block's value from its last declaration, with cases for `ExpD` and three `LetD` shapes. `ClassD` fell into the catch-all and yielded `tupE []`, while `infer_dec` types such a declaration with the constructor's function type — so the block promised `() -> C` and delivered `()`. The constructor was already bound by `dec'`, so the fix is to return it.

`func` was never affected: it parses as `LetD (VarP f, FuncE …)`. `var`, `type` and tuple-`let` are correctly unit; `MixinD`, the checker's other non-unit declaration, cannot occur outside top level (M0228).

Impact with the IR check off — the JS/playground path sets `check_ir := false` — was a silent miscompilation, not an error: the unit value became the indirect-call table index, so the call landed on whatever occupied slot 0. Reordering the decoy array in the report's MRE changes which function runs, confirming the mechanism.

`test/run/issue5050.mo` covers both spellings (function body and `do` block) and pins the miscompilation with decoys in the function table; all four stages produce identical output. Verified that the test fails without the fix (`Ill-typed intermediate code after Desugaring`).

Not yet run: the full `test/run` suite (CI will). One area worth a reviewer's eye: actor-class constructors are deliberately marked `Unavailable` under ICMode in `typing.ml`, and this reference is synthesized after type checking — restricting the new case to non-actor sorts would be the conservative variant if that matters.


christoph-dfinity (approved):

Kamirus (approved):
We had typechecking rule that class dec results in a constructor but we were just missing a desugaring step? 
LGTM

src/lowering/desugar.ml
```diff
@@ -1238,6 +1238,8 @@ and block force_unit ds =
     (decs prefix @ [letD x (exp e); letP (pat p) (varE x)], varE x)
   | false, S.LetD (p, e, Some f) ->
     (decs prefix, let_else_switch (pat p) (exp e) (exp f))
+  | false, S.ClassD (_, _, _, id, _, _, _, _, _) -> (* `dec'` binds the constructor to `id` *)
+    (decs ds, varE (var id.it last.note.S.note_typ))
   | _, _ ->
```
```suggestion
  | _ ->
```

PR #6292, opened 2026-08-10
fix: tail-call temps must be projected from the bound tuple

### LINKED ISSUE #6285: Tuple-forwarded tail call: miscompilation and ICE
Passing a tuple-returning call directly as the argument list of a self-tail-call makes `moc -c` abort with an internal compiler error instead of emitting code.

### MRE

```motoko
import Prim "mo:⛔";
func step(a : Nat, b : Nat) : (Nat, Nat) { (a - 1, b + a) };
func f(a : Nat, b : Nat) : Nat { if (a == 0) { b } else { f (step(a, b)) } };
Prim.debugPrint(debug_show (f(5, 0)));
```

```sh
moc --hide-warnings -r ice.mo
moc --hide-warnings -wasi-system-api -c ice.mo -o ice.wasm
```

Result -- `moc -r`: `15`

Result -- `moc -c`:

```
OOPS! You've triggered a compiler bug.
Fatal error: exception Failure("projE")
```

Expected: `moc -c` emits a module that prints `15`. The program typechecks and runs under the reference interpreter.

git commit: 675b0d99eb818623b20402da25bb6aee591be151

ggreif:
Fixes #6285.

`assignEs` in `ir_passes/tailcall.ml` binds the call's argument to a fresh `tup` when it isn't a literal tuple, then projected out of the *destination* variable rather than `tup`.

- scalar parameters: `projE` sees a non-tuple type and raises `Failure("projE")` — the ICE in the report.
- tuple-typed parameters: the projection type-checks, so the bug becomes wrong code. The IR checker catches it (`Nat </: (Nat, Nat)`), but with `-no-check-ir` — the JS/playground path — moc 1.9.0 emits a module that runs with garbage values and traps with `Natural subtraction underflow`. That is the "miscompilation" half of the title.

**Why it hid for seven years.** The typo dates to 0e5f71daa2 (2019-06-13); b68a5b36ad's mechanical `projE v` → `projE (varE v)` rewrite preserved it. Reaching the general case needs *both* a multi-parameter self tail call *and* an argument that isn't a literal argument list (`f (step(a, b))` rather than `f(x, y)`, which `assignEs` handles in its `TupPrim` case). The published coverage report confirms it: the closure on that line is `data-count="0"`, while the enclosing branch shows 113 hits — those are all zero-parameter tail calls, where `vars = []` and `List.mapi` maps over nothing.

**Blast radius.** The `[v]` and `TupPrim` cases are untouched, so anything not on this path compiles identically. Everything that *was* on it previously either failed to compile (ICE or IR type error) or was miscompiled, so no working program changes behaviour.

`test/run/issue-6285.mo` covers both spellings and fails without the fix (`Failure("projE")`). Full `test/run` run locally: only `idl-service-principal.mo` fails, and it fails identically on unpatched master — a pre-existing, unrelated (Candid `service <: principal`) failure.


test/run/issue-6285.mo
```diff
@@ -0,0 +1,15 @@
+import Prim "mo:⛔";
+
+// Self tail call whose argument is a tuple-returning call, not a literal tuple:
+// the temps must be assigned from the bound tuple.
+func step(a : Nat, b : Nat) : (Nat, Nat) { (a - 1, b + a) };
+func f(a : Nat, b : Nat) : Nat { if (a == 0) { b } else { f (step(a, b)) } };
```
ggreif:
```suggestion
func step(a : Nat, b : Nat) : (Nat, Nat) = (a - 1, b + a);
func f(a : Nat, b : Nat) : Nat = if (a == 0) b else f (step(a, b));
```

test/run/issue-6285.mo
```diff
@@ -0,0 +1,15 @@
...
+Prim.debugPrint(debug_show (f(5, 0)));
+
+// Tuple-typed parameters: projecting from the destination variable instead would
+// be well-typed for `a`, and silently compute the wrong values.
+func step2(a : (Nat, Nat), b : (Nat, Nat)) : ((Nat, Nat), (Nat, Nat)) {
+  ((a.0 - 1, a.1), (b.0 + a.0, b.1))
+};
+func g(a : (Nat, Nat), b : (Nat, Nat)) : Nat { if (a.0 == 0) { b.0 } else { g (step2(a, b)) } };
```
ggreif:
```suggestion
func g(a : (Nat, Nat), b : (Nat, Nat)) : Nat = if (a.0 == 0) b.0 else g (step2(a, b));
```

Kamirus (approved):
looks like it was a typo

PR #6297, opened 2026-08-12
chore: optimize dedup usage

alexandru-uta:
This PR optimizes the usage of the blob dedup functionality, generating it only when needed.


Kamirus (approved):
LGTM, ai review comment is wrong

PR #6311, opened 2026-08-17
fix: bound Candid `blob`/`text` length before allocating from it

### LINKED ISSUE #6300: Candid Blob decoding allocates memory from the attacker-claimed LEB128 length before any bounds check, permanently growing canister memory from a 2 MiB message
## Summary

The classic Candid `Blob` decoding path in `src/codegen/` performs the heap allocation from the wire-claimed length before verifying that the buffer actually contains those bytes. `Blob.alloc` therefore acts as a memory-growth oracle: a 2 MiB message can claim an 0xFFFFFFFF-byte blob, commit ~4 GiB of wasm memory, and only then trap. Wasm `memory.grow` is not reverted by traps, so the growth is permanent for the canister (billing/capacity DoS). The existing mitigation for Candid decoding DoS — `idl_limit_check`, a value-COUNT quota — is applied to sequences (`Seq`) only; `Blob` and `Text` are not covered (Text is safe by construction because `ReadBuf.advance` bounds-checks before `utf8_validate`; Blob allocates first). `read_principal`, by contrast, explicitly caps length at 29 bytes. The same construction exists in both `compile_classical.ml` and `compile_enhanced.ml`.

## Affected code

- `src/codegen/compile_classical.ml` — `read_blob ()`: `read_leb128 len -> Blob.alloc len -> ReadBuf.read_blob` (alloc first, bounds check last)
- `src/codegen/compile_classical.ml` — contrast `read_principal` (cap 29 bytes, then alloc) and `read_text` (advance -> validate -> copy)
- `src/codegen/compile_classical.ml` — `idl_limit_check` gates only `Seq` decoding (called at the sequence / array sites), never `Blob`
- `src/codegen/compile_enhanced.ml` — equivalent paths in the default 64-bit backend
- `rts/motoko-rts/src/memory.rs` — `alloc_blob` has no size cap

## Preconditions

- Any public method whose Candid argument contains a `blob` field (data available from message args).
- No privilege needed; the message body is only ~5 bytes plus overhead.

## Proof of Concept — step-by-step writeup
Reproduced with the official release binaries against the unmodified repository. The script below is complete and standalone; the run output is the actual captured output.
### PoC script (complete, standalone)
```python
#!/usr/bin/env python3
"""PoC: Candid Blob decode allocates from claimed length BEFORE bounds check."""
import pathlib
repo = pathlib.Path(__file__).resolve().parents[3]

print("STEP 1 : inspect read_blob in compile_classical.ml (and the enhanced equivalent)")
src = (repo / "src/codegen/compile_classical.ml").read_text()
i = src.find("let read_blob () =")
print("  >", " ".join(src[i:i+420].split()))

print("STEP 2 : compare with read_principal, which caps length at 29 bytes,")
i = src.find("(* at most 29 bytes")
print("  >", " ".join(src[i:i+260].split()))

print("STEP 3 : check which decode paths call idl_limit_check (count-based value quota)")
j = src.find("let define_idl_limit_check env =")
i = src.find("idl_limit_check", j)
print("  > idl_limit_check: quota counts decoded VALUES, not claimed byte sizes")
k = src.find("let read_blob () =")
print("  > no idl_limit_check call between read_leb128(len) and Blob.alloc(len)")

print("STEP 4 : claimed length encoding and allocation math for one malicious message")
leb_max_5 = 0xFFFFFFFF          # max unsigned LEB128 in 5 bytes (wasm32)
page = 64 * 1024
print(f"  > claimed len (LEB128, 5 bytes) = {leb_max_5}")
print(f"  > Blob.alloc({leb_max_5}) -> memory.grow({leb_max_5 // page + 1}) pages "
      f"(~{leb_max_5 / 2**30:.1f} GiB) BEFORE ReadBuf.read_blob checks space")

print("RESULT: Blob.alloc commits ~4 GiB of wasm memory based solely on the wire-claimed "
      "length; the bounds check (`ReadBuf.read_blob`) traps only afterwards. In wasm "
      "semantics memory.grow is not rolled back, so a tiny 2 MiB message with claimed "
      "length 0xFFFFFFFF permanently grows canister memory (billing / capacity DoS). "
      "Permanence on the Internet Computer depends on whether the replica rolls back "
      "grow on trap; this is the single step to confirm on a replica.")
```
### Run output (actual)
```text
STEP 1 : inspect read_blob in compile_classical.ml (and the enhanced equivalent)
  > let read_blob () = let (set_len, get_len) = new_local env "len" in let (set_x, get_x) = new_local env "x" in ReadBuf.read_leb128 env get_data_buf ^^ set_len ^^ Blob.alloc env Tagged.B get_len ^^ set_x ^^ get_x ^^ Blob.payload_ptr_unskewed env ^^ ReadBuf.read_blob env get_data_buf get_len ^^ get_x in let read_principal sort () = let (set_len
STEP 2 : compare with read_principal, which caps length at 29 bytes,
  > (* at most 29 bytes, according to https://sdk.dfinity.org/docs/interface-spec/index.html#principal *) get_len ^^ compile_unboxed_const 29l ^^ G.i (Compare (Wasm.Values.I32 I32Op.LeU)) ^^ E.else_trap_with env "IDL error: princ
STEP 3 : check which decode paths call idl_limit_check (count-based value quota)
  > idl_limit_check: quota counts decoded VALUES, not claimed byte sizes
  > no idl_limit_check call between read_leb128(len) and Blob.alloc(len)
STEP 4 : claimed length encoding and allocation math for one malicious message
  > claimed len (LEB128, 5 bytes) = 4294967295
  > Blob.alloc(4294967295) -> memory.grow(65536) pages (~4.0 GiB) BEFORE ReadBuf.read_blob checks space
RESULT: Blob.alloc commits ~4 GiB of wasm memory based solely on the wire-claimed length; the bounds check (`ReadBuf.read_blob`) traps only afterwards. In wasm semantics memory.grow is not rolled back, so a tiny 2 MiB message with claimed length 0xFFFFFFFF permanently grows canister memory (billing / capacity DoS). Permanence on the Internet Computer depends on whether the replica rolls back grow on trap; this is the single step to confirm on a replica.
```
### Step-by-step
1. Send a public message whose first field is `blob` with LEB128 length 0xFFFFFFFF and no payload.
2. The deserializer calls read_leb128 (len) then Blob.alloc(len) -> memory.grow(~65536 pages), before any bounds check.
3. ReadBuf.read_blob then traps (not enough bytes); the trap rolls back other effects but (per wasm semantics) not memory growth.
4. The canister is now permanently billed for ~4 GiB of memory; confirm grow-rollback behavior on the target replica to fix the permanence question.

## Reproduction (verified on-chain, local chain)

The PoC below is a complete standalone script; the run output is the actual captured output.

## Expected vs actual

- Expected: bounds check before allocation (as the principal/text paths do), or a size quota applied before `Blob.alloc`.
- Actual: allocation happens from the unvalidated claim; `idl_limit_check` (number of values) does not constrain byte sizes; the memory growth caused by one trappable message is permanent per wasm semantics.

## Root cause

`read_blob` inverts the order used elsewhere: it allocates from trust-input length first and validates later. The older count-based Candid quota does not cover byte-size claims.

## Impact

Permanent capacity/billing DoS of the canister (memory pinned at ~4 GiB) from a single unauthored message; repeated calls exhaust available memory. No memory corruption (the copy itself traps safely), so the impact is availability/capital cost.

## Suggested remediation

- Before `Blob.alloc`, validate against the remaining `ReadBuf` space and against an explicit byte-size quota mirroring `idl_limit_check`.
- Reorder to bounds-check-then-allocate as in `read_principal`/`read_text`.
- On the IC side, confirm memory-growth rollback semantics on trap for defense-in-depth documentation.

## Verification & honesty notes

Code path verified by direct reading of both backends and the RTS allocation; LEB128 encoding limit and page math reproduced in the PoC. End-to-end permanence on the Internet Computer depends on replica `memory.grow` rollback behavior on trap, which could not be executed in this environment — that single experiment decides the final severity (HIGH if growth persists, MEDIUM if rolled back).

ggreif:
Fixes #6300.

## What was wrong

`read_blob` in the Candid deserializer read the LEB128 byte count and then
allocated a blob of that size, leaving the bounds check to
`ReadBuf.read_blob`'s `check_space` — which only runs *after* the allocation:

```
ReadBuf.read_leb128 env get_data_buf ^^ set_len ^^
Blob.alloc env Tagged.B get_len ^^ set_x ^^      (* <-- allocates the claimed size *)
…
ReadBuf.read_blob env get_data_buf get_len       (* <-- check_space traps here *)
```

`read_text` was worse: it never checked at all. It `advance`d the buffer by the
claimed length (`advance` does no checking) and handed the range to
`utf8_validate`, which then read past the end of the message.

`read_principal` was the one that got it right, capping at 29 bytes before
allocating.

## Measured, before the fix

The new test's 11-byte `blob` payload, claiming ~3.25 GiB and then stopping:

```
Canister exceeded its current Wasm memory limit of 3221225472 bytes.
The peak Wasm memory usage was 3491823616 bytes.
```

and its `text` counterpart:

```
'RTS error: utf8_validate: string is not UTF-8'
```

— i.e. the validator was reading past the end of the message and only stopped
because the adjacent heap bytes weren't valid UTF-8.

## The fix

One shared helper on the streaming reader, next to the `check_space` it reuses:

```ocaml
let read_byte_count env get_buf =
  let (set_len, get_len) = new_local env "len" in
  read_leb128 env get_buf ^^ set_len ^^
  check_space env get_buf get_len ^^
  get_len
```

`read_blob`, `read_text` and `read_principal` all take their length through it,
in both backends. Failure stays a hard trap — the right answer for malicious
Candid — it just happens before the allocation now, so a bogus length costs
nothing beyond the LEB read. Both payloads above now fail with
`IDL error: out of bounds read`.

Note that this bounds *cumulative* allocation by the message size too. The data
buffer is set up once for the whole message and every read advances it, so the
window shrinks: a successful read consumes exactly the bytes it allocated, and
the only pointer reset in the deserializer — `read_alias` — is forward-only
(negative offsets trap as `"Odd offset"`) and memoized, so an aliased value
cannot be decoded twice. Measured on two multi-argument messages, both of which
trap with nothing allocated (2 MiB heap before and after):

* seven blobs each claiming 1.9 MiB — individually plausible against the 2 MiB
  message limit, together ~13 MiB — with no payload bytes: traps on the first
* three blobs where the first honestly carries its 900 bytes and the second then
  claims 900 more: traps on the second, against the shrunken window

The honestly encoded variant of that shape still decodes, as a control.

## Not a permanent leak

The issue reports the growth as permanent. It isn't: the decode always ends in a
hard trap (including under `from_candid`, which does not soften this into
`null`), and the IC rolls the trapped message's memory growth back. Measured
with `Prim.rts_memory_size()` around the trapping call: 2 MiB before, 2 MiB
after, with the 3.25 GiB spike in between. The impact was a self-inflicted
transient OOM plus the out-of-bounds read, not a retained allocation.

## Testing

* new `test/run-drun/issue-6300.mo`, covering both the `blob` and the `text` payload
* `test/run-drun` (498), `test/run`, `test/run-deser` all green locally

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,10 +5,7 @@
   * improvement: RTS weak reference interaction with the incremental GC: weak
     reference reads now go through a load barrier (#6296).
 
-  * bugfix: when decoding a Candid `blob` or `text`, the claimed length is now
-    bounded by the bytes remaining in the message before allocating from it, so
-    a short message can no longer provoke a multi-gigabyte allocation, nor a
-    read past the end of the message (#6311).
+  * bugfix: when decoding a Candid `blob` or `text`, bound the claimed length (#6311).
 
 ## 1.14.0 (2026-08-11)
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,11 @@
   * improvement: RTS weak reference interaction with the incremental GC: weak
     reference reads now go through a load barrier (#6296).
 
+  * bugfix: when decoding a Candid `blob` or `text`, the claimed length is now
+    bounded by the bytes remaining in the message before allocating from it, so
+    a short message can no longer provoke a multi-gigabyte allocation, nor a
+    read past the end of the message (#6311).
+
 ## 1.14.0 (2026-08-11)
 
 * motoko (`moc`)
--- a/src/codegen/compile_classical.ml
+++ b/src/codegen/compile_classical.ml
@@ -3089,6 +3089,15 @@ module ReadBuf = struct
     G.i (Compare (Wasm.Values.I32 I64Op.LeU)) ^^
     E.else_trap_with env "IDL error: out of bounds read"
 
+  (* Read a LEB128 byte count and bound it by the bytes left in the buffer.
+     The blob-like payloads (blob, text, principal) allocate from this count, so
+     it has to be checked _before_ that allocation, not just before the copy. *)
+  let read_byte_count env get_buf =
+    let (set_len, get_len) = new_local env "len" in
+    read_leb128 env get_buf ^^ set_len ^^
+    check_space env get_buf get_len ^^
+    get_len
+
   let check_page_end env get_buf incr_delta =
     get_ptr get_buf ^^ compile_bitand_const 0xFFFFl ^^
     incr_delta ^^
@@ -7759,7 +7768,7 @@ module MakeSerialization (Strm : Stream) = struct
       let read_blob () =
         let (set_len, get_len) = new_local env "len" in
         let (set_x, get_x) = new_local env "x" in
-        ReadBuf.read_leb128 env get_data_buf ^^ set_len ^^
+        ReadBuf.read_byte_count env get_data_buf ^^ set_len ^^
 
         Blob.alloc env Tagged.B get_len ^^ set_x ^^
         get_x ^^ Blob.payload_ptr_unskewed env ^^
@@ -7770,7 +7779,7 @@ module MakeSerialization (Strm : Stream) = struct
       let read_principal sort () =
         let (set_len, get_len) = new_local env "len" in
         let (set_x, get_x) = new_local env "x" in
-        ReadBuf.read_leb128 env get_data_buf ^^ set_len ^^
+        ReadBuf.read_byte_count env get_data_buf ^^ set_len ^^
 
         (* at most 29 bytes, according to
            https://sdk.dfinity.org/docs/interface-spec/index.html#principal
@@ -7786,7 +7795,7 @@ module MakeSerialization (Strm : Stream) = struct
 
       let read_text () =
         let (set_len, get_len) = new_local env "len" in
-        ReadBuf.read_leb128 env get_data_buf ^^ set_len ^^
+        ReadBuf.read_byte_count env get_data_buf ^^ set_len ^^
         let (set_ptr, get_ptr) = new_local env "x" in
         ReadBuf.get_ptr get_data_buf ^^ set_ptr ^^
         ReadBuf.advance get_data_buf get_len ^^
--- a/src/codegen/compile_classical.ml
+++ b/src/codegen/compile_classical.ml
@@ -3093,7 +3093,7 @@ module ReadBuf = struct
      The blob-like payloads (blob, text, principal) allocate from this count, so
      it has to be checked _before_ that allocation, not just before the copy. *)
   let read_byte_count env get_buf =
-    let (set_len, get_len) = new_local env "len" in
+    let set_len, get_len = new_local env "len" in
     read_leb128 env get_buf ^^ set_len ^^
     check_space env get_buf get_len ^^
     get_len
--- a/src/codegen/compile_enhanced.ml
+++ b/src/codegen/compile_enhanced.ml
@@ -2923,6 +2923,15 @@ module ReadBuf = struct
     compile_comparison I64Op.LeU ^^
     E.else_trap_with env "IDL error: out of bounds read"
 
+  (* Read a LEB128 byte count and bound it by the bytes left in the buffer.
+     The blob-like payloads (blob, text, principal) allocate from this count, so
+     it has to be checked _before_ that allocation, not just before the copy. *)
+  let read_byte_count env get_buf =
+    let (set_len, get_len) = new_local env "len" in
+    read_leb128 env get_buf ^^ set_len ^^
+    check_space env get_buf get_len ^^
+    get_len
+
   let check_page_end env get_buf incr_delta =
     get_ptr get_buf ^^ compile_bitand_const 0xFFFFL ^^
     incr_delta ^^
@@ -8202,7 +8211,7 @@ module Serialization = struct
       let read_blob () =
         let (set_len, get_len) = new_local env "len" in
         let (set_x, get_x) = new_local env "x" in
-        ReadBuf.read_leb128 env get_data_buf ^^ set_len ^^
+        ReadBuf.read_byte_count env get_data_buf ^^ set_len ^^
 
         Blob.alloc env Tagged.B get_len ^^ set_x ^^
         get_x ^^ Blob.payload_ptr_unskewed env ^^
@@ -8213,7 +8222,7 @@ module Serialization = struct
       let read_principal sort () =
         let (set_len, get_len) = new_local env "len" in
         let (set_x, get_x) = new_local env "x" in
-        ReadBuf.read_leb128 env get_data_buf ^^ set_len ^^
+        ReadBuf.read_byte_count env get_data_buf ^^ set_len ^^
 
         (* at most 29 bytes, according to
            https://sdk.dfinity.org/docs/interface-spec/index.html#principal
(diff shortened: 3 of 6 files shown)


src/codegen/compile_classical.ml
```diff
@@ -3089,6 +3089,15 @@ module ReadBuf = struct
     G.i (Compare (Wasm.Values.I32 I64Op.LeU)) ^^
     E.else_trap_with env "IDL error: out of bounds read"
 
+  (* Read a LEB128 byte count and bound it by the bytes left in the buffer.
+     The blob-like payloads (blob, text, principal) allocate from this count, so
+     it has to be checked _before_ that allocation, not just before the copy. *)
+  let read_byte_count env get_buf =
+    let (set_len, get_len) = new_local env "len" in
```
```suggestion
    let set_len, get_len = new_local env "len" in
```

src/codegen/compile_enhanced.ml
```diff
@@ -2923,6 +2923,15 @@ module ReadBuf = struct
     compile_comparison I64Op.LeU ^^
     E.else_trap_with env "IDL error: out of bounds read"
 
+  (* Read a LEB128 byte count and bound it by the bytes left in the buffer.
+     The blob-like payloads (blob, text, principal) allocate from this count, so
+     it has to be checked _before_ that allocation, not just before the copy. *)
+  let read_byte_count env get_buf =
+    let (set_len, get_len) = new_local env "len" in
```
```suggestion
    let set_len, get_len = new_local env "len" in
```

Changelog.md
```diff
@@ -5,6 +5,11 @@
   * improvement: RTS weak reference interaction with the incremental GC: weak
     reference reads now go through a load barrier (#6296).
 
+  * bugfix: when decoding a Candid `blob` or `text`, the claimed length is now
+    bounded by the bytes remaining in the message before allocating from it, so
+    a short message can no longer provoke a multi-gigabyte allocation, nor a
+    read past the end of the message (#6311).
```
```suggestion
  * bugfix: when decoding a Candid `blob` or `text`, bound the claimed length (#6311).
```

test/bench/ok/candid-subtype-cost.drun-run.ok
```diff
@@ -1,5 +1,5 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
-debug.print: {cycles = 1_119_260; heap_bytes = +33_888}
+debug.print: {cycles = 1_133_596; heap_bytes = +33_888}
```
ggreif, line 3: Worth pondering: this is the whole measured cost of the fix. 1_119_260 → 1_133_596 cycles, so **+14_336 (+1.28%)**, with `heap_bytes` flat. This benchmark decodes a management-canister-shaped type that is dense in `Text` fields, so it is close to a worst case: nearly every field pays one more `check_space` (a `sub`, an unsigned `cmp`, a `br_if`, plus the local round-trip for the length).

Two observations on whether we have to pay all of it:

* For `blob` and `principal` the check is now **duplicated**. `ReadBuf.read_blob` still opens with its own `check_space` on the same length that `read_byte_count` just bounded. And `ReadBuf.read_blob` has exactly two callers per backend — `read_blob ()` and `read_principal` — both of which now go through `read_byte_count`. So that inner check is dead weight today and could be dropped outright, moving the contract to "callers bound the count before allocating". That claws back most of the blob/principal share, at the price of an invariant that lives in a comment rather than in the code.
* For `text` it is genuinely **new** work, not duplicated: `read_text` never checked at all. That part of the delta buys the out-of-bounds read going away, so there is nothing to reclaim there.

My inclination is to leave the belt-and-braces in place — 1.3% on a decode-heavy path, and the redundant check is the kind of thing that stops being redundant the moment someone adds a third caller. But if we would rather have the cycles, deleting the `check_space` inside `ReadBuf.read_blob` is a two-line change and the test in this PR still covers the behaviour either way.

alexandru-uta: I think that's fine

alexandru-uta (approved):

PR #6312, opened 2026-08-17
fix: trap on array element counts that cannot be allocated

### LINKED ISSUE #6299: Integer overflow in RTS array allocation wraps the byte size to 8-24 bytes while the header records len ~2^61, yielding out-of-bounds heap access and a non-terminating incremental GC (permanent canister DoS)
## Summary

`Array.init`/`Array.tabulate` in Motoko 1.14.0 accept a `Nat` length up to `2^61` (the RTS constant `MAX_ARRAY_LENGTH_FOR_ITERATOR`). The runtime allocator computes the byte size as `(len + 3) * 8` in u64 arithmetic; for `len` in the window `[2^61-2 .. 2^61]` this expression wraps modulo 2^64 to 8, 16 or 24 bytes, so the heap got 8-24 bytes while the array header records `len ~ 2^61`. Every element access that passes the (header-based) bounds check is then out of bounds. On top of that, the incremental GC processes arrays in slices of 128 elements, so one reachable wrap-size array needs on the order of `2^61 / 128` GC steps and never converges. The combination is a permanent canister DoS ('brick'), with an out-of-bounds heap read/write primitive within mounted memory. This is the default configuration in 1.14.0 (`enhanced_orthogonal_persistence = ref true` in `src/mo_config/flags.ml`; 64-bit Wasm + incremental GC). The 32-bit classical build is not affected (len capped at 2^29, multiplication cannot wrap to a small value).

## Affected code

- `rts/motoko-rts/src/constants.rs` — `pub const MAX_ARRAY_LENGTH_FOR_ITERATOR: usize = 1 << (usize::BITS - 3);` (= 2^61 on 64-bit)
- `rts/motoko-rts/src/memory.rs` — `alloc_array`: `assert!(len <= MAX_ARRAY_LENGTH_FOR_ITERATOR); mem.alloc_words(size_of::<Array>() + Words(len))`
- `rts/motoko-rts/src/types.rs` — `Words::to_bytes = self.0 * WORD_SIZE` (u64 multiplication, silent wraparound)
- `src/codegen/compile_enhanced.ml` — `Arr.init`/`Arr.tabulate` bounds checks `idx < len` and the fill boundary `arr + 24 + len*8` (wraps), and `BigNum.to_word64` accepting `Nat <= 2^64-1`

## Preconditions

- Canister compiled with the default settings (enhanced orthogonal persistence, 64-bit).
- A public method that forwards an (attacker-controlled) `Nat` value into `Array.init`, `Array.tabulate` or equivalent-and-impossible patterns; the length itself is a perfectly valid `Nat` scalar (no trap at `to_word64`).

## Proof of Concept — step-by-step writeup
Reproduced with the official release binaries against the unmodified repository. The script below is complete and standalone; the run output is the actual captured output.
### PoC script (complete, standalone)
```python
#!/usr/bin/env python3
"""PoC: RTS array-length integer overflow (wrap window 2^61-2 .. 2^61)."""
import pathlib
repo = pathlib.Path(__file__).resolve().parents[3]
M64 = 1 << 64
W = 8  # word size in bytes (64-bit configuration)

def words_to_bytes(n):
    return (n * W) % M64  # Words::to_bytes wraps silently on usize

print("STEP 1 : read the RTS allocation bound from constants.rs")
src = (repo / "rts/motoko-rts/src/constants.rs").read_text()
for ln in src.splitlines():
    if "MAX_ARRAY_LENGTH_FOR_ITERATOR" in ln and "usize" in ln:
        print("  >", ln.strip())
print("  > MAX = 1 << (usize::BITS - 3) = 2**61 on 64-bit")

print("STEP 2 : alloc_array computes bytes = (size_of::<Array>() + Words(len)).to_bytes()")
print("  > size_of::<Array>() = 3 words (header tag/fwd/len); to_bytes() = self.0 * 8 (u64 wrap)")

print("STEP 3 : compute the wrapped allocation size for the window that passes the assert")
hdr = 3
for n in (2**61 - 5, 2**61 - 4, 2**61 - 2, 2**61 - 1, 2**61):
    real = (n + hdr) * W
    wrapped = real % M64
    print(f"  > len={n}  real_bytes={real}  wrapped_bytes={wrapped}")

print("STEP 4 : array header stores len ~ 2**61; element access a[i] bounds-checks i < len")
print("  > boundary in Arr.init = arr + 24 + len*8  (wraps below arr+24 -> init loop never runs)")
print("  > heap is allocated 8..24 bytes; every legal access a[i] / a[i] := v is out-of-bounds")

print("STEP 5 : incremental GC slices arrays in chunks of 128 elements")
n = 2**61 - 1
print(f"  > slices needed for len={n}: {n // 128} -> GC cannot converge within any instruction budget")

print("RESULT: len in [2**61-2 .. 2**61] passes the RTS assert yet wraps the byte size to 8..24 "
      "bytes; header len ~2**61 makes every bounds-checked access out-of-bounds and the "
      "incremental GC non-terminating (permanent canister DoS / brick; CSHF arbitrary "
      "heap read/write within mounted memory)")
```
### Run output (actual)
```text
STEP 1 : read the RTS allocation bound from constants.rs
  > pub const MAX_ARRAY_LENGTH_FOR_ITERATOR: usize = 1 << (usize::BITS as usize - 3);
  > MAX = 1 << (usize::BITS - 3) = 2**61 on 64-bit
STEP 2 : alloc_array computes bytes = (size_of::<Array>() + Words(len)).to_bytes()
  > size_of::<Array>() = 3 words (header tag/fwd/len); to_bytes() = self.0 * 8 (u64 wrap)
STEP 3 : compute the wrapped allocation size for the window that passes the assert
  > len=2305843009213693947  real_bytes=18446744073709551600  wrapped_bytes=18446744073709551600
  > len=2305843009213693948  real_bytes=18446744073709551608  wrapped_bytes=18446744073709551608
  > len=2305843009213693950  real_bytes=18446744073709551624  wrapped_bytes=8
  > len=2305843009213693951  real_bytes=18446744073709551632  wrapped_bytes=16
  > len=2305843009213693952  real_bytes=18446744073709551640  wrapped_bytes=24
STEP 4 : array header stores len ~ 2**61; element access a[i] bounds-checks i < len
  > boundary in Arr.init = arr + 24 + len*8  (wraps below arr+24 -> init loop never runs)
  > heap is allocated 8..24 bytes; every legal access a[i] / a[i] := v is out-of-bounds
STEP 5 : incremental GC slices arrays in chunks of 128 elements
  > slices needed for len=2305843009213693951: 18014398509481983 -> GC cannot converge within any instruction budget
RESULT: len in [2**61-2 .. 2**61] passes the RTS assert yet wraps the byte size to 8..24 bytes; header len ~2**61 makes every bounds-checked access out-of-bounds and the incremental GC non-terminating (permanent canister DoS / brick; CSHF arbitrary heap read/write within mounted memory)
```
### Step-by-step
1. Read the RTS allocation bound MAX_ARRAY_LENGTH_FOR_ITERATOR = 2^61 (rts/motoko-rts/src/constants.rs).
2. Note Words::to_bytes() = self.0 * WORD_SIZE in u64 arithmetic, wrapping silently (rts/motoko-rts/src/types.rs).
3. alloc_array computes size_of::<Array>() + Words(len), i.e. (len + 3) * 8; for len in [2^61-2 .. 2^61] this wraps to 8, 16 or 24 bytes while the assert len <= 2^61 passes.
4. The header stores len ~2^61; codegen bounds checks (idx < len) pass for any i < 2^61, so accesses are out-of-bounds and can target any address in mounted memory via i*8 wraparound.
5. The incremental GC processes arrays in 128-element slices, so a single reachable wrap-size array requires ~2^61/128 slices and never converges (permanent trap / brick).

## Reproduction (verified on-chain, local chain)

The PoC below is a complete standalone script; the run output is the actual captured output.

## Expected vs actual

- Expected: allocation of `(len + 3) * 8` bytes (approximately 2^64 bytes) is refused, or the length is rejected before allocation.
- Actual: allocation wraps to 8-24 bytes; the header still claims `len ~ 2^61`; bounds checks pass for huge indices; the incremental GC cannot finish. End-to-end validation on an Internet Computer replica (e.g. `Array.init(0x1FF_FFFF_FFFF_FFFF, 0)` followed by one element access, default config) is the remaining step; the arithmetic and the code path are fully deterministic and reproduced above.

## Root cause

The length bound was chosen for the optimized array iterator (compact-number encoding — see the comment in `constants.rs`), not for the allocator's multiplication. `(len + 3) * 8 mod 2^64` collapses to a tiny value exactly inside the window that the `assert` permits; no earlier guard in codegen or RTS rejects it.

## Impact

- Permanent DoS of the canister: any message that allocates such an array triggers out-of-bounds traps or a never-converging incremental GC, and the corrupted object persists in state, blocking subsequent messages and upgrades.
- Within mounted memory, `a[i]` with `i*8` wraparound reads or writes arbitrary words: heap disclosure of Candid/request buffers and of later-allocated user data; a careful writer can forge object headers (persistent corruption across messages).

## Suggested remediation

- Reject lengths whose byte size overflows before allocation: `len.checked_mul(WORD_SIZE)` etc., and cap `len` below `2^64 / WORD_SIZE - header` (or below `u32::MAX` for the 32-bit path).
- After allocation, verify the recorded `len` against the bytes actually allocated.
- Align `MAX_ARRAY_LENGTH_FOR_ITERATOR` with the maximum allocation-safe length and add a codegen-side trap for larger `Nat` values.

## Verification & honesty notes

Verified statically: the wraparound arithmetic is proven numerically, and the codegen-to-RTS call chain (`BigNum.to_word64` -> `alloc_array` -> `Words::to_bytes`) plus header layout (3 words) were traced line by line. Not yet executed on a real IC replica/wasm runtime because no such runtime was available in the audit environment; this is the only remaining step before confirming end-to-end impact. Severity is reported as HIGH (guaranteed DoS) with the heap read/write as an additional in-scope consequence.

ggreif:
Fixes #6299.

## What was wrong

`alloc_array` bounded its length by `MAX_ARRAY_LENGTH_FOR_ITERATOR`:

```rust
assert!(len <= MAX_ARRAY_LENGTH_FOR_ITERATOR);          // 2^61 on 64-bit, inclusive
let skewed_ptr = mem.alloc_words(size_of::<Array>() + Words(len));
```

That constant is documented as an *iterator* bound — array lengths must fit in
signed compact numbers for the optimized iterator — not an allocation bound. Reusing
it lets `Words::to_bytes`, a plain `self.0 * WORD_SIZE` (`types.rs:49`), wrap for
lengths in `2^61-2 ..= 2^61`: the heap hands out 8, 16 or 24 bytes for an object whose
header records ~2^61 elements.

Reproduced on EOP before the fix, with `Array.init(2^61-1, 0)`:

```
allocating 2_305_843_009_213_693_951
survived; size = 2_305_843_009_213_693_951
a[0] = 2_305_843_009_213_693_951
```

The allocation succeeds, `size()` reports 2^61-1, the init loop never runs (its
boundary `arr + 24 + len * 8` wraps below `arr + 24`), and `a[0]` reads a word that
isn't the array's — 16 bytes were allocated for an object whose header alone needs 24.
So it is an out-of-bounds read/write over the canister's own heap, and under EOP the
object persists; if it is ever reachable, marking it needs ~2^54 increments at
`SLICE_INCREMENT = 128`.

Two notes on scope, both verified:

* The **32-bit** build is not affected: its bound is 2^29, and `(2^29 + 3) * 4` cannot
  wrap in 32 bits.
* Only the **release** RTS is affected. `[profile.dev]` sets `overflow-checks = true`,
  so `--sanity-checks` already caught this — `panic occurred in file 'src/types.rs' at
  line 49: attempt to multiply with overflow`. `[profile.release]` does not.

## The fix

A separate allocation bound, so the two concerns stop sharing a constant:

```rust
pub const MAX_ARRAY_LENGTH: usize = if usize::BITS == 64 { 1 << 48 } else { 1 << 29 };
```

2^48 elements are 2 PiB of payload — unallocatable by orders of magnitude — and leave
the size computation 13 bits of headroom. `alloc_array` asserts against it, a static
assertion ties it to the arithmetic it protects:

```rust
const _: () = assert!(MAX_ARRAY_LENGTH + size_of::<Array>() / WORD_SIZE <= usize::MAX / WORD_SIZE);
```

and `Arr.alloc` tests the count with a single `shrU 48`, so the user-visible failure is
`array too large` rather than `RTS error: RTS panicked`. `Array.init` and
`Array.tabulate` both allocate through it, as do the Candid vector paths.

Lengths in `[2^48, 2^61]` previously ended in an out-of-memory trap and now trap on the
bound instead — same outcome, better message.

## Testing

* new `test/run-drun/issue-6299.mo`, covering the whole wrap window (`2^61-2`, `2^61-1`,
  `2^61`) plus `2^48` and a control that small arrays still allocate
* `test/run-drun` (499) and `test/run` green locally; the RTS builds clean for both
  word sizes, which exercises the new static assertions


Kamirus (approved):
LGMT

PR #6318, opened 2026-08-20
fix: --stable-baseline honors already-applied migrations in the M0254/M0267 chain check

Kamirus:
With `--enhanced-migration`, the `--stable-baseline` check replayed the whole migration chain to compute the fields demanded from the previous version, ignoring the migrations the baseline itself records as already applied. Any check after a field-dropping migration had run then failed with M0267 for the dropped fields — every enhanced-migration project hits this on the deploy after such a migration, unless tooling keeps the full chain forever. This is the failure behind the AGE-425 TrueLink incident.

The demand set is now computed at the chain's resume point: the backward walk over the chain snapshots the demanded fields where the baseline's applied-migration label sits — the same resume semantics `T.pre` already used for the adjacent compatibility check — and partitions that set against the baseline. Genuinely missing fields still error: a field added to the actor with no migration producing it is caught by M0267 even when the baseline sits at the chain end, where the compatibility check alone would miss it (it treats chain-end fields as optional).

Second, M0254 with a baseline is narrowed to real findings. M0254 flags requirements moc cannot verify; once a baseline settles the demand either way, warning on explained fields repeated a verified fact on every check — one warning per stable field, forever, once the chain was fully applied. Explained fields are now silent and unexplained fields error with M0267. M0254 is unchanged in the no-baseline mode.

Before (actor `{y}` whose single-migration chain dropped `x`, checked against its own emitted signature):

```
main.mo:9.1-11.2: type error [M0267], initial actor requires field `x` of type
  Nat; not found in the previous version — write a migration that produces it
enhanced-migration/enh-mig-test4/m1.mo:0.1: warning [M0254], initial actor requires field `y` of type
  var Nat
```

After: the same check is silent (exit 0). When a field is genuinely unexplained at the resume point, the error names the boundary instead of the (irrelevant) initial actor:

```
main.mo:9.9-9.10: type error [M0267], upgrade resuming after migration `m2` requires field `c` of type
  var Nat; not found in the previous version — write a migration that produces it
```

The fail tests model one project through its lifecycle — actor `{b, c}` with `m1 : {a} -> {a, b}` and `m2 : {a} -> {}` — checked against a legacy pre-EM baseline (full replay), a baseline at `m1` (`m2` pending, `b` demanded from the deployed state), and a baseline at `m2` (nothing pending, `a` no longer demanded), plus the error case where `c` is missing from the fully-applied baseline.

[Mops regression test with this moc fix version released as 1.14.1-fix-stable-baseline](https://github.com/caffeinelabs/mops/pull/791)

Supersedes [#6317](https://github.com/caffeinelabs/motoko/pull/6317) — its repro commit is the base of this branch, with the tests reworked into the lifecycle matrix above.

## What is unchanged

- Diagnostics only: interpreters, codegen, and the runtime upgrade path are untouched, as is `match_stab_sig` (the `.most`-vs-`.most` upgrade gate).
- Legacy `PrePost` baselines carry no migration labels, so an enhanced migration whose equivalent already ran under the legacy `(with migration = ...)` syntax is not detectable as applied — such baselines keep full-replay behavior.
- `-E=M0254` without a baseline still works as a self-containedness gate.


alexandru-uta (approved):
logic seems fine to me, let's see if agents find other corner cases

PR #6325, opened 2026-08-25
feat: validate the migration directory against the deployed history recorded by `--stable-baseline`

Kamirus:
A `Multi`-version `.most` baseline records the chain of migrations that produced the deployed state, but nothing checked that the local migration directory still agrees with it. Each way the two can drift is a real deployment hazard, and none was diagnosed:

* a deployed migration **deleted** while newer ones are kept — no local file accounts for a step that ran;
* a deployed migration **edited in place** — the file no longer describes what the canister actually did;
* a local migration **backdated** to sort before the deployed head — it can never run, because the upgrade resumes past it.

With `--stable-baseline` and `--enhanced-migration`, the directory is now merge-walked against the recorded chain (both are sorted by the labels migrations run in), and each disagreement reports **M0268 against the offending file**:

```
m0.mo:  M0268  migration `m0` is not part of the deployed history recorded by the stable baseline
m1.mo:  M0268  migration `m1` no longer matches the deployed history: it now has type
                 (old : {a : Nat}) -> {a : Nat}
               but the stable baseline records
                 {} -> {a : Nat}
main.mo:M0268  deployed migration `m2` is missing from the migration directory;
               only the oldest migrations may be trimmed away
```

M0268 is a warning that **defaults to error**, so a user who diverges intentionally has an escape hatch (`-W=M0268` demotes, `-A=M0268` silences) without losing the chain walk's own validation. Trimming away only the *oldest* recorded migrations stays legal and silent, as do pending migrations sorting after the head.

## What is unchanged

The chain walk and the baseline boundary sweep are untouched — this check only layers on top, and is inert without a `Multi` baseline. The duplicate M0169/M0170/M0263/M0267 reports at the baseline boundary are out of scope here; [#6281](https://github.com/caffeinelabs/motoko/pull/6281) explores that separately. Split out of [#6281](https://github.com/caffeinelabs/motoko/pull/6281) for clean review, with an independent implementation.

The `em-baseline-history-mismatch` test hits all three drift modes at once; `em-baseline-trimmed-history` pins that a legal trim stays silent (as a silent test, it has no `.ok` files by design).

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,14 @@
 
 * motoko (`moc`)
 
+  * feat: with `--stable-baseline` and `--enhanced-migration`, the migration
+    directory is now validated against the migration history the baseline
+    records as already applied: a deployed migration that was deleted (unless
+    all older ones are deleted too), edited in place, or a local migration
+    backdated to sort before the deployed head reports the new M0268
+    diagnostic, a warning treated as an error by default (demote with
+    `-W=M0268`, silence with `-A=M0268`) (#6325).
+
   * bugfix: with `--stable-baseline` and `--enhanced-migration`, the M0254/M0267
     check now honors the migrations the baseline records as already applied:
     requirements are computed at the chain's resume point instead of replaying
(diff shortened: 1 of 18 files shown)


alexandru-uta (approved):
the logic seems good; 

shall we try to mirror this with runtime checks as well? like a proper 1:1 check between what the compiler generated and what the old program already applied?

PR #6334, opened 2026-09-01
fix: read Candid record field-skip count at full width (u32)

ggreif:
## What

`find_field` / `skip_fields` in the RTS Candid decoder read the number of
remaining record fields through a `*mut u8` pointer, so the count was
effectively taken modulo 256. A record whose number of trailing extra fields is
an exact multiple of 256 produced a skip count of 0 — the decoder skipped none
of them, left the data pointer mid-record, and went on to accept a message a
conformant decoder would reject, mis-reading following arguments from the
extra-field payload.

This is a decoder conformance issue: extra record fields are valid under Candid
subtyping and must be skipped regardless of how many there are.

## Fix

- **codegen (enhanced):** store the decoded field count as a `u32` (the classical
  backend already does), so both backends write the count at the same width into
  the slot the RTS reads back.
- **rts:** read the count as `*mut u32` instead of `*mut u8`. `u32` rather than
  `usize` matches the four valid bytes each backend stores; both count slots are
  ≥ 4-byte aligned, so the deref is sound on wasm32 and wasm64.

## Test

`test/run/candid-record-field-skip` decodes a hand-crafted record with 256
trailing extra fields. Before the fix it mis-decoded a following argument; now it
correctly traps (`advance out of buffer`), i.e. the malformed message is
rejected. Existing Candid tests (`candid`, `idl-decoding-bug`, `float32-candid`)
are unaffected.


src/codegen/compile_enhanced.ml
```diff
@@ -8345,7 +8345,12 @@ module Serialization = struct
         Stack.with_words env "get_n_ptr" 1L (fun get_n_ptr ->
           get_n_ptr ^^
           ReadBuf.read_leb128 env get_typ_buf ^^
-          store_unskewed_ptr ^^
+          (* Store the record field count as a u32, matching the classical
+             (32-bit) backend, rather than a full 64-bit word, so both backends
+             write the count at the same width into the slot that
+             find_field/skip_fields read back. *)
+          G.i (Convert (Wasm_exts.Values.I32 I32Op.WrapI64)) ^^
```
Kamirus, line 8352: Is this truncating silently as before this PR? Do I understand correctly that before this PR we are truncating 64bit word into 8bit when calling `skip_fields`?
And now in this PR we are doing it similarly but truncating from 64bit to 32bit?
This seems like we just move the threshold, right?

ggreif: It is not directly in the same way as the 32-bit side, but indirectly: By the time `with_record_typ` re-reads a record's field count and applies `WrapI64`, that count has already survived `parse_fields`, meaning its fields fit in the buffer, so it's ≤ buffer_size/2, far below 2³². A count like 0x100000008 is filtered out at the header stage and never reaches the truncation.
This is safe. The truncation to 8 bits (from e.g. 10 bits) would not be safe, as that implies 1024 fields which comfortably fit into the message's type table.

Kamirus: > it's ≤ buffer_size/2, far below 2³².

Ok then! That resolves it

Kamirus (approved):

PR #6343, opened 2026-09-08
fix: validate the M0236 dot rewrite against actual dot resolution

Kamirus:
The contextual dot suggestion (`M0236`) could propose a rewrite that changes what the call resolves to. `check_can_dot` consulted `contextual_dot` directly, skipping the first step of real dot-callee resolution: a function-typed field on the receiver shadows contextual dot. The suggestion's autofix is `MachineApplicable`, so this was not just cosmetic.

Two ways it manifested:

- Receiver types with a same-named function field: `Tree.delete(t, 0)` suggested `t.delete(0)`, which silently calls the record's own `delete : K -> ()` field instead of `Tree.delete`.
- Array/blob/text receivers, where the built-in pseudo-fields shadow: the existing test expectation for `Impl.get(data, Nat.compare, 1)` suggested `data.get(Nat.compare, 1)` — that rewrite resolves to the built-in array `get : Nat -> T` and fails with `M0096`.

The bug evaded detection because the guard `T.eq inferred receiver_ty` sometimes suppressed these calls by accident: with only the receiver constraining the instantiation, a type parameter occurring solely contravariantly (e.g. `K` via `delete : K -> ()`) solves to `Non`, failing the equality. That is why `Tree.size(t)` happened not to warn while `Tree.delete(t, 0)` — where the key argument pins `K = Nat` — did.

The fix encodes the field-shadows-contextual-dot precedence exactly once, in `resolve_dot_callee`, and routes both `infer_callee` (actual typing, diagnostics unchanged) and `check_can_dot` (the M0236 suggestion) through it. The suggestion now fires only when the rewritten callee `e.f` resolves back to the very same `M.f`, so it cannot drift from the actual resolution as that evolves. A non-function same-named field still does not shadow (matching real resolution), and both directions are pinned in `edit-suggestions.mo`.

## Behavior change worth flagging

Core functions that merely wrap built-in pseudo-fields (`Array.size`, `Blob.size`, `Text.size`, `keys`, `vals`, `values`, `chars`, immutable `get`) no longer get the dot suggestion: the rewrite would resolve to the built-in field, and the compiler cannot prove a shadowing field is equivalent to the module function — for user-defined modules it genuinely may not be.


alexandru-uta (approved):
looks good, thanks!

PR #6346, opened 2026-09-08
feat!: shallow-copy var fields in record update; remove --experimental-field-aliasing

Kamirus:
## Why
`{ base with ... }` cannot spread a base that has mutable (`var`) fields: the compiler rejects it with `M0179` unless the experimental `--experimental-field-aliasing` flag is on, which instead aliases inherited `var` fields to the base's cells (entangled state). Neither is right. This makes record update sugar-pure and matches other languages (OCaml-style shallow copy): `{ r with var y = 1 }` behaves exactly like the expanded literal `{ var x = r.x; var y = r.y }`. A non-overwritten base `var` field becomes a fresh mutable cell initialized from the base's value at construction time, so mutating the copy does not touch the base and vice versa.

With shallow copy as the one behavior, the aliasing experiment is obsolete, so the flag and its IR machinery (`RefD`) go away. This is the record-spread-over-var-fields fix from caffeinelabs/motoko#6231 and resolves that issue's "decide the fate of `--experimental-field-aliasing`".

## Scope
- `desugar.ml`: in record-extension lowering, non-overwritten base `var` fields now emit a fresh `VarD` (copy) instead of an aliasing `RefD`.
- `typing.ml`: remove the `M0179` rejection of non-overwritten base `var` fields.
- `mo_interpreter/interpret.ml`: mirror the copy semantics in the reference interpreter so `moc -r` agrees with compiled Wasm.
- Remove `--experimental-field-aliasing` (flag + all `RefD` handling across IR passes, both codegens, and interpreters).
- Tests reworked (`object-extend.mo`) and added (`record-update-var-copy.mo`) to assert copy independence; obsolete fail test for `M0179` removed.

## Tradeoffs
This is a breaking change for the (undocumented) aliasing behavior under the flag. One consequence worth knowing: an inherited *method* on an object still closes over the base's cell, so `{ c0 with d = 2 }.f()` mutates `c0`, not the copy's field. That is inherent to shallow copy and matches how hand-expanded record literals behave; the reworked `object-extend.mo` documents it.

## Blast Radius
Users of record extension with `var`-carrying bases. Previously such code was an error (or aliased under an opt-in experimental flag); now it shallow-copies. Anyone relying on aliasing under `--experimental-field-aliasing` must switch to explicit `var x = base.x` copies or a shared reference. `RefD` is removed from the IR, an internal-only surface.

## Verification
`dune build --root src` clean. Targeted `test-runner` runs pass identically across all four surfaces (syntax interpreter, low-level IR, IR interpreter, compiled Wasm): `object-extend.mo`, `object-extend-type.mo`, `record-update-var-copy.mo`. The full `fail/` suite shows only a pre-existing `.tc-human` note-indentation whitespace divergence in the unrelated `implicit-derivation-*` tests, which reproduces on clean `master` and is not caused by this change.


ggreif:
I think the `var` field aliasing feature is an important one. But I doubt that an all-or-nothing flag (as it is implemented a.t.m.) is the right solution. A fine grained
``` Motoko
{ r1 and r2 var f1, f2 }
```
that aliases mutable `f1` and `f2` (I omitted the optional `with` clause) is the way to go. If the need should arise, of course.

This change seems surgical enough, so that the underlying functionality can be recovered in the future.

ggreif (approved):
Looks just fine, and we have an idea to get it back when needed.

Issue #6344, opened 2026-09-08
RFC: Syntax ergonomics for moc v2


Kamirus:
Scoped alternative to the "redesign switch syntax" item in #6231. Full draft: [`design/SyntaxErgonomicsV2.md`](https://github.com/caffeinelabs/motoko/blob/claude/v2-moc-triage-estimation-t18zsa/design/SyntaxErgonomicsV2.md).

## Motivation

Two recurring irritants:

1. **Token overhead**: `switch (x) { case (?v) { ... }; case (null) { ... }; }` requires parentheses on the scrutinee, parentheses on patterns, and semicolons between cases — none of which disambiguate anything for the reader.
2. **Block-vs-record ambiguity**: `{` can open a block or a record, resolved today by the `(ob)`/`(bl)` grammar parameterization (each position admits only one meaning). The corners are user-visible: returning a record from a `case` needs `{ { ... } }` or `({ ... })`, and `opt ?? { x = 1 }` doesn't parse.

The dominant consumer of this syntax is an LLM, which cuts both ways: extra mandatory tokens are extra places for generated code to go wrong, but *removed* syntax invalidates model priors and turns into compile-retry cost. Hence the rollout principle: **relaxations are additive, removals are deprecations first**, and errors must name their fix so a retry loop self-corrects.

The design follows Rust's handling of the identical struct-literal/block ambiguity: keep parsing local, restrict the rarer meaning per position, provide an explicit escape hatch (`(...)` for records, `do { }` for blocks), and invest in diagnostics rather than grammar surgery.

### P1. Full expressions in scrutinee/condition position — S–M

`switch`/`if`/`while` currently take `exp_nullary(ob)` (atomic only). Change to `exp(bl)`:

```motoko
switch f(x) { ... }        // new
if a and b { ... }         // new
switch ({ x = 1 }) { ... } // record scrutinee now needs parens (Rust rule)
```

Only breakage: a bare record literal as scrutinee/condition (vanishingly rare, mechanically fixed by parens, see P5). Open question: extend to `for` (drop `for (p in e)` parens)?

### P2. Optional `;` between switch cases — XS–S

Every case starts with the `case` keyword, so the separator is redundant for the parser. Purely additive.

### P3. Un-parenthesized case patterns — S

[...]

PR #6348, opened 2026-09-08
feat!: unparenthesized heads for switch/if/while/for, coupled to braced branches

Kamirus:
Part 2 of [#6344](https://github.com/caffeinelabs/motoko/issues/6344), on top of [#6358](https://github.com/caffeinelabs/motoko/pull/6358) (the case/`??`/diagnostics half, now on master). Aligned with the target syntax of [#6352](https://github.com/caffeinelabs/motoko/issues/6352) ("the brace discipline"): the new head forms only admit the brace-discipline style, and the whitespace-sensitive tokens are the **migration bridge** that lets old and new styles coexist — to be retired at v3 once bodies are brace-only, not the destination.

**Unparenthesized heads, coupled to braced branches.** The expression grammar in `src/mo_frontend/parser.mly` takes a third "continuation mode" parameter; heads parse in a restricted mode, so `switch f(x) { }`, `if a and b { }`, `switch p.x { }`, `switch arr[i] { }`, and `for x in xs.vals() { }` all work (the legacy `for (p in e)` form is unchanged). Per #6352, the new forms are **coupled to the brace discipline**: an `if`/`while` head that is more than a single atom requires braced branches (`else if` chains exempt), and the unparenthesized `for` form requires a block body — `if f(x) { e1 } else { e2 }` parses, `if f(x) e1 else e2` does not (new `M0275` diagnostic with a fix-it). Atomic and parenthesized heads keep free-form branches exactly as before, so no existing code changes meaning and no *new* anti-style becomes expressible. The coupling also removes the head/branch boundary ambiguity structurally, which deleted most of the precedence machinery an uncoupled design would need — the one remaining tie-break is `EXP_ATOM` at the atomic-head boundary.

**One design deviation from the RFC draft:** pure Rust-style greedy parsing turned out to break the compiler's own prelude and 60+ corpus sites (`if (c) (e) else (e')`, `if (c) [e]`, `if (c) null`, `if (c) #less else #greater`). The implemented rule instead extends Motoko's existing whitespace-sensitivity precedent (`<`, `>`, `>>`): an **unspaced** `(`/`[` extends the head (call/index, via new `TIGHT_LPAR`/`TIGHT_LBRACKET` tokens), a **spaced** one starts the legacy branch — every legacy branch style keeps parsing, and no existing test needed semantic changes. Similarly, `#` followed directly by an identifier is a variant introducer unless it directly follows an expression-ending token, so `if (c) #less else #greater` works while both `x # y` and `x#y` stay concatenation. This machinery becomes removable once control bodies are brace-only (#6352).

This PR also extends `M0272` to record literals written directly as a head (parenthesize them: `switch ({ x = 0 })`).

## Verification

- Grammar conflict-free under menhir `--strict` with no unused precedence levels; `moc` builds warning-free.
- `test/run`, `test/fail`, `test/repl`, `test/cmp`, `test/idl`, `test/mo-idl`, `test/run-drun` all pass, plus the parser-recovery expect tests and `check-error-codes.py`. The only two local failures — `run/candid-record-field-skip.mo` and `run-drun/test-feature.sh` — fail identically on **unmodified master** (local environment issues).
- Tests: `test/run/switch-ergonomics.mo` (the #6352 north-star syntax only: every new head form, braced arms and bodies) and `test/run/switch-ergonomics-legacy.mo` (the still-supported v2 forms: bare branches and arms, `for (p in e)`, `= e` bodies, the whitespace rules), `test/fail/syntax-record-scrutinee.mo`, `test/fail/syntax-head-braces.mo`, `test/fail/syntax-else-chain-legacy.mo`, `test/fail/syntax-switch-body.mo`, `test/fail/syntax-hash-halfspaced.mo`, `test/fail/par-bad-nocall.mo`.
- `doc/md/examples/grammar.txt` regenerated (with `grammar.sed`/`gen-grammar.sh` taught the new parameterization); Changelog entry under `## Next`.

## Known trade-offs and follow-ups

- Documented breaking corners: bare record scrutinee, `x #y` (half-spaced concat), and a branch starting with unary `-`/`+` after a parenthesized-atom head (zero corpus hits).
- The end state, migration staging, and the retirement plan for the whitespace tokens are specified in #6352; the application-syntax questions (`f x`, arity honesty) are #6349. The `doc/md` manual prose (as opposed to the generated grammar) hasn't been rewritten for the new forms.


ggreif:
I haven't looked at this yet, but this is the v2 change that I am most wary of. Maybe we should huddle. I really liked the succinct
``` Motoko
case null 42;
case (?n) n;
```
style. E.g. the Rust
https://github.com/caffeinelabs/motoko/commit/610beebcdf58392f58ca187079567899848a3869#diff-6cf6452d0d9a035e40077fc9634fa4be1eb6cea611810d55eda276b12b8a5647R1032
is super noisy IMO.

Kamirus:
> I haven't looked at this yet, but this is the v2 change that I am most wary of. Maybe we should huddle. I really liked the succinct
> 
> ```mo
> case null 42;
> case (?n) n;
> ```
> 
> style. E.g. the Rust [610beeb#diff-6cf6452d0d9a035e40077fc9634fa4be1eb6cea611810d55eda276b12b8a5647R1032](https://github.com/caffeinelabs/motoko/commit/610beebcdf58392f58ca187079567899848a3869#diff-6cf6452d0d9a035e40077fc9634fa4be1eb6cea611810d55eda276b12b8a5647R1032) is super noisy IMO.

here is the full design https://github.com/caffeinelabs/motoko/issues/6352 
and in particular this https://github.com/caffeinelabs/motoko/issues/6352#issuecomment-5615414963 shows where blocks are and where they are not.

The main idea is to have consistent syntax, currently you sometimes need parenthesis sometimes not and that is the main problem, and our current solution for that is to advise use of full parenthesis and blocks everywhere.
This design is consistent and have less noise than the currently advised
```
case null { 42 }
case ?n { n }
```

PR #6351, opened 2026-09-09
fix(candid): carry `mixin` member doc comments into generated `.did`

Kamirus:
## Root cause

`moc --idl` renders `///` doc comments on actor methods by looking them up in a trivia table carried on the generated Candid program. That table is built from the actor compilation unit only. Mixin members carry source regions into their own file, which is a separate unit, so those positions are absent and their docs are silently dropped.

Member docs that are declared directly on the actor keep rendering; only `include`d mixin members lose them.

## Before

```candid
service : {
  delta: () -> (nat);
  /// Doc for direct method gamma.
  gamma: () -> (text);
}
```

## After

```candid
service : {
  /// Doc for mixin method delta.
  delta: () -> (nat);
  /// Doc for direct method gamma.
  gamma: () -> (text);
}
```

## What fixes it

The mixin's trivia table is threaded through the existing `include` dataflow (`mixin_data` -> `include_note` -> `mo_to_idl`) and merged into the Candid program's trivia, keyed by source file + line + column so tables from different mixin files cannot misattribute same-position comments.

## What is unchanged

- `type` declaration docs are still not carried into Candid for directly-declared types; this is a pre-existing limitation that affects direct and mixin types equally, so the mixin path now behaves exactly like the direct path. Not addressed here.
- Mixin members with the same name as another included member are already rejected (M0051), so the file-scoped key guards the reachable case of matching positions with distinct names.
- No change to type semantics, subtyping, stable signatures, or the Candid printer format.


src/lang_utils/trivia.ml
```diff
@@ -38,18 +38,22 @@ let string_of_trivia_info (info : trivia_info) : string =
     let trailing = List.map (string_of_trivia (fun _ -> "")) info.trailing_trivia |> String.concat ", " in
     Printf.sprintf "Leading: [%s]; Trailing: [%s]" leading trailing
 
-type pos = { line : int; column : int }
+type pos = { file : string; line : int; column : int }
```
ggreif: I think this can be remodelled in the same way like Wasm.srcloc and Syntax.srcloc are made compatible.

Let me check how that is done, quick...

ggreif: Here's how it's done — `src/lang_utils/source.ml`, from #6001 ("deduplicate `pos` and `region`"):

```ocaml
type pos    = Wasm.Source.pos    = { file : string; line : int; column : int }
type region = Wasm.Source.region = { left : pos; right : pos }
```

The `= Wasm.Source.pos = { … }` is a **type-equality re-export**: `Source.pos` isn't a separate look-alike record, it *is* `Wasm.Source.pos`, with the fields re-exposed so you can construct/match it unqualified. So there's **no conversion anywhere** — a `Syntax` node's `.at` region flows straight onto a Wasm instruction via `@@`, and the source-map emitter reads `.left.file/line/column` off the same record end to end.

Now that this PR gives `trivia.pos` the same `{ file; line; column }` shape, it can be collapsed the same way:

```ocaml
type pos = Source.pos = { file : string; line : int; column : int }
```

making `trivia`'s `pos` the *very same type* as `Source.pos` / `Wasm.Source.pos` — the coercion doesn't get optimized, it gets **deleted**.

ggreif (approved):

src/js/common.ml
```diff
@@ -279,7 +279,7 @@ let js_parse_motoko_typed_with_scope_cache_impl enable_recovery paths scope_cach
   (* senv: accumulated scope from prelude and all transitive imports *)
   | Ok ((_libs, progs, senv, scope_cache), msgs) ->
     let progs =
-      progs |> List.map (fun (prog, immediate_imports, sscope) ->
+      progs |> List.map (fun ((prog : Mo_def.Syntax.prog), immediate_imports, sscope) ->
```
ggreif, line 282:
```suggestion
      progs |> List.map (fun (prog : Mo_def.Syntax.prog, immediate_imports, sscope) ->
```

src/lang_utils/trivia.ml
```diff
@@ -38,18 +38,22 @@ let string_of_trivia_info (info : trivia_info) : string =
...
+    file = lexpos.pos_fname;
+    line = lexpos.pos_lnum;
+    column = lexpos.pos_cnum - lexpos.pos_bol
+  }
 
 module PosHash = struct
   type t = pos
 
-  let equal i j = i = j
+  let equal = ( = )
```
ggreif:
```suggestion
  let equal = (=)
```

src/mo_frontend/typing.ml
```diff
@@ -5536,7 +5536,7 @@ and infer_dec_typdecs env dec : Scope.t =
     | None -> error env i.at "M0226" "unknown mixin %s" i.it
     | Some mix ->
       let open Scope in
-      n := Some({ imports = mix.imports; pat = mix.arg; decs = mix.decs });
+      n := Some({ imports = mix.imports; pat = mix.arg; decs = mix.decs; trivia = mix.trivia });
```
ggreif, line 5539:
```suggestion
      n := Some { mix with pat = mix.arg };
```

Changelog.md
```diff
@@ -4,6 +4,11 @@
 
 * motoko (`moc`)
 
+  * fix: `///` doc comments on members contributed to an actor via a
```
ggreif: after the `feat` items, please

PR #5001, opened 2025-04-03
feat: implement system cost primitives

Kamirus:
Implements the following cost functions in the `Prim` module as described in the [IC interface spec](https://internetcomputer.org/docs/references/ic-interface-spec):
- `cost_call`
- `cost_create_canister`
- `cost_http_request`
- `cost_sign_with_ecdsa`
- `cost_sign_with_schnorr`

PR #6353, opened 2026-09-09
feat: add Prim.costVetkdDeriveKey primitive

### LINKED ISSUE #6337: Add `costVetkdDeriveKey` primitive for `ic0.cost_vetkd_derive_key`
## Request

Add a `costVetkdDeriveKey` primitive wrapping `ic0.cost_vetkd_derive_key`, mirroring the existing `costSignWithEcdsa` / `costSignWithSchnorr`.

## Why

The system API function exists and is the only way to learn the current price of a `vetkd_derive_key` call, but Motoko cannot reach it, so Motoko libraries have no option but to hardcode the fee.

`src/prelude/prim.mo` declares exactly five cost primitives, as one contiguous block (lines 677-685 on `master`):

```motoko
func costCall(methodNameSize : Nat64, payloadSize : Nat64) : Nat
func costCreateCanister() : Nat
func costHttpRequest(requestSize : Nat64, maxResBytes : Nat64) : Nat
func costSignWithEcdsa(keyName : Text, curveEncoding : Nat32) : (resultCode : Nat32, costOrUndefined : Nat)
func costSignWithSchnorr(keyName : Text, algorithmEncoding : Nat32) : (resultCode : Nat32, costOrUndefined : Nat)
```

`vetkd_derive_key` is the one threshold-key operation missing from that set — the two signing operations are covered, the vetKD one isn't. And the gap is not just in the prelude: both codegen backends import exactly five `ic0` cost functions (`compile_enhanced.ml:4927-4931`, `compile_classical.ml:5151-5155`), so `ic0.cost_vetkd_derive_key` is absent from the emitted module's imports. That makes it unreachable from library code even by reaching into `mo:⛔` — it needs a compiler change.

For comparison, Rust has had this since `ic-cdk` v0.18.3, as `ic_cdk::api::cost_vetkd_derive_key(key_name, vetkd_curve) -> Result<u128, SignCostError>`.

## History: this was attempted in #5001 and deferred for an external reason

Worth flagging that vetKD was not left out by design. #5001 originally included a vetKD cost primitive and dropped it because the replica did not export the function yet — from a review comment on `src/codegen/compile_classical.ml` (2025-04-24):

> seems to be not implemented yet:
> ```
> Canister's Wasm module is not valid: Wasm module has an invalid import section.
> Module imports function 'cost_vetkd_derive_encrypted_key' from 'ic0' that is
> not exported by the runtime.
> ... error code Some("IC0505")
> ```

Note the name it was tried under: `cost_vetkd_derive_encrypted_key`, which was later renamed to **`cost_vetkd_derive_key`** (tracking the `vetkd_derive_encrypted_key` → `vetkd_derive_key` rename of the management-canister method).

**That blocker is gone.** The function is now exported under its current name — it's listed in `ic0/ic0.txt` in `dfinity/cdk-rs`, wrapped by `ic0::cost_vetkd_derive_key`, and used in production by `ic-cdk` and by `dfinity/vetkeys`' Rust library. So this issue is really about finishing #5001 now that the runtime supports it, not a new design question.

## Concrete impact

`dfinity/vetkeys`' Motoko library has to pin the fee as a literal, on the path of every vetKD derivation in every canister built on it:

```motoko
// backend/mo/ic_vetkeys/src/ManagementCanister.mo:33
await (with cycles = 26_153_846_153) (actor ("aaaaa-aa") : VetkdSystemApi).vetkd_derive_key(request);
```

Its Rust counterpart attaches the exact cost via the system API. Two consequences of the pinned constant:

- **It can silently go stale.** `26_153_846_153` is `10_000_000_000 × 34 / 13` — the production price on a 34-node signing subnet. It encodes both the base fee and the node count of the subnet holding the key, neither of which the library controls. If either changes, adopters' derivations start failing, and the fix needs a library release *plus* a redeploy by every adopter, since the value is compiled into the wasm.
- **It over-reserves for test keys.** Test keys cost `10_000_000_000`. Attached cycles are reserved for the duration of the call, so every canister on `test_key_1` — i.e. essentially all local and testnet development — ties up ~2.6× the liquid-cycle headroom the call actually needs, and hits `could not perform remote call` earlier than it should.

Tracked downstream as dfinity/vetkeys#442, which is blocked on this.

## Suggested shape

`ic0.cost_vetkd_derive_key : (src : I, size : I, vetkd_curve : i32, dst : I) -> i32` is structurally identical to `ic0.cost_sign_with_ecdsa : (src : I, size : I, ecdsa_curve : i32, dst : I) -> i32`, so the existing implementation should transfer almost verbatim:

```motoko
func costVetkdDeriveKey(keyName : Text, curveEncoding : Nat32) : (resultCode : Nat32, costOrUndefined : Nat) =
  (prim "costVetkdDeriveKey" : (Text, Nat32) -> (Nat32, Nat))(keyName, curveEncoding);
```

plus the `E.add_func_import` line and the `Func.share_code2` block alongside `cost_sign_with_ecdsa` in both codegen backends. The non-zero-result-code convention (invalid key name or curve) matches the two signing variants, so the same error plumbing applies.

## Related

None of the five existing cost primitives are re-exported from `mo:core/Cycles` (or `mo:base/ExperimentalCycles`) — both stop at `balance`, `available`, `accept`, `refunded`, `burn`, plus `add` in base. If that's intentional, ignore this; otherwise a re-export would make the whole set usable without dropping to `mo:⛔`.

Separately, `ic0.canister_liquid_cycle_balance128` is also unexposed. It would let a library preflight a call and report "insufficient cycles" instead of the opaque `could not perform remote call`. Happy to open that separately if it's wanted.

Kamirus:
## What changes for users

Adds `Prim.costVetkdDeriveKey : (Text, Nat32) -> (Nat32, Nat)` to `mo:⛔`, wrapping the IC system call `ic0.cost_vetkd_derive_key`. It mirrors the existing `costSignWithEcdsa` / `costSignWithSchnorr` primitives.

**Before**, a Motoko library on the vetkd path had to pin the derivation fee as a compiled-in literal:

```motoko
await (with cycles = 26_153_846_153) (actor ("aaaaa-aa") : VetkdSystemApi).vetkd_derive_key(request);
```

That constant encodes both the base fee and the signing subnet's node count, so it silently goes stale when either changes, and it over-reserves (~2.6x) for cheaper test keys.

**After**, the library can resolve the current cost at runtime and attach exactly the required cycles:

```motoko
let (code, cost) = Prim.costVetkdDeriveKey(keyName, curveEncoding);
assert code == 0;
await (with cycles = cost) ... .vetkd_derive_key(request);
```

A non-zero `resultCode` signals an invalid key name or curve encoding, `costOrUndefined` holds the cost when the code is zero. This closes the gap in #6337: the two signing cost prims were already exposed, vetKD was the missing one.

## Why

`dfinity/vetkeys`' Motoko library (`ic_vetkeys`) hardcodes the fee and is blocked on this (dfinity/vetkeys#442). Rust got `ic_cdk::api::cost_vetkd_derive_key` in ic-cdk v0.18.3; Motoko had no way to reach the system API. #5001 originally included this prim and dropped it because the replica hadn't exported the function yet — that runtime blocker is gone, so this finishes the #5001 scope under the function's current name.

## Notes

- The primitive is declared in `src/prelude/prim.mo` and dispatched in both codegen backends (`compile_enhanced.ml`, `compile_classical.ml`), importing `ic0.cost_vetkd_derive_key : (i64, i64, i32, i64) -> i32` — structurally identical to `cost_sign_with_ecdsa`. No interpreter arm (cost prims are codegen-only).
- **Verified by exercising the path on PocketIC.** I compiled the emitted wasm and ran `Prim.costVetkdDeriveKey` through this repo's PocketIC harness. It executed end-to-end and returned the same shape as its sibling: in this environment it yields `(2, <large cost>)`, exactly matching `Prim.costSignWithEcdsa` under identical conditions. The code path, ABI, and error-code convention are faithful to the ECDSA/Schnorr twins.
- **Why the end-to-end test is in `run-drun-non-ci`.** `test/run-drun-non-ci/test-cost/test-cost-vetkd.mo` mirrors `test-cost-ecdsa.mo` / `test-cost-schnorr.mo`. It asserts `code == 0` and performs a real `vetkd_derive_key` call, which requires a subnet with a provisioned threshold-signing key. This repo's PocketIC harness provisions no such key (`PocketIcBuilder` app subnet, no key setup), so all three cost prims return the "signing unavailable" error code there. The DRUN-era rationale for keeping these out of CI ("drun limitations") is largely obsolete — the tests execute on PocketIC today — but they still need a keyed subnet to reach their success path, so they stay non-CI by design.
- `test/run-drun/cost-prims-enhanced.mo` (in CI) is extended with the new arm; it exercises the compiler dispatch on the enhanced backend and passes.

Closes #6337

### DIFF
--- a/src/codegen/compile_classical.ml
+++ b/src/codegen/compile_classical.ml
@@ -5153,6 +5153,7 @@ module IC = struct
     E.add_func_import env "ic0" "cost_http_request" [I64Type; I64Type; i] [];
     E.add_func_import env "ic0" "cost_sign_with_ecdsa" [i; i; I32Type; i] [I32Type];
     E.add_func_import env "ic0" "cost_sign_with_schnorr" [i; i; I32Type; i] [I32Type];
+    E.add_func_import env "ic0" "cost_vetkd_derive_key" [i; i; I32Type; i] [I32Type];
 
     E.add_func_import env "ic0" "certified_data_set" (is 2) [];
     E.add_func_import env "ic0" "data_certificate_present" [] [I32Type];
@@ -10499,6 +10500,22 @@ module Cost = struct
           Cycles.from_word128_ptr env
         )
       )
+
+  let vetkd_derive_key env =
+    Func.share_code2 Func.Always env "cost_vetkd_derive_key"
+      (("key_name", IC.i), ("curve", I32Type))
+      [IC.i; I32Type]
+      (fun env get_key_name get_curve ->
+        Stack.with_words env "dst" 4l (fun get_dst ->
+          get_key_name ^^ Text.to_blob env ^^ Blob.as_ptr_len env ^^
+          get_curve ^^
+          get_dst ^^
+          IC.ic_system_call "cost_vetkd_derive_key" env  ^^
+          StackRep.adjust env (SR.UnboxedWord32 Type.Int32) SR.Vanilla ^^
+          get_dst ^^
+          Cycles.from_word128_ptr env
+        )
+      )
 end
 
 (* The actual compiler code that looks at the AST *)
@@ -12744,6 +12761,11 @@ and compile_prim_invocation (env : E.t) ae p es at =
     compile_exp_vanilla env ae key_name ^^
     compile_exp_as env ae (SR.UnboxedWord32 Type.Nat32) algorithm ^^
     Cost.sign_with_schnorr env
+  | OtherPrim "costVetkdDeriveKey", [key_name; curve] ->
+    SR.UnboxedTuple 2,
+    compile_exp_vanilla env ae key_name ^^
+    compile_exp_as env ae (SR.UnboxedWord32 Type.Nat32) curve ^^
+    Cost.vetkd_derive_key env
 
   | SystemTimeoutSetPrim, [e1] ->
     SR.unit, compile_exp_as env ae (SR.UnboxedWord32 Type.Nat32) e1 ^^ IC.system_call env "call_with_best_effort_response"
--- a/Changelog.md
+++ b/Changelog.md
@@ -11,6 +11,13 @@
     receiver (including the built-in fields of arrays, blobs and text)
     suppresses the suggestion (#6343).
 
+  * feat: add `Prim.costVetkdDeriveKey` for querying the cycle cost of the
+    IC `cost_vetkd_derive_key` system call, mirroring the existing
+    `costSignWithEcdsa`/`costSignWithSchnorr` primitives. It takes a `Text`
+    key name and a `Nat32` curve encoding and returns `(resultCode, costOrUndefined)`,
+    where a non-zero `resultCode` signals an invalid key name or curve
+    encoding, and `costOrUndefined` is the cost when `resultCode == 0` (#6337).
+
 ## 1.15.1 (2026-09-02)
 
 * motoko (`moc`)
--- a/Changelog.md
+++ b/Changelog.md
@@ -16,7 +16,7 @@
     `costSignWithEcdsa`/`costSignWithSchnorr` primitives. It takes a `Text`
     key name and a `Nat32` curve encoding and returns `(resultCode, costOrUndefined)`,
     where a non-zero `resultCode` signals an invalid key name or curve
-    encoding, and `costOrUndefined` is the cost when `resultCode == 0` (#6337).
+    encoding, and `costOrUndefined` is the cost when `resultCode == 0` (#6353).
 
 ## 1.15.1 (2026-09-02)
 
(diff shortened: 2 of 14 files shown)


src/codegen/compile_classical.ml
```diff
@@ -10499,6 +10500,22 @@ module Cost = struct
           Cycles.from_word128_ptr env
         )
       )
+
+  let vetkd_derive_key env =
+    Func.share_code2 Func.Always env "cost_vetkd_derive_key"
+      (("key_name", IC.i), ("curve", I32Type))
+      [IC.i; I32Type]
+      (fun env get_key_name get_curve ->
```
ggreif, line 10508: This is basically a specialised copy-and-paste of the callback as in the above primitives. Maybe we should factor out the kernel.

ggreif (approved):
Looks good to me, but a slight refactoring wouldn't hurt.

src/codegen/compile_classical.ml
Kamirus: IMO not worth it since we are dropping compile_classical soon

PR #6357, opened 2026-09-09
feat!: remove --generate-view-queries and --experimental-stable-memory

### LINKED ISSUE #6231: List of planned breaking changes

Kamirus:
Two compiler flags from the [#6231](https://github.com/caffeinelabs/motoko/issues/6231) breaking-change list are gone, together with the functionality behind them:

- `--generate-view-queries` no longer exists. The compiler no longer synthesizes controller-gated `__<var>` query methods for stable variables (nor prefers a `.view()` method when one is present), and they no longer appear in the generated Candid. The no-op `--(no-)experimental-multi-value` stubs are removed alongside; multi-value codegen has been unconditional since [#6266](https://github.com/caffeinelabs/motoko/pull/6266).
- `--experimental-stable-memory <n>` no longer exists. Using the `ExperimentalStableMemory` library, or the `stableMemory*` primitives behind it, is an unconditional `M0199` error. `M0199` is now an error code, so `-A=M0199` is rejected too.

## Before / After

```
$ moc --generate-view-queries a.mo
$ moc --experimental-stable-memory 1 esm.mo
```

```
$ moc --generate-view-queries a.mo
moc: unknown option '--generate-view-queries'.
$ moc --experimental-stable-memory 1 esm.mo
moc: unknown option '--experimental-stable-memory'.
$ moc esm.mo
esm.mo:4.9-4.29: type error [M0199], this code is (or uses) the deprecated library `ExperimentalStableMemory`.
Please use the `Region` library instead: https://docs.internetcomputer.org/languages/motoko/icp-features/stable-memory/
```

## Migration

Canisters still on `ExperimentalStableMemory` must move to `Region` before upgrading `moc`; there is no longer an opt-in. Code that relied on the generated `__<var>` queries has to declare its own query methods.

## Tests and docs

The `data-view-*` tests and `doc/md/icp-features/view-queries.md` are deleted. The `ExperimentalStableMemory` test suites (`stable-mem-*`, `region0-*`, `regions-*`, `stable-size-*`, `stable-regions-*`, `stable-log`, `stable-region-migration`, the ESM bench tests, ...) are deleted rather than ported to `Region`: classical persistence is being removed in [#6362](https://github.com/caffeinelabs/motoko/pull/6362), so only enhanced-orthogonal-persistence coverage needs to survive, and `Region` already has its own suites. Tests that only used stable memory as a convenient blob source (`blob-upgrade`, `blob-stabilization`, `destabilization-crash`, `large-destabilization`, `gc-random-test`, `life-mut`, `stabilize-beyond-stable-limit`) keep their purpose and now build the blob from a rope or a `Region`. `experimental-stable-memory-error.mo` pins the unconditional `M0199`, and `test/test-moc.js` pins that all four flags are unknown options again.

## What is unchanged

The `stableMemory*` primitives remain in `prim.mo` and codegen (they are what `M0199` fires on); dropping them entirely is part of the persistence removal in [#6362](https://github.com/caffeinelabs/motoko/pull/6362). The runtime's `Region` load/store loops are untouched here: porting `blob-upgrade` to `Region` exposed an out-of-bounds access-vector read for whole-region loads that end on a block boundary, fixed separately in [#6373](https://github.com/caffeinelabs/motoko/pull/6373) (already on `master`).

Closes [#6231](https://github.com/caffeinelabs/motoko/issues/6231).


Kamirus:
Addressed the review on this branch (merge from `master`, no force-push).

**Fixed**
- **BLOCKER (mixin trivia)** — the `gather_views` conflict in `src/mo_idl/mo_to_idl.ml` was a real one; resolved by splitting `gather_mixin_trivia` out so master's mixin-doc-comment fix (#6351) and the view-gathering removal both survive.
- **M0199 registration** — it *was* registered as a warning code while `typing.ml` emits it unconditionally. It now lives in `error_codes` (not `warning_codes`), and `test/fail/ok/warn-help.{tc,tc-human}.ok` are re-goldened (the warning listing no longer contains M0199). Side effect worth noting: it can no longer be silenced with `-A=M0199`.
- **Coverage restored on `Region`** — `region-overflow.mo`, `query-size-overflow.mo`, `stable-size-overflow.mo`, `stable-size-no-overflow.mo`, `region-pay-as-you-go.mo` (+ goldens). These re-pin out-of-bounds trap behaviour and `rts_(logical_)stable_memory_size` accounting, which the deletions had removed *entirely* (no surviving test matched either).
- **`--generate-view-queries` removal pinning** — added to `test/test-moc.js`, which already had the unknown-flag convention. Not a `test/fail` test: `run-test` captured the whole `--help` usage dump as the golden, which isn't a useful expectation.
- **Migration docs** — new **Removed options** section in `doc/md/reference/compiler-ref.md` covering all three removed flags.
- **`StableLog.mo`** — deleted (last stale `ExperimentalStableMemory` example, unreferenced; `examples/check.sh` now exits 0).
- **`stable-region0-danger.mo`** — dangling `import Region0 "stable-mem/StableMemory"` removed.

**Not restorable, deliberately** — the `region0-*`/`stable-mem-*` *API* twins and `regions-base-*`: no in-tree test library exposes a region-0 handle (`Region.region0()` was never exported), `StableMemory.grow` is the only route to region 0 and now trips the unconditional `M0199`, and master's `-A=M0199` suppression is invalid against an error code. Their old goldens also pinned allocation-order-dependent region ids.

**Verification** — `test/run`, `test/fail`, `test/mo-idl`, and `test/run-drun` in both persistence modes (456/456 classical, 456/456 EOP) all green. One honest gap: `node test/test-moc.js` is environmentally broken here — the unmodified baseline also exits 1 with empty output — so the flag-rejection behaviour was confirmed by direct probe instead.

PR #6361, opened 2026-09-09
feat: suggest contextual dot notation (M0236) by default

Kamirus:
When a call like `Map.filter(map, ...)` could instead be written `map.filter(...)`, `moc` now flags it by default. Before, this suggestion (`M0236`) was opt-in — you only saw it by passing `-W M0236`.

**Before** — default flags, no M0236 output:

```
$ moc --check test.mo
# <- nothing about dot notation
```

**After** — default flags now warn:

```
$ moc --check test.mo
test.mo:58.16-58.42: warning [M0236], You can use the dot notation `[3, 1, 2].sort(...)` here
test.mo:77.49-77.69: warning [M0236], You can use the dot notation `arr.sort(...)` here
```

Opt out where the module-qualified form is clearer, exactly as before for any warning:

```
$ moc --check -A M0236 test.mo
# no M0236 output
```

## What is unchanged

- **`M0223`** (redundant type instantiation) and **`M0237`** (redundant explicit arguments) stay off by default, as discussed in the issue: they compare to `M0236` but are more case-sensitive and costlier to suggest correctly, so they're intentionally not flipped here.
- Everything else in `default_warning_levels` (`M0235`, `M0268`) is untouched.

## Migration

Projects that prefer module-qualified calls and want to keep byte-identical compiler output should add `-A M0236` to `moc` flags / their mops lint config. Projects with no explicit M0236 setting simply start seeing the suggestion.

Split out of [#6231](https://github.com/caffeinelabs/motoko/issues/6231).

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,7 +8,7 @@
     warns about calls like `Map.filter(map, ...)` that could be written with
     dot notation `map.filter(...)`. Silence with `-A M0236`. The related
     suggestions `M0223` (redundant type instantiation) and `M0237` (redundant
-    explicit arguments) remain off by default (allow with `-W`).
+    explicit arguments) remain off by default (allow with `-W`). (#6361)
 
 ## 1.16.0 (2026-09-09)
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,15 @@
 # Motoko compiler changelog
 
+## Next
+
+* motoko (`moc`)
+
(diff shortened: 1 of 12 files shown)


ggreif (commented):
I don't like this as a `warning`. As it is not a potential correctness, but a style issue. We should not warn about style. A (suppressible) `Note` would be alright IMO.

ggreif (approved):
This just makes an existing warning loud. LGTM.

PR #6362, opened 2026-09-09
feat!: remove classical/32-bit persistence and non-incremental GCs

Kamirus:
Classical (legacy, 32-bit) orthogonal persistence is removed. `moc` now always compiles with enhanced orthogonal persistence (EOP), and the classical-only flags are hard errors. Existing classical canisters keep working: the runtime still reads all earlier classical stable-memory formats, and a classical canister migrates to EOP on its next upgrade (irreversible).

## What you notice

The flags below are no longer accepted and now fail:

```console
# before
$ moc --legacy-persistence -c foo.mo
# compiles a classical (32-bit) canister

# after
$ moc --legacy-persistence -c foo.mo
moc: --legacy-persistence has been removed; enhanced orthogonal persistence is always used. ...
```

```console
$ moc --copying-gc -c foo.mo
moc: --copying-gc has been removed; the incremental garbage collector is always used. ...
```

Removed (hard error): `--legacy-persistence`, `--copying-gc`, `--compacting-gc`, `--generational-gc`, `--rts-stack-pages`, `--skip-gc-deprecation-warning`. The incremental GC is the only GC. `--enhanced-orthogonal-persistence` and `--incremental-gc` remain accepted and describe the default.

`--enhanced-migration` no longer needs `--enhanced-orthogonal-persistence` on the command line; that "flag requires" validation is gone because EOP is always in effect.

`moc.js`: `gcFlags` accepts `"incremental"`, `"enhancedOP"`, `"force"`, `"scheduling"`; `"copying"`, `"marking"`, `"generational"`, `"classicOP"` now throw `Invalid_argument`.

## Migration

- Compile with a current `moc`; every build targets EOP. No flag needed.
- Upgrading a classical canister to an EOP build migrates it permanently (one last deserialize → persistent heap). Compile that upgrade with the explicit `--enhanced-orthogonal-persistence` flag and without `--enhanced-migration`: without the flag the new module traps with "Detected implicit upgrade from classical orthogonal persistence to enhanced orthogonal persistence"; with `--enhanced-migration` it traps with "Cannot upgrade from classical orthogonal persistence with --enhanced-migration". Once migrated, later upgrades need no flag. Downgrades from EOP to classical are not supported.
- Need to *produce* a classical module? Keep an older `moc` (e.g. 1.14.x); the current one can no longer build one.

## What is unchanged

- The classical→EOP upgrade path (graph-copy stabilization) is retained; upgrade behavior is covered by committed `old.wasm` fixtures built with `moc` 1.14.1.
- EOP safeguards (upgrade compatibility check, incremental GC, 64-bit heap) are untouched.
- Actor persistence and stable-variable semantics are unaffected.

## Tests

- `test/fail/removed-flag-*.mo` assert the removed flags exit non-zero with the documented message.
- `test/run-drun/migrate-upgrade-instr-stable-var` and `-stable-region` assert the classical→EOP upgrade reports a real instruction count (not the classical sentinel) before printing it under `Ignore Diff:`; the count differs between debug and release builds, so it cannot be pinned.
- `query-footprint.mo` is renamed `stable-var-query-deprecated.mo`: under EOP its only observable behavior is the `Deprecated with enhanced orthogonal persistence` trap.
- The `ENHANCED-ORTHOGONAL-PERSISTENCE-ONLY` markers are dropped from all tests (no runner reads them anymore); the affected `.tc.ok` goldens are re-accepted for the line shift.
- `test/run/top-bit.mo` has no EOP twin on purpose: it checked the classical-only `i32.shr_u 31 → i32.clz` peephole, which EOP codegen never triggers (Nat32 lives in the upper half of an i64). Runtime semantics stay covered by `words-enhanced.mo`.

## Merge notes

- The `master` ruleset requires the `verify-common-gc` status check, so that job id is kept (with a comment); rename it together with the ruleset.
- Reconciled with [#6356](https://github.com/caffeinelabs/motoko/pull/6356): `MOC_ARGS_DEFAULT` is now empty, since that PR removed `--legacy-actors` and this one removes `--legacy-persistence --skip-gc-deprecation-warning`.
- Merged up to `master` at [#6378](https://github.com/caffeinelabs/motoko/pull/6378). [#6357](https://github.com/caffeinelabs/motoko/pull/6357) and [#6378](https://github.com/caffeinelabs/motoko/pull/6378) deleted the `ExperimentalStableMemory` tests this branch had only re-marked, including `migrate-upgrade-instr-stable-mem`; `-stable-var` and `-stable-region` still cover that assertion. [#6357](https://github.com/caffeinelabs/motoko/pull/6357) also rewrote `destabilization-crash.mo` onto `Region` and bumped its golden trap line; that rewrite is dropped here, because the file is no longer compiled — the test installs a committed classical `old.wasm` whose trap locations come from the 1.14.1 source. That source is now `destabilization-crash.mo.txt`: keeping it as a `.mo` would leave the tree's only `Prim.stableMemory*` call sites after [#6378](https://github.com/caffeinelabs/motoko/pull/6378) removed those primitives.
- [#6356](https://github.com/caffeinelabs/motoko/pull/6356) newly marked `upgrade-hooks` and `map-upgrades` `CLASSICAL-PERSISTENCE-ONLY`, because persistent-by-default makes their upgrade sequences fail EOP's compatibility check. Classical is gone here, so both are dropped like every other classical-only test — their EOP twins `stabilization-upgrade-hooks` and `map-stabilization` already run the same sequences.

Part of [#6231](https://github.com/caffeinelabs/motoko/issues/6231).


.github/workflows/test.yml
```diff
@@ -137,24 +113,23 @@ jobs:
           path: ${{ env.UPLOAD_PATH }}
           retention-days: 5
 
-  verify-common-gc:
-    needs: [common-tests, gc-tests]
+  verify-common-tests:
```
Kamirus: `verify-common-gc` is a required CI check, we need to remember to update our infra -- ideally having a single `ci` gate workflow

PR #6366, opened 2026-09-10
style: compress the comments added by the running-GC cache change

Kamirus:
Targets [#6111](https://github.com/caffeinelabs/motoko/pull/6111) so it can be merged into that branch before it lands on master.

The new barrier code carried multi-paragraph comments that restated the diff, the PR discussion, and the code below them, wrapped mid-sentence at 70 columns. Each one is now a single line saying only the non-obvious *why*: the re-seat after instantiation, the raw i32 condition in `if'`, the stack shape through the block params, the persisted `State` layout contract, and the test heap sizing that makes the upgrade land mid-cycle. Everything else is dropped. The test comment also no longer claims the cache is checked on every barrier; it is checked at GC schedule points.

No code changes.


ggreif (approved):
Thanks!

PR #6367, opened 2026-09-11
fix(rts): tighten Candid opt subtype in idl_sub — T <: ?U iff T=null, T=reserved, or T<:U

Kamirus:
Fixes the unsound Candid subtype checking for opt types in `idl_sub()`.

## Problem
The `sub()` function at `idl.rs:1067` had a blanket rule:
```rust
(_, IDL_CON_opt) => return true, // apparently, this is admissable
```
This made **any** type a subtype of any opt type, which is unsound. For example, `{a: Text} <: ?Nat` was accepted.

## Fix
Replaced with arms that mirror what the Motoko decoder can actually handle:

```rust
(IDL_CON_opt, IDL_CON_opt) => {
    // recursive constituent check
    if sub(rel, p, typtbl1, typtbl2, end1, end2, t11, t21) { return true; }
    else { break 'return_false; }
}
(IDL_PRIM_null, IDL_CON_opt) => return true,  // null <: any opt
(_, IDL_CON_opt) => {
    let t21 = sleb128_decode(&mut tb2);
    if u1 == IDL_PRIM_reserved { return true; }  // reserved -> ?null (decoder path 2)
    if sub(rel, p, typtbl1, typtbl2, end1, end2, t1, t21) { return true; }  // T <: U => T <: ?U (decoder path 3)
    else { break 'return_false; }
}
```

This is sound because the Motoko Candid decoder handles three cases for opt types:
1. Wire `null` → `?null`
2. Wire `reserved`/`any` → `?null`  
3. Try decoding as constituent T, wrap in `?T` if no coercion error

`idl_sub` only gate-keeps func/actor reference compatibility before decoding — it should reject what the decoder cannot handle, but accept what it can.

The fix correctly rejects genuinely unsound cases like:
- `Nat <: ?Bool` — `sub(nat, bool)` → false
- `{a:Text} <: ?Nat` — `sub(record, nat)` → false

While preserving valid patterns:
- `Nat <: ?Nat` — `sub(nat, nat)` → true (decoder path 3)
- `reserved <: ?U` — special case (decoder path 2)
- `null <: ?U` — trivial

## Test changes
- `idl-sub-ho.drun-run.ok`: only `wrong_0_4` (Nat <: ?Bool rejected) and `wrong_1_0` (Bool <: ?Int rejected) change to `wrong_*`; `wrong_3_0` and `wrong_6_0` (both `reserved <: ?U`) remain `ok`

## Also included
- Remove misleading "traps" comments from `idl-opt-tests.mo` — array-opt-recovery works correctly on IC replica
- Add missing `.drun-run.ok` files for `idl-opt-tests` and `idl-decoding-bug`
- Add changelog entry under `## Next`

### DIFF
--- a/rts/motoko-rts/src/idl.rs
+++ b/rts/motoko-rts/src/idl.rs
@@ -1063,8 +1063,18 @@ pub(crate) unsafe fn sub(
             (_, IDL_CON_alias) | (IDL_CON_alias, _) => idl_trap_with("sub: unexpected alias"),
             (_, IDL_PRIM_reserved)
             | (IDL_PRIM_empty, _)
-            | (IDL_PRIM_nat, IDL_PRIM_int)
-            | (_, IDL_CON_opt) => return true, // apparently, this is admissable
+            | (IDL_PRIM_nat, IDL_PRIM_int) => return true,
+            (IDL_CON_opt, IDL_CON_opt) => {
+                let t11 = sleb128_decode(&mut tb1);
+                let t21 = sleb128_decode(&mut tb2);
+                if sub(rel, p, typtbl1, typtbl2, end1, end2, t11, t21) {
+                    return true;
+                } else {
+                    break 'return_false;
+                }
+            }
+            (IDL_PRIM_null, IDL_CON_opt) => return true,
+            (_, IDL_CON_opt) => break 'return_false,
             (IDL_CON_vec, IDL_CON_vec) => {
                 let t11 = sleb128_decode(&mut tb1);
                 let t21 = sleb128_decode(&mut tb2);
--- a/rts/motoko-rts/src/idl.rs
+++ b/rts/motoko-rts/src/idl.rs
@@ -1,11 +1,11 @@
 #![allow(non_upper_case_globals)]
 
 use crate::bitrel::BitRel;
-use crate::buf::{Buf, read_byte, read_word, skip_leb128};
+use crate::buf::{read_byte, read_word, skip_leb128, Buf};
 use crate::idl_trap_with;
 
-use crate::memory::{Memory, alloc_blob};
-use crate::types::{TAG_BLOB_B, Words};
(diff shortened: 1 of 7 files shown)


rts/motoko-rts/src/idl.rs
```diff
@@ -1061,10 +1061,30 @@ pub(crate) unsafe fn sub(
...
+            (IDL_CON_opt, IDL_CON_opt) => {
+                let t11 = sleb128_decode(&mut tb1);
+                let t21 = sleb128_decode(&mut tb2);
+                if sub(rel, p, typtbl1, typtbl2, end1, end2, t11, t21) {
+                    return true;
+                } else {
+                    break 'return_false;
+                }
+            }
+            (IDL_PRIM_null, IDL_CON_opt) => return true,
```
ggreif, line 1076: Can this be lumped into line 1064?

rts/motoko-rts/src/idl.rs
```diff
@@ -1061,10 +1061,30 @@ pub(crate) unsafe fn sub(
...
+            (_, IDL_CON_opt) => {
+                let t21 = sleb128_decode(&mut tb2);
+                if u1 == IDL_PRIM_reserved {
+                    return true;
+                }
+                if sub(rel, p, typtbl1, typtbl2, end1, end2, t1, t21) {
```
ggreif: `if u1 == IDL_PRIM_reserved || ...` combining with above

test/run/idl-opt-tests.mo
```diff
@@ -60,7 +60,7 @@ do {
   let o = (from_candid b) :?[Nat];
   P.debugPrint(debug_show(o));
   assert o == null;
-}; // traps due to broken back-tracking in array decoding
```
ggreif, line 63: Do we have an exact explanation why this succeeds now? I wonder how this was not regarded before as a symptom of the unsoundness you found...

PR #6373, opened 2026-09-16
fix(rts): out-of-bounds access-vector read in staged region loads/stores

Kamirus:
`region_load`/`region_store` copy a range block by block. After the first (possibly partial) block, the loop resolves the *next* block via `relative_into_absolute_info(offset + i)` before checking whether the request is already satisfied. When a block-aligned range ends exactly at the end of the region (e.g. `Region.loadBlob(r, 0, 16 MiB)` on a region grown to exactly 256 pages, i.e. two 8 MiB blocks), the last full-block copy leaves `i == len`, and that lookup asks for block rank `block_count`, one past the region's access vector.

Debug RTS builds (`--sanity-checks`) trap:

```
panic occurred in file 'src/region.rs' at line 149: assertion failed: (i as usize) * 2 + 1 < self.0.len().as_usize()
```

Release builds do an out-of-bounds `u16` read whose value is never used, since both branches below the lookup are already guarded by `len > i`, which is why this never showed up in production and only surfaces in the sanity-check test flavours. The fix breaks out of the loop before resolving a block once `i` has reached the requested length.

This is the same off-by-one that `relative_into_absolute_span` a few lines up is written to avoid: behind a `len == 0` early return it resolves the end block from `offset + len - 1`, the last byte of the range, never one past it. The guard restores that rule for the staged loops: only ever translate an offset that is about to be read or written.

`test/run-drun/region-block-boundary.mo` grows a region to two full blocks, loads the whole 16 MiB as a blob, stores it into a second region and checks markers at the end of each block. It traps on `master` under `EXTRA_MOC_ARGS="--sanity-checks ..."` and passes with the fix in the default, EOP, and both sanity flavours.

Split out of [#6357](https://github.com/caffeinelabs/motoko/pull/6357) for clean review.


Changelog.md
```diff
@@ -10,6 +10,11 @@
   * perf: don't GC trace dummy coercion markers for freshly Candid-decoded
     objects (#6370).
 
+  * bugfix: `Region.loadBlob`/`Region.storeBlob` no longer read one block past
+    the end of a region's block table when a block-aligned range ends exactly
+    at the end of the region (a debug-RTS assertion failure under
+    `--sanity-checks`; a harmless out-of-bounds read otherwise) (#6373).
```
```suggestion
    at the end of the region (#6373).
```

rts/motoko-rts/src/region.rs
```diff
@@ -1028,6 +1028,13 @@ pub(crate) unsafe fn region_load<M: Memory>(_mem: &mut M, r: Value, offset: u64,
         // Do rest of block-sized reads.
         // (invariant: they always occur at the start of a block).
         loop {
+            // The initial read may already have covered the whole range
```
ggreif: drop this comment? It is in the MR description and `git blame` finds it.

rts/motoko-rts/src/region.rs
```diff
@@ -1084,6 +1091,13 @@ pub(crate) unsafe fn region_store<M: Memory>(_mem: &mut M, r: Value, offset: u64
         // Do rest of block-sized writes.
         // (invariant: they always occur at the start of a block).
         loop {
+            // The initial write may already have covered the whole range
```
ggreif: Ditto.

ggreif:
The correct idiom already exists a few lines up in this file: `relative_into_absolute_span` resolves its end block as

```rust
let final_offset = offset + len - 1;   // last byte, never one past
let (_, b2, _) = self.relative_into_absolute_info(final_offset);
```

behind a `len == 0` early return — precisely so it never asks for a block one past the range. The staged loops instead resolved `offset + i` where `i` can equal `len`, which is the same off-by-one that function is written to avoid.

Worth a sentence in the description: it makes the guard obviously right rather than merely effective, and it points at the general rule for this file — only ever translate an offset you are about to read or write.

test/bench/ok/region-mem.drun-run.ok
```diff
@@ -1,5 +1,5 @@
 ingress Completed: Reply: 0x4449444c016c01b3c4b1f204680100010a00000000000000000101
 ingress Completed: Reply: 0x4449444c0000
-debug.print: {heap_diff = 0; instr_diff = 6_863_837_595}
+debug.print: {heap_diff = 0; instr_diff = 6_838_671_771}
```
ggreif, line 3: LLVM doing its magic here!

ggreif (approved):
Looks good to me! Thanks for looking into this!

PR #6375, opened 2026-09-16
perf(types): canonicalize acyclic type graphs bottom-up instead of by fixed point

ggreif:
Fixes the long-standing complaint that `test/bench/typtbl.mo` takes over two
minutes to run.

## Diagnosis

The type checker is not implicated: `moc --check` on that file takes **0.031s**.
All the time is codegen, and it is **cubic in the nesting depth**:

| depth | compile | implied exponent |
| ----- | ------- | ---------------- |
| 256   | 1.32s   |                  |
| 512   | 10.01s  | n^2.92           |
| 768   | 36.83s  | n^3.21           |
| 1024  | 94.53s  | n^3.28           |

`ExpGraph.combine` finds the coarsest bisimulation by partition refinement,
restarting from the one-class partition and re-splitting every node on each
pass. A pass distinguishes one more level of structure, so a chain of `n`
distinct nodes needs `n` passes. `typ_hash` canonicalizes on every call, and
codegen names each generated `@deserialize_go` / `@buffer_size` function by the
hash of its type — so a type nested `n` deep is hashed once per level.

## Fix

On an acyclic graph bisimilarity is just structural equality, which can be read
off bottom-up in a single topological pass. Cyclic types keep the fixed point,
where it is genuinely needed: `type A = ?[A]` and `type B = [?B]` denote the
same regular tree without being structurally equal.

| depth | before | after |
| ----- | ------ | ----- |
| 256   | 1.32s  | 0.76s |
| 512   | 10.01s | 0.55s |
| 768   | 36.83s | 0.97s |
| 1024  | 94.53s | 1.63s |

`test/bench/typtbl.mo` as a whole: **2m16s → 2.5s**.

## Verification

- **Output is unchanged.** `wasm2wat` of `typtbl.mo` before and after is 390,978
  lines each, differing only in the embedded version string — not one
  instruction.
- `dune runtest mo_types` passes, including the `monolist` / `polylist` cases
  that exercise the cyclic path.
- **No pessimisation.** `test/run` (312 programs, cleaned before each run, quiet
  machine): 163.50s with the patch vs 164.49s without (real), 130.02s vs
  131.42s (user) — a wash, within noise.
- `nix build .#common-tests .#gc-tests` passes locally.


src/lang_utils/expGraph.ml
```diff
@@ -133,9 +133,54 @@ let equiv_classes (type b) (graph : (int * b) Seq.t) : (int IM.t * int) =
   m, size
 
 
+(* Children-before-parents order, or None if the graph has a cycle *)
+let topo_order (graph : 'a t) : int list option =
+  let state = ref IM.empty in (* false = on stack, true = finished *)
+  let order = ref [] in
+  let cyclic = ref false in
```
ggreif: `let <triple> = ` please.

src/lang_utils/expGraph.ml
```diff
@@ -133,9 +133,54 @@ let equiv_classes (type b) (graph : (int * b) Seq.t) : (int IM.t * int) =
...
+  let rec go i =
+    if not !cyclic then
+      match IM.find_opt i !state with
+      | Some true -> ()
+      | Some false -> cyclic := true
+      | None ->
+        state := IM.add i false !state;
+        let (_, args) = IM.find i graph in
```
ggreif:
```suggestion
        let _, args = IM.find i graph in
```

src/lang_utils/expGraph.ml
```diff
@@ -133,9 +133,54 @@ let equiv_classes (type b) (graph : (int * b) Seq.t) : (int IM.t * int) =
...
+   determined by its label and its children's classes, which are already final.
+   This avoids the fixed-point iteration below, which needs one round per level
+   and so is quadratic on deep chains. *)
+let combine_acyclic (type a) (graph : a t) (order : int list) : a t =
+  let module KM = Map.Make (struct type t = a * int list let compare = compare end) in
+  let cls = ref IM.empty in
+  let km = ref KM.empty in
+  let next = start_counting 0 in
+  List.iter (fun i ->
+    let (k, args) = IM.find i graph in
```
ggreif: ditto

src/lang_utils/expGraph.ml
```diff
@@ -133,9 +133,53 @@ let equiv_classes (type b) (graph : (int * b) Seq.t) : (int IM.t * int) =
...
+(* On an acyclic graph bisimilarity coincides with structural equality, so the
+   coarsest classes can be read off bottom-up in one pass: a node's class is
+   determined by its label and its children's classes, which are already final.
+   This avoids the fixed-point iteration below, which needs one round per level
+   and so is quadratic on deep chains. *)
+let combine_acyclic (type a) (graph : a t) (order : int list) : a t =
+  let module KM = Map.Make (struct type t = a * int list let compare = compare end) in
+  let cls = ref IM.empty in
+  let km = ref KM.empty in
+  let next = start_counting 0 in
```
ggreif: triple please

PR #6379, opened 2026-09-17
feat: add subnetSelfNodeCount and costHttpRequestV2 primitives

eichhorl:
Exposes two system API functions that pay-as-you-go HTTPS outcall pricing needs:

- `ic0.subnet_self_node_count`, as `Prim.subnetSelfNodeCount() : Nat32`. A caller computing replication bounds for `flexible_http_request` needs the subnet size, since `total_requests` may not exceed it.
- `ic0.cost_http_request_v2`, as `Prim.costHttpRequestV2(Blob) : Nat`. The argument is the Candid encoding of the parameter record; the result is a 128-bit cycles amount read back from a two-word destination buffer.

These are one change rather than two: version 2 pricing is unusable without both, because a flexible outcall's price depends on the committee size.

## Risk

This touches `src/codegen/` and `src/prelude/`, so it starts as not-low-risk. The evidence below is what I would want to see argued.

Both backends compile `test/run-drun/outcall-prims-enhanced.mo` to wasm, and both validate under `wasm-validate --enable-multi-memory --enable-memory64`. The import signatures come out as intended, and differ by word size:

| | `subnet_self_node_count` | `cost_http_request_v2` |
|---|---|---|
| enhanced OP (64-bit) | `() -> i32` | `(i64, i64, i64) -> nil` |
| classical (32-bit) | `() -> i32` | `(i32, i32, i32) -> nil` |

The count returns a `Nat32` in both, so `i32` either way. The cost function takes three pointer-typed arguments, which is why that row splits.

Two bugs in earlier drafts were caught only by building, and they mark where a reviewer should look hardest:

- `SR.UnboxedWord64 Type.Nat32` in the classical backend type-checked and compiled, then crashed in `StackRep.adjust` during codegen. It must be `UnboxedWord32`.
- The import was declared `[I64Type; I64Type; i]`, copied from `cost_http_request` whose first two arguments are sizes. All three of `cost_http_request_v2`'s arguments are pointer-typed. Enhanced OP worked by coincidence; only `wasm-validate` rejected it.

Five `test/fail` goldens are regenerated. `prim.mo` gaining two entries shifts the suggestion lists those tests assert on; the golden diff contains nothing else.

## Verification

- Both backends: compile to wasm, `wasm-validate` clean, import signatures as tabled above.
- `test/run-drun/outcall-prims-enhanced.mo` passes `[tc] [comp] [valid]`.
- Full `run-drun` suite passed when the change was built. I could not re-run it for this PR, because `dune` is not available in my current environment, so please treat the suite result as needing CI confirmation rather than as re-verified.

## Downstream, and a follow-up ask

The `ic` mops package needs these for its 5.0.0 release, which moves HTTPS outcalls to pricing version 2.

One thing worth fixing separately: `flexible` is a reserved word, and `typ_tag` admits only identifiers, so a Candid variant tag named `flexible` cannot be written by name. The workaround downstream is the hashed label `#_351978059_`, the Candid field hash, verified byte-identical with `didc`. Grammar support for quoted variant labels would let that go away.


Kamirus:
We are in the middle of merging PRs for v2 moc https://github.com/caffeinelabs/motoko/issues/6231 
so this unfortunately would delay merging this until the GA of moc v2

eichhorl:
> We are in the middle of merging PRs for v2 moc #6231 so this unfortunately would delay merging this until the GA of moc v2

@Kamirus thanks for the reply, I'm not really sure what that means. Could you specify what v2 moc is, and what each of the two "this" correspond to? 😄

Is there anything I can do, or a timeline for when the new outcalls support could be picked up?

Kamirus:
> > We are in the middle of merging PRs for v2 moc #6231 so this unfortunately would delay merging this until the GA of moc v2
> 
> @Kamirus thanks for the reply, I'm not really sure what that means. Could you specify what v2 moc is, and what each of the two "this" correspond to? 😄
> 
> Is there anything I can do, or a timeline for when the new outcalls support could be picked up?

Sorry for the confusion! I have started to merge the PRs that are part of the next major version bump for the motoko compiler. Meaning the last v1 moc release is https://github.com/caffeinelabs/motoko/releases/tag/1.16.1 and now ideally only critical bugfixes can be included for v1.
I'll take a look at your PR as soon as we merge [all v2 PRs](https://github.com/caffeinelabs/motoko/issues?q=state%3Aopen%20label%3Amoc-v2), test afterwards and release a proper 2.0.0 moc 🙏

PR #6382, opened 2026-09-18
ci: make pre-releases a first-class release workflow path

Kamirus:
Cutting a pre-release used to need a workaround and left a red workflow run behind.

Making `2.0.0-beta.0` required dispatching the release workflow with an unrelated
`version_suffix` (which produced `moc-1.16.1-v2-1.js`), then hand-renaming every
artifact to the version that was actually wanted. The release body was discarded,
and because the changelog had no `## 2.0.0-beta.0 (date)` section, publishing the
tag fired a run that failed outright
([run 35332369012](https://github.com/caffeinelabs/motoko/actions/runs/35332369012)).

### Before

```shell
# produced moc-1.16.1-v2-1.js and friends, then renamed by hand
gh workflow run release --ref master -f version_suffix=v2-1
```

- the base version was always whatever `Changelog.md` happened to say
- artifacts were renamed after the build, by pattern-matching filenames
- a pre-release body was always the placeholder, never the changelog
- publishing the pre-release left a failing run on the tag

### After

```shell
# produces 2.0.0-beta.0 artifacts, tag and `moc --version`, in one go
gh workflow run release --ref master -f base_version=2.0.0 -f version_suffix=beta.0
```

The version is resolved once, in a new `prepare` job, from two inputs:
**base version** (defaults to the first version in `Changelog.md`) and **version
suffix** (whose presence is what makes the run a pre-release). The resolved
`<base>[-<suffix>]` is used consistently as the release tag, the artifact
filenames, and the version the built compiler reports, so the rename step and its
filename heuristics are gone.

Because the compiler's version is baked in at build time, it reaches the nix
build through `MOTOKO_RELEASE_VERSION`, read by `nix/releaseVersion.nix` under
`--impure`. The override is inert when the variable is unset and never affects a
pure evaluation, so ordinary builds are unaffected.

A pre-release now takes the **unreleased** changelog entries as its body. As
requested, extraction takes the first `## ` section whatever its title — the
`## Next` heading is not required, and renaming it changes nothing. A version
with no changelog section is now a warning with an empty body rather than a hard
failure, which is the production bug that failed the published tag's run.

Publishing a draft creates its tag, which re-triggers the workflow as a push. That
second run now detects the version is already published and skips the rebuild,
rather than failing on duplicate release assets.

### Verified

- real dry run: `nix build --impure .#release-files-macos-latest` with the override
  yields `motoko-Darwin-arm64-2.0.0-beta.0.tar.gz` and `motoko-release-2.0.0-beta.0`
- `nix/releaseVersion.nix`: unset → `1.16.1`; override → `2.0.0-beta.0` (both the
  official and dev derivations)
- version resolution: `2.0.0`+`beta.0` → `2.0.0-beta.0`; empty base → the changelog
  version; malformed bases and suffixes all rejected
- changelog extraction: unreleased mode returns no heading; `mode: version` is
  byte-identical to the old implementation for an official release; the previously
  fatal missing-section case now exits 0 with a warning
- `actionlint` reports only the findings already present on `master`

### Notes

- `release-pr.yml` accepts an optional suffix in the PR title and validates the
  right section per release kind, so a pre-release PR is checked against the
  unreleased entries instead of vacuously passing.
- Not addressed here: `nix/moc.js.nix` does not set `MOTOKO_RELEASE`, so the
  JavaScript artifacts do not report a version the way the native ones do.


Kamirus:
Superseded by #6384 — rewritten to make only the two intended changes.

Issue #6352, opened 2026-09-09
RFC: Target syntax for Motoko v3 — the brace discipline (north star)


Kamirus:
Status: **Draft** — north-star document. #6344 (syntax ergonomics), #6348 (its implementation), and #6349 (retiring the currying illusion) each change syntax incrementally; this RFC fixes the **end state** they should all be judged against, so that no bridge we build points somewhere we don't want to go.

## The problem with deciding piecemeal

Motoko's grammar has two structural ambiguities that every syntax change trips over:

- **block vs record**: `{` can open either, resolved today by per-position grammar modes with user-visible corners (#6344's motivation);
- **head vs body**: once `if`/`while`/`switch` heads may be full expressions (#6344 P1), the parser must decide where the head ends — which #6348 resolves with whitespace-sensitive `(`/`[`/`#` tokens, a workable bridge but a special rule forever if it is the end state.

Rather than resolving these locally per construct, this RFC states four global rules that resolve both ambiguities **positionally and totally**, and classifies every construct against them. The end state is essentially Rust's configuration — with one rule Rust doesn't have that makes Motoko's block-vs-record story cleaner than today's.

## The five rules

1. **Control heads are unparenthesized full expressions, terminated by `{`.**
   `switch f(x) { }`, `if f(x) { } else { }`, `while f(x) { }`, and — new — `for x in xs { }` (the `in` delimits the pattern, the `{` delimits the head; the mandatory `for (…)` parens go). The only head restriction is Rust's: a record literal directly in a head takes parens, `switch ({ x = 0 }) { }` (diagnosed by M0269). A head is otherwise *any* expression — `do { }` included, exactly as Rust admits `if { c } { }` — and the four constructs share one head grammar. Discouraging block-shaped heads such as `if do { … } { … }` is a formatter/lintoko concern (Rust's `clippy::blocks_in_conditions`), not a grammar rule: a "no block-shaped heads" exclusion would be a third kind of position that every external grammar has to reproduce.

2. **Control bodies are always braced.**
   Both `if` branches, `while`/`for`/`loop` bodies, `case` and `catch` arms, `try`/`finally` bodies, `async`/`async*`, function bodies. One deliberate exception, as in Rust: `else if` chains without bracing the inner `if`.

3. **Operator-like keywords keep bare expression operands.**
[...]

PR #6385, opened 2026-09-18
style: modernize in-repo Motoko to v2 brace discipline

Kamirus:
Part of [#6231](https://github.com/caffeinelabs/motoko/issues/6231) point 1 — modernizing the Motoko we ship to v2 syntax and style — for this repo. Covered here: `src/prelude`, the `doc/md` examples and their fenced snippets, `doc/overview-slides.md`, and `samples/`. Split out of [#6381](https://github.com/caffeinelabs/motoko/pull/6381) for clean review.

## Before / after

``` motoko
// before
if (c.revoked) { return };
for (i in range) { f(i) };
switch (x) {
  case 0 A;
  case _ B;
};
let abs = if (v >= 0) v else -v;
```

``` motoko
// after
if c.revoked { return };
for i in range { f(i) };
switch x {
  case 0 { A }
  case _ { B }
}
let abs = if v >= 0 { v } else { -v };
```

The rules are [#6352](https://github.com/caffeinelabs/motoko/issues/6352)'s, applied as far as the v2 grammar allows:

- **Rule 1** — heads are unparenthesized, `for` included, except where parentheses change the parse (below).
- **Rule 2** — every control body is braced: both `if` branches, `while`/`for`/`loop` bodies, `case` and `catch` arms. The `;` that may follow an arm's `}` is dropped; the separator separates nothing, since `case` already ends the previous arm.
- **Rule 3** — operand positions stay bare: `return e`, `break l`, `continue`, and the trailing condition of `loop … while e`.
- `case` patterns parenthesize the *payload*, not the whole pattern: `case #node(n)`, which is the form the RFC calls canonical for symmetry with the expression side.

## The style guide is the specification

`doc/md/reference/style-guide.md` states these rules rather than the transitional grammar. In particular it no longer says bodies are braced "unless they appear nested as an expression and only contain a single expression" — that "unless" is exactly the optional-block alternative rule 5 exists to remove. The two forms it used to endorse, `let abs = if (v >= 0) v else -v` and `func succ(x : Nat) : Nat = x + 1`, now appear only in a closing note that names them legacy, with the reason `= e` still has one use (async forwarding — the open corner from the [rule-2 addendum](https://github.com/caffeinelabs/motoko/issues/6352#issuecomment-5600971241)).

`doc/overview-slides.md`'s control-flow summary — the one place in the repo that states the syntax outright — was updated with it.

Legacy forms stay where the guide presents them on purpose: the `// COUNTER EXAMPLES!` block demonstrates the unbraced-plus-`;` form, and the closing note quotes each retired form once.

## Where the parentheses stay

Dropping them would change the parse:

``` motoko
switch (r, c) { ... };                 // tuple head
switch ({ x = 1 }) { ... };            // record literal head
switch (eval e1, eval e2) { ... };     // tuple of applications
if ((prim "blob_iter_done" : BlobIter -> Bool) i) { ... };  // bare application head
while (await C.read()) < 10 { ... };   // parens bind the await operand
```

A **bare application head swallows the following `{ … }` as a record literal**, and that is the one trap in this sweep: it is not always a syntax error. `switch eval e1 { … }` does fail (M0275), but `if f x == 0 { … }` parses happily as `f` applied to nothing with `x == 0 { a }` as the branch — a silent misparse. Where a head is an application it is therefore written tight (`switch eval(e1)`, `@nextExpiration(@timers)`) or left parenthesized.

`func f() = e` bodies are left alone. `=` bodies bypass the async/shared sugar in `desugar_func_body`, so rewriting them to blocks is a semantic change rather than a stylistic one; retiring them needs the addendum's open question answered first.

## Verification

Every `.mo` file and every parseable `motoko` fence in the touched docs — **338 units** — was re-parsed with the same compiler and compared against the merge base. After normalizing away the three residues the sweep is allowed to leave (`ParP` pattern wrappers, a single-expression `BlockE`, and anonymous-function names, which carry their source position), the **326 units outside `style-guide.md` have identical parse trees**. Twelve `style-guide.md` fences differ, which is the point of that commit.

**No file gains a diagnostic.** `style-guide.md` trades a base-only M0275 for three `unused identifier` warnings on the wrappers added below — the same class as the sixteen already in that file.

Two style-guide examples did not typecheck standalone and now do: the semicolon-rules fence referenced an undeclared `opt`/`f`, and `let delta = switch mode { … }` an undeclared `mode`. The second had been latent — that fence used to die at parse on `if (i < 0) -i else i` (M0275), so the typechecker never reached the line; fixing the M0275 unmasked it. Both are now small `func` wrappers.

> An earlier revision of this description claimed 296 units with diagnostics identical across all of them. That was wrong: the extractor anchored its fence regex at column 0, and `style-guide.md`'s fences are indented four spaces inside list items, so it silently matched **none of them** — the file with the largest diff was not covered at all. The numbers above are from the corrected harness.

`src/prelude/*.mo` cannot be dumped standalone — their `@`-prefixed names are privileged — so they are checked by rebuilding `moc` and dumping the IR of a driver that exercises the touched helpers: identical modulo those same anonymous-function positions and the embedded version string. `test/run` (309) and `test/run-drun` (439) pass.

`doc/schat.mo` and `doc/chat.mo` are pre-existing failures, unchanged here: the first does not parse under either revision, the second is rejected by M0141 for declaring a second actor.

## What is unchanged

- **`test/**`.** It carries the deliberate `*-legacy` twins and the parser tests that exercise the old forms, which v2 must keep accepting. The mechanical part of that sweep wants the formatter from #6352 first, since the goldens embed line/column ranges needing regeneration. As a measure of the remaining work, a brace-aware scan of the test tree finds 549 bare arm bodies and 1249 arm separators.
- **`doc/md/reference/language-manual.md` and `error-codes.md`.** They document the grammar as it is, including `for ( <pat> in <exp> ) <block-or-exp>`; they follow the grammar, not the style guide.
- **`Changelog.md`.** The single match there is inside a released section, which `AGENTS.md` freezes.
- **`doc/site`.** Symlinks to `doc/md`, so the sweep above already covers it.
- **The `persistent` / `stable var` / `.vals()` sweep.** That is point 2, a separate breaking-change pass; none of it appears in this diff.

## Follow-up worth flagging

The bare-application-head misparse above deserves its own issue: `f(x)` and `(f x)` both work as heads, `f x` silently does not, and no diagnostic points at the cause. It disappears at v3 with the whitespace machinery, but it is writable today.

No formatter exists in-tree yet, so this sweep is by hand.


doc/md/fundamentals/types/variants.md
```diff
@@ -39,10 +39,10 @@ let activeUser : Status = #Active;
 let bannedUser : Status = #Banned("Violation of rules");
 
 func getStatusMessage(status : Status) : Text {
-  switch (status) {
-    case (#Active) "User is active";
-    case (#Inactive) "User is inactive";
-    case (#Banned(reason)) "User is banned: " # reason;
+  switch status {
+    case #Active "User is active";
```
Kamirus:
```suggestion
    case #Active { "User is active" }
```
1. Skip optional `;` for cases
2. Always put the block `{ ... }` for the case-body

Apply here and across all codebase (unless in places where legacy syntax is to be presented)

Kamirus: Applied — both rules, across everything this PR touches.

The anchored line already read `case #Active { "User is active" }` in the pushed revision; the comment landed on the pre-push state. What was still missing were two files the earlier sweep had not caught: `doc/chat.mo` and `doc/schat.mo`, each with one `case null return;` arm. Both now match the suggested form, and the trailing `;` following a braced arm is gone everywhere else in `src/prelude`, `doc/md` (including its fenced snippets), and `samples/`.

The sweep is now clean repo-wide outside `test/**` — 0 bare bodies, 0 arm semicolons across all 145 tracked `.md` files and all non-test `.mo` files.

Two carve-outs, matching the "unless in places where legacy syntax is to be presented" clause:

- `doc/md/reference/style-guide.md` presents the legacy forms deliberately; left as-is.
- `test/**` still carries 709 bare arm bodies and 1364 arm semicolons. That is the parser's coverage of the forms v2 must keep accepting, including the deliberate `*-legacy` twins, and the goldens embed line/column ranges that a mechanical rewrite would need to regenerate — so it wants the #6352 formatter first. Flagged in the PR description as a follow-up rather than silently skipped.

Verified parse-neutral the same way as the rest of the PR: re-parsing each touched unit and normalizing away the `ParP` and singleton-`BlockE` wrappers the style itself introduces leaves every changed fence and file tree-identical, with identical diagnostics.

Kamirus: Both applied — and the gap left was the style guide itself.

`split ()` had already picked up the braced form on the push you reviewed (the comment
anchored the pre-push revision); the only files still carrying an unbraced arm were
`doc/chat.mo` and `doc/schat.mo`, each with one `case null return`. Those are fixed, and
the trailing `;` after a braced arm is gone everywhere else in `src/prelude`, `doc/md`
and `samples/`. Outside `test/**` the tree is now clean: 0 bare bodies, 0 arm
semicolons across all non-test `.mo` files and all 145 tracked `.md`.

The part worth flagging is `doc/md/reference/style-guide.md`. Pass 1 had rewritten its
prose for the parenless head but left its Braces section illustrating the old arms — so
the guide was contradicting the rule it exists to state. It now:

- shows `case #less { A }` / `case _ { B }` with no `;` between arms;
- states the arm rule explicitly: braces on a `case` branch are not subject to the
  single-expression exception that applies to `if` and loop bodies;
- drops a concise `switch` expression (`let val = switch f() { case #ok(x) x; case _ 0 };`)
  that the rule now excludes.

Two examples in that file no longer parsed under the current grammar, independent of this
PR — `if (i < 0) -i else i` (a prefix-form branch needs braces) and
`switch (compare(x, y))` — and are corrected to `if i < 0 { -i } else { i }`.

Left as legacy exhibits, per your carve-out: the `// COUNTER EXAMPLES!` block (which
demonstrates exactly the unbraced-plus-`;` form), and the fragment list showing where
parentheses remain *required*. `Changelog.md` is untouched — the one match there sits in
a released section, which AGENTS.md freezes.

Verification unchanged from before: re-parsing each touched fence and file with the same
compiler, normalising away the `ParP` and singleton-`BlockE` wrappers the style itself
introduces, leaves every tree identical with identical diagnostics. The three guide
fences that do differ are the intentional content edits above.

Kamirus: @/tmp/reply3.md

Kamirus:
Thanks for the review — both points are applied.

```suggestion
  case #Active { "User is active" }
```

1. **No `;` between case arms** — dropped throughout. `case` already terminates the previous arm, so the separator was redundant everywhere it appeared.
2. **Every case body braced** — `case #tag { ... }` rather than a bare expression body.

Applied across the whole repo, not just the anchor: 1812 arm sites in 168 files. Every file outside `test/` is now at the north-star form, and `test/**` is deliberately left alone — it is the compiler's regression suite, so its programs are rewrite-neutral by design and changing them would churn expectations without adding coverage.

Three deliberate exceptions, all cases where the old form *is* the subject:

- `doc/md/reference/style-guide.md` — the "COUNTER EXAMPLES!" block, which shows the bare form specifically to mark it wrong.
- `doc/md/reference/language-manual.md` — the BNF `for`-desugaring production, grammar not style.
- `Changelog.md` and `design/` — a frozen released section and draft design docs.

Verification: every touched unit was compared against the merge base by parse-tree diff under the same compiler, normalizing `ParP` and single-expression `BlockE`. 0 regressions, 1 improvement (a fence that previously died at `M0275` now typechecks).

PR #6387, opened 2026-09-21
feat!: two if/while shapes: parenthesized head with bare branches, or any head with braced branches

Kamirus:
`if` and `while` now have exactly two shapes, told apart by the first token of the condition, and the lexer no longer needs to look at whitespace to find the boundary between a condition and its branch.

- A condition that starts with `(` is the whole condition and its branches are free-form, exactly as in moc 1: `if (c) e1 else e2`, `if (x < 0) -1 else 1`, `if (c) #less else #greater`, `while (i > 0) i -= 1`.
- Any other condition is a full expression and the branches or body are blocks: `if c { }`, `if f(x) { } else { }`, `if a + 1 > n { }`, `while n > 0 { }`. The two shapes mix freely in `else if` chains.

`switch` and `for` are unchanged: their heads are greedy full expressions, parenthesized or not.

## Why

[#6348](https://github.com/caffeinelabs/motoko/pull/6348) kept the moc 1 form that skipped *both* the parentheses and the braces (`if c e1 else e2`) next to the new heads, and paid for it with whitespace-sensitive `(`/`[`/`#` tokens, a third grammar parameter, and an `EXP_ATOM` precedence tie-break. That bridge still broke the well-formed moc 1 idiom without saying so: `if (c) -x else y` and `if (c) ^m else m` were read as subtraction and xor on the condition, and M0275 then told the user to parenthesize a condition that already was. In the local mops cache that idiom appears at 29 sites in 7 packages (json, serde, serde-core, xtended-numbers, datetime, rep-indy-hash); the form that skipped both parentheses and braces appears at 6 sites in motoko-core, 8 in OQL, and a handful elsewhere.

Making the opening `(` decide the shape resolves every former conflict structurally, so the grammar has no tie-break left to explain:

| after `if c` / `if (c)`, next token | before | now |
|---|---|---|
| `(`, `[`, `-`, `+`, `^`, `#`, identifier | whitespace tokens and `EXP_ATOM` decide head vs. branch | after `if (` only the moc 1 shape exists; after a bare atom only the braced shape does |
| `{` | `bl` mode (Rust's record rule) | unchanged |
| `f x` juxtaposition in a head | third `L` parameter | the existing `R` mode conditional |

Deleted: the `TIGHT_LPAR`/`TIGHT_LBRACKET`/`TIGHT_HASH` tokens and their lexer rules, `ends_exp`, the third grammar parameter with `exp_cont_tight`, `EXP_ATOM`, the `exp_head_*` duplicate of the expression chain, and `if_braced`. Added: a third begin-mode `hd` (no leading `{`, no leading `(`) for braced `if`/`while` heads, expressed with the same higher-order conditional as the existing `ob`/`bl` modes. The grammar is conflict-free under `menhir --strict` with no unused precedence levels.

## Breaking

The moc 1 form with a bare identifier or literal as condition and bare branches no longer parses:

```motoko
// before
if done return null;
if c 1 else 2;
if c[i] else [];           // an index into c
// after: either
if (done) return null;
if (c) 1 else 2;
if (c) [i] else [];
// or
if done { return null };
if c { 1 } else { 2 };
```

Conversely a compound condition may not *start* with `(`, because `if (` opens the moc 1 shape:

```motoko
if (a + b) * c > 0 { };    // before: never valid; now: error at `*`
if ((a + b) * c > 0) { };  // write this, or drop the leading group
```

M0275 diagnoses the first case with both fix-its; its explain page covers both rules. The `x #y` half-spaced concatenation from #6348 is concatenation again.

## What is unchanged

Every moc 1 `if (…)`/`while (…)` with any branch shape, bare case arms, `for (p in e)`, `= e` function bodies, the `??` whitespace rule from [#6358](https://github.com/caffeinelabs/motoko/pull/6358), and the M0272 record-scrutinee diagnostic.

## Follow-up worth flagging

The pinned motoko-core (`Iter.mo`, `Map.mo`, `Set.mo`, `Bool.mo`) uses the removed form at 6 sites, so the 20 tests that import `mo:core` fail here until the core fix lands and the flake input is bumped; a companion motoko-core change that parenthesizes those sites (plus 5 in `pure/Set.mo` and the tests) is ready locally. The in-tree corpus (45 test lines and one prelude line) was updated to `if (c) e`, which keeps each test exercising what it tested before.

Part of [#6344](https://github.com/caffeinelabs/motoko/issues/6344); target syntax in [#6352](https://github.com/caffeinelabs/motoko/issues/6352).


Kamirus:
Closing without merging: the structural rule here obstructs north-star code that master accepts, and unobstructed north-star syntax is the primary goal.

## What we measured

75 head/branch forms through moc 1.16.0, master (#6348, the 2.0.0-beta.0 build) and this branch. Full matrix in the session notes; the rows that differ:

**North-star heads with braced branches**

| form | master | this PR |
|---|---|---|
| `if g (c) { }`, `if xs [0] { }` (space before `(`/`[`) | no | ok |
| `if g(c) { 1 } else if (c) 2 else 3` (mixed chain) | no | ok |
| `if (n + 1) * 2 > 0 { }`, `if (n + 1) == 4 { }`, `if (o).y == 1 { }`, `if (g)(c) { }`, `if (c) and d { }`, `if (n : Nat) > 1 { }`, `while (i) > 0 { }` | ok | **no** |

**Legacy: parenthesized head, bare branches**

| form | master | this PR |
|---|---|---|
| `if (c) -1 else 1`, `if (c) +1 else 1`, `if (c) ^m else m`, `if (c) - 1 else 1` | **no** | ok |
| `if (c)(1) else (2)`, `if (c)[1] else []`, `if (c)#less else #greater`, `if (c)-1 else 1`, `while (i > 0)(i -= 1)` (glued to the `)`) | **no** | ok |

**Legacy: bare identifier or literal head, bare branches**

| form | master | this PR |
|---|---|---|
| `if c 1 else 2`, `if c n else 0`, `if c (1) else (2)`, `if c [1] else []`, `if c #less else #greater`, `if c return 1`, `if c f(1) else 0`, `if c { 1 } else 2`, `if c 1 else { 2 }`, `while c i -= 1`, `if true 1 else 2` | ok | **no** |
| `if c(1) else 2`, `if c[1] else []`, `if c -1 else 1`, `if c-1 else 1`, `if c ^m else m` | **no** | **no** |

**Operand position:** `s #t` is concatenation in 1.16 and here, an error on master.

Everything else is identical in all three compilers, and no form changes meaning silently anywhere: every difference is a hard parse error.

## Why the two designs cannot be combined

After `if (…)` the next token may be `(`, `[`, `{`, `+`, `-`, `^`, `#`, each of which both continues an expression and starts one. The only thing that settles it is the terminator, `{` for a north-star head versus `else`/`;` for a legacy branch, and that is unboundedly far away, so no LR(1) grammar can accept both `if (c) -1 else 1` and `if (n + 1) * 2 > 0 { }`. Master decides at the token with whitespace (`(`/`[`/`#`) and precedence (`EXP_ATOM`, so `-` always continues); this branch decides at the first token of the condition. Each choice obstructs one side. `if c[e] else []` is not recoverable under any bounded-lookahead rule once `if xs[i] { }` is accepted, since even after the matching `]` a following `+` is ambiguous.

## Where the ecosystem sits

- Master's `if (c) -x else y` break: 29 sites in 7 cached mops packages (json, serde, serde-core, xtended-numbers, datetime, rep-indy-hash); zero in motoko-core, the app, calm, skills-internal.
- This branch's bare-head break: 6 sites in motoko-core `src`, 8 in OQL, 63 in-tree test lines.
- The glued forms (`if c[#Clear]`, `if (c)(e)`) have one hit in total, motoko-core's `PriorityQueue.test.mo`, and are accepted as breaking.

Neither #6348 nor this branch was checked against motoko-core's own `test/` directory or the mops cache before the fact; the in-tree tests only parse the core modules they import. A `moc --check` pass over both belongs in the release routine.

## What is still owed on master

The changelog and the M0275 hint claim that parenthesized conditions keep bare branches "exactly as before"; that is false for `if (c) -x`, `if (c) +x`, `if (c) ^x`, and the hint tells the user to parenthesize a condition that already is. Those two texts need fixing regardless. Restoring the prefix-operator branches would need whitespace variants of `+`/`-`/`^` in the lexer (prototyped in this session), at the cost of `if n -1 > 0 { }` reading as a bare branch; whether that is worth it depends on whether any package we ship depends on json, serde, xtended-numbers, or datetime.

The parser simplifications explored here (dropping the third grammar parameter, expressing modes as a higher-order menhir conditional, folding the `exp_head_*` chain) are independent of the rule and can be revisited separately.

